PR #8225, opened 2019-04-19
[FLINK-10984]Remove flink-shaded-hadoop from flink

sunjincheng121:

To allow reasonable dependency management we should move `flink-shaded-hadoop` to `flink-shaded`, then Remove `flink-shaded-hadoop` from flink. 

## Brief change log
  - Remove `flink-shaded-hadoop` from flink.
  - Remove `flink-shaded-hadoop/flink-shaded-hadoop2` from flink.
  - Move out `flink-shaded-yran-test` from `flink-shaded-hadoop`.
  - Modify the scripts in `flink-dist`, i.e. remove the logic about `hadoop`.
  - Modify some pom config about `flink-shaded-hadoop`.

## Verifying this change

  - The serializers: (no)

## Documentation


sunjincheng121:
The CI will turn green after `flink-shaded` releases the 7.0 version.

flink-connectors/flink-connector-filesystem/pom.xml
```diff
@@ -52,8 +52,8 @@ under the License.
 
 		<dependency>
 			<groupId>org.apache.flink</groupId>
-			<artifactId>flink-shaded-hadoop2</artifactId>
-			<version>${hadoop.version}-${project.version}</version>
+			<artifactId>flink-shaded-hadoop-2</artifactId>
+			<version>${hadoop.version}-7.0</version>
```
zentol: use `${shaded.version}`instead of `7.0`.

flink-dist/pom.xml
```diff
@@ -428,9 +411,6 @@ under the License.
 									<goal>single</goal>
 								</goals>
 								<configuration>
-									<descriptors>
-										<descriptor>src/main/assemblies/hadoop.xml</descriptor>
```
zentol: you have to rework this to use the `maven-dependency-plugin` to download the jar beforehand, and then place it into the distribution, see https://github.com/apache/flink/pull/6167 for example on how this can be done.

flink-dist/pom.xml
```diff
@@ -399,23 +399,6 @@ under the License.
...
 			</activation>
-			<dependencies>
-				<!--
-                    The Hadoop 2 Uber jar should not go into the Flink dist jar, but
-                    sit next to it. Hence, we set it to 'provided' here.
-                -->
-
-				<dependency>
-					<groupId>org.apache.flink</groupId>
-					<artifactId>flink-shaded-hadoop2-uber</artifactId>
```
zentol: we want to retain the ability to bundle hadoop, for the time being at least, as such this should stay (without it the entire profile makes no sense anyway)

pom.xml
```diff
@@ -234,7 +233,7 @@ under the License.
 						asm,
 						netty
 					here. It will overwrite Hadoop's guava dependency (even though we handle it
-			separatly in the flink-shaded-hadoop module).
+			separatly in the flink-shaded-hadoop-2).
```
zentol: `...hadoop-2 dependency`

flink-dist/pom.xml
sunjincheng121: Yes, I think we should keep `include-hadoop` available. and we can download the jar by change the `hadoop.xml`. i.e.: `files` ->`dependencySet`.  and `Rework flink-dist assemblies` changes can in a new PR. What do you think?

sunjincheng121:
Thanks for your Review @zentol!

I have had updated the PR accordingly.  And about download the `uber` JRA,  I think your right! we should keep `include-hadoop` available. and we can download the jar by changing the `hadoop.xml`. i.e.: `files` ->`dependencySe`t. and Rework `flink-dist assemblies `changes can in a new PR. What do you think?

Best,
Jincheng

flink-dist/src/main/assemblies/hadoop.xml
```diff
@@ -28,13 +28,14 @@ under the License.
...
-		<file>
-			<source>../flink-shaded-hadoop/flink-shaded-hadoop2-uber/target/flink-shaded-hadoop2-uber-${hadoop.version}-${project.version}.jar</source>
+	<dependencySets>
+		<dependencySet>
 			<outputDirectory>lib/</outputDirectory>
-			<destName>flink-shaded-hadoop2-uber-${project.version}-${hadoop.version}.jar</destName>
-			<fileMode>0644</fileMode>
-		</file>
-	</files>
+			<useTransitiveDependencies>true</useTransitiveDependencies>
```
zentol: this is the default and can be omitted

flink-dist/src/main/assemblies/hadoop.xml
```diff
@@ -28,13 +28,14 @@ under the License.
...
+			<scope>provided</scope>
```
zentol: we could add `<outputFileNameMapping>${artifact.artifactId}-${artifact.baseVersion}.${artifact.extension}</outputFileNameMapping>
`, so that (for SNAPSHOT builds) the jar is always called `flink-shaded-hadoop2-uber-2.4.1-1.9-SNAPSHOT.jar`, and not `flink-shaded-hadoop2-uber-2.4.1-1.9-629462874621478328407-23.jar`

sunjincheng121: I think If we add the dependency for `flink-shaded-hadoop-2-uber` with `<version>${hadoop.version}-${flink.shaded.version}</version>`. we never using the SANAPSHOT JRA. e.g.: `flink-shaded-hadoop-2-uber-2.4.1-7.0.jar`Is there anything I misunderstand?

sunjincheng121:
The flink-shaded 7.0 already released, I'll rebase the code and let the CI passed.

pom.xml
```diff
@@ -121,7 +121,7 @@ under the License.
 		<chill.version>0.7.6</chill.version>
 		<zookeeper.version>3.4.10</zookeeper.version>
 		<curator.version>2.12.0</curator.version>
-		<jackson.version>2.7.9</jackson.version>
+		<jackson.version>2.9.8</jackson.version>
```
zentol: we're mixing things here. Let's open one PR that bumps the flink-shaded version and related dependencies, and migrate to hadoop separately.

sunjincheng121: Agree, So, we merge this PR first, right?

zentol:
could you rebase the PR? I merged the flink-shaded version bump made in #8605.

flink-shaded-yarn-tests/pom.xml
```diff
@@ -146,42 +172,42 @@ under the License.
 							<relocations>
 								<relocation>
 									<pattern>com.google</pattern>
-									<shadedPattern>org.apache.flink.hadoop.shaded.com.google</shadedPattern>
+									<shadedPattern>org.apache.flink.hadoop2.shaded.com.google</shadedPattern>
```
zentol: I don't think we have to modify these.

sunjincheng121: Without this chanage, the test of `YARNHighAvailabilityITCase` cannot get success.

sunjincheng121:
Thanks for the Review @zentol 
I have rebased the code and wait for CI change green.

sunjincheng121:
I appreciate if you can have another look :) @zentol

sunjincheng121:
Hi, @zentol Is there are any other comments from you? :)

flink-shaded-yarn-tests/pom.xml
zentol: I'm currently looking into why that is the case. I'd very much prefer if this module would be independent of whatever hadoop version is in flink-dist.

PR #8229, opened 2019-04-20
[FLINK-12273]The default value of CheckpointRetentionPolicy should be …

Mr-Nineteen:
…RETAIN_ON_FAILURE, otherwise it cannot be recovered according to checkpoint after failure.]

## Brief change log

*(for example:)*

## Verifying this change

*(or)*

*(or)*

*(example:)*

## Documentation

### DIFF
--- a/flink-streaming-java/src/main/java/org/apache/flink/streaming/api/graph/StreamingJobGraphGenerator.java
+++ b/flink-streaming-java/src/main/java/org/apache/flink/streaming/api/graph/StreamingJobGraphGenerator.java
@@ -627,7 +627,7 @@ public class StreamingJobGraphGenerator {
 					CheckpointRetentionPolicy.RETAIN_ON_FAILURE :
 					CheckpointRetentionPolicy.RETAIN_ON_CANCELLATION;
 		} else {
-			retentionAfterTermination = CheckpointRetentionPolicy.NEVER_RETAIN_AFTER_TERMINATION;
+			retentionAfterTermination = CheckpointRetentionPolicy.RETAIN_ON_FAILURE;
 		}
 
 		CheckpointingMode mode = cfg.getCheckpointingMode();


flink-streaming-java/src/main/java/org/apache/flink/streaming/api/graph/StreamingJobGraphGenerator.java
```diff
@@ -627,7 +627,7 @@ private void configureCheckpointing() {
 					CheckpointRetentionPolicy.RETAIN_ON_FAILURE :
 					CheckpointRetentionPolicy.RETAIN_ON_CANCELLATION;
 		} else {
-			retentionAfterTermination = CheckpointRetentionPolicy.NEVER_RETAIN_AFTER_TERMINATION;
+			retentionAfterTermination = CheckpointRetentionPolicy.RETAIN_ON_FAILURE;
```
klion26: Why do you want to change the default value here, instead of set an external checkpoint cleanup?

Mr-Nineteen: Hundreds of flink production applications, most applications do not have set an external checkpoint cleanup. After the application fails, it cannot be recovered according to the checkpoint, and the recovery also brings a lot of trouble, which brings huge losses to the production service. Therefore, the default is RETAIN_ON_FAILURE is more reasonable.

StephanEwen:
I think this may be a good new default, but we need to discuss that on the mailing list first, because it affects existing setups.

We also need to mention this in the release notes for the next release.

StephanEwen:
@Mr-Nineteen Can you start this discussion?

I would suggest to close this PR until the discussion has consensus and open it after that.

Mr-Nineteen:
@StephanEwen 
 Ok.
I think it's important to change this default.If some users don't know much about flink, it's easy to bring huge losses in production.Therefore, flink needs to consider this aspect and also bring a better experience to users.

PR #8236, opened 2019-04-22
[FLINK-12289][Runtime] Fix java doc inconsistencies and style issues in Memory manager

liyafan82:

Fix some bugs and typos in memory manager

## Brief change log

  - Fix the bugs and typos in class MemoryManager
  - Add a test case to verify the change in MemoryManagerTest.testReleaseNull

## Verifying this change

  - When the input argument is null, the method should throw a NullPointerException.
  
  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/memory/MemoryManager.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/memory/MemoryManager.java
@@ -176,7 +176,7 @@ public class MemoryManager {
 	}
 
 	/**
-	 * Checks if the memory manager all memory available.
+	 * Checks if the memory manager's memory is completely available (nothing allocated at the moment).
 	 *
 	 * @return True, if the memory manager is empty and valid, false if it is not empty or corrupted.
 	 */
@@ -328,8 +328,7 @@ public class MemoryManager {
 	 * the memory pool of its type, increasing its available limit for the later allocations.
 	 *
 	 * @param segments The segments to be released.
-	 * @throws NullPointerException Thrown, if the given collection is null.
-	 * @throws IllegalArgumentException Thrown, id the segments are of an incompatible type.
+	 * @throws IllegalArgumentException Thrown, if the segments are of an incompatible type.
 	 */
 	public void release(Collection<MemorySegment> segments) {
 		if (segments == null) {


flink-runtime/src/main/java/org/apache/flink/runtime/memory/MemoryManager.java
```diff
@@ -387,11 +387,11 @@ public void release(MemorySegment segment) {
 	 *
 	 * @param segments The segments to be released.
 	 * @throws NullPointerException Thrown, if the given collection is null.
-	 * @throws IllegalArgumentException Thrown, id the segments are of an incompatible type.
+	 * @throws IllegalArgumentException Thrown, if the segments are of an incompatible type.
 	 */
 	public void release(Collection<MemorySegment> segments) {
 		if (segments == null) {
-			return;
+			throw new NullPointerException();
```
Aitozi: If we  get a `null` to release, why can't we just skip it? I think it's a defensive code not a bug. Or can you post some bad situation about this?

liyafan82: > If we get a `null` to release, why can't we just skip it? I think it's a defensive code not a bug. Or can you post some bad situation about this?

Hi Aitozi, thanks a lot for your comment. The JavaDoc for this method declares "@throws NullPointerException Thrown, if the given collection is null." So we should either throw an NPE, or change the JavaDoc, to make it consistent. 
By referencing other methods in this class, I think it is preferable to make the method throw NPE, because the methods should follow a consistent style.

Aitozi: It's truly inconsistent, but this fix will change the default behaviour of the release operation, and the `release(MemorySegment segment)` also skip the null...  I think may be it need @StephanEwen author of the `MemoryManager.java` to confirm, Thanks.

liyafan82: That sounds reasonable. Thank you for your comments. Hi @StephanEwen, would you please give some comments?

StephanEwen: Unless there is a real bug caused by this in some setup, I would not change this.
Such changes often have unforeseen side effects, so I would avoid doing them for "looks better" or "looks more correct" reasons.

StephanEwen: Let's change the JavaDocs instead to match the implementation.

flink-runtime/src/main/java/org/apache/flink/runtime/memory/MemoryManager.java
```diff
@@ -229,7 +229,7 @@ public boolean isShutdown() {
 	}
 
 	/**
-	 * Checks if the memory manager all memory available.
+	 * Checks if the memory manager's all memory is available.
```
StephanEwen: "Checks if the memory manager's memory is completely available (nothing allocated at the moment)."

flink-runtime/src/main/java/org/apache/flink/runtime/memory/MemoryManager.java
liyafan82: @StephanEwen Sounds reasonable. I have revised the PR accordingly. Please take a look.

flink-runtime/src/main/java/org/apache/flink/runtime/memory/MemoryManager.java
liyafan82: Resolved. Thanks.

liyafan82:
> @liyafan82
> Please, rebase on master, squash the commits and adjust commit, PR and jira issue description to reflect the actual changes. Otherwise, looks good.

Rebased. Please take a look. Thanks.

PR #8242, opened 2019-04-23
[FLINK-6227][network] Introduce the PartitionException for downstream task failure

zhijiangW:

*It is part of FLIP-1.*

*We define a new abstract special exception to indicate the downstream task failure in consuming upstream partition. This special exception is reported to `JobMaster` as a hint for unavailable partition data,  and then `JobMaster` could track the upstream executions to restart for reproducing the partition data.*

*We make current `PartitionNotFoundException` extend `PartitionException`, because it is reasonable to restart producer task when the requested partition is already released.*

## Brief change log

  - *Refactor the previous `PartitionException` to `PartitionUpdateException.*
  - *Define a new abstract `PartitionException`.*
  - *Make current `PartitionNotFoundException` extend `PartitionException`.*
  - *Add some new unit tests for covering the cases of `PartitionNotFoundException`.*

## Verifying this change

  - *The whole process could be verified via existing `TaskExecutorSubmissionTest#testRemotePartitionNotFound` and `TaskExecutorSubmissionTest#testLocalPartitionNotFound`.*
  - *Added some new unit tests to further verify the separate components related to `PartitionNotFoundException`:*
      -  *Add new test for `ResultPartitionManager#createSubpartitionView`.*
      -  *Add new test for `PartitionRequestServerHandler` sending `PartitionNotFoundException`.*
      -  *Add new test for `LocalInputChannel#requestSubpartition`.*
      -  *Add new test for `RemoteInputChannel#retriggerSubpartitionRequest`.*
      -  *Add new test for `SingleInputGate#getNextBufferOrEvent`.*

## Documentation


zhijiangW:
Some thoughts in implementation:

1. The `DataConsumptionException` should be general for both streaming and batch jobs, although we always restart the whole in pipelined mode atm no matter with this `DataConsumptionException`. 

2. We decide to wrap this exception on receiver side, so the sender is not aware of it. The advantages are that it could cover all the cases and reflect the semantics from the aspect of consumer side. This exception is not only involved in the process of `ResultSubpartitionView#getNextBuffer` on sender side, but also involved in the processes of network connection, partition request and data transport.

3. This exception is just a hint for `JobMaster` to decide whether to restart the upstream executions. In future `JobMaster` could also query the partition state from `ShuffleMaster`, then it could consider these two overall factors to make the final decision.

zhijiangW:

zhijiangW:
Thanks for your professional reviews and valuable suggestions @tillrohrmann .

 I think there are mainly two issues to be confirmed:
1 . Whether we need a new special `DataConsumptionException` or why not use existing `PartitionNotFoundException`?
In semantic/conceptional aspect, `PartitionNotFoundException` is actually a subtype of `DataConsumptionException`. In functional aspect they provide the same information currently, so I agree to maintain only one type of exception for simplification.  That means to use `PartitionNotFoundException` directly or rename it for covering more semantics.

2. In which cases we should report this `PartitionNotFoundException` for JobMaster to restart the producer?

Your above mentioned cases are very guidable and I am trying to further specify every case:

-> TaskExecutor no longer reachable
- TE unavailable during requesting sub partition: Executor shutdown/killed would cause exception in establishing network connection or requesting sub partition, which is covered by `RemoteInputChannel#requestSubpartition` in PR.

- TE unavailable during transferring partition data: Executor shutdown/killed would cause network channel inactive, which is also covered by `RemoteInputChannel#onError` in PR.

-> TaskExecutor reachable but result partition not available
It is covered in `RemoteInputChannel#failPartitionRequest` in PR.

-> TaskExecutor reachable, result partition available but sub partition view cannot be created (spilled file has been removed)
The producer would response `ErrorResponse` for requesting sub partition, then the consumer would call `RemoteInputChannel#onError` for covering this case. Because this `ErrorResponse` is fatal error, then it would bring all the remote input channels report `DataConsumptionException`. So maybe it is better to wrap `DataConsumptionException` on producer side in this case to avoid affecting all the remote input channels.

Besides above three cases, there also exists another cause of partition data corrupt/deleted during transferring, which is covered by `PartitionRequestQueue#exceptionCaught` in PR ATM. Then the consumer would receive `ErrorResponse` to call `RemoteInputChannel#onError`. 

In this case I ever considered wrapping `DataConsumptionException` directly on producer side in specific `SpilledSubpartitionView#getNextBuffer` to avoid reusing the general `ErrorResponse`, because you might think some internal network exceptions via `ErrorResponse` do not need to restart the producer side. My previous thought was some network exceptions including environment/hardware issues might also be suitable to restart producer in other executors, otherwise the consumer might still encounter the same problem after restarting to consume data from previous executor. So the `ErrorResponse` with fatal property is easy to be reused to represent the partition problem here. Or we could regard network environment/hardware as separate issue to be solved, then wrap the special `DataConsumptionException` in specific views to distinguish with existing `ErrorResponse`.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/consumer/RemoteInputChannel.java
```diff
@@ -162,10 +164,13 @@ void assignExclusiveSegments(List<MemorySegment> segments) {
 	public void requestSubpartition(int subpartitionIndex) throws IOException, InterruptedException {
 		if (partitionRequestClient == null) {
 			// Create a client and request the partition
-			partitionRequestClient = connectionManager
-				.createPartitionRequestClient(connectionId);
+			try {
+				partitionRequestClient = connectionManager.createPartitionRequestClient(connectionId);
+			} catch (RemoteTransportException ex) {
```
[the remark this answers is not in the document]
zhijiangW: It should check `IOException` here and only wrap for `RemoteTransportException` instance.
For the case of `LocalTransportException` it should not need to restart producer side.

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/consumer/LocalInputChannelTest.java
```diff
@@ -389,25 +401,80 @@ public void testGetNextAfterPartitionReleased() throws Exception {
...
+	/**
+	 * Tests the {@link DataConsumptionException} is wrapped when {@link ResultSubpartitionView#getNextBuffer()}
+	 * throws an exception.
+	 */
+	@Test
+	public void testDataConsumptionExceptionDuringGetNextBuffer() throws Exception {
+		Tuple2<Integer, Integer> backoff = new Tuple2<>(0, 0);
+
+		ResultSubpartitionView view = mock(ResultSubpartitionView.class);
+		when(view.getNextBuffer()).thenThrow(new IOException("Expected test exception"));
```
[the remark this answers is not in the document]
zhijiangW: yes

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/consumer/LocalInputChannel.java
```diff
@@ -127,7 +128,7 @@ void requestSubpartition(int subpartitionIndex) throws IOException, InterruptedE
 					if (increaseBackoff()) {
```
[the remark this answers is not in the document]
zhijiangW: The `IOException` should be wrapped into `DataConsumptionException` here, not only for `PartitionNotFound`.

tillrohrmann:
Hi @zhijiangW, I think we should not restart the producer if the data transfer caused and exception after establishing the connection and looking up the sub partitions. Otherwise we might restart producers even though there has been only a transient network problem. Thus, I'm not sure whether we should react to the `RemoteInputChannel#onError` wrt to restarting producers.

What kind of environment/hardware issues do you have in mind that could cause repeated failures of reading the result data? Ideally, we would eventually realize that we need to reproduce the result data and, hence, restart the producers.

Did I understand you correctly, that the `PartitionNotFoundException` is good enough and, thus, we don't need to introduce a new exception?

My proposal would be the following:

1. Make sure that `PartitionNotFoundException` is always thrown if lookup fails
2. Add tests for `PartitionNotFoundException`
3. If there are problems during data transmission which require the restart of producers then let's introduce a new exception, e.g. `DataTransferException` (being subtype of `DataConsumptionException`, `PartitionNotFoundException` could then also be a subtype of the `DataConsumptionException`)
4. Update the failover strategy to also consider `DataTransferException` for restarting producers (this could also mean that this exception must occur x times before restarting the producer unlike in the `PartitionNotFoundException` case where it happens immediately)

When you update the PR, please also rebase onto the latest master.

zhijiangW:
@tillrohrmann thanks for the further suggestions! I agree with your overall ideas.

->What kind of environment/hardware issues do you have in mind that could cause repeated failures of reading the result data? 

I mean the disk/network hardware problems on producer side which could not be restored in short time. So it is better to restart the producer in another machine. We ever encountered this corner case in production.

-> Did I understand you correctly, that the PartitionNotFoundException is good enough and, thus, we don't need to introduce a new exception?

The information in current `PartitionNotFoundException` is enough for `JobMaster` restarting the producer, but it can not cover all the cases. So I would like to list all the possible cases to confirm with you firstly:

- a. Tcp connection fail: it might be caused by producer TM lost or network hardware issue. We might introduce `DataConnectionException` for this.

- b. PartitionNotFound: `ResultPartition` is released from `ResultPartitionManager` which needs to restart producer immediately.

- c. `ResultSupartition#createReaderView`  throw `IOException: it might be caused by disk file corrupt/deleted for `BlockingResultSubpartition`. It could also be wrapped into existing `PartitionNotFound`.

- d. `BlockingResultSubpartitionView#getNextBuffer` thrown IOException: the reason is the same as above c. `PartitionNotFound` might also be used here.

- e. Network server exception during transferring data: it seems more complicated here. The reason might be caused by producer TM lost, or temporary network problem or server hardware environment issue, etc. The consumer as client might be sensitive via inactive network channel or `ErrorResponse` from server. We could introduce `DataTransferException` for covering all these.

The above {b, c, d} might be determined to restart producer immediately and the current `PartitionNotFound` could be used for covering them.

For the cases of a and e, we might introduce new exceptions to cover them and failover strategy might have different rules for considering them.

So I think there are two options for considering: 

- If we want to cover all {a, b, c, d, e} ATM, it might be necessary to define an abstract `DataConsumptionException` as parent of  above `PartitionNotFoundException`, `DataConnectionException` and `DataTransferException`. 

- Or we only concern on {b, c, d} in the first step (a and e might be done future or in other ways), then the current `PartitionNotFound` is enough and no need new exceptions ATM. 

Both two options make sense for me, so I would like to take your final opinion or you have other options. :)

tillrohrmann:
> I mean the disk/network hardware problems on producer side which could not be restored in short time. So it is better to restart the producer in another machine. We ever encountered this corner case in production.

I guess that this problem won't be solely solved by restarting the producer. Imagine that the restarted producer will be placed on the same machine again. Then it will run into the same problem eventually. Additionally, one would need to blacklist this machine so that these resources are no longer used. Since this is a bigger issue, I would like to keep it out of the scope of this PR and address it with a follow up.

> d. BlockingResultSubpartitionView#getNextBuffer thrown IOException: the reason is the same as above c. PartitionNotFound might also be used here.

I'm not 100% sure that this should throw an `PartitionNotFound` exception. The reason for the `IOException` could maybe be something transient as well (e.g. if you read from a NFS). I would propose that this should be a `DataTransferException` or something similar. If the file were removed, then the next lookup (due to the task restart) of the subpartition should fail with a `PartitionNotFound` exception which would then trigger the restart of the producer.

In general I would like to distinguish `DataTransferException`, which signals a problem after a connection has been established and after the partition has been found, from `PartitionNotFoundException` which signals that a partition could not be found.

I think we are missing another case:

* f. The TM is not reachable because it failed. Hence it is not possible to open a connection to it. After the retry time, this should also fail with a `PartitionNotFoundException` or a `ConnectionTimeoutException` which would be treated like a `PartitionNotFoundException`.

I would propose to start with covering the cases `{b, c, f}` and address the other cases as a follow up because that way we keep the scope narrow without preventing us from covering the other cases later.

zhijiangW:
Yes, the blacklist could solve the hardware problems future.

As for d, you are right that the `IOException` might be caused by multiple factors, then we do not cover it in this PR.

As for f, if I understand correctly, the scenario is similar as the case `a` I mentioned above. 
If the TM is unreachable, the first phase of establishing connection would fail by `ConnectionTimeoutException` or `ConnectionRefusedException` etc. I proposed `DataConnectionException` not `PartitionNotFound` in `a`, because it might be caused by network temporary problem. After retrying the connection later it might be success. 

In other words, from the `ConnectionTimeoutException` or `ConnectionRefusedException` received by consumer, it could not estimate whether the producer TM is really lost or the temporary network issue. 

After the connection established, the second phase is to send logic `PartitionRequest` message, if at this time the producer TM is unreachable to cause any network exceptions, we still can not distinguish whether it is caused by TM really lost or temporary network issue, only if the consumer receives specific `PartitionNotFoundException` via network which is the cases of b, c.

So we might only focus on b, c currently?

tillrohrmann:
I think we actually need to handle case `f` one way or the other. It is not impossible that a `TaskExecutor` completely dies. So we either return a `DataConnectionException` which is handled by the failover strategy or say that after trying to connect to the remote host for some time we fail with a `PartitionNotFoundException`. Of course, it can still be the case that the `TaskExecutor` is reachable afterwards but at some point we simply need to consider the `TaskExecutor` as dead.

zhijiangW:
Yes, I think we are on the same page now.

I would focus on `b, c` in this PR, and launch separate PRs for the other cases future. Especially for `DataConnectionException` we might add the retry mechanism if possible during connecting the server and could throw `PartitionNotFoundException` or `DataConnectionException` after retry fails.

I already updated the codes in two aspects:

-  Not transform the received `PartitionNotFoundException` on consumer side.

- Send the `PartitionNotFoundException` on producer side in the process of creating reader view. If the `ResultPartition` is not removed from `ResultPartitionManager`, only the `SpillableSubpartition#createView` might throw `IOException` which could indicate the required data file not open correctly. So I wrap the `PartitionNotFoundException` in upper layer which seems unified for all the `ResultSubpartition` instances besides the  new `BoundedBlockingSubpartition` Stephan proposed.

After you confirm this way is correct, then I would add new unit tests for covering these cases. :)

zhijiangW:
Yes, I think we are on the same page now.

I would focus on `b, c` in this PR, and launch separate PRs for the other cases future. Especially for `DataConnectionException` we might add the retry mechanism if possible during connecting the server and could throw `PartitionNotFoundException` or `DataConnectionException` after retry fails.

I already updated the codes for the following points:

-  Extend the `PartitionNotFoundException` to add internal `Throwable` information. This exception might be caused by some `IOException` during opening blocking result partition file, so the internal throwable could be used for tracing /debugging the root problem.

- Not transform the received `PartitionNotFoundException` to `IOException` on consumer side.

- Catch parent `IOException` instead of `PartitionNotFoundException` in `LocalInputChannel` or `PartitionRequestServerHandler` during requesting partition. `IOException` could only happen in blocking result partition during opening disk file which might indicate the file corrupted/deleted.  So the above cases of `b, c` are unified in this process.  But there might exist concerns here: 

1.  For the pipelined mode, the consumer might ask JM for partition state to retrigger request after receiving `PartitionNotFoundException`. 
2. For the blocking mode, the consumer should fail directly because the partition state is already known as `FINISHED`. 

So I think we should adjust the logic of `triggerPartitionProducerStateCheck` during receiving `PartitionNotFoundException` on consumer side. Otherwise this PR change might bring unnecessary retry if it is caused by partition corrupted not released.

After you confirm this way is correct, then I would add new unit tests for covering these cases. :)

zhijiangW:
Thanks for the further review @tillrohrmann . Let me try to explain above two concerns.

- Whether to always transform `IOException` to `PartitionNotFoundException`? Actually I would like to do this wrapper in the construtor of `SpilledSubpartitionView` which might cause `IOException` during opening disk file. And based on the current codes, this is the only process which might cause `IOExeption` during creating `ResultSubpartitionView`. But considering the `SpillableSubpartitoin` and its reader view would be removed by stephan's new `BoundedBlockingSubpartition` soon, so I do this wrapper as the way in PR which might also suitable for the new `BoundedBlockingSubpartition`. But in strict way it would   be better to transform in the specific process instead of covering the whole process. If you have concerns on this, I could adjust it to the internal process even though it would be removed soon. Or we do not focus on the case `c` until `BoundedBlockingSubpartition` is merged.

- Whether to check file exist on producer side and throw `PartitionNotFoundException` there?  I think the check could only be done on producer side during opening the file. Maybe there are other options but I have not thought of now. But the producer might not throw `PartitionNotFoundException` after checking file nonexist. Another option is the producer throws another special exception called `PartitionOpenException`, then it would be sent to consumer via `ErrorResponse` in network. The consumer would fail directly and wrap it into `PartitionNotFoundException` for JM. To do so we could avoid the process of checking partition state on consumer side when receiving `PartitionNotFoundException`. But we need another new exception defination. My current way in PR is easy and actually for blocking partition the consumer could also fail directly after receiving `PartitionNotFoundException` as I mentioned before.

I could create a ticket jira for the connection issue later.

The unit tests I have not focused yet, because it is related to the process we want to implement. After you approval the current way, I would fix the existing tests and add new tests to cover the cases.

zhijiangW:
@tillrohrmann thanks for your further great suggestions!

I totally agree the above points and the new exception category would make the semantics more clearly. 

To double confirm If my understanding is correct:

- The new proposed `PartitionException` has three sub-type instances as `PartitionNotFoundException`, `PartitionCorruptionException` and `PartitionLookupException`. The job master would check the `PartitionException` for determining to restart the producer. Especially for instance of `PartitionLookupException`, the producer might need to be restarted in another machine.

- Considering the current `SpilledSubpartitionView` would be replace by Stephan's new `BoundedBlockingSubpartition`, do you think it is necessary to throw `PartitionCorruptionException` in current code or focus on it after `BoundedBlockingSubpartition` is merged?

- The producer could send `PartitionException` if fails to create sub partition view. On consumer side, if the exception is instanceof `PartitionNotFoundException`, it would keep the current logic of retrying to request partition. If instanceof `PartitionCorruptionException`, the consumer could fail directly.

- I already created a separate jira for covering `PartitionLookupException`, and the connection lost after established would not be considered in this issue.

tillrohrmann:
Yes @zhijiangW, this is what I had in mind. 

For your second point, Stephan merged the blocking result partition PR and removed the `SpillableSubpartition`. Hence, we can make it how it's best for the new `BoundedBlockingSubpartition`.

zhijiangW:
Great, I have no other concerns now, then I could continue on this PR. :)

zhijiangW:
@tillrohrmann I have updated the codes for the following issues:

- Refactor the previous existing `PartitionException` to `PartitionUpdateException`

- Introduce the new abstract `PartitionException`

- Make existing `PartitionNotFoundException` extend `PartitionException`

- Although the existing tests of `TaskExecutorSubmissionTest#testRemotePartitionNotFound` and `TaskExecutorSubmissionTest#testLocalPartitionNotFound` could verify the whole process, but I still added some new tests to verify tiny process in separate component.
       -  Adds new test for `ResultPartitionManager#createSubpartitionView` to make sure `ResultPartitionManager` throw `PartitionNotFoundException`.
       - Adds new test for `PartitionRequestServerHandler` to make sure the server would not swallow the `PartitionNotFoundException` thrown by `ResultPartitionManager`.
       - Adds new test for `LocalInputChannel#requestSubpartition` to make sure local channel throw `PartitionNotFoundException`.
       - Adds new test for `RemoteInputChannel#retriggerSubpartitionRequest` to make sure remote channel throw `PartitionNotFoundException`.
       - Adds new test for `SingleInputGate#getNextBufferOrEvent` to make sure `PartitionNotFoundException` would not be swallowed by `SingleInputGate`.
       - I am not sure whether it is necessary to further verify every component between task and `SingleInputGate` in order to confirm the `PartitionNotFoundException` would not be swallowed by any intermediate components, such as `UnionInputGate`, `BarrierBuffer/Tracker`, `Stream(Two)InputProcessor`, various of instances of `StreamTask`.

- `PartitionCorruptionException` would not happen in current `BoundedBlockingSubpartition#createSubpartitionView`, so this PR does not focus on it now.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/PartitionException.java
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * Exception for covering all the scenarios of consuming partition failure
+ * which causes the consumer task failed, and the job master would decide
+ * whether to restart the producer based on this exception.
+ */
+public abstract class PartitionException extends IOException {
+
+	private static final long serialVersionUID = 0L;
+
+	protected final ResultPartitionID partitionId;
```
tillrohrmann: Why not `private`?

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/netty/DummyConnectionManager.java
```diff
@@ -0,0 +1,67 @@
...
+
+import java.io.IOException;
+
+import static org.apache.flink.runtime.io.network.partition.InputChannelTestUtils.STUB_CONNECTION_ID;
+
+/**
+ * A dummy implementation of the {@link ConnectionManager} which is mainly used for tests
+ * to avoid mock {@link PartitionRequestClient}.
+ */
+public class DummyConnectionManager implements ConnectionManager {
```
tillrohrmann: I would suggest to call it `TestingConnectionManager` which is a bit more in line with recent additions.

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/netty/PartitionRequestServerHandlerTest.java
```diff
@@ -0,0 +1,69 @@
...
+
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+/**
+ * Tests for {@link PartitionRequestServerHandler}.
+ */
+public class PartitionRequestServerHandlerTest {
```
tillrohrmann: `extends TestLogger` is missing

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/netty/PartitionRequestServerHandlerTest.java
```diff
@@ -0,0 +1,69 @@
...
+
+	/**
+	 * Tests that {@link PartitionRequestServerHandler} responses {@link ErrorResponse} with wrapped
```
tillrohrmann: typo *responds*

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/netty/PartitionRequestServerHandlerTest.java
```diff
@@ -0,0 +1,69 @@
...
+		final ResultPartitionID partitionId = new ResultPartitionID();
+
+		// Write the message of partition request to server
+		channel.writeInbound(new PartitionRequest(partitionId, 0, new InputChannelID(), 2));
+		channel.runPendingTasks();
+
+		// Read the response message after handling partition request
+		final Object msg = channel.readOutbound();
+
+		assertEquals(msg.getClass(), ErrorResponse.class);
```
tillrohrmann: It's not super critical, but I would suggest to use Hamcrest's Matchers for assertions because they offer a bit more expressive assertion messages. With Hamcrest it could look like: `assertThat(msg, Matchers.instanceOf(ErrorResponse.class))` or `assertThat(msg, Matchers.isA(ErrorResponse.class))`

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/netty/PartitionRequestServerHandlerTest.java
```diff
@@ -0,0 +1,69 @@
...
+
+		final ErrorResponse err = (ErrorResponse) msg;
+
+		assertTrue(err.cause instanceof PartitionNotFoundException);
```
tillrohrmann: Here the same

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/netty/PartitionRequestServerHandlerTest.java
```diff
@@ -0,0 +1,69 @@
...
+		assertEquals(partitionId, ((PartitionNotFoundException) err.cause).getPartitionId());
```
tillrohrmann: With hamcrest this would be `assertThat(partitionId, is(partitionNotFoundException.getPartitionId())`

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/ResultPartitionManagerTest.java
```diff
@@ -0,0 +1,65 @@
...
+
+import java.io.IOException;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.fail;
+
+/**
+ * Tests for {@link ResultPartitionManager}.
+ */
+public class ResultPartitionManagerTest {
```
tillrohrmann: `extends TestLogger` is missing

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/ResultPartitionManagerTest.java
```diff
@@ -0,0 +1,65 @@
...
+	 * successful if this partition was already registered before.
+	 */
+	@Test
+	public void testCreateViewForRegisteredPartition() {
+		final ResultPartitionManager partitionManager = new ResultPartitionManager();
+		final ResultPartition partition = PartitionTestUtils.createPartition();
+		try {
+			partitionManager.registerResultPartition(partition);
+			partitionManager.createSubpartitionView(partition.getPartitionId(), 0, new NoOpBufferAvailablityListener());
+		} catch (IOException e) {
```
tillrohrmann: No need to catch the exception. Let it simply bubble up. This is enough to fail the test.

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/netty/DummyConnectionManager.java
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	@Override
+	public PartitionRequestClient createPartitionRequestClient(ConnectionID connectionId) throws IOException {
+		final CreditBasedPartitionRequestClientHandler handler = new CreditBasedPartitionRequestClientHandler();
+		final EmbeddedChannel channel = new EmbeddedChannel(handler);
+		final PartitionRequestClientFactory factory = new PartitionRequestClientFactory(
+			new NettyClient(NettyTestUtil.createConfig()));
+
+		return new PartitionRequestClient(channel, handler, STUB_CONNECTION_ID, factory);
```
tillrohrmann: Would it make sense to extract the `PartitionRequestClient` interface and have a `NettyPartitionRequestClient` implementation? Then we could also have a `TestingPartitionRequestClient` for which we don't have to instantiate the Netty to run a test.

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/consumer/LocalInputChannelTest.java
```diff
@@ -269,6 +270,24 @@ public void testProducerFailedException() throws Exception {
...
+		final LocalInputChannel localChannel = createLocalInputChannel(inputGate, new ResultPartitionManager());
+
+		try {
+			localChannel.requestSubpartition(0);
+
+			fail("Should throw PartitionNotFoundException while requesting unregistered partition.");
+		} catch (PartitionNotFoundException notFound) {
+			assertEquals(localChannel.getPartitionId(), notFound.getPartitionId());
+		}
+	}
```
tillrohrmann: Maybe we could also check that the correct error is set if there is a small backoff (e.g. `10 ms`) and no partition. This is a slightly different code path.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/PartitionException.java
zhijiangW: I also thought whether to keep private before, and get this property only via `getter` way. 
The `protected` was considered to be referenced directly by subclass if possible. I could change it to `private`.

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/ResultPartitionManagerTest.java
zhijiangW: That is true. I just thought to give a detail failed message. I could remove to make it simple.

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/consumer/LocalInputChannelTest.java
zhijiangW: Yes, you are right, we should add this path.

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/netty/DummyConnectionManager.java
zhijiangW: Yes, I think it makes sense to do so. The `PartitionRequestClient` interface might only need one method of `requestSubpartition` for using by `RemoteInputChannel` atm.

zhijiangW:
@tillrohrmann , I have updated the codes for addressing the above comments.

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/consumer/LocalInputChannelTest.java
```diff
@@ -269,6 +272,73 @@ public void testProducerFailedException() throws Exception {
...
+		final SingleInputGate inputGate = createSingleInputGate(1);
+		final LocalInputChannel localChannel = createLocalInputChannel(inputGate, new ResultPartitionManager());
+
+		final Timer timer = new Timer(true) {
+			@Override
+			public void schedule(TimerTask task, long delay) {
+				task.run();
+
+				try {
+					localChannel.checkError();
```
tillrohrmann: `fail` is missing

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/consumer/LocalInputChannelTest.java
```diff
@@ -269,6 +272,73 @@ public void testProducerFailedException() throws Exception {
...
+	public void testRetriggerPartitionRequestWhilePartitionNotFound() throws Exception {
+		final SingleInputGate inputGate = createSingleInputGate(1);
+		final LocalInputChannel localChannel = createLocalInputChannel(
+			inputGate, new ResultPartitionManager(), 1, 1);
+
+		inputGate.setInputChannel(localChannel.getPartitionId().getPartitionId(), localChannel);
+		localChannel.requestSubpartition(0);
+
+		// The timer should be initialized at the first time of retriggering partition request.
+		assertNotNull(inputGate.getRetriggerLocalRequestTimer());
```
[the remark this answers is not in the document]
zhijiangW: It is actually not asserted explicitly. I just want to verify `SingleInputGate#retriggerPartitionRequest` is triggered via assert non-null timer. And it must satisfy two conditions for making this trigger happen. One is throwing `PartitionNotFoundException`, and the other is within backoff.

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/netty/ClientTransportErrorHandlingTest.java
```diff
@@ -134,12 +133,10 @@ public Void answer(InvocationOnMock invocation) throws Throwable {
 		}).when(rich[1]).onError(isA(LocalTransportException.class));
 
 		// First request is successful
-		ChannelFuture f = requestClient.requestSubpartition(new ResultPartitionID(), 0, rich[0], 0);
-		assertTrue(f.await().isSuccess());
+		requestClient.requestSubpartition(new ResultPartitionID(), 0, rich[0], 0);
 
 		// Second request is *not* successful
-		f = requestClient.requestSubpartition(new ResultPartitionID(), 0, rich[1], 0);
-		assertFalse(f.await().isSuccess());
```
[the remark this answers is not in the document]
zhijiangW: The `rich[1],onError` is actually verified above, then it would call `sync.countDown()`. Otherwise the test would fail in timeout.

In `PartitionRequestClient#requestSubpartition`, the `delayMs` is 0 then it would not schedule `tcpChannel.writeAndFlush(request)`. So I thought it wrote the request in the same thread before. After double checking this, the write is actually done in netty thread, so I would consider the race conditions here.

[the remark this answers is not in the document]
zhijiangW: I move the verify of `rich[0]` behind `sync.await` to confirm the future operations are done. Because the end of `sync.await` indicates the rich[1] is set on error, and the request of rich[0] which is in front of rich[1] should also be done.

zhijiangW:
@tillrohrmann thanks for further reviews, and I already updated the codes to address the comments.

PR #8243, opened 2019-04-23
[FLINK-12299] ExecutionConfig#setAutoWatermarkInterval should check param(interval should not less than zero)

carryxyh:
We should check autowatermarkinterval field.

In any scenario, `autoWatermarkInterval` should not be less than or equal to zero.

First of all, this does not correspond to the meaning of `autoWatermarkInterval`.

Second, in the case where `autoWatermarkInterval` is less than 0, we will not be able to register ourselves in `TimestampsAndPeriodicWatermarksOperator#open`, which will result in the water level of this stream being kept at the lowest level.

## Brief change log
add check to ExecutionConfig#setAutoWatermarkInterval's param

## Verifying this change

no

## Documentation

### DIFF
--- a/flink-core/src/main/java/org/apache/flink/api/common/ExecutionConfig.java
+++ b/flink-core/src/main/java/org/apache/flink/api/common/ExecutionConfig.java
@@ -220,6 +220,7 @@ public class ExecutionConfig implements Serializable, Archiveable<ArchivedExecut
 	 */
 	@PublicEvolving
 	public ExecutionConfig setAutoWatermarkInterval(long interval) {
+		Preconditions.checkArgument(interval >= 0, "Auto watermark interval must not be negative.");
 		this.autoWatermarkInterval = interval;
 		return this;
 	}


carryxyh:
I am sorry to have modified some formats inadvertently. It seems that my idea has done some things myself. But they have no effect, it is not my intention.

flink-core/src/main/java/org/apache/flink/api/common/ExecutionConfig.java
```diff
@@ -220,6 +219,7 @@ public boolean isClosureCleanerEnabled() {
 	 */
 	@PublicEvolving
 	public ExecutionConfig setAutoWatermarkInterval(long interval) {
+		Preconditions.checkArgument(interval > 0, "The interval must be greater than 0.");
```
Matrix42: interval >= 0

carryxyh: Hi, I don't think this can be equal to 0.

In the case of equal to 0, it actually causes the watermark to remain at Long.MIN_VALUE, which is not what we want.

zentol:
Please revert the formatting changes.

carryxyh:
@zentol 
Fix it soon. Thx for your review.

carryxyh:
Hi, I try my best to revert format changes.
But I was failed. I even change this file directly on github.com. But it still tells me "nothing changed for this file" when I just try to add some space.

@zentol

zentol:
I've reverted the formatting changes by manually adding spaces/tabs through the github UI, reverted unrelated changes in `ExecutionConfig#getRegisteredKryoTypes()/registerPojoType()`, and modified the check to allow 0, since `org.apache.flink.streaming.api.environment.StreamExecutionEnvironment#setStreamTimeCharacteristic` sets it to 0. Since 0 is the default value we should also accept 0 as input.

carryxyh:
Very kind of u to help for this.
IMO, `0` is not appropriate for the `EventTime` mode.
Perhaps a correlation check is necessary. For example, in the `EventTime` mode, we need to ensure that `auto watermark` is enabled(means auto watermark should > 0).

How do u think about it?

zentol:
It is true that it's difficult to find a valid use-case for the interval being 0, but the sad reality is that all existing classes handle this case gracefully and there might be someone out there relying on it, possibly even just in tests (like we do).

Hence I'd rather not change that. Will merge the PR as is once travis is green.

carryxyh:
Ok, thank you for your help and reply.

PR #8248, opened 2019-04-24
[FLINK-12309][JDBC] JDBCOoutputFormat and JDBCAppendTableSink float behavior is not align

FaxianZhao:

JDBCOoutputFormat and JDBCAppendTableSink float behavior is not align
When we use JDBCAppendTableSinkBuilder#setParameterTypes() config the jdbc types, it bind BasicTypeInfo#FLOAT_TYPE_INFO to java.sql.Types#FLOAT.
But JDBCOutputFormat#writeRecord will force cast the float value to double type, it will throw ClassCastException like "java.lang.Float cannot be cast to java.lang.Double".

## Verifying this change

  - Added a test case org.apache.flink.api.java.io.jdbc.JDBCOutputFormatTest#testCastFloatToDoubleType

  - The serializers: (don't know)

## Documentation

### DIFF
--- a/flink-connectors/flink-jdbc/src/main/java/org/apache/flink/api/java/io/jdbc/JDBCOutputFormat.java
+++ b/flink-connectors/flink-jdbc/src/main/java/org/apache/flink/api/java/io/jdbc/JDBCOutputFormat.java
@@ -156,9 +156,9 @@ public class JDBCOutputFormat extends RichOutputFormat<Row> {
 									upload.setLong(index + 1, (long) row.getField(index));
 									break;
 								case java.sql.Types.REAL:
+								case java.sql.Types.FLOAT:
 									upload.setFloat(index + 1, (float) row.getField(index));
 									break;
-								case java.sql.Types.FLOAT:
 								case java.sql.Types.DOUBLE:
 									upload.setDouble(index + 1, (double) row.getField(index));
 									break;
--- a/flink-connectors/flink-jdbc/src/test/java/org/apache/flink/api/java/io/jdbc/JDBCOutputFormatTest.java
+++ b/flink-connectors/flink-jdbc/src/test/java/org/apache/flink/api/java/io/jdbc/JDBCOutputFormatTest.java
@@ -133,6 +133,32 @@ public class JDBCOutputFormatTest extends JDBCTestBase {
 		jdbcOutputFormat.writeRecord(row);
 	}
 
+	@Test
+	public void testCastFloatToDoubleType() throws IOException {
+		jdbcOutputFormat = JDBCOutputFormat.buildJDBCOutputFormat()
+			.setDrivername(DRIVER_CLASS)
+			.setDBUrl(DB_URL)
+			.setQuery(String.format(INSERT_TEMPLATE, INPUT_TABLE))
+			.setSqlTypes(new int[] {
+				Types.INTEGER,
+				Types.VARCHAR,
+				Types.VARCHAR,
+				Types.FLOAT,
+				Types.INTEGER})
+			.finish();
+		jdbcOutputFormat.open(0, 1);
+
+		Row row = new Row(5);
+		row.setField(0, 4);
+		row.setField(1, "hello");
+		row.setField(2, "world");
+		row.setField(3, 0.99f);
+		row.setField(4, 1);
+
+		jdbcOutputFormat.writeRecord(row);
+		jdbcOutputFormat.close();
+	}
+
 	@Test(expected = RuntimeException.class)
 	public void testExceptionOnClose() throws IOException {
 


flink-connectors/flink-jdbc/src/test/java/org/apache/flink/api/java/io/jdbc/JDBCOutputFormatTest.java
```diff
@@ -133,6 +133,32 @@ public void testExceptionOnInvalidType() throws IOException {
 		jdbcOutputFormat.writeRecord(row);
 	}
 
+	@Test(expected = RuntimeException.class)
```
walterddr: I am assuming you are trying to catch the `"imthewrongtype"` on the last column. but if the jdbc cast float to double fails. It will throw the same exception and the test will still pass. I suggest let this pass by setting the last field to `INTEGER`

flink-connectors/flink-jdbc/src/test/java/org/apache/flink/api/java/io/jdbc/JDBCOutputFormatTest.java
```diff
@@ -133,6 +133,32 @@ public void testExceptionOnInvalidType() throws IOException {
...
+				Types.FLOAT,
+				Types.VARCHAR})
+			.finish();
+		jdbcOutputFormat.open(0, 1);
+
+		Row row = new Row(5);
+		row.setField(0, 4);
+		row.setField(1, "hello");
+		row.setField(2, "world");
+		row.setField(3, 0.99f); // jdbcOutputFormat will cast float to double
```
walterddr: I usually avoid trailing comments (it's already pretty self-explanatory from your test function name

flink-connectors/flink-jdbc/src/main/java/org/apache/flink/api/java/io/jdbc/JDBCOutputFormat.java
```diff
@@ -160,7 +160,11 @@ public void writeRecord(Row row) throws IOException {
 									break;
 								case java.sql.Types.FLOAT:
 								case java.sql.Types.DOUBLE:
-									upload.setDouble(index + 1, (double) row.getField(index));
+									if (row.getField(index).getClass() == Float.class) {
```
walterddr: I am not sure this is the right approach. shouldn't we do something like:
```
case java.sql.Types.FLOAT:
  upload.setFloat(index + 1, (float) row.getField(index));
  break;
case java.sql.Types.DOUBLE:
  upload.setDouble(index + 1, (double) row.getField(index));
  break;
```
that would make more sense to me since the row is float, the type mapping is also float.

flink-connectors/flink-jdbc/src/test/java/org/apache/flink/api/java/io/jdbc/JDBCOutputFormatTest.java
FaxianZhao: agree.

flink-connectors/flink-jdbc/src/test/java/org/apache/flink/api/java/io/jdbc/JDBCOutputFormatTest.java
FaxianZhao: agree

flink-connectors/flink-jdbc/src/main/java/org/apache/flink/api/java/io/jdbc/JDBCOutputFormat.java
FaxianZhao: agree

afedulov:
Hi @FaxianZhao, I would like to facilitate merging this fix. I know it has been a while, but wanted to ask you if you would still be willing to rebase your changes on the current master. FYI: the code in question moved to `JdbcUtils`.

FaxianZhao:
> Hi @FaxianZhao, I would like to facilitate merging this fix. I know it has been a while, but wanted to ask you if you would still be willing to rebase your changes on the current master. FYI: the code in question moved to `JdbcUtils`.

It seems this bug has fixed. I will close this pr.
https://github.com/apache/flink/blob/66d9188a368dc02aa36255523c6ba7a7d3847f82/flink-connectors/flink-connector-jdbc/src/main/java/org/apache/flink/connector/jdbc/utils/JdbcUtils.java#L109

afedulov:
@FaxianZhao are you sure? Your fix added an explicit cast of `java.sql.Types.FLOAT`to (float). In the current version of the `JdbcUtils` it will fall through the switch caasae still be cast to (double), as in the original ticket.
https://github.com/apache/flink/blob/master/flink-connectors/flink-connector-jdbc/src/main/java/org/apache/flink/connector/jdbc/utils/JdbcUtils.java#L109

FaxianZhao:
@afedulov Yes, you're right. I misunderstand my original code.

FaxianZhao:
@afedulov hello, I re-think this issue.
For the oracle standard. java.sql.Types#FLOAT is better to mapping JDBC Double. 
[](https://docs.oracle.com/javase/1.5.0/docs/guide/jdbc/getstart/mapping.html)

#10731 and #17459  already fix the original mapping issue. So, I think we can keep current behavior.

afedulov:
@FaxianZhao thanks for the clarification. Could you please close the PR?

PR #8251, opened 2019-04-24
[FLINK-12316][jdbc] Relax assertions

zentol:

Relaxes the exception message assertion in `JDBCFullTest#testEnrichedClassCastException` such that it works under java 8 and 9+. The test was failing since the class names for Strings, Doubles etc. are different under Java 9.

The assertion now only checks for the part of the exception that we actually control.


tillrohrmann:
@flinkbot approve all

PR #8250, opened 2019-04-24
[FLINK-12247][runtime] fix NPE when writing the archive json file to FileSystem

lamber-ken:
## What is the purpose of the change

if a flink-job reaches the max attempt count, the flink job will write an archive file to FileSystem and shut down.but when `SubtaskExecutionAttemptDetailsHandler` handle the detail attempt info of subtask, met NEP.

for more detail, please see [FLINK-12247](https://issues.apache.org/jira/browse/FLINK-12247).

## Brief change log

  - `SubtaskExecutionAttemptAccumulatorsHandler` add judgment null condition for `AccessExecution`.
  - `SubtaskExecutionAttemptDetailsHandler` add judgment null condition for `AccessExecution`.

## Verifying this change

This change is already covered by existing tests
- *SubtaskExecutionAttemptAccumulatorsHandlerTest*.
- *SubtaskExecutionAttemptDetailsHandlerTest*.

## Does this pull request potentially affect one of the following parts:

  - Dependencies (does it add or upgrade a dependency): (no)
  - The public API, i.e., is any changed class annotated with `@Public(Evolving)`: (no)
  - The serializers: (no)
  - The runtime per-record code paths (performance sensitive): (yes)
  - Anything that affects deployment or recovery: JobManager (and its components), Checkpointing, Yarn/Mesos, ZooKeeper: (yes)
  - The S3 file system connector: (no)

## Documentation

  - Does this pull request introduce a new feature? (no)
  - If yes, how is the feature documented? (not applicable )

PR #8254, opened 2019-04-24
[FLINK-12219][runtime] Yarn application can't stop when flink job failed in per-job yarn cluste mode

lamber-ken:

In our flink(1.6.3) product env, I often encounter a scene that yarn application can't stop when flink job failed in per-job yarn cluste mode.

When a flink job fail, system will write an archive file to a FileSystem through `MiniDispatcher#archiveExecutionGraph` method, then notify `YarnJobClusterEntrypoint` to shutDown. But, if `MiniDispatcher#archiveExecutionGraph` throw exceptions during execution, it affect the following calls.

So I open [FLINK-12247](https://issues.apache.org/jira/browse/FLINK-12247) to solve NEP bug when system write archive to FileSystem. But We still need to consider other exceptions, so we should catch Exception / Throwable not just IOExcetion.

for more detail, please see [FLINK-12219-JIRA](https://issues.apache.org/jira/browse/FLINK-12219).

## Brief change log

`JsonResponseHistoryServerArchivist#archiveExecutionGraph` change Exception to IOException.

## Verifying this change

This change is already covered by existing tests, such as `TestingDispatcher`.

  - The serializers: (no)

## Documentation


lamber-ken:
@tillrohrmann, thanks for review. Let me think how to provide a right test case.

lamber-ken:
@tillrohrmann, I add a junit case `MiniDispatcherTest#testTerminationAfterJobCompletionWithExceptionHistoryServerArchivist`. At this case, miniDispatcher was in detached mode( like to per-job on yarn) and we will see `JobTerminationFuture` will nerver terminate until timeout.

lamber-ken:
hi @tillrohrmann, a minimal reproducible flink program can simulate this bug quickly.

### ENV
- flink-1.6.3 or up version 
- run job on yarn cluster in detached mode.
- restart strategy
```
restart-strategy: fixed-delay
restart-strategy.fixed-delay.attempts: 20
restart-strategy.fixed-delay.delay: 2 s
```

### Flink Job
```
public static void main(String[] args) throws Exception {

    StreamExecutionEnvironment env = StreamExecutionEnvironment.getExecutionEnvironment();

    DataStream<String> text = env.addSource(new SourceFunction<String>() {
        @Override
        public void run(SourceContext<String> ctx) throws Exception {
            while (true) {
                ctx.collect("aaaa");
                Thread.sleep(100);
            }
        }

        @Override
        public void cancel() {

        }
    });

    text.addSink(new SinkFunction<String>() {
        @Override
        public void invoke(String value, Context context) throws Exception {
            System.out.println(1 / 0);
        }
    });

    env.execute();

}
```

tillrohrmann:
The `HistoryServerArchivist` is not supposed to throw any exceptions. If it does, then it is clearly a programming fault. I don't understand which case you try to solve here. What is the original exception you are seeing why you think that an `IOException` is not good enough.

lamber-ken:
hi, @tillrohrmann 

**First**, the original exception is NEP which as [FLINK-12247](https://github.com/apache/flink/pull/8250) described. When the number of retries more than `MAX_ATTEMPTS_HISTORY_SIZE`, it'll throw NEP. it'll cause yarn application hang.

**Second**, the [FsJobArchivist#archiveJob](https://github.com/apache/flink/blob/6cbc9bf2cd6142c34074deaec56bb339d30b3b93/flink-runtime/src/main/java/org/apache/flink/runtime/history/FsJobArchivist.java#L71) method not only throw `IOException` but also throw `Exception`. Include at least the above exception currently. 
So I used `ExceptionHistoryServerArchivist` to simulate other unknown exceptions which may be happen.


image

lamber-ken:
@tillrohrmann, here is original NPE stacktrace.
```
java.lang.NullPointerException
   at org.apache.flink.runtime.rest.handler.util.MutableIOMetrics.addIOMetrics(MutableIOMetrics.java:88)
   at org.apache.flink.runtime.rest.handler.job.SubtaskExecutionAttemptDetailsHandler.createDetailsInfo(SubtaskExecutionAttemptDetailsHandler.java:140)
   at org.apache.flink.runtime.rest.handler.job.SubtaskExecutionAttemptDetailsHandler.archiveJsonWithPath(SubtaskExecutionAttemptDetailsHandler.java:120)
   at org.apache.flink.runtime.webmonitor.WebMonitorEndpoint.archiveJsonWithPath(WebMonitorEndpoint.java:780)
   at org.apache.flink.runtime.dispatcher.JsonResponseHistoryServerArchivist.archiveExecutionGraph(JsonResponseHistoryServerArchivist.java:57)
   at org.apache.flink.runtime.dispatcher.Dispatcher.archiveExecutionGraph(Dispatcher.java:758)
   at org.apache.flink.runtime.dispatcher.Dispatcher.jobReachedGloballyTerminalState(Dispatcher.java:730)
   at org.apache.flink.runtime.dispatcher.MiniDispatcher.jobReachedGloballyTerminalState(MiniDispatcher.java:138)
   at org.apache.flink.runtime.dispatcher.Dispatcher.lambda$startJobManagerRunner$6(Dispatcher.java:341)
   at java.util.concurrent.CompletableFuture.uniWhenComplete(CompletableFuture.java:760)
   at java.util.concurrent.CompletableFuture$UniWhenComplete.tryFire(CompletableFuture.java:736)
   at java.util.concurrent.CompletableFuture$Completion.run(CompletableFuture.java:442)
   at org.apache.flink.runtime.rpc.akka.AkkaRpcActor.handleRunAsync(AkkaRpcActor.java:332)
   at org.apache.flink.runtime.rpc.akka.AkkaRpcActor.handleRpcMessage(AkkaRpcActor.java:158)
   at org.apache.flink.runtime.rpc.akka.FencedAkkaRpcActor.handleRpcMessage(FencedAkkaRpcActor.java:70)
   at org.apache.flink.runtime.rpc.akka.AkkaRpcActor.onReceive(AkkaRpcActor.java:142)
   at org.apache.flink.runtime.rpc.akka.FencedAkkaRpcActor.onReceive(FencedAkkaRpcActor.java:40)
   at akka.actor.UntypedActor$$anonfun$receive$1.applyOrElse(UntypedActor.scala:165)
   at akka.actor.Actor$class.aroundReceive(Actor.scala:502)
   at akka.actor.UntypedActor.aroundReceive(UntypedActor.scala:95)
   at akka.actor.ActorCell.receiveMessage(ActorCell.scala:526)
   at akka.actor.ActorCell.invoke(ActorCell.scala:495)
   at akka.dispatch.Mailbox.processMailbox(Mailbox.scala:257)
   at akka.dispatch.Mailbox.run(Mailbox.scala:224)
   at akka.dispatch.Mailbox.exec(Mailbox.scala:234)
   at scala.concurrent.forkjoin.ForkJoinTask.doExec(ForkJoinTask.java:260)
   at scala.concurrent.forkjoin.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:1339)
   at scala.concurrent.forkjoin.ForkJoinPool.runWorker(ForkJoinPool.java:1979)
   at scala.concurrent.forkjoin.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:107)
```

lamber-ken:
> The `HistoryServerArchivist` is not supposed to throw any exceptions. If it does, then it is clearly a programming fault. I don't understand which case you try to solve here. What is the original exception you are seeing why you think that an `IOException` is not good enough.

you're right, may be it's a clearly a programming falut. but once an exception occurs, it will cause the yarn application hang. Because this affects the execution of the method `ClusterEntrypoint#shutDownAsync` which will notify yarn resourcemanager to stop applicaiton.

lamber-ken:
@tillrohrmann, BTW, if you want to get the NPE stacktrace, you need to use log4j and change `IOException` to `Exception`. than use command `yarn logs -applicationId xxx`.
```
@Override
public CompletableFuture<Acknowledge> archiveExecutionGraph(AccessExecutionGraph executionGraph) {
   try {
      FsJobArchivist.archiveJob(archivePath, executionGraph.getJobID(), jsonArchivist.archiveJsonWithPath(executionGraph));
      return CompletableFuture.completedFuture(Acknowledge.get());
   } catch (Exception e) {
      LOG.error("archive exception",e);
      return FutureUtils.completedExceptionally(e);
   }
}
```

lamber-ken:
@tillrohrmann, I have a new idea to fix the bug. The most essential reason is when the `Dispatcher#jobReachedGloballyTerminalState` method throw exception witch we don't catched. it affect jobterminal callback. we can catch exception and move callback to finally code block.

### Current code
`MiniDispatcher#jobReachedGloballyTerminalState`
```
protected void jobReachedGloballyTerminalState(ArchivedExecutionGraph archivedExecutionGraph) {

    super.jobReachedGloballyTerminalState(archivedExecutionGraph);

    if (executionMode == ClusterEntrypoint.ExecutionMode.DETACHED) {
        // shut down since we don't have to wait for the execution result retrieval
        jobTerminationFuture.complete(ApplicationStatus.fromJobStatus(archivedExecutionGraph.getState()));
    }
}
```
### Change to new code
```
protected void jobReachedGloballyTerminalState(ArchivedExecutionGraph archivedExecutionGraph) {

    try {
        super.jobReachedGloballyTerminalState(archivedExecutionGraph);
    } catch (Exception e) {
        log.error("jobReachedGloballyTerminalState exception", e);
    } finally {
        if (executionMode == ClusterEntrypoint.ExecutionMode.DETACHED) {
            // shut down since we don't have to wait for the execution result retrieval
            jobTerminationFuture.complete(ApplicationStatus.fromJobStatus(archivedExecutionGraph.getState()));
        }
    }
}

```

tillrohrmann:
Hi @lamber-ken, I think the inner `Exception` is not a problem because it is actually only a `IOException`. Thus, the `archiveJob` method can only throw an `IOException`.

The NPE is clearly a Flink bug and as such we should rather fix the bug than guard against it. What we could do, though, is to run `jobTerminationFuture.complete(ApplicationStatus.fromJobStatus(archivedExecutionGraph.getState()));` in a `finally` block. But we should not catch the exception there. Per contract `jobReachedGloballyTerminalState` should not throw any exception. If it does, then it violates the contract and Flink gives no guarantees how it behaves afterwards.

lamber-ken:
> Hi @lamber-ken, I think the inner `Exception` is not a problem because it is actually only a `IOException`. Thus, the `archiveJob` method can only throw an `IOException`.
> 
> The NPE is clearly a Flink bug and as such we should rather fix the bug than guard against it. What we could do, though, is to run `jobTerminationFuture.complete(ApplicationStatus.fromJobStatus(archivedExecutionGraph.getState()));` in a `finally` block. But we should not catch the exception there. Per contract `jobReachedGloballyTerminalState` should not throw any exception. If it does, then it violates the contract and Flink gives no guarantees how it behaves afterwards.

Hi, @tillrohrmann, you are right! I had move `jobTerminationFuture.complete(ApplicationStatus.fromJobStatus(archivedExecutionGraph.getState()))` to a finally block, and log the exception for analysis, then throw it continue.

lamber-ken:
hi, @tillrohrmann, I had move jobTerminationFuture.complete(ApplicationStatus.fromJobStatus(archivedExecutionGraph.getState())) to a finally block, and log the exception for analysis, then throw it continue.

lamber-ken:
Thanks for your review and suggestion @tillrohrmann. I had update the patch, can you take some time to check?

lamber-ken:
> I don't think that this solves the problem @lamber-ken. Please provide a test case which verifies it.

lamber-ken:
> I don't think that this solves the problem @lamber-ken. Please provide a test case which verifies it.

done.

tillrohrmann:
Thanks for updating the PR @lamber-ken. I actually think that we cannot guard against all possible unchecked exceptions in Flink. If such an exception occurs, then this is clearly an indicator that there is a bug in Flink. If there is a bug in Flink, then we don't give guarantees on the its behaviour. Thus, I think we don't need to add this `finally` block. The problem should have been solved by fixing #8250. Hence, I will close this PR and the corresponding JIRA issue. Please re-open if you strongly object.

lamber-ken:
Thanks for your comment @tillrohrmann, as another issue [FLINK-12183](https://issues.apache.org/jira/browse/FLINK-12183) says, it's hard to find that because here it only catches IOException, and we can't get any stacktrace information from the log file (because there's no place to record the exception log). If some uncatched exception happens in the futher, it'll cause yarn application can't stop again even if the task has failed. BTW, I have no right to re-open this issue again.

Best.

tillrohrmann:
FLINK-12183 is the same issue as FLINK-12247. If the problem is debuggability, then we should think about adding a `UncaughtExceptionHandler` to log these exceptions. This should be the proper fix for this problem, because it can potentially also happen in other places. If you want to re-open the issue, then please ping me or open a new issue.

lamber-ken:
hi @tillrohrmann, yes, at least we need to record the exception record, so I log the exception and re-throw it again.
```
protected void jobReachedGloballyTerminalState(ArchivedExecutionGraph archivedExecutionGraph) {

    try {
        super.jobReachedGloballyTerminalState(archivedExecutionGraph);
    } catch (Exception e) {
        log.error("jobReachedGloballyTerminalState exception", e);
        throw e;
    } finally {
        if (executionMode == ClusterEntrypoint.ExecutionMode.DETACHED) {
            // shut down since we don't have to wait for the execution result retrieval
            jobTerminationFuture.complete(ApplicationStatus.fromJobStatus(archivedExecutionGraph.getState()));
        }
    }
}
```

lamber-ken:
@tillrohrmann, no button to click to re-open, just a `Comment` button.
image

lamber-ken:
The `super.jobReachedGloballyTerminalState(archivedExecutionGraph);` method need to call `WebMonitorEndpoint#archiveJsonWithPath` to acquire archive jsons, in that method, will call many archivingHandlers. We can't guarantee that there is no problem in these handles in the futher. Although it's may be a minor bug for these handlers, but this will cause the yarn application can't stop. 

For the user, when open the flink-ui page, we only see the fink-job fail, but it is not clear why the yarn-application does not stop in per-job detached mode.

```
protected void jobReachedGloballyTerminalState(ArchivedExecutionGraph archivedExecutionGraph) {

    super.jobReachedGloballyTerminalState(archivedExecutionGraph);

    if (executionMode == ClusterEntrypoint.ExecutionMode.DETACHED) {
        // shut down since we don't have to wait for the execution result retrieval
        jobTerminationFuture.complete(ApplicationStatus.fromJobStatus(archivedExecutionGraph.getState()));
    }
}

```

tillrohrmann:
As I've said before, the archiving handlers belong to Flink and if they are broken, then the system does not give any guarantees. If they were user provided, then I would agree that Flink should not fail in case of exceptions because they are not under its control.

The problem is that we don't see this exception. The reason seems to be that we execute the `jobReachedGloballyTerminalState` in a `whenCompleteAsync` callback which returns a future containing the exception. Since we don't check this future, we will never see it. That's also why the `UncaughtExceptionHandler` is never called.

I would suggest to add a utility `FutureUtils#assertNoException` which checks returned futures for normal completion. Please take a look at this commit https://github.com/tillrohrmann/flink/commit/524ead44f68ff02d52e8ff561c9e235e5b353123 and let me know whether you would accept this as a fix for the problem.

lamber-ken:
Your solution is fantastic @tillrohrmann! 
I known the reason，but I'm sorry that I didn't describe it well. `FutureUtils#assertNoException` is a good tool so solve this problem.
My point is when flink job failed on yarn in detach mode，the yarn application should failed too.

lamber-ken:
but, another problem is I can't re-open this issue again

tillrohrmann:
It is true that Yarn could try to restart the cluster depending on the configuration. Flink would then try to re-execute the job again since it is part of the Yarn application (no problem with the submitted job graph store). Depending on whether the Flink bug is transient or not the failure would happen again until all restarts are depleted or eventually the job will succeed (potentially with producing duplicate results).

Given that this is caused by a bug in Flink, I think it is ok to say that we don't give hard guarantees in this case. The important bit is that we report the problem.

I would open a PR based on my commit to add this utility.

lamber-ken:
ok，I agree with you

lamber-ken:
I opened a issuse before, it describe a scene that containts two steps.
first，the flink job failed and the yarn application don't stop. 
second，the yarn resourcemanager restarted for some reason, it will attempt to restart all applications. the job in the first step also restarted, but no jobs recover because the zk meta was removed by first step. it'll failed. but the finalState status is UNDEFIEND.

these two issues are associated，if you have time，help me review it, thanks.

tillrohrmann:
FYI @lamber-ken I've opened the PR with the utility #8334

lamber-ken:
hi，@tillrohrmann，what I need to do next? need I update this pr when #8334 merged?

tillrohrmann:
No, this PR has been closed and should no longer be relevant, right?

lamber-ken:
right, but when and how to fix this issue when #8334 merged? need to open a new pr?

tillrohrmann:
which issue do you wanna fix? That the Yarn application can't stop? I think this one should be fixed with #8250.

lamber-ken:
hi, @tillrohrmann, what this commit [tillrohrmann@524ead4](https://github.com/tillrohrmann/flink/commit/524ead44f68ff02d52e8ff561c9e235e5b353123) mean?
If you insist that it has been resolved with #8250, I don't agree with you very much, but I can only obey you.

As another issue FLINK-12183 says, it's hard to find that because here it only catches IOException, and we can't get any stacktrace information from the log file (because there's no place to record the exception log). If some uncatched exception happens in the futher, it'll cause yarn application can't stop again even if the task has failed.

Best.

lamber-ken:
hi, @tillrohrmann, I think your fix is fantastic [tillrohrmann@524ead4](https://github.com/tillrohrmann/flink/commit/524ead44f68ff02d52e8ff561c9e235e5b353123).

like the [FLINK-12247](https://github.com/apache/flink/pull/8250), We used flink from 1.2.1 to 1.6.3 and we have been puzzled by this problem for a long time. Because it‘s needs the count of flink job failed attempt up to the max attempt history size, and no error log information that we can find the reason why the flink job failed but the yarn application don't stop in detach mode. 

May be other users had met the same bug, but they did not mention it or fix it. So I think it's better to log the error information and stop the application like your solution [tillrohrmann@524ead4](https://github.com/tillrohrmann/flink/commit/524ead44f68ff02d52e8ff561c9e235e5b353123).

If don't do that, any unexpected exception happens, the yarn application will can not stop again. It will need users to find out the reason, it's hard to users.  

Best.

lamber-ken:
hi, @tillrohrmann,

Sorry, it's ok, I just think you just add a future tool just now. you had fix the issue #8334 too.

Best.

tillrohrmann:
The commit message and the PR title were a bit misleading. I've updated them.

lamber-ken:
ok

PR #8258, opened 2019-04-24
[FLINK-12297]Harden ClosureCleaner to handle the wrapped function

Aitozi:

ClosureCleaner is used to clean the implicit `this$x` point to the outer class in the user defined function to reduce the case of serialization function failed.

But if the function is wrapped by user or the flink framework, it will escape the clean operation. This pr harden the ClosureCleaner to handle this. And I think it is also a feasible approach for [[FLINK-12113]](https://issues.apache.org/jira/browse/FLINK-12113)

## Brief change log

check the field if need to further clean and apply

## Verifying this change

1. Add the case in the issue to confirm
2. Add the wrapFunction test case in ClosuerCleanerTest


Aitozi:
close for the failed travis ci and will reopen then

PR #8259, opened 2019-04-24
[FLINK-12325][metrics] Fix bug in statsd exporter when using negative values

Xeli:

Fix a bug in the statsd reporter. With this PR flink is able to correctly send negative metric values to statsd. See FLINK-12325 for the exact details what is going wrong.

## Brief change log
- * Change in StatsdReporter class. When sending negative values to statsd first reset the metric to 0

## Verifying this change

This change added an unit tests.

  - The serializers: no

## Documentation

### DIFF
--- a/flink-metrics/flink-metrics-statsd/src/main/java/org/apache/flink/metrics/statsd/StatsDReporter.java
+++ b/flink-metrics/flink-metrics-statsd/src/main/java/org/apache/flink/metrics/statsd/StatsDReporter.java
@@ -127,14 +127,20 @@ public class StatsDReporter extends AbstractReporter implements Scheduled {
 	// ------------------------------------------------------------------------
 
 	private void reportCounter(final String name, final Counter counter) {
-		send(name, String.valueOf(counter.getCount()));
+		send(name, counter.getCount());
 	}
 
 	private void reportGauge(final String name, final Gauge<?> gauge) {
 		Object value = gauge.getValue();
-		if (value != null) {
-			send(name, value.toString());
+		if (value == null) {
+			return;
 		}
+
+		if (value instanceof Number) {
+			send(numberIsNegative((Number) value), name, value.toString());
+		}
+
+		send(name, value.toString());
 	}
 
 	private void reportHistogram(final String name, final Histogram histogram) {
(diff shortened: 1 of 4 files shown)


zentol:
@flinkbot approve architecture

flink-metrics/flink-metrics-statsd/src/test/java/org/apache/flink/metrics/statsd/StatsDReporterTest.java
```diff
@@ -262,6 +262,69 @@ public void testStatsDMetersReporting() throws Exception {
...
+			int port = receiver.getPort();
+
+			MetricConfig config = new MetricConfig();
+			config.setProperty(ConfigConstants.METRICS_REPORTER_INTERVAL_SUFFIX, "1 SECONDS");
+			config.setProperty("host", "localhost");
+			config.setProperty("port", String.valueOf(port));
+
+			registry = new MetricRegistryImpl(
+				MetricRegistryConfiguration.defaultMetricRegistryConfiguration(),
+				Collections.singletonList(ReporterSetup.forReporter("test", config, new StatsDReporter())));
```
zentol: You can skip the creation of the registry. You can extract the reporter from the ReporterSetup and call report explicitly.

flink-metrics/flink-metrics-statsd/src/test/java/org/apache/flink/metrics/statsd/StatsDReporterTest.java
```diff
@@ -262,6 +262,69 @@ public void testStatsDMetersReporting() throws Exception {
...
+			TaskManagerMetricGroup metricGroup = new TaskManagerMetricGroup(registry, "localhost", "tmId");
+			TestMeter meter = new TestMeter(-50, -5.3);
```
zentol: should also test other metric types

flink-metrics/flink-metrics-statsd/src/test/java/org/apache/flink/metrics/statsd/StatsDReporterTest.java
Xeli: Done, it's a lot cleaner now. I've adjusted the other unit tests in the class like this as well

flink-metrics/flink-metrics-statsd/src/test/java/org/apache/flink/metrics/statsd/StatsDReporterTest.java
Xeli: Done

Xeli:
Hi @zentol,

I've added Gauge support by checking if the type is (a subclass of) `Numeric`.

You make a good point about the risk of sending 2 udp packages. I'm seeing some mixed documentation about how we might solve this. Most do not mention anything about sending multiple metrics in a single udp package at all. One explicitly states newlines can be used, while another states only one metric per udp can be used.

I've tried a couple of implementations' docker images to see if it would work with the new lines. One seemed to support it, but most did not properly.

I think a proper way of supporting this is to add the ability of statsd through tcp, and keep this udp implementation as a sort of best-effort. What do you think?

zentol:
hmm, that's unfortunate. Let's use it as a best-effort implementation then.

We can look into adding TCP support in a separate issue.

flink-metrics/flink-metrics-statsd/src/main/java/org/apache/flink/metrics/statsd/StatsDReporter.java
```diff
@@ -216,4 +238,14 @@ public String filterCharacters(String input) {
 
 		return chars == null ? input : new String(chars, 0, pos);
 	}
+
+	private boolean numberIsNegative(Number input) {
+		try {
+			return new BigDecimal(input.toString()).compareTo(BigDecimal.ZERO) == -1;
```
zentol: I'm wondering whether we could use `return Double.compare(input.doubleValue(), 0) < 0` instead. This should be sufficient for our use-cases, and we get around the repeated convert-to-string -> parse-from-string operations.

Xeli: Yeah good call, in this case the rounding that doubleValue might do wouldn't matter anyhow. Changed it!

flink-metrics/flink-metrics-statsd/src/main/java/org/apache/flink/metrics/statsd/StatsDReporter.java
```diff
@@ -240,12 +240,6 @@ public String filterCharacters(String input) {
...
 
 	private boolean numberIsNegative(Number input) {
-		try {
-			return new BigDecimal(input.toString()).compareTo(BigDecimal.ZERO) == -1;
-		} catch (Exception e) {
-			//not all Number's can be converted to a BigDecimal, such as Infinity or NaN
-			//in this case we just say it isn't a negative number
-			return false;
-		}
+		return input.doubleValue() < 0;
```
zentol: Is there a reason you went for the simple comparison instead of the suggested `Double#compareTo`? The javadocs list a number of special cases that are handled differently:

```
     * Compares two {@code Double} objects numerically.  There
     * are two ways in which comparisons performed by this method
     * differ from those performed by the Java language numerical
     * comparison operators ({@code <, <=, ==, >=, >})
     * when applied to primitive {@code double} values:
     *          {@code Double.NaN} is considered by this method 
     *               to be equal to itself and greater than all other
     *               {@code double} values (including
     *               {@code Double.POSITIVE_INFINITY}).
     *          {@code 0.0d} is considered by this method to be greater
     *               than {@code -0.0d}.
```

Xeli: Ah sorry, no reason! 

I've changed it to Double.compare(), I've also changed line #190 to use this method so the double check there works in the same way.

PR #8263, opened 2019-04-25
[FLINK-12296][StateBackend]Data loss silently in RocksDBStateBackend because of local directory collision

klion26:
Fix the data loss silently problem in RocksDBStateBackend when more than one stateful operators chained in a single task.

When preparing the local directory we'll delete the local directory if it already exists.

Currently, if more than one stateful operators chained in a single task, they'll share the same local directory path, then the local directory will be deleted unexpected, and we'll get data loss silently.

## Brief change log

Change the local data path from
`/local_state_root/allocatio_id/job_id/jobvertext_id_subtask_id/chk_id/rocksdb`
to
`/local_state_root/allocatio_id/job_id/jobvertext_id_subtask_id/chk_id/operatorIdentifier`

## Verifying this change

- `StatefulOperatorChainedTaskTest#testRestore()`

  - The serializers: (**no**)
  - Anything that affects deployment or recovery: JobManager (and its components), Checkpointing, Yarn/Mesos, ZooKeeper: (**yes**)

## Documentation


klion26:
@flinkbot attention @StefanRRichter

link3280:
@klion26 Thanks a lot for fixing this problem! It seems that all the statebackends are affected, I think it would be good to reflect that in the description.

flink-runtime/src/main/java/org/apache/flink/runtime/util/OperatorSubtaskDescriptionText.java
```diff
@@ -29,6 +32,9 @@
 	/** Cached description result. */
 	private final String description;
 
+	/** Pattern used to extract operatorId. */
+	public static final Pattern EXTRACT_OPERATOR_PATTERN = Pattern.compile("^.*_([0-9a-f]+)_\\([1-9]+/[1-9]+\\)$");
```
link3280: Why zero is not taken into account when matching the subtask index and task number?

flink-state-backends/flink-statebackend-rocksdb/src/main/java/org/apache/flink/contrib/streaming/state/snapshot/RocksIncrementalSnapshotStrategy.java
```diff
@@ -182,7 +186,14 @@ private SnapshotDirectory prepareLocalSnapshotDirectory(long checkpointId) throw
...
+			String operatorId = OperatorSubtaskDescriptionText.getOperatorIdByOperatorIdentifier(operatorIdentifier);
+			if (operatorId == null) {
+				operatorId = operatorIdentifier;
+				LOG.warn("Can not extract operatorId from {}, the expected format is {}.",
+					operatorIdentifier,
+					EXTRACT_OPERATOR_PATTERN.toString());
+			}
+			File directory = directoryProvider.subtaskSpecificCheckpointDirectory(checkpointId, operatorId);
 
 			if (directory.exists()) {
```
StefanRRichter: As commented on the JIRA already, instead of applying all the changes in the PR, wouln't it be sufficient to remove this deletion an existing root directory of the checkpoint? This looks like the actual cause of the problem to me. What do you think? We could check and delete an existing subdirectory for the particular operator instead.

flink-runtime/src/main/java/org/apache/flink/runtime/util/OperatorSubtaskDescriptionText.java
klion26: Thanks for your review, the implementation has been changed.

klion26:
@StefanRRichter as discussed offline, I've just adjusted the change and added a test

klion26:
Rebased and squashed the commits to change the commit history according to the change

flink-state-backends/flink-statebackend-rocksdb/src/main/java/org/apache/flink/contrib/streaming/state/snapshot/RocksIncrementalSnapshotStrategy.java
```diff
@@ -184,17 +189,19 @@ private SnapshotDirectory prepareLocalSnapshotDirectory(long checkpointId) throw
...
-			if (!directory.mkdirs()) {
+			if (!directory.exists() && !directory.mkdirs()) {
 				throw new IOException("Local state base directory for checkpoint " + checkpointId +
 					" already exists: " + directory);
 			}
 
 			// introduces an extra directory because RocksDB wants a non-existing directory for native checkpoints.
-			File rdbSnapshotDir = new File(directory, "rocks_db");
+			// append operatorIdentifier here to solve directory collision problem when two stateful operators chained in one task.
+			String subDir = operatorIdentifier.replaceAll("[^a-zA-Z0-9\\-]", "_");
```
StefanRRichter: I would suggest to move the string sanatizing into a separate method and we could use a precompiled regex that we can reuse every time over `replaceAll`.

StefanRRichter: Thinking about it a bit more, I wonder if it is a good idea to the the whole operator identifier string. It might be better to just use the operator id.

Pros:
- No need to handle illegal characters.
- The description string can become quiete length. Given that the dir name is also already long, we can avoid running into any cases that paths go beyond the limits of certain filessystems.

Cons:
- This string is a bit more descriptive.

What do you think?

klion26: I think the Pros make sense to me.

If we use the operator id as the directory name, we need to extract the operator id from operator identifier text, in this case, I want to add an extract method in `OperatorSubtaskDescriptionText`.

In the other hand, I found there is a `backendUID` in `RocksIncrementalSnapshotStrategy`, How about using 'backendUID' as the directory name? If we think the `backendUID`(an UUID.randomUUID()) is too long, we can erasure all the '-' character.

What do you think?
```
operatorID.toString()         ---> d1c4061a511396721732590877b90cc6
UUID.randomUUID().toString()  ---> 3dff9490-2957-431b-a682-c7cbb995e872
after erasuring '-'           ---> 3dff94902957431ba682c7cbb995e872
```

StefanRRichter: I would have suggested to pass the operator id into the backend instead of string operations, but you are right, using the backend UUID can also be used to fix the problem. The only downside is that we can not conclude anymore about which opertor the directory belongs to. But that might be ok.

StefanRRichter:
Thanks @klion26! LGTM 👍 Merging.

klion26:
@StefanRRichter thanks for the review.
I have another question, do I need to cherry-pick the merged commit to release-1.6 and release-1.7, or just the release manager will do this?

StefanRRichter:
@klion26 I will do this. If there is need for a bigger rebase I might get back to you.

klion26:
@StefanRRichter ok, thanks

StefanRRichter:
@klion26 I just tried to cherry-pick the changes. It worked easily for 1.8 but we need a larger rebase for 1.7 because of some refactorings. Can you create a separate PR for the rebase of this to 1.7?

klion26:
@StefanRRichter ok, I'll create a separate PR for 1.7.

PR #8268, opened 2019-04-25
[FLINK-12246][runtime] Fix can't get max_attempts_history_size value when create ExecutionVertex

lamber-ken:

fix can't get `MAX_ATTEMPTS_HISTORY_SIZE` value when create ExecutionVertex.

## Brief change log

the `MAX_ATTEMPTS_HISTORY_SIZE` read from the jobManagerConfig in the ExecutionGraph and then passed to the ExecutionJobVertex.

## Verifying this change

This change is already covered by existing tests, `ExecutionGraphDeploymentTest`.

  - The serializers: (no)

## Documentation

  - If yes, how is the feature documented? (not applicable )

### DIFF
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/ExecutionGraph.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/ExecutionGraph.java
@@ -26,6 +26,7 @@ import org.apache.flink.api.common.accumulators.Accumulator;
 import org.apache.flink.api.common.accumulators.AccumulatorHelper;
 import org.apache.flink.api.common.time.Time;
 import org.apache.flink.configuration.Configuration;
+import org.apache.flink.configuration.JobManagerOptions;
 import org.apache.flink.runtime.JobException;
 import org.apache.flink.runtime.accumulators.AccumulatorSnapshot;
 import org.apache.flink.runtime.accumulators.StringifiedAccumulatorResult;
@@ -255,6 +256,9 @@ public class ExecutionGraph implements AccessExecutionGraph {
 	 * from results than need to be materialized. */
 	private ScheduleMode scheduleMode = ScheduleMode.LAZY_FROM_SOURCES;
 
+	/** The maximum number of prior execution attempts kept in history. */
+	private final int maxPriorAttemptsHistoryLength;
+
 	// ------ Execution status and progress. These values are volatile, and accessed under the lock -------
 
 	private final AtomicInteger verticesFinished;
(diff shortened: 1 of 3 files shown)


flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/ExecutionGraph.java
```diff
@@ -816,6 +817,9 @@ public void attachJobGraph(List<JobVertex> topologiallySorted) throws JobExcepti
 
 		final ArrayList<ExecutionJobVertex> newExecJobVertices = new ArrayList<>(topologiallySorted.size());
 		final long createTimestamp = System.currentTimeMillis();
+		int maxPriorAttemptsHistoryLength = jobInformation.getJobConfiguration().getInteger(
```
tillrohrmann: This is also the job's configuration and not the cluster configuration you are using here.

lamber-ken: @tillrohrmann, I not only use run junit tests to validate, but also run a single flink job to verify that it is correct today. May be wrong somewhere, I'll check it as your suggestion. thanks.

lamber-ken: hi @tillrohrmann, we need to discuss about the `maxPriorAttemptsHistoryLength` option where should we get it? 

**First,**, the option is used for `ExecutionJobVertex`, so it should be extracted from job's configuration. The cluster configuration is like `jobmanager.rpc.address`, `jobmanager.rpc.port` and so on.

**Second**, the key of the option is `jobmanager.execution.attempts-history-size`, I think it's better to be changed to `job.execution.attempts-history-size`. because the option is not used for `jobmanager`

tillrohrmann: I think the maximum prior attempts history length is a cluster configuration option because it affects cluster operations. Therefore, it should come from the cluster's configuration which is accessible when you construct the `ExecutionGraph` in the `ExecutionGraphBuilder`.

tillrohrmann: I would not change the name of the config option atm.

lamber-ken: @tillrohrmann, thanks for your detail explain. I'll think about it again.

lamber-ken: @tillrohrmann, hi, I have question that `ExecutionGraphBuilder.buildGraph` was called from `JobMaster`, but there is only one `jobMasterConfiguration` in JobMaster class. Can jobMasterConfiguration represent ClusterConfigration?

tillrohrmann: yes it does

lamber-ken: > yes it does

done.

lamber-ken:
Thanks for your review and suggestion @tillrohrmann. I had update the patch.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/ExecutionGraph.java
```diff
@@ -255,6 +256,9 @@
 	 * from results than need to be materialized. */
 	private ScheduleMode scheduleMode = ScheduleMode.LAZY_FROM_SOURCES;
 
+	/** The maximum number of prior execution attempts kept in history. */
+	private int maxPriorAttemptsHistoryLength = JobManagerOptions.MAX_ATTEMPTS_HISTORY_SIZE.defaultValue();
```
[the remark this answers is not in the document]
lamber-ken: ok, I see, I'll update it later.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/ExecutionGraph.java
```diff
@@ -464,6 +468,10 @@ public ScheduleMode getScheduleMode() {
 		return scheduleMode;
 	}
 
+	public void setMaxPriorAttemptsHistoryLength(int maxPriorAttemptsHistoryLength) {
+		this.maxPriorAttemptsHistoryLength = maxPriorAttemptsHistoryLength;
+	}
```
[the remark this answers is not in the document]
lamber-ken: ok, I see, I'll update it later.

lamber-ken:
Sorry for late for some reason，I can update this pr tomorrow.

lamber-ken:
> Thanks for addressing my comments @lamber-ken. LGTM. Merging once Travis gives green light.

you're welcome!

PR #8269, opened 2019-04-25
[FLINK-12285][test] Harden savepoint relevant test cases by explicitly canceling test job

ifndef-SleePy:

* Harden `SavepointITCase` and `SavepointMigrationTestBase`

## Brief change log

* Cancel test job explicitly before shutting down cluster

## Verifying this change

* This change is about test cases

  - The serializers: no

## Documentation

### DIFF
--- a/flink-tests/src/test/java/org/apache/flink/test/checkpointing/SavepointITCase.java
+++ b/flink-tests/src/test/java/org/apache/flink/test/checkpointing/SavepointITCase.java
@@ -27,7 +27,6 @@ import org.apache.flink.api.common.restartstrategy.RestartStrategies;
 import org.apache.flink.api.common.state.ValueState;
 import org.apache.flink.api.common.state.ValueStateDescriptor;
 import org.apache.flink.api.common.time.Deadline;
-import org.apache.flink.api.common.time.Time;
 import org.apache.flink.api.java.functions.KeySelector;
 import org.apache.flink.client.program.ClusterClient;
 import org.apache.flink.configuration.CheckpointingOptions;
@@ -35,14 +34,13 @@ import org.apache.flink.configuration.Configuration;
 import org.apache.flink.configuration.TaskManagerOptions;
 import org.apache.flink.core.testutils.OneShotLatch;
 import org.apache.flink.runtime.client.JobExecutionException;
-import org.apache.flink.runtime.concurrent.FutureUtils;
 import org.apache.flink.runtime.jobgraph.JobGraph;
 import org.apache.flink.runtime.jobgraph.JobStatus;
 import org.apache.flink.runtime.jobgraph.JobVertex;
 import org.apache.flink.runtime.jobgraph.SavepointRestoreSettings;
 import org.apache.flink.runtime.messages.FlinkJobNotFoundException;
-import org.apache.flink.runtime.testingUtils.TestingUtils;
 import org.apache.flink.runtime.testtasks.BlockingNoOpInvokable;
+import org.apache.flink.runtime.testutils.CommonTestUtils;
 import org.apache.flink.runtime.testutils.MiniClusterResourceConfiguration;
 import org.apache.flink.streaming.api.checkpoint.ListCheckpointed;
 import org.apache.flink.streaming.api.datastream.DataStream;
@@ -192,7 +190,11 @@ public class SavepointITCase extends TestLogger {
 
 			StatefulCounter.getProgressLatch().await();
 
-			return client.cancelWithSavepoint(jobId, null);
+			final String savepointPath = client.cancelWithSavepoint(jobId, null);
+
+			waitJobCanceled(client, jobId);
+
+			return savepointPath;
 		} finally {
 			cluster.after();
 			StatefulCounter.resetForTest(parallelism);
@@ -240,13 +242,7 @@ public class SavepointITCase extends TestLogger {
 
 			client.cancel(jobId);
 
-			FutureUtils.retrySuccessfulWithDelay(
-				() -> client.getJobStatus(jobId),
-				Time.milliseconds(50),
-				Deadline.now().plus(Duration.ofSeconds(30)),
-				status -> status == JobStatus.CANCELED,
-				TestingUtils.defaultScheduledExecutor()
-			);
+			waitJobCanceled(client, jobId);
 
 			client.disposeSavepoint(savepointPath)
 				.get();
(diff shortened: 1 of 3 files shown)


flink-tests/src/test/java/org/apache/flink/test/checkpointing/SavepointITCase.java
```diff
@@ -324,6 +334,9 @@ public void testTriggerSavepointWithCheckpointingDisabled() throws Exception {
 			assertTrue(ExceptionUtils.findThrowable(e, IllegalStateException.class).isPresent());
 			assertTrue(ExceptionUtils.findThrowableWithMessage(e, graph.getJobID().toString()).isPresent());
 			assertTrue(ExceptionUtils.findThrowableWithMessage(e, "is not a streaming job").isPresent());
+
+			client.cancel(graph.getJobID());
+			waitJobCanceled(client, graph.getJobID());
```
zentol: should not be necessary since the cluster is shut down afterwards

flink-tests/src/test/java/org/apache/flink/test/checkpointing/SavepointITCase.java
```diff
@@ -439,7 +452,11 @@ public void testCanRestoreWithModifiedStatelessOperators() throws Exception {
 			StatefulCounter.getProgressLatch().await(deadline.timeLeft().toMillis(), TimeUnit.MILLISECONDS);
 
 			savepointPath = client.triggerSavepoint(jobID, null).get();
+
 			LOG.info("Retrieved savepoint: " + savepointPath + ".");
+
+			client.cancel(jobID);
+			waitJobCanceled(client, jobID);
```
zentol: should not be necessary since the cluster is shut down afterwards

flink-tests/src/test/java/org/apache/flink/test/checkpointing/SavepointITCase.java
```diff
@@ -489,6 +506,9 @@ public void testCanRestoreWithModifiedStatelessOperators() throws Exception {
 
 			// Await some progress after restore
 			StatefulCounter.getProgressLatch().await(deadline.timeLeft().toMillis(), TimeUnit.MILLISECONDS);
+
+			client.cancel(modifiedJobGraph.getJobID());
+			waitJobCanceled(client, modifiedJobGraph.getJobID());
```
zentol: should not be necessary since the cluster is shut down afterwards

flink-tests/src/test/java/org/apache/flink/test/checkpointing/SavepointITCase.java
```diff
@@ -240,13 +242,7 @@ private void restoreJobAndVerifyState(String savepointPath, MiniClusterResourceF
...
 			client.cancel(jobId);
 
-			FutureUtils.retrySuccessfulWithDelay(
-				() -> client.getJobStatus(jobId),
-				Time.milliseconds(50),
-				Deadline.now().plus(Duration.ofSeconds(30)),
-				status -> status == JobStatus.CANCELED,
-				TestingUtils.defaultScheduledExecutor()
-			);
+			waitJobCanceled(client, jobId);
```
zentol: should not be necessary since the cluster is shut down afterwards

flink-tests/src/test/java/org/apache/flink/test/checkpointing/utils/SavepointMigrationTestBase.java
```diff
@@ -184,6 +185,9 @@ protected final void executeAndSavepoint(
 		} else {
 			FileUtils.moveFile(jobManagerSavepoint, new File(savepointPath));
 		}
+
+		client.cancel(jobGraph.getJobID());
+		waitJobCanceled(client, jobGraph.getJobID());
```
zentol: should not be necessary since the cluster is shut down afterwards

flink-tests/src/test/java/org/apache/flink/test/checkpointing/SavepointITCase.java
```diff
@@ -701,9 +719,7 @@ public Integer map(Integer value) throws Exception {
 			}
 
 			client.cancel(jobGraph.getJobID());
-			while (!client.getJobStatus(jobGraph.getJobID()).get().isGloballyTerminalState()) {
-				Thread.sleep(100);
-			}
+			waitJobCanceled(client, jobGraph.getJobID());
```
zentol: should not be necessary since the cluster is shut down afterwards

flink-tests/src/test/java/org/apache/flink/test/checkpointing/SavepointITCase.java
```diff
@@ -192,7 +190,11 @@ private String submitJobAndTakeSavepoint(MiniClusterResourceFactory clusterFacto
 
 			StatefulCounter.getProgressLatch().await();
 
-			return client.cancelWithSavepoint(jobId, null);
+			final String savepointPath = client.cancelWithSavepoint(jobId, null);
+
+			waitJobCanceled(client, jobId);
```
zentol: should not be necessary since the cluster is shut down afterwards

flink-tests/src/test/java/org/apache/flink/test/checkpointing/utils/SavepointMigrationTestBase.java
```diff
@@ -244,5 +248,19 @@ protected final void restoreAndExecute(
 		if (!done) {
 			fail("Did not see the expected accumulator results within time limit.");
 		}
+
+		client.cancel(jobGraph.getJobID());
+		waitJobCanceled(client, jobGraph.getJobID());
```
zentol: should not be necessary since the cluster is shut down afterwards

shixiaogang:
> most of these should have no positive effect since the `MiniClusterResource` is shut down at the end of the test.

@zentol The problem here is that MiniCluster cannot correctly release memory when it's shutdown. When there exist tasks running in TaskExecutors, TaskExecutors will throw exceptions and will not release direct memory for network buffers and memory manager. Succeeding tests running in the same JVM process may fail due to insufficient amount of direct memory.

The problem can be reproduced when changing the settings of MiniCluster, including the number of network buffers, the number of task managers, the number of slots in task managers, and slot sharing.

ifndef-SleePy:
Hi @zentol ,
Currently `MiniCluster` does not clean the running tasks properly, see [FLINK-11630](https://issues.apache.org/jira/browse/FLINK-11630). As described in this FLINK-12285 JIRA issue, IMO we should explicitly canceling the job before shutting down cluster. Elegant exit is always a good practice, especially in test case which there might be other cases run in the same process.

BTW, even though [FLINK-11630](https://issues.apache.org/jira/browse/FLINK-11630) should solve the memory leaking problem (sadly the PR doesn't work well), it is probably a best-effort solution. I think we still need this explicitly canceling.

What do you think?

zentol:
Instead of requiring the test to explicitly shut down all jobs, which will inevitable cause issues in other tests that forget to do this, we should instead modify the MiniClusterResource to shut down all jobs in the shutdown method.

zentol:
something like this in `after()` should work:
```
try {
	final Deadline deadline = Deadline.fromNow(Duration.ofSeconds(30));

	List<CompletableFuture<Acknowledge>> jobCancellationFutures = miniCluster.listJobs()
		.get(deadline.timeLeft().toMillis(), TimeUnit.MILLISECONDS)
		.stream()
		.filter(status -> !status.getJobState().isGloballyTerminalState())
		.map(status -> miniCluster.cancelJob(status.getJobId()))

	FutureUtils.waitForAll(jobCancellationFutures)
		.get(deadline.timeLeft().toMillis(), TimeUnit.MILLISECONDS);
} catch (InterruptedException | ExecutionException e) {
	log.warn("Exception while shutting down remaining jobs.", e);
} catch (TimeoutException e) {
	log.warn("Shutting down remaining jobs resulted in a timeout.", e);
}
```

ifndef-SleePy:
Hi @zentol ,
Sorry for the late response. I'm in vacation recently.
Thank you for feedback. Your solution makes sense to me, and might be helpful for other cases. I will update the PR later.

PR #8271, opened 2019-04-25
 [FLINK-8640][build][travis] Enable japicmp on Java 9 

zentol:

The `japicmp-plugin` is usable again on java 9+.

## Brief change log

* add explicit Java EE dependencies that the plugin requires, which are no longer available by default on Java 9+
* modify `hadoop-common` `jdk.tools` exclusion; this (system-provided) dependency is no longer available on java 9+ and thus has to be excluded. Defining the dependencies in the `dependencyManagement` section without a version lead to build failures (for _some_ reason), so we now also specify the hadoop version which just works out. Also adds such an exclusion for `hadoop-annotations`, which was simply missing. Note that this is an issue during **compile**  time; there are no repercussions on the runtime (by virtue of being about system dependencies).
* enable japicmp in our java 9 builds

## Verifying this change

Run plugin locally or on travis on java 9.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -189,39 +189,39 @@ jobs:
       jdk: "openjdk9"
       stage: compile
       script: ./tools/travis_controller.sh compile
-      env: PROFILE="-Dhadoop.version=2.8.3 -Pinclude-kinesis -Dinclude_hadoop_aws -Dscala-2.11 -Djdk9 -Djapicmp.skip=true"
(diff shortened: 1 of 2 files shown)


StephanEwen:
Looks good. +1

I assume that this seems to be a temporary fix until the Hadoop dependencies are part of `flink-shaded`?

zentol:
no, this is permanent. This doesn't just affect `flink-shaded-hadoop`; for example `flink-hbase` has the same issue; likely any module that pulls in hadoop (even as a test dependency) should be affected.

PR #8272, opened 2019-04-25
[FLINK-12274] Fix documentation to enable Queryable State

PierreZ:

Queryable state, since 1.8, now need to be enable through configuration. This add the right documentation to help people using it.

## Brief change log

  - add reference to configuration parameter

  - The serializers:  no

## Documentation

  - If yes, how is the feature documented? now, yes!

### DIFF
--- a/docs/dev/stream/state/queryable_state.md
+++ b/docs/dev/stream/state/queryable_state.md
@@ -69,10 +69,12 @@ response back to the client.
 
 ## Activating Queryable State
 
-To enable queryable state on your Flink cluster, you just have to copy the 
-`flink-queryable-state-runtime{{ site.scala_version_suffix }}-{{site.version }}.jar` 
+To enable queryable state on your Flink cluster, you need to do the following:
+
+ 1. copy the `flink-queryable-state-runtime{{ site.scala_version_suffix }}-{{site.version }}.jar` 
 from the `opt/` folder of your [Flink distribution](https://flink.apache.org/downloads.html "Apache Flink: Downloads"), 
-to the `lib/` folder. Otherwise, the queryable state feature is not enabled. 
+to the `lib/` folder.
+ 2. set the property `queryable-state.enable` to `true`. See the [Configuration]({{ site.baseurl }}/ops/config.html#queryable-state) documentation for details and additional parameters.
 
 To verify that your cluster is running with queryable state enabled, check the logs of any 
 task manager for the line: `"Started the Queryable State Proxy Server @ ..."`.


docs/dev/stream/state/queryable_state.md
```diff
@@ -69,10 +69,12 @@ response back to the client.
...
 
-To enable queryable state on your Flink cluster, you just have to copy the 
-`flink-queryable-state-runtime{{ site.scala_version_suffix }}-{{site.version }}.jar` 
+To enable queryable state on your Flink cluster, you need to do the following:
+
+ 1. copy the `flink-queryable-state-runtime{{ site.scala_version_suffix }}-{{site.version }}.jar` 
 from the `opt/` folder of your [Flink distribution](https://flink.apache.org/downloads.html "Apache Flink: Downloads"), 
-to the `lib/` folder. Otherwise, the queryable state feature is not enabled. 
+to the `lib/` folder.
+ 2. set the property `queryable-state.enable` to `true` in `./conf/flink-conf.yaml`. See the [Configuration]({{ site.baseurl }}/ops/config.html#queryable-state) documentation for details and additional parameters.
```
Myasuka: Just a minor suggestion, we could configure `queryable-state.enable` not only have to change the `./conf/flink-conf.yaml` file but also when we submit jobs. I think it would be better to remove the description of "in ./conf/flink-conf.yaml".

Myasuka: Just a minor suggestion, we could configure `queryable-state.enable` not only have to change the `./conf/flink-conf.yaml` file but also when we submit jobs. I think it would be better to remove the description of "in ./conf/flink-conf.yaml".

rmetzger:
@flinkbot approve description

rmetzger:
@flinkbot attention @eaglewatcherwb

@eaglewatcherwb: do you agree with this change? (since you introduced the config flag)

docs/dev/stream/state/queryable_state.md
Myasuka: In my mind, "set the property `queryable-state.enable` to `true` in `./conf/flink-conf.yaml`" might give user a subconscious hint that we could only change properties by editing the `./conf/flink-conf.yaml`. Actually, we could change related properties through other ways such as `-yD <property=value>` when submitting jobs in YARN.

In other words, I think describe this by only speaking of "set the property `queryable-state.enable` to `true`" should be enough.

Last but not least, this is a just a `comment` not a `request change` from a non-native English speaker understanding.

rmetzger: Okay, I see your point. I leave it to @PierreZ to make the final decision.

PierreZ: Thanks to both of you, I'm dropping the last part of the sentence!

eaglewatcherwb:
> @flinkbot attention @eaglewatcherwb
> 
> @eaglewatcherwb: do you agree with this change? (since you introduced the config flag)

It seems good to me.

rmetzger:
Thank you!

@flinkbot approve all

I'll merge the PR then

PR #8278, opened 2019-04-26
[FLINK-12335][Table-runtime]Remove useless code in class SegmentsUtil

liyafan82:
…entsUtil

Improve the performance of SegmentsUtil class :

To evaluate the offset, an integer is bitand with a mask to clear to low bits, and then shift right. The bitand is useless:
((index & BIT_BYTE_POSITION_MASK) >>> 3)

## Brief change log

  - Change class SegmentsUtil to fix the above 2 issues.

## Verifying this change

Verified by manually running existing test cases.

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/util/SegmentsUtil.java
+++ b/flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/util/SegmentsUtil.java
@@ -36,8 +36,6 @@ public class SegmentsUtil {
 	 */
 	public static final boolean LITTLE_ENDIAN = ByteOrder.nativeOrder() == ByteOrder.LITTLE_ENDIAN;
 
-	private static final int BIT_BYTE_POSITION_MASK = 0xfffffff8;
-
 	private static final int BIT_BYTE_INDEX_MASK = 7;
 
 	/**
@@ -62,13 +60,17 @@ public class SegmentsUtil {
 	public static byte[] allocateReuseBytes(int length) {
 		byte[] bytes = BYTES_LOCAL.get();
 
-		if (bytes == null || bytes.length < length) {
-			bytes = new byte[length];
-
+		if (bytes == null) {
 			if (length <= MAX_BYTES_LENGTH) {
+				bytes = new byte[MAX_BYTES_LENGTH];
 				BYTES_LOCAL.set(bytes);
+			} else {
+				bytes = new byte[length];
 			}
+		} else if (bytes.length < length) {
+			bytes = new byte[length];
 		}
+
 		return bytes;
 	}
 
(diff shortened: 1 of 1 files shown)


flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/util/SegmentsUtil.java
```diff
@@ -445,7 +443,7 @@ public static void bitSet(MemorySegment segment, int baseOffset, int index) {
 	 * @param index bit index from base offset.
 	 */
 	public static boolean bitGet(MemorySegment segment, int baseOffset, int index) {
-		int offset = baseOffset + ((index & BIT_BYTE_POSITION_MASK) >>> 3);
+		int offset = baseOffset + (index  >>> 3);
```
JingsongLi: Can you and a util method like this:
```
/**
     * Given a bit index, return word index containing it.
     */
    private static int wordIndex(int bitIndex) {
        return bitIndex >> ADDRESS_BITS_PER_WORD;
    }
```

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/util/SegmentsUtil.java
```diff
@@ -417,7 +415,7 @@ private static boolean inFirstSegment(MemorySegment[] segments, int offset, int
 	 * @param index bit index from base offset.
 	 */
 	public static void bitUnSet(MemorySegment segment, int baseOffset, int index) {
-		int offset = baseOffset + ((index & BIT_BYTE_POSITION_MASK) >>> 3);
+		int offset = baseOffset + (index >>> 3);
 		byte current = segment.get(offset);
 		current &= ~(1 << (index & BIT_BYTE_INDEX_MASK));
```
JingsongLi: It seems that `BIT_BYTE_INDEX_MASK` is not needed either.

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/util/SegmentsUtil.java
liyafan82: Good suggestion. Thank you.

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/util/SegmentsUtil.java
liyafan82: Good point. Thanks for your comments.

liyafan82:
> Thanks @liyafan82 for you code, maybe you can change the title to `remove useless code in SegmentsUtil`, I think the jdk/jvm will optimizer it, so maybe performance keep same.

Thanks for your comments. 

I have made some investigations to analyze the assembly generated by JIT. 
It seems that JIT successfully removes BIT_BYTE_POSITION_MASK, but fails to remove BIT_BYTE_INDEX_MASK. So there can be performance overhead.

image

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/util/SegmentsUtil.java
JingsongLi: Can you do it in this PR?

JingsongLi:
Thanks @liyafan82 , LGTM +1

JingsongLi:
@flinkbot attention @wuchong

wuchong:
LGTM.

wuchong:
LGTM.

KurtYoung:
Cool, thanks @liyafan82 for your contributions.

PR #8280, opened 2019-04-26
[FLINK-12297]Harden ClosureCleaner to handle the wrapped function

Aitozi:

ClosureCleaner is used to clean the implicit `this$x` point to the outer class in the user defined function to reduce the case of serialization function failed.

But if the function is wrapped by user or the flink framework, it will escape the clean operation. This pr harden the ClosureCleaner to handle this. And I think it is also a feasible approach for [[FLINK-12113]](https://issues.apache.org/jira/browse/FLINK-12113)

## Brief change log

check the field if need to further clean and apply

## Verifying this change

1. Add the case in the issue to confirm
2. Add the wrapFunction test case in ClosuerCleanerTest

### DIFF
--- a/flink-core/src/main/java/org/apache/flink/api/common/ExecutionConfig.java
+++ b/flink-core/src/main/java/org/apache/flink/api/common/ExecutionConfig.java
@@ -90,7 +90,7 @@ public class ExecutionConfig implements Serializable, Archiveable<ArchivedExecut
 	/** Defines how data exchange happens - batch or pipelined */
 	private ExecutionMode executionMode = ExecutionMode.PIPELINED;
 
-	private ClosureCleanerLevel closureCleanLevel = ClosureCleanerLevel.RECURSIVE;
+	private ClosureCleanerLevel closureCleanerLevel = ClosureCleanerLevel.RECURSIVE;
 
 	private int parallelism = PARALLELISM_DEFAULT;
 
@@ -188,7 +188,7 @@ public class ExecutionConfig implements Serializable, Archiveable<ArchivedExecut
 	 * User code must be serializable because it needs to be sent to worker nodes.
 	 */
 	public ExecutionConfig enableClosureCleaner() {
-		this.closureCleanLevel = ClosureCleanerLevel.RECURSIVE;
+		this.closureCleanerLevel = ClosureCleanerLevel.RECURSIVE;
 		return this;
 	}
 
@@ -198,7 +198,7 @@ public class ExecutionConfig implements Serializable, Archiveable<ArchivedExecut
 	 * @see #enableClosureCleaner()
 	 */
 	public ExecutionConfig disableClosureCleaner() {
-		this.closureCleanLevel = ClosureCleanerLevel.NONE;
+		this.closureCleanerLevel = ClosureCleanerLevel.NONE;
 		return this;
 	}
 
@@ -208,28 +208,23 @@ public class ExecutionConfig implements Serializable, Archiveable<ArchivedExecut
 	 * @see #enableClosureCleaner()
 	 */
 	public boolean isClosureCleanerEnabled() {
-		return !(closureCleanLevel == ClosureCleanerLevel.NONE);
+		return !(closureCleanerLevel == ClosureCleanerLevel.NONE);
 	}
 
 	/**
-	 * Sets the level of ClosureCleaner. It has three levels:
-	 * NONE: means disable the clean.
-	 * TOP_LEVEL: means only clean the top level of the class passed by user.
-	 * RECURSIVE: means do the clean recursively with looping over all the needed fields, which is the default clean level.
-	 *
-	 * @param level
-	 * @return
+	 * Configures the closure cleaner. Please see {@link ClosureCleanerLevel} for details on the
+	 * different settings.
 	 */
-	public ExecutionConfig setClosureCleanLevel(ClosureCleanerLevel level) {
-		this.closureCleanLevel = level;
+	public ExecutionConfig setClosureCleanerLevel(ClosureCleanerLevel level) {
+		this.closureCleanerLevel = level;
 		return this;
 	}
 
 	/**
-	 * @return The ClosureClean level.
+	 * Returns the configured {@link ClosureCleanerLevel}.
 	 */
-	public ClosureCleanerLevel getClosureCleanLevel() {
-		return closureCleanLevel;
+	public ClosureCleanerLevel getClosureCleanerLevel() {
+		return closureCleanerLevel;
 	}
 
 	/**
@@ -979,7 +974,7 @@ public class ExecutionConfig implements Serializable, Archiveable<ArchivedExecut
 
 			return other.canEqual(this) &&
 				Objects.equals(executionMode, other.executionMode) &&
-				closureCleanLevel == other.closureCleanLevel &&
+				closureCleanerLevel == other.closureCleanerLevel &&
 				parallelism == other.parallelism &&
 				((restartStrategyConfiguration == null && other.restartStrategyConfiguration == null) ||
 					(null != restartStrategyConfiguration && restartStrategyConfiguration.equals(other.restartStrategyConfiguration))) &&
@@ -1009,7 +1004,7 @@ public class ExecutionConfig implements Serializable, Archiveable<ArchivedExecut
 	public int hashCode() {
 		return Objects.hash(
 			executionMode,
-			closureCleanLevel,
+				closureCleanerLevel,
 			parallelism,
 			restartStrategyConfiguration,
 			forceKryo,
@@ -1078,21 +1073,21 @@ public class ExecutionConfig implements Serializable, Archiveable<ArchivedExecut
 	}
 
 	/**
-	 * The clean level determines whether need to clean or whether need to clean recursively.
+	 * Configuration settings for the closure cleaner.
 	 */
 	public enum ClosureCleanerLevel {
 		/**
-		 * 	Do not perform clean for closure function.
+		 * Disable the closure cleaner completely.
 		 */
 		NONE,
 
 		/**
-		 * Just clean the top level the closure function.
+		 * Clean only the top-level class without recursing into fields.
 		 */
 		TOP_LEVEL,
 
 		/**
-		 * Try to clean the closure function recursively.
+		 * Clean all the fields recursively.
 		 */
 		RECURSIVE
 	}
--- a/flink-core/src/main/java/org/apache/flink/api/common/ExecutionConfig.java
+++ b/flink-core/src/main/java/org/apache/flink/api/common/ExecutionConfig.java
@@ -90,7 +90,7 @@ public class ExecutionConfig implements Serializable, Archiveable<ArchivedExecut
 	/** Defines how data exchange happens - batch or pipelined */
 	private ExecutionMode executionMode = ExecutionMode.PIPELINED;
 
-	private boolean useClosureCleaner = true;
+	private ClosureCleanerLevel closureCleanLevel = ClosureCleanerLevel.RECURSIVE;
 
 	private int parallelism = PARALLELISM_DEFAULT;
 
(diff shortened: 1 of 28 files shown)


Aitozi:
@dawidwys @ifndef-SleePy could you help review this PR

Aitozi:
May be the author of the `ClosureClose.java` should also keep an eye on this PR, thanks.  @flinkbot attention @aljoscha

StephanEwen:
This PR does a few things at the same time:
  1. It recurses into fields, cleaning them.
  2. It skips closure cleaning of functions that have a `writeObject` method or are `Externalizable`.
  3. It adds checks to skip classes that are not inner classes.

Recursion into non-transient and non-static fields probably solves the "wrapping function" problem, but it does much more than that. While this can be a good addition, we definitely need a way to turn this off, so we need to extend the closure cleaner configuration to support "don't clean", "clean", and "clean recursive".

I think that (2) and (3) are good additions, but the check should probably be at the top level, not only before checking fields. Especially (3) would help make this faster, because many classes can be skipped right away.

This check should be simpler, though:
```java
public static void clean(Object function) {
    if (function == null) {
        return
    }

    final Class<?> clazz = function.getClass();
    if (clazz.getEnclosingClass() == null || !Modifiers.isStatic(clazz.getModifiers())) {
        return;
    }

    // do the cleaning
}
```

StephanEwen:

Aitozi:
Thanks @StephanEwen for your valuable suggestion, I will address your comments as far as i have time (about at this weekend).

Aitozi:
Hi, @StephanEwen 
I have followed your suggestion to add some fix. Still two have to confim:
1. check at the top level with the condition you mentioned may skip some class wrong, think about a  class has no enclosingClass it may also have to be cleaned, so i don't add the check.
2. I found that the scala class clean is working with `ClosureCleaner.scala`. I think it should be consistent with the java clean action. I am not quiet familiar with the scala, but i will try to look into it.

Aitozi:
ping @StephanEwen

aljoscha:
Thanks for your contribution, I'm on vacation for the end of this week but I will review when I get back next week. 👌

flink-java/src/main/java/org/apache/flink/api/java/ClosureCleaner.java
```diff
@@ -101,6 +132,29 @@ public static void clean(Object func, boolean checkSerializable) {
 		}
 	}
 
+	public static void clean(Object func, boolean checkSerializable) {
```
aljoscha: Do we strictly need the old `clean()` methods that has the default of "recursion". It might be more straightforward in the code to always require the config.

flink-java/src/main/java/org/apache/flink/api/java/ClosureCleaner.java
```diff
@@ -101,6 +132,29 @@ public static void clean(Object func, boolean checkSerializable) {
...
+		clean(func, ExecutionConfig.CLEANLEVEL.RECURSION, checkSerializable);
+	}
+
+	private static boolean isNeedToDig(Field f, Object fo, Class<?> cls) {
```
aljoscha: I think this should be called `needsRecursion` or something like it that makes the recursion clearer.

flink-java/src/main/java/org/apache/flink/api/java/ClosureCleaner.java
```diff
@@ -101,6 +132,29 @@ public static void clean(Object func, boolean checkSerializable) {
...
+		return fo != null &&
+			!checkCustomSerializeMethod(cls) &&
+			!Modifier.isStatic(f.getModifiers()) &&
+			!Modifier.isTransient(f.getModifiers()) &&
+			((fo.getClass().isAnonymousClass()) ||
+				(fo.getClass().isLocalClass()) ||
+				(fo.getClass().isMemberClass()));
+	}
+
+	private static boolean checkCustomSerializeMethod(Class<?> cls) {
```
aljoscha: This could be called `usesCustomSerialization()`.

flink-core/src/main/java/org/apache/flink/api/common/ExecutionConfig.java
```diff
@@ -1054,4 +1075,13 @@ public T getSerializer() {
 			return Collections.emptyMap();
 		}
 	}
+
+	/**
+	 * The clean level determines whether need to clean or whether need to clean recursively.
+	 */
+	public enum CLEANLEVEL {
```
aljoscha: This should be in line with regular java style, i.e. the name should be something like `ClosureCleanerLevel` or `ClosureCleanerConfig`. The individual options could be something like `NONE`, `TOP_LEVEL`, or `RECURSIVE`, with javadoc that describes what the options do inserted here.

flink-java/src/main/java/org/apache/flink/api/java/ClosureCleaner.java
```diff
@@ -73,6 +78,32 @@ public static void clean(Object func, boolean checkSerializable) {
 			if (f.getName().startsWith("this$")) {
 				// found a closure referencing field - now try to clean
 				closureAccessed |= cleanThis0(func, cls, f.getName());
+			} else {
+				Object fo;
```
aljoscha: This could be `fieldObject`, to make the code more readable.

aljoscha:
I think overall the changes are good!

One thing I thought about is what happens when objects that need to be cleaned are nested, i.e. if you have a structure like `OuterStaticClass.InnerStaticClass.InnerAnonymousOrLocalClass`. I think the current approach doesn't catch this, and checking always at the top level, as @StephanEwen suggested would also not catch this problem. Any thoughts?

flink-java/src/main/java/org/apache/flink/api/java/ClosureCleaner.java
Aitozi: I meant to keep the old `clean()` to avoid changing too much clean method, I can remove this.

Aitozi:
Thanks @aljoscha for your review. I have addressed your comments. About the case you mentioned `OuterStaticClass.InnerStaticClass.InnerAnonymousOrLocalClass`, I add a test to cover this, and i think it can work with this approach, you can check the test case and leave more comments.

aljoscha:
Ah, so this shows that there are quite a few places where we use `RECURSIVE` and the closure cleaner level is not actually configurable. Do you think we could do something to fix that as well?

Aitozi:
I know your mean, and can we remove all the scattered `clean()` used in functions like `connector`, `Pattern`. Because IMO with the `RECURSIVE` level clean at the entrance showing below, I think other clean function can be removed, or make this less radical by use `TOP_LEVEL` in these scattered code path to keep old way. WDYT ? 

```
public <F> F clean(F f) {
		if (getConfig().isClosureCleanerEnabled()) {
			ClosureCleaner.clean(f, getConfig().getClosureCleanLevel(), true);
		}
		ClosureCleaner.ensureSerializable(f);
		return f;
	}
```

Aitozi:
I think it over again, it can be removed directly. And after this the clean method should not be directly used, it's should use with one entrance `environment#clean()`

Aitozi:
One way to my mind: set the scattered `clean()` method with the `TOP_LEVEL` to keep the old way, And set `checkSerializable` to false. All the user function will be passed to clean by `environment#clean()`, we can config the clean level there and checkSerializable at the entrance.

```
@Internal
public <F> F clean(F f) {
	if (getConfig().isClosureCleanerEnabled()) {
		ClosureCleaner.clean(f, getConfig().getClosureCleanLevel(), true);
	}
	ClosureCleaner.ensureSerializable(f);
	return f;
}
```

aljoscha:
It's tough, I think `RECURSIVE` might be the better default strategy because it cleans more.

aljoscha:
I still think the approach doesn't work for closures that are nested in "non-closures". I.e when you have a closure as a field in an object of a static class. This test case shows it:

```
	@Test
	public void testRealOuterStaticClassInnerStaticClassInnerAnonymousOrLocalClass() {
		MapFunction<Integer, Integer> nestedMap = new OuterMapCreator().getMap();

		MapFunction<Integer, Integer> wrappedMap = new WrapperMapFunction(nestedMap);

		Tuple1<MapFunction<Integer, Integer>> tuple = new Tuple1<>(wrappedMap);

		ClosureCleaner.clean(tuple, ExecutionConfig.ClosureCleanerLevel.RECURSIVE, true);

		ClosureCleaner.ensureSerializable(tuple);
	}
```

The problem is that recursion stops as soon as it finds a field of a static class, which doesn't take into account the fact that this nested static class can have fields that are themselves closures.

Aitozi:
Agree, but I think keep the  scattered `clean()` with old `TOP_LEVEL` and use configurable `RECURSIVE` at the entrance  will make this more flexible if user want to rollback the clean action.  WDYT @aljoscha.

Aitozi:
I test this case failed due to this check, because the `WrapperMapFunction` is top level, so it doesn't run recursively.

```
(fo.getClass().isAnonymousClass()) ||
(fo.getClass().isLocalClass()) ||
(fo.getClass().isMemberClass())
```

Aitozi:
Sorry for late response @sunjincheng121 , I’m out for these two days, I will further look into this today.

Aitozi:
Hi @aljoscha  The case you post can not past is due to the check below, because a closure can be hidden in any class as you say.

```
(fo.getClass().isAnonymousClass()) ||
(fo.getClass().isLocalClass()) ||
(fo.getClass().isMemberClass())
```

But I can not simply remove the check, because it will run into the StackOverFlowError, So i check the serializability before recursively clean the member field.  What do you think of the solution now? The travis failed seems unrelated.

aljoscha:
I simplified the recursion and fixed the stack overflow error (the problem was that the ClosureCleaner was recursing into primitive types/wrappers and would get stuck on `Long.MIN_VALUE`. You can see this yourself by printing the name of the field when recursing). You can see my changes here: https://github.com/aljoscha/flink/commits/pr/8280-closure-cleaner. If Travis is green on this I would merge, if you're OK with the changes.

Aitozi:
Thanks for your help @aljoscha  It's true that it was stuck by the `Long.MIN_VALUE` or (`Integer.MAX_VALUE` I met in my local test) and so on. I did't to this check because i am not very sure about whether this cover all case (may be it's enough). 
Another doubt that why you remove the static and transient modifier check. I think the static class can not be skipped check but the field with static/transient modifier should be skipped.

aljoscha:
You're exactly right! I'll not remove those checks.

Aitozi:
OK, and the travis failed seems due to the python word count e2e test ?

Aitozi:
I think the check for primitive type maybe not enough, do you see the last travis ci result <https://api.travis-ci.org/v3/job/544637574/log.txt>, but it seems disappear during the second trigger. No matter the travis result , I think a class below will also run into stack overflow error, right?

```
class A {
    A fieldA;
}
```

Do you think the solution to check the serializability before recursive is not a good idea?  I think this can solve the problem because we only have to recursively clean the unserializable fields, WDYT @aljoscha ?

aljoscha:
I added a test for a recursive class to the test in my branch: https://github.com/aljoscha/flink/commits/pr/8280-closure-cleaner

As I expected, this is fine, because the recursion naturally stops at some point. The only problem I could imagine if we had a type that references itself. But that's a bit of an edge case that users should avoid. But we could add code for dealing with it if we want.

Could you maybe take the changes from my branch and add them to this PR, i.e. force push those changes?

Aitozi:
OK, I will take a look tonight, thanks.

Aitozi:
> The only problem I could imagine if we had a type that references itself. 

I check the case, and i think you are right👍.  I have cherry-pick the last three commits from your branch and squashed, thanks for your help @aljoscha . Before merging,  I will also check the last failure in your branch <https://api.travis-ci.org/v3/job/544637574/log.txt> by running in my local environment and also waiting the travis to be green.

sunjincheng121:
The CI failed, I restarted the CI due to It looks the EVN problem. :)

sunjincheng121:
Thanks for the review @aljoscha! And thanks for your efforts @Aitozi 
I'll merge this after CI change to green! :)
@flinkbot approve all

aljoscha:
I merged this! Thanks again to the two of you for working on this. 👍

PR #8282, opened 2019-04-26
[FLINK-12338] Update Apache Avro test to use try-with-resources

Fokko:
Hi all,

I've updated the Apache Avro test suite to use try-with-resources. Right now some resources aren't close at all. Having the try-with-resources increases the readability of the code.

The cause of this PR is that I was trying to test the RC of Avro 1.9.0 against the Flink codebase: https://lists.apache.org/thread.html/a856e074b67f24c4352967054dbaece78d38f68e5b86b908f4307713@%3Cdev.avro.apache.org%3E

Cheers, Fokko

Make the tests of the Avro format more readable and make sure that all the resources are closed properly.

## Verifying this change

  - The serializers: no

## Documentation

### DIFF
--- a/flink-formats/flink-avro/src/test/java/org/apache/flink/formats/avro/AvroKryoSerializerRegistrationsTest.java
+++ b/flink-formats/flink-avro/src/test/java/org/apache/flink/formats/avro/AvroKryoSerializerRegistrationsTest.java
@@ -56,8 +56,9 @@ public class AvroKryoSerializerRegistrationsTest {
 	public void testDefaultKryoRegisteredClassesDidNotChange() throws Exception {
 		final Kryo kryo = new KryoSerializer<>(Integer.class, new ExecutionConfig()).getKryo();
 
-		try (BufferedReader reader = new BufferedReader(new InputStreamReader(
-				getClass().getClassLoader().getResourceAsStream("flink_11-kryo_registrations")))) {
+		try (InputStreamReader input = new InputStreamReader(
+				getClass().getClassLoader().getResourceAsStream("flink_11-kryo_registrations"));
+			BufferedReader reader = new BufferedReader(input)) {
 
 			String line;
 			while ((line = reader.readLine()) != null) {
--- a/flink-formats/flink-avro/src/test/java/org/apache/flink/formats/avro/AvroOutputFormatTest.java
+++ b/flink-formats/flink-avro/src/test/java/org/apache/flink/formats/avro/AvroOutputFormatTest.java
@@ -107,23 +107,23 @@ public class AvroOutputFormatTest {
 			outputFormat.setSchema(schema);
 		}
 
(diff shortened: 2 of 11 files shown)


flink-formats/flink-avro/src/test/java/org/apache/flink/formats/avro/AvroKryoSerializerRegistrationsTest.java
```diff
@@ -56,8 +56,9 @@
 	public void testDefaultKryoRegisteredClassesDidNotChange() throws Exception {
 		final Kryo kryo = new KryoSerializer<>(Integer.class, new ExecutionConfig()).getKryo();
 
-		try (BufferedReader reader = new BufferedReader(new InputStreamReader(
```
StephanEwen: I think changes like this one are not necessary. The `BufferedReader` forwards the `close()` call to the nested `InputStreamReader`. So there is no need to have them separate in the `try(...)` clause.

I would keep the original pattern, which seems easier to read/format.

flink-formats/flink-avro/src/test/java/org/apache/flink/formats/avro/AvroOutputFormatTest.java
```diff
@@ -107,23 +107,23 @@ private void serializeAndDeserialize(final AvroOutputFormat.Codec codec, final S
...
-
-			assertTrue(codec != null ? restoredCodec == codec : restoredCodec == null);
-			assertTrue(schema != null ? restoredSchema.equals(schema) : restoredSchema == null);
+		try (final ByteArrayOutputStream bos = new ByteArrayOutputStream()) {
+			// when
+			try (final ObjectOutputStream oos = new ObjectOutputStream(bos)) {
+				oos.writeObject(outputFormat);
+			}
+			try (final ByteArrayInputStream bais = new ByteArrayInputStream(bos.toByteArray());
+				final ObjectInputStream ois = new ObjectInputStream(bais)) {
```
StephanEwen: same as above, would not separate the nested streams/readers

flink-formats/flink-avro/src/test/java/org/apache/flink/formats/avro/EncoderDecoderTest.java
```diff
@@ -66,27 +66,28 @@ public void testComplexStringsDirecty() {
...
 
 				byte[] data = baos.toByteArray();
 
 				// deserialize
 				{
-					ByteArrayInputStream bais = new ByteArrayInputStream(data);
-					DataInputStream dataIn = new DataInputStream(bais);
-					DataInputDecoder decoder = new DataInputDecoder();
-					decoder.setIn(dataIn);
+					try (ByteArrayInputStream bais = new ByteArrayInputStream(data)) {
```
StephanEwen: same as above

PR #8288, opened 2019-04-27
[FLINK-12345] [table-planner-blink] Add support for generating optimized logical plan for stream window aggregate

godfreyhe:

*Add support for generating optimized logical plan for stream window aggregate*

## Brief change log

  - *refactor batch window aggregate*
  - *add RelTimeIndicatorConverter to materialize rowtime/proctime type to timestamp type*
  - *add WindowAggregateReduceFunctionsRule to reduce complex aggregate function to simpler forms*
  - *add StreamExecGroupWindowAggregateRule to convert logical window aggregate to stream physical window aggregate*

## Verifying this change

  - *Added WindowAggregateTest that validates physical plan for window aggregate queries*
  - *Added RelTimeIndicatorConverterTest that validates plan after materialized*

  - The serializers: (no)

## Documentation


flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/api/TableConfig.scala
```diff
@@ -53,6 +53,30 @@ class TableConfig {
...
+
+  /**
+    * The early firing interval in milli second, early fire is the emit strategy
+    * before watermark advanced to end of window.
+    *
+    * < 0 means no early fire
+    * 0 means no delay (fire on every element).
+    * > 0 means the fire interval
+    */
+  private var earlyFireInterval = DEFAULT_FIRE_INTERVAL
```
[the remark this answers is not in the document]
wuchong: I think we can support it if we expose early&late interval through options in `TableConfig`.

I think it's nice to have it.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/plan/nodes/physical/stream/StreamExecGroupWindowAggregate.scala
```diff
@@ -0,0 +1,139 @@
...
+import org.apache.calcite.rel.{RelNode, RelWriter, SingleRel}
+
+import java.util
+
+import scala.collection.JavaConversions._
+
+/**
+  * Streaming group window aggregate physical node which will be translate to window operator.
+  *
+  * If requirements satisfied, it will be translated into minibatch window operator, otherwise,
```
wuchong: The minibatch-window is still in beta in Blink. I think we can support it after 1.9 release.

What do you think about removing minibatch-window relative comments and codes?

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/functions/sql/ProctimeSqlFunction.java
```diff
@@ -37,16 +39,24 @@ public ProctimeSqlFunction() {
 		super(
 			"PROCTIME",
 			SqlKind.OTHER_FUNCTION,
-			ReturnTypes.explicit(new ProctimeRelProtoDataType()),
-			null,
-			OperandTypes.NILADIC,
-			SqlFunctionCategory.TIMEDATE);
+			ReturnTypes.cascade(ReturnTypes.explicit(SqlTypeName.TIMESTAMP), SqlTypeTransforms.TO_NULLABLE),
```
wuchong: `PROCTIME()` is a built-in function used to generate proctime attribute field, whose return type must be proctime time indicator.

I think what you want is `ProctimeMaterializeSqlFunction` which materialize a proctime field into `Timestamp` field.

So I think we don't need to change this file. We should use `FlinkSqlOperatorTable.PROCTIME_MATERIALIZE` in `RelTimeIndicatorConverter`.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/calcite/RelTimeIndicatorConverter.scala
```diff
@@ -0,0 +1,667 @@
...
+  override def visit(union: LogicalUnion): RelNode =
+    visitSetOp(union)
+
+  override def visit(aggregate: LogicalAggregate): RelNode = convertAggregate(aggregate)
+
+  override def visit(minus: LogicalMinus): RelNode =
+    visitSetOp(minus)
+
+  override def visit(sort: LogicalSort): RelNode = {
+
```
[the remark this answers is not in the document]
godfreyhe: ok

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/plan/nodes/physical/stream/StreamExecGroupWindowAggregate.scala
godfreyhe: It makes sense to me

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/functions/sql/ProctimeSqlFunction.java
godfreyhe: I didn't notice `ProctimeMaterializeSqlFunction`'s existence before, i will revert the changes in `ProctimeSqlFunction`

wuchong:
Thanks @godfreyhe ,  it looks good to me now.

PR #8290, opened 2019-04-27
[FLINK-12070] [network] Implement new bounded blocking subpartitions

StephanEwen:

This PR changes the implementation of the bounded blocking result partitions (batch shuffles) to
make the implementation simpler, more efficient, and most importantly: allow for multiple (and concurrent) consumers of that data.

Rather than using a limited set of Flink network buffers and manually managing them and spilling them
to disk, this implementation simply creates a temporary memory mapped file, writes the data to that
memory region, and serves results from there.

This has multiple advantages:
  - Simple implementation that stays in memory as much as possible, but is not limited to memory. The memory mapped file asynchronously spills to disk.
  - Result does not occupy Flink network buffers, saving us the complexity of reasoning about the buffer resource consumption of the result when keeping it around for longer (for batch recovery and interactive queries).
  - Fewer memory copies of the data.

## Open Architecture Questions

@pnowojski @NicoK @zhijiangW Any of you could probably help here.

The biggest open question in this implementation is: when and how to "release" the readers (`ResultSubpartitionView`). To give some background:

  - By Java's default, memory mapped files are unmapped during GC, when no one references the ByteBuffer any more. This is similar to the way Java deals with DirectByteBuffers.
  - This means that unmapping can happen rather late. When one creates many memory mapped files and GC does not happen frequently (which is what we try to achieve), the memory mapped files linger. This can cause problems as it means occupying resources overly long.
  - Netty for example aggressively releases DirectByteBuffers and unmaps memory mapped files to prevent that. They have some utilities for that.
  - This implementation follows the same approach, to make sure we release resources fast.
  - Concretely, this means that files are unmapped once the `BoundedBlockingSubpartition` is released and all readers (ResultSubpartitionViews) are released.

Previously, the readers were tightly coupled to the partitions, once the partition was disposed, the readers were immediately disposed as well. This definitely makes sense for the `PipelinedSubpartition`, because the producer and consumer run concurrently. For the `BoundedBlockingSubpartition`, this is not strictly required, especially since we don't occupy a strictly limited resource (like network buffers).

We could go various routes here:

  1. The SubPartitionView (readers) can continue when the ResultPartition is released. No new readers can be created, and once all existing readers are released, the data file will be unmapped from memory and get deleted. This is the version implemented in this PR.
  
     - This option is simple and efficient. The assumption is that by the time the consumer (Netty or local channel) calls "releaseAllResources()" it won't access buffers any more, so we don't need to track buffer reference counts before unmapping the file. That seems to be strictly true throughout all existing code.

     - We assume that once the consumer tasks are cancelled, the SubPartitionView gets released within reasonable time. That is definitely true for local input channels, however, I could not verify this in the netty code that ships data. Netty might end up shipping large amounts of data, which gets discarded on the receiver side. In a clean design, when the receiver closes the logical channel, the sender side should notice and release the SubPartionView, making this approach work nicely.
  
  ==> This is my favoured design, modulo clarifying the network stack behavior as described above.

  2. The SubPartitionView (readers) get immediately disposed when the ResultSubpartition is disposed. The ResultSubpartition forwards the release() call to all readers.
  
     - To prevent Segfaults, we can only mark the reader as disposed, but need to defer unmapping until all buffers that reference the memory are released. That means keeping a reference counter (increase when slicing a buffer out of the memory mapped byte buffer, decrease when that buffer is recycled). This is possible, but extra overhead.
     
     - This avoids making the assumption that Netty cancels the logical channel when the receiver ends (cancels the partition request).

  ==> I would like to avoid this route, to avoid fine grained buffer reference tracking.

  3. When the `BoundedBlockingSubpartition` is disposed, it does not dispose the readers, but sets a flag such that the readers error out upon the next read request.
  
     - This seems like a hack to me, caused by unclean designs in other components.

     - This avoids making the assumption that Netty cancels the logical channel when the receiver ends (cancels the partition request).

     - Currently, reading from the subpartition must not throw an exception, the network stack will close the full TCP channel and not just cancel the one logical channel. We would need to change the Netty stack to
        - Either handle exceptions during reading from the subpartition views.
        - Or return "null" (which can indicate 'empty' or 'erroneous') and then make sure Netty calls "release()" even through "isReleased()" is already true (super unclean, should totally avoid this option).
        - Or return "null" (which can indicate 'empty' or 'erroneous') and then check a new status "close requested" and let Netty call "release()" in that case.

  ==> I would like to avoid this option, because it seems hacking around symptoms of a problems caused by unclear defined behavior/responsibilities.

## Further ToDos and Follow-ups

Aside from answering the above question, the main remaining task is to benchmark this at large scale.

## Brief change log

  - f239f4a Minor fix to release unpooled event buffers on Netty's sender side. Fixes segfault in a special case when using direct buffers and attaching a debugger

  - acb1613 , df0bd91 , 886d545 , 1fe12c2 : Various minor cleanups of tests, such as mocking and consolidating utils.

  - 252b187 , dd1b205 : Adjusts some assumptions in the shared tests for all Subpartition implementations to cover only what want to universally assume for all implementations.
  Some tests become thus specific to the `PipelinedSubpartition` implementation.
  
  - 053aefc , bd4301a : Remove old SpillableSubpartition code and move code in the base class that was specific to pipelined and previous spillable implementation into the `PipelinedSubpartition` class.

  - 9cc9579 : Implementation of new `BoundedBlockingSubpartition`
  
  - 13d6121 : Blocking partitions no longer occupy network buffers after writing, and thus tests that check the reaction to `releaseMemory()` calls are no longer applicable.
  
  - 33029c5 : Tests for the new implementation of `BoundedBlockingSubpartition`

## Verifying this change

The code can be tested with any batch job that does blocking data exchanges.
To force batch data exchanges, add `env.getConfig().setExecutionMode(ExecutionMode.BATCH_FORCED);`
to any DataSet program. Otherwise many data exchanges will currently still be pipelined.

To test this in-IDE, take the *WordCount* example and add the above line.
To test the network code, use a local environment and pass a configuration that starts multiple local TaskManagers.

This adds also several untit tests:
  - `BoundedBlockingSubpartitionTest` for common behavior tests around releasing and buffer disposal.
  - `BoundedBlockingSubpartitionWriteReadTest` for producing data into a partition and consuming it. This test also tests that partitions can be consumed multiple times and can be read by multiple concurrent readers.

  - The serializers: **no**
  - The runtime per-record code paths (performance sensitive): **yes, batch data exchange**
  - Anything that affects deployment or recovery: JobManager (and its components), Checkpointing, Yarn/Mesos, ZooKeeper: **yes, batch intermediate results used during recovery**

## Documentation


rmetzger:
@flinkbot approve description

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/PartitionTestUtils.java
```diff
@@ -0,0 +1,77 @@
...
+import org.apache.flink.runtime.taskmanager.TaskActions;
+
+import static org.mockito.Mockito.mock;
+
+/**
+ * This class should consolidate all mocking logic for ResultPartitions.
+ * While using Mockito internally (for now), the use of Mockito should not
+ * leak out of this class.
+ */
+public class PartitionTestUtils {
```
pnowojski: Could you drop mockito usage from this class? You are not only consolidating the mockito usage, but also later you are introducing new tests that are using this methods.

I think all of those `mock`'ed classes, are quite easy to construct manually (for example `new ResultPartitionManager()`, `IOManager` or `new NoOpTaskActions()`).

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/PartitionTestUtils.java
```diff
@@ -0,0 +1,77 @@
...
+
+	public static ResultPartition createMockPartition() {
+		return mock(ResultPartition.class);
```
pnowojski: Can not this method recursively call `createPartition(...)` with some defaults? (% ioManager which has to be closed, but this also could be solved by making this class a factory class with `close()` closing the `IOManager` and making those methods non static.

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartitionWriteReadTest.java
```diff
@@ -0,0 +1,223 @@
...
+		final ResultSubpartitionView reader = subpartition.createReadView(() -> {});
+		readLongs(reader, numLongs, subpartition.getBuffersInBacklog());
+
+		// cleanup
+		reader.releaseAllResources();
+		subpartition.release();
+	}
+
+	@Test
+	public void testWriteAndReadDataAcrossRegions() throws Exception {
```
pnowojski: `public testWriteAndReadDataAcrossRegions()` and `public testWriteAndReadData()` could be deduplicated to `private testWriteAndReadData(int numLongs, int regionSize)`.

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartitionWriteReadTest.java
```diff
@@ -0,0 +1,223 @@
...
+		reader.releaseAllResources();
+		subpartition.release();
+	}
+
+	@Test
+	public void testRead10ConsumersSequential() throws Exception {
+		final int numLongs = 10_000_000;
+
+		// setup
+		final BoundedBlockingSubpartition subpartition = createSubpartition();
```
pnowojski: This 3 setup lines are also could be deduplicated in all of those tests (extracted to for example `createAndFillSubpartition(int numLongs, int regionSize)`)

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/SubpartitionTestBase.java
```diff
@@ -36,10 +38,24 @@
...
+	public void createReaderAfterDispose() throws Exception {
+		final ResultSubpartition subpartition = createSubpartition();
+		subpartition.release();
+
+		try {
+			subpartition.createReadView(() -> {});
+			fail("expected an exception");
+		}
+		catch (IllegalStateException e) {
+			// expected
```
pnowojski: nit: it would be better to check the exception message/error code to distinguish between the really expected `IllegalStateException` vs something else (this is quite generic exception)

pnowojski:
Could someone from your team @zhijiangW look at this and do the final review? Also this definitely deserves a benchmark coverage. Could we add the benchmarks referenced in the ticket to this PR and `flink-benchmark` repository?

zhijiangW:
@pnowojski I already watched this ticket before and am really looking forward to review this PR.
Because of many other works in hand atm, the review is delayed some days. I think I would take it next week (public holiday this week) and my colleague YingjieCao would also review together and supplement the benchmark to cover.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartitionReader.java
```diff
@@ -0,0 +1,191 @@
...
+/**
+ * The reader (read view) of a BoundedBlockingSubpartition.
+ */
+class BoundedBlockingSubpartitionReader implements ResultSubpartitionView {
+
+	/** The result subpartition that we read. */
+	private final BoundedBlockingSubpartition parent;
+
+	/** Further byte buffers, to handle cases where there is more data than fits into
+	 * one mapped byte buffer (2GB = Integer.MAX_VALUE). */
```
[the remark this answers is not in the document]
StephanEwen: There is a case that artificially lowers the limit to 10MB (see extra test constructor in `BoundedBlockingSubpartition`) and tests going across various buffers.

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartitionWriteReadTest.java
StephanEwen: I did not want to do this, to leave the common case without supplying an explicit region size. The fact that a region size parameter is exposed is purely to optimize the cross region tests, but should not bleed into the default tests.

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/SubpartitionTestBase.java
StephanEwen: I think this is targeted enough here to be fine.
We often do not have messages in IllegalArgumentExceptions, if the `checkState` is meant as the equivalent of an always active assert statement. This is mainly for tracing error root causes in case of legitimate bugs, and the stack trace defines the exception even better than any error message.

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartitionWriteReadTest.java
StephanEwen: I was looking into alternative byte buffer implementations that would not have that limit, and then we could simply remove that parameter and test, rather than altering other tests as well.

There is a trade-off here between "the test should only see methods/parameters related to what it is testing" and code de-duplication. In that case, I leaned towards the first aspect more.

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartitionWriteReadTest.java
StephanEwen: can do

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/PartitionTestUtils.java
StephanEwen: Can add a later cleanup commit here. Did not want to go too deep down the mockito hole when rearranging this.

StephanEwen: Using a proper IOManager is tricky, because it needs I/O Worker Thread shutdown. We could use a test mock here that does not start threads.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/consumer/SingleInputGate.java
```diff
@@ -565,7 +565,13 @@ public void requestPartitions() throws IOException, InterruptedException {
 			return Optional.of(new BufferOrEvent(buffer, currentChannel.getChannelIndex(), moreAvailable));
 		}
 		else {
-			final AbstractEvent event = EventSerializer.fromBuffer(buffer, getClass().getClassLoader());
+			final AbstractEvent event;
+			try {
+				event = EventSerializer.fromBuffer(buffer, getClass().getClassLoader());
+			}
+			finally {
+				buffer.recycleBuffer();
```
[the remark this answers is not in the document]
StephanEwen: can look into this

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/PartitionTestUtils.java
StephanEwen: The best solution would be to implement a synchronous I/O Manager without additional threads, actually. Maybe that is not too hard, I can try and look into that some time...

StephanEwen:
I tried to keep each commit self-contained.

However, there was some later rebasing involved with some merge conflicts which might have messed this up. Which commit did you find not self-contained?

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/PartitionTestUtils.java
pnowojski: I would still prefer not to expand the mockito usage. Adding a trivial no-op implementation of `IOManager` is easy and we already have one that throws exceptions (`TestIOManager`).  Rest of the fields are already taken care of:

```
		return new ResultPartition(
			"TestTask",
			new NoOpTaskActions(),
			new JobID(),
			new ResultPartitionID(),
			PIPELINED,
			1,
			1,
			new ResultPartitionManager(),
			new NoOpResultPartitionConsumableNotifier(),
			new NoOpIOManager(),
			false);
```

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartition.java
```diff
@@ -0,0 +1,416 @@
...
+		private void rollOverToNextBuffer() throws IOException {
+			if (currentBuffer != null) {
+				// we need to remember the original buffers, not any slices.
+				// slices have no cleaner, which we need to trigger explicit unmapping
+				fullBuffers.add(currentBuffer.complete());
+			}
+
+			final ByteBuffer mapped = file.map(MapMode.READ_WRITE, nextMappingOffset, mappingSize);
+			currentBuffer = new Writer(mapped);
+			nextMappingOffset += mappingSize;
```
wsry: The previous mapped bytebuffer may not be full, this may lead to file holes between data blocks, though the read view can be created correctly, is it by design.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartition.java
```diff
@@ -0,0 +1,416 @@
...
+			Files.delete(filePath);
+		}
+
+		/**
+		 * Gets the number of bytes of all written data (including the metadata in the buffer headers).
+		 */
+		long getSize() {
+			long size = 0L;
+			for (ByteBuffer bb : fullBuffers) {
+				size += bb.capacity();
```
wsry: The size may not be right if the bytebuffer is not full, using limit instead of capacity is better.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartition.java
```diff
@@ -0,0 +1,416 @@
...
+
+	/** All created and not yet released readers. */
+	@GuardedBy("lock")
+	private final Set<BoundedBlockingSubpartitionReader> readers;
+
+	/** Counter for the number of data buffers (not events!) written. */
+	private int numDataBuffersWritten;
+
+	/** The counter for the number of data buffers and events. */
+	private int numBuffersAndEventsWritten;
```
Ryantaocer: Sum of data buffers number and events?  The var comments could be more clear.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartition.java
```diff
@@ -0,0 +1,416 @@
...
+	}
+
+	// ------------------------------------------------------------------------
+
+	/**
+	 * Checks if writing is finished.
+	 * Readers cannot be created until writing is finished, and no further writes can happen after that.
+	 */
+	public boolean isFinished() {
+		return isFinished;
```
Ryantaocer: Could add `volatile` for `isFinished` and `isReleased` or use AtomicBoolean type for thread safety.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartition.java
```diff
@@ -0,0 +1,416 @@
...
+	}
+
+	@Override
+	public boolean add(BufferConsumer bufferConsumer) throws IOException {
+		if (isFinished()) {
+			bufferConsumer.close();
+			return false;
+		}
+
+		flushCurrentBuffer();
```
Ryantaocer: Should catch `IOException` from `flushCurrentBuffer` just like `flush()`?

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartition.java
```diff
@@ -0,0 +1,416 @@
...
+		currentBuffer = bufferConsumer;
+		return true;
+	}
+
+	@Override
+	public void flush() {
```
Ryantaocer: Could add synchronized (lock) here.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartitionMemory.java
```diff
@@ -0,0 +1,130 @@
...
+		public boolean writeBuffer(Buffer buffer) {
+			final ByteBuffer memory = this.memory;
+			final int bufferSize = buffer.getSize();
+
+			if (memory.remaining() < bufferSize + HEADER_LENGTH) {
+				return false;
+			}
+
+			memory.putInt(buffer.isBuffer() ? HEADER_VALUE_IS_BUFFER : HEADER_VALUE_IS_EVENT);
+			memory.putInt(bufferSize);
```
Ryantaocer: these two `int` are written as metadata at the front of memory have total lenght `4+4=8` denoted by `HEADER_LENGTH `, am I right?

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartitionMemory.java
```diff
@@ -0,0 +1,130 @@
...
+final class BoundedBlockingSubpartitionMemory {
+
+	// all fields and methods below here have package-private access to avoid bridge
+	// methods when accessing them from the nested classes
+
+	static final int HEADER_LENGTH = 8;
+
+	static final int HEADER_VALUE_IS_BUFFER = 0;
+
+	static final int HEADER_VALUE_IS_EVENT = 1;
```
Ryantaocer: what's the difference between these two types of header `HEADER_VALUE_IS_BUFFER` and `HEADER_VALUE_IS_EVENT`? what is the meaning of it if a ByteBuffer is `HEADER_VALUE_IS_EVENT` ?

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartitionMemory.java
```diff
@@ -0,0 +1,130 @@
...
+				return null;
+			}
+
+			final int header = memory.getInt();
+			final int size = memory.getInt();
+
+			memory.limit(memory.position() + size);
+			ByteBuffer buf = memory.slice();
+			memory.position(memory.limit());
+			memory.limit(memory.capacity());
```
Ryantaocer: here, is it necessary to `memory.limit(memory.capacity())` ?

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartitionReader.java
```diff
@@ -0,0 +1,191 @@
...
+			return next;
+		}
+
+		if (!furtherData.hasNext()) {
+			data = null;
+			return null;
+		}
+
+		data = new BoundedBlockingSubpartitionMemory.Reader(furtherData.next());
+		return fillNextBuffer();
```
Ryantaocer: Is it recursively to find next buffer untill it is not null? We could use a simple for-loop to avoid temporary `BoundedBlockingSubpartitionMemory.Reader` creation.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartitionReader.java
```diff
@@ -0,0 +1,191 @@
...
+
+	@Override
+	public void notifySubpartitionConsumed() throws IOException {
+		parent.onConsumedSubpartition();
+	}
+
+	@Override
+	public void releaseAllResources() throws IOException {
+		// mark released and prevent double release
+		if (isReleased) {
```
Ryantaocer: We should add lock here to avoid mulitple thread entering when `isReleased` is false.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartition.java
```diff
@@ -0,0 +1,416 @@
...
+		// need to do this discouraged pattern of runtime exception wrapping
+		try {
+			flushCurrentBuffer();
+		}
+		catch (IOException e) {
+			throw new FlinkRuntimeException(e.getMessage(), e);
+		}
+	}
+
+	private void flushCurrentBuffer() throws IOException {
```
zhijiangW: `flushCurrentBuffer` might be called by both task thread and flusher thread, so it exists race condition here. The variable `currentBuffer` should be visible for both threads.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartition.java
```diff
@@ -0,0 +1,416 @@
...
+		if (currentBuffer != null) {
+			writeAndCloseBufferConsumer(currentBuffer);
+			currentBuffer = null;
```
zhijiangW: It might be better to put `currentBuffer = null`  into finally part.
If `writeAndCloseBufferConsumer` is trigged by task thread and throws exception, the following flusher thread do not need to flush this `currentBuffer` again to bring unnecessary trouble. (This `currentBuffer` is already recycled during exception)

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartition.java
```diff
@@ -0,0 +1,416 @@
...
+		}
+	}
+
+	@Override
+	public ResultSubpartitionView createReadView(BufferAvailabilityListener availability) throws IOException {
+		synchronized (lock) {
+			checkState(!isReleased, "data partition already released");
+			checkState(isFinished, "writing of blocking partition not yet finished");
+
+			availability.notifyDataAvailable();
```
zhijiangW: `availability.notifyDataAvailable()` is better to be called as late as possible, before return reader.
Because after notification of netty stack, the netty thread would call `subpartitionView.isAvailable()` from `CreditBasedSequenceNumberingViewReader`, but this `subpartitionView` might not be returned to `CreditBasedSequenceNumberingViewReader` at that moment.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartition.java
```diff
@@ -0,0 +1,416 @@
...
+			ResultPartition parent,
+			Path filePath,
+			int maxMMapRegionSize) throws IOException {
+
+		super(index, parent);
+
+		final FileChannel fc = FileChannel.open(filePath,
+				StandardOpenOption.READ, StandardOpenOption.WRITE, StandardOpenOption.CREATE_NEW);
+
+		this.memory = new MemoryMappedBuffers(filePath, fc, maxMMapRegionSize);
```
zhijiangW: The `FileChannel` could be opened inside the constructor of `MemoryMappedBuffers` via `filePath`, so no need to pass two parameters here?

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartitionReader.java
```diff
@@ -0,0 +1,191 @@
...
+		if (current == null) {
+			// as per contract, we must return null when the reader is empty,
+			// but also in case the reader is disposed (rather than throwing an exception)
+			return null;
+		}
+		if (current.isBuffer()) {
+			bufferBacklog--;
+		}
+
+		Buffer next = fillNextBuffer();
```
zhijiangW: Why not use `nextBuffer = fillNextBuffer()` directly?

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartitionReader.java
```diff
@@ -0,0 +1,191 @@
...
+
+		return new BufferAndBacklog(
+				current,
+				next != null,
+				bufferBacklog,
+				next != null && !next.isBuffer());
+	}
+
+	@Nullable
+	private Buffer fillNextBuffer() throws IOException {
```
zhijiangW: never throw `IOException` here?

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartitionReader.java
```diff
@@ -0,0 +1,191 @@
...
+		}
+
+		Buffer next = fillNextBuffer();
+		nextBuffer = next;
+
+		return new BufferAndBacklog(
+				current,
+				next != null,
+				bufferBacklog,
+				next != null && !next.isBuffer());
```
zhijiangW: Use `nextBufferIsEvent()` to replace `next != null && !next.isBuffer()` here?

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartitionReader.java
```diff
@@ -0,0 +1,191 @@
...
+
+	@Override
+	public void releaseAllResources() throws IOException {
+		// mark released and prevent double release
+		if (isReleased) {
+			return;
+		}
+		isReleased = true;
+
+		// nulling these fields means thet read method and will fail fast
```
zhijiangW: thet typo?

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/ResultPartition.java
```diff
@@ -466,4 +465,35 @@ private void notifyPipelinedConsumers() {
 			hasNotifiedPipelinedConsumers = true;
 		}
 	}
+
+	private static void initializeBoundedBlockingPartitions(
```
zhijiangW: I guess this method would have other usages such as tests in future? Otherwise we do not need to pass these parameters here.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartition.java
```diff
@@ -0,0 +1,416 @@
...
+	}
+
+	@Override
+	public boolean isReleased() {
+		return isReleased;
+	}
+
+	@Override
+	public boolean add(BufferConsumer bufferConsumer) throws IOException {
+		if (isFinished()) {
```
zhijiangW: It seems better to check `isReleased()` instead of `isFinished()` here. From functional aspect, they could both work, because we also tag `isFinished` flag during `release()`. But `isReleased()` here seems easy to understand, because in another case of `finish()` to tag `isFinished` flag, this `add` should never be called.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartition.java
```diff
@@ -0,0 +1,416 @@
...
+			}
+		}
+		finally {
+			bufferConsumer.close();
+		}
+	}
+
+	@Override
+	public void finish() throws IOException {
+		synchronized (lock) {
```
zhijiangW: Base on existing processes, there should be no race conditions between `finish()` and `release()`. That means if `finish()` is called by task thread, the `release()` should never be called via `failPartition` by rpc thread meanwhile.

So we might not need `synchronized` in `finish()` method. `release()` and `createReadView()` could be `synchronized` for rpc and netty threads.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartitionReader.java
```diff
@@ -0,0 +1,191 @@
...
+		this(parent, Collections.singleton(data), numBuffers);
+	}
+
+	/**
+	 * The general constructor that takes a queue of buffers containing the data.
+	 */
+	BoundedBlockingSubpartitionReader(
+			BoundedBlockingSubpartition parent,
+			Iterable<ByteBuffer> data,
+			int numBuffers) throws IOException {
```
zhijiangW: Seems never throw `IOException` here?

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartition.java
```diff
@@ -0,0 +1,416 @@
...
+			checkState(!isReleased, "data partition already released");
+			checkState(!isFinished, "data partition already finished");
+
+			isFinished = true;
+			flushCurrentBuffer();
+			writeAndCloseBufferConsumer(EventSerializer.toBufferConsumer(EndOfPartitionEvent.INSTANCE));
+			memory.finishWrite();
```
zhijiangW: `memory.finishWrite()` might be in finally part.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartition.java
StephanEwen: Right, it should be limit. Need to add a test to catch that.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartition.java
StephanEwen: That is by design. If you fill the buffers fully in the writer, the reader need to stitch together buffers across boundaries, which I wanted to avoid. The complexity seems lower this way and the "holes" in the file should be very small, meaning less than one buffer size.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartition.java
StephanEwen: You mean "sum of counters"? Sounds easier the way it is, in my opinion.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartition.java
StephanEwen: Some critical sections are under lock (finish / dispose) and there volatile makes no difference.
For all other section, it adds no guarantees.
I see not what it would help, but it would introduce a performance penalty.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartitionMemory.java
StephanEwen: yes, it is, the code will not work without

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartitionReader.java
StephanEwen: This is at buffer rollover, which once per 2GB. Code simplicity is more important in that context over avoiding a single object creation.

StephanEwen:
Some comment on thread-safety and synchronization:

We don't assume arbitrary concurrent access, there is a single writer thread for the `BoundedBlockingSubpartition` and one thread that operates the reader. If that happens nonetheless (bug in the system), exceptions occur. That is preferable to the performance overhead of locks everywhere.

The only concurrency is concurrent reader creation and concurrent release() calls. We need to guard those to avoid segfaults due to accessing unmapped memory. That's what the implementation does.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartition.java
StephanEwen: You mean in a finally block so that it also happens when exceptions are thrown? Why that? When an exception happens, we only want to release later.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartitionReader.java
StephanEwen: True, can remove that from the signature. Leftover from a previous version.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartition.java
StephanEwen: I think we want to assume here that this is either always coming from the same thread, or it is externally synchronized. We should assume single writer here. The notion of a flusher thread should not leak into this implementation.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartition.java
StephanEwen: Can do that, but in any case, once an exception on the writer side occurred, we cannot support any good operation any more. So this would by trying to improve one aspect, but since the whole would still not work, I would prefer to avoid this, for simplicity. If we put finally there, we suggest to the code reader that we recover something, but we actually do not recover the functionality.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartition.java
StephanEwen: I see the point, but this seems like making assumptions about a specific user of this class, which one should usually avoid.

My understanding is that on the netty side, this should go through the same event loop thread, so the `notifyDataAvailable()` enqueues a Runnable into the event loop, which should be fine. Is that correct?

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartition.java
StephanEwen: Do you think there is a concrete abstraction or separation-or-concerns problem behind this?

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartitionReader.java
StephanEwen: Caching a reference on the stack is often a good pattern. Avoids implicit multiple "this." de-referencing. Good for performance and consistency. Not really required here, but also seems simple enough to not hurt readability.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartitionReader.java
StephanEwen: yes

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartitionReader.java
StephanEwen: Can remove that, leftover from previous version.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartitionReader.java
StephanEwen: Similar as above, heap references versus stack references. Not a critical optimization, but seems to also not hurt here.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/ResultPartition.java
StephanEwen: yes

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartition.java
StephanEwen: We need to check for `isFinished()` in any case, as per the contract that buffers are released in that case.
We also need to check whether the partition is released, the question is whether to do this explicitly or collapse the checks (as I did here). This is covered by tests, so I felt comfortable collapsing the checks.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartition.java
StephanEwen: We probably don't strictly need it here.

The reason why I added it is that its only called once (so no performance issue) and it is part of the bookkeeping that ensures consistency around the bookkeeping that triggers the file unmapping eventually. We could remove it, to make it clear we don't assume multiple writer threads.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartition.java
zhijiangW: Yes, it is true. `notifyDataAvailable` and `createReadView` would be from the same netty thread. So we could ignore this issue.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartition.java
zhijiangW: I just felt a bit redundant in the constructor of `MemoryMappedBuffers` before.  But the advantage is that it seems more flexible for passing `FileChannel` outside for defining different `StandardOpenOption` in other possible usages such as tests. So I accept the current way now.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartition.java
zhijiangW: Thanks for the explanation! That makes sense for me.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartition.java
zhijiangW: I just wondered the file handle would not be closed if `writeAndCloseBufferConsumer` throws any exceptions before. After double check the codes, the `release()` method would be called after task failed to release all the resources. So it is no problem as current way here.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartition.java
zhijiangW: Yes, that sounds reasonable. My previous concern was the root exception might bring other following exceptions which might bring noise for log tracing. Actually this has two sides so I understand your thought now.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartition.java
zhijiangW: Yes, it has no problem if the flusher thread never touches here.

zhijiangW:
Hey @StephanEwen , after thinking through your concerns of when and how to release readers, I tried to understand your proposed three options. In general I also prefer to the second option because of the following reasons:

- When partition is released explicitly, that probably means the corresponding reader/view should not be consumable any more from the semantic aspect. So the first option to make existing readers still consumable is just limited by previous mechanism/implementation of releasing view.

- `CancelPartition` is not always reliable by design ATM. It is only determined by internal TCP mechanism, and we have no other ack mechanisms in flink stack. Different from `BufferResponse` we add the sequence number to avoid the message lost in network.  In previous failover strategy, if the consumer fails and sends `CancelPartition` to the producer, even though this cancel message is not received by producer, the JM could also cancel the producer to avoid this issue.

-  Even though we enhance to make `CancelPartition` reliable, we could not determine the reasonable time when the reader is disposed. Because `CancelPartition` would be sent when consumer fails or `EndOfPartitionEvent` is read from some `RemoteInputChannel`. There might exist the scenario to cause reader disposed very delay.

- In third option we still want to release view via netty thread, then we set the flag to be seen by netty in the next read. This is very similar with the way when consumer netty is aware of any exception/errors, it would call `RemoteInputChannel#onError` to trigger task thread sensitive during `getNextBuffer`. But I am not sure whether we could confirm the next read happen or when to happen. In the case of `RemoteInputChannel#onError`, it must happen via following call `notifyChannelNonEmpty`. Also there seems a bit hacky in logic for this option as you mentioned .

In summary I think the first option could still make sense in current PR for most of the cases. It is easy to not consider the specific references. Then we could further improve it future if necessary after finding any serious problems in practice.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartition.java
StephanEwen: Adding synchronization arbitrarily to some methods is not the right approach.
The class follows a strict contract (which I will describe more detailed in the JavaDocs)
  - Single threaded writer
  - Multiple single threaded readers

The only synchronized sections in that model are where reader creation and disposal meet.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartition.java
```diff
@@ -0,0 +1,416 @@
...
+					.map(ByteBuffer::slice)
+					.collect(Collectors.toList());
+		}
+
+		/**
+		 * Finishes the current region and prevents further writes.
+		 */
+		void finishWrite() throws IOException {
+			fullBuffers.add(currentBuffer.complete());
+			currentBuffer = null; // fail further writes fast
```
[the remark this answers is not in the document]
StephanEwen: Because that class is a private inner class, I assume a collaboration on the contract with the outer class. The outer class has checks in place that this situation cannot occur.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartitionMemory.java
StephanEwen: yes, that is correct

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartitionMemory.java
StephanEwen: This is used to tag the result buffer payload accordingly as a buffer or an event.
The logic is similar to the spilling logic in the previous implementation.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartitionMemory.java
StephanEwen: If we don't set the limit forward, the successive `readInt()` calls for the header will fail.
I played around with some alternatives, but could not find any where the code overall would become simpler.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartitionReader.java
StephanEwen: The threading model here assumes only a singe thread as the user of the reader, so we do not need that lock.

All operations here are also idempotent, except the final `parent.releaseReaderReference()`, which is actually thread-safe, so this method can work without synchronization.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartition.java
StephanEwen: Sure, will move it to that constructor.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/PageSizeUtil.java
```diff
@@ -0,0 +1,108 @@
...
+			ExceptionUtils.rethrowIfFatalError(t);
+			return PAGE_SIZE_UNKOWN;
+		}
+	}
+
+	/**
+	 * Tries to get the system page size. If the page size cannot be determined, this
+	 * returns the {@link #DEFAULT_PAGE_SIZE}.
+	 */
+	public static int getSystemPageSizeOrDefault() {
```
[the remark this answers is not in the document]
StephanEwen: This util is meant as a more general util

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/PageSizeUtil.java
```diff
@@ -0,0 +1,108 @@
...
+		final int pageSize = getSystemPageSize();
+		return pageSize == PAGE_SIZE_UNKOWN ? DEFAULT_PAGE_SIZE : pageSize;
+	}
+
+	/**
+	 * Tries to get the system page size. If the page size cannot be determined, this
+	 * returns the {@link #CONSERVATIVE_PAGE_SIZE_MULTIPLE}.
+	 */
+	public static int getSystemPageSizeOrConservativeMultiple() {
```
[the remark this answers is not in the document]
StephanEwen: This util is meant as a more general util

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/PageSizeUtil.java
```diff
@@ -0,0 +1,108 @@
...
+import java.security.AccessController;
+import java.security.PrivilegedAction;
+
+/**
+ * Utility for accessing the system page size.
+ */
+final class PageSizeUtil {
+
+	/** Value indicating an unknown page size. */
+	public static final int PAGE_SIZE_UNKOWN = -1;
```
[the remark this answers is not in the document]
StephanEwen: will fix that

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartitionMemory.java
```diff
@@ -0,0 +1,130 @@
...
+
+		private final ByteBuffer memory;
+
+		Reader(ByteBuffer memory) {
+			this.memory = checkAndConfigureByteBuffer(memory);
+		}
+
+		@Nullable
+		public Buffer sliceNextBuffer() {
+			final ByteBuffer memory = this.memory;
```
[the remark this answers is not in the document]
StephanEwen: Localize the scope. Makes it clear to the reader that this is strictly local access, and often helps Java avoid some instructions (avoid indirection via implicit `this.`)

StephanEwen:
@zhijiangW Thanks for the detailed review and good comments.

StephanEwen:
@eaglewatcherwb This implementation makes no assumption when a partition will be released. That is the responsibility of the network environment.

StephanEwen:
Thank you for the reviews.

I refactored some code:
  - Pulled out the `MemoryMappedBuffers`) class to better test `size()` and region spanning writes and avoid leaking the region configuration into the `BoundedBlockingSubpartition`.
  - Added comments to make the treading model assumptions more clear
  - Addressed comments on typos, etc.

The code staged to be merged is in https://github.com/StephanEwen/incubator-flink/commit/a7cf24383be9f310fb5ccc5a032721421fa45791

I think overall, this is now in a really good shape. Waiting for Travis to pass, will then merge this.

Followups are to to actually make use of the "multiple read" functionality.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartitionReader.java
```diff
@@ -0,0 +1,191 @@
...
+	/** The remaining number of data buffers (not events) in the result. */
+	private int bufferBacklog;
+
+	/** Flag whether this reader is released. Atomic, to avoid double release. */
+	private boolean isReleased;
+
+	/**
+	 * Convenience constructor that takes a single buffer.
+	 */
+	BoundedBlockingSubpartitionReader(
```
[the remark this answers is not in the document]
StephanEwen: have removed it

StephanEwen:
@zhijiangW Our reasoning about the three options seems similar.

My understanding is that even if CancelPartition is not reliable, it can only fail when the TCP connection fails. In that case, all associated SubpartitionViews are released immediately, which should work.

StephanEwen:
Manually merged in a7cf24383be9f310fb5ccc5a032721421fa45791

PR #8302, opened 2019-04-28
[FLINK-12269][table-blink] Support Temporal Table Join in blink planner and runtime

wuchong:

Support translate "FOR SYSTEM_TIME AS OF" query into temporal table join for both Batch and Stream.

## Brief change log

  - Introduce some API interface to support lookup table source
    - `LookupableTableSource`, `AsyncTableFunction`, `DefinedPrimaryKey`, `DefinedTableIndex`, `TableIndex`, `LookupConfig`
 - Support for generating optimized logical & physical plan for temporal table join for STREAM and BATCH.
 - Introduce temporal table join operators and async join operators for runtime.
 
Some differences between this PR and blink branch.

 - Support returns multiple lines for async temporal table join
 - Only allow `FOR SYSTEM_TIME AS OF` on left table's proctime field, not a `PROCTIME()` builtin function. This makes syntax clean.
 - Lookup function support external types (i.e. `String`, `Timestamp`...).

## Verifying this change

- Add several validation tests, plan tests , and integration tests.

  - The serializers: (no)

## Documentation


wuchong:
Hi @godfreyhe , I add the `JoinPushExpressionsRule.INSTANCE` rule into predicate pushdown stage to make sure the calculation expression can be pushed down (to make sure `ON a+1 = b` is an equi join).  That's why several results of plan test are changed. I have checked, the changes are reasonable.

It would be nice if you can have a look too.

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/runtime/join/temporal/TemporalTableJoinAsyncRunner.java
```diff
@@ -0,0 +1,268 @@
...
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+import java.util.concurrent.ArrayBlockingQueue;
+import java.util.concurrent.BlockingQueue;
+
+/**
+ * The async join runner to lookup the dimension table.
+ */
+public class TemporalTableJoinAsyncRunner extends RichAsyncFunction<BaseRow, BaseRow> {
```
KurtYoung: add tests for these runners?

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/codegen/CalcCodeGenerator.scala
```diff
@@ -69,6 +71,42 @@ object CalcCodeGenerator {
     new OneInputOperatorWrapper(genOperator)
   }
 
+  private[flink] def generateFunction[T <: Function](
```
KurtYoung: why don't adding this to `FunctionCodeGenerator`?

flink-table/flink-table-planner-blink/src/test/resources/org/apache/flink/table/plan/batch/sql/join/TemporalTableJoinTest.xml
```diff
@@ -0,0 +1,335 @@
...
++- TemporalTableJoin(table=[TestTemporalTable(id, name, age)], joinType=[InnerJoin], async=[false], on=[a=id], select=[a, b, c, id, name, age])
+   +- BoundedStreamScan(table=[[T0]], fields=[a, b, c])
+]]>
+    </Resource>
+  </TestCase>
+  <TestCase name="testJoinTemporalTableWithFilterPushDown">
+    <Resource name="sql">
+      <![CDATA[
+SELECT * FROM MyTable AS T
+JOIN temporalTest FOR SYSTEM_TIME AS OF T.proctime AS D
```
KurtYoung: Why we want to attach the proctime to table `T`, as opposed to just use `proctime()`?

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/codegen/CalcCodeGenerator.scala
wuchong: Because it actually generate code for a `CalcProgram`, it needs to access the private method `generateProcessCode` in this file. And `FunctionCodeGenerator` only accepts code body as  parameter, not the `CalcProgram.

What do you think about renaming the method name to `generateCalcFunction` to align with `generateCalcOperator` ?

flink-table/flink-table-planner-blink/src/test/resources/org/apache/flink/table/plan/batch/sql/join/TemporalTableJoinTest.xml
wuchong: Because I think using `proctime()` here is a little semantic ambiguous. Temporal table join is joining the snapshot of temporal table, the version of snapshot is determined by left table's time field. Say `left.proctime`, it means joining the current version of temporal table.

But `temporalTest FOR SYSTEM_TIME AS OF PROCTIME()` is not related to left table, it only returns one (latest) version of temporal table and then be joined with left table. It is the same with 
```sql
SELECT * 
FROM left AS L, 
JOIN (
  SELECT * FROM temporalTest FOR SYSTEM_TIME AS OF PROCTIME()
) AS R ON L.id = R.id
```

That's why I want to restrict it a bit more in the first version. We can discuss it, and can support `PROCTIME()` if we think it is needed. It should be supported easily with several lines change.

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/runtime/join/temporal/TemporalTableJoinAsyncRunner.java
wuchong: Sure

flink-table/flink-table-planner-blink/src/test/resources/org/apache/flink/table/plan/batch/sql/join/TemporalTableJoinTest.xml
KurtYoung: I think you are right. The example you gave will cause different behavior with what we expected. The `protime()` can be translated to `the processing time we begin to scan right table`, and in this case, right table will become a static table. This is not we want.

wuchong:
Hi @KurtYoung , I have addressed the review comments.

1. add unit tests for all join runners
2. fix code generation problem when temporal join with udf filter (a bug found recently in internal)

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/sources/LookupableTableSource.java
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * A {@link TableSource} which supports for lookup accessing via key column(s).
+ * For example, MySQL TableSource can implement this interface to support lookup accessing.
+ * When temporal join this MySQL table, the runtime behavior could be in a lookup fashion.
+ *
+ * @param <T> type of the result
+ */
+@PublicEvolving
+public interface LookupableTableSource<T> extends TableSource<T> {
```
KurtYoung: What's the relationship between this `LookupableTableSource` and `DefinedIndexes` and `DefinedPrimaryKey`? Does this table source support getting lookup function for random keys, even if the columns have no index on them or not the primary key?

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/plan/rules/physical/common/CommonTemporalTableJoinRule.scala
```diff
@@ -0,0 +1,134 @@
...
+      operand(classOf[FlinkLogicalRel], any()),
+      operand(classOf[FlinkLogicalSnapshot],
+        operand(classOf[TableScan], any()))),
+    description)
+          with CommonTemporalTableJoinRule {
+
+  override def matches(call: RelOptRuleCall): Boolean = {
+    val join = call.rel[FlinkLogicalJoin](0)
+    val tableScan = call.rel[TableScan](3)
+    matches(join, tableScan)
```
KurtYoung: we should also check if it's snapshotted by proctime?

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/sources/LookupableTableSource.java
wuchong: Currently, the lookup key of `LookupableTableSource` must be either primary key or index. Therefore, a concrete `LookupableTableSource` should also implement `DefinedIndexes` or `DefinedPrimaryKey`.

> Does this table source support getting lookup function for random keys, even if the columns have no index on them or not the primary key?

No. It will throw exceptions.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/plan/rules/physical/common/CommonTemporalTableJoinRule.scala
wuchong: Currently, whether it is snapshotted by proctime or rowtime will both be translated into temporal table join. And will throw exception if it is rowtime when physical translation, because we don't support rowtime temporal join currently.

So I think we don't need to check proctime here.

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/sources/LookupableTableSource.java
KurtYoung: Do we have any better solution for this? It's hard for user to realize that `LookupableTableSource` should be used with `DefinedIndexes` or `DefinedPrimaryKey`

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/plan/rules/physical/common/CommonTemporalTableJoinRule.scala
KurtYoung: I noticed current TemporalTableJoin is a `SingleRel`, which is not suitable for further extension after we support scanning data into state and provide event time join. So i think it's inappropriate to translate snapshotted with rowtime to `TemporalTableJoin` and then throw exception inside this physical operator.

wuchong: Make sense. 

I'm also rethinking about the physical node name of `TemporalTableJoin` and `TemporalTableFunctionJoin` (i.e. temporal join in Flink). The node names are really confused to users. Actually, the `TemporalTableJoin` is joining a dimension table which is a SingleRel. 

So how about renaming `TemporalTableJoin` to `DimensionTableJoin`, and renaming `TemporalTableFunctionJoin` to `TemporalJoin`. And we can change the translation rule to:

- snapshot on proctime & table source supports `LookupableTableSource` ==> `DimensionTableJoin` 
- snapshot on rowtime & table source ONLY supports `LookupableTableSource` ==> exception
- snapshot on proctime/rowtime & table source supports scanning ==> `TemporalJoin` 

What do you think ?

KurtYoung: +1 to this idea. How about rename `DimensionTableJoin` to `LookupJoin`, and rename `TemporalJoin` to `TemporalTableJoin`. 

Conceptually `LookupJoin` is very similar with `NestedLoopJoin` but is a `SingleRel`.

wuchong: +1 to `LookupJoin`.

The only concern of `TemporalTableJoin` is that it is actually joining two streams, but there is a `Table` in the name. Do you think it matters?

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/sources/LookupableTableSource.java
wuchong: As discussed offline, we will keep `DefinedIndexes` and `DefinedPrimaryKey` in the first version. And integrate them into table source v2 interface in the future.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/plan/rules/physical/common/CommonTemporalTableJoinRule.scala
KurtYoung: Because i think TemporalTable is more consistent with sql standard. `TemporalJoin` will confuse others about what is temporal? It's not intuitive that this is actually used for temporal table join.  You can consider this name as `TemporalTableJoin ' Join`

wuchong: Ok, let's go forward with `LookupJoin` and `TemporalTableJoin`.

wuchong:
Hi @KurtYoung , I renamed the node name to `LookupJoin` and move the snapshot on proctime verifying to rules in the second commit: `rename to LookupJoin`.

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/runtime/collector/TableFunctionCollector.java
```diff
@@ -0,0 +1,89 @@
...
+	 *
+	 * @return True if {@link #collect(Object)} has been called.
+	 */
+	public boolean isCollected() {
+		return collected;
+	}
+
+	@Override
+	public void collect(T record) {
+		this.collected = true;
```
KurtYoung: Shouldn't this method call `collector.collect`?

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/runtime/collector/TableFunctionResultFuture.java
```diff
@@ -0,0 +1,68 @@
...
+	 * which will be used to cross join with the result of right table.
+	 */
+	public Object getInput() {
+		return input;
+	}
+
+	/**
+	 * Sets the current collector, which used to emit the final row.
+	 */
+	public void setCollector(ResultFuture<?> resultFuture) {
```
KurtYoung: setResultFuture?

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/runtime/collector/TableFunctionResultFuture.java
```diff
@@ -0,0 +1,68 @@
...
+		this.resultFuture = resultFuture;
+	}
+
+	/**
+	 * Gets the internal collector which used to emit the final row.
+	 */
+	public ResultFuture<?> getCollector() {
```
KurtYoung: getResultFuture?

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/runtime/collector/TableFunctionCollector.java
KurtYoung: and i just found that the sub-class of this class `RowToBaseRowCollector` is actually calling `      getCollector.asInstanceOf[Collector[BaseRow]].collect(result)`,

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/runtime/collector/TableFunctionResultFuture.java
```diff
@@ -0,0 +1,68 @@
...
+
+package org.apache.flink.table.runtime.collector;
+
+import org.apache.flink.api.common.functions.AbstractRichFunction;
+import org.apache.flink.streaming.api.functions.async.ResultFuture;
+
+/**
+ * The basic implementation of collector for {@link ResultFuture} in table joining.
+ */
+public abstract class TableFunctionResultFuture<T> extends AbstractRichFunction implements ResultFuture<T> {
```
KurtYoung: The responsibility of this class and `TableFunctionCollector` is not clean. It looks like they want to do some encapsulation but actually only contain some get and set methods. And the interface contract also seems not consistent, see the comment i left in `TableFunctionCollector`

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/runtime/join/lookup/AsyncLookupJoinRunner.java
```diff
@@ -0,0 +1,274 @@
...
+
+		// fetcher has copied the input field when object reuse is enabled
+		fetcher.asyncInvoke(input, outResultFuture);
+	}
+
+	public TableFunctionResultFuture<BaseRow> createFetcherResultFuture() throws Exception {
+		TableFunctionResultFuture<BaseRow> resultFuture = generatedResultFuture.newInstance(
+			getRuntimeContext().getUserCodeClassLoader());
+		FunctionUtils.setFunctionRuntimeContext(resultFuture, getRuntimeContext());
+		FunctionUtils.openFunction(resultFuture, new Configuration());
```
JingsongLi: Consider saving `parameters` of `open` for use here?

JingsongLi: Consider close `resultFuture`?

wuchong:
Comments addressed

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/runtime/collector/TableFunctionResultFuture.java
wuchong: The responsibility of them are used to generate a collector which will do some filters or projections on both left input row and right table row. That's why we have `setInput`, `getInput` and a `setCollector` to set the real underlying collector.

The only difference between them is `TableFunctionCollector` has an additional `collected` flag used to indicate whether right table is empty. Because UDTF may call `collect(T)` zero or several times, we need a way to know whether it is called zero times, so that we can emit an null row for left join. However, `ResultFuture.complete(Collection<OUT> result)` will be called exactly once, so that we don't need a flag to indicate the "zero call", an empty or null `result` is the "zero call".

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/runtime/join/lookup/AsyncLookupJoinRunner.java
wuchong: Sure

wuchong: Ok

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/runtime/collector/TableFunctionResultFuture.java
wuchong: OK

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/runtime/collector/TableFunctionCollector.java
KurtYoung: What i meant is this class already holds a `private Collector<?> collector;`, and also implements `Collector<T>` interface. In this `collect(T record)` method, why don't you collect the record with your own collector, but rely on the caller to first `getCollector()` and collect records?

wuchong: Because `TableFunctionCollector` abstract class does't know the what is the record type `T`. It might be a `BaseRow`, but can also be a `Row` (see `RowToBaseRowCollector`). As a result, `TableFunctionCollector` can't collect by himself, because he doesn't know what the result record is.

wuchong:
Rebased.

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/runtime/collector/TableFunctionCollector.java
KurtYoung: Then i think this class shouldn't be a collector, aka shouldn't implementing `Collector<T>` interface

wuchong: We have to keep the `collect(T record)` interface, because it used to accept UDTF's result value.

What about provide a `outputResult` method used to collect final result. And remove `getCollector()` method and `collect(T record)` implementation. 

```java
	public void outputResult(Object result) {
		this.collected = true;
		collector.collect(result);
	}
```

All the sub-classes can use `outputResult(result)` to collect final result instead of `getCollector().collect(result)`.

KurtYoung: yes, let's try this.

wuchong:
@KurtYoung  `TableFunctionCollector` has been refactored.

KurtYoung:
+1

PR #8304, opened 2019-04-28
[hotfix]Harden TextInputFormatTest to avoid failure by uncleaned files

Aitozi:

When I run test locally, it happened to failed and with the uncleaned temp file leaving in the tmp dir. So the test case can not pass ever until i delete the correspond files.

It use `deleteOnExit` to clean up the temp file created by test case, But the method may not work when JVM crash or killed. I think it may make the test case fragile.   

## Brief change log

Add the @Before and @After prepare / cleanup method

### DIFF
--- a/flink-java/src/test/java/org/apache/flink/api/java/io/TextInputFormatTest.java
+++ b/flink-java/src/test/java/org/apache/flink/api/java/io/TextInputFormatTest.java
@@ -22,6 +22,7 @@ import org.apache.flink.configuration.Configuration;
 import org.apache.flink.core.fs.FileInputSplit;
 import org.apache.flink.core.fs.FileSystem;
 import org.apache.flink.core.fs.Path;
+import org.apache.flink.util.TestLogger;
 
 import org.junit.Rule;
 import org.junit.Test;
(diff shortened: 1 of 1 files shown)


Aitozi:
And I see there are quite a lot other tests rely on `deleteOnExit` to do clean up. Should I do the fix for them, too?  Please take a look when you free , Thanks @zentol .

flink-java/src/test/java/org/apache/flink/api/java/io/TextInputFormatTest.java
```diff
@@ -44,14 +48,34 @@
  * Tests for {@link TextInputFormat}.
  */
 public class TextInputFormatTest {
+
+	private File baseDir;
+	private final String tempDirName = "tempDir";
+
+	@Before
+	public void setUp() throws IOException {
```
zentol: use a `TemporaryFolder` instead

Aitozi: Seems better, I will change it then. But a little question, why my approach can not ensure the file cleaned (I check and clean the related file before execute and after test method), Thanks.

zentol: if the JVM crashes before the delete calls are executed the files will remain.

Aitozi:
I update the PR, please take a look when you are free, thanks @zentol

PR #8306, opened 2019-04-28
[FLINK-12342] Add fast-heartbeat-delay yarn config for  jobs with large number of containers

HuangZhenQiu:

Fix the issue of requesting too many containers issue in FlinkResourceManager. Original the new request issue is bounded by the number of pending containers in AMRMClientAsync. As every time AMRMClientAsync will issue request of pendingContainers  + 1 containers, it always ask more than we need. Especially when an Flink application needs 100+ containers, the number total containers will be in thousands. The issue will be enlarged when we do cluster maintenance and restart 1000+ applications as soon as possible. Too many request will overload yarn resource manager and introduce extra delay for restarting applications.

## Brief change log

  - record the total number of containers are requested and still in pending status
  - use the totalRequestedPendingContainers as upper bound fore new container acquirement

## Verifying this change

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-yarn/src/main/java/org/apache/flink/yarn/YarnResourceManager.java
+++ b/flink-yarn/src/main/java/org/apache/flink/yarn/YarnResourceManager.java
@@ -86,10 +86,6 @@ public class YarnResourceManager extends ResourceManager<YarnWorkerNode> impleme
 
 	/** YARN container map. Package private for unit test purposes. */
 	private final ConcurrentMap<ResourceID, YarnWorkerNode> workerNodeMap;
-
-	/** The heartbeat interval while the resource master is waiting for containers. */
-	private static final int FAST_YARN_HEARTBEAT_INTERVAL_MS = 500;
-
 	/** Environment variable name of the final container id used by the YarnResourceManager.
 	 * Container ID generation may vary across Hadoop versions. */
 	static final String ENV_FLINK_CONTAINER_ID = "_FLINK_CONTAINER_ID";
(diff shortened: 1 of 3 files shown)


HuangZhenQiu:
@tillrohrmann @rmetzger 
As using the config and set it to 3000 milliseconds, the job with 256 containers can be successfully launched with only 1000+ total requested containers. The number can be further reduced by using larger number, such as 5000 or even higher. So, for small jobs with 32 containers, user should just default value for sending out request as soon as possible. For large jobs, user need to tune the parameter to trade-off the fast request and negative impact of repetitively as more containers.

flink-yarn/src/main/java/org/apache/flink/yarn/YarnResourceManager.java
```diff
@@ -514,8 +514,7 @@ private void requestYarnContainer() {
 		resourceManagerClient.addContainerRequest(getContainerRequest());
 
 		// make sure we transmit the request fast and receive fast news of granted allocations
-		resourceManagerClient.setHeartbeatInterval(FAST_YARN_HEARTBEAT_INTERVAL_MS);
-
+		resourceManagerClient.setHeartbeatInterval(containerRequestHeartbeatIntervalMillis);
```
lamber-ken: Hi, @tillrohrmann  @HuangZhenQiu, `resourceManagerClient.setHeartbeatInterval(containerRequestHeartbeatIntervalMillis);` need set the value when `requestYarnContainer` called each time?  thanks.

tillrohrmann: It is not optimal but it should also not hurt too much. Moreover, at the moment this method is the common place which is called for every container request. Changing it should be out of scope for this PR.

lamber-ken: > It is not optimal but it should also not hurt too much. Moreover, at the moment this method is the common place which is called for every container request. Changing it should be out of scope for this PR.

ok, I see.

PR #8307, opened 2019-04-28
[FLINK-12359][metrics][tests] Harden SystemResourcesMetricsITCase

zentol:

Hardens the `SystemResourcesMetricsITCase` against
* slow starts of a TaskManager, in which case metrics could be registered later than the test expects
* the host portion of the metric identifier not being `localhost`

We now exclude the host from the identifier, and generate a future in the reporter that we wait on instead.

### DIFF
--- a/flink-tests/src/test/java/org/apache/flink/runtime/metrics/SystemResourcesMetricsITCase.java
+++ b/flink-tests/src/test/java/org/apache/flink/runtime/metrics/SystemResourcesMetricsITCase.java
@@ -19,9 +19,12 @@
 package org.apache.flink.runtime.metrics;
 
 import org.apache.flink.configuration.Configuration;
-import org.apache.flink.metrics.Gauge;
+import org.apache.flink.configuration.MetricOptions;
+import org.apache.flink.metrics.Metric;
 import org.apache.flink.metrics.MetricConfig;
-import org.apache.flink.metrics.reporter.AbstractReporter;
+import org.apache.flink.metrics.MetricGroup;
+import org.apache.flink.metrics.reporter.MetricReporter;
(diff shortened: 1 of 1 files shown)


rmetzger:
@flinkbot approve description consensus

flink-tests/src/test/java/org/apache/flink/runtime/metrics/SystemResourcesMetricsITCase.java
```diff
@@ -136,8 +126,17 @@ public void close() {
 			OPENED_REPORTERS.remove(this);
 		}
 
-		public Map<Gauge<?>, String> getGauges() {
-			return gauges;
+		@Override
+		public void notifyOfAddedMetric(Metric metric, String metricName, MetricGroup group) {
+			for (final String expectedPattern : patternFutures.keySet()) {
+				if (group.getMetricIdentifier(metricName, name -> name).matches(expectedPattern)) {
```
tillrohrmann: I think we could move `group.getMetricIdentifier()` outside of the loop

flink-tests/src/test/java/org/apache/flink/runtime/metrics/SystemResourcesMetricsITCase.java
```diff
@@ -102,8 +94,8 @@ public void startTaskManagerAndCheckForRegisteredSystemMetrics() throws Exceptio
 		};
 
 		String[] expectedHosts = new String[] {
```
tillrohrmann: Could be replaced with array initializer.

tillrohrmann: The same applies to the `expectedGauges`.

PR #8309, opened 2019-04-29
[FLINK-12229] [runtime] Implement LazyFromSourcesScheduling Strategy

eaglewatcherwb:

*This pull request implements the LAZY_FROM_SOURCES scheduling strategy based on the new scheduling interface.*

## Brief change log

  - *Implement the LazyFromSourcesSchedulingStrategy which schedules vertices when the input data are available*
  - *Add InputDependencyConstraintChecker class to efficiently check input dependency constraint*
  - *Add getInputDependencyConstraint in SchedulingExecutionVertex interface*

## Verifying this change

  - *Add unit test in LazyFromSourcesSchedulingStrategyTest*
  - *Add unit test in InputDependencyConstraintCheckerTest*

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/strategy/InputDependencyConstraintChecker.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/strategy/InputDependencyConstraintChecker.java
@@ -1,0 +1,180 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.runtime.scheduler.strategy;
+
+import org.apache.flink.api.common.InputDependencyConstraint;
+import org.apache.flink.runtime.jobgraph.IntermediateDataSet;
+import org.apache.flink.runtime.jobgraph.IntermediateDataSetID;
+import org.apache.flink.runtime.jobgraph.IntermediateResultPartitionID;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+import static org.apache.flink.api.common.InputDependencyConstraint.ALL;
+import static org.apache.flink.api.common.InputDependencyConstraint.ANY;
+import static org.apache.flink.runtime.io.network.partition.ResultPartitionType.BLOCKING;
+import static org.apache.flink.runtime.scheduler.strategy.SchedulingResultPartition.ResultPartitionState.DONE;
+import static org.apache.flink.runtime.scheduler.strategy.SchedulingResultPartition.ResultPartitionState.PRODUCING;
+
+/**
+ * A wrapper class for {@link InputDependencyConstraint} checker.
+ */
+public class InputDependencyConstraintChecker {
+	private final SchedulingIntermediateDataSetManager intermediateDataSetManager =
+		new SchedulingIntermediateDataSetManager();
+
+	public boolean check(final SchedulingExecutionVertex schedulingExecutionVertex) {
+		final InputDependencyConstraint inputConstraint = schedulingExecutionVertex.getInputDependencyConstraint();
+		if (schedulingExecutionVertex.getConsumedResultPartitions().isEmpty() || ALL.equals(inputConstraint)) {
+			return checkAll(schedulingExecutionVertex);
+		} else if (ANY.equals(inputConstraint)) {
+			return checkAny(schedulingExecutionVertex);
+		} else {
+			throw new IllegalArgumentException();
+		}
+	}
+
+	List<SchedulingResultPartition> markSchedulingResultPartitionFinished(SchedulingResultPartition srp) {
+		return intermediateDataSetManager.markSchedulingResultPartitionFinished(srp);
+	}
+
+	void resetSchedulingResultPartition(SchedulingResultPartition srp) {
+		intermediateDataSetManager.resetSchedulingResultPartition(srp);
+	}
+
+	void addSchedulingResultPartition(SchedulingResultPartition srp) {
+		intermediateDataSetManager.addSchedulingResultPartition(srp);
+	}
+
+	private boolean checkAll(final SchedulingExecutionVertex schedulingExecutionVertex) {
+		return schedulingExecutionVertex.getConsumedResultPartitions()
+			.stream()
+			.allMatch(this::partitionConsumable);
+	}
+
+	private boolean checkAny(final SchedulingExecutionVertex schedulingExecutionVertex) {
+		return schedulingExecutionVertex.getConsumedResultPartitions()
+			.stream()
+			.anyMatch(this::partitionConsumable);
+	}
+
+	private boolean partitionConsumable(SchedulingResultPartition partition) {
+		if (BLOCKING.equals(partition.getPartitionType())) {
+			return intermediateDataSetManager.allPartitionsFinished(partition);
+		} else {
+			SchedulingResultPartition.ResultPartitionState state = partition.getState();
+			return PRODUCING.equals(state) || DONE.equals(state);
+		}
+	}
+
+	private static class SchedulingIntermediateDataSetManager {
+
+		private final Map<IntermediateDataSetID, SchedulingIntermediateDataSet> intermediateDataSets = new HashMap<>();
+
+		List<SchedulingResultPartition> markSchedulingResultPartitionFinished(SchedulingResultPartition srp) {
+			SchedulingIntermediateDataSet intermediateDataSet = getSchedulingIntermediateDataSet(srp.getResultId());
+			if (intermediateDataSet.markPartitionFinished(srp.getId())) {
+				return intermediateDataSet.getSchedulingResultPartitions();
+			}
+			return Collections.emptyList();
+		}
+
+		void resetSchedulingResultPartition(SchedulingResultPartition srp) {
+			SchedulingIntermediateDataSet sid = getSchedulingIntermediateDataSet(srp.getResultId());
+			sid.resetPartition(srp.getId());
+		}
+
+		void addSchedulingResultPartition(SchedulingResultPartition srp) {
+			SchedulingIntermediateDataSet sid = getOrCreateSchedulingIntermediateDataSetIfAbsent(srp.getResultId());
+			sid.addSchedulingResultPartition(srp);
+		}
+
+		boolean allPartitionsFinished(SchedulingResultPartition srp) {
+			SchedulingIntermediateDataSet sid = getSchedulingIntermediateDataSet(srp.getResultId());
+			return sid.allPartitionsFinished();
+		}
+
+		private SchedulingIntermediateDataSet getSchedulingIntermediateDataSet(
+				final IntermediateDataSetID intermediateDataSetId) {
+			return getSchedulingIntermediateDataSetInternal(intermediateDataSetId, false);
+		}
+
+		private SchedulingIntermediateDataSet getOrCreateSchedulingIntermediateDataSetIfAbsent(
+				final IntermediateDataSetID intermediateDataSetId) {
+			return getSchedulingIntermediateDataSetInternal(intermediateDataSetId, true);
+		}
+
+		private SchedulingIntermediateDataSet getSchedulingIntermediateDataSetInternal(
+				final IntermediateDataSetID intermediateDataSetId,
+				boolean createIfAbsent) {
+
+			return intermediateDataSets.computeIfAbsent(
+				intermediateDataSetId,
+				(key) -> {
+					if (createIfAbsent) {
+						return new SchedulingIntermediateDataSet();
+					} else {
+						throw new IllegalArgumentException("can not find data set for " + intermediateDataSetId);
+					}
+				});
+		}
+	}
+
+	/**
+	 * Representation of {@link IntermediateDataSet}.
+	 */
+	private static class SchedulingIntermediateDataSet {
+
+		private final List<SchedulingResultPartition> partitions;
+
+		private final Set<IntermediateResultPartitionID> producingPartitionIds;
+
+		SchedulingIntermediateDataSet() {
+			partitions = new ArrayList<>();
+			producingPartitionIds = new HashSet<>();
+		}
+
+		boolean markPartitionFinished(IntermediateResultPartitionID partitionId) {
+			producingPartitionIds.remove(partitionId);
+			return producingPartitionIds.isEmpty();
+		}
+
+		void resetPartition(IntermediateResultPartitionID partitionId) {
+			producingPartitionIds.add(partitionId);
+		}
+
+		boolean allPartitionsFinished() {
+			return producingPartitionIds.isEmpty();
+		}
+
+		void addSchedulingResultPartition(SchedulingResultPartition partition) {
+			partitions.add(partition);
+			producingPartitionIds.add(partition.getId());
+		}
+
+		List<SchedulingResultPartition> getSchedulingResultPartitions() {
+			return Collections.unmodifiableList(partitions);
+		}
+	}
+}
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/strategy/LazyFromSourcesSchedulingStrategy.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/strategy/LazyFromSourcesSchedulingStrategy.java
@@ -1,0 +1,180 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.runtime.scheduler.strategy;
+
+import org.apache.flink.runtime.execution.ExecutionState;
+import org.apache.flink.runtime.io.network.partition.ResultPartitionID;
+import org.apache.flink.runtime.jobgraph.JobGraph;
+import org.apache.flink.runtime.scheduler.DeploymentOption;
+import org.apache.flink.runtime.scheduler.ExecutionVertexDeploymentOption;
+import org.apache.flink.runtime.scheduler.SchedulerOperations;
+
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
+
+import static org.apache.flink.runtime.execution.ExecutionState.CREATED;
+import static org.apache.flink.runtime.execution.ExecutionState.FINISHED;
+import static org.apache.flink.util.Preconditions.checkNotNull;
+
+/**
+ * {@link SchedulingStrategy} instance for batch job which schedule vertices when input data are ready.
+ */
+public class LazyFromSourcesSchedulingStrategy implements SchedulingStrategy {
+
+	private static final Predicate<SchedulingExecutionVertex> IS_IN_CREATED_EXECUTION_STATE = schedulingExecutionVertex -> CREATED == schedulingExecutionVertex.getState();
+
+	private final SchedulerOperations schedulerOperations;
+
+	private final SchedulingTopology schedulingTopology;
+
+	private final Map<ExecutionVertexID, DeploymentOption> deploymentOptions;
+
+	private final InputDependencyConstraintChecker inputConstraintChecker;
+
+	public LazyFromSourcesSchedulingStrategy(
+			SchedulerOperations schedulerOperations,
+			SchedulingTopology schedulingTopology) {
+
+		this.schedulerOperations = checkNotNull(schedulerOperations);
+		this.schedulingTopology = checkNotNull(schedulingTopology);
+		this.deploymentOptions = new HashMap<>();
+		this.inputConstraintChecker = new InputDependencyConstraintChecker();
+	}
+
+	@Override
+	public void startScheduling() {
+		final DeploymentOption updateOption = new DeploymentOption(true);
+		final DeploymentOption nonUpdateOption = new DeploymentOption(false);
+
+		for (SchedulingExecutionVertex schedulingVertex : schedulingTopology.getVertices()) {
+			DeploymentOption option = nonUpdateOption;
+			for (SchedulingResultPartition srp : schedulingVertex.getProducedResultPartitions()) {
+				if (srp.getPartitionType().isPipelined()) {
+					option = updateOption;
+				}
+				inputConstraintChecker.addSchedulingResultPartition(srp);
+			}
+			deploymentOptions.put(schedulingVertex.getId(), option);
+		}
+
+		allocateSlotsAndDeployExecutionVertexIds(getAllVerticesFromTopology());
+	}
+
+	@Override
+	public void restartTasks(Set<ExecutionVertexID> verticesToRestart) {
+		// increase counter of the dataset first
+		verticesToRestart
+			.stream()
+			.map(this::getSchedulingVertex)
+			.flatMap(vertex -> vertex.getProducedResultPartitions().stream())
+			.forEach(inputConstraintChecker::resetSchedulingResultPartition);
+
+		allocateSlotsAndDeployExecutionVertexIds(verticesToRestart);
+	}
+
+	@Override
+	public void onExecutionStateChange(ExecutionVertexID executionVertexId, ExecutionState executionState) {
+		if (!FINISHED.equals(executionState)) {
+			return;
+		}
+
+		final Set<SchedulingExecutionVertex> verticesToSchedule = getSchedulingVertex(executionVertexId)
+			.getProducedResultPartitions()
+			.stream()
+			.flatMap(partition -> inputConstraintChecker.markSchedulingResultPartitionFinished(partition).stream())
+			.flatMap(partition -> partition.getConsumers().stream())
+			.collect(Collectors.toSet());
+
+		allocateSlotsAndDeployExecutionVertices(verticesToSchedule);
+	}
+
+	@Override
+	public void onPartitionConsumable(ExecutionVertexID executionVertexId, ResultPartitionID resultPartitionId) {
+		final SchedulingResultPartition resultPartition = schedulingTopology
+				.getResultPartition(resultPartitionId.getPartitionId())
+				.orElseThrow(() -> new IllegalStateException("can not find scheduling result partition for "
+						+ resultPartitionId));
+
+		if (!resultPartition.getPartitionType().isPipelined()) {
+			return;
+		}
+
+		final SchedulingExecutionVertex producerVertex = getSchedulingVertex(executionVertexId);
+		if (!producerVertex.getProducedResultPartitions().contains(resultPartition)) {
+			throw new IllegalStateException("partition " + resultPartitionId
+					+ " is not the produced partition of " + executionVertexId);
+		}
+
+		allocateSlotsAndDeployExecutionVertices(resultPartition.getConsumers());
+	}
+
+	private SchedulingExecutionVertex getSchedulingVertex(final ExecutionVertexID executionVertexId) {
+		return schedulingTopology.getVertex(executionVertexId)
+			.orElseThrow(() -> new IllegalStateException("can not find scheduling vertex for " + executionVertexId));
+	}
+
+	private void allocateSlotsAndDeployExecutionVertexIds(Set<ExecutionVertexID> verticesToSchedule) {
+		allocateSlotsAndDeployExecutionVertices(
+			verticesToSchedule
+				.stream()
+				.map(this::getSchedulingVertex)
+				.collect(Collectors.toList()));
+	}
+
+	private void allocateSlotsAndDeployExecutionVertices(final Collection<SchedulingExecutionVertex> schedulingExecutionVertices) {
+		schedulerOperations.allocateSlotsAndDeploy(
+			schedulingExecutionVertices
+				.stream()
+				.filter(isInputConstraintSatisfied().and(IS_IN_CREATED_EXECUTION_STATE))
+				.map(SchedulingExecutionVertex::getId)
+				.map(executionVertexID -> new ExecutionVertexDeploymentOption(
+					executionVertexID,
+					deploymentOptions.get(executionVertexID)))
+				.collect(Collectors.toSet()));
+	}
+
+	private Predicate<SchedulingExecutionVertex> isInputConstraintSatisfied() {
+		return inputConstraintChecker::check;
+	}
+
+	private Set<ExecutionVertexID> getAllVerticesFromTopology() {
+		return StreamSupport
+			.stream(schedulingTopology.getVertices().spliterator(), false)
+			.map(SchedulingExecutionVertex::getId)
+			.collect(Collectors.toSet());
+	}
+
+	/**
+	 * The factory for creating {@link LazyFromSourcesSchedulingStrategy}.
+	 */
+	public static class Factory implements SchedulingStrategyFactory {
+		@Override
+		public SchedulingStrategy createInstance(
+				SchedulerOperations schedulerOperations,
+				SchedulingTopology schedulingTopology,
+				JobGraph jobGraph) {
+			return new LazyFromSourcesSchedulingStrategy(schedulerOperations, schedulingTopology);
+		}
+	}
+}
--- a/flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/strategy/InputDependencyConstraintCheckerTest.java
+++ b/flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/strategy/InputDependencyConstraintCheckerTest.java
@@ -1,0 +1,289 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.runtime.scheduler.strategy;
+
+import org.apache.flink.api.common.InputDependencyConstraint;
+import org.apache.flink.runtime.io.network.partition.ResultPartitionType;
+import org.apache.flink.runtime.jobgraph.IntermediateDataSetID;
+import org.apache.flink.runtime.jobgraph.JobVertexID;
+import org.apache.flink.util.TestLogger;
+
+import org.junit.Test;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+import static org.apache.flink.api.common.InputDependencyConstraint.ALL;
+import static org.apache.flink.api.common.InputDependencyConstraint.ANY;
+import static org.apache.flink.runtime.io.network.partition.ResultPartitionType.BLOCKING;
+import static org.apache.flink.runtime.io.network.partition.ResultPartitionType.PIPELINED;
+import static org.apache.flink.runtime.scheduler.strategy.SchedulingResultPartition.ResultPartitionState.DONE;
+import static org.apache.flink.runtime.scheduler.strategy.SchedulingResultPartition.ResultPartitionState.EMPTY;
+import static org.apache.flink.runtime.scheduler.strategy.SchedulingResultPartition.ResultPartitionState.PRODUCING;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+/**
+ * Unit tests for {@link InputDependencyConstraintChecker}.
+ */
+public class InputDependencyConstraintCheckerTest extends TestLogger {
+
+	@Test
+	public void testCheckInputVertex() {
+		final TestingSchedulingExecutionVertex vertex = addSchedulingExecutionVertex().finish();
+		final InputDependencyConstraintChecker inputChecker = createInputDependencyConstraintChecker(Collections.emptyList());
+
+		assertTrue(inputChecker.check(vertex));
+	}
+
+	@Test
+	public void testCheckEmptyPipelinedInput() {
+		final List<TestingSchedulingResultPartition> partitions = addResultPartition()
+			.withPartitionType(PIPELINED)
+			.withPartitionState(EMPTY)
+			.finish();
+		final TestingSchedulingExecutionVertex vertex = addSchedulingExecutionVertex()
+			.withConsumedPartitions(partitions)
+			.finish();
+
+		final InputDependencyConstraintChecker inputChecker = createInputDependencyConstraintChecker(partitions);
+
+		assertFalse(inputChecker.check(vertex));
+	}
+
+	@Test
+	public void testCheckProducingPipelinedInput() {
+		final List<TestingSchedulingResultPartition> partitions = addResultPartition()
+			.withPartitionType(PIPELINED)
+			.withPartitionState(PRODUCING)
+			.finish();
+		final TestingSchedulingExecutionVertex vertex = addSchedulingExecutionVertex()
+			.withConsumedPartitions(partitions)
+			.finish();
+
+		final InputDependencyConstraintChecker inputChecker = createInputDependencyConstraintChecker(partitions);
+
+		assertTrue(inputChecker.check(vertex));
+	}
+
+	@Test
+	public void testCheckDoneBlockingInput() {
+		final List<TestingSchedulingResultPartition> partitions = addResultPartition()
+			.withPartitionCntPerDataSet(2)
+			.finish();
+		final TestingSchedulingExecutionVertex vertex = addSchedulingExecutionVertex()
+			.withConsumedPartitions(partitions)
+			.finish();
+
+		final InputDependencyConstraintChecker inputChecker = createInputDependencyConstraintChecker(partitions);
+
+		for (TestingSchedulingResultPartition srp : partitions) {
+			inputChecker.markSchedulingResultPartitionFinished(srp);
+		}
+
+		assertTrue(inputChecker.check(vertex));
+	}
+
+	@Test
+	public void testCheckPartialDoneBlockingInput() {
+		final List<TestingSchedulingResultPartition> partitions = addResultPartition()
+			.withPartitionCntPerDataSet(2)
+			.finish();
+		final TestingSchedulingExecutionVertex vertex = addSchedulingExecutionVertex()
+			.withConsumedPartitions(partitions)
+			.finish();
+
+		final InputDependencyConstraintChecker inputChecker = createInputDependencyConstraintChecker(partitions);
+
+		inputChecker.markSchedulingResultPartitionFinished(partitions.get(0));
+
+		assertFalse(inputChecker.check(vertex));
+	}
+
+	@Test
+	public void testCheckResetBlockingInput() {
+		final List<TestingSchedulingResultPartition> partitions = addResultPartition()
+			.withPartitionCntPerDataSet(2)
+			.finish();
+		final TestingSchedulingExecutionVertex vertex = addSchedulingExecutionVertex()
+			.withConsumedPartitions(partitions)
+			.finish();
+
+		final InputDependencyConstraintChecker inputChecker = createInputDependencyConstraintChecker(partitions);
+
+		for (TestingSchedulingResultPartition srp : partitions) {
+			inputChecker.markSchedulingResultPartitionFinished(srp);
+		}
+
+		for (TestingSchedulingResultPartition srp : partitions) {
+			inputChecker.resetSchedulingResultPartition(srp);
+		}
+
+		assertFalse(inputChecker.check(vertex));
+	}
+
+	@Test
+	public void testCheckAnyBlockingInput() {
+		final List<TestingSchedulingResultPartition> partitions = addResultPartition()
+			.withDataSetCnt(2)
+			.finish();
+		final TestingSchedulingExecutionVertex vertex = addSchedulingExecutionVertex()
+			.withConsumedPartitions(partitions)
+			.finish();
+
+		final InputDependencyConstraintChecker inputChecker = createInputDependencyConstraintChecker(partitions);
+
+		inputChecker.markSchedulingResultPartitionFinished(partitions.get(0));
+
+		assertTrue(inputChecker.check(vertex));
+	}
+
+	@Test
+	public void testCheckAllBlockingInput() {
+		final List<TestingSchedulingResultPartition> partitions = addResultPartition()
+			.withDataSetCnt(2)
+			.finish();
+		final TestingSchedulingExecutionVertex vertex = addSchedulingExecutionVertex()
+			.withInputDependencyConstraint(ALL)
+			.withConsumedPartitions(partitions)
+			.finish();
+
+		final InputDependencyConstraintChecker inputChecker = createInputDependencyConstraintChecker(partitions);
+
+		for (TestingSchedulingResultPartition srp : partitions) {
+			inputChecker.markSchedulingResultPartitionFinished(srp);
+		}
+
+		assertTrue(inputChecker.check(vertex));
+	}
+
+	@Test
+	public void testCheckAllPartialDatasetBlockingInput() {
+		final List<TestingSchedulingResultPartition> partitions = addResultPartition()
+			.withDataSetCnt(2)
+			.finish();
+		final TestingSchedulingExecutionVertex vertex = addSchedulingExecutionVertex()
+			.withInputDependencyConstraint(ALL)
+			.withConsumedPartitions(partitions)
+			.finish();
+
+		final InputDependencyConstraintChecker inputChecker = createInputDependencyConstraintChecker(partitions);
+
+		inputChecker.markSchedulingResultPartitionFinished(partitions.get(0));
+		assertFalse(inputChecker.check(vertex));
+	}
+
+	@Test
+	public void testCheckAllPartialPartitionBlockingInput() {
+		final List<TestingSchedulingResultPartition> partitions = addResultPartition()
+			.withDataSetCnt(2)
+			.withPartitionCntPerDataSet(2)
+			.finish();
+		final TestingSchedulingExecutionVertex vertex = addSchedulingExecutionVertex()
+			.withInputDependencyConstraint(ALL)
+			.withConsumedPartitions(partitions)
+			.finish();
+
+		final InputDependencyConstraintChecker inputChecker = createInputDependencyConstraintChecker(partitions);
+
+		for (int idx = 0; idx < 3; idx++) {
+			inputChecker.markSchedulingResultPartitionFinished(partitions.get(idx));
+		}
+
+		assertFalse(inputChecker.check(vertex));
+	}
+
+	private static TestingSchedulingExecutionVertexBuilder addSchedulingExecutionVertex() {
+		return new TestingSchedulingExecutionVertexBuilder();
+	}
+
+	private static class TestingSchedulingExecutionVertexBuilder {
+		private static final JobVertexID jobVertexId = new JobVertexID();
+		private InputDependencyConstraint inputDependencyConstraint = ANY;
+		private List<TestingSchedulingResultPartition> partitions = Collections.emptyList();
+
+		TestingSchedulingExecutionVertexBuilder withInputDependencyConstraint(InputDependencyConstraint constraint) {
+			this.inputDependencyConstraint = constraint;
+			return this;
+		}
+
+		TestingSchedulingExecutionVertexBuilder withConsumedPartitions(List<TestingSchedulingResultPartition> partitions) {
+			this.partitions = partitions;
+			return this;
+		}
+
+		TestingSchedulingExecutionVertex finish() {
+			return new TestingSchedulingExecutionVertex(jobVertexId, 0, inputDependencyConstraint, partitions);
+		}
+	}
+
+	private static TestingSchedulingResultPartitionBuilder addResultPartition() {
+		return new TestingSchedulingResultPartitionBuilder();
+	}
+
+	private static InputDependencyConstraintChecker createInputDependencyConstraintChecker(
+		List<TestingSchedulingResultPartition> partitions) {
+
+		InputDependencyConstraintChecker inputChecker = new InputDependencyConstraintChecker();
+		for (SchedulingResultPartition partition : partitions) {
+			inputChecker.addSchedulingResultPartition(partition);
+		}
+		return inputChecker;
+	}
+
+	private static class TestingSchedulingResultPartitionBuilder {
+		private int dataSetCnt = 1;
+		private int partitionCntPerDataSet = 1;
+		private ResultPartitionType partitionType = BLOCKING;
+		private SchedulingResultPartition.ResultPartitionState partitionState = DONE;
+
+		TestingSchedulingResultPartitionBuilder withDataSetCnt(int dataSetCnt) {
+			this.dataSetCnt = dataSetCnt;
+			return this;
+		}
+
+		TestingSchedulingResultPartitionBuilder withPartitionCntPerDataSet(int partitionCnt) {
+			this.partitionCntPerDataSet = partitionCnt;
+			return this;
+		}
+
+		TestingSchedulingResultPartitionBuilder withPartitionType(ResultPartitionType type) {
+			this.partitionType = type;
+			return this;
+		}
+
+		TestingSchedulingResultPartitionBuilder withPartitionState(SchedulingResultPartition.ResultPartitionState state) {
+			this.partitionState = state;
+			return this;
+		}
+
+		List<TestingSchedulingResultPartition> finish() {
+			List<TestingSchedulingResultPartition> partitions = new ArrayList<>(dataSetCnt * partitionCntPerDataSet);
+			for (int dataSetIdx = 0; dataSetIdx < dataSetCnt; dataSetIdx++) {
+				IntermediateDataSetID dataSetId = new IntermediateDataSetID();
+				for (int partitionIdx = 0; partitionIdx < partitionCntPerDataSet; partitionIdx++) {
+					partitions.add(new TestingSchedulingResultPartition(dataSetId, partitionType, partitionState));
+				}
+			}
+
+			return partitions;
+		}
+	}
+}
--- a/flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/strategy/LazyFromSourcesSchedulingStrategyTest.java
+++ b/flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/strategy/LazyFromSourcesSchedulingStrategyTest.java
@@ -1,0 +1,382 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.runtime.scheduler.strategy;
+
+import org.apache.flink.runtime.execution.ExecutionState;
+import org.apache.flink.runtime.executiongraph.ExecutionAttemptID;
+import org.apache.flink.runtime.io.network.partition.ResultPartitionID;
+import org.apache.flink.runtime.io.network.partition.ResultPartitionType;
+import org.apache.flink.util.TestLogger;
+
+import org.hamcrest.Description;
+import org.hamcrest.Matcher;
+import org.hamcrest.TypeSafeDiagnosingMatcher;
+import org.junit.Test;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import static org.apache.flink.api.common.InputDependencyConstraint.ALL;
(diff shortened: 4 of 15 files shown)


flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/strategy/LazyFromSourcesSchedulingStrategy.java
```diff
@@ -0,0 +1,183 @@
...
+	public void onPartitionConsumable(ExecutionVertexID executionVertexId, ResultPartitionID resultPartitionId) {
+		// No need to react to this notification.
+	}
+
+	private void scheduleBlockingConsumers(List<SchedulingVertex> vertices) {
+		List<ExecutionVertexDeploymentOption> executionVertexDeploymentOptions = new LinkedList<>();
+		for (SchedulingVertex schedulingVertex : vertices) {
+			final ExecutionState state = schedulingVertex.getState();
+
+			if (CREATED.equals(state)) {
```
Ryantaocer: use ExecutionState.CREATED ?

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/strategy/LazyFromSourcesSchedulingStrategy.java
```diff
@@ -0,0 +1,183 @@
...
+				final InputDependencyConstraint inputDependency = jobGraph.findVertexByID(schedulingVertex
+					.getJobVertexId()).getInputDependencyConstraint();
+
+				boolean shouldSchedule = true;
+				if (!InputDependencyConstraint.ANY.equals(inputDependency)) {
+					for (SchedulingResultPartition srp : schedulingVertex.getConsumedResultPartitions()) {
```
Ryantaocer: This check loop can optimized during collecting the SchedulingVertex to schedule in the loop above.

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/strategy/SchedulingIntermediateDataSet.java
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+package org.apache.flink.runtime.scheduler.strategy;
+
+import org.apache.flink.runtime.executiongraph.IntermediateResult;
+
+/**
+ * Representation of {@link IntermediateResult}.
+ */
+public interface SchedulingIntermediateDataSet {
```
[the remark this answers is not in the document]
eaglewatcherwb: In origin scheduler, `LazyFromSourceScheduler` schedules vertices when ANY/ALL input `dataset` are consumable, and when all the `partitions` are finished the `dataset` is consumable. Without `SchedulingIntermediateDataSet`,  the input `SchedulingResultPartitions` may be checked O(n^2) times, while, using a counter in `SchedulingIntermediateDataSet` only check O(n) times.

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/strategy/LazyFromSourcesSchedulingStrategy.java
```diff
@@ -0,0 +1,187 @@
...
+		List<ExecutionVertexDeploymentOption> executionVertexDeploymentOptions = new ArrayList<>(verticesNeedingRestart.size());
+		for (ExecutionVertexID executionVertexID : verticesNeedingRestart) {
+			executionVertexDeploymentOptions.add(
+				new ExecutionVertexDeploymentOption(executionVertexID, deploymentOption));
+		}
+		schedulerOperations.allocateSlotsAndDeploy(executionVertexDeploymentOptions);
+	}
+
+	@Override
+	public void onExecutionStateChange(ExecutionVertexID executionVertexId, ExecutionState executionState) {
```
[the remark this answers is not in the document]
eaglewatcherwb: Reacting to onPartitionConsumable is much more complicated to check whether ANY/ALL the input `dataset` are consumable which depends on all the `partitions` are finished of `BLOCKING` type and data are producing of `PIPELINED` type. It seems that reacting to onExecutionStateChange is much easier to handle both `PIPELINED` and `BLOCKING` result type. In the discussion of slack channel, relying on onExecutionStateChange should be an easier and final way.

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/strategy/LazyFromSourcesSchedulingStrategy.java
eaglewatcherwb: done

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/strategy/LazyFromSourcesSchedulingStrategy.java
eaglewatcherwb: These two loops can not be merged since the first loop check the produced partitions while the second check the consumed partitions.

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/strategy/LazyFromSourcesSchedulingStrategyTest.java
```diff
@@ -0,0 +1,194 @@
...
+	 *             vertex#4
+	 *                |
+	 *                |
+	 *                |
+	 *            (PIPELINED)
+	 *             vertex#5
+	 */
+	@Test
+	public void testOnExecutionStateChange() {
+		TestingSchedulingTopology schedulingTopology = new TestingSchedulingTopology();
```
GJL: I think this method is a bit too long. It could be split into more units (see Clean Code by Robert C Martin, Chapter 3).

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/strategy/LazyFromSourcesSchedulingStrategyTest.java
```diff
@@ -0,0 +1,194 @@
...
+				vertices[i][j] = new TestingSchedulingVertex(jobVertices[i].getID(), j);
+				schedulingTopology.addSchedulingVertex(vertices[i][j]);
+
+				if (i != 5) {
+					ResultPartitionType type = i == 4 ? ResultPartitionType.PIPELINED : ResultPartitionType.BLOCKING;
+					partitions[i][j] = new TestingSchedulingResultPartition(dataSets[i].getId(), new IntermediateResultPartitionID(),
+						type, vertices[i][j], null);
+					List<SchedulingResultPartition> producedPartitions = new ArrayList<>(1);
+					producedPartitions.add(partitions[i][j]);
+					vertices[i][j].setProducedPartitions(producedPartitions);
```
GJL: For the future, if you need a list holding a single element, you can use `Collections.singletonList(partitions[i][j])`

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/strategy/LazyFromSourcesSchedulingStrategyTest.java
```diff
@@ -0,0 +1,194 @@
...
+		TestingSchedulingVertex[][] vertices = new TestingSchedulingVertex[6][2];
+		TestingSchedulingResultPartition[][] partitions = new TestingSchedulingResultPartition[6][2];
+		for (int i = 5; i >= 0; i--) {
+			for (int j = 0; j < 2; j++) {
+				vertices[i][j] = new TestingSchedulingVertex(jobVertices[i].getID(), j);
+				schedulingTopology.addSchedulingVertex(vertices[i][j]);
+
+				if (i != 5) {
+					ResultPartitionType type = i == 4 ? ResultPartitionType.PIPELINED : ResultPartitionType.BLOCKING;
+					partitions[i][j] = new TestingSchedulingResultPartition(dataSets[i].getId(), new IntermediateResultPartitionID(),
```
GJL: I think it is not ideal that it is currently easily possible to create contradicting topologies. For example, it is possible to create an instance of `TestingSchedulingResultPartition` that links to a producer vertex with empty _"producedResultPartitions"_. This is because we currently have to create the links between partitions and vertices in two separate steps.

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/strategy/LazyFromSourcesSchedulingStrategy.java
```diff
@@ -0,0 +1,227 @@
...
+			SchedulingTopology schedulingTopology,
+			JobGraph jobGraph) {
+			return new LazyFromSourcesSchedulingStrategy(schedulerOperations, schedulingTopology, jobGraph);
+		}
+	}
+
+	/**
+	 * Representation of {@link IntermediateDataSet}.
+	 */
+	class SchedulingIntermediateDataSet {
```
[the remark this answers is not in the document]
eaglewatcherwb: OK

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/strategy/LazyFromSourcesSchedulingStrategy.java
```diff
@@ -0,0 +1,227 @@
...
+				}
+			}
+		}
+		if (!executionVertexDeploymentOptions.isEmpty()) {
+			schedulerOperations.allocateSlotsAndDeploy(executionVertexDeploymentOptions);
+		}
+	}
+
+	@Override
+	public void restartTasks(Set<ExecutionVertexID> verticesNeedingRestart) {
```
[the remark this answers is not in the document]
eaglewatcherwb: OK, I thought the `verticesNeedingRestart` of `restartTasks` are well chosen for directly scheduling vertices before.
There would be another little difference between `startScheduling` and `restartTasks` in implementation, the former schedules the vertices without input result partition of all vertices in the topology, and the latter schedules vertices with empty input result partition or ANY/ALL input result  partition is ready in `verticesNeedingRestart`.

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/strategy/LazyFromSourcesSchedulingStrategy.java
```diff
@@ -0,0 +1,227 @@
...
+		this.schedulingTopology = checkNotNull(schedulingTopology);
+		this.jobGraph = checkNotNull(jobGraph);
+		this.intermediateDataSets = new HashMap<>();
+	}
+
+	@Override
+	public void startScheduling() {
+		List<ExecutionVertexDeploymentOption> executionVertexDeploymentOptions = new LinkedList<>();
+		for (SchedulingVertex schedulingVertex : schedulingTopology.getVertices()) {
+			if (jobGraph.findVertexByID(schedulingVertex.getJobVertexId()).isInputVertex()) {
```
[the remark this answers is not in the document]
eaglewatcherwb: Yes, `schedulingVertex.getConsumedResultPartitions().isEmpty()` would work.

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/strategy/LazyFromSourcesSchedulingStrategy.java
```diff
@@ -0,0 +1,227 @@
...
+				new ExecutionVertexDeploymentOption(executionVertexID, deploymentOption));
+		}
+		schedulerOperations.allocateSlotsAndDeploy(executionVertexDeploymentOptions);
+	}
+
+	@Override
+	public void onExecutionStateChange(ExecutionVertexID executionVertexId, ExecutionState executionState) {
+		Optional<SchedulingVertex> schedulingVertex = schedulingTopology.getVertex(executionVertexId);
+		if (!schedulingVertex.isPresent()) {
+			throw new IllegalStateException("can not find scheduling vertex for " + executionVertexId);
```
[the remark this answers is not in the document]
eaglewatcherwb: OK

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/strategy/TestingSchedulingResultPartition.java
```diff
@@ -0,0 +1,84 @@
...
+		return producer;
+	}
+
+	@Override
+	public Collection<SchedulingVertex> getConsumers() {
+		return consumers;
+	}
+
+	public void setConsumers(Collection<SchedulingVertex> consumers) {
+		this.consumers = consumers;
```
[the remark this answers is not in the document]
eaglewatcherwb: OK

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/strategy/TestingSchedulingResultPartition.java
```diff
@@ -0,0 +1,84 @@
...
+public class TestingSchedulingResultPartition implements SchedulingResultPartition {
+	private final IntermediateDataSetID intermediateDataSetID;
+
+	private final IntermediateResultPartitionID intermediateResultPartitionID;
+
+	private final ResultPartitionType partitionType;
+
+	private final SchedulingVertex producer;
+
+	private Collection<SchedulingVertex> consumers;
```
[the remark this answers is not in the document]
eaglewatcherwb: OK

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/strategy/LazyFromSourcesSchedulingStrategy.java
```diff
@@ -0,0 +1,227 @@
...
+			List<SchedulingResultPartition> finishedPartitions = null;
+			for (SchedulingResultPartition srp : schedulingVertex.get().getProducedResultPartitions()) {
+				if (srp.getPartitionType().isBlocking()) {
+					SchedulingIntermediateDataSet intermediateDataSet = intermediateDataSets.get(srp.getResultId());
+					if (intermediateDataSet == null) {
+						throw new IllegalStateException("can not find scheduling intermediate dataset for " + srp.getResultId());
+					}
+					final int refCnt = intermediateDataSet.decrementNumberOfRunningProducersAndGetRemaining();
+					if (refCnt < 0) {
+						throw new IllegalStateException("Decremented number of unfinished producers below 0. "
```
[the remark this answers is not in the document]
eaglewatcherwb: OK

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/strategy/LazyFromSourcesSchedulingStrategy.java
```diff
@@ -0,0 +1,227 @@
...
+			Collection<SchedulingResultPartition> partitions = schedulingVertex.getProducedResultPartitions();
+			if (partitions != null) {
+				for (SchedulingResultPartition srp : partitions) {
+					SchedulingIntermediateDataSet sid = intermediateDataSets.computeIfAbsent(srp.getResultId(),
+						(key) -> new SchedulingIntermediateDataSet());
+					sid.increaseProducerCnt();
+				}
+			}
+		}
+		if (!executionVertexDeploymentOptions.isEmpty()) {
```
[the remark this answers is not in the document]
eaglewatcherwb: OK

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/strategy/LazyFromSourcesSchedulingStrategy.java
```diff
@@ -0,0 +1,227 @@
...
+		List<ExecutionVertexDeploymentOption> executionVertexDeploymentOptions = new LinkedList<>();
+		for (SchedulingVertex schedulingVertex : schedulingTopology.getVertices()) {
+			if (jobGraph.findVertexByID(schedulingVertex.getJobVertexId()).isInputVertex()) {
+				// schedule vertices without consumed result partition
+				executionVertexDeploymentOptions.add(
+					new ExecutionVertexDeploymentOption(schedulingVertex.getId(), deploymentOption));
+			}
+
+			Collection<SchedulingResultPartition> partitions = schedulingVertex.getProducedResultPartitions();
+			if (partitions != null) {
```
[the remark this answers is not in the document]
eaglewatcherwb: OK

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/strategy/LazyFromSourcesSchedulingStrategy.java
```diff
@@ -0,0 +1,227 @@
...
+				}
+
+				if (shouldSchedule) {
+					executionVertexDeploymentOptions.add(
+						new ExecutionVertexDeploymentOption(schedulingVertex.getId(), deploymentOption));
+				}
+			}
+		}
+
+		if (!executionVertexDeploymentOptions.isEmpty()) {
```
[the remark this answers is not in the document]
eaglewatcherwb: OK

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/strategy/LazyFromSourcesSchedulingStrategy.java
```diff
@@ -0,0 +1,227 @@
...
+						throw new IllegalStateException("can not find scheduling intermediate dataset for " + srp.getResultId());
+					}
+					final int refCnt = intermediateDataSet.decrementNumberOfRunningProducersAndGetRemaining();
+					if (refCnt < 0) {
+						throw new IllegalStateException("Decremented number of unfinished producers below 0. "
+							+ "This is most likely a bug in the execution state/intermediate result "
+							+ "partition management.");
+					} else if (refCnt == 0) {
+						if (finishedPartitions == null) {
+							finishedPartitions = new LinkedList<>();
```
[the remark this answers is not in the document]
eaglewatcherwb: OK

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/strategy/LazyFromSourcesSchedulingStrategy.java
```diff
@@ -0,0 +1,227 @@
...
+						}
+						finishedPartitions.add(srp);
```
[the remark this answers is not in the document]
eaglewatcherwb: Yes, you are right. Should find the finished `IntermediateResult` first and then find all the result partitions of the `IntermediateResult`.

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/strategy/LazyFromSourcesSchedulingStrategy.java
```diff
@@ -0,0 +1,227 @@
...
+		// No need to react to this notification.
+	}
+
+	private void scheduleBlockingConsumers(List<SchedulingVertex> vertices) {
+		List<ExecutionVertexDeploymentOption> executionVertexDeploymentOptions = new LinkedList<>();
+		for (SchedulingVertex schedulingVertex : vertices) {
+			final ExecutionState state = schedulingVertex.getState();
+
+			if (ExecutionState.CREATED.equals(state)) {
+				final InputDependencyConstraint inputDependency = jobGraph.findVertexByID(schedulingVertex
```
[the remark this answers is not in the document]
eaglewatcherwb: Yes, job graph would not be needed in scheduling strategy if adding input dependency constraints to `SchedulingTopology`.

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/strategy/LazyFromSourcesSchedulingStrategyTest.java
eaglewatcherwb: OK

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/strategy/LazyFromSourcesSchedulingStrategyTest.java
eaglewatcherwb: OK

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/strategy/LazyFromSourcesSchedulingStrategyTest.java
eaglewatcherwb: OK, I will refine the structure.

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/strategy/LazyFromSourcesSchedulingStrategy.java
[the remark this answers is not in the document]
eaglewatcherwb: Since using vertex state transitions to schedule vertices has the benefit of avoiding flood `onPartitionConsumable` notifications, while there may be operators that take a very long time until they produce a result of PIPELINED shuffle mode. So, I think we could keep both the benefit by relying on both vertex state transitions and  `onPartitionConsumable` notifications.
1. DeploymentOption#sendScheduleOrUpdateConsumerMessage set to true only the vertex has PIPELINED produced result partition.
2. Schedule vertices with BLOCKING input result partition using vertex state transition.
3. Schedule vertices with PIPELINED input result partitions using `onPartitionConsumable` notification.
What do you think?

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/strategy/LazyFromSourcesSchedulingStrategy.java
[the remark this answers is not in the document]
GJL: > I thought the verticesNeedingRestart of restartTasks are well chosen for directly scheduling vertices before.

The vertices are well chosen but in some cases all vertices need to be restarted.

> the latter schedules vertices with [...] ANY/ALL input result partition is ready in verticesNeedingRestart

That's correct but if a vertex has 0 input result partitions, _all_ of the vertex's input partitions are _trivially_ ready. Maybe this helps to simplify the code and use virtually the same code path for both methods, or at least reuse some of the code.

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/strategy/LazyFromSourcesSchedulingStrategy.java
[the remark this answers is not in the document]
GJL: cc: @tillrohrmann

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/strategy/LazyFromSourcesSchedulingStrategy.java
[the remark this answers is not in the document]
eaglewatcherwb: @GJL @tillrohrmann  What do you think of using both vertex state transition and `onPartitionConsumable` notification ?

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/strategy/LazyFromSourcesSchedulingStrategy.java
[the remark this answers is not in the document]
eaglewatcherwb: Thanks for the excellent example, it really simplifies `restartTasks` code.

eaglewatcherwb:
Hi, @GJL @tillrohrmann , I have updated the PR based on the latest master. Would you mind to begin another round of code review ?

As discussed in the first round of code review, there are still some open questions, I make a summary to be  convenient for discussion.

1.  [[SchedulePolicy](https://github.com/apache/flink/pull/8309#discussion_r281049742)] Using vertex state transitions to schedule vertices has the benefit of avoiding flood `onPartitionConsumable` notifications, while there may be idle-waiting-result of PIPELINED shuffle mode. So, I think we could keep the benefit by relying on both vertex state transitions and `onPartitionConsumable` notifications.
	1) `DeploymentOption#sendScheduleOrUpdateConsumerMessage` set to true if the vertex has PIPELINED produced result partition and set to false if all the produced result partitions are BLOCKING
	2) Schedule vertices with BLOCKING input result partition using vertex state transition.
	3) Schedule vertices with PIPELINED input result partitions using `onPartitionConsumable` notification.

2. [[JobGraph Usage](https://github.com/apache/flink/pull/8309#discussion_r281037134)] The only usage of `JobGraph` is to provide `InputDependencyConstraint` in `LazyFromSourcesSchedulingStrategy`, while, it is not used in `EagerSchedulingStrategy`. Maybe we could remove `JobGraph` from `SchedulingStrategyFactory#createInstance` and add `InputDependencyConstraint` information into `SchedulingTopology`, which need an new interface in `SchedulingTopology`:
`InputDependencyConstraint getInputDependencyConstraint(JobVertexID jobVertexId)`?

3. [[ANY/ALL Schedule Granularity](https://issues.apache.org/jira/browse/FLINK-12229)]  In the original scheduler, the schedule granularity is ANY/ALL the IntermediateDataSet finishes, and using granularity of result partition could speedup deployments but may involve flood of partition update network communication and resource deadlock. Thus, in this PR my implementation is consistent with the original logic. 
However, we are wondering we could use some methods to keep both speedup deployments advantage and avoiding flood partition update and resource deadlock. Based on our production experience, we propose to introduce a new trigger `InputDependencyConstraint#Progress`, which is a float between 0.0~1.0 and identifies the percentage of the finish result partitions. 1.0 means ALL the input result partitions finish and we configured it to 0.8 as default to balance the speedup advantage and flood partition update, possible resource deadlock.

tillrohrmann:
Thanks for updating the PR @eaglewatcherwb.

1. This sounds like how we do lazy from sources scheduling at the moment. Thus +1 for doing it that way.

2. I think it is ok to add the `InputDependencyConstraint` to the `SchedulingTopology`.

3. I would not do this as part of this PR. We should rather do it as a follow up task.

eaglewatcherwb:
Thanks @tillrohrmann for the comment. Current PR has implemented as 1 and 2, and without 3. We could implement 3 as a future optimization.

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/strategy/LazyFromSourcesSchedulingStrategy.java
```diff
@@ -0,0 +1,296 @@
...
+			final SchedulingExecutionVertex schedulingVertex = schedulingTopology.getVertex(executionVertexId)
+				.orElseThrow(() -> new IllegalStateException("can not find scheduling vertex for " + executionVertexId));
+			final InputDependencyConstraint constraint = schedulingTopology
+				.getInputDependencyConstraint(schedulingVertex.getId().getJobVertexId())
+				.orElseThrow(() -> new IllegalStateException("can not find input dependency constraint for "
+					+ executionVertexId));
+
+			Collection<SchedulingResultPartition> resultPartitions = schedulingVertex.getConsumedResultPartitions();
+			boolean allConsumable = resultPartitions.isEmpty() ||
+				(resultPartitions.iterator().next().getPartitionType().isBlocking() && InputDependencyConstraint.ALL.equals(constraint));
```
GJL: Why is it required to check if the partition type is blocking, i.e., why is `InputDependencyConstraint.ALL.equals(constraint))` not enough? Why are we only checking the first result partition (`resultPartitions.iterator().next()`)?

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/strategy/LazyFromSourcesSchedulingStrategy.java
```diff
@@ -0,0 +1,296 @@
...
+
+		if (!producerVertex.getProducedResultPartitions().contains(resultPartition)) {
+			throw new IllegalStateException("partition " + resultPartitionId
+				+ " is not the produced partition of " + executionVertexId);
+		}
+
+		List<ExecutionVertexDeploymentOption> executionVertexDeploymentOptions = new ArrayList<>();
+		for (SchedulingExecutionVertex schedulingVertex : resultPartition.getConsumers()) {
+			if (ExecutionState.CREATED.equals(schedulingVertex.getState()) &&
+				resultPartition.getPartitionType().isPipelined() &&
```
GJL: What is the importance of this condition? I think the logic would be still correct without it.

```
resultPartition.getPartitionType().isPipelined()
```

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/strategy/LazyFromSourcesSchedulingStrategy.java
```diff
@@ -0,0 +1,296 @@
...
+				(PRODUCING.equals(resultPartition.getState()) || DONE.equals(resultPartition.getState()))) {
```
GJL: I would say the contract of `onPartitionConsumable` should be that if the method is called, the partition has data, i.e., it’s `PRODUCING` or `DONE`.

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/strategy/LazyFromSourcesSchedulingStrategyTest.java
```diff
@@ -0,0 +1,342 @@
...
+		partitionTypes[1] = BLOCKING;
+		inputDependencyConstraints[0] = ANY;
+		inputDependencyConstraints[1] = ANY;
+		inputDependencyConstraints[2] = InputDependencyConstraint.ALL;
+
+		initVerticesAndPartitions(schedulingTopology, partitionTypes, inputDependencyConstraints, vertices, partitions);
+
+		for (int i = 0; i < taskCnt; i++) {
+			for (int j = 0; j < taskCnt; j++) {
+				connectVerticesToPartition(vertices[0][i], vertices[2][i], partitions[0][j]);
```
GJL: It looks like that `vertices[0][0]` and `vertices[0][1]` produce the same partitions. This is not a legal topology. Any intermediate result partition has exactly one producer.

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/strategy/StrategyTestUtil.java
```diff
@@ -0,0 +1,102 @@
...
+		for (int i = 0; i < jobVertexCnt; i++) {
+			for (int j = 0; j < taskCnt; j++) {
+				vertices[i][j] = new TestingSchedulingExecutionVertex(jobVertexIds[i], j);
+				schedulingTopology.addSchedulingExecutionVertex(vertices[i][j]);
+
+				if (i != jobVertexCnt - 1) {
+					partitions[i][j] = new TestingSchedulingResultPartition(dataSetIds[i], new IntermediateResultPartitionID(),
+						partitionTypes[i], vertices[i][j]);
+					resultPartitionIds[i][j] = new ResultPartitionID(partitions[i][j].getId(), new ExecutionAttemptID());
+					schedulingTopology.addResultPartition(partitions[i][j].getId(), partitions[i][j]);
```
GJL: The method `addResultPartition` should not exist. It should only be allowed to add new partitions to the topology via `SchedulingExecutionVertex`.

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/strategy/LazyFromSourcesSchedulingStrategy.java
```diff
@@ -0,0 +1,296 @@
...
+
+	@Override
+	public void restartTasks(Set<ExecutionVertexID> verticesToRestart) {
+		// increase counter of the dataset first
+		for (ExecutionVertexID executionVertexId : verticesToRestart) {
+			final SchedulingExecutionVertex schedulingVertex = schedulingTopology.getVertex(executionVertexId)
+				.orElseThrow(() -> new IllegalStateException("can not find scheduling vertex for " + executionVertexId));
+
+			for (SchedulingResultPartition srp : schedulingVertex.getProducedResultPartitions()) {
+				if (srp.getPartitionType().isBlocking() && DONE.equals(srp.getState())) {
```
GJL: If I understand this code correctly, the condition `DONE.equals(srp.getState())` is needed to avoid overcounting in `SchedulingIntermediateDataSet`. Maybe it makes sense in `SchedulingIntermediateDataSet` to keep track of the `IntermediateResultPartitionIDs` that are still being produced to avoid overcounting.

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/strategy/LazyFromSourcesSchedulingStrategy.java
eaglewatcherwb: OK, `resultPartitions.isEmpty() || InputDependencyConstraint.ALL.equals(constraint)` is enough.

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/strategy/LazyFromSourcesSchedulingStrategy.java
eaglewatcherwb: OK

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/strategy/LazyFromSourcesSchedulingStrategy.java
eaglewatcherwb: Great, the code could be more simple if having the contract .

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/strategy/LazyFromSourcesSchedulingStrategyTest.java
eaglewatcherwb: OK

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/strategy/StrategyTestUtil.java
eaglewatcherwb: OK

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/strategy/LazyFromSourcesSchedulingStrategy.java
eaglewatcherwb: Yes, the condition of `DONE.equals(srp.getState())` is to avoid overcounting. One reason of my original thought of not keeping track of `IntermediateResultPartitionIDs` is to save the additional memory of the variables, but now I realize that there may be inconsistency between the counter and the partition state, e.g., restart vertex when partition state is DONE but having not increase the counter, the counter will be over decreased.

I think we should keep track of the `IntermediateResultPartitionIDs` of `DONE` state, which is added in `onExecutionStateChange`. When `restartTasks`, we could use `donePartitionIds.remove(srp.getId())` instead of `DONE.equals(srp.getState())`.

What do you think?

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/strategy/LazyFromSourcesSchedulingStrategyTest.java
```diff
@@ -0,0 +1,337 @@
...
+		scheduledVertices = testingSchedulerOperation.getScheduledVertices().get(1);
+		assertThat(getExecutionVertexIdsFromDeployOptions(scheduledVertices),
+			containsInAnyOrder(toBeScheduledVertices.toArray()));
+	}
+
+	/**
+	 * Tests that when partition consumable notified will start available {@link PIPELINED} downstream vertices.
+	 */
+	@Test
+	public void testPIPELINEDPartitionConsumable() {
```
GJL: This test and other tests suffer from a few issues:

- Tests are still a bit too long
- `initVerticesAndPartitions()` adds some automation but it is difficult to use and hard to understand 
  - One of the issues of this function is that it uses output arguments (see Clean Code by Robert C Martin, Chapter 3, page 45)
- Intuitively, I would say that if `SchedulingIntermediateDataSet` was unit tested separately, we do not have to make all test topologies parallel, i.e., only one execution vertex per operator
- This test sets all producer vertices to `RUNNING`, which is not needed for this test.

I have thought about how to improve the test, and you can find my prototype below. I am not sure if my approach can cover everything needed for all required test scenarios. Please have a look and let me know what you think.

```
//
// actual test code 
//

@Test
public void testPipelinedPartitionConsumable() {
	final TestingSchedulingTopology testingSchedulingTopology = new TestingSchedulingTopology();

	final List<TestingSchedulingExecutionVertex> producers = testingSchedulingTopology.addExecutionVertices().withParallelism(2).finish();
	final List<TestingSchedulingExecutionVertex> consumers = testingSchedulingTopology.addExecutionVertices().withParallelism(2).finish();
	testingSchedulingTopology.connectAllToAll(producers, consumers).withResultPartitionType(PIPELINED).finish();

	final LazyFromSourcesSchedulingStrategy schedulingStrategy = startScheduling(testingSchedulingTopology);

	final TestingSchedulingExecutionVertex producer1 = producers.get(0);
	final SchedulingResultPartition partition1 = producer1.getProducedResultPartitions().iterator().next();

	schedulingStrategy.onExecutionStateChange(producer1.getId(), ExecutionState.RUNNING);
	schedulingStrategy.onPartitionConsumable(producer1.getId(), new ResultPartitionID(partition1.getId(), new ExecutionAttemptID()));

        // an idea on how to make the assertions more readible
        // not insisting on it, could be added later
	assertThat(testingSchedulerOperation, hasScheduledVertices(consumers)); 
}

//
// hamcrest matcher
//

private static Matcher<TestingSchedulerOperations> hasScheduledVertices(final List<TestingSchedulingExecutionVertex> consumers) {

	final Matcher<Iterable<? extends ExecutionVertexID>> vertexIdMatcher = containsInAnyOrder(consumers.stream()
		.map(SchedulingExecutionVertex::getId)
		.toArray(ExecutionVertexID[]::new));

	return new TypeSafeDiagnosingMatcher<TestingSchedulerOperations>() {

		@Override
		protected boolean matchesSafely(final TestingSchedulerOperations item, final Description mismatchDescription) {
			final boolean matches = vertexIdMatcher.matches(toExecutionVertexIDs(item.getLatestScheduledVertices()));
			if (!matches) {
				vertexIdMatcher.describeMismatch(item.getLatestScheduledVertices(), mismatchDescription);
			}
			return matches;
		}

		@Override
		public void describeTo(final Description description) {
			vertexIdMatcher.describeTo(description);
		}
	};
}

//
// TestingSchedulingTopology
//

public SchedulingExecutionVerticesBuilder addExecutionVertices() {
		return new SchedulingExecutionVerticesBuilder();
	}

	public ProducerConsumerConnectionBuilder connectAllToAll(
			final List<TestingSchedulingExecutionVertex> producers,
			final List<TestingSchedulingExecutionVertex> consumers) {

		return new ProducerConsumerAllToAllConnectionBuilder(producers, consumers);
	}

	public abstract class ProducerConsumerConnectionBuilder {

		protected final List<TestingSchedulingExecutionVertex> producers;

		protected final List<TestingSchedulingExecutionVertex> consumers;

		protected ResultPartitionType resultPartitionType;

		protected ProducerConsumerConnectionBuilder(
				final List<TestingSchedulingExecutionVertex> producers,
				final List<TestingSchedulingExecutionVertex> consumers) {
			this.producers = producers;
			this.consumers = consumers;
		}

		public ProducerConsumerConnectionBuilder withResultPartitionType(final ResultPartitionType resultPartitionType) {
			this.resultPartitionType = resultPartitionType;
			return this;
		}

		public List<TestingSchedulingResultPartition> finish() {
			final List<TestingSchedulingResultPartition> resultPartitions = connect();

			TestingSchedulingTopology.this.addSchedulingExecutionVertices(producers);
			TestingSchedulingTopology.this.addSchedulingExecutionVertices(consumers);

			return resultPartitions;
		}

		protected TestingSchedulingResultPartition.Builder initTestingSchedulingResultPartitionBuilder() {
			return new TestingSchedulingResultPartition.Builder()
				.withResultPartitionType(resultPartitionType);
		}

		protected abstract List<TestingSchedulingResultPartition> connect();

	}

		private class ProducerConsumerAllToAllConnectionBuilder extends ProducerConsumerConnectionBuilder {

		private ProducerConsumerAllToAllConnectionBuilder(
				final List<TestingSchedulingExecutionVertex> producers,
				final List<TestingSchedulingExecutionVertex> consumers) {
			super(producers, consumers);
		}

		@Override
		protected List<TestingSchedulingResultPartition> connect() {
			final List<TestingSchedulingResultPartition> resultPartitions = new ArrayList<>();
			final IntermediateDataSetID intermediateDataSetId = new IntermediateDataSetID();

			for (TestingSchedulingExecutionVertex producer : producers) {

				final TestingSchedulingResultPartition resultPartition = initTestingSchedulingResultPartitionBuilder()
					.withIntermediateDataSetID(intermediateDataSetId)
					.build();
				resultPartition.setProducer(producer);
				producer.addProducedPartition(resultPartition);

				for (TestingSchedulingExecutionVertex consumer : consumers) {
					consumer.addConsumedPartition(resultPartition);
					resultPartition.addConsumer(consumer);
					resultPartitions.add(resultPartition);
				}
			}

			return resultPartitions;
		}
	}

	public class SchedulingExecutionVerticesBuilder {

		private final JobVertexID jobVertexId = new JobVertexID();

		private int parallelism = 1;

		private InputDependencyConstraint inputDependencyConstraint;

		public SchedulingExecutionVerticesBuilder withParallelism(final int parallelism) {
			this.parallelism = parallelism;
			return this;
		}

		public SchedulingExecutionVerticesBuilder withInputDependencyConstraint(final InputDependencyConstraint inputDependencyConstraint) {
			this.inputDependencyConstraint = inputDependencyConstraint;
			return this;
		}

		public List<TestingSchedulingExecutionVertex> finish() {
			final List<TestingSchedulingExecutionVertex> vertices = new ArrayList<>();
			for (int subtaskIndex = 0; subtaskIndex < parallelism; subtaskIndex++) {
				vertices.add(new TestingSchedulingExecutionVertex(jobVertexId, subtaskIndex, inputDependencyConstraint));
			}

			TestingSchedulingTopology.this.addSchedulingExecutionVertices(vertices);

			return vertices;
		}
	}

```

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/strategy/SchedulingTopology.java
```diff
@@ -50,4 +52,11 @@
 	 * @return Optional containing the respective scheduling result partition or none if the partition does not exist
 	 */
 	Optional<SchedulingResultPartition> getResultPartition(IntermediateResultPartitionID intermediateResultPartitionId);
+
+	/**
+	 * Looks up the {@link InputDependencyConstraint} for the given {@link JobVertexID}.
+	 * @param jobVertexId identifying the respective job vertex id
+	 * @return Optional containing the respective input dependency constraint or none if the job vertex does not exist
+	 */
+	Optional<InputDependencyConstraint> getInputDependencyConstraint(JobVertexID jobVertexId);
```
GJL: Since `InputDependencyConstraint` is a property of the `JobVertex`, I think it makes more sense to add this to the `SchedulingExecutionVertex` (without the `Optional`)

eaglewatcherwb: OK

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/strategy/LazyFromSourcesSchedulingStrategyTest.java
eaglewatcherwb: Thanks a lot for the valuable comments and the brilliant examples. I will add necessary methods for appropriate test scenarios.

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/strategy/LazyFromSourcesSchedulingStrategyTest.java
```diff
@@ -0,0 +1,230 @@
...
+
+		Set<ExecutionVertexID> verticesToRestart = producers.stream().map(TestingSchedulingExecutionVertex::getId)
+			.collect(Collectors.toSet());
+		verticesToRestart.addAll(consumers.stream().map(
+			TestingSchedulingExecutionVertex::getId).collect(Collectors.toSet()));
+
+		schedulingStrategy.restartTasks(verticesToRestart);
+		assertThat(testingSchedulerOperation, hasScheduledVertices(producers));
+	}
+
```
GJL: Strictly speaking we are missing some tests for `restartTasks()`. For example, we are not testing that a vertex should be scheduled if the partition is `PIPELINED` and in status `PRODUCING`. I think testing all combinations/branches from within `LazyFromSourcesSchedulingStrategyTest` will be difficult. I thought about extracting the `InputDependencyConstraint` checking into an extra class, which then can be tested in isolation. Find below an outline of what I mean:

```
class InputDependencyConstraintTester {

	private final SchedulingIntermediateDataSetManager schedulingIntermediateDataSetManager;

	public boolean test(final SchedulingExecutionVertex schedulingExecutionVertex) {
		if (schedulingExecutionVertex.getInputDependencyConstraint() == InputDependencyConstraint.ALL) {
			return testAll(schedulingExecutionVertex);
		} else if (schedulingExecutionVertex.getInputDependencyConstraint() == InputDependencyConstraint.ANY) {
			return testAny(schedulingExecutionVertex);
		} else {
			throw new IllegalArgumentException();
		}
	}

	private static boolean testAll(final SchedulingExecutionVertex schedulingExecutionVertex) {
		return schedulingExecutionVertex.getConsumedResultPartitions()
			.stream()
			.allMatch(this::partitionConsumable);
	}

	private static boolean testAny(final SchedulingExecutionVertex schedulingExecutionVertex) {
		return schedulingExecutionVertex.getConsumedResultPartitions()
			.stream()
			.anyMatch(this::partitionConsumable);
	}

	private boolean partitionConsumable(SchedulingResultPartition partition) {
		// ...
	}
}

class SchedulingIntermediateDataSetManager {

	private final Map<IntermediateDataSetID, SchedulingIntermediateDataSet> intermediateDataSets = new HashMap<>();

	// throws exception if dataset does not exist
	public SchedulingIntermediateDataSet getSchedulingIntermediateDataSet(final IntermediateDataSetID intermediateDataSetID);

}
```

I haven't tried whether this actually works. Let me know what you think.

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/strategy/LazyFromSourcesSchedulingStrategy.java
```diff
@@ -0,0 +1,287 @@
...
+	}
+
+	private boolean partitionConsumable(SchedulingResultPartition partition) {
+		SchedulingResultPartition.ResultPartitionState state = partition.getState();
+		if (partition.getPartitionType().isPipelined()) {
+			return PRODUCING.equals(state) || DONE.equals(state);
+		} else {
+			SchedulingIntermediateDataSet intermediateDataSet = intermediateDataSets.get(partition.getResultId());
+			if (intermediateDataSet == null) {
+				throw new IllegalStateException("can not find scheduling intermediate dataset for "
```
GJL: This check is duplicated. See my other comment in `LazyFromSourcesSchedulingStrategyTest` for a possible solution.

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/strategy/LazyFromSourcesSchedulingStrategy.java
```diff
@@ -0,0 +1,287 @@
...
+
+	private void scheduleBlockingConsumers(List<SchedulingExecutionVertex> vertices) {
+		Map<ExecutionVertexID, ExecutionVertexDeploymentOption> executionVertexDeploymentOptions = new HashMap<>();
+		for (SchedulingExecutionVertex schedulingVertex : vertices) {
+			final ExecutionState state = schedulingVertex.getState();
+
+			if (ExecutionState.CREATED.equals(state)) {
+				final InputDependencyConstraint inputDependency = schedulingVertex.getInputDependencyConstraint();
+
+				boolean shouldSchedule = true;
```
GJL: This looks similar to what happens in line 123:
```
shouldSchedule = resultPartitions.stream().allMatch(this::partitionConsumable);
```
Also see my other comment in `LazyFromSourcesSchedulingStrategyTest`

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/strategy/LazyFromSourcesSchedulingStrategyTest.java
```diff
@@ -0,0 +1,230 @@
...
+	public void testRestartTasks() {
+		final TestingSchedulingTopology testingSchedulingTopology = new TestingSchedulingTopology();
+
+		final List<TestingSchedulingExecutionVertex> producers = testingSchedulingTopology.addExecutionVertices().finish();
+		final List<TestingSchedulingExecutionVertex> consumers = testingSchedulingTopology.addExecutionVertices().finish();
+		testingSchedulingTopology.connectAllToAll(producers, consumers).finish();
+
+		LazyFromSourcesSchedulingStrategy schedulingStrategy = startScheduling(testingSchedulingTopology);
+
+		assertThat(testingSchedulerOperation, hasScheduledVertices(producers));
```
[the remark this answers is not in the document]
eaglewatcherwb: OK

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/strategy/LazyFromSourcesSchedulingStrategyTest.java
```diff
@@ -0,0 +1,230 @@
...
+			.map(SchedulingExecutionVertex::getId)
+			.toArray(ExecutionVertexID[]::new));
+
+		return new TypeSafeDiagnosingMatcher<TestingSchedulerOperations>() {
+
+			@Override
+			protected boolean matchesSafely(final TestingSchedulerOperations item, final Description mismatchDescription) {
+				final boolean matches = vertexIdMatcher.matches(getExecutionVertexIdsFromDeployOptions(item.getLatestScheduledVertices()));
+				if (!matches) {
+					vertexIdMatcher.describeMismatch(item.getLatestScheduledVertices(), mismatchDescription);
```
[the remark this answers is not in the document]
eaglewatcherwb: OK

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/strategy/LazyFromSourcesSchedulingStrategyTest.java
eaglewatcherwb: `SchedulingIntermediateDataSetManager` is a good idea, I will wrap up all the  `SchedulingIntermediateDataSet` functions into `SchedulingIntermediateDataSetManager`.

For `restartTasks` tests I expand `ProducerConsumerConnectionBuilder`, It seems easy to combine various `ResultPartitionType` and `ResultPartitionState` user cases. Thus I still add tests in `LazyFromSourcesSchedulingStrategyTest`, what do you think?

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/strategy/LazyFromSourcesSchedulingStrategy.java
eaglewatcherwb: OK,I will wrap it in `SchedulingIntermediateDataSetManager`.

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/strategy/LazyFromSourcesSchedulingStrategy.java
eaglewatcherwb: OK, I will wrap it in `SchedulingIntermediateDataSetManager`.

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/strategy/LazyFromSourcesSchedulingStrategy.java
```diff
@@ -0,0 +1,197 @@
...
+		}
+
+		schedulerOperations.allocateSlotsAndDeploy(executionVertexDeploymentOptions);
+	}
+
+	@Override
+	public void restartTasks(Set<ExecutionVertexID> verticesToRestart) {
+		// increase counter of the dataset first
+		for (ExecutionVertexID executionVertexId : verticesToRestart) {
+			final SchedulingExecutionVertex schedulingVertex = schedulingTopology.getVertex(executionVertexId)
```
GJL: Code is duplicated in line 101. For now I think it's ok to introduce a helper method inside the strategy. Eventually we have to think whether `Optional` return type makes sense, or if `schedulingTopology` should just throw an exception`.
```
	private SchedulingExecutionVertex getSchedulingVertex(final ExecutionVertexID executionVertexId) {
		return schedulingTopology.getVertex(executionVertexId)
			.orElseThrow(() -> new IllegalStateException("can not find scheduling vertex for " + executionVertexId));
	}
```

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/strategy/LazyFromSourcesSchedulingStrategy.java
```diff
@@ -0,0 +1,197 @@
...
+			deploymentOptions.put(schedulingVertex.getId(), option);
+
+			if (schedulingVertex.getConsumedResultPartitions().isEmpty()) {
+				// schedule vertices without consumed result partition
+				executionVertexDeploymentOptions.add(
+						new ExecutionVertexDeploymentOption(schedulingVertex.getId(), option));
+			}
+		}
+
+		schedulerOperations.allocateSlotsAndDeploy(executionVertexDeploymentOptions);
```
GJL: There is some common structure in `startScheduling` and `restartTasks`:

1. Initialization of data structures
1. scheduling of _"candidate"_ vertices

At least the 2nd part should be virtually the same:

```
@Override
	public void startScheduling() {
		// ... init, i.e., building the deployment options map, etc.

		allocateSlotsAndDeployExecutionVertexIds(getAllVerticesFromTopology());
	}

	@Override
	public void restartTasks(Set<ExecutionVertexID> verticesToRestart) {
		// ... init, i.e., resetting partitions

		allocateSlotsAndDeployExecutionVertexIds(verticesToRestart);
	}
```

There are also some similarities between `onExecutionStateChange` and `onPartitionConsumable`. Both schedule the consumers that are:

- in created state and
- have a satisfied input dependency 

So the outline of these functions looks something like this:
```
	@Override
	public void onExecutionStateChange(ExecutionVertexID executionVertexId, ExecutionState executionState) {
		if (!ExecutionState.FINISHED.equals(executionState)) {
			return;
		}

		final SchedulingExecutionVertex producer = getSchedulingVertex(executionVertexId);

		List<SchedulingResultPartition> finishedPartitions = new ArrayList<>();
		// ... populate finishedPartitions 

		final Set<SchedulingExecutionVertex> verticesToSchedule = finishedPartitions.stream()
			.flatMap(schedulingResultPartition -> schedulingResultPartition.getConsumers().stream())
			.collect(Collectors.toSet());

		allocateSlotsAndDeployExecutionVertices(verticesToSchedule);
	}

	@Override
	public void onPartitionConsumable(ExecutionVertexID executionVertexId, ResultPartitionID resultPartitionId) {
		final SchedulingResultPartition resultPartition = // ... look up result partition

		if (!resultPartition.getPartitionType().isPipelined()) {
			return;
		}

		final SchedulingExecutionVertex producerVertex = getSchedulingVertex(executionVertexId);

		// ... checkState

		allocateSlotsAndDeployExecutionVertices(resultPartition.getConsumers());
	}
```

We can remove redundancy by moving logic into some helper functions:
```
private SchedulingExecutionVertex getSchedulingVertex(final ExecutionVertexID executionVertexId) {
		return schedulingTopology.getVertex(executionVertexId)
			.orElseThrow(() -> new IllegalStateException("can not find scheduling vertex for " + executionVertexId));
	}

	private void allocateSlotsAndDeployExecutionVertexIds(final Collection<ExecutionVertexID> executionVertexIds) {
		allocateSlotsAndDeployExecutionVertices(executionVertexIds
			.stream()
			.map(this::getSchedulingVertex)
			.collect(Collectors.toList()));
	}

	private void allocateSlotsAndDeployExecutionVertices(final Collection<SchedulingExecutionVertex> schedulingExecutionVertices) {
		schedulerOperations.allocateSlotsAndDeploy(schedulingExecutionVertices
			.stream()
			.filter(isInputConstraintSatisfied().and(isInCreatedExecutionState()))
			.map(SchedulingExecutionVertex::getId)
			.map(executionVertexID -> new ExecutionVertexDeploymentOption(executionVertexID, deploymentOptions.get(executionVertexID)))
			.collect(Collectors.toSet()));
	}

	private Predicate<SchedulingExecutionVertex> isInCreatedExecutionState() {
		return schedulingExecutionVertex -> schedulingExecutionVertex.getState() == ExecutionState.CREATED;
	}

	private Predicate<SchedulingExecutionVertex> isInputConstraintSatisfied() {
		return inputConstraintChecker::check;
	}

	private Set<ExecutionVertexID> getAllVerticesFromTopology() {
		return StreamSupport
			.stream(schedulingTopology.getVertices().spliterator(), false)
			.map(SchedulingExecutionVertex::getId)
			.collect(Collectors.toSet());
	}
```

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/strategy/InputDependencyConstraintCheckerTest.java
```diff
@@ -0,0 +1,264 @@
...
+ */
+public class InputDependencyConstraintCheckerTest extends TestLogger {
+
+	@Test
+	public void testCheckInputVertex() {
+		final TestingSchedulingTopology testingSchedulingTopology = new TestingSchedulingTopology();
+
+		final List<TestingSchedulingExecutionVertex> producers = testingSchedulingTopology.addExecutionVertices().finish();
+
+		final InputDependencyConstraintChecker inputChecker = testingSchedulingTopology
```
GJL: I think it's better to create the `InputDependencyConstraintChecker` instance without `TestingSchedulingTopology` because:
- there are no complex relationships with other objects, such as `SchedulingExecutionVertex`
- InputDependencyConstraintChecker is only used inside this test but not in other tests

A static factory method `createInputDependencyConstraintChecker(List<? extends SchedulingResultPartition> resultPartitions)` inside the unit test should be enough.

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/strategy/LazyFromSourcesSchedulingStrategy.java
```diff
@@ -0,0 +1,197 @@
...
+
+	@Override
+	public void restartTasks(Set<ExecutionVertexID> verticesToRestart) {
+		// increase counter of the dataset first
+		for (ExecutionVertexID executionVertexId : verticesToRestart) {
+			final SchedulingExecutionVertex schedulingVertex = schedulingTopology.getVertex(executionVertexId)
+					.orElseThrow(() -> new IllegalStateException("can not find scheduling vertex for " + executionVertexId));
+
+			for (SchedulingResultPartition srp : schedulingVertex.getProducedResultPartitions()) {
+				if (srp.getPartitionType().isBlocking()) {
```
GJL: Do you think the code can be simplified if we track the intermediate data sets for pipelined partitions? I would be willing to pay the extra memory costs.

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/strategy/LazyFromSourcesSchedulingStrategy.java
eaglewatcherwb: Yes, I also think that `Optional` may be a little code redundant since `orElseThrow` is always checked. May be we could optimize this in another JIRA ticket.

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/strategy/LazyFromSourcesSchedulingStrategy.java
eaglewatcherwb: OK

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/strategy/InputDependencyConstraintCheckerTest.java
eaglewatcherwb: OK, involving `TestingSchedulingTopology` into `InputDependencyConstraintChecker` is not a good choice.

As I would like to test various combinations of `ResultPartitionType`, `ResultPartitionState` and `InputDependencyConstraint`, I add simple `TestingSchedulingExecutionVertexBuilder` and `TestingSchedulingResultPartitionBuilder` in `InputDependencyConstraintCheckerTest`.

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/strategy/LazyFromSourcesSchedulingStrategy.java
eaglewatcherwb: OK

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/strategy/LazyFromSourcesSchedulingStrategy.java
GJL: I think we can add `getVertexOrThrow(ExecutionVertexID)` to the topology. Can you create a ticket for that?

eaglewatcherwb: OK. [FLINK-12608](https://issues.apache.org/jira/browse/FLINK-12608)

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/strategy/TestingSchedulingResultPartition.java
```diff
@@ -0,0 +1,120 @@
...
+		this.producer = checkNotNull(producer);
+	}
+
+	/**
+	 * Builder for {@link TestingSchedulingResultPartition}.
+	 */
+	public static final class Builder {
+		private IntermediateDataSetID intermediateDataSetId;
+		private ResultPartitionType resultPartitionType;
+		private ResultPartitionState resultPartitionState;
```
tillrohrmann: Could be initialized with default values in order to avoid `null` value assignments.

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/strategy/InputDependencyConstraintChecker.java
```diff
@@ -0,0 +1,170 @@
...
+			sid.resetPartition(srp.getId());
+		}
+
+		void addSchedulingResultPartition(SchedulingResultPartition srp) {
+			SchedulingIntermediateDataSet sid = getSchedulingIntermediateDataSet(srp.getResultId(), true);
+			sid.addSchedulingResultPartition(srp);
+		}
+
+		boolean allPartitionsFinished(SchedulingResultPartition srp) {
+			SchedulingIntermediateDataSet sid = getSchedulingIntermediateDataSet(srp.getResultId(), false);
```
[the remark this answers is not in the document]
eaglewatcherwb: OK

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/strategy/InputDependencyConstraintChecker.java
```diff
@@ -0,0 +1,170 @@
...
+			return sid.allPartitionsFinished();
+		}
+
+		SchedulingIntermediateDataSet getSchedulingIntermediateDataSet(
+			final IntermediateDataSetID intermediateDataSetId,
+			boolean createIfAbsent) {
```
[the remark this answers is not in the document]
eaglewatcherwb: OK

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/strategy/InputDependencyConstraintChecker.java
```diff
@@ -0,0 +1,170 @@
...
+		boolean allPartitionsFinished() {
+			return producingPartitionIds.isEmpty();
+		}
+
+		void addSchedulingResultPartition(SchedulingResultPartition partition) {
+			partitions.add(partition);
+			producingPartitionIds.add(partition.getId());
+		}
+
+		List<SchedulingResultPartition> getSchedulingResultPartition() {
```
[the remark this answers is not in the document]
eaglewatcherwb: OK

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/strategy/TestingSchedulingResultPartition.java
eaglewatcherwb: OK

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/strategy/LazyFromSourcesSchedulingStrategy.java
```diff
@@ -0,0 +1,182 @@
...
+				.filter(isInputConstraintSatisfied().and(isInCreatedExecutionState()))
+				.map(SchedulingExecutionVertex::getId)
+				.map(executionVertexID -> new ExecutionVertexDeploymentOption(
+					executionVertexID,
+					deploymentOptions.get(executionVertexID)))
+				.collect(Collectors.toSet()));
+	}
+
+	private Predicate<SchedulingExecutionVertex> isInCreatedExecutionState() {
+		return schedulingExecutionVertex -> CREATED.equals(schedulingExecutionVertex.getState());
```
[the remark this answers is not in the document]
eaglewatcherwb: OK

PR #8310, opened 2019-04-29
[FLINK-12331][network] Introduce partition/gate setup to decouple task registration with NetworkEnvironment

zhijiangW:

*In order to decouple task with `NetworkEnvironment` completely, we introduce interface methods for `InputGate#setup` and `ResultPartitionWriter#setup`. Then the task could call the setup method for the created partitions/gates directly instead of calling current `registerTask` via `NetworkEnvironment`.*

## Brief change log

  - *Introduce `InputGate#setup` and `ResultPartitionWriter#setup` separately*
  - *Refactor the process of setup partition/gate in task to replace `registerTask`*
  - *Refactor the processes of creating `LocalBufferPool` in `SingleInputGate` and `ResultPartition`*

## Verifying this change

## Documentation


zhijiangW:

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/consumer/UnionInputGate.java
```diff
@@ -146,6 +146,13 @@ public boolean isFinished() {
 		return true;
 	}
 
+	@Override
+	public void setup() throws IOException {
+		for (InputGate inputGate : inputGates) {
```
azagrebin: I would keep this method empty if it is not really supposed to be called.

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/Task.java
```diff
@@ -599,20 +599,23 @@ else if (current == ExecutionState.CANCELING) {
 			}
 
 			// ----------------------------------------------------------------
-			// register the task with the network stack
+			// setup partitions and gates before task running
```
azagrebin: Move to separate method `Task.setupPartitionsAndGates()`?

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/InputGateFairnessTest.java
```diff
@@ -266,13 +281,14 @@ public void testFairConsumptionRemoteChannels() throws Exception {
 	//  Utilities
 	// ------------------------------------------------------------------------
 
-	private SingleInputGate createFairnessVerifyingInputGate(int numberOfChannels) {
+	private SingleInputGate createFairnessVerifyingInputGate(int numberOfChannels, NetworkBufferPool bufferPool) {
```
azagrebin: do we need to pass `bufferPool` here always as argument if it is a class field?

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/consumer/SingleInputGate.java
```diff
@@ -290,27 +304,38 @@ public String getOwningTaskName() {
...
 
 	/**
 	 * Assign the exclusive buffers to all remote input channels directly for credit-based mode.
-	 *
-	 * @param networkBufferPool The global pool to request and recycle exclusive buffers
-	 * @param networkBuffersPerChannel The number of exclusive buffers for each channel
 	 */
-	public void assignExclusiveSegments(NetworkBufferPool networkBufferPool, int networkBuffersPerChannel) throws IOException {
-		checkState(this.isCreditBased, "Bug in input gate setup logic: exclusive buffers only exist with credit-based flow control.");
-		checkState(this.networkBufferPool == null, "Bug in input gate setup logic: global buffer pool has" +
```
azagrebin: Maybe, we could keep this check `networkBufferPool == null` here where it is really relevant and allow nulls in the constructor? I would also keep the existing methods used in tests as they are now to avoid so many changes in tests at the moment because it looks like another refactoring which could be done at least in another commit.

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/InputGateFairnessTest.java
zhijiangW: Not need this field actually.  In my first version for this test it was not a class field. After refactoring to class field I forgot to change this part.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/consumer/SingleInputGate.java
zhijiangW: In the constructor we already check not null for `NetworkBufferPool`. Actually `assignExclusiveSegments` should be called only from `setup` which also checks `networkBufferPool == null` already, so I removed this check here and it have not changed anything in practice.

Considering not changing existing tests so much atm, so I retain this method public now. Do you mean if this method is used in tests separately, this removed check should be valid to be restored?

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/ResultPartition.java
```diff
@@ -131,6 +140,9 @@ public ResultPartition(
 		ResultPartitionType partitionType,
 		int numberOfSubpartitions,
 		int numTargetKeyGroups,
+		int networkBuffersPerChannel,
```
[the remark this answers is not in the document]
zhijiangW: That might be a good idea from the aspect of minimal changes especially for tests. I would consider and have a try.

zhijiangW:
Thanks for reviews @azagrebin !
I think your suggestion of migrating the creation of partition/gate to `TaskExecutor` is great and I would further consider it in a separate commit or PR. :)

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/consumer/SingleInputGate.java
zhijiangW: Considering the above comments of providing `Supplier` for `ResultPartition` and `SingleInputGate`, we do not need to adjust the existing logics in partition/gate.

zhijiangW:
I have updated the codes with squashed commit to address above comments.  The way of `Supplier` really simplifies the previous tests. :) @azagrebin

zhijiangW:
Thanks for further reviews @azagrebin 

My previous understanding of providing `Supplier` for partition/gate was not changing the existing main processes much, then the unit tests could almost work after the refactoring. 

But from the aspect of circular dependency you mentioned, I guess you mean the `NetworkEnvironment` relies on `ResultPartition` and `SingleInputGate` during setup `BufferPool`, meanwhile the constructors of `ResultPartition` and `SingleInputGate` also rely on some components of `NetworkEnvironment`. 

To be honest, I am not very sure what is the benefits of decoupling among them, because the `ResultPartition/SingleInputGate` are both from `NetworkEnvironment`. It seems reasonable to make them dependent with each other. 

- The proposed `MemorySegmentProvider` could avoid relying on `NetworkBufferPool` in partition/gate, but partition/gate would also rely on other components `ResultPartitionManager`, `ConnectionManager`, `IOManager` etc from `NetworkEnvironment`, so it still could not solve circular dependency completely. 

- We could not rely on `ResultPartition/SingleInputGate` directly in `NetworkEnvironment#setupPartition/Gate` by passing specific parameters `ResultPartitionType`, `numberOfChannels` etc. But considering not affecting tests, I have not changed this.

- The proposed `MemorySegmentProvider` interface seems a bit overlap with current `BufferProvider` in semantics. Furthermore we might have only one type of implementation in `NetworkEnvironment`. So it seems not very reasonable to define a new interface to replace specific `NetworkBufferPool` here.

Maybe there are any other concerns I have not caught up with? Nevertheless I update the codes based on your above suggestions of introducing `MemorySegmentProvider` and `Supplier<BufferPool>` (I have not fixed the tests), then we could further confirm whether this way is suitable.

PR #8312, opened 2019-04-29
[FLINK-12366][table] Clean up Catalog APIs to make them more consiste…

xuefuz:
…nt and coherent

Clean up existing catalog APIs to make them more consistent and coherent

## Brief change log

  - Clean up API signature regarding exception thrown
  - Improve error handling
 
## Verifying this change

This change is already covered by existing tests. Tests are modified to reflect the proposed changes.

  - The serializers: (no)

## Documentation

  - If yes, how is the feature documented? (avaDocs)

### DIFF
--- a/flink-table/flink-table-api-java/src/test/java/org/apache/flink/table/catalog/GenericInMemoryCatalogTest.java
+++ b/flink-table/flink-table-api-java/src/test/java/org/apache/flink/table/catalog/GenericInMemoryCatalogTest.java
@@ -474,12 +474,12 @@ public class GenericInMemoryCatalogTest extends CatalogTestBase {
 		catalog.createDatabase(db1, createDb(), false);
 		catalog.createTable(path1, table, false);
 
-		assertTrue(catalog.listAllPartitions(path1).isEmpty());
+		assertTrue(catalog.listPartitions(path1).isEmpty());
 
 		CatalogPartitionSpec partitionSpec = createPartitionSpec();
 		catalog.createPartition(path1, partitionSpec, createPartition(), false);
 
-		assertEquals(Arrays.asList(partitionSpec), catalog.listAllPartitions(path1));
+		assertEquals(Arrays.asList(partitionSpec), catalog.listPartitions(path1));
 		assertEquals(Arrays.asList(partitionSpec), catalog.listPartitions(path1, createPartitionSpecSubset()));
 		CatalogTestUtil.checkEquals(createPartition(), catalog.getPartition(path1, createPartitionSpec()));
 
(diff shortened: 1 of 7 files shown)


flink-table/flink-table-api-java/src/test/java/org/apache/flink/table/catalog/GenericInMemoryCatalogTest.java
```diff
@@ -731,32 +727,36 @@ public void testDropPartition() throws Exception {
...
+		assertEquals(Arrays.asList(createPartitionSpec()), catalog.listAllPartitions(path1));
 
 		catalog.dropPartition(path1, createPartitionSpec(), false);
 
-		assertEquals(Arrays.asList(), catalog.listPartitions(path1));
+		assertEquals(Arrays.asList(), catalog.listAllPartitions(path1));
 	}
 
 	@Test
 	public void testDropPartition_TableNotExistException() throws Exception {
```
bowenli86: We need to update the unit test names to reflect what case the test is actually about. E.g. this UT can be renamed to something like `testDropParition_tableNotExist()` or `testDropParition_ParitionNotExistException_tableNotExist()`.

Same thing for the following changed unit tests in this file.

xuefuz:
Thanks, Bowen! Updated the PR to address above your review feedback.

bowenli86:
@KurtYoung Would be good to have you take a look, and help to merge if there's no more concerns. Thanks!

flink-table/flink-table-common/src/main/java/org/apache/flink/table/catalog/ReadableCatalog.java
```diff
@@ -174,14 +172,11 @@
...
 	 *
-	 * @throws TableNotExistException thrown if the table does not exist in the catalog
-	 * @throws TableNotPartitionedException thrown if the table is not partitioned
-	 * @throws PartitionSpecInvalidException thrown if the given partition spec is invalid,
 	 * @throws PartitionNotExistException thrown if the partition is not partitioned
 	 * @throws CatalogException	in case of any runtime exception
 	 */
 	CatalogPartition getPartition(ObjectPath tablePath, CatalogPartitionSpec partitionSpec)
-		throws TableNotExistException, TableNotPartitionedException, PartitionSpecInvalidException, PartitionNotExistException, CatalogException;
+		throws PartitionNotExistException, CatalogException;
```
KurtYoung: I don't quite get the idea why this function will not throw `TableNotExistException`, but `listPartitions` will. These two methods are both some kind of "get some partition from table".

xuefuz: In short, partition not existing includes the case of table not existing. getPartition() attempts to get a specific partition for a table. If the table doesn't exist, then of course, the partition doesn't exist.

On the other hand, listPartition() fails (throws an exception) if the table doesn't exist, which is reasonable.

Just like getTable(), which doesn't throw databasenotexistEx when db doesn't exist. The change makes the behavior consistent across board.

KurtYoung:
LGTM, merging this

PR #8313, opened 2019-04-29
[FLINK-12184][Coordination]HistoryServerArchieFetcher incompatible with old version

klion26:

If we have old verison json files in history server, the HistoryServerArchiveFetcher can't convert legacy job overview. It will throw the NullPointerException when trying to convert legacy job overview because the tasks don't have the "pending" field.

There are actually two paths for the legacy job overview to consider, one is coming from Flink < 1.4 where we actually wrote the pending field and the other is for Flink >= 1.4 (legacy mode) which split the pending field up into scheduled, deployed and created.

## Brief change log
Add a condition check and add the right state.

## Verifying this change

`HistoryServerTest#testHistoryServerIntegration`

  - The serializers: (**no**)

## Documentation

@tillrohrmann @yumengz5

### DIFF
--- a/flink-runtime-web/src/main/java/org/apache/flink/runtime/webmonitor/history/HistoryServerArchiveFetcher.java
+++ b/flink-runtime-web/src/main/java/org/apache/flink/runtime/webmonitor/history/HistoryServerArchiveFetcher.java
@@ -248,7 +248,23 @@ class HistoryServerArchiveFetcher {
 
 		JsonNode tasks = job.get("tasks");
 		int numTasks = tasks.get("total").asInt();
-		int pending = tasks.get("pending").asInt();
+		JsonNode pendingNode = tasks.get("pending");
+		// for flink version < 1.4 we have pending field,
+		// when version >= 1.4 pending has been split into scheduled, deploying, and created.
+		boolean versionLessThan14 = pendingNode != null;
+		int created = 0;
+		int scheduled;
+		int deploying = 0;
+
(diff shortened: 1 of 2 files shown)


flink-runtime-web/src/main/java/org/apache/flink/runtime/webmonitor/history/HistoryServerArchiveFetcher.java
```diff
@@ -258,7 +269,13 @@ private static String convertLegacyJobOverview(String legacyOverview) throws IOE
...
 		// pending is a mix of CREATED/SCHEDULED/DEPLOYING
 		// to maintain the correct number of task states we have to pick one of them
-		tasksPerState[ExecutionState.SCHEDULED.ordinal()] = pending;
+		if (versionLessThan14) {
+			tasksPerState[ExecutionState.SCHEDULED.ordinal()] = pending;
+		} else {
+			tasksPerState[ExecutionState.SCHEDULED.ordinal()] = scheduled;
+			tasksPerState[ExecutionState.CREATED.ordinal()] = created;
+			tasksPerState[ExecutionState.DEPLOYING.ordinal()] = deploying;
+		}
```
tillrohrmann: Do we need the `versionLessThan14` distinction here? We could say that we assign `scheduled = pendingNode.asInt()` in line 257 and get completely rid of `pending`.

klion26:
@tillrohrmann thanks for the quick review.

PR #8314, opened 2019-04-29
[FLINK-12365][table] Add stats related catalog APIs

xuefuz:

Added stats related catalog APIs. Made table/column stats first-level meta object in catalog. Removed stats related info from table definition.

## Brief change log

*(for example:)*
  - Added stats related catalog APIs
  - Added stats related classes
  - Implemented these APIs in GenericInMemoryCatalog
  - Adapted CatalogTable definition for the new changes
  - Added tests

## Verifying this change

Added unit test coverage.

  - The serializers: (no)

## Documentation

  - Does this pull request introduce a new feature? (Yes)


flink-table/flink-table-common/src/main/java/org/apache/flink/table/catalog/ReadableCatalog.java
```diff
@@ -224,4 +224,54 @@ CatalogPartition getPartition(ObjectPath tablePath, CatalogPartitionSpec partiti
 	 */
 	boolean functionExists(ObjectPath functionPath) throws CatalogException;
 
+	// ------ Statistics ------
+
+	/**
+	 * Get the statistics of a table.
+	 *
+	 * @param tablePath path of the table
+	 * @return the statistics of the given table, null if its statistics don't exist or are unknown.
```
bowenli86: Stats should probably always exist. It's how Hive handles it. Also, the behavior of "returns null when the stats doesn't exist" doesn't seem to match behaviors of other metadata APIs and would confuse people, for example `getTable()` throws `TableNotExistException` when the request table doesn't exist. It's just the default, "unknown" stats will be with some default value, for example, rowCount, fileCount, totalSize will all be 0.

This will also better explain the signatures of these APIs.

What do you think?

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/GenericInMemoryCatalog.java
```diff
@@ -224,6 +235,8 @@ public void dropTable(ObjectPath tablePath, boolean ignoreIfNotExists) throws Ta
 			tables.remove(tablePath);
```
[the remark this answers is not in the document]
xuefuz: They are removed in line#238 and #239 below.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/GenericInMemoryCatalog.java
```diff
@@ -246,6 +259,16 @@ public void renameTable(ObjectPath tablePath, String newTableName, boolean ignor
 				if (partitions.containsKey(tablePath)) {
 					partitions.put(newPath, partitions.remove(tablePath));
 				}
+
```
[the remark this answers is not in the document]
xuefuz: Line # 263-271 does that.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/catalog/CatalogColumnStatistics.java
```diff
@@ -0,0 +1,125 @@
...
+
+	/**
+	 * max length of column values.
+	 */
+	private final Integer maxLen;
+
+	/**
+	 * max value of column values.
+	 */
+	private final Number max;
```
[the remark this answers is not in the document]
xuefuz: Good point! This is merely copied from Flink's column stats definition. If we think rework is needed, I think it's better to do it in a followup JIRA.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/catalog/ReadableCatalog.java
xuefuz: Make sense. I will leave out the "null" comment in those APIs and let implementations decide. It doesn't seem making sense to throw "stats not exist" exception, though.

flink-table/flink-table-api-java/src/test/java/org/apache/flink/table/catalog/GenericInMemoryCatalogTest.java
```diff
@@ -1104,6 +1104,43 @@ public void testDropFunction_FunctionNotExist_ignored() throws Exception {
 		catalog.dropDatabase(db1, false);
 	}
 
+	// ------ statistics ------
```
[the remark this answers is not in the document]
xuefuz: Well, I think the added test is written differently as others, which gives such an expression. However, it covers all the added implementation methods, as you can see. What's missing, though, is some negative test cases, which I will create a followup JIRA for this.

flink-table/flink-table-api-java/src/test/java/org/apache/flink/table/catalog/GenericInMemoryCatalogTest.java
```diff
@@ -1104,6 +1104,43 @@ public void testDropFunction_FunctionNotExist_ignored() throws Exception {
...
+		CatalogPartitionSpec partitionSpec = createPartitionSpec();
+		catalog.createPartition(path2, partitionSpec, createPartition(), false);
+		assertTrue(catalog.getPartitionStatistics(path2, partitionSpec) == null);
+		assertTrue(catalog.getPartitionColumnStatistics(path2, partitionSpec) == null);
+		catalog.alterPartitionStatistics(path2, partitionSpec, tableStatistics, false);
+		CatalogTestUtil.checkEquals(tableStatistics, catalog.getPartitionStatistics(path2, partitionSpec));
+		catalog.alterPartitionColumnStatistics(path2, partitionSpec, columnStatistics, false);
+		CatalogTestUtil.checkEquals(columnStatistics, catalog.getPartitionColumnStatistics(path2, partitionSpec));
+
+		// Clean up
```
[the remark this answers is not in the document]
xuefuz: Yeah. I guess they are harmless to be here and make this testcase complete on its own.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/catalog/ReadableCatalog.java
bowenli86: Yeah, it doesn't make sense to throw `stats not exist exception`. Just using the table API and an example to demonstrate the behavior difference.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/GenericInMemoryCatalog.java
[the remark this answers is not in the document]
bowenli86: Aren't line 238/239 only drops partition stats and partition column stats?

flink-table/flink-table-common/src/main/java/org/apache/flink/table/catalog/ReadableCatalog.java
```diff
@@ -224,4 +224,54 @@ CatalogPartition getPartition(ObjectPath tablePath, CatalogPartitionSpec partiti
...
+	/**
+	 * Get the statistics of a table.
+	 *
+	 * @param tablePath path of the table
+	 * @return the statistics of the given table
+	 *
+	 * @throws TableNotExistException if the table does not exist in the catalog
+	 * @throws CatalogException	in case of any runtime exception
+	 */
+	CatalogTableStatistics getTableStatistics(ObjectPath tablePath) throws TableNotExistException, CatalogException;
```
bowenli86: What happens if this API is called on a partitioned table?

flink-table/flink-table-common/src/main/java/org/apache/flink/table/catalog/ReadableCatalog.java
```diff
@@ -224,4 +224,54 @@ CatalogPartition getPartition(ObjectPath tablePath, CatalogPartitionSpec partiti
...
+	CatalogColumnStatistics getTableColumnStatistics(ObjectPath tablePath) throws TableNotExistException, CatalogException;
+
+	/**
+	 * Get the statistics of a partition.
+	 *
+	 * @param tablePath path of the table
+	 * @param partitionSpec partition spec of the partition
+	 * @return the statistics of the given partition
+	 *
+	 * @throws PartitionNotExistException if the partition is not partitioned
```
bowenli86: the wordage "if the partition is not partitioned" needs some rework

flink-table/flink-table-common/src/main/java/org/apache/flink/table/catalog/ReadableCatalog.java
```diff
@@ -224,4 +224,54 @@ CatalogPartition getPartition(ObjectPath tablePath, CatalogPartitionSpec partiti
...
+	/**
+	 * Get the column statistics of a table.
+	 *
+	 * @param tablePath path of the table
+	 * @return the column statistics of the given table
+	 *
+	 * @throws TableNotExistException if the table does not exist in the catalog
+	 * @throws CatalogException	in case of any runtime exception
+	 */
+	CatalogColumnStatistics getTableColumnStatistics(ObjectPath tablePath) throws TableNotExistException, CatalogException;
```
bowenli86: What does this API try to accomplish?  If this is for all columns of the table, should it return a list of `CatalogColumnStatistics`? If it's for a single column, should it take a column name?

Besides, what happens if this API is called on a partitioned table?

flink-table/flink-table-common/src/main/java/org/apache/flink/table/catalog/ReadableCatalog.java
```diff
@@ -224,4 +224,54 @@ CatalogPartition getPartition(ObjectPath tablePath, CatalogPartitionSpec partiti
...
+	 * Get the statistics of a partition.
+	 *
+	 * @param tablePath path of the table
+	 * @param partitionSpec partition spec of the partition
+	 * @return the statistics of the given partition
+	 *
+	 * @throws PartitionNotExistException if the partition is not partitioned
+	 * @throws CatalogException	in case of any runtime exception
+	 */
+	CatalogTableStatistics getPartitionStatistics(ObjectPath tablePath, CatalogPartitionSpec partitionSpec)
```
bowenli86: what happens if this API is called on a non-partitioned table?

flink-table/flink-table-common/src/main/java/org/apache/flink/table/catalog/ReadableCatalog.java
```diff
@@ -224,4 +224,54 @@ CatalogPartition getPartition(ObjectPath tablePath, CatalogPartitionSpec partiti
...
+	 * Get the column statistics of a partition.
+	 *
+	 * @param tablePath path of the table
+	 * @param partitionSpec partition spec of the partition
+	 * @return the column statistics of the given partition
+	 *
+	 * @throws PartitionNotExistException if the partition is not partitioned
+	 * @throws CatalogException	in case of any runtime exception
+	 */
+	CatalogColumnStatistics getPartitionColumnStatistics(ObjectPath tablePath, CatalogPartitionSpec partitionSpec)
```
bowenli86: ditto. What does this API try to accomplish?

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/GenericInMemoryCatalog.java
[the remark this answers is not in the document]
xuefuz: Yeah, you're right. Good catch.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/GenericInMemoryCatalog.java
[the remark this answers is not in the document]
xuefuz: Got it. Good catch!

flink-table/flink-table-common/src/main/java/org/apache/flink/table/catalog/ReadableCatalog.java
xuefuz: Partitioned table has its own, table-level stats, though it may not be complete or accurate, as in Hive's case.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/catalog/ReadableCatalog.java
xuefuz: good thinking. Let me rethink this. It seems to me that column stats of a table should be a list of stats for all columns. This probably requires some rework. Stay tuned.

xuefuz: Called on partitioned table is fine, though it may not exist or be accurate.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/catalog/ReadableCatalog.java
xuefuz: partition not exist exception.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/catalog/ReadableCatalog.java
bowenli86: ok, no problem. Please also take the APIs in ReadableWritableCatalog into consideration

flink-table/flink-table-api-java/src/test/java/org/apache/flink/table/catalog/GenericInMemoryCatalogTest.java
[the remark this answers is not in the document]
bowenli86: > Well, I think the added test is written differently as others, which gives such an expression. However, it covers all the added implementation methods, as you can see. What's missing, though, is some negative test cases, which I will create a followup JIRA for this.

I think we need to have a consistent style for unit tests in this class, otherwise it's hard for other developers to figure out the two styles and extend/modify the testing code

xuefuz:
I updated the PR with the following reworking:
1. Modified CatalogColumnStatistics class to enclose statistical info of all columns for a table/partition.
2. Added CatalogColumnStatisticsDataBase as a base class for column stats values that can be subclassed for different data types.
3. Added two stats value types, boolean and long, as examples. More types will be added in followup JIRAs.
4. Return "unknown" stats for GenericInMemoryCatalog when stats are not available for a table/partition.
5. Modfied tests and test utils accordingly.

bowenli86:
Hi @xuefuz , can you properly rebase to master? the commit history has some unrelated commits

xuefuz:
> Hi @xuefuz , can you properly rebase to master? the commit history has some unrelated commits

What unrelated commits you saw? I just rebased and rebuilt, and everything seems working fine to me.

xuefuz:
did a forced push and now it seems good.

xuefuz:
Thanks for the review, @bowenli86 ! I pushed another commit addressing above review comments. Plus, I will create a followup JIRA to add more test cases.

bowenli86:

flink-table/flink-table-common/src/main/java/org/apache/flink/table/catalog/stats/CatalogColumnStatisticsDataLong.java
```diff
@@ -0,0 +1,73 @@
...
+
+	/**
+	 * max value.
+	 */
+	private final long max;
+
+	/**
+	 * number of distinct values.
+	 */
+	private final long ndv;
```
KurtYoung: move ndv to `CatalogColumnStatisticsDataBase`? I think it's valid for all types of columns.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/catalog/stats/CatalogColumnStatisticsDataLong.java
```diff
@@ -0,0 +1,73 @@
...
+
+package org.apache.flink.table.catalog.stats;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Column statistics value of long type.
+ */
+public class CatalogColumnStatisticsDataLong extends CatalogColumnStatisticsDataBase {
```
KurtYoung: If we want to have a sub-class for each type we supported, i think we need to add them at once. The reason is after we introduce something like `CatalogColumnStatisticsDataInt`, `CatalogColumnStatisticsDataShort`, we may need to do some code refactor to have `CatalogColumnStatisticsNumericBase<T>`. It would be more clear if you introduce all types we support at the first step.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/catalog/stats/CatalogTableStatistics.java
```diff
@@ -0,0 +1,98 @@
...
+
+package org.apache.flink.table.catalog.stats;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Statistics for a non-partitioned table or a partition of a partitioned table.
+ */
+public class CatalogTableStatistics {
```
KurtYoung: how about letting this class holding `CatalogColumnStatistics`? It will be easier when converting catalog statistics to planner stats, we only need to pass in one variable.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/catalog/stats/CatalogColumnStatisticsDataLong.java
xuefuz: Maybe true, but Hive doesn't give ndv for boolean. I'm pretty much following Hive's stats definition. Let me know if you insist putting ndv for all types.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/catalog/stats/CatalogColumnStatisticsDataLong.java
xuefuz: Actually we will not have generic typed stats. Thus, adding more types are purely incremental.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/catalog/stats/CatalogColumnStatisticsDataLong.java
KurtYoung: I'm okay for this, and i just realized that it's also weird for `double` and `float` have ndv

flink-table/flink-table-common/src/main/java/org/apache/flink/table/catalog/stats/CatalogTableStatistics.java
xuefuz: Column stats are stored separately from table stats, and getting it needs additional trips and cost. Nevertheless, the purpose of this PR is to decouple stats related catalog classes from those used by planner, so planner will has its own stats definition (probably in its current forms), which might be different from what were defined here.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/catalog/stats/CatalogColumnStatisticsDataLong.java
xuefuz: Yeah. Hive uses double stats for both "double" and "float". Theoretically, ndvs for double might not be accurate, but from statistical point view, the value is still there. It's just that it cannot be used to answer queries such as select count(distinct).

xuefuz:
I have added more column stats types, leaving out Decimal because that's more involved to a different JIRA after this. @KurtYoung could you please take another look and merge this so that we can settle on the APIs at least. I have rebased this PR for quite a few times already. Thanks.

KurtYoung:
Changes LGTM, please fix the test failure:
21:22:05.992 [ERROR] testStatistics(org.apache.flink.table.catalog.GenericInMemoryCatalogTest)  Time elapsed: 0.011 s  <<< FAILURE!
java.lang.AssertionError: Use assertEquals(expected, actual, delta) to compare floating-point numbers
	at org.apache.flink.table.catalog.GenericInMemoryCatalogTest.testStatistics(GenericInMemoryCatalogTest.java:604)

xuefuz:
Pushed fix for the test. Thanks for the review, @KurtYoung .

PR #8315, opened 2019-04-29
[FLINK-12368] add subtask index to FlinkKafkaConsumerBase logging, wh…

stevenzwu:
…ich can be very useful when debugging problem


stevenzwu:
@Myasuka that is good to know.enable thread name in each log line can be an option. but I am not sure if we always want to enable thread in each log line just for this purpose.

`FlinkKafkaConsumerBase` already have many existing log lines with "Consumer subtask " prefix. this PR is to fill this useful info in other important lines where it is missing.

Myasuka:
@stevenzwu I found existing `logback.xml` in Flink already contains `thread name` information, and I create a jira [FLINK-12380](https://issues.apache.org/jira/browse/FLINK-12380) to track this.

In my opinion, developers would always feel bothered if we want sub-task index information to help debug but have to change the source code.

StephanEwen:
I think this is good to merge. I agree with @stevenzwu that this just continues a pattern that is in the consumer anyways.

In the long run, it may make sense to think about exposing loggers that always pre-pend a short operator name, like `[<operator_name>-<subtask>]` or so.

StephanEwen:
@flinkbot approve all

StephanEwen:
Merging this...

PR #8318, opened 2019-04-30
[FLINK-12231][runtime] Introduce Scheduler interface and adapt ExecutionGraph to it

GJL:

*This PR introduces the SchedulerNG interface.*

## Brief change log

  - *Introduce `SchedulerNG` interface*
  - *Remove `ExecutionGraph` from JobMaster*

## Verifying this change

This change is already covered by existing tests, such as *JobMasterTest*.

## Documentation


flink-runtime/src/main/java/org/apache/flink/runtime/jobmaster/JobMaster.java
```diff
@@ -1006,101 +760,58 @@ private Acknowledge suspendExecution(final Exception cause) {
...
+	private void resetAndStartScheduler() throws Exception {
 		validateRunsInMainThread();
 
-		final CompletableFuture<Void> executionGraphAssignedFuture;
+		final CompletableFuture<Void> schedulerAssignedFuture;
 
-		if (executionGraph.getState() == JobStatus.CREATED) {
-			executionGraphAssignedFuture = CompletableFuture.completedFuture(null);
-			executionGraph.start(getMainThreadExecutor());
+		if (schedulerNG.requestJobStatus() == JobStatus.CREATED) {
```
GJL: Here we call a method on the scheduler without setting the main thread executor.

flink-runtime/src/main/java/org/apache/flink/runtime/jobmaster/LegacyScheduler.java
```diff
@@ -0,0 +1,639 @@
...
+
+				if (throwable != null) {
+					log.info("Failed during stopping job {} with a savepoint. Reason: {}", jobGraph.getJobID(), throwable.getMessage());
+					throw new CompletionException(throwable);
+				} else if(jobstatus != JobStatus.FINISHED) {
+					log.info("Failed during stopping job {} with a savepoint. Reason: Reached state {} instead of FINISHED.", jobGraph.getJobID(), jobstatus);
+					throw new CompletionException(new FlinkException("Reached state " + jobstatus + " instead of FINISHED."));
+				}
+				return jobstatus;
+			}, mainThreadExecutor);
```
GJL: Is the main thread executor really needed here? We don't modify state inside `handleAsync()`.

flink-runtime/src/main/java/org/apache/flink/runtime/jobmaster/LegacyScheduler.java
```diff
@@ -0,0 +1,639 @@
...
+
+		final CompletableFuture<String> savepointFuture = checkpointCoordinator
+			.triggerSynchronousSavepoint(now, advanceToEndOfEventTime, targetDirectory)
+			.handleAsync((completedCheckpoint, throwable) -> {
+				if (throwable != null) {
+					log.info("Failed during stopping job {} with a savepoint. Reason: {}", jobGraph.getJobID(), throwable.getMessage());
+					throw new CompletionException(throwable);
+				}
+				return completedCheckpoint.getExternalPointer();
+			}, mainThreadExecutor);
```
GJL: Is the main thread executor really needed here? We don't modify state inside `handleAsync()`.

flink-runtime/src/main/java/org/apache/flink/runtime/jobmaster/LegacyScheduler.java
```diff
@@ -0,0 +1,639 @@
...
+		mainThreadExecutor.assertRunningInMainThread();
+		return WebMonitorUtils.createDetailsForJob(executionGraph);
+	}
+
+	@Override
+	public KvStateLocation requestKvStateLocation(final JobID jobId, final String registrationName) throws UnknownKvStateLocation, FlinkJobNotFoundException {
+		mainThreadExecutor.assertRunningInMainThread();
+
+		// sanity check for the correct JobID
+		if (jobGraph.getJobID().equals(jobId)) {
```
GJL: `jobId` isn't really needed at this point.

flink-runtime/src/main/java/org/apache/flink/runtime/jobmaster/JobMaster.java
```diff
@@ -392,43 +379,10 @@ public JobMaster(
...
-			final byte[] serializedInputSplit = InstantiationUtil.serializeObject(nextInputSplit);
-			return CompletableFuture.completedFuture(new SerializedInputSplit(serializedInputSplit));
-		} catch (Exception ex) {
-			log.error("Could not serialize the next input split of class {}.", nextInputSplit.getClass(), ex);
-			IOException reason = new IOException("Could not serialize the next input split of class " +
-					nextInputSplit.getClass() + ".", ex);
-			vertex.fail(reason);
-			return FutureUtils.completedExceptionally(reason);
+			return CompletableFuture.completedFuture(schedulerNG.requestNextInputSplit(vertexID, executionAttempt));
+		} catch (IOException e) {
```
tillrohrmann: I would also add a logging statement logging the `e` because it might potentially leave the distributed component here.

flink-runtime/src/main/java/org/apache/flink/runtime/jobmaster/JobMaster.java
```diff
@@ -437,43 +391,20 @@ public JobMaster(
...
-				} else {
-					return FutureUtils.completedExceptionally(new PartitionProducerDisposedException(resultPartitionId));
-				}
-			} else {
-				return FutureUtils.completedExceptionally(new IllegalArgumentException("Intermediate data set with ID "
-						+ intermediateResultId + " not found."));
-			}
+		try {
+			return CompletableFuture.completedFuture(schedulerNG.requestPartitionState(intermediateResultId, resultPartitionId));
+		} catch (PartitionProducerDisposedException e) {
```
tillrohrmann: I would also log the exception here before it leaves the component.

flink-runtime/src/main/java/org/apache/flink/runtime/jobmaster/JobMaster.java
```diff
@@ -501,76 +432,21 @@ public void acknowledgeCheckpoint(
...
-				return FutureUtils.completedExceptionally(new UnknownKvStateLocation(registrationName));
-			}
-		} else {
-			if (log.isDebugEnabled()) {
-				log.debug("Request of key-value state location for unknown job {} received.", jobId);
-			}
-			return FutureUtils.completedExceptionally(new FlinkJobNotFoundException(jobId));
+		try {
+			return CompletableFuture.completedFuture(schedulerNG.requestKvStateLocation(jobId, registrationName));
+		} catch (UnknownKvStateLocation | FlinkJobNotFoundException e) {
```
tillrohrmann: logging statement missing

flink-runtime/src/main/java/org/apache/flink/runtime/jobmaster/JobMaster.java
```diff
@@ -582,26 +458,12 @@ public void declineCheckpoint(DeclineCheckpoint decline) {
...
-			}
-		} else {
-			if (log.isDebugEnabled()) {
-				log.debug("Notification about key-value state registration for unknown job {} received.", jobId);
-			}
-			return FutureUtils.completedExceptionally(new FlinkJobNotFoundException(jobId));
+		try {
+			schedulerNG.notifyKvStateRegistered(jobId, jobVertexId, keyGroupRange, registrationName, kvStateId, kvStateServerAddress);
+			return CompletableFuture.completedFuture(Acknowledge.get());
+		} catch (FlinkJobNotFoundException e) {
```
tillrohrmann: same here

flink-runtime/src/main/java/org/apache/flink/runtime/jobmaster/JobMaster.java
```diff
@@ -611,26 +473,11 @@ public void declineCheckpoint(DeclineCheckpoint decline) {
...
-			}
-		} else {
-			if (log.isDebugEnabled()) {
-				log.debug("Notification about key-value state deregistration for unknown job {} received.", jobId);
-			}
-			return FutureUtils.completedExceptionally(new FlinkJobNotFoundException(jobId));
+		try {
+			schedulerNG.notifyKvStateUnregistered(jobId, jobVertexId, keyGroupRange, registrationName);
+			return CompletableFuture.completedFuture(Acknowledge.get());
+		} catch (FlinkJobNotFoundException e) {
```
tillrohrmann: and here

flink-runtime/src/main/java/org/apache/flink/runtime/jobmaster/JobMaster.java
```diff
@@ -809,80 +625,18 @@ public void heartbeatFromResourceManager(final ResourceID resourceID) {
...
 	public CompletableFuture<OperatorBackPressureStatsResponse> requestOperatorBackPressureStats(final JobVertexID jobVertexId) {
-		final ExecutionJobVertex jobVertex = executionGraph.getJobVertex(jobVertexId);
-		if (jobVertex == null) {
-			return FutureUtils.completedExceptionally(new FlinkException("JobVertexID not found " +
-				jobVertexId));
+		try {
+			final Optional<OperatorBackPressureStats> operatorBackPressureStats = schedulerNG.requestOperatorBackPressureStats(jobVertexId);
+			return CompletableFuture.completedFuture(OperatorBackPressureStatsResponse.of(
+				operatorBackPressureStats.orElse(null)));
+		} catch (FlinkException e) {
```
tillrohrmann: logging statement missing

flink-runtime/src/main/java/org/apache/flink/runtime/jobmaster/JobMaster.java
tillrohrmann: This should be gone once https://issues.apache.org/jira/browse/FLINK-11719 has been resolved.

flink-runtime/src/main/java/org/apache/flink/runtime/jobmaster/LegacyScheduler.java
```diff
@@ -0,0 +1,639 @@
...
+			if (log.isDebugEnabled()) {
+				log.debug("Can not find Execution for attempt {}.", executionAttempt);
+			}
+			// but we should TaskManager be aware of this
+			throw new IllegalArgumentException("Can not find Execution for attempt " + executionAttempt);
+		}
+
+		final ExecutionJobVertex vertex = executionGraph.getJobVertex(vertexID);
+		if (vertex == null) {
+			log.error("Cannot find execution vertex for vertex ID {}.", vertexID);
```
tillrohrmann: I think we could remove the error logging statements and log them on the `JobMaster` level.

flink-runtime/src/main/java/org/apache/flink/runtime/jobmaster/LegacyScheduler.java
tillrohrmann: I don't think so. I guess it should simply be `handle`

flink-runtime/src/main/java/org/apache/flink/runtime/jobmaster/LegacyScheduler.java
tillrohrmann: Same here.

flink-runtime/src/main/java/org/apache/flink/runtime/jobmaster/LegacyScheduler.java
GJL: After the `throw`, we potentially leave the _distributed component_, which means we should also catch `IllegalArgumentException` in `JobMaster#requestNextInputSplit()` to log the exception before returning an exceptional future. It would be consequent to catch all `RuntimeExceptions` on `JobMaster` side before leaving the _distributed component_, which is evidently not done for most of the methods. Why not log exceptional futures in the rpc framework?

flink-runtime/src/main/java/org/apache/flink/runtime/jobmaster/JobMaster.java
GJL: I guess it's ok not to protect `requestJobStatus()` against invocations from outside the main thread for now.

flink-runtime/src/main/java/org/apache/flink/runtime/jobmaster/JobMaster.java
GJL: Then I wonder what we should do about unchecked exceptions. See my other comment.

flink-runtime/src/main/java/org/apache/flink/runtime/jobmaster/SchedulerNG.java
```diff
@@ -0,0 +1,109 @@
...
+import org.apache.flink.runtime.taskmanager.TaskExecutionState;
+import org.apache.flink.util.FlinkException;
+
+import javax.annotation.Nullable;
+
+import java.io.IOException;
+import java.net.InetSocketAddress;
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+
```
shuai-xu: Missing java doc. What's NG standing for?

GJL: NG stands for **N**ew **G**eneration. This suffix will be removed as soon as we renamed the existing `Scheduler` interface.

GJL: Added Javadoc.

flink-runtime/src/main/java/org/apache/flink/runtime/jobmaster/JobMaster.java
```diff
@@ -582,26 +461,13 @@ public void declineCheckpoint(DeclineCheckpoint decline) {
...
-				log.error("Failed to notify KvStateRegistry about registration {}.", registrationName, e);
-				return FutureUtils.completedExceptionally(e);
-			}
-		} else {
-			if (log.isDebugEnabled()) {
-				log.debug("Notification about key-value state registration for unknown job {} received.", jobId);
-			}
-			return FutureUtils.completedExceptionally(new FlinkJobNotFoundException(jobId));
+		try {
+			schedulerNG.notifyKvStateRegistered(jobId, jobVertexId, keyGroupRange, registrationName, kvStateId, kvStateServerAddress);
```
tillrohrmann: As a follow up refactoring one could get rid of the `jobId` and thereby getting rid of the `FlinkJobNotFoundException`. Then we would not have to log anything.

GJL:
Rebased to latest master. Merging as soon as build is green.

PR #8320, opened 2019-04-30
[FLINK-12201][network,metrics] Introduce InputGateWithMetrics in Task to increment numBytesIn metric

zhijiangW:

* Incrementing of `numBytesIn` metric in `SingleInputGate` does not depend on shuffle service and can be moved out of network internals into `Task`. `Task` could wrap `InputGate` provided by `ShuffleService` with `InputGateWithMetrics` which would increment numBytesIn metric.*

## Brief change log

  - *Introduce `InputGateWithMetrics` to wrap `InputGate` from `ShuffleService` in `Task`*
  - *Remove the counter from `SingleInputGate` and related creation.*

## Verifying this change

## Documentation

### DIFF
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/buffer/Buffer.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/buffer/Buffer.java
@@ -159,17 +159,6 @@ public interface Buffer {
 	void setReaderIndex(int readerIndex) throws IndexOutOfBoundsException;
 
 	/**
-	 * Returns the size of the written data, i.e. the <tt>writer index</tt>, of this buffer in an
-	 * non-synchronized fashion.
-	 *
-	 * <p>This is where writable bytes start in the backing memory segment.
-	 *
-	 * @return writer index (from 0 (inclusive) to the size of the backing {@link MemorySegment}
-	 * (inclusive))
-	 */
-	int getSizeUnsafe();
-
-	/**
 	 * Returns the size of the written data, i.e. the <tt>writer index</tt>, of this buffer.
 	 *
 	 * <p>This is where writable bytes start in the backing memory segment.
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/consumer/BufferOrEvent.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/consumer/BufferOrEvent.java
@@ -18,6 +18,7 @@
 
 package org.apache.flink.runtime.io.network.partition.consumer;
 
+import org.apache.flink.annotation.VisibleForTesting;
 import org.apache.flink.runtime.event.AbstractEvent;
 import org.apache.flink.runtime.io.network.buffer.Buffer;
 
@@ -43,26 +44,32 @@ public class BufferOrEvent {
 
 	private int channelIndex;
 
-	BufferOrEvent(Buffer buffer, int channelIndex, boolean moreAvailable) {
+	private final int size;
+
+	public BufferOrEvent(Buffer buffer, int channelIndex, boolean moreAvailable) {
 		this.buffer = checkNotNull(buffer);
 		this.event = null;
 		this.channelIndex = channelIndex;
 		this.moreAvailable = moreAvailable;
+		this.size = buffer.getSize();
 	}
 
-	BufferOrEvent(AbstractEvent event, int channelIndex, boolean moreAvailable) {
+	public BufferOrEvent(AbstractEvent event, int channelIndex, boolean moreAvailable, int size) {
 		this.buffer = null;
 		this.event = checkNotNull(event);
 		this.channelIndex = channelIndex;
 		this.moreAvailable = moreAvailable;
+		this.size = size;
 	}
 
+	@VisibleForTesting
 	public BufferOrEvent(Buffer buffer, int channelIndex) {
 		this(buffer, channelIndex, true);
 	}
 
+	@VisibleForTesting
 	public BufferOrEvent(AbstractEvent event, int channelIndex) {
-		this(event, channelIndex, true);
+		this(event, channelIndex, true, 0);
 	}
 
 	public boolean isBuffer() {
@@ -96,11 +103,15 @@ public class BufferOrEvent {
 
 	@Override
 	public String toString() {
-		return String.format("BufferOrEvent [%s, channelIndex = %d]",
-				isBuffer() ? buffer : event, channelIndex);
+		return String.format("BufferOrEvent [%s, channelIndex = %d, size = %d]",
+				isBuffer() ? buffer : event, channelIndex, size);
 	}
 
 	public void setMoreAvailable(boolean moreAvailable) {
 		this.moreAvailable = moreAvailable;
+	}
+
+	public int getSize() {
+		return size;
 	}
 }
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/consumer/SingleInputGate.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/consumer/SingleInputGate.java
@@ -19,7 +19,6 @@
 package org.apache.flink.runtime.io.network.partition.consumer;
 
 import org.apache.flink.annotation.VisibleForTesting;
-import org.apache.flink.metrics.Counter;
 import org.apache.flink.runtime.deployment.InputChannelDeploymentDescriptor;
 import org.apache.flink.runtime.deployment.ResultPartitionLocation;
 import org.apache.flink.runtime.event.AbstractEvent;
@@ -165,8 +164,6 @@ public class SingleInputGate extends InputGate {
 	/** A timer to retrigger local partition requests. Only initialized if actually needed. */
 	private Timer retriggerLocalRequestTimer;
 
-	private final Counter numBytesIn;
-
 	private final SupplierWithException<BufferPool, IOException> bufferPoolFactory;
 
 	private final CompletableFuture<Void> closeFuture;
@@ -178,7 +175,6 @@ public class SingleInputGate extends InputGate {
 		int consumedSubpartitionIndex,
 		int numberOfInputChannels,
 		PartitionProducerStateProvider partitionProducerStateProvider,
-		Counter numBytesIn,
 		boolean isCreditBased,
 		SupplierWithException<BufferPool, IOException> bufferPoolFactory) {
 
@@ -199,8 +195,6 @@ public class SingleInputGate extends InputGate {
 		this.enqueuedInputChannelsWithData = new BitSet(numberOfInputChannels);
 
 		this.partitionProducerStateProvider = checkNotNull(partitionProducerStateProvider);
-
-		this.numBytesIn = checkNotNull(numBytesIn);
 
 		this.isCreditBased = isCreditBased;
 
@@ -566,7 +560,6 @@ public class SingleInputGate extends InputGate {
 			Buffer buffer,
 			boolean moreAvailable,
 			InputChannel currentChannel) throws IOException, InterruptedException {
-		numBytesIn.inc(buffer.getSizeUnsafe());
 		if (buffer.isBuffer()) {
 			return new BufferOrEvent(buffer, currentChannel.getChannelIndex(), moreAvailable);
 		}
@@ -596,7 +589,7 @@ public class SingleInputGate extends InputGate {
 				currentChannel.releaseAllResources();
 			}
 
-			return new BufferOrEvent(event, currentChannel.getChannelIndex(), moreAvailable);
+			return new BufferOrEvent(event, currentChannel.getChannelIndex(), moreAvailable, buffer.getSize());
 		}
 	}
 
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/InputGateWithMetrics.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/InputGateWithMetrics.java
@@ -1,0 +1,107 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.runtime.taskmanager;
+
+import org.apache.flink.metrics.Counter;
+import org.apache.flink.runtime.event.TaskEvent;
+import org.apache.flink.runtime.io.network.partition.consumer.BufferOrEvent;
+import org.apache.flink.runtime.io.network.partition.consumer.InputGate;
+import org.apache.flink.runtime.metrics.groups.TaskIOMetricGroup;
+
+import java.io.IOException;
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+
+import static org.apache.flink.util.Preconditions.checkNotNull;
+
+/**
+ * This class wraps {@link InputGate} provided by shuffle service and it is mainly
+ * used for increasing general input metrics from {@link TaskIOMetricGroup}.
+ */
+public class InputGateWithMetrics extends InputGate {
+
+	private final InputGate inputGate;
+
+	private final Counter numBytesIn;
+
+	public InputGateWithMetrics(InputGate inputGate, Counter numBytesIn) {
+		this.inputGate = checkNotNull(inputGate);
+		this.numBytesIn = checkNotNull(numBytesIn);
+	}
+
+	@Override
+	public CompletableFuture<?> isAvailable() {
+		return inputGate.isAvailable();
+	}
+
+	@Override
+	public int getNumberOfInputChannels() {
+		return inputGate.getNumberOfInputChannels();
+	}
+
+	@Override
+	public String getOwningTaskName() {
+		return inputGate.getOwningTaskName();
+	}
+
+	@Override
+	public boolean isFinished() {
+		return inputGate.isFinished();
+	}
+
+	@Override
+	public void setup() throws IOException {
+		inputGate.setup();
+	}
+
+	@Override
+	public void requestPartitions() throws IOException, InterruptedException {
+		inputGate.requestPartitions();
+	}
+
+	@Override
+	public Optional<BufferOrEvent> getNextBufferOrEvent() throws IOException, InterruptedException {
+		return updateMetrics(inputGate.getNextBufferOrEvent());
+	}
+
+	@Override
+	public Optional<BufferOrEvent> pollNextBufferOrEvent() throws IOException, InterruptedException {
+		return updateMetrics(inputGate.pollNextBufferOrEvent());
+	}
+
+	@Override
+	public void sendTaskEvent(TaskEvent event) throws IOException {
+		inputGate.sendTaskEvent(event);
+	}
+
+	@Override
+	public int getPageSize() {
+		return inputGate.getPageSize();
+	}
+
+	@Override
+	public void close() throws Exception {
+		inputGate.close();
+	}
+
+	private Optional<BufferOrEvent> updateMetrics(Optional<BufferOrEvent> bufferOrEvent) {
+		bufferOrEvent.ifPresent(b -> numBytesIn.inc(b.getSize()));
+		return bufferOrEvent;
+	}
+}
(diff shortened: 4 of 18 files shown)


zhijiangW:

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/Task.java
```diff
@@ -191,7 +191,7 @@
 
 	private final ResultPartition[] producedPartitions;
 
-	private final SingleInputGate[] inputGates;
+	private final InputGate[] inputGates;
```
azagrebin: I would keep it as `SingleInputGate[]` to avoid that copying loop `SingleInputGate[]` -> `InputGate[]` in Task constructor at the moment.

Later, we could move `inputGatesById` completely to `NetworkEnviroment`, when we refactor:
- `TaskExecutor.updatePartitions` to use `NetworkEnviroment.update(PartitionInfo)`
- try to move `triggerPartitionProducerStateCheck` to `SingleInputGate` and pass `partitionProducerStateChecker` to `SingleInputGate` instead of having it in `TaskActions`

`inputGatesById` looks like the last blocker to change `SingleInputGate[]` -> `InputGate[]` in Task.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/consumer/BufferOrEvent.java
```diff
@@ -43,26 +46,30 @@
 
 	private int channelIndex;
 
-	BufferOrEvent(Buffer buffer, int channelIndex, boolean moreAvailable) {
+	public BufferOrEvent(Buffer buffer, int channelIndex, boolean moreAvailable, int size) {
 		this.buffer = checkNotNull(buffer);
 		this.event = null;
 		this.channelIndex = channelIndex;
 		this.moreAvailable = moreAvailable;
+		this.size = size;
```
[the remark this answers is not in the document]
zhijiangW: Yes, I think it is available before.

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/Task.java
zhijiangW: I agree with the following refactors for `updatePartitions` and `triggerPartitionProducerStateCheck` which could help remove `inputGatesById` in task.

But the change of `SingleInputGate[]` to `InputGate[]` is necessary here because we have to loop the `SingleInputGate[]` from `NetworkEnvironment#createInputGates` to generate `InputGateWithMetrics` which implements `InputGate`. The task would finally maintain `InputGate[]` instead of `SingleInputGate`, so this change would be consistent with future plans.

Did I misunderstand your concern?

azagrebin: Oh, sorry, you are right, too fast review :)

azagrebin:
@flinkbot approve all

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/InputGateWithMetrics.java
```diff
@@ -0,0 +1,119 @@
...
+
+	@Override
+	public void requestPartitions() throws IOException, InterruptedException {
+		inputGate.requestPartitions();
+	}
+
+	@Override
+	public Optional<BufferOrEvent> getNextBufferOrEvent() throws IOException, InterruptedException {
+		Optional<BufferOrEvent> bufferOrEvent = inputGate.getNextBufferOrEvent();
+		if (bufferOrEvent.isPresent()) {
```
azagrebin: As an alternative:
`bufferOrEvent.ifPresent(b -> numBytesInCounter.inc(b.getSize()));`

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/InputGateWithMetrics.java
```diff
@@ -0,0 +1,119 @@
...
+			numBytesInCounter.inc(bufferOrEvent.get().getSize());
+		}
+
+		return bufferOrEvent;
+	}
+
+	@Override
+	public Optional<BufferOrEvent> pollNextBufferOrEvent() throws IOException, InterruptedException {
```
azagrebin: It seems we also need to increment here.

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/InputGateWithMetrics.java
```diff
@@ -0,0 +1,119 @@
...
+import java.util.Optional;
+
+import static org.apache.flink.util.Preconditions.checkNotNull;
+
+/**
+ * The InputGateWithMetrics is wrapped in task with {@link InputGate} provided by
+ * shuffle service. It is mainly used for increasing general input metrics from
+ * {@link TaskIOMetricGroup}.
+ */
+public class InputGateWithMetrics implements InputGate, InputGateListener {
```
azagrebin: Not sure why we need to implement `InputGateListener` as well.
Could we not just forward `registerListener` to `inputGate`?
Who is going to call `notifyInputGateNonEmpty` in this class?

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/InputGateWithMetrics.java
```diff
@@ -0,0 +1,119 @@
...
+		inputGate.registerListener(this);
+	}
+
+	@Override
+	public int getPageSize() {
+		return inputGate.getPageSize();
+	}
+
+	@Override
+	public void close() throws Exception {
```
azagrebin: should we not forward `close` to `inputGate` as well?

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/InputGateWithMetrics.java
```diff
@@ -0,0 +1,119 @@
...
+import org.apache.flink.runtime.io.network.partition.consumer.InputGateListener;
+import org.apache.flink.runtime.metrics.groups.TaskIOMetricGroup;
+
+import java.io.IOException;
+import java.util.Optional;
+
+import static org.apache.flink.util.Preconditions.checkNotNull;
+
+/**
+ * The InputGateWithMetrics is wrapped in task with {@link InputGate} provided by
```
azagrebin: `This class wraps {@link InputGate} provided by ...`

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/Task.java
```diff
@@ -637,6 +636,11 @@ else if (current == ExecutionState.CANCELING) {
 
 			TaskKvStateRegistry kvStateRegistry = kvStateService.createKvStateTaskRegistry(jobId, getJobVertexId());
 
+			InputGate[] gates = new InputGate[inputGates.length];
+			for (int i = 0; i < inputGates.length; i++) {
```
azagrebin: To be honest, I liked wrapping in constructor more. This way we keep in `Task` and `RuntimeEnvironment` same objects and there is less risk for them to diverge and forget that we call real gate and not wrapper. What do you think?

I would at least move the wrapping loop to a separate static method.

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/InputGateWithMetrics.java
zhijiangW: In task class we would finally maintain only the array of `InputGateWithMetrics` finally, so the close called from task would focus on `InputGateWithMetrics`. Otherwise the task would maintain two arrays `SingleInputGate` and `InputGateWithMetrics` separately.

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/InputGateWithMetrics.java
zhijiangW: I also not considered this issue before, only found it when existing ITCase fails.

The `UnionInputGate` could only see `InputGateWithMetrics` and then put them into internal `inputGateToIndexOffsetMap`. When the `SingleInputGate` receives data from network, it would notify the listener(`UnionInputGate`) and then the `SingleInputGate` would be added into `UnionInputGate#inputGatesWithData`. When `UnionInputGate#getNextBufferOrEvent`, the `SingleInputGate` is fetched from `inputGatesWithData`, but it can not be found from `inputGateToIndexOffsetMap` which would cause NPE issue.

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/InputGateWithMetrics.java
zhijiangW: If we change to wrap the `InputGateWithMetrics` in the constructor as you mentioned below, we should forward `close` to `InputGate` as well.

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/Task.java
zhijiangW: Yes, I also did it in the constructor in first version. Later I think it might be better to make `InputGateWithMetrics` involved in less logics such as `close`, so I change it as now to keep the previous arrays of `SingleInputGate` in task.

I think it also makes sense to do it in constructor, then I would fix it.

zhijiangW: Forgot another concern for not doing in the constructor. It would make maintain two arrays of `SingleInputGate` and `InputGate` separately. Because we still need to return the array of `SingleInputGate` for `NetworkEnvironment#setup` currently. What do you think?

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/InputGateWithMetrics.java
azagrebin: I see, thanks for explanation!

zhijiangW:
@azagrebin I have rebased the master to solve the conflicts.

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/io/benchmark/StreamNetworkBenchmarkEnvironment.java
```diff
@@ -261,8 +260,7 @@ private InputGate createInputGate(
 					jobId,
 					gateDescriptor,
 					new NoOpTaskActions(),
-					InputChannelTestUtils.newUnregisteredInputChannelMetrics(),
```
azagrebin: We should actually also wrap `gate` here to take wrapping into account in benchmark tests. Probably with a dummy metric counter, at least forcing the wrapper to be called.

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/Task.java
```diff
@@ -382,19 +382,21 @@ public Task(
 			buffersGroup);
 
 		// consumed intermediate result partitions
-		this.inputGates = networkEnvironment.createInputGates(
+		SingleInputGate[] gates = networkEnvironment.createInputGates(
```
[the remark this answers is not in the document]
zhijiangW: Yes, no matter which one is merged first, then the other one has to rebase to solve the conflicts for concerning this issue. :)

zhijiangW:

I pushed one fixup commit for addressing above comments!  And I would submit the micro benchmark comparison later.

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/io/benchmark/StreamNetworkBenchmarkEnvironment.java
```diff
@@ -249,18 +250,21 @@ private InputGate createInputGate(
...
 				environment.getConfiguration(),
 				environment.getConnectionManager(),
 				environment.getResultPartitionManager(),
 				new TaskEventDispatcher(),
-				environment.getNetworkBufferPool())
-				.create(
+				environment.getNetworkBufferPool());
+
+			final InputGate gate = new InputGateWithMetrics(
+				gateFactory.create(
```
azagrebin: nit: maybe have a separate variable for `gateFactory.create`
also channel loop body could be refactored into two methods: `createInputGateDeploymentDescriptor` and `createSingleInputGateWithMetrics` to reduce the current method

zhijiangW: Actually gate factory should not be created inside loop, and it could be created only once together with `NetworkEnvironment`. I would submit a hotfix commit firstly for refactoring this method, and then append my PR based on it.

zhijiangW:
I have squashed the commits for addressing the comments. @azagrebin

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/io/benchmark/StreamNetworkBenchmarkEnvironment.java
```diff
@@ -227,39 +229,22 @@ protected ResultPartitionWriter createResultPartition(
 	}
 
 	private InputGate createInputGate(
-			IntermediateDataSetID dataSetID,
-			ExecutionAttemptID executionAttemptID,
 			final TaskManagerLocation senderLocation,
-			NetworkEnvironment environment,
+			final SingleInputGateFactory gateFactory,
```
azagrebin: is `gateFactory` not already available as a class field in the method?
strange that the field is protected, can be private

zhijiangW: I also found this issue before, but considering all the fields in this class are defined as protected, so I keep the way as others. 
Also for the class field factory, I keep the way as field `channels`  to pass because it is not initialized in the constructor. Actually I also think many other fields in this class could not be passed as well.
Do you think we should further refactoring this?

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/InputGateWithMetrics.java
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	@Override
+	public void requestPartitions() throws IOException, InterruptedException {
+		inputGate.requestPartitions();
+	}
+
+	@Override
+	public Optional<BufferOrEvent> getNextBufferOrEvent() throws IOException, InterruptedException {
+		Optional<BufferOrEvent> bufferOrEvent = inputGate.getNextBufferOrEvent();
```
pnowojski:
```
getNextBufferOrEvent() {
  return updateMetrics(inputGate.getNextBufferOrEvent());
}

pollNextBufferOrEvent() {
  return updateMetrics(inputGate.pollNextBufferOrEvent());
}
```

to deduplicate the code a bit and decrease chances of methods being out of sync (as pointed by @azagrebin in another comment)?

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/consumer/BufferOrEvent.java
```diff
@@ -43,26 +44,32 @@
 
 	private int channelIndex;
 
-	BufferOrEvent(Buffer buffer, int channelIndex, boolean moreAvailable) {
+	private final int size;
+
+	public BufferOrEvent(Buffer buffer, int channelIndex, boolean moreAvailable, int size) {
```
pnowojski: why are we passing this new field explicitly? Why not
```
this.size = buffer.getSize();
```
?

Further more, why do we need `size` field at all? Why not implement:
```
int getSize() {
  return buffer == null ? 0 : buffer.getSize();
}
```

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/consumer/SingleInputGate.java
```diff
@@ -566,9 +560,9 @@ private BufferOrEvent transformToBufferOrEvent(
 			Buffer buffer,
 			boolean moreAvailable,
 			InputChannel currentChannel) throws IOException, InterruptedException {
-		numBytesIn.inc(buffer.getSizeUnsafe());
+		final int size = buffer.getSizeUnsafe();
```
pnowojski: Hmm, it looks like we can drop `getSizeUnsafe()` method altogether, since it doesn't differ from `getSize()` - both are not synchronized (also like `setSize()`). It seems like I or @NicoK missed that when we were implementing low latency changes and dropping the synchronisation in the `Buffer` class. Or am I missing something?

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/io/benchmark/StreamNetworkBenchmarkEnvironment.java
```diff
@@ -270,4 +266,18 @@ private InputGateDeploymentDescriptor createInputGateDeploymentDescriptor(
...
+		InputGateDeploymentDescriptor gateDescriptor,
+		int channelIndex) {
+
+		final SingleInputGate singleGate = gateFactory.create(
+			"receiving task[" + channelIndex + "]",
+			gateDescriptor,
+			SingleInputGateBuilder.NO_OP_PRODUCER_CHECKER,
+			InputChannelTestUtils.newUnregisteredInputChannelMetrics());
+
+		return new InputGateWithMetrics(singleGate, new SimpleCounter());
```
pnowojski: Could we somehow deduplicate the code of wrapping the input gate between production and benchmarking code? So that they follow the same path? For example moving the wrapping for both code paths to the `SingleInputGateFactory`?

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/InputGateWithMetrics.java
zhijiangW: We could deduplicate one common line here, but I am not clear  why decrease changes out of sync. Are there any syncs related to here?

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/consumer/BufferOrEvent.java
zhijiangW: For event buffer, we might not return 0, so we have the explicit size field, but we do not need pass the size for buffer constructor as you suggested.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/consumer/SingleInputGate.java
zhijiangW: Yes, it could be replaced by `getSize` here. I could submit a separate hotfix commit for it?

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/io/benchmark/StreamNetworkBenchmarkEnvironment.java
zhijiangW: Yes, we could make `SingleInputGateFactory` always return wrapped `InputGateWithMetrics`

zhijiangW: There has one concern here. 
`SingleInputGateFactory` is `NetworkEnvironment`/`ShuffleService` specific, not a general factory. `InputGateWithMetrics` is not belonging to `ShuffleService`, and it should be in task scope. So it might be not proper to let `SingleInputGateFactory` wrap for task related components, otherwise it seems no need to introduce `InputGateWithMetrics` in this PR. WDYT?

azagrebin: I agree that `InputGateWithMetrics` does not belong Netty implementation of shuffle service this is why we are doing it change to keep this notification concern outside.

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/InputGateWithMetrics.java
pnowojski: Not that I'm aware of :) I meant for the future, to avoid similar bugs/issues as pointed out by @azagrebin

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/consumer/SingleInputGate.java
pnowojski: Yes, separate hotfix commit would be best (sorry for nit picking here, just noticed this here, since I got curious whether you should be using unsafe or safe version)

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/io/benchmark/StreamNetworkBenchmarkEnvironment.java
pnowojski: Ok, I understand. In that case maybe we could consider some another layer of factory/builder? One that takes a concrete NetworkEnvironment and creates InputGates wrapped with metrics? I'm not sure if that's worth, but you see the code duplication here? And the risks that if those two code parts will drift apart, the benchmarks might become inaccurate.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/buffer/Buffer.java
```diff
@@ -158,17 +158,6 @@
 	 */
 	void setReaderIndex(int readerIndex) throws IndexOutOfBoundsException;
 
-	/**
```
pnowojski: Nit: typo in the commit message (reductant -> redundant)

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/io/benchmark/StreamNetworkBenchmarkEnvironment.java
zhijiangW: I could understand your concern here. 
ATM the `InputGateWithMetrics` is only created in the benchmark which would reflect the real process. Considering this wrapper is very easy to construct and the duplication here is not obvious, just a local `singleGate` definition, I think we could introduce another factory for this wrapped `InputGate` if necessary later?

pnowojski: Ok

PR #8328, opened 2019-05-01
[FLINK-12380] Add thread name in the log4j.properties

Myasuka:

This is inspired by FLINK-12368 when users want to add sub-task index information in the source code. We could add thread name, which already contains sub-task index information, in the logs to avoid have to change the source code.

Moreover, I found existing `logback.xm`l in Flink already contains thread name information. We should also add this in the `log4j.properties`.

## Brief change log

Add thread name in the log4j.properties

## Verifying this change

  - The serializers: **no**

## Documentation

### DIFF
--- a/flink-dist/src/main/flink-bin/conf/log4j-cli.properties
+++ b/flink-dist/src/main/flink-bin/conf/log4j-cli.properties
@@ -23,7 +23,7 @@ log4j.appender.file=org.apache.log4j.FileAppender
 log4j.appender.file.file=${log.file}
 log4j.appender.file.append=false
 log4j.appender.file.layout=org.apache.log4j.PatternLayout
-log4j.appender.file.layout.ConversionPattern=%d{yyyy-MM-dd HH:mm:ss,SSS} %-5p %-60c %x - %m%n
+log4j.appender.file.layout.ConversionPattern=%d{yyyy-MM-dd HH:mm:ss,SSS} [%t] %-5p %-60c %x - %m%n
 
 
 # Log output from org.apache.flink.yarn to the console. This is used by the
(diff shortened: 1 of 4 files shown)


Myasuka:
@link3280 This PR is inspired by FLINK-12368 when users want to add sub-task index information in the source code. Actually, we could add thread name, which already contains sub-task index information, in the logs to avoid have to change the source code. And `logback` in Flink already configured `thread name` in the pattern (you could refer to the [doc of logginng](https://ci.apache.org/projects/flink/flink-docs-stable/monitoring/logging.html#configuring-logback) to confirm).

StephanEwen:
I would like to not change the default log format. We kept it stable for a while now, which seems to have been generally appreciated by users.

In proper production setups, most users customize the log format anyways to their company standards.

Myasuka:
@StephanEwen As you are the founder of Flink and the most-contributed coder, I believe and respect your choice.
However, your reply actually did not convince me. I struggled to hesitate whether to voice my thoughts. Since Apache Flink is an open source software, I believe every word counts.

1. >  We kept it stable for a while now, which seems to have been generally appreciated by users.

I am not sure about other users but we might have the largest Flink cluster in the world but add thread name explicitly in `log4j.properties`.

2. Actually [logback.xml](https://github.com/apache/flink/blob/a7cf24383be9f310fb5ccc5a032721421fa45791/flink-dist/src/main/flink-bin/conf/logback.xml#L24) already contains the thread name.

Last but not least, if your choice is based on "Even though `log4j` and `logback` have different log pattern, but the log system for Flink works well so far, we should not touch this without any strong demand." I think this would be acceptable for me, and hope for no more demands come out to add sub-task index information in logs.

PR #8331, opened 2019-05-02
[FLINK-11987][flink-connector-kafka-0.11] convert NEXT_TRANSACTIONAL_ID_HINT_DESCRIPTOR to a non-static variabl…

Zor-X-L:
…e to avoid StateDescriptor#initializeSerializerUnlessSet throws NullPointerException

To fix a problem in flink-connector-kafka-0.11 which can cause StateDescriptor#initializeSerializerUnlessSet to throw NullPointerException

## Brief change log

*(for example:)*

convert NEXT_TRANSACTONAL_ID_HINT_DESCRIPTOR to a non-static final variable

## Verifying this change

*(or)*

*(or)*

*(example:)*

No.

## Documentation

### DIFF
--- a/flink-connectors/flink-connector-kafka-0.11/src/main/java/org/apache/flink/streaming/connectors/kafka/FlinkKafkaProducer011.java
+++ b/flink-connectors/flink-connector-kafka-0.11/src/main/java/org/apache/flink/streaming/connectors/kafka/FlinkKafkaProducer011.java
@@ -173,7 +173,7 @@ public class FlinkKafkaProducer011<IN>
 	/**
 	 * Descriptor of the transactional IDs list.
 	 */
-	private static final ListStateDescriptor<NextTransactionalIdHint> NEXT_TRANSACTIONAL_ID_HINT_DESCRIPTOR =
+	private final ListStateDescriptor<NextTransactionalIdHint> nextTransactionalIdHintDescriptor =
 		new ListStateDescriptor<>("next-transactional-id-hint", TypeInformation.of(NextTransactionalIdHint.class));
 
 	/**
@@ -815,7 +815,7 @@ public class FlinkKafkaProducer011<IN>
 		}
 
 		nextTransactionalIdHintState = context.getOperatorStateStore().getUnionListState(
-			NEXT_TRANSACTIONAL_ID_HINT_DESCRIPTOR);
+			nextTransactionalIdHintDescriptor);
 		transactionalIdsGenerator = new TransactionalIdsGenerator(
 			getRuntimeContext().getTaskName() + "-" + ((StreamingRuntimeContext) getRuntimeContext()).getOperatorUniqueID(),
 			getRuntimeContext().getIndexOfThisSubtask(),


StephanEwen:
The Pull Request description seems a bit off. Could you add a proper description, as per the common contribution guidelines?

Zor-X-L:
> The Pull Request description seems a bit off. Could you add a proper description, as per the common contribution guidelines?

description edited.

Should I change my PR to fix StateDescriptor instead? Maybe that's indeed a better way.

becketqin:
@Zor-X-L Thanks for the patch. The issue is essentially `StateDescriptor` is mutable and not thread safe. The patch fixes the issue in `FlinkKafkaProducer011`. There might be somewhere else in the code having the same issue.

I agree with @StephanEwen that a better fix would be in `StateDescriptor` itself. There are two possible fixes:

Option 1: 
Ideally,  `StateDescriptor` should be immutable. The `typeInfo` field is only used to create `TypeSerializer`. We can probably do this in the constructor instead of saving the `typeInfo` and creating the serializer lazily. 

However, because creating `TypeSerializer` from `TypeInfo` depends on the execution config which is usually not available when the `StateDescriptor` is constructed, the TypeSerializer have to be created lazily at runtime, unless we make some larger interface change.

Option 2:
Fix `StateDescriptor` to make it thread safe. This is more feasible at this point. There are currently three fields that might be modified after construction. `ttlConfig`, `querableName` and `typeSerializer`. I think we can just synchronize on them to make sure they are only set once. It still has caveats that users may set multiple different values while only the first value will take effect, but it is no worse than the current situation.

@StephanEwen what do you think?

carp84:
Thanks for the analysis @becketqin , and I also agree that we'd better fix the issue from inside `StateDescriptor`.

Regarding the options, besides the already proposed two, there's a 3rd option: making `StateDescriptor.serializer` singleton with CAS plus removing the nulls out of `StateDescriptor.typeInfo` to prevent NPE caused by race condition, as suggested by @StephanEwen in JIRA. IMHO this is a more light-weight and quick fix for the time being. Wdyt? Thanks.

sunjincheng121:
Thanks for your comments @becketqin @carp84!
Hi, @Zor-X-L, Thanks for the PR, and do you want to continue to solve this problem?

becketqin:
> Thanks for the analysis @becketqin , and I also agree that we'd better fix the issue from inside `StateDescriptor`.
> 
> Regarding the options, besides the already proposed two, there's a 3rd option: making `StateDescriptor.serializer` singleton with CAS plus removing the nulls out of `StateDescriptor.typeInfo` to prevent NPE caused by race condition, as suggested by @StephanEwen in JIRA. IMHO this is a more light-weight and quick fix for the time being. Wdyt? Thanks.

@carp84 Yes, CAS + not nullify works. It is essentially one way to synchronize on the fields. On the other hand, given that performance is probably not a concern here, personally I feel adding a `synchronized` block seems less work and more readable.

carp84:
Thanks for the reply @becketqin .

Since this is a blocker issue and it seems @Zor-X-L hasn't replied yet, how about I create another PR for a quick fix @sunjincheng121  @becketqin @StephanEwen ? Thanks.

sunjincheng121:
It sounds pretty good! 
IMO, Feel free to open the PR! :) 
Great thanks!

carp84:
Thanks for the confirmation @sunjincheng121 , have just created [PR#8570](https://github.com/apache/flink/pull/8570) to supersede this one, FYI.

aljoscha:
I'm closing this as "Abandoned", since there is no more activity and the code base has moved on quite a bit. Please re-open this if you feel otherwise and work should continue.

The associated Jira issue is also closed by now.

PR #8332, opened 2019-05-02
[FLINK-12390][build] Fully migrate to flink-shaded-asm-6

zentol:

Migrates the remaining usages of asm5 to version 6.

IIRC, when we introduced asm6 initially we only touched scala-related code.

## Verifying this change

https://travis-ci.org/zentol/flink/builds/516207673


GJL:
@flinkbot approve all

PR #8333, opened 2019-05-02
[FLINK-12391][travis] Limit transfer.sh upload to 1 minute

zentol:
This PR adds a 1 minute timeout to the curl `transfer.sh` upload. We usually spend between 10-20 seconds on the upload; anything going beyond 1 minute usually meant that the upload was broken and would never finish.


StephanEwen:
LGTM

+1

PR #8346, opened 2019-05-06
[FLINK-12405] [DataSet] Introduce BLOCKING_PERSISTENT result partition type

Xpray:
Introducing a new``` ResultPartitionType```, result partition of this type can be consumed for several times and will be used for Interactive Programming.

## Brief change log
Here is the brief changes:

1.  Introduce ```ResultPartitionType``` called ```BLOCKING_PERSISTENT```
2. Introduce ```BlockingShuffleOutputFormat```  which contains a user specified ```IntermediateDataSetID``` (passed from TableAPI in later PR)
3. when ```JobGraphGenerator``` sees a ```GenericDataSinkBase``` with ```BlockingShuffleOutputFormat```, it creates a ```IntermediateDataSet``` with this id, then add it to its predecessor, the ```OutputFormatVertex``` for  this sink node will be excluded in ```JobGraph```
4. So the JobGraph may contains some JobVertex which has more ```IntermediateDataSet``` than its downstream consumers.

Here are some design notes:
* Why modify DataSet/JobGraphGenerator
    Since Blink Planner is not ready yet, and Batch Table is running on Flink Planner(based on DataSet).
    There will be another implementation once Blink Planner is ready.
* Why use a special ```OutputFormat``` as placeholder
   We could add a ```cache()``` method for DataSet, but we do not want to change DataSet API any more. so a special ```OutputFormat``` as placeholder seems reasonable.

## Verifying this change

```org.apache.flink.optimizer.plantranslate.JobGraphGeneratorTest#testGeneratingJobGraphWithUnconsumedResultPartition```

  - The serializers: no

## Documentation

FLIP: 
https://cwiki.apache.org/confluence/display/FLINK/FLIP-36%3A+Support+Interactive+Programming+in+Flink
DOC: 
https://docs.google.com/document/d/1g0VgixNpz2z-v-8AyUIatuRfX-tr6zafOE32Tr0sSjI/edit?ts=5cce964d#heading=h.rz8dxo5goczr

### CODE REVIEW

File: flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/ResultPartitionType.java [hunk 1]
```diff
@@ -49,13 +59,17 @@
 	/** Does this partition use a limited number of (network) buffers? */
 	private final boolean isBounded;
 
+	/** This partition will not be released after consuming if 'isPersistent' is true. */
+	private final boolean isPersistent;
```
zhijiangW: I think we want to introduce a property to indicate this partition is global across jobs. So `isPersistent` seems not reflect this behavior, because `BLOCKING` type is also for persistent file from the semantics.  Maybe called `isGlobal`?

flink-java/src/main/java/org/apache/flink/api/java/io/BlockingShuffleOutputFormat.java
```diff
@@ -0,0 +1,69 @@
...
+		this.intermediateDataSetId = Preconditions.checkNotNull(intermediateDataSetId, "intermediateDataSetId is null");
+	}
+
+	public static <T> BlockingShuffleOutputFormat<T> createOutputFormat(AbstractID intermediateDataSetId) {
+		return new BlockingShuffleOutputFormat<>(intermediateDataSetId);
+	}
+
+	@Override
+	public void configure(Configuration parameters) {
+
```
zhijiangW: remove the empty line

flink-java/src/main/java/org/apache/flink/api/java/io/BlockingShuffleOutputFormat.java
```diff
@@ -0,0 +1,69 @@
...
+import org.apache.flink.util.Preconditions;
+
+import java.io.IOException;
+
+/**
+ * This is the inner OutputFormat used for specify the BLOCKING_PERSISTENT result partition type of coming edge.
+ * @param <T>
+ */
+@Internal
+public final class BlockingShuffleOutputFormat<T> implements OutputFormat<T> {
```
zhijiangW: maybe extends `DiscardingOutputFormat` then we could avoid implement the following empty methods

flink-java/src/main/java/org/apache/flink/api/java/io/BlockingShuffleOutputFormat.java
```diff
@@ -0,0 +1,69 @@
...
+import org.apache.flink.util.AbstractID;
+import org.apache.flink.util.Preconditions;
+
+import java.io.IOException;
+
+/**
+ * This is the inner OutputFormat used for specify the BLOCKING_PERSISTENT result partition type of coming edge.
+ * @param <T>
+ */
+@Internal
```
zhijiangW: I am not sure this should be @Internal. Does it be used for users?

flink-java/src/main/java/org/apache/flink/api/java/io/BlockingShuffleOutputFormat.java
```diff
@@ -0,0 +1,69 @@
...
+public final class BlockingShuffleOutputFormat<T> implements OutputFormat<T> {
+
+	private final AbstractID intermediateDataSetId;
+
+	private BlockingShuffleOutputFormat(AbstractID intermediateDataSetId) {
+		this.intermediateDataSetId = Preconditions.checkNotNull(intermediateDataSetId, "intermediateDataSetId is null");
```
zhijiangW: `Preconditions.checkNotNull` could be ignored here and be done in `createOutputFormat` part.

flink-optimizer/src/main/java/org/apache/flink/optimizer/plantranslate/JobGraphGenerator.java
```diff
@@ -480,6 +482,32 @@ public void postVisit(PlanNode node) {
...
 				return;
 			}
+
+			// if this is a blocking shuffle vertex, we add one IntermediateDataSetID to its predecessor and return
+			if (node instanceof SinkPlanNode) {
+				Object userCodeObject = node.getProgramOperator().getUserCodeWrapper().getUserCodeObject();
+				if (userCodeObject instanceof BlockingShuffleOutputFormat) {
+					PlanNode precedentNode = node.getInputs().iterator().next().getSource();
+					JobVertex precedentVertex;
+					if (this.vertices.containsKey(precedentNode)) {
```
zhijiangW: I think we should not use `this` in the detail process in the new format except in the constructor.

flink-optimizer/src/main/java/org/apache/flink/optimizer/plantranslate/JobGraphGenerator.java
```diff
@@ -480,6 +482,32 @@ public void postVisit(PlanNode node) {
 			if (node instanceof SourcePlanNode || node instanceof NAryUnionPlanNode || node instanceof SolutionSetPlanNode) {
 				return;
 			}
+
+			// if this is a blocking shuffle vertex, we add one IntermediateDataSetID to its predecessor and return
+			if (node instanceof SinkPlanNode) {
+				Object userCodeObject = node.getProgramOperator().getUserCodeWrapper().getUserCodeObject();
+				if (userCodeObject instanceof BlockingShuffleOutputFormat) {
+					PlanNode precedentNode = node.getInputs().iterator().next().getSource();
```
zhijiangW: Do we need to check `node.getInputs().iterator().hasNext()` before `getSource()`?

flink-optimizer/src/test/java/org/apache/flink/optimizer/plantranslate/JobGraphGeneratorTest.java
```diff
@@ -265,6 +274,55 @@ public void testArtifactCompression() throws IOException {
...
+
+		ExecutionEnvironment env = ExecutionEnvironment.getExecutionEnvironment();
+
+		DataSet<Tuple2<Long, Long>> input = env.fromElements(new Tuple2<>(1L, 2L))
+			.setParallelism(1);
+
+		DataSet ds = input.map((MapFunction<Tuple2<Long, Long>, Object>) value -> new Tuple2<>(value.f0 + 1, value.f1))
+			.setParallelism(3);
+
+		AbstractID intermediateDataSetID = new AbstractID();
```
zhijiangW: intermediateDataSetID -> abstractID?

flink-java/src/main/java/org/apache/flink/api/java/io/BlockingShuffleOutputFormat.java
Xpray: This class should not be used by users.

flink-java/src/main/java/org/apache/flink/api/java/io/BlockingShuffleOutputFormat.java
Xpray: DiscardingOutputFormat should really acts like a real sink instead of a placeholder, that's way we should introduce a new OutputFormat.

flink-java/src/main/java/org/apache/flink/api/java/io/BlockingShuffleOutputFormat.java
Xpray: Good advice, I'll take care of it.

flink-optimizer/src/main/java/org/apache/flink/optimizer/plantranslate/JobGraphGenerator.java
Xpray: Since it's a SinkPlanNode, it contains one and only one input, I can add some sanity check here.

flink-optimizer/src/test/java/org/apache/flink/optimizer/plantranslate/JobGraphGeneratorTest.java
```diff
@@ -265,6 +274,55 @@ public void testArtifactCompression() throws IOException {
...
+		Plan plan = env.createProgramPlan();
+		Optimizer pc = new Optimizer(new Configuration());
+		OptimizedPlan op = pc.compile(plan);
+
+		JobGraphGenerator jgg = new JobGraphGenerator();
+		JobGraph jobGraph = jgg.compileJobGraph(op);
+
+		Assert.assertEquals(3, jobGraph.getVerticesSortedTopologicallyFromSources().size());
+
+		JobVertex inputVertex = jobGraph.getVerticesSortedTopologicallyFromSources().get(0);
```
zhijiangW: the `inputVertex` seems not be used in following parts.

flink-optimizer/src/test/java/org/apache/flink/optimizer/plantranslate/JobGraphGeneratorTest.java
```diff
@@ -265,6 +274,55 @@ public void testArtifactCompression() throws IOException {
...
+		JobVertex mapVertex = jobGraph.getVerticesSortedTopologicallyFromSources().get(1);
+		JobVertex outputVertex = jobGraph.getVerticesSortedTopologicallyFromSources().get(2);
+
+		Assert.assertTrue(outputVertex instanceof OutputFormatVertex);
```
zhijiangW: It is better to use `hamcrest` to verify `instanceof` here.

flink-optimizer/src/main/java/org/apache/flink/optimizer/plantranslate/JobGraphGenerator.java
Xpray: I found ``this`` shows up all over the ``JobGraphGenerator``, maybe a refactor needed later.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/ResultPartitionType.java
Xpray: I think `BLOCKING ` is opposite to `PIPELINE` only describes the flow pattern of data.
But persistence describes the life cycle of data.
What do you think?

flink-optimizer/src/test/java/org/apache/flink/optimizer/plantranslate/JobGraphGeneratorTest.java
Xpray: > It is better to use `hamcrest` to verify `instanceof` here.

Great, it seemes `hamcrest ` is more readable here.

Xpray:
@zhijiangW Thanks for you review and advice，I've updated the PR.

flink-optimizer/src/main/java/org/apache/flink/optimizer/plantranslate/JobGraphGenerator.java
zhijiangW: Yes, we have some history inconsistent issues. But it should not bring that in new codes. For the history codes we might solve in a separate hotfix or just keep that.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/ResultPartitionType.java
zhijiangW: We want to define a property to give a hint for partition release. But the term of `isPersistent` seems not very proper for this property, and it looks like to describe whether the data is saved in memory or persistent file. So from this point, `BLOCKING` also saves the data in persistent file. I think `isGlobal` seems more suitable for the hint of partition release.

Xpray: How about changing `BLOCKING` to `BLOCKING_INTERNAL` and `BLOCKING_PERSISTENT` to `BLOCKING_EXTERNAL`

Xpray:
@GJL  please have a look at this.

GJL:
@Xpray Unfortunately, I am not familiar with the topic of interactive programming, and I currently do not have capacities to look into it.

StephanEwen:
Looks good, with some minor suggested changes, like moving some code into helper methods.

I would go ahead and merge this and apply these changes myself. That is not the usual way to go, but I understand this is blocking a lot of follow-up work, so I want to help speed this up...

flink-java/src/main/java/org/apache/flink/api/java/io/BlockingShuffleOutputFormat.java
StephanEwen: I agree, this should be a separate one which actually throws `UnsupportedOperationException` in all methods. That way we ensure that this never accidentally gets used during runtime, but enforce to only have it as a compile time placeholder.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/ResultPartitionType.java
StephanEwen: I agree that BLOCKING_PERSISTENT is not the best name.
The specific thing is that it lives across jobs, but I am not sure what a good name for that would be.
"global" and "external" both go in that direction...

Xpray: It seems "external" might be better than "global" since "internal" is better than "local"?

PR #8349, opened 2019-05-06
[FLINK-12411][table-planner][tests] Workaround limited support of not nullable fields in window aggregation

dawidwys:

This does not fix the problem. It just makes the e2e test pass. For a
proper fix of the underlying problem see [FLINK-12249](https://issues.apache.org/jira/browse/FLINK-12249).

## Brief change log

  - Produce nullable type in `CASE ... WHEN` statement. The added `NULL` value will never be generated. It's sole purpose is to deceive the planner.

## Verifying this change

It changes only `test_streaming_sql.sh`.

  - The runtime per-record code paths (performance sensitive): (yes / **no /** don't know)

## Documentation

### DIFF
--- a/flink-end-to-end-tests/flink-stream-sql-test/src/main/java/org/apache/flink/sql/tests/StreamSQLTestProgram.java
+++ b/flink-end-to-end-tests/flink-stream-sql-test/src/main/java/org/apache/flink/sql/tests/StreamSQLTestProgram.java
@@ -106,7 +106,9 @@ public class StreamSQLTestProgram {
 		String tumbleQuery = String.format(
 			"SELECT " +
 			"  key, " +
-			"  CASE SUM(cnt) / COUNT(*) WHEN 101 THEN 1 ELSE 99 END AS correct, " +
+			//TODO: The "WHEN -1 THEN NULL" part is a temporary workaround, to make the test pass, for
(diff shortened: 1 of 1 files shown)


dawidwys:
@twalthr This is a temporary fix for the failling `test_streaming_sql.sh`. WDYT? should we add that until we fix it properly in [FLINK-12249](https://issues.apache.org/jira/browse/FLINK-12249)

flink-end-to-end-tests/flink-stream-sql-test/src/main/java/org/apache/flink/sql/tests/StreamSQLTestProgram.java
```diff
@@ -106,7 +106,7 @@ public static void main(String[] args) throws Exception {
 		String tumbleQuery = String.format(
 			"SELECT " +
 			"  key, " +
-			"  CASE SUM(cnt) / COUNT(*) WHEN 101 THEN 1 ELSE 99 END AS correct, " +
+			"  CASE SUM(cnt) / COUNT(*) WHEN 101 THEN 1 WHEN -1 THEN NULL ELSE 99 END AS correct, " +
```
twalthr: Can you add a comment and link to the related issue?

PR #8339, opened 2019-05-04
[FLINK-12240][hive] Support view related operations in GenericHiveMetastoreCatalog

bowenli86:
## What is the purpose of the change

This PR added support for views in GenericHiveMetastoreCatalog, operations include create, drop, alter, rename, list.

## Brief change log

- Implemented view related APIs in GenericHiveMetastoreCatalog
- Reused and moved view related unit tests from GenericInMemoryCatalog to CatalogTestBase

## Verifying this change

This change added tests and can be verified as follows:

  - Tests moved to CatalogTestBase

## Does this pull request potentially affect one of the following parts:

  - Dependencies (does it add or upgrade a dependency): (no)
  - The public API, i.e., is any changed class annotated with `@Public(Evolving)`: (no)
  - The serializers: (no)
  - The runtime per-record code paths (performance sensitive): (no)
  - Anything that affects deployment or recovery: JobManager (and its components), Checkpointing, Yarn/Mesos, ZooKeeper: (no)
  - The S3 file system connector: (no)

## Documentation

  - Does this pull request introduce a new feature? (yes)
  - If yes, how is the feature documented? (JavaDocs)

PR #8353, opened 2019-05-06
[FLINK-12233][hive] Support table related operations in HiveCatalog

bowenli86:

This PR introduced HiveCatalogTable and implemented table related catalog APIs in HiveCatalog.

## Brief change log

- Created a basic `HiveCatalogTable`, and we can enrich this class later
- Implemented table related catalog APIs in `HiveCatalog`
- Extracted common logic between `HiveCatalog` and `GenericHiveMetastoreCatalog` into `HiveCatalogBase`
- Moved a few util classes from package `org.apache.flink.table.catalog.hive` to `org.apache.flink.table.catalog.hive.util`

## Verifying this change

Reuse tests in CatalogTestBase.

  - The serializers: (no)

## Documentation

This PR depends on https://github.com/apache/flink/pull/8339


bowenli86:
cc @xuefuz @KurtYoung @JingsongLi

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalog.java
```diff
@@ -85,49 +86,39 @@ public void alterDatabase(String name, CatalogDatabase newDatabase, boolean igno
...
+			HiveCatalogUtil.createHiveTable(tablePath, table),
+			ignoreIfExists);
 	}
 
 	@Override
-	public List<String> listViews(String databaseName) throws DatabaseNotExistException, CatalogException {
-		throw new UnsupportedOperationException();
+	public void alterTable(ObjectPath tablePath, CatalogBaseTable newTable, boolean ignoreIfNotExists)
+			throws TableNotExistException, CatalogException {
+		validateHiveCatalogTable(newTable);
```
xuefuz: Wouldn't we want to do similar validation for GenericHiveMetastoreCatalog?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
```diff
@@ -197,4 +204,140 @@ protected void alterHiveDatabase(String name, Database newHiveDatabase, boolean
...
+	public void alterTable(ObjectPath tablePath, CatalogBaseTable newTable, boolean ignoreIfNotExists)
+			throws TableNotExistException, CatalogException {
+		if (!tableExists(tablePath)) {
+			if (!ignoreIfNotExists) {
+				throw new TableNotExistException(catalogName, tablePath);
+			}
+		} else {
+			// IMetastoreClient.alter_table() requires the table to have a valid location, which it doesn't in this case
+			// Thus we have to translate alterTable() into (dropTable() + createTable())
+			dropTable(tablePath, false);
```
xuefuz: I think we needs to be very careful here. dropTable() might also drop the data, but alterTable() really is supposed to change the metadata only.

We should probably solve the location problem rather than taking this workaround.

xuefuz: We also need to note that view and table are sharing this call, so it probably doesn't make sense to change a view to a regular table and vise versa.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogTable.java
```diff
@@ -0,0 +1,111 @@
...
+
+	@Override
+	public String getComment() {
+		return comment;
+	}
+
+	@Override
+	public CatalogBaseTable copy() {
+		return new HiveCatalogTable(
+			tableSchema, new ArrayList<>(partitionKeys), new HashMap<>(properties), comment);
```
xuefuz: Don't we need to copy tableSchema?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveTableConfig.java
```diff
@@ -19,11 +19,15 @@
 package org.apache.flink.table.catalog.hive;
 
 /**
- * Configs for Flink tables stored in Hive metastore.
+ * Configs for tables in Hive metastore.
  */
 public class HiveTableConfig {
 
-	// Description of the Flink table
+	// -------------------
```
xuefuz: Do the three lines (26-28) provide any additional info above the class name here?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/util/HiveCatalogUtil.java
```diff
@@ -0,0 +1,148 @@
...
+
+/**
+ * Utils to convert meta objects between Flink and Hive for HiveCatalog.
+ */
+public class HiveCatalogUtil extends HiveCatalogBaseUtil {
+
+	private HiveCatalogUtil() {
+	}
+
+	// ------ Utils ------
```
xuefuz: Nit: The class name is called "utils", and is this line still necessary?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalog.java
```diff
@@ -85,49 +86,39 @@ public void alterDatabase(String name, CatalogDatabase newDatabase, boolean igno
...
-	}
-
-	@Override
-	public void renameTable(ObjectPath tablePath, String newTableName, boolean ignoreIfNotExists)
-			throws TableNotExistException, TableAlreadyExistException, CatalogException {
-		throw new UnsupportedOperationException();
-	}
-
 	@Override
 	public void createTable(ObjectPath tablePath, CatalogBaseTable table, boolean ignoreIfExists)
```
xuefuz: I think we can keep the implementation only in HiveCatalogBase and define a new API called createHiveTable(). Each subcalss only needs to implement createHiveTable() (because they may create in different ways). I don't see why we need HiveCatalogUtil and related classes for this.

This might be cleaner.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalog.java
xuefuz: Similar to createTable(), I don't see we need to provide implementations for alterTable() in each subclass because I don't really see much difference. If validation is different, then we can define a new interface in the base and have each subclass implement it.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalog.java
```diff
@@ -85,49 +86,39 @@ public void alterDatabase(String name, CatalogDatabase newDatabase, boolean igno
...
+			throws TableNotExistException, CatalogException {
+		validateHiveCatalogTable(newTable);
+
+		super.alterTable(tablePath, newTable, ignoreIfNotExists);
 	}
 
 	@Override
-	public CatalogBaseTable getTable(ObjectPath objectPath) throws TableNotExistException, CatalogException {
-		throw new UnsupportedOperationException();
+	public CatalogBaseTable getTable(ObjectPath tablePath)
```
xuefuz: Similar comment here. I don't see we need getTable() here. getTable() is the same for both catalogs, and the only diff is the translation, which can be captured in a new API.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
bowenli86: good catch, fixed

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveTableConfig.java
bowenli86: removed

bowenli86:
Thanks @xuefuz . The reason I use util methods is because as the logic complexity grows, the size of two catalog classes can be quite big. I'm fine to remove util methods and keep shared logic in HiveCatalogBase and special logic of each catalog in its own class now, and we can reevaluate whenever necessary.

Looking at each commit would probably be easier. To summarize the changes:

- moved shared logic to HiveCatalogBase
- added a few interfaces in HiveCatalogBase for subclasses to implement
- updated test util and a few tests to cater to above changes

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/GenericHiveMetastoreCatalog.java
```diff
@@ -42,23 +46,35 @@
...
  * A catalog that persists all Flink streaming and batch metadata by using Hive metastore as a persistent storage.
  */
 public class GenericHiveMetastoreCatalog extends HiveCatalogBase {
 	private static final Logger LOG = LoggerFactory.getLogger(GenericHiveMetastoreCatalog.class);
 
+	// Prefix used to distinguish properties created by Hive and Flink,
+	// as Hive metastore has its own properties created upon table creation and migration between different versions of metastore.
+	private static final String FLINK_PROPERTY_PREFIX = "flink.";
+
+	// Flink tables should be stored as 'external' tables in Hive metastore
```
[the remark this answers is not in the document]
bowenli86: You are right, I verified it locally. I think using Hive managed tables would be fine here.

bowenli86:
@xuefuz @lirui-apache thanks for your review. please take another look.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/GenericHiveMetastoreCatalog.java
```diff
@@ -95,144 +106,103 @@ public void alterDatabase(String name, CatalogDatabase newDatabase, boolean igno
...
-		} catch (NoSuchObjectException e) {
-			if (!ignoreIfNotExists) {
-				throw new TableNotExistException(catalogName, tablePath);
-			}
-		} catch (TException e) {
-			throw new CatalogException(
-				String.format("Failed to drop table %s", tablePath.getFullName()), e);
+	public void validateCatalogBaseTable(CatalogBaseTable table) throws IllegalArgumentException {
+		// TODO: invalidate HiveCatalogView
+		if (table instanceof HiveCatalogTable) {
```
xuefuz: I'm wondering if we should specify what we are expecting and throw exceptions if condition isn't met.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/GenericHiveMetastoreCatalog.java
```diff
@@ -95,144 +106,103 @@ public void alterDatabase(String name, CatalogDatabase newDatabase, boolean igno
...
-		Table hiveTable = getHiveTable(tablePath);
+		Table hiveTable = new Table();
+		hiveTable.setDbName(tablePath.getDatabaseName());
+		hiveTable.setTableName(tablePath.getObjectName());
+		hiveTable.setCreateTime((int) (System.currentTimeMillis() / 1000));
 
-		return GenericHiveMetastoreCatalogUtil.createCatalogTable(hiveTable);
-	}
+		// Table properties
+		hiveTable.setParameters(buildFlinkProperties(properties));
```
xuefuz: Nit: remove the leading spaces.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/GenericHiveMetastoreCatalog.java
```diff
@@ -363,4 +333,23 @@ public CatalogColumnStatistics getPartitionColumnStatistics(ObjectPath tablePath
...
+	private Map<String, String> retrieveFlinkProperties(Map<String, String> hiveTableParams) {
+		return hiveTableParams.entrySet().stream()
+			.filter(e -> e.getKey().startsWith(FLINK_PROPERTY_PREFIX))
+			.collect(Collectors.toMap(e -> e.getKey().replace(FLINK_PROPERTY_PREFIX, ""), e -> e.getValue()));
+	}
+
+	/**
+	 * Add a prefix to Flink-created properties to distinguish them from Hive-created properties.
+	 */
+	private Map<String, String> buildFlinkProperties(Map<String, String> properties) {
```
xuefuz: Maybe we can use words "mask" and "unmask" instead of "build" and "retrieve".

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
```diff
@@ -89,6 +101,34 @@ private static IMetaStoreClient getMetastoreClient(HiveConf hiveConf) {
...
 
+	// ------ APIs ------
+
+	/**
+	 * Validate input base table.
+	 *
+	 * @param catalogBaseTable the base table to be validated
+	 * @throws IllegalArgumentException thrown if the input base table is invalid.
+	 */
+	public abstract void validateCatalogBaseTable(CatalogBaseTable catalogBaseTable)
```
xuefuz: Use "protected" if we don't expect external callers.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
```diff
@@ -89,6 +101,34 @@ private static IMetaStoreClient getMetastoreClient(HiveConf hiveConf) {
...
+		throws IllegalArgumentException;
+
+	/**
+	 * Create a CatalogBaseTable from a Hive table.
+	 *
+	 * @param hiveTable a Hive table
+	 * @return a CatalogBaseTable
+	 */
+	public abstract CatalogBaseTable createCatalogTable(Table hiveTable);
```
xuefuz: Can we use word "convertTo" instead of "create"?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
```diff
@@ -89,6 +101,34 @@ private static IMetaStoreClient getMetastoreClient(HiveConf hiveConf) {
...
+
+	/**
+	 * Create a Hive table from a CatalogBaseTable.
+	 *
+	 * @param tablePath path of the table
+	 * @param table a CatalogBaseTable
+	 * @return a Hive table
+	 */
+	public abstract Table createHiveTable(ObjectPath tablePath, CatalogBaseTable table);
```
xuefuz: Same as above.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
```diff
@@ -197,4 +237,193 @@ protected void alterHiveDatabase(String name, Database newHiveDatabase, boolean
...
+		} catch (TException e) {
+			throw new CatalogException(
+				String.format("Failed to rename table %s", tablePath.getFullName()), e);
+		}
+	}
+
+	@Override
+	public void alterTable(ObjectPath tablePath, CatalogBaseTable newCatalogTable, boolean ignoreIfNotExists)
+			throws TableNotExistException, CatalogException {
+		try {
```
xuefuz: Do we need to validate newCatalogTable?

xuefuz: Also, how do we ensure no alter for table to view and vise versa?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
```diff
@@ -197,4 +237,193 @@ protected void alterHiveDatabase(String name, Database newHiveDatabase, boolean
...
+		} catch (TException e) {
+			throw new CatalogException(
+				String.format("Failed to get table %s from Hive metastore", tablePath.getFullName()), e);
+		}
+	}
+
+	/**
+	 * Create a Flink's TableSchema from Hive table's columns and partition keys.
+	 */
+	protected TableSchema createTableSchema(List<FieldSchema> cols, List<FieldSchema> partitionKeys) {
```
xuefuz: This also seems to be conversions. Can we also use word "convertTo" instead of "create"? Same for the next method below.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/GenericHiveMetastoreCatalog.java
bowenli86: seems to be a github presentation issue, the code doesn't have spaces

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/GenericHiveMetastoreCatalog.java
bowenli86: I feel "mask" is a good name. Not sure about "unmask" because the retrieval not only unmasks flink properties but also filters out hive properties.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
bowenli86: I feel "create" is better because it has more complexity than just simply convert one object to another and keep naming convention consistent. Also feels a bit weird to "convert" two things into one.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
bowenli86: In memory catalog also doesn't check that currently. We should solve that in "FLINK-12452: alterTable() in all catalogs should ensure existing base table and the new one are of the same type"

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
bowenli86: ditto

bowenli86:
> Thanks for refactoring the code. It looks much cleaner now.
> 
> However, I still noticed there are still a lot of commonality between the two implementations of createHiveTable() and createCatalogTable(). Maybe we can share more code between them and reduce the code duplication.

@xuefuz they currently look similar mainly because HiveCatalog's implementation is only preliminary at the moment as mentioned in comment. As we enrich HiveCatalog, I expect they will diverge a lot. It would be better to try to unify their common parts in the end.

xuefuz:
ok. sounds good. thanks.

bowenli86:
@xuefuz @lirui-apache thanks for your review! The build failure is unrelated (python api issue). If there's no more comment, I will merge tonight

bowenli86:
Rebased to latest master to resolve conflicts. Will wait for a green build before proceeding

xuefuz:
> @xuefuz @lirui-apache thanks for your review! The build failure is unrelated (python api issue). If there's no more comment, I will merge tonight
I think I will go thru one more time on this. Thanks.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/GenericHiveMetastoreCatalog.java
```diff
@@ -363,4 +332,23 @@ public CatalogColumnStatistics getPartitionColumnStatistics(ObjectPath tablePath
 		throw new UnsupportedOperationException();
 	}
 
+	// ------ utils ------
+
+	/**
+	 * Filter out Hive-created properties, and return Flink-created properties.
+	 */
+	private Map<String, String> retrieveFlinkProperties(Map<String, String> hiveTableParams) {
```
xuefuz: static?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/GenericHiveMetastoreCatalog.java
```diff
@@ -363,4 +332,23 @@ public CatalogColumnStatistics getPartitionColumnStatistics(ObjectPath tablePath
...
+		return hiveTableParams.entrySet().stream()
+			.filter(e -> e.getKey().startsWith(FLINK_PROPERTY_PREFIX))
+			.collect(Collectors.toMap(e -> e.getKey().replace(FLINK_PROPERTY_PREFIX, ""), e -> e.getValue()));
+	}
+
+	/**
+	 * Add a prefix to Flink-created properties to distinguish them from Hive-created properties.
+	 */
+	private Map<String, String> maskFlinkProperties(Map<String, String> properties) {
```
xuefuz: Static?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
```diff
@@ -89,6 +101,34 @@ private static IMetaStoreClient getMetastoreClient(HiveConf hiveConf) {
...
 
+	// ------ APIs ------
+
+	/**
+	 * Validate input base table.
+	 *
+	 * @param catalogBaseTable the base table to be validated
+	 * @throws IllegalArgumentException thrown if the input base table is invalid.
+	 */
+	protected abstract void validateCatalogBaseTable(CatalogBaseTable catalogBaseTable)
```
xuefuz: Even though this is not a public API, I feel it should be more formal. Specifically, we should probably can let this return true or false for validation result, rather than relying on throwing IllegalArgumentException. Secondly, validation can be more more complicated than just check the instance type. At this level, it shouldn't care what kind of validation the sub classes might do.

I think it's acceptable if the base class doesn't care if the sub class does validation or how it does it. So it's okay if we remove this and let each subclass do its validation when create/alter table. This way might be even cleaner.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
```diff
@@ -197,4 +237,196 @@ protected void alterHiveDatabase(String name, Database newHiveDatabase, boolean
...
+			colTypes[i] = HiveTypeUtil.toFlinkType(TypeInfoUtils.getTypeInfoFromTypeString(fs.getType()));
+		}
+
+		return new TableSchema(colNames, colTypes);
+	}
+
+	/**
+	 * Create Hive columns from Flink TableSchema.
+	 */
+	protected List<FieldSchema> createHiveColumns(TableSchema schema) {
```
xuefuz: static?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
xuefuz: static?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
bowenli86: This would go against the previous suggestion you made:

> Similar to createTable(), I don't see we need to provide implementations for alterTable() in each subclass because I don't really see much difference. If validation is different, then we can define a new interface in the base and have each subclass implement it.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/GenericHiveMetastoreCatalog.java
bowenli86: What's the key point of making it static given no other class uses it?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
bowenli86: I think this is fine as it's used and only used by its two subclasses, not any other classes. Later on we may be able to abstract column and partition creation from subclasses to base class, and this would save us from debating whether it's should be base class or a util class back and forth.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
xuefuz: I'm not backtracking what I said earlier. It's fine to defined an API, but the API needs to be well-thought, which doesn't seem to be the case here. That's why I said we can forgo the API. If we think an API is the way to go, then let's define the API better and parent class shouldn't define what validation a subclass might do. Personally, I don't feel the API is sufficiently sound, but feel free to get 2nd opinion though.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/GenericHiveMetastoreCatalog.java
xuefuz: The difference is that a static method doesn't require an instance of the class in order to invoke it. For example, if another static method of this class needs to call this method, without this one defined as static, one has to create an instance of the class to invoke the method. The general principle is that if a method doesn't access any of the object variables/methods, make it static.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
xuefuz: It probably doesn't matter much if one develops both parent class and child classes. Image the base class is defined by one different person, who probably doesn't have much idea what the child class needs. Putting the two utility methods here suggests the parent class needs to envision what's needed by the child classes, which doesn't meet the criteria of separation of concerns. I think moving the two methods to static and out of the parent class would make the code cleaner.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/GenericHiveMetastoreCatalog.java
bowenli86: make sense to me. Fixed

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
bowenli86: Make sense. Though IllegalArgumentException can be thrown when any part of the input table is invalid, a better signature may just be throwing CatalogException to conform with that of other existing catalog APIs, since we've defined CatalogException's use case as "in case of any runtime exception".

I don't think returning boolean is a good signature as the check can be fairly complicated and the caller won't know which part is invalid. Returning boolean would be a good option for a single criteria check like "bool isTable()".

bowenli86:

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
xuefuz: Yeah. While inmemory is currently doing that, it's achievable, which might not be the case for Hive. This is no consistency required among external catalogs. Nevertheless, I think we should if check if Hive allows it. It's fine to do this in a separate JIRA.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
xuefuz: If you're concerned about giving user precise, useful information, throwing an exception isn't the best choice. For that, logging is probably better. Neverthelesss, CatalogException certainly sounds better, though it doesn't change the nature of the problem. Let's do this way for now.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
```diff
@@ -89,6 +96,34 @@ private static IMetaStoreClient getMetastoreClient(HiveConf hiveConf) {
...
+	protected abstract void validateCatalogBaseTable(CatalogBaseTable catalogBaseTable)
+		throws CatalogException;
+
+	/**
+	 * Create a CatalogBaseTable from a Hive table.
+	 *
+	 * @param hiveTable a Hive table
+	 * @return a CatalogBaseTable
+	 */
+	protected abstract CatalogBaseTable createCatalogBaseTable(Table hiveTable);
```
xuefuz: I think we can just name the method createCatalogTable() which sounds more general, and as a direct counterpart of createHiveTable(). However, this is very minor though.

bowenli86: This method will return either a CatalogTable or CatalogView, that's why it's named "createCatalogBaseTable() " and "createCatalogTable()" may be a bit misleading

bowenli86:
merging...

PR #8357, opened 2019-05-07
[FLINK-12175] Change filling of typeHierarchy in analyzePojo, for cor…

andbul:

This pull request fixes bug with creating TypeInfo for POJO Classes with generic superclasses which has fields of that genric type. The problem is when we building typeHierarchy for Pojo with generic superclass we already have values in typeHierarchy array, we stopped building typehierahy and coud not extract type for fields later.

## Brief change log

  - Removed condition on if typeHierarchy <= 1. Changed to simple else statement cause we shoudn't stop building type hierarchy.

## Verifying this change

  - Added *PojoParametrizedTypeExtractionTest* test that validates that TypeInfo for generic field correctlly exctracted by TypeExtractor

  - The serializers: yes 
  - The S3 file system connector:  no 

## Documentation

### DIFF
--- a/flink-core/src/main/java/org/apache/flink/api/java/typeutils/TypeExtractor.java
+++ b/flink-core/src/main/java/org/apache/flink/api/java/typeutils/TypeExtractor.java
@@ -1837,8 +1837,8 @@ public class TypeExtractor {
 		if (parameterizedType != null) {
 			getTypeHierarchy(typeHierarchy, parameterizedType, Object.class);
 		}
-		// create a type hierarchy, if the incoming only contains the most bottom one or none.
-		else if (typeHierarchy.size() <= 1) {
+		// create a type hierarchy, for fields types extraction
+		else {
 			getTypeHierarchy(typeHierarchy, clazz, Object.class);
 		}
 
--- a/flink-core/src/test/java/org/apache/flink/api/java/typeutils/PojoParametrizedTypeExtractionTest.java
+++ b/flink-core/src/test/java/org/apache/flink/api/java/typeutils/PojoParametrizedTypeExtractionTest.java
@@ -1,0 +1,101 @@
+package org.apache.flink.api.java.typeutils;
+
+import org.apache.flink.api.common.functions.MapFunction;
+import org.apache.flink.api.common.typeinfo.TypeInformation;
+import org.junit.Assert;
+import org.junit.Test;
+
+import java.io.Serializable;
+
+/**
+ *  Tests concerning type extraction of Parametrized Pojo and its superclasses.
+ */
+public class PojoParametrizedTypeExtractionTest {
+
+	@Test
+	public void testCreateTypeInfo(){
+		/// Init
+		final TypeInformation<ParametrizedParentImpl> directTypeInfo = TypeExtractor.createTypeInfo(ParametrizedParentImpl.class);
+
+		/// Check
+		Assert.assertTrue(directTypeInfo instanceof PojoTypeInfo);
+	}
+
+	@Test
+	public void testFieldExtraction(){
+		/// Init
+		final TypeInformation<ParametrizedParentImpl> directTypeInfo = TypeExtractor.createTypeInfo(ParametrizedParentImpl.class);
+
+		// Extract
+		TypeInformation<?> directPojoFieldTypeInfo = ((PojoTypeInfo) directTypeInfo)
+			.getPojoFieldAt(0)
+			.getTypeInformation();
+
+		// Check
+		Assert.assertTrue(directPojoFieldTypeInfo instanceof PojoTypeInfo);
+	}
+
+	@Test
+	public void testMapReturnTypeInfo(){
+		/// Init
+		final TypeInformation<ParametrizedParentImpl> directTypeInfo = TypeExtractor.createTypeInfo(ParametrizedParentImpl.class);
+
+		// Extract
+		TypeInformation<ParametrizedParentImpl> mapReturnTypeInfo = TypeExtractor
+			.getMapReturnTypes(new ConcreteMapFunction(), directTypeInfo);
+
+		// Check
+		Assert.assertTrue(mapReturnTypeInfo instanceof PojoTypeInfo);
+		Assert.assertEquals(directTypeInfo, mapReturnTypeInfo);
+	}
+
+	@Test
+	public void testMapReturnFieldTypeInfo(){
+		/// Init
+		final TypeInformation<ParametrizedParentImpl> directTypeInfo = TypeExtractor.createTypeInfo(ParametrizedParentImpl.class);
+
+		// Extract
+		TypeInformation<ParametrizedParentImpl> mapReturnTypeInfo = TypeExtractor
+			.getMapReturnTypes(new ConcreteMapFunction(), directTypeInfo);
+		TypeInformation<?> mapReturnPojoFieldTypeInfo = ((PojoTypeInfo) mapReturnTypeInfo)
+			.getPojoFieldAt(0)
+			.getTypeInformation();
+
+		// Check
+		Assert.assertTrue(mapReturnPojoFieldTypeInfo instanceof PojoTypeInfo);
+
+	}
+
+	/**
+	 * Representation of Pojo class with 2 fields.
+	 */
+	public static class Pojo implements Serializable {
+		public int digits;
+		public String letters;
+	}
+
+	/**
+	 * Representation of class which is parametrized by some pojo.
+	 */
+	public static class ParameterizedParent<T extends Serializable> implements Serializable {
+		public T pojoField;
+	}
+
+	/**
+	 * Implementation of ParametrizedParent parametrized by Pojo.
+	 */
+	public static class ParametrizedParentImpl extends ParameterizedParent<Pojo> {
+		public double precise;
+	}
+
+	/**
+	 * Representation of map function for type extraction.
+	 */
+	public static class ConcreteMapFunction implements MapFunction<ParametrizedParentImpl, ParametrizedParentImpl> {
+		@Override
+		public ParametrizedParentImpl map(ParametrizedParentImpl value) throws Exception {
+			return null;
+		}
+	}
+
+}


flink-core/src/test/java/org/apache/flink/api/java/typeutils/PojoParametrizedTypeExtractionTest.java
```diff
@@ -0,0 +1,101 @@
+package org.apache.flink.api.java.typeutils;
```
gaoyunhaii, line 1: Needs the licence.

flink-core/src/test/java/org/apache/flink/api/java/typeutils/PojoParametrizedTypeExtractionTest.java
```diff
@@ -0,0 +1,101 @@
...
+
+import org.apache.flink.api.common.functions.MapFunction;
+import org.apache.flink.api.common.typeinfo.TypeInformation;
```
gaoyunhaii, line 4: blank lines required.

flink-core/src/test/java/org/apache/flink/api/java/typeutils/PojoParametrizedTypeExtractionTest.java
```diff
@@ -0,0 +1,101 @@
...
+import java.io.Serializable;
+
+/**
+ *  Tests concerning type extraction of Parametrized Pojo and its superclasses.
+ */
+public class PojoParametrizedTypeExtractionTest {
+
+	@Test
+	public void testCreateTypeInfo(){
+		/// Init
```
gaoyunhaii, line 17: I tend to remove the comments and the following `Init`, `Extract` and `Check` comments, since the current tests seem to be not very complex and easy to understand.

flink-core/src/test/java/org/apache/flink/api/java/typeutils/PojoParametrizedTypeExtractionTest.java
```diff
@@ -0,0 +1,101 @@
...
+	/**
+	 * Representation of map function for type extraction.
+	 */
+	public static class ConcreteMapFunction implements MapFunction<ParametrizedParentImpl, ParametrizedParentImpl> {
+		@Override
+		public ParametrizedParentImpl map(ParametrizedParentImpl value) throws Exception {
+			return null;
+		}
+	}
+
```
gaoyunhaii, line 100: remove the blank lines

flink-core/src/test/java/org/apache/flink/api/java/typeutils/PojoParametrizedTypeExtractionTest.java
```diff
@@ -0,0 +1,101 @@
...
+		// Check
+		Assert.assertTrue(mapReturnPojoFieldTypeInfo instanceof PojoTypeInfo);
+
+	}
+
+	/**
+	 * Representation of Pojo class with 2 fields.
+	 */
+	public static class Pojo implements Serializable {
```
gaoyunhaii, line 72: How about removing the ` implements Serializable` ? Since it should be irrelevant to this test.

flink-core/src/test/java/org/apache/flink/api/java/typeutils/PojoParametrizedTypeExtractionTest.java
```diff
@@ -0,0 +1,101 @@
...
+		public int digits;
+		public String letters;
+	}
+
+	/**
+	 * Representation of class which is parametrized by some pojo.
+	 */
+	public static class ParameterizedParent<T extends Serializable> implements Serializable {
```
gaoyunhaii, line 80: And also remove the `extends Serializable` and ` implements Serializable` here ? (Together with the comment in Line 72).

flink-core/src/test/java/org/apache/flink/api/java/typeutils/PojoParametrizedTypeExtractionTest.java
```diff
@@ -0,0 +1,101 @@
...
+ */
+public class PojoParametrizedTypeExtractionTest {
+
+	@Test
+	public void testCreateTypeInfo(){
+		/// Init
+		final TypeInformation<ParametrizedParentImpl> directTypeInfo = TypeExtractor.createTypeInfo(ParametrizedParentImpl.class);
+
+		/// Check
+		Assert.assertTrue(directTypeInfo instanceof PojoTypeInfo);
```
gaoyunhaii, line 21: `assertTrue` better also give the `message field`.

flink-core/src/test/java/org/apache/flink/api/java/typeutils/PojoParametrizedTypeExtractionTest.java
```diff
@@ -0,0 +1,101 @@
...
+	}
+
+	@Test
+	public void testFieldExtraction(){
```
gaoyunhaii, line 25: I tend to merge this PR with the above `testCreateTypeInfo` and rename to `testDirectlyCreateTypeInfo`, since here seems to have the same action with the above test and with more asserts. This makes them more like the same test.

If we merge the two tests, we may create an explicit `PojoTypeInfo` object as the expected result and use the `assertEquals` to compare the result with the expected result.

flink-core/src/test/java/org/apache/flink/api/java/typeutils/PojoParametrizedTypeExtractionTest.java
```diff
@@ -0,0 +1,101 @@
...
+		/// Init
+		final TypeInformation<ParametrizedParentImpl> directTypeInfo = TypeExtractor.createTypeInfo(ParametrizedParentImpl.class);
+
+		// Extract
+		TypeInformation<ParametrizedParentImpl> mapReturnTypeInfo = TypeExtractor
+			.getMapReturnTypes(new ConcreteMapFunction(), directTypeInfo);
+
+		// Check
+		Assert.assertTrue(mapReturnTypeInfo instanceof PojoTypeInfo);
+		Assert.assertEquals(directTypeInfo, mapReturnTypeInfo);
```
gaoyunhaii, line 49: I tend to use an explicit `PojoTypeInfo` as the expected result (namely we create a PojoTypeInfo directly). Since both the two type extraction logic are based on similar logic, it is possible that some bad implementation may cause the same errors for the two processes.

Besides, if we use explicit expected result, we could also remove the previous `instanceof` assert.

flink-core/src/test/java/org/apache/flink/api/java/typeutils/PojoParametrizedTypeExtractionTest.java
```diff
@@ -0,0 +1,101 @@
...
+	}
+
+	@Test
+	public void testMapReturnFieldTypeInfo(){
```
gaoyunhaii, line 53: Similarly, I think we might merge `testMapReturnFieldTypeInfo` with the above `testMapReturnTypeInfo`, as commented in Line 25.

gaoyunhaii:
Very thanks @andbul and @dawidwys for the PR! I also agree with that we could remove the condition and always compute the type hierarchy. There are cases that the hierarchy get repeated, for example, if we have 
```
class MyTuple extends Tuple2<String, String>{
   public String anotherField;
}
```
We will first try to analyze it as a Tuple and calculated the hierarchy from MyTuple -> Tuple2, however we will fail since it has additional fields, the we will turn to `analyzePojo` and if we calculated the hierarchy from MyTuple -> Object again, the hierarchy will be repeated.

However, I don't think the repeating will cause errors. Besides, it is hard to detect if there are repeating with the number of hierarchy, since MyTuple -> Object may have arbitrary numbers of types.

I also left some comments for the tests.

dawidwys:
Thank you @andbul for the work and I am sorry nobody reached out to you before. I think it addresses a valid case. Let us know if you'd like to continue with this work. If not we can take over from here.

flink-core/src/main/java/org/apache/flink/api/java/typeutils/TypeExtractor.java
```diff
@@ -1837,8 +1837,8 @@ private boolean isValidPojoField(Field f, Class<?> clazz, ArrayList<Type> typeHi
 		if (parameterizedType != null) {
 			getTypeHierarchy(typeHierarchy, parameterizedType, Object.class);
 		}
-		// create a type hierarchy, if the incoming only contains the most bottom one or none.
-		else if (typeHierarchy.size() <= 1) {
+		// create a type hierarchy, for fields types extraction
```
dawidwys, line 1840: I would use this opportunity to simplify the code a bit. Actually the `clazz` and ` parameterizedType` are very tightly coupled and one originiates from the other. I think we could simplify the method slightly if we removed the `parameterizedType` parameter. Than the beginning of this method could like this:

```
	protected <OUT, IN1, IN2> TypeInformation<OUT> analyzePojo(
			Type type,
			List<Type> typeHierarchy,
			TypeInformation<IN1> in1Type,
			TypeInformation<IN2> in2Type) {

		Class<OUT> clazz = typeToClass(type);
		if (!Modifier.isPublic(clazz.getModifiers())) {
			LOG.info("Class " + clazz.getName() + " is not public so it cannot be used as a POJO type " +
				"and must be processed as GenericType. Please read the Flink documentation " +
				"on \"Data Types & Serialization\" for details of the effect on performance.");
			return new GenericTypeInfo<OUT>(clazz);
		}

		// add the hierarchy of the POJO
		getTypeHierarchy(typeHierarchy, type, Object.class);

...
```

dawidwys:
@gaoyunhaii I think we could easily solve the problem of duplicated hierarchy if we worked on a copy of `typeHierarchy` when determining types for subtypes of a POJO.

gaoyunhaii:
I also agree with that we could duplication the hierarchy to eliminate the duplication. As a whole, I think it would be good if type hierarchy passed as parameters could be kept unchanged.

dawidwys:
Closed with a8847061c40bf8ca17e22e6e412a378f53b8b82d

PR #8360, opened 2019-05-07
[FLINK-12393][table-common] Add the user-facing classes of the new type system

twalthr:

This PR introduces the `DataType` class and subclasses. Users are able to do a star import of `DataTypes` and declare types like: `MULTISET(MULTISET(INT())`. Very close to SQL. As mentioned in FLIP-37, data types allow to specify format hints to the planner using `TIMESTAMP(9).bridgedTo(java.sql.Timestamp.class)`.

## Brief change log

- `DataType` structure
- `DataTypes` enumeration of types

## Verifying this change

See `org.apache.flink.table.types.DataTypesTest`.

  - The serializers: no

## Documentation

### DIFF
--- a/flink-table/flink-table-common/src/main/java/org/apache/flink/table/api/DataTypes.java
+++ b/flink-table/flink-table-common/src/main/java/org/apache/flink/table/api/DataTypes.java
@@ -22,7 +22,11 @@ import org.apache.flink.annotation.PublicEvolving;
 import org.apache.flink.api.common.ExecutionConfig;
 import org.apache.flink.api.common.typeinfo.TypeInformation;
 import org.apache.flink.api.common.typeutils.TypeSerializer;
+import org.apache.flink.table.types.AtomicDataType;
+import org.apache.flink.table.types.CollectionDataType;
 import org.apache.flink.table.types.DataType;
+import org.apache.flink.table.types.FieldsDataType;
+import org.apache.flink.table.types.KeyValueDataType;
 import org.apache.flink.table.types.logical.AnyType;
 import org.apache.flink.table.types.logical.ArrayType;
 import org.apache.flink.table.types.logical.BigIntType;
@@ -81,8 +85,8 @@ public final class DataTypes {
 	 *
 	 * @see CharType
 	 */
-	public static DataType.AtomicDataType CHAR(int n) {
-		return new DataType.AtomicDataType(new CharType(n));
+	public static DataType CHAR(int n) {
+		return new AtomicDataType(new CharType(n));
 	}
 
 	/**
@@ -92,8 +96,8 @@ public final class DataTypes {
 	 *
 	 * @see VarCharType
 	 */
-	public static DataType.AtomicDataType VARCHAR(int n) {
-		return new DataType.AtomicDataType(new VarCharType(n));
+	public static DataType VARCHAR(int n) {
+		return new AtomicDataType(new VarCharType(n));
 	}
 
 	/**
@@ -102,7 +106,7 @@ public final class DataTypes {
 	 *
 	 * @see VarCharType
 	 */
-	public static DataType.AtomicDataType STRING() {
+	public static DataType STRING() {
 		return VARCHAR(Integer.MAX_VALUE);
 	}
 
@@ -111,8 +115,8 @@ public final class DataTypes {
 	 *
 	 * @see BooleanType
 	 */
-	public static DataType.AtomicDataType BOOLEAN() {
-		return new DataType.AtomicDataType(new BooleanType());
+	public static DataType BOOLEAN() {
+		return new AtomicDataType(new BooleanType());
 	}
 
 	/**
@@ -122,8 +126,8 @@ public final class DataTypes {
 	 *
 	 * @see BinaryType
 	 */
-	public static DataType.AtomicDataType BINARY(int n) {
-		return new DataType.AtomicDataType(new BinaryType(n));
+	public static DataType BINARY(int n) {
+		return new AtomicDataType(new BinaryType(n));
 	}
 
 	/**
@@ -133,8 +137,8 @@ public final class DataTypes {
 	 *
 	 * @see VarBinaryType
 	 */
-	public static DataType.AtomicDataType VARBINARY(int n) {
-		return new DataType.AtomicDataType(new VarBinaryType(n));
+	public static DataType VARBINARY(int n) {
+		return new AtomicDataType(new VarBinaryType(n));
 	}
 
 	/**
@@ -143,7 +147,7 @@ public final class DataTypes {
 	 *
 	 * @see VarBinaryType
 	 */
-	public static DataType.AtomicDataType BYTES() {
+	public static DataType BYTES() {
 		return VARBINARY(Integer.MAX_VALUE);
 	}
 
@@ -155,8 +159,8 @@ public final class DataTypes {
 	 *
 	 * @see DecimalType
 	 */
-	public static DataType.AtomicDataType DECIMAL(int precision, int scale) {
-		return new DataType.AtomicDataType(new DecimalType(precision, scale));
+	public static DataType DECIMAL(int precision, int scale) {
+		return new AtomicDataType(new DecimalType(precision, scale));
 	}
 
 	/**
@@ -164,8 +168,8 @@ public final class DataTypes {
 	 *
 	 * @see TinyIntType
 	 */
-	public static DataType.AtomicDataType TINYINT() {
-		return new DataType.AtomicDataType(new TinyIntType());
+	public static DataType TINYINT() {
+		return new AtomicDataType(new TinyIntType());
 	}
 
 	/**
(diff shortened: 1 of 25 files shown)


flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/DataType.java
```diff
@@ -0,0 +1,427 @@
...
+				Preconditions.checkNotNull(newConversionClass, "New conversion class must not be null."));
+		}
+	}
+
+	/**
+	 * A data type that contains an element type (e.g. {@code ARRAY} or {@code MULTISET}).
+	 *
+	 * @see DataTypes for a list of supported data types
+	 */
+	public static final class ElementDataType extends DataType {
```
wuchong: CollectionDataType? 

It is wired that the `ElementDataType` contains an element type named `elementDataType`.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/DataType.java
```diff
@@ -0,0 +1,427 @@
...
+
+	/**
+	 * Adds a hint that null values are expected in the data for this type (default behavior).
+	 *
+	 * <p>This method exists for explicit declaration of the default behavior or for invalidation of
+	 * a previous call to {@link #notNull()}.
+	 *
+	 * @return a new, reconfigured data type instance
+	 */
+	public abstract DataType andNull();
```
wuchong: How about rename `andNull()` --> `asNullable()` and `notNull()` --> `asNotNull()`?

flink-table/flink-table-common/src/main/java/org/apache/flink/table/api/DataTypes.java
```diff
@@ -0,0 +1,703 @@
...
+	 */
+	public static Resolution HOUR() {
+		return new Resolution(Resolution.IntervalUnit.HOUR);
+	}
+
+	/**
+	 * Resolution in days. The precision is the number of digits of days. It must have a value
+	 * between 1 and 6 (both inclusive). If no precision is specified, it is equal to 2 by default.
+	 */
+	public static Resolution DAY(int precision) {
```
wuchong: What about to provide a `DAY()` method which uses 2 precision as default? The same as YEAR and SECOND.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/DataType.java
twalthr: The naming is on purpose as `MapType` is also a collection data type. The naming should not reflect the SQL type families but just serve as a container name.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/DataType.java
twalthr: 1. The goal was to be close to SQL: `INT NOT NULL` -> `INT().notNull()`.
2. `notNull` and `andNull` are shorter than `asNullable` and `asNotNullable`.

Alternatively, what about `withNull()` or `withoutNull`?

flink-table/flink-table-common/src/main/java/org/apache/flink/table/api/DataTypes.java
twalthr: Sounds good.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/DataType.java
wuchong: The only concern to me is the class name `ElementDataType` and the member field `elementDataType` is confusing. 

In JDK, `Collection` is a group of elements, and `Map` is not a collection. IMO, `ARRAY` and `MULTISET` are collections of elements with same type.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/DataType.java
wuchong: It makes sense to be close to SQL. 

In SQL, a nullable field is declared like `name INT NULL`. But it's not good choice in API `INT().null()`.

What about `notNull()` and `nullable()` ?  It's easy to understand and close to SQL.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/DataType.java
twalthr: You are right. Java is also a bit confusing in this sense as methods such as `java.util.Collections#emptyMap()` are also put under collections. I will correct that. Also in `org.apache.flink.table.types.logical.LogicalTypeRoot#MAP`. Thanks for the feedback.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/DataType.java
twalthr: Yes, I'm also ok with `notNull()` and `nullable()`. Any third opinion? @dawidwys

flink-table/flink-table-common/src/test/java/org/apache/flink/table/types/DataTypesTest.java
```diff
@@ -0,0 +1,133 @@
...
+
+	@Test
+	public void testAtomicDataTypes() {
+		testLogicalType(new CharType(2), DataTypes.CHAR(2));
+		testLogicalType(new VarCharType(2), DataTypes.VARCHAR(2));
+		testLogicalType(new VarCharType(Integer.MAX_VALUE), DataTypes.STRING());
+
+		testLogicalType(
+			new DayTimeIntervalType(DayTimeResolution.MINUTE_TO_SECOND, DayTimeIntervalType.DEFAULT_DAY_PRECISION, 2),
+			DataTypes.INTERVAL(DataTypes.MINUTE(), DataTypes.SECOND(2)));
```
dawidwys: nit: Not sure about that, but maybe we could static import the `DataTypes`? It makes it easier to read.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/api/DataTypes.java
```diff
@@ -0,0 +1,703 @@
...
+	// we use SQL-like naming for data types and avoid Java keyword clashes
+	// CHECKSTYLE.OFF: MethodName
+
+	/**
+	 * Data type of a fixed-length character string {@code CHAR(n)} where {@code n} is the number
+	 * of code points. {@code n} must have a value between 1 and 255 (both inclusive).
+	 *
+	 * @see CharType
+	 */
+	public static DataType.AtomicDataType CHAR(int n) {
```
dawidwys: Can we return just the `DataType` rather than more concrete subclasses? I think we should try to expose as little internal structure as possible. This would give us more flexibility over the API, and it would still be quite easy to expose more if needed in the future.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/DataType.java
```diff
@@ -0,0 +1,427 @@
...
+	 * @see DataTypes for a list of supported data types
+	 */
+	public static final class FieldsDataType extends DataType {
+
+		private final Map<String, DataType> fieldDataTypes;
+
+		public FieldsDataType(
+				LogicalType logicalType,
+				@Nullable Class<?> conversionClass,
+				Map<String, DataType> fieldDataTypes) {
```
JingsongLi: It's a bit strange to use `Map<String, DataType>` to represent fields, because RowType is sequential, is it better to use array to express fields?

wuchong: I agree. At least we should use LinkedHashMap to preserve the fields order. And it would be nice to provide `getFieldNames()` and `getFieldTypes()`  for convenience.

twalthr: The field order is defined by the logical type. The map adds just metadata.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/api/DataTypes.java
twalthr: Good point. First I was a bit skeptical because it would require casting in util methods but we can also provide a visitor.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/DataType.java
twalthr: If we would allow a `LinkedHashMap` we would also need to verify the order with the logical type. The logical type should be the single source of truth.

wuchong: OK, I'm just not sure what the `getFieldDataTypes` will be used for.

twalthr: This method will only used by us to figure out which format the user requested. It will be used in  the next PR around FLINK-12254.

twalthr:
Thanks everyone for the feedback. I will merge this now.

PR #8370, opened 2019-05-08
[FLINK-12445][yarn] Cancel application on failure

zentol:
Fixes an issue in the `YARNSessionCapacitySchedulerITCase` where the application was not explicitly stopped if an assertion fails, resulting in misleading surefire output.


StephanEwen:
Looks good, +1 to merge this

PR #6259, opened 2018-07-04
[FLINK-9679] Implement AvroSerializationSchema

medcv:
## What is the purpose of the change

Provides implementation of AvroSerializationSchema that write records serialized as avro and also provides version that uses Confluent Schema Registry to write the record.

This is following AvroDESerializationSchema implementation patterns to have a consistent code base for Ser/Des.

## Brief change log

- Implemented AvroSerializationSchema / RegistryAvroSerializationSchema / ConfluentRegistryAvroSerializationSchema

## Does this pull request potentially affect one of the following parts:

  - Dependencies (does it add or upgrade a dependency): (yes)
  - The public API, i.e., is any changed class annotated with `@Public(Evolving)`: (no)
  - The serializers: (yes)
  - The runtime per-record code paths (performance sensitive): (no)
  - Anything that affects deployment or recovery: JobManager (and its components), Checkpointing, Yarn/Mesos, ZooKeeper: (no)
  - The S3 file system connector: (no)

## Documentation

  - Does this pull request introduce a new feature? (yes)
  - If yes, how is the feature documented? (JavaDocs)

PR #8371, opened 2019-05-08
[FLINK-9679] - Add AvroSerializationSchema

Wosin:

Provides implementation of AvroSerializationSchema that write records serialized as avro and also provides version that uses Confluent Schema Registry to write the record.

This is following AvroDESerializationSchema implementation patterns to have a consistent code base for Ser/Des.

This is follow-up of #6259 

## Brief change log
Provides implementation of AvroSerializationSchema that write records serialized as avro and also provides version that uses Confluent Schema Registry to write the record.

Implemented AvroSerializationSchema / RegistryAvroSerializationSchema / ConfluentRegistryAvroSerializationSchema

## Verifying this change

- Added Tests that verify proper serialization of messages for both Specific and Generic case.

  - The serializers: no

## Documentation

### DIFF
--- a/flink-formats/flink-avro-confluent-registry/pom.xml
+++ b/flink-formats/flink-avro-confluent-registry/pom.xml
@@ -29,6 +29,10 @@ under the License.
 
 	<artifactId>flink-avro-confluent-registry</artifactId>
 
+	<properties>
+		<confluent.schema.registry.version>4.1.0</confluent.schema.registry.version>
+	</properties>
+
 	<repositories>
 		<repository>
 			<id>confluent</id>
@@ -40,7 +44,7 @@ under the License.
 		<dependency>
 			<groupId>io.confluent</groupId>
 			<artifactId>kafka-schema-registry-client</artifactId>
-			<version>3.3.1</version>
+			<version>${confluent.schema.registry.version}</version>
 			<exclusions>
 				<exclusion>
 					<groupId>org.apache.avro</groupId>
--- a/flink-formats/flink-avro-confluent-registry/src/main/java/org/apache/flink/formats/avro/registry/confluent/ConfluentRegistryAvroSerializationSchema.java
+++ b/flink-formats/flink-avro-confluent-registry/src/main/java/org/apache/flink/formats/avro/registry/confluent/ConfluentRegistryAvroSerializationSchema.java
@@ -1,0 +1,143 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.formats.avro.registry.confluent;
+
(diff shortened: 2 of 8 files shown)


dawidwys:

Wosin:
There is one error here that I have found. I will make fix commit today.

Wosin:
Hey, 
Any update on this ??

aljoscha:
@Wosin could you please rebase and squash the various commits into some commits that make sense, i.e. separate pre-requisites like version upgrade into a separate commit and then have the actual implementation commit? I'd like to review this.

Wosin:
@aljoscha 
Sorry It took so long.... I have made the rebase & squash.

soumyasmruti:
Using ConfluentRegistryAvroSerialization schema auto registers the avro schema in bytes. Also, when an avro schema is registered in advance when sending message getting the following error. 

```Caused by: org.apache.kafka.common.errors.SerializationException: Error registering Avro schema: "bytes"
Caused by: io.confluent.kafka.schemaregistry.client.rest.exceptions.RestClientException: Schema being registered is incompatible with an earlier schema; error code: 409; error code: 409
```

aljoscha:
@soumyasmruti So you're saying the PR in the current state does not work? @Wosin Do you have a comment on that?

derjust:
I used this PR fairly successful over a week producing ~50 mil messages in total.

There is only one catch in the PR that needs to be fixed: https://github.com/apache/flink/pull/8371/files#diff-c60754c7ce564f4229c22913d783c339R128

```java
int schemaId= schemaCoderProvider.get()
					.writeSchema(getSchema());
```

This creates a new http client each time a message is serialized which creates a significant overhead due to establishing the HTTP connections but also never leveraging the schema caching 
The `schemaCoderProvider.get()` should be cached/kept local after it has been fetched and used for all subsequent `serialize()` calls.
In fact, it should be done in the same fashion as in the existing source implementation:

```java
  @Override
  protected void checkAvroInitialized() {
    super.checkAvroInitialized();
    if (schemaCoder == null) {
      this.schemaCoder = schemaCoderProvider.get();
    }

```

Wosin:
@soumyasmruti Can You elaborate on Your problem ? I have never observed this behavior before and I have used this schema on multiple occasions. Accoring to the confluent docs https://docs.confluent.io/current/schema-registry/schema_registry_tutorial.html this is the expected behavior if the schemas are not compatible. Maybe that's the issue ?

aljoscha:
There are build failures because of checkstyle: https://travis-ci.com/flink-ci/flink/builds/137108234

soumyasmruti:
@Wosin I was having problems while Serializing using SpecificRecord but I haven't tried it in a while. As far as I remember Generic Record Serialization was working fine. But you can ignore my comments if you find everything is working fine.

aljoscha:
@Wosin Is there a reason `RegistryAvroSerializationSchema.serialize()` and `ConfluentRegistryAvroSerializationSchema.serialize()` are different? Shouldn't both register the scheam and serialize the schema id?

Also, in `RegistryAvroSerializationSchema` we don't use the `schemaCoder` that is set in `checkAvroInitialized()`. I can fix that while merging because I also fixed the checkstyle problems in my local checkout of the PR.

Wosin:
Hey @aljoscha, 
I have some issues with Travis, but I will try to fix the rest of the checkstyle issues today:) As for the schema coder I will take a look at this and try to fix that too. 

As for the difference between those classes, technically we could make `SchemaCoder` return the `byte[]` instead of int and append it to the serialized message. I haven't thought of that before tbh.
Generally, the difference comes from the fact that there are some custom bytes in Confluent serialized messages.

aljoscha:
@Wosin could you please rebase on master and squash it into sensible commits, as you would like to have them merged in the end.

aljoscha:
I pushed a branch that squashes commits and adds some simplification. I'm now working on extending `test_confluent_schema_registry.sh` to also test writing to a schema registry because that part of the code is completely untested as of now.

I'm also wondering about the new `subject` parameter of `ConfluentSchemaRegistryCoder`, we use this when writing the schema but never use it when reading to schema.

aljoscha:
I merged, thanks a lot for your work and patience! 😅

PR #8373, opened 2019-05-08
[FLINK-11922][metrics] Support MetricReporter factories 

zentol:

Adds support for `MetricReporter` factories. With this change reporters no longer have to adhere to restrictions imposed by generic instantiation (public no arg constructor, initializing things lazily in open()).

Factories are loaded through a `ServiceLoader` in the `ReporterSetup`. A migration to the new `Plugin` mechanism should be trivial.

Factories are configured similarly to reporter classes ( `reporter.*.factory.class` instead of `reporter.*.class`).

This PR additionally migrates the `JMXReporter` to use factories instead, serving as a sanity-check.

### Backwards-compatibility
The existing mechanism for instantiating reporters via reflection still exists for the sake of backwards-compatibility, but is likely to be removed in the future.
Reporters may support both factories and reflection , but are neither required nor encouraged to do so.
For backwards compatibility reporters may be annotated with `InstantiateViaFactory(String factoryClassName)` to automatically forward reflection-based instantiation attempts to the factory.

## Verifying this change

The `ReporterSetupTest` was extended to cover the handling of factory configurations.

Added:
* JMXReporterFactoryTest

Existing tests/E2E tests use (JMX) reporters and thus ensure that
* reflection-based reporter still work
* (now) out-dated JMX configurations are still honored

## Documentation

  - If yes, how is the feature documented? (docs / JavaDocs)

### DIFF
--- a/flink-metrics/flink-metrics-jmx/src/main/java/org/apache/flink/metrics/jmx/JMXReporter.java
+++ b/flink-metrics/flink-metrics-jmx/src/main/java/org/apache/flink/metrics/jmx/JMXReporter.java
@@ -26,6 +26,7 @@ import org.apache.flink.metrics.Meter;
 import org.apache.flink.metrics.Metric;
 import org.apache.flink.metrics.MetricConfig;
 import org.apache.flink.metrics.MetricGroup;
+import org.apache.flink.metrics.reporter.InstantiateViaFactory;
 import org.apache.flink.metrics.reporter.MetricReporter;
 import org.apache.flink.runtime.metrics.groups.AbstractMetricGroup;
 import org.apache.flink.runtime.metrics.groups.FrontMetricGroup;
@@ -34,6 +35,7 @@ import org.apache.flink.util.NetUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import javax.annotation.Nullable;
 import javax.management.InstanceAlreadyExistsException;
 import javax.management.InstanceNotFoundException;
 import javax.management.MBeanServer;
@@ -55,6 +57,7 @@ import java.util.HashMap;
 import java.util.Hashtable;
 import java.util.Iterator;
 import java.util.Map;
+import java.util.Optional;
 
 /**
  * {@link MetricReporter} that exports {@link Metric Metrics} via JMX.
@@ -62,11 +65,10 @@ import java.util.Map;
  * <p>Largely based on the JmxReporter class of the dropwizard metrics library
  * https://github.com/dropwizard/metrics/blob/master/metrics-core/src/main/java/io/dropwizard/metrics/JmxReporter.java
  */
+@InstantiateViaFactory(factoryClassName = "org.apache.flink.metrics.jmx.JMXReporterFactory")
 public class JMXReporter implements MetricReporter {
 
 	static final String JMX_DOMAIN_PREFIX = "org.apache.flink.";
-
-	public static final String ARG_PORT = "port";
 
 	private static final Logger LOG = LoggerFactory.getLogger(JMXReporter.class);
 
@@ -86,33 +88,24 @@ public class JMXReporter implements MetricReporter {
 	private final Map<Metric, ObjectName> registeredMetrics;
 
 	/** The server to which JMX clients connect to. Allows for better control over port usage. */
-	private JMXServer jmxServer;
+	@Nullable
+	private final JMXServer jmxServer;
 
-	public JMXReporter() {
+	JMXReporter(@Nullable final String portsConfig) {
 		this.mBeanServer = ManagementFactory.getPlatformMBeanServer();
 		this.registeredMetrics = new HashMap<>();
-	}
-
-	// ------------------------------------------------------------------------
-	//  life cycle
-	// ------------------------------------------------------------------------
-
-	@Override
-	public void open(MetricConfig config) {
-		String portsConfig = config.getString(ARG_PORT, null);
 
 		if (portsConfig != null) {
 			Iterator<Integer> ports = NetUtils.getPortRangeFromString(portsConfig);
 
-			JMXServer server = new JMXServer();
-			while (ports.hasNext()) {
+			JMXServer successfullyStartedServer = null;
+			while (ports.hasNext() && successfullyStartedServer == null) {
+				JMXServer server = new JMXServer();
 				int port = ports.next();
 				try {
 					server.start(port);
 					LOG.info("Started JMX server on port " + port + ".");
-					// only set our field if the server was actually started
-					jmxServer = server;
-					break;
+					successfullyStartedServer = server;
 				} catch (IOException ioe) { //assume port conflict
 					LOG.debug("Could not start JMX server on port " + port + ".", ioe);
 					try {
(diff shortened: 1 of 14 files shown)


flink-runtime/src/main/java/org/apache/flink/runtime/metrics/ReporterSetup.java
```diff
@@ -131,46 +137,119 @@ private static ReporterSetup createReporterSetup(String reporterName, MetricConf
...
-				final String reporterClassName = reporterConfig.getString(ConfigConstants.METRICS_REPORTER_CLASS_SUFFIX, null);
-				if (reporterClassName == null) {
-					LOG.error("No reporter class set for reporter " + reporterName + ". Metrics might not be exposed/reported.");
-					continue;
-				}
-
-				Class<?> reporterClass = Class.forName(reporterClassName);
-				MetricReporter reporter = (MetricReporter) reporterClass.newInstance();
+				Optional<MetricReporter> metricReporterOptional = loadReporter(reporterName, reporterConfig, reporterFactories);
+				if (metricReporterOptional.isPresent()) {
```
tillrohrmann: `metricReporterOptional.ifPresent()` is a bit more succinct.

flink-runtime/src/main/java/org/apache/flink/runtime/metrics/ReporterSetup.java
```diff
@@ -131,46 +137,119 @@ private static ReporterSetup createReporterSetup(String reporterName, MetricConf
...
 				LOG.error("Could not instantiate metrics reporter {}. Metrics might not be exposed/reported.", reporterName, t);
 			}
 		}
 		return reporterArguments;
 	}
+
+	private static Collection<MetricReporterFactory> loadReporterFactories() {
+		final ServiceLoader<MetricReporterFactory> serviceLoader = ServiceLoader.load(MetricReporterFactory.class);
+
+		final Collection<MetricReporterFactory> reporterFactories = new ArrayList<>();
```
tillrohrmann: Specify capacity.

flink-runtime/src/main/java/org/apache/flink/runtime/metrics/ReporterSetup.java
```diff
@@ -131,46 +137,119 @@ private static ReporterSetup createReporterSetup(String reporterName, MetricConf
...
+			throws ClassNotFoundException, IllegalAccessException, InstantiationException {
+
+		final String reporterClassName = reporterConfig.getString(ConfigConstants.METRICS_REPORTER_CLASS_SUFFIX, null);
+		final String factoryClassName = reporterConfig.getString(ConfigConstants.METRICS_REPORTER_FACTORY_CLASS_SUFFIX, null);
+
+		if (factoryClassName != null) {
+			return loadViaFactory(factoryClassName, reporterName, reporterConfig, reporterFactories);
+		}
+
+		if (reporterClassName != null) {
```
tillrohrmann: The method could be split up here to make it a bit simpler.

flink-runtime/src/main/java/org/apache/flink/runtime/metrics/ReporterSetup.java
```diff
@@ -131,46 +137,119 @@ private static ReporterSetup createReporterSetup(String reporterName, MetricConf
...
+
+		LOG.warn("No reporter class nor factory set for reporter {}. Metrics might not be exposed/reported.", reporterName);
+		return Optional.empty();
+	}
+
+	private static Optional<MetricReporter> loadViaFactory(
+			final String factoryClassName,
+			final String reporterName,
+			final Configuration reporterConfig,
+			final Collection<MetricReporterFactory> reporterFactories) {
```
tillrohrmann: I think it would make sense to construct a `Map<String, MetricReporterFactory>` for easier lookups at an earlier place.

flink-runtime/src/main/java/org/apache/flink/runtime/metrics/ReporterSetup.java
```diff
@@ -131,46 +137,119 @@ private static ReporterSetup createReporterSetup(String reporterName, MetricConf
...
+
+		final Optional<MetricReporterFactory> factory = reporterFactories.stream()
+			.filter(factoryCandidate -> factoryClassName.equals(factoryCandidate.getClass().getName()))
+			.findAny();
+
+		if (!factory.isPresent()) {
+			LOG.warn("The reporter factory ({}) could not be found for reporter {}.", factoryClassName, reporterName);
+			return Optional.empty();
+		} else {
```
tillrohrmann: Here we could use `factory.map(f -> f.createMetricReporter())`

flink-metrics/flink-metrics-jmx/src/main/java/org/apache/flink/metrics/jmx/JMXReporterFactory.java
```diff
@@ -0,0 +1,37 @@
...
+
+import java.util.Optional;
+import java.util.Properties;
+
+/**
+ * {@link MetricReporterFactory} for {@link JMXReporter}.
+ */
+public class JMXReporterFactory implements MetricReporterFactory {
+
+	static final String ARG_PORT = "port";
```
tillrohrmann: This is out of scope of this PR but maybe something to consider for the future. I think the `MetricReporter` configuration parameters are hard to discover. Maybe we could use `ConfigOptions` and their utilities to bundle definition and description as well as to generate the reporter documentation automatically. We would also get type checks for free. One could, for example define a `JMXReporterConfigOptions` class which contains all the different `ConfigOptions`. Then only needs to include the generated config documentation in the `metrics.md` file.

flink-metrics/flink-metrics-jmx/src/main/java/org/apache/flink/metrics/jmx/JMXReporter.java
```diff
@@ -122,13 +115,24 @@ public void open(MetricConfig config) {
...
-			if (jmxServer == null) {
-				throw new RuntimeException("Could not start JMX server on any configured port. Ports: " + portsConfig);
+			if (successfullyStartedServer == null) {
+				throw new RuntimeException("Could not start JMX server on any configured port. Ports: " + portsConfig.get());
 			}
+			this.jmxServer = successfullyStartedServer;
+		} else {
+			this.jmxServer = null;
 		}
 		LOG.info("Configured JMXReporter with {port:{}}", portsConfig);
```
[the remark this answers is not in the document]
zentol: no, the purpose here is to log the parameters that have been evaluated by the reporter. If you configured a port, but the port is not listed here, you'll see the mismatch.

flink-metrics/flink-metrics-jmx/src/main/java/org/apache/flink/metrics/jmx/JMXReporterFactory.java
zentol: I brought up this very point when we added the InfluxDB reporter.

We cannot do this at the moment.
The docs generator enforces that all config option keys are well-defined; i.e., for a given key there is exactly 1 description and default value.

The parameters here are not fully qualified however but only suffixes, and would thus likely clash with other reporters that also have a "port" parameter.
I haven't found the time yet to add support for this.

PR #8377, opened 2019-05-09
[FLINK-12455][python] Move the packaging of pyflink to flink-dist

dianfu:

This pull request moves the packaging of pyflink to flink-dist for the following reasons:
 - *flink-dist is responsible for the packaging of flink, moving the packaging logic there is more straightforward*
 - *The pom.xml under flink-python can be removed and this makes flink-python a pure python module*

## Brief change log

  - *Moves the packaging logic to flink-dist*
  - *Removes the pom.xml under flink-python*

## Verifying this change

  - The serializers: (no)

## Documentation


flink-dist/pom.xml
```diff
@@ -588,6 +588,31 @@ under the License.
...
+					<execution>
+						<id>package</id>
+						<phase>package</phase>
+						<goals>
+							<goal>run</goal>
+						</goals>
+						<configuration>
+							<target>
+								<delete>
+									<fileset dir="${project.basedir}/../flink-python/pyflink" includes="**/*.pyc"/>
```
zentol: whenever you have to move into other modules like this it is a good sign that you're doing something that you really shouldn't. We've run into _so_ many issues due to things like this that I'm inclined to reject this PR.

sunjincheng121:
@flinkbot approve all
Merging

flink-dist/pom.xml
zentol: I mean that this logic should continue to be encapsulated in flink-python. Flink-dist should only do 1 thing: pick assembled artifacts and copy them somewhere.

What problem are we solving here? What's so bad about having a separate pom in flink-python?

zentol: The batch Python API for example bundles the python classes as a zip in the jar; which has the nice advantage that flink-dist doesn't even know about this.

sunjincheng121: Ohh, My initial idea was to put it in` flink-python`, and I think there's only one motivation for this change: to make `flink-python` purer. But I agree that `flink-dist` also has a pure responsibility for only copying the files it needs. 

What do you think @dianfu

dianfu: The original motivation of this PR is that pom is a Java kind of things and we want to get rid of it in flink-python module and make flink-python module a pure Python module. But considering that this may introduce another problem just as @zentol are concerned, I'm fine to keep things there it is.

PR #8379, opened 2019-05-09
[FLINK-12457][table-planner-blink] Remove char support in blink

JingsongLi:

Sql types don't support java char. CharType(InternalType) in blink and CharType of the new type system are very confusing, we should remove the current support for Java char in blink.

  - The serializers: (no)

## Documentation


JingsongLi:
> LGTM，btw does flink planner support CHAR right?

In flink planner:
There are some useless code (three lines) in `CodeGenUtils.primitiveTypeTermForTypeInfo` and etc..
But `FlinkTypeFactory` don't support `CHAR_TYPE_INFO` with `throw new TableException("Character type is not supported.")`.
So flink planner does not support `CHAR_TYPE_INFO`.

PR #8385, opened 2019-05-09
[FLINK-12463][tests] Deduplicate empty .out log file checks

zentol:
Deduplicate logic for checking whether .out files are empty. The existing function in `common.sh` accounts for output unavoidable on Java 9, while the others do not, resulting in test failures.

### DIFF
--- a/flink-end-to-end-tests/test-scripts/common_ha.sh
+++ b/flink-end-to-end-tests/test-scripts/common_ha.sh
@@ -18,6 +18,8 @@
 # limitations under the License.
 ################################################################################
 
+source "${END_TO_END_DIR}"/test-scripts/common.sh
+
 # flag indicating if we have already cleared up things after a test
 CLEARED=0
 
@@ -53,13 +55,12 @@ function verify_num_occurences_in_logs() {
 }
 
 function verify_logs() {
-    local OUTPUT=$FLINK_DIR/log/*.out
     local JM_FAILURES=$1
     local EXIT_CODE=0
     local VERIFY_CHECKPOINTS=$2
 
     # verify that we have no alerts
-    if ! [ `cat ${OUTPUT} | wc -l` -eq 0 ]; then
+    if ! check_logs_for_non_empty_out_files; then
         echo "FAILURE: Alerts found at the general purpose job."
         EXIT_CODE=1
     fi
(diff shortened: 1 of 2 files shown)


flink-end-to-end-tests/test-scripts/common_ha.sh
```diff
@@ -59,10 +61,8 @@ function verify_logs() {
     local VERIFY_CHECKPOINTS=$2
 
     # verify that we have no alerts
-    if ! [ `cat ${OUTPUT} | wc -l` -eq 0 ]; then
```
tillrohrmann: let's remove the `OUTPUT` variable if possible.

flink-end-to-end-tests/test-scripts/common_ha.sh
```diff
@@ -59,10 +61,8 @@ function verify_logs() {
...
-        echo "FAILURE: Alerts found at the general purpose job."
-        EXIT_CODE=1
-    fi
+    check_logs_for_non_empty_out_files
+    EXIT_CODE=$(($EXIT_CODE+$?))
```
tillrohrmann: Wouldn't it be better to fail early with the actual exit code instead of summing it up and potentially obfuscate what the actual problem was?

flink-end-to-end-tests/test-scripts/test_ha_per_job_cluster_datastream.sh
```diff
@@ -69,10 +69,8 @@ function verify_logs_per_job() {
     local EXIT_CODE=0
 
     # verify that we have no alerts
-    if ! [ `cat ${OUTPUT} | wc -l` -eq 0 ]; then
```
tillrohrmann: Same here.

flink-end-to-end-tests/test-scripts/test_ha_per_job_cluster_datastream.sh
```diff
@@ -69,10 +69,8 @@ function verify_logs_per_job() {
...
-        echo "FAILURE: Alerts found at the general purpose job."
-        EXIT_CODE=1
-    fi
+    check_logs_for_non_empty_out_files
+    EXIT_CODE=$(($EXIT_CODE+$?))
```
tillrohrmann: And here.

flink-end-to-end-tests/test-scripts/common_ha.sh
zentol: I made a small change to again print something if this check fails, restoring the existing state.
I think it's fine to fail late; if a test fails you can end up with something like:
```
<log output>
FAILURE: Alerts found at the general purpose job.
FAILURE: A JM did not take over.
FAILURE: A JM did not execute the job.
One or more tests FAILED.
```
I think this isn't obfuscating anything.

PR #8387, opened 2019-05-09
[FLINK-11982] [table] File System Connector's support JSON Format and JSON file BatchTableSource

ambition119:

fix [ISSUE #11982](https://issues.apache.org/jira/browse/FLINK-11982) 

## Brief change log

-  [flink-connector-json] module to support File System Connector's JSON Format.
-  JsonBatchTableSourceFactory is a BatchTableSourceFactory
-  JsonBatchTableSource is a BatchTableSource

## Verifying this change

Unit tests for add java class.

*[**flink json format doc**](https://ci.apache.org/projects/flink/flink-docs-release-1.8/dev/table/connect.html#json-format) : File system connector use JSON Format example:*
```java
FileSystem connector = new FileSystem().path(path);

// or withFormatSchema
tEnv.connect(connector)
       .withFormat(new Json()
				.schema(new RowTypeInfo(FIELD_TYPES, FIELD_NAMES))
				.failOnMissingField(true)
      )
      .registerTableSource("jsonTable");

// or withFormatJsonSchema
tEnv.connect(connector)
	.withFormat(new Json()
				.jsonSchema("{\"type\":\"object\",\"properties\":{" +
				"\"author\":{\"type\":\"string\"}," +
				"\"title\":{\"type\":\"string\"}," +
				"\"price\":{\"type\":\"number\"}," +
				"\"zip_code\":{\"type\":\"string\"}" +
				"}}")
				.failOnMissingField(true)
        )
	.registerTableSource("jsonTable");

//or withScheam
tEnv.connect(connector)
	.withFormat(new Json().failOnMissingField(true).deriveSchema())
	.withSchema(
				new Schema()
					.field("author", Types.STRING())
					.field("title", Types.STRING())
					.field("price", Types.DOUBLE())
					.field("zip_code", Types.STRING())
          )
	.registerTableSource("jsonTable");

String fullSql = "select * FROM jsonTable";
Table fullTable = tEnv.sqlQuery(fullSql);
``` 

  - Dependencies (does it add or upgrade a dependency): (**yes**)
  - The serializers: (**no**)

## Documentation

  - If yes, how is the feature documented? (example)

### DIFF
--- a/flink-connectors/flink-connector-json/pom.xml
+++ b/flink-connectors/flink-connector-json/pom.xml
@@ -1,0 +1,166 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+		 xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+		 xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+	<modelVersion>4.0.0</modelVersion>
+
+	<parent>
+		<groupId>org.apache.flink</groupId>
+		<artifactId>flink-connectors</artifactId>
+		<version>1.9-SNAPSHOT</version>
+		<relativePath>..</relativePath>
+	</parent>
+
+	<artifactId>flink-connector-json_${scala.binary.version}</artifactId>
+	<name>flink-connector-json</name>
+
+	<packaging>jar</packaging>
+
+	<dependencies>
+
+		<!-- core dependencies -->
+		<dependency>
+			<groupId>org.apache.flink</groupId>
+			<artifactId>flink-streaming-scala_${scala.binary.version}</artifactId>
+			<version>${project.version}</version>
+			<scope>provided</scope>
+		</dependency>
+
+		<!-- Table ecosystem -->
+		<!-- Projects depending on this project won't depend on flink-table-*. -->
+		<dependency>
+			<groupId>org.apache.flink</groupId>
+			<artifactId>flink-table-api-java-bridge_${scala.binary.version}</artifactId>
+			<version>${project.version}</version>
+			<scope>provided</scope>
+			<optional>true</optional>
+		</dependency>
+
+		<dependency>
+			<groupId>org.apache.flink</groupId>
+			<artifactId>flink-table-common</artifactId>
+			<version>${project.version}</version>
+			<scope>provided</scope>
+			<optional>true</optional>
+		</dependency>
+
+		<!-- A planner dependency won't be necessary once FLIP-32 has been completed. -->
+		<dependency>
+			<groupId>org.apache.flink</groupId>
+			<artifactId>flink-table-planner_${scala.binary.version}</artifactId>
+			<version>${project.version}</version>
+			<scope>provided</scope>
+			<optional>true</optional>
+		</dependency>
+
+		<dependency>
+			<groupId>org.apache.flink</groupId>
+			<artifactId>flink-json</artifactId>
+			<version>${project.version}</version>
+			<scope>provided</scope>
+			<optional>true</optional>
+		</dependency>
+
+		<!-- test dependencies -->
+
+		<dependency>
+			<groupId>org.apache.flink</groupId>
+			<artifactId>flink-test-utils-junit</artifactId>
+		</dependency>
+
+		<dependency>
+			<groupId>org.apache.flink</groupId>
+			<artifactId>flink-test-utils_${scala.binary.version}</artifactId>
+			<version>${project.version}</version>
+			<scope>test</scope>
+		</dependency>
+
+		<dependency>
+			<groupId>org.apache.flink</groupId>
+			<artifactId>flink-clients_${scala.binary.version}</artifactId>
+			<version>${project.version}</version>
+			<scope>test</scope>
+		</dependency>
+
+		<dependency>
+			<groupId>org.apache.flink</groupId>
+			<artifactId>flink-core</artifactId>
+			<version>${project.version}</version>
+			<scope>test</scope>
+			<type>test-jar</type>
+		</dependency>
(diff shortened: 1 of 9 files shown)


StephanEwen:
In this PR, the format "JSON" and the connector "file" are mixed into one aspect.

In the design for sources, we want to keep format and connector separate, so they can be combined (i.e., combine file with Avro, combine Kafka with JSON, etc.).
The way to support JSON files would be to combine the "file" Connector with the Json Format.

There is also a [flink-json](https://github.com/apache/flink/tree/master/flink-formats/flink-json) module already.

Does that approach currently not work?

@twalthr or @dawidwys Could you share some details here?

ambition119:
> In this PR, the format "JSON" and the connector "file" are mixed into one aspect.
> 
> In the design for sources, we want to keep format and connector separate, so they can be combined (i.e., combine file with Avro, combine Kafka with JSON, etc.).
> The way to support JSON files would be to combine the "file" Connector with the Json Format.
> 
> There is also a [flink-json](https://github.com/apache/flink/tree/master/flink-formats/flink-json) module already.
> 
> Does that approach currently not work?
> 
> @twalthr or @dawidwys Could you share some details here?

yes , I dev also use [flink-json module](https://github.com/apache/flink/blob/c1b8f6757cee56ee9744f0533d2ec0118ef82801/flink-connectors/flink-connector-json/pom.xml#L77) like [JsonRowDeserializationSchema](https://github.com/apache/flink/blob/c1b8f6757cee56ee9744f0533d2ec0118ef82801/flink-formats/flink-json/src/main/java/org/apache/flink/formats/json/JsonRowDeserializationSchema.java) in JsonRowInputFormat.

But [org.apache.flink.formats.json.JsonRowFormatFactory](https://github.com/apache/flink/blob/c1b8f6757cee56ee9744f0533d2ec0118ef82801/flink-formats/flink-json/src/main/resources/META-INF/services/org.apache.flink.table.factories.TableFactory) not support File System Connector [ISSUE #11982](https://issues.apache.org/jira/browse/FLINK-11982), so dev JsonBatchTableSourceFactory  and JsonBatchTableSource. 

This implementation has been validated and run for a while in our flink production environment. thanks.

dawidwys:
Hi @ambition119 
I agree with @StephanEwen that this PR mixes the concepts of `connector`(file) and `format`(json).

I don't necessarily understand the comment that `JsonRowFormatFactory` does not support File System Connector. First, there is no file system connector, and second of all this is a format factory thus it should have no notion of connector.

I think you could rework your `JsonBatchTableSource` and `JsonRowInputFormat` to accept any `DeserializationSchema<Row>` similar to how `org.apache.flink.streaming.connectors.kafka.KafkaTableSourceBase` works.

Another issue on a more conceptual layer is that the proposed `RowInputFormat` assumes that each line in a file is a separate record. I agree this is probably the most common case, but not the only one. 
Files can be also written with a different layout (e.g. Parquet), thus we should differentiate that also on the connector level.

I'm guessing you might have been inspired by `CsvTableSource`, but it was one of the first `TableSource`s that was written before we decided to split connectors and formats and its design is flawed. It also uses `RowCsvInputFormat` that as I said in the previous paragraph applies custom block(in this case line) splitting based on configurable delimiter.

ambition119:
> Hi @ambition119
> I agree with @StephanEwen that this PR mixes the concepts of `connector`(file) and `format`(json).
> 
> I don't necessarily understand the comment that `JsonRowFormatFactory` does not support File System Connector. First, there is no file system connector, and second of all this is a format factory thus it should have no notion of connector.
> 
> I think you could rework your `JsonBatchTableSource` and `JsonRowInputFormat` to accept any `DeserializationSchema<Row>` similar to how `org.apache.flink.streaming.connectors.kafka.KafkaTableSourceBase` works.
> 
> Another issue on a more conceptual layer is that the proposed `RowInputFormat` assumes that each line in a file is a separate record. I agree this is probably the most common case, but not the only one.
> Files can be also written with a different layout (e.g. Parquet), thus we should differentiate that also on the connector level.
> 
> I'm guessing you might have been inspired by `CsvTableSource`, but it was one of the first `TableSource`s that was written before we decided to split connectors and formats and its design is flawed. It also uses `RowCsvInputFormat` that as I said in the previous paragraph applies custom block(in this case line) splitting based on configurable delimiter.

Thank you for your reply and discussion. 

First of all, the naming of this PR is a bit confusing. The code function mainly supports json files, such as local or hdfs, so you can rename it to JsonFileXxx.

Second, flink does exist File System Connector, the official document is
https://ci.apache.org/projects/flink/flink-docs-release-1.8/dev/table/connect.html#file-system-connector, provide [CsvBatchTableSourceFactory](https://github.com/apache/flink/blob/master/flink-table/flink-table-planner/src/main/resources/META-INF/services/org.apache.flink.table.factories.TableFactory#L16)  to support reading of csv files.
image

unit test [FileSystem connector CSV Format](https://github.com/apache/flink/blob/master/flink-table/flink-table-planner/src/test/scala/org/apache/flink/table/descriptors/TableDescriptorTest.scala#L59)
add example:
```java
public static void main(String[] args) throws Exception {
		ExecutionEnvironment env = ExecutionEnvironment.getExecutionEnvironment();
		BatchTableEnvironment tEnv = BatchTableEnvironment.create(env);
		FileSystem connector = new FileSystem().path("/path/to/csv/in.csv");

		OldCsv oldCsv = new OldCsv()
			.field("a", Types.STRING())
			.field("b", Types.INT())
			.field("c", Types.BOOLEAN());
		Schema schema = new Schema()
			.field("a", Types.STRING())
			.field("b", Types.INT())
			.field("c", Types.BOOLEAN());

		tEnv.connect(connector)
			.withFormat(oldCsv)
			.withSchema(schema)
			.registerTableSource("csvTable");

		String fullSql = "select * FROM csvTable";
		Table fullTable = tEnv.sqlQuery(fullSql);

		DataSet<Row> fullDS = tEnv.toDataSet(fullTable, Row.class);
		List<Row> result = fullDS.collect();

		System.out.println(result);
}
```

Third, for example, the hive support json file also is one line  a separate record, otherwise it is not supported. eg:
```json
{
	"fruit": [{
		"weight": 8,
		"type": "apple"
	}, {
		"weight": 9,
		"type": "pear"
	}],
	"author": "J. R. R. Tolkien",
	"title": "The Lord of the Rings",
	"category": "fiction",
	"price": 22.99,
	"isbn": "0-395-19395-8",
	"email": "amy@only_for_json_udf_test.net",
	"owner": "amy",
	"zip_code": "94025",
	"fb_id": "1234"
}
```
thanks.

dawidwys:
Ad. 1 Once again we aim to separate connector and formats => JsonFile = Json + File
Ad. 2  It provides `FileSystem` descriptor not connector. See my comments about `CsvTableSource`. It's design is broken and we do not want to repeat this error
Ad. 3 Good to know, hive is not the source of truth though. Unfortunately we cannot accept general purpose code that works in just a single very narrow setup.

ambition119:
> Ad. 1 Once again we aim to separate connector and formats => JsonFile = Json + File
> Ad. 2 It provides `FileSystem` descriptor not connector. See my comments about `CsvTableSource`. It's design is broken and we do not want to repeat this error
> Ad. 3 Good to know, hive is not the source of truth though. Unfortunately we cannot accept general purpose code that works in just a single very narrow setup.

@StephanEwen @dawidwys  Thank you for your suggestions and responses.  I am developing based on flink 1.6.x, and now I understand what you are talking about, so I will close this PR.

PR #8388, opened 2019-05-09
[FLINK-12164][runtime] Harden JobMasterTest against timeouts

zentol:
Hardens `JobMasterTest#testJobFailureWhenTaskExecutorHeartbeatTimeout`

The reported failure occurred when offering slots, indicating that at this time the TM was no longer registered at the JM. However, the TM is being registered right before the slot offer.

The only explanation I could find is that due so some freak timing thing a heartbeat times out in between these 2 calls. `testJobFailureWhenTaskExecutorHeartbeatTimeout` uses a very small heartbeat interval (1ms) and timeout (5ms).
I was only able to reproduce the issue locally after reducing the timeout to 2ms.

This PR doubles the timeout to reduce the likely-hood of this happening again.


StephanEwen:
Does this have any implication on test run times?

zentol:
I could not determine any significant increase in test times.

Tests that do not rely on timeouts are unaffected, and only few tests do not fall into this category.

PR #8389, opened 2019-05-09
[FLINK-12399][table] Fix FilterableTableSource does not change after applyPredicate

walterddr:

This PR fixes the problem: FilterableTableSource does not generate a new digest after the predicate push down, unless `explainSource()` API is explicitly override. 

## Brief change log

  - Changed the `explainTerm` API from `FlinkLogicalTableSourceScan` to include auxiliary source description.
  - Changed the `TestFilterableTableSource` to include both the none explainSource and with explainSource version.
  - Added test to both stream and batch cases.

## Verifying this change

  - Tests for both stream and batch cases for predicate push down without explainSource override.
  - Others are covered for current tests.

  - The serializers: no

## Documentation

  - If yes, how is the feature documented? n/a

### DIFF
--- a/flink-table/flink-table-planner/src/main/scala/org/apache/flink/table/plan/nodes/logical/FlinkLogicalTableSourceScan.scala
+++ b/flink-table/flink-table-planner/src/main/scala/org/apache/flink/table/plan/nodes/logical/FlinkLogicalTableSourceScan.scala
@@ -29,7 +29,7 @@ import org.apache.flink.table.api.TableException
 import org.apache.flink.table.calcite.FlinkTypeFactory
 import org.apache.flink.table.plan.nodes.FlinkConventions
 import org.apache.flink.table.plan.schema.TableSourceSinkTable
-import org.apache.flink.table.sources.{FilterableTableSource, TableSource, TableSourceUtil}
+import org.apache.flink.table.sources.{FilterableTableSource, ProjectableTableSource, TableSource, TableSourceUtil}
 
 import scala.collection.JavaConverters._
 
@@ -81,7 +81,13 @@ class FlinkLogicalTableSourceScan(
     val terms = super.explainTerms(pw)
         .item("fields", tableSource.getTableSchema.getFieldNames.mkString(", "))
 
-    val sourceDesc = tableSource.explainSource()
+    val auxiliarySourceDesc = tableSource match {
+      case fts: FilterableTableSource[_] =>
+        s"FilterPushDown=${fts.isFilterPushedDown.toString}"
+      case _ => ""
+    }
+
+    val sourceDesc = tableSource.explainSource() + auxiliarySourceDesc
     if (sourceDesc.nonEmpty) {
       terms.item("source", sourceDesc)
     } else {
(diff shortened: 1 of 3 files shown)


KurtYoung:
@flinkbot attention @godfreyhe

KurtYoung:
@flinkbot attention @beyond1920

beyond1920:
@walterddr good catch, we also encountered the problem in Blink before. 
The key question is that the explainSource value of new TableSource should be different with the original one after partition prune or filter push down or query push done, however the method name (explainSource) cannot indicate the meaning explicitly. 
The solution to pattern match FilterableTableSource in `FlinkLogicalTableSourceScan` is not perfect, since it ignores `PartitionableTableSource`,  `AggregatePushDownTableSource` and so on.
Maybe it's better to refactor current API `explainSource` in `TableSource` to explicitly indicates the value should be an identifier for this `Tablesource`, the identifier must be unique. That meanings, after project push down or filter push down or query pushdown, the identifier value of new TableSource
should be different with the original one.

walterddr:
Thanks for the review @beyond1920 . Yes I agree that the solution is mostly a bare minimum to fix the bug at this moment. I was actually trying to re-architecture the way how `TableSource` interfaces works, since the `explainSource` extends from `TableSource` it doesn't actually contain any information of whether this is a predicate push-down source type. 

However changing the API might take some time for designing. I will start a design doc and attached to the JIRA discussion, what do you think?

flink-table/flink-table-planner/src/main/scala/org/apache/flink/table/plan/nodes/logical/FlinkLogicalTableSourceScan.scala
```diff
@@ -81,7 +81,13 @@ class FlinkLogicalTableSourceScan(
     val terms = super.explainTerms(pw)
         .item("fields", tableSource.getTableSchema.getFieldNames.mkString(", "))
 
-    val sourceDesc = tableSource.explainSource()
+    val auxiliarySourceDesc = tableSource match {
+      case fts: FilterableTableSource[_] =>
+        s"FilterPushDown=${fts.isFilterPushedDown.toString}"
```
godfreyhe: `s"FilterPushDown=${fts.isFilterPushedDown.toString}"` is not the digest of a `FilterableTableSource `,  consider the following case: there are several different filters to be pushed down. 
so, i think the digest should be the string value of the pushed filters.

godfreyhe:
@walterddr thanks for the fix.  It's a little difficult for users to understand `explainSource` and implement it, even though we rename `explainSource` to `getSourceDigest`(or other names). could we keep the logic about `re-compute digest after push down` in rules ?

flink-table/flink-table-planner/src/main/scala/org/apache/flink/table/plan/nodes/logical/FlinkLogicalTableSourceScan.scala
walterddr: hmm. I think this is valid. I original think since there could only be one "match" and "onMatch", all the different filters either can be pushed down all cannot (e.g. always the same set of filter being push down).

but I think yes, there's nothing preventing calcite to choose to partially push down (not partially possible, but partially push down some of the filter because of volcano planner thinks it is more efficient than another). I doubt this is possible at this moment.

walterddr:
@godfreyhe yeah you are right. this is not exactly "explain source", ideally speaking we need to put "what filter has been pushed down" in the `explainSource` result since that actually "explain" the data source. 

I will draft a design to change the API in the JIRA. how about we move the discussion there? I will close this PR if we all agree.

walterddr:
close based on comments, will open a new one @godfreyhe @beyond1920

PR #8390, opened 2019-05-09
[FLINK-12469][table] Clean up catalog API on default/current database

xuefuz:

This PR separates concepts of "current database of a user session" and "default database of a catalog". 

## Brief change log

*(for example:)*
  - Removed get/setCurrentDatabase() from Catalog API
  - Added getDefaultDatabase() in Catalog API
  - Provided implementations in sub classes

## Verifying this change

This change is already covered by existing tests

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
+++ b/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
@@ -48,23 +48,26 @@ import static org.apache.flink.util.Preconditions.checkNotNull;
  */
 public abstract class HiveCatalogBase implements Catalog {
 	private static final Logger LOG = LoggerFactory.getLogger(HiveCatalogBase.class);
-
-	public static final String DEFAULT_DB = "default";
+	private static final String DEFAULT_DB = "default";
 
 	protected final String catalogName;
 	protected final HiveConf hiveConf;
 
-	protected String currentDatabase = DEFAULT_DB;
+	private final String defaultDatabase;
 	protected IMetaStoreClient client;
 
 	public HiveCatalogBase(String catalogName, String hivemetastoreURI) {
-		this(catalogName, getHiveConf(hivemetastoreURI));
+		this(catalogName, DEFAULT_DB, getHiveConf(hivemetastoreURI));
 	}
 
 	public HiveCatalogBase(String catalogName, HiveConf hiveConf) {
+		this(catalogName, DEFAULT_DB, hiveConf);
+	}
+
+	public HiveCatalogBase(String catalogName, String defaultDatabase, HiveConf hiveConf) {
 		checkArgument(!StringUtils.isNullOrWhitespaceOnly(catalogName), "catalogName cannot be null or empty");
 		this.catalogName = catalogName;
-
+		this.defaultDatabase = checkNotNull(defaultDatabase, "defaultDatabase cannot be null");
 		this.hiveConf = checkNotNull(hiveConf, "hiveConf cannot be null");
 	}
 
@@ -109,19 +112,8 @@ public abstract class HiveCatalogBase implements Catalog {
 	// ------ databases ------
 
 	@Override
-	public String getCurrentDatabase() throws CatalogException {
-		return currentDatabase;
-	}
-
-	@Override
-	public void setCurrentDatabase(String databaseName) throws DatabaseNotExistException, CatalogException {
-		checkArgument(!StringUtils.isNullOrWhitespaceOnly(databaseName));
-
-		if (!databaseExists(databaseName)) {
-			throw new DatabaseNotExistException(catalogName, databaseName);
-		}
-
-		currentDatabase = databaseName;
+	public String getDefaultDatabase() throws CatalogException {
+		return defaultDatabase;
 	}
 
 	@Override
--- a/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
+++ b/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
@@ -99,6 +99,11 @@ public abstract class HiveCatalogBase implements Catalog {
 			client = getMetastoreClient(hiveConf);
 			LOG.info("Connected to Hive metastore");
 		}
+
+		if (!databaseExists(defaultDatabase)) {
+			throw new CatalogException(String.format("Configured default database %s doesn't exist in catalog %s.",
+				defaultDatabase, catalogName));
+		}
 	}
 
 	@Override
--- a/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
+++ b/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
@@ -66,6 +66,7 @@ public abstract class HiveCatalogBase implements Catalog {
 
 	public HiveCatalogBase(String catalogName, String defaultDatabase, HiveConf hiveConf) {
 		checkArgument(!StringUtils.isNullOrWhitespaceOnly(catalogName), "catalogName cannot be null or empty");
+		checkArgument(!StringUtils.isNullOrWhitespaceOnly(defaultDatabase), "defaultDatabase cannot be null or empty");
 		this.catalogName = catalogName;
 		this.defaultDatabase = checkNotNull(defaultDatabase, "defaultDatabase cannot be null");
 		this.hiveConf = checkNotNull(hiveConf, "hiveConf cannot be null");
--- a/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
+++ b/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
@@ -68,7 +68,7 @@ public abstract class HiveCatalogBase implements Catalog {
 		checkArgument(!StringUtils.isNullOrWhitespaceOnly(catalogName), "catalogName cannot be null or empty");
 		checkArgument(!StringUtils.isNullOrWhitespaceOnly(defaultDatabase), "defaultDatabase cannot be null or empty");
 		this.catalogName = catalogName;
-		this.defaultDatabase = checkNotNull(defaultDatabase, "defaultDatabase cannot be null");
+		this.defaultDatabase = defaultDatabase;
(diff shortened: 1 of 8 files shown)


flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
```diff
@@ -48,23 +48,26 @@
...
 
 	public HiveCatalogBase(String catalogName, HiveConf hiveConf) {
+		this(catalogName, DEFAULT_DB, hiveConf);
+	}
+
+	public HiveCatalogBase(String catalogName, String defaultDatabase, HiveConf hiveConf) {
 		checkArgument(!StringUtils.isNullOrWhitespaceOnly(catalogName), "catalogName cannot be null or empty");
 		this.catalogName = catalogName;
-
+		this.defaultDatabase = checkNotNull(defaultDatabase, "defaultDatabase cannot be null");
```
bowenli86: can defaultDatabase be empty string?

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/GenericInMemoryCatalog.java
```diff
@@ -65,9 +64,14 @@
...
 
 	public GenericInMemoryCatalog(String name) {
+		this(name, DEFAULT_DB);
+	}
+
+	public GenericInMemoryCatalog(String name, String defaultDatabase) {
 		checkArgument(!StringUtils.isNullOrWhitespaceOnly(name), "name cannot be null or empty");
 
 		this.catalogName = name;
+		this.defaultDatabase = defaultDatabase;
```
bowenli86: check whether defaultDatabase is null or empty string?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
xuefuz: I think it cannot be either null or empty string. I will add additional checks.

xuefuz:
> Thanks @xuefuz for the PR.
> 
> Currently this PR doesn't seem to verify whether user-set "defaultDatabase" in catalog impls' constructors actually exists or not upon initialization. Do we need to ensure it actually exists? if so, how? I have something in mind but want to listen to your ideas first.

default DB is most likely from configuration and it cannot be verified until a catalog is opened. This shouldn't be a problem as error will be reported later when the db is actually referenced.

bowenli86:
> default DB is most likely from configuration and it cannot be verified until a catalog is opened. This shouldn't be a problem as error will be reported later when the db is actually referenced.

I wonder if we should verify its existence upon catalog opening and fail early if the db doesn't exist? The reason being that the default db cannot be changed at runtime. If SQL CLI users realize it's nonexistent in the later stage of their workflow, e.g. the default db name they put has a typo, they may have several options: 1) not use any feature related to default db, 2) reconfig and restart their clients and lost all their previous progress in memory, 3) create that db (not a good option for a typo name). None seems very user-friendly.

xuefuz:
> > default DB is most likely from configuration and it cannot be verified until a catalog is opened. This shouldn't be a problem as error will be reported later when the db is actually referenced.
> 
> I wonder if we should verify its existence upon catalog opening and fail early if the db doesn't exist? The reason being that the default db cannot be changed at runtime. If SQL CLI users realize it's nonexistent in the later stage of their workflow, e.g. the default db name they put has a typo, they may have several options: 1) not use any feature related to default db, 2) reconfig and restart their clients and lost all their previous progress in memory, 3) create that db (not a good option for a typo name). None seems very user-friendly.

Okay. Sounds good.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/GenericInMemoryCatalog.java
```diff
@@ -65,9 +64,15 @@
 	private final Map<ObjectPath, Map<CatalogPartitionSpec, CatalogColumnStatistics>> partitionColumnStats;
 
 	public GenericInMemoryCatalog(String name) {
+		this(name, DEFAULT_DB);
+	}
+
+	public GenericInMemoryCatalog(String name, String defaultDatabase) {
```
dawidwys: How about `GenericInMemoryCatalog(String name, String defaultDatabaseName, CatalogDatabase defaultDatabase)`? I guess we don't need to register additional database if user wants to set a different one.

I think we don't need to check for db existence in the `open` method if we do that.

xuefuz: Hi @dawidwys Thanks for your review. To clarify, do you mean we don't want user to set a  default database other than DEFAULT_DB? I'm fine with this because there is the only one it has when the catalog is initialized.

Even for GenericInMemoryCatlog, it confirms the catalog API, allowing user to create different databases.

dawidwys: No, what I mean is to allow user to provide the defaultDB. In other words to have a ctor that does not create the defaultDb, but uses the one provided by the user.

So the ctor would like:
```
	public GenericInMemoryCatalog(String name, String defaultDbName, CatalogDatabase defaultDb) {
		checkArgument(!StringUtils.isNullOrWhitespaceOnly(name), "name cannot be null or empty");

		this.currentDatabase = defaultDbName;
		this.catalogName = name;
		this.databases = new LinkedHashMap<>();
		this.databases.put(defaultDbName, defaultDb);
		this.tables = new LinkedHashMap<>();
		this.functions = new LinkedHashMap<>();
		this.partitions = new LinkedHashMap<>();
		this.tableStats = new LinkedHashMap<>();
		this.tableColumnStats = new LinkedHashMap<>();
		this.partitionStats = new LinkedHashMap<>();
		this.partitionColumnStats = new LinkedHashMap<>();
	}
```

xuefuz: Cool. Got it. Will update. Thanks.

bowenli86: +1, I like Xuefu's approach as it conforms more to yaml config.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
```diff
@@ -48,23 +48,27 @@
...
 	public HiveCatalogBase(String catalogName, HiveConf hiveConf) {
+		this(catalogName, DEFAULT_DB, hiveConf);
+	}
+
+	public HiveCatalogBase(String catalogName, String defaultDatabase, HiveConf hiveConf) {
 		checkArgument(!StringUtils.isNullOrWhitespaceOnly(catalogName), "catalogName cannot be null or empty");
+		checkArgument(!StringUtils.isNullOrWhitespaceOnly(defaultDatabase), "defaultDatabase cannot be null or empty");
 		this.catalogName = catalogName;
-
+		this.defaultDatabase = checkNotNull(defaultDatabase, "defaultDatabase cannot be null");
```
bowenli86: Since we already have `!StringUtils.isNullOrWhitespaceOnly(defaultDatabase)` above, shall we remove the `checkNotNull`?

xuefuz: Adopted suggestion.

bowenli86:
+1

@dawidwys  can you take another look? if you don't have other comments, I'll merge this once CI is green

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/GenericInMemoryCatalog.java
dawidwys: Hmm, not sure if it conforms to yaml more. How do you define database properties in the yaml file? How do you set the properties of the default db in yaml?

I could imagine sth like:
```
catalogs:
  - name: ABC
    type: in-memory
    property-version: 1
    default-database: 
        name: mydb
        properties:
            prop1: "prop1"
            ....
```

bowenli86: I'm not in favor of that for a couple reasons. 1) Currently yaml file is misused as some kind of "file-based catalog" where users can define metadata like tables, views, etc. The effort of unified catalog API is to correct that misuse by moving metadata out and only keep configuration values in yaml. The above proposed format in which allows users to define database would lead us back to the misuse case. 2)  in-memory catalog is mainly used for back-compatibility and when users want to experiment with Flink. We don't expect/recommend users to set an in-memory catalog in yaml file besides the default one. 

To take a step back, I think we can clearly define that users can only set default db for a persistent catalog, and not in-memory catalog, as no db except the "build-in" db will be existent upon initialization of in-memory catalog. What do you think?

xuefuz:
@bowenli86 Can we move this forward? The YAML discussion should have its own thread. Thanks.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/GenericInMemoryCatalog.java
dawidwys: I agree we should not get pulled into the discussion around yaml file.

I still believe thought from the clean code perspective we should allow for passing a CatalogDatabase (or GenericInMemoryCatalogDatabse if we want to limit the scope) in the ctor. Such approach helps to write a testable code(http://misko.hevery.com/attachments/Guide-Writing%20Testable%20Code.pdf or  http://misko.hevery.com/code-reviewers-guide/) (which in turn is easier to use, as tests also verify that the interfaces are usable). I see no reason why we should not allow for that. 

But in the end I will leave it for your call.

bowenli86:
merging...

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/GenericInMemoryCatalog.java
xuefuz: Since we will have PR coming for YAML file structure for catalogs, let's defer the discussion until we see it. Now as far as I am concerned, I don't see a use case for a user to create a default DB before creating a catalog. It's also common for a db product to create default db automatically. If user needs there specific databases, they can create their own (possibly via DDL). Our in-memory catalog is following the suite except we allow user specific a name other than "default" with the change here. This addition seems to me even optional. Nevertheless, if we find it necessary that taking a database created by user to instantiate the in-memory DB, we can add it at that time.

PR #8396, opened 2019-05-10
[FLINK-12468][yarn] Unregister application from the YARN Resource Man…

jiasheng55:
…ager with a valid appTrackingUrl

Currently when a Flink job on Yarn finished, it's tracking URL is not set. As a result, we can not jump to Flink history server directly from Yarn.
This PR aims to provide a valid "Tracking URL".

## Brief change log

Add a new configuration to set the "Tracking URL" of a finished Yarn application with the address of a Flink history server.

## Verifying this change

All the existing integration tests.

  - The serializers: (no)

## Documentation

  - If yes, how is the feature documented? (**docs**)

### DIFF
--- a/flink-yarn/src/main/java/org/apache/flink/yarn/configuration/YarnConfigOptions.java
+++ b/flink-yarn/src/main/java/org/apache/flink/yarn/configuration/YarnConfigOptions.java
@@ -167,15 +167,6 @@ public class YarnConfigOptions {
 		.defaultValue("")
 		.withDescription("A comma-separated list of tags to apply to the Flink YARN application.");
 
-	/**
-	 * The address of Flink HistoryServer.
-	 * This configuration parameter allows setting the appTrackingUrl for finished YARN applications.
-	 */
-	public static final ConfigOption<String> APPLICATION_HISTORY_SERVER_ADDRESS =
-		key("yarn.history.server.address")
-		.defaultValue("")
-		.withDescription("The address of Flink HistoryServer.");
-
 	// ------------------------------------------------------------------------
 
 	/** This class is not meant to be instantiated. */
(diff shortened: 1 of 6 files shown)


flink-yarn/src/main/java/org/apache/flink/yarn/configuration/YarnConfigOptions.java
```diff
@@ -167,6 +167,15 @@
 		.defaultValue("")
 		.withDescription("A comma-separated list of tags to apply to the Flink YARN application.");
 
+	/**
+	 * The address of Flink HistoryServer.
+	 * This configuration parameter allows setting the appTrackingUrl for finished YARN applications.
+	 */
+	public static final ConfigOption<String> APPLICATION_HISTORY_SERVER_ADDRESS =
```
tillrohrmann: Should we rather use `HistoryServerOptions#HISTORY_SERVER_WEB_ADDRESS`?

jiasheng55: I have thought about that, but `HistoryServerOptions#HISTORY_SERVER_WEB_ADDRESS` is a hostname, what if we want to set a domain name instead?

_http://host:port vs http://domain-name_

jiasheng55: @tillrohrmann Any suggestion? Should we stick to `HistoryServer#HISTORY_SERVER_WEB_ADDRESS` ?

tillrohrmann: I think we could start with `hostname:port` and extend it if we see that there is a problem with that. At the moment people would need to connect to the `HistoryServer` that way anyway.

tillrohrmann: I'll apply the change and merge the PR.

PR #8397, opened 2019-05-10
[FLINK-11421][Table SQL/Runtime]Add compilation options to allow comp…

liyafan82:
…iling generated code with JDK compiler

By default, Flink compiles the generated source code by Janino. This is fast, as the compilation often finishes in hundreds of milliseconds. The generated Java Byte Code, however, is of poor quality. To illustrate, we use Java Compiler API (JCA) to compile the generated code. Experiments on TPC-H (1 TB) queries show that the E2E time can be more than 10% shorter, when operators are compiled by JCA, despite that it takes more time (a few seconds) to compile with JCA.

Therefore, in this PR we introduce another option to compile generated source code - compiling by JCA. 

## Brief change log

  - We introduce class JCACompilationUtil as a utility to compile generated code by JCA.
  - In JCACompilationUtil, we provide a class cache to reuse compiled classes, so as to improve the performance. 
  - We introduce a global configuration, compiler.code-gen.compilation.option, to allow the users to configure the compilation option. The default configuration option is compiling by Janino, which is the same as before. 

## Verifying this change

  - Class JCACompilationUtilTest contains a number of test cases for the basic functionalities of utility JCACompilationUtil.
  - We extend org/apache/flink/table/runtime/batch/sql/join/JoinITCase.scala class as parameterized unit tests, so we can reuse existing test cases to test the functionalities of this PR.
  
  - The serializers: (no)

## Documentation


liyafan82:
Hi @JingsongLi , would you please help take a look at this PR?

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/generated/CompilationOption.java
```diff
@@ -0,0 +1,45 @@
...
+
+	/**
+	 * Compiling Java code by Janino.
+	 * The compilation is fast, but the generated binary code is of low quality.
+	 */
+	JANINO,
+
+	/**
+	 * Compiling Java code by Java Compiler API (JCA)
+	 * The compilation is slow, but the generated binary code is of high quality.
```
Xpray: Are there any proofs that JCA is always better than janino, I think for long running jobs, JIT is the key for performance instead of fresh generated code. And more tests may need for this conclusion.

liyafan82: @Xpray , thanks a lot for your comments.
We have some proofs that JCA can be better than Janino. For example, when trying to vectorize Flink, we found that by using JCA, the performance of TPC-H Q1 (1 TB) can be improved by over 20% (27-28s to 20s). Similar performance improvements are also witness for other queries like Q12 and Q18.

However, we cannot claim that JCA always outperforms Janino. Because JCA takes longer time for compiling, the performance of JCA can even be worse in some cases. 

Therefore, the meaning of this issue is not to give users a guarantee of better performance. Instead, we give the users a new choice, which may help them in their specific scenarios.

Xpray: If that is the case, I think these comments ' The compilation is slow, but the generated binary code is of high quality' should be removed.

liyafan82: Removed. Thanks for your suggestion.

liyafan82:
Hi @KurtYoung , would you please help review this PR? It also fix a bug in the code-gen logic.

JingsongLi:
I think you should add some detailed benchmarks, as we discussed in JIRA.

KurtYoung:
@liyafan82 We will also try to verify the conclusions you provided in our benchmark environment, will let you know whether the results align with yours.

liyafan82:
Hi @JingsongLi, thank you so much for your comments. I will provide the benchmark results, as we have discussed.
The bug refers to OperatorCodeGenrator#generateTwoInputStreamOperator method. There should be a parameter with type TableConfig, so as to be align with generateOneInputStreamOperator. 
I am not sure if you have left it out intentionally, or forgotten about it.

Hi @KurtYoung , thanks a lot for your attention, and thank you for your effort in verifying our results.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/codegen/OperatorCodeGenerator.scala
```diff
@@ -130,6 +133,7 @@ object OperatorCodeGenerator extends Logging {
       endInputCode2: String,
       input1Type: InternalType,
       input2Type: InternalType,
+      config: TableConfig,
```
JingsongLi: It is not a bug... You can get config from `CodeGeneratorContext.tableConfig`.
`generateOneInputStreamOperator` have `config` parameter is redundant.

liyafan82: I see. Thanks for the clarification. So the config parameter in generateOneInputStreamOperator is also redundant, right?

liyafan82:
Hi @KurtYoung @JingsongLi , I have evaluated some benchmarks, and the results are as follows. Please give your valuable feedback. 

To evaluate the effects of the compilation options, we use TPC-H Q1 as the benchmark, because it is a relatively simple query, which makes it easier for demonstration. The evaluations are based on our vectorized code, because it is more notable on the vectorized code. 

The execution plan of Q1 is as follows:
image

It can be seen that the following operators are based on code-gen: SourceConversion (ID = 4), Calc (ID = 5), LongHashAggregate (ID = 6), GlobalHashAggregate (ID = 8), Calc (ID = 11), SinkConversion (ID = 12).

However, we exclude some operators from our analysis:
  1) Operators SourceConversion (ID = 4) and SinkConversion (ID = 12) are excluded, because their code is quite simple.
  2) Operators GlobalHashAggregate (ID = 8), Calc (ID = 11) are excluded, because their input is too small (only 4 records), and processing them takes very short time (< 1ms)

Therefore, we only consider Calc (ID = 5), LongHashAggregate (ID = 6) in our analysis.

The table below shows the average time (in ms) for each operator for processing Q1 (1TB) in our cluster:

Operator\Compiler | JCA | Janino
-- | -- | --
Calc (ID = 5) | 3858.07 | 7146.9
LongHashAggregate (ID = 6) | 2493.75 | 3923.17

It can be seen that the code compiled by JCA runs much faster. 

The following table shows the compilation time (in ms) with different compilers. Note that compilation times for other operators are also given for reference. 

Operator\Compiler | JCA | Janino
-- | -- | --
SourceConversion (ID = 4) | 92 | 4
Calc (ID = 5) | 148 | 15
LongHashAggregate (ID = 6) | 854 | 44
GlobalHashAggregate (ID = 8) | 200 | 58
Calc (ID = 11) | 117 | 10
SinkConversion (ID = 12) | 99 | 4

It can be seem that the compilation time for JCA is one magnitude large than the compilation time for Janino. To hide such latency by two techniques:
1)	By using a class cache, tasks in the same JVM can reuse the compilation results.
2)	The time for compiling a batch is much smaller than the sum of compiling classes individually. 

The latter is not implemented in the current PR, because it involves changes to other parts of the code based. We will implement in another separate JIRA. 

To investigate why the code produced by JCA is better, we dumped all class files, and it can be easily seen that the file sizes are different:


Operator\Compiler | JCA | Janino
-- | -- | --
SourceConversion (ID = 4) | 2 KB | 2 KB
Calc (ID = 5) | 6 KB | 5 KB
LongHashAggregate (ID = 6) | 13 KB | 9 KB
GlobalHashAggregate (ID = 8) | 18 KB | 12 KB
Calc (ID = 11) | 7 KB | 5 KB
SinkConversion (ID = 12) | 3 KB | 2 KB

By analyzing the byte code, we found there are slight differences in the code structure. For example, the following figure shows the bytecode of the main processing method for Calc (ID = 5):

image

JingsongLi:
Hi @liyafan82 Thank for your nice works~ I think I have a clearer understanding now.
I think your benchmarks illustrates two points:
1. JCA is much better than Janino in vector computation. (150% - 200%)
2. JCA compilation time is about 0.1 second - 1 second slower.

But we probably don't have any vector computing right now. Can you continue to benchmark how our table works right now? (It would be better if you could provide reproducible code)

liyafan82:
> Hi @liyafan82 Thank for your nice works~ I think I have a clearer understanding now.
> I think your benchmarks illustrates two points:
> 
> 1. JCA is much better than Janino in vector computation. (150% - 200%)
> 2. JCA compilation time is about 0.1 second - 1 second slower.
> 
> But we probably don't have any vector computing right now. Can you continue to benchmark how our table works right now? (It would be better if you could provide reproducible code)

@JingsongLi , thanks for your comments. 
Sure. I will provide benchmark results on table workloads.

liyafan82:
> Hi @liyafan82 Thank for your nice works~ I think I have a clearer understanding now.
> I think your benchmarks illustrates two points:
> 
> 1. JCA is much better than Janino in vector computation. (150% - 200%)
> 2. JCA compilation time is about 0.1 second - 1 second slower.
> 
> But we probably don't have any vector computing right now. Can you continue to benchmark how our table works right now? (It would be better if you could provide reproducible code)

Hi @JingsongLi and @KurtYoung , I have evaluated the benchmark on table work load. The results are as follows. Please give your valuable feedback:

We use the same benchmark (TPC-H Q1, 1TB) to evaluate the effects of compilation options. This time, we use the original Flink runtime engine, instead of vectorization. The effects are notable, but not as significant as with vectorization. 

Likewise, we only consider Calc (ID = 5), LongHashAggregate (ID = 6) in our analysis. 

The table below shows the average time (in ms) for each operator for processing Q1 in our cluster:


Operator\Compiler | JCA | Janino
-- | -- | --
Calc (ID = 5) | 4902.18 | 5482.86
LongHashAggregate (ID = 6) | 2967.2 | 3257.92

It can be seen that the code compiled by JCA runs about 9.8% faster. 

The following table shows the compilation time (in ms) with different compilers. The results are similar to the previous benchmark results. 


Operator\Compiler | JCA | Janino
-- | -- | --
Calc (ID = 5) | 124 | 12
LongHashAggregate (ID = 6) | 850 | 31
GlobalHashAggregate (ID = 8) | 225 | 100
Calc (ID = 11) | 105 | 14
SinkConversion (ID = 12) | 100 | 5

Investigations on compiled class files show that, different compiles produce class files with different sizes. 


Operator\Compiler | JCA | Janino
-- | -- | --
Calc (ID = 5) | 4 KB | 3 KB
LongHashAggregate (ID = 6) | 10 KB | 8 KB
GlobalHashAggregate (ID = 8) | 17 KB | 12 KB
Calc (ID = 11) | 4 KB | 3 KB
SinkConversion (ID = 12) | 2 KB | 2 KB

By analyzing the byte code, we found there are differences in the code structure of bytecodes. For example, the following figure shows the bytecode of the processElement method for Calc (ID = 5):

image

To make it more convenient to reproduce the results, we have attached the source code for generated operators. Just by compiling the code with different compilers and generating some test data set, the  above results should be reproduced locally (Our results are derived from the cluster). 

[code.zip](https://github.com/apache/flink/files/3185415/code.zip)

JingsongLi:
@liyafan82  The benchmarks results were exciting.
Can you use CodeGen's code to build some local benchmarks? You can put your code to your github branch. In this way, we can more easily reproduce.

liyafan82:
> @liyafan82 The benchmarks results were exciting.
> Can you use CodeGen's code to build some local benchmarks? You can put your code to your github branch. In this way, we can more easily reproduce.

Hi @JingsongLi and @KurtYoung , thank you for the good suggestion.

Evaluating local benchmarks can be time-consuming. To make the results reliable, a large amounts of data must be evaluated for many times times. In addition, a performance benchmark framework should be employed.

If you really insist on this, I'd recommend JMH, which is the de factor standard. However, that may involve substantial changes, including opening a separate project,  so I think we should do it in a separate JIRA. What do you think?

JingsongLi:
I think JMH benchmark component is a further thing. It's under discussion.
For this PR, I think you can do a simpler benchmark rather than a complicate component.
You can dependent https://github.com/dataArtisans/flink-benchmarks or other things, just let us reproduce your benchmark easier.
(you can also include testing for vector computation in this local benchmarks)

liyafan82:
> I think JMH benchmark component is a further thing. It's under discussion.
> For this PR, I think you can do a simpler benchmark rather than a complicate component.
> You can dependent https://github.com/dataArtisans/flink-benchmarks or other things, just let us reproduce your benchmark easier.
> (you can also include testing for vector computation in this local benchmarks)

@JingsongLi Thanks for your suggestion.
flink-benchmarks is based on JMH. To add a performance benchmark, we need a separate project, not simply adding a dependency. So we need to open a separate project (if you know a way to add an individual performance benchmark without opening a new project, please let me know), and that is substantial change, which deserves a separate JIRA.

liyafan82:
> I think JMH benchmark component is a further thing. It's under discussion.
> For this PR, I think you can do a simpler benchmark rather than a complicate component.
> You can dependent https://github.com/dataArtisans/flink-benchmarks or other things, just let us reproduce your benchmark easier.
> (you can also include testing for vector computation in this local benchmarks)

Hi @JingsongLi and @KurtYoung ,

I have tried to provide a local benchmark, as you suggested. Unfortunately, I could not reproduce the results in the cluster environment when the operators are processed in the jobs. 

So let me show you what I did, and what results I got.

I setup a JMH project (I believe this is a reliable way to conduct performance evaluations), and use randomly generated data to test the Calc operator method above, in a previous discussion. 

The results are as follows:
1) Compilation time:

Compilation time for JCA: 713ms
Compilation time for JANINO: 267ms

The compilation time for JCA is ok, but the time for Janino is not right. In the job environment, compilation by Janino never exceeds 100ms. This result is stably reproducible.

2) Processing time:

The general performance improvements of JCA is about 2% to 7%. A typical report is given below:

Benchmark                        Mode  Cnt       Score      Error  Units
CompilationBenchmark.testJCA     avgt    5   94611.913 ±  868.944  us/op
CompilationBenchmark.testJanino  avgt    5  100778.892 ± 1367.429  us/op

If you are interested, please use this [branch](https://github.com/liyafan82/flink/tree/fly_11421_test) to reproduce the results. 

Anyway, I do not think it is a good idea to reproduce the results in a local machine. This is because, it is extremely difficult to simulate the environment in the cluster when a job is running. The following factors will be different from the local environment:

1. In the job environment, the CPU cache may be polluted by other processes (like HDFS, Yarn), or threads from the process (like Task Manager, Job Manager, source threads, etc.)
2. In the job environment, there are many concurrent worker threads processing the same operator and accessing the same resources.
3. The JVM has been running for a while (it has been running since the session is started, or even the cluster is started), so the JIT can be in a different state from the local JIT. 

In my opinion, the performance in the job environment matters the most, because that is what we want ultimately. I guess you must have such experiences: a improvement well-tested locally does not perform as expected in the production environment, and vice versa. 

So I think what we have is sufficient to prove the merit of this issue. If you are still in doubt, I sincerely invite you to take a look in our environment.

liyafan82:
> > I think JMH benchmark component is a further thing. It's under discussion.
> > For this PR, I think you can do a simpler benchmark rather than a complicate component.
> > You can dependent https://github.com/dataArtisans/flink-benchmarks or other things, just let us reproduce your benchmark easier.
> > (you can also include testing for vector computation in this local benchmarks)
> 
> Hi @JingsongLi and @KurtYoung ,
> 
> I have tried to provide a local benchmark, as you suggested. Unfortunately, I could not reproduce the results in the cluster environment when the operators were run in jobs.
> 
> So let me show you what I did, and what results I got.
> 
> I setup a JMH project (I believe this is a reliable way to conduct performance evaluations), and use randomly generated data to test the Calc operator method above, in a previous discussion.
> 
> The results are as follows:
> 
> 1. Compilation time:
> 
> Compilation time for JCA: 713ms
> Compilation time for JANINO: 267ms
> 
> The compilation time for JCA is ok, but the time for Janino is not right. In the job environment, compilation by Janino never exceeds 100ms. This result is stably reproducible.
> 
> 1. Processing time:
> 
> The general performance improvements of JCA is about 2% to 7%. A typical report is given below:
> 
> Benchmark Mode Cnt Score Error Units
> CompilationBenchmark.testJCA avgt 5 94611.913 ± 868.944 us/op
> CompilationBenchmark.testJanino avgt 5 100778.892 ± 1367.429 us/op
> 
> If you are interested, please use this [branch](https://github.com/liyafan82/flink/tree/fly_11421_test) to reproduce the results.
> 
> Anyway, I do not think it is a good idea to reproduce the results in a local machine. This is because, it is extremely difficult to simulate the environment in the cluster when a job is running. The following factors will be different from the local environment:
> 
> 1. In the job environment, the CPU cache may be polluted by other processes (like HDFS, Yarn), or threads from the process (like Task Manager, Job Manager, source threads, etc.)
> 2. In the job environment, there are many concurrent worker threads processing the same operator and accessing the same resources.
> 3. The JVM has been running for a while (it has been running since the session is started, or even the cluster is started), so the JIT can be in a different state from the local JIT.
> 
> In my opinion, the performance in the job environment matters the most, because that is what we want ultimately. I guess you must have such experiences: a improvement well-tested locally does not perform as expected in the production environment, and vice versa.
> 
> So I think what we have is sufficient to prove the merit of this issue. If you are still in doubt, I sincerely invite you to take a look in our environment.

JingsongLi:
@liyafan82 Thanks your local benchmark.
I think it is important to try to reproduce the results in a local machine. As you say, cluster environment is very complicate. In that case, the test data may not tell where optimizations exist and where regression occurs.
The compiler optimizer is easy to reproduce on a single machine. Everything is mockable.

Can you add some other patterns? Like vector computation benchmark?

liyafan82:
> @liyafan82 Thanks your local benchmark.
> I think it is important to try to reproduce the results in a local machine. As you say, cluster environment is very complicate. In that case, the test data may not tell where optimizations exist and where regression occurs.
> The compiler optimizer is easy to reproduce on a single machine. Everything is mockable.
> 
> Can you add some other patterns? Like vector computation benchmark?

Hi @JingsongLi , I agree with you that the local performance benchmarks can prevent performance regressions. 

However, performance benchmark can be difficult for Java. Since local and job environments differ greatly, JIT may apply different optimizations in different times. To reproduce the results in job environment, the local environment must apply exactly the same optimizations enabled in the job environment, and the optimizations must be applied in the same timing.

Benchmark for vector computation can be a good suggestion. However, vector related code has not been merged into Flink, so how can we introduce the benchmarks based on Flink?

JingsongLi:
You can simplify the codes, maybe you just need a simple VectorBatch class.

KurtYoung:
From all the discussions above, seems we have two conclusions so far:
1. JCA only improves vectorized execution codes.
2. The improvement about vectorization using JCA cannot be reproduced with JMH and can only be observed in cluster environment. 

Am I right about this?

liyafan82:
> From all the discussions above, seems we have two conclusions so far:
> 
> 1. JCA only improves vectorized execution codes.
> 2. The improvement about vectorization using JCA cannot be reproduced with JMH and can only be observed in cluster environment.
> 
> Am I right about this?

Not exactly.

For processing time:
1. In cluster environment, the performance improvement is 30% - 40% for vectorization, and about 10% for row-store.
2. In local, the performance improvement is 2% - 7% for row-store, and the improvements for vectorization is unknown.

For compilation time:
1. In cluster environment, the compilation time for JCA is more than 10 times longer than that of Janino.
2. In local, the compilation time for JCA is about 3 times longer.

KurtYoung:
@liyafan82 Thanks for the clarification. As long as we don't have vectorization mode yet, I propose we focus on current row mode first. 
2% - 7% improvements looks not bad, but the compilation time is also a noticeable factor. 
BTW, have we evaluated this option in our benchmark environment @JingsongLi ? I think it will be a supplement to making the decision about whether it's worthy to have this compile option in current row execution mode.

JingsongLi:
@KurtYoung @liyafan82 
1.In cluster environment: I tried to test TPCDS 10T, but it turned out to be a bit regression. I think it is related to compile time.
2.In local: I test https://github.com/liyafan82/flink/tree/fly_11421_test branch, but after ten times test, the processing time results of JCA and Janino are basically the same. compilation time is same to @liyafan82  (VM version: JDK 1.8.0_102, OpenJDK 64-Bit Server VM, 25.102-b52)

KurtYoung:
Just in case if you may mislead by unreliable JMH test, here is a good article to read and follow: http://tutorials.jenkov.com/java-performance/jmh.html#writing-good-benchmarks

liyafan82:
@JingsongLi Thank you so much for taking time to verify our results.
For 1). I think it is generally expected. JCA takes 10x longer for compilation. We have some techniques to hide this time, but they are not implemented in this PR.
For 2). I think at least, your results reveal that the improvements are unstable. Even if you can reproduce the 2% - 7% performance improvements, such improvements are not significant enough to help improve the performance of row-store. So I think more investigations are required for us. 

@KurtYoung the article is impressive. Thank you so much for your suggestion.

Anyway, I would like to thank you so much for all your helpful suggestions and valuable comments. Let's work together to make row-store better, to make Flink better.

JingsongLi:
Hi @liyafan82 , Thank you very much for your research on compilation, because we have generally agreed that JCA is not very useful for the current FLINK.
Can you close this PR temporarily?

liyafan82:
> Hi @liyafan82 , Thank you very much for your research on compilation, because we have generally agreed that JCA is not very useful for the current FLINK.
> Can you close this PR temporarily?

Sure. Thanks for your kind reminder. I think we still need a way to prove the performance benefits of JCA.

PR #8398, opened 2019-05-10
[FLINK-10466][build] Add flink-dist dependency to yarn-test

zentol:
`flink-yarn-tests` makes use of `flink-dist`, but doesn't declare it as a dependency.


azagrebin:
@flinkbot approve all

PR #8402, opened 2019-05-10
[FLINK-12473][ml] Add the interface of ML pipeline and ML lib

c4emmmm:

This pull request introduces the major interfaces for ML pipeline and ML lib.

## Brief change log

  - Create flink-ml module under flink root
  - Add flink-ml-api module, and introduce the major API interfaces, including PipelineStage, Estimator, Transformer, Model, Pipeline, etc.

## Verifying this change

This change is pure interface design without any test coverage.

This PR only adds ML pipeline & ML lib interface. The implementations on these interface will be added in the next PR (please refer to “implementation plan” section of [FLIP-39 doc](https://docs.google.com/document/d/1StObo1DLp8iiy0rbukx8kwAJb0BwDZrQrMWub3DzsEo) where we will add the test cases. 

  - The serializers: (no)

## Documentation

  - If yes, how is the feature documented? 
    - JavaDocs 
    - Flink document (will submit via a separate PR)


flink-ml/flink-ml-api/src/main/java/org/apache/flink/ml/util/persist/MLStageFactory.java
```diff
@@ -0,0 +1,67 @@
...
+	public static <T extends Persistable> T createFromJson(String json) {
+		Gson gson = new Gson();
+		JsonObject jobj = gson.fromJson(json, JsonObject.class);
+		String className = jobj.get("stageClassName").getAsString();
+		if (className == null) {
+			throw new RuntimeException(
+					"Can not create a PipelineStage with a json without stageClassName signed");
+		}
+		try {
+			Persistable s = (Persistable) Class.forName(className).newInstance();
```
Xpray: it's better to supply a method which user can specify a {{ClassLoader}}, and use {{Class.forName(className, classLoader)}}, since class might only be found in other ClassLoader.

flink-ml/flink-ml-api/src/main/java/org/apache/flink/ml/api/core/Pipeline.java
```diff
@@ -0,0 +1,213 @@
...
+	public boolean needFit() {
+		return this.getIndexOfLastEstimator() >= 0;
+	}
+
+	public Params getParams() {
+		return params;
+	}
+
+	//find the last Estimator or Pipeline that needs fit in stages, -1 stand for no Estimator in Pipeline
+	private int getIndexOfLastEstimator() {
```
Xpray: travel in reverse order?

flink-ml/flink-ml-api/src/main/java/org/apache/flink/ml/api/misc/param/ParamInfo.java
```diff
@@ -0,0 +1,145 @@
...
+		this.description = description;
+		this.isOptional = isOptional;
+		this.hasDefaultValue = hasDefaultValue;
+		this.defaultValue = defaultValue;
+		this.validator = validator;
+		this.valueClass = valueClass;
+	}
+
+	/**
+	 * Returns the name of the parameter.
```
walterddr: might be good to explain `name` needs to be unique within a pipeline (since it is used for the paramMap as key)

flink-ml/flink-ml-api/src/main/java/org/apache/flink/ml/api/misc/param/Params.java
```diff
@@ -0,0 +1,145 @@
...
+	 * @param <V>  the type of the specific parameter
+	 * @return the value of the specific parameter, or default value defined in the {@code info} if
+	 * this Params doesn't contain the parameter
+	 * @throws RuntimeException if the Params doesn't contains the specific parameter, while the
+	 *                          param is not optional but has no default value in the {@code info}
+	 */
+	@SuppressWarnings("unchecked")
+	public <V> V get(ParamInfo<V> info) {
+		V value = (V) paramMap.getOrDefault(info.getName(), info.getDefaultValue());
+		if (value == null && !info.isOptional() && info.getDefaultValue() == null) {
```
walterddr: This is a bit confusing from your previous `hasDefaultValue()` javadoc. if `hasDefaultValue()` is added and here defaultValue cannot be null if it is not optional. it seems like:
```
info.hasDefaultValue() === (!info.isOptional() && info.getDefaultValue() == null)
```

flink-ml/flink-ml-api/src/main/java/org/apache/flink/ml/util/persist/MLStageFactory.java
```diff
@@ -0,0 +1,67 @@
...
+import org.apache.flink.ml.api.misc.persist.Persistable;
+
+import com.google.gson.Gson;
+import com.google.gson.JsonObject;
+
+/**
+ * Utility to restore a PipelineStage from a stage json.
+ */
+@PublicEvolving
+public class MLStageFactory {
```
walterddr: How about just calling it `PipelineStageFactory` ?

flink-ml/flink-ml-api/src/main/java/org/apache/flink/ml/api/misc/param/ParamInfo.java
```diff
@@ -0,0 +1,145 @@
...
+		return name;
+	}
+
+	/**
+	 * Returns the aliases of the parameter. The alias will be an empty string array by default.
+	 *
+	 * @return the aliases of the parameter
+	 */
+	public String[] getAlias() {
+		return new String[0];
```
ex00: probably better to use a variable?

flink-ml/flink-ml-api/src/main/java/org/apache/flink/ml/api/core/Pipeline.java
```diff
@@ -0,0 +1,213 @@
...
+ * </ul>
+ *
+ * <p>In addition, a pipeline can also be used as a {@link PipelineStage} in another pipeline, just
+ * like an ordinary {@link Estimator} or {@link Transformer} as describe above.
+ */
+@PublicEvolving
+public final class Pipeline implements Estimator<Pipeline, Pipeline>, Transformer<Pipeline>,
+	Model<Pipeline> {
+	private static final long serialVersionUID = 1L;
+	private List<PipelineStage> stages;
```
ex00: is it final?

flink-ml/pom.xml
```diff
@@ -0,0 +1,39 @@
...
+	<modelVersion>4.0.0</modelVersion>
+
+	<parent>
+		<groupId>org.apache.flink</groupId>
+		<artifactId>flink-parent</artifactId>
+		<version>1.9-SNAPSHOT</version>
+		<relativePath>..</relativePath>
+	</parent>
+
+	<artifactId>flink-ml</artifactId>
```
walterddr: Also I think naming this module as `flink-ml` might've conflict with the one in `flink-library/flink-ml` which is not a scala free project, thus travis complains. it might be a good idea to change it to some other naming ?

flink-ml/flink-ml-api/pom.xml
```diff
@@ -0,0 +1,45 @@
...
+	<artifactId>flink-ml-api</artifactId>
+
+	<dependencies>
+		<dependency>
+			<groupId>org.apache.flink</groupId>
+			<artifactId>flink-table-api-java</artifactId>
+			<version>${project.version}</version>
+		</dependency>
+		<dependency>
+			<groupId>com.google.code.gson</groupId>
```
bowenli86: gson is a pretty common lib that users have in their code. We may need to shade gson to avoid with conflict with user code. We can do it in this or another PR.

flink-ml/flink-ml-api/src/main/java/org/apache/flink/ml/api/core/Pipeline.java
bowenli86: can't we eagerly check and record that in `Pipeline appendStage(PipelineStage)` with a bool member variable?

flink-ml/flink-ml-api/src/main/java/org/apache/flink/ml/api/misc/persist/Persistable.java
```diff
@@ -0,0 +1,39 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.ml.api.misc.persist;
+
+/**
+ * An interface to allow PipelineStage persistence and reload. As of now, we are using JSON as
```
bowenli86: nit:

```suggestion
 * An interface to allow PipelineStage persist and reload. As of now, we are using JSON as
```

flink-ml/pom.xml
shaoxuan-wang: Thanks for the comments, @walterddr. Can you please suggest a proper name for this root module. I think flink-ml is the most appropriate one, and prefer to rename the legacy flink-ml (`flink-library/flink-ml`) module to flink-dataset-ml (I raised this [discussion](http://apache-flink-mailing-list-archive.1008284.n3.nabble.com/DISCUSS-FLIP-39-Flink-ML-pipeline-and-ML-libs-td28633.html) in ML, let us see what the others suggest). Or we can investigate if it is possible to eliminate the travis complains via changing the verification scripts to scan the scala package by obsolete package name: `flink-library/flink-ml` needs a scala suffix, while `flink-ml` does not.

c4emmmm:
Example libs and tests are pushed to a new branch called flink-ml-preview, including a kmeans estimator/model, two transformers and a simple kmeans test case. Though there's trouble running this test case, it should be helpful for understanding how to develop ml libs under flink-ml's core api and how the whole flink-ml works. Please check if necessary.

flink-ml/pom.xml
walterddr: I actually like the idea to change the current `flink-ml` to `flink-dataset-ml` or `flink-mllib`. If we were treating `flink-ml` like a first class citizen similar to how we moved `flink-table` out of the library submodule. I think the top level should take the most generic name of `flink-ml`.

flink-ml/flink-ml-api/pom.xml
walterddr: +1. I was actually also thinking whether we can use the already shaded jackson library in flink-shaded?

flink-ml/flink-ml-api/src/main/java/org/apache/flink/ml/util/persist/MLStageFactory.java
c4emmmm: Good idea. I'll rename it.

flink-ml/flink-ml-api/src/main/java/org/apache/flink/ml/api/misc/param/Params.java
c4emmmm: In my opinion, null may be a valid default value. Some estimator may treat null value as a special case. It works like this: when getting a param which is not configured,
if hasDefaultValue, use defaultValue
if !hasDefaultValue && isOptional, use null
if !hasDefaultValue && !isOptional, throw exception, this means the params must be set manually
I will add @Nullable at the setter of defaultValue. Or do you have any question about this behavior or this prerequisite(null can be valid param value)?

flink-ml/flink-ml-api/src/main/java/org/apache/flink/ml/api/misc/param/ParamInfo.java
c4emmmm: Thanks, I will add to the javadoc.

flink-ml/flink-ml-api/src/main/java/org/apache/flink/ml/api/misc/param/ParamInfo.java
c4emmmm: Yes, this is an omission. Since the constructors of ParamInfo are already chaos, I will refactor this with a new ParamInfoFactory class, and make the alias configurable.

flink-ml/flink-ml-api/src/main/java/org/apache/flink/ml/api/core/Pipeline.java
```diff
@@ -0,0 +1,213 @@
...
+	private Params params;
+
+	public Pipeline() {
+		this(new ArrayList<>());
+	}
+
+	public Pipeline(List<PipelineStage> stages) {
+		this.stages = stages;
+		this.params = new Params();
```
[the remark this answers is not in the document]
c4emmmm: The params is currently useless, but only as the return of the getParams() method. All of the information the pipeline needs is in the PipelineStage list. Neither can we set the params of these stages via this params.
We did think about how to use this Params, like using it to store all params in this Params and sharing among these stages, but gave up for fare of the conflicting. We are still finding its usage.
Removing the WithParams interface from Pipeline means removing it from all basic stages, and some default implementation will be unable to provide. I prefer to maintain the status quo.

c4emmmm:
Thank @Xpray , @walterddr , @bowenli86 , @ex00 for your advice. I have pushed a new commit modifying Pipeline. The changes are:
Make stages final, and getStages() returns an immutable copy of the stages. 
New pipeline with a pipelineStages list and restoring from json also use appendStage().
(These make the stages only editable via the appendStage() method.)
Validate and record index of last estimator in appendStage(), then all stages will be guaranteed valid and there'll be no need to find the index of last estimator many times.

flink-ml/flink-ml-api/src/main/java/org/apache/flink/ml/api/core/Pipeline.java
walterddr: Yes, I think that's also a possibility. Just we need to carefully maintain the code to limit what API can alter the "pipeline" (e.g. each time pipeline changes, we need to update the state). 

One more thinking on this is, we can always make the pipeline final, since every time we `fit` the pipeline, we create a new one. what do you think @c4emmmm

c4emmmm: I have renewed the Pipeline add left a comment. It's outside the code page so maybe you didn't see it. Sorry that i'm not familiar with github, maybe there's a better way to let you know.
The new Pipeline has only final fields, except lastEstimatorIndex which is an internal cache. Every stage must be added via appendStage() method including in constructor or fromJson() method. The return of getStages() will be an immutable copy. 
These makes it impossible to modify the stages list other than append a new stage via  appendStage(). So validation or finding out the last estimator can be done eagerly when a new stage is added. This should solve all your problems.

flink-ml/flink-ml-api/pom.xml
c4emmmm: It's good not to introduce new dependency, I'll try to use jackson instead of gson.

c4emmmm: When treating maps with flexible values(like the 'paramMap' in Params), jackson seems not as friendly as gson, which can get a JsonObject and parse each value with different class hint. Since the flink-ml code mostly runs at driver side only once rather than per record, ease of use is more important than efficiency and I'd like to shade gson and use it. Do you have any idea about better ways to use jackson? @walterddr

flink-ml/flink-ml-api/src/main/java/org/apache/flink/ml/api/core/PipelineStage.java
```diff
@@ -0,0 +1,70 @@
...
+ * <p>Each pipeline stage is with parameters and meanwhile persistable.
+ *
+ * @param <T> The class type of the PipelineStage implementation itself, used by {@link
+ *            org.apache.flink.ml.api.misc.param.WithParams}
+ * @see WithParams
+ */
+interface PipelineStage<T extends PipelineStage<T>> extends WithParams<T>, Serializable,
+		Persistable {
+
+	default String toJson() {
```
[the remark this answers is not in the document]
c4emmmm: Do you mean just using these names or serializing to a byte array like most of serialization?
We hope the serialized String is human-readable and end-users can easily understand and edit it. This makes it possible to change some params and rerun a pipeline without recompiling the code. Json seems to be a standard and friendly way to express the params of a stage. 
On the other hand, className is needed when restoring. It's now appended to the json with signWithClass() method in PipelineStageFactory. It's hard to provide such tool method if the serialized String is not a standard json map format.
That's why we explicitly use 'json' in the method name and why we use json as the serialize result. But we do wonder whether it's good to use json in serializing. What do you think?

flink-ml/flink-ml-api/src/main/java/org/apache/flink/ml/api/misc/persist/Persistable.java
```diff
@@ -0,0 +1,39 @@
...
+ * format.
+ */
+public interface Persistable {
```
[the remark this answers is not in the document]
c4emmmm: Thanks for your advice.
1.I'll add it to the javadoc.
2.Besides serialization, we hope there's a human-readable string to describe the pipeline/stage, and end-users can easily understand and edit it. This makes it possible to change some params and rerun a pipeline without recompiling the code. I believe such string is enough for serialization, and I choose to use json as the unified serialized string, which is standard and friendly.
But I'm also thinking whether it's good to use json, what do you think?
3.I'm working on adding tests, including this. It can be done today.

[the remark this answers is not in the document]
wuchong: Thanks for the explanation. It makes sense to me. I think it's fine to use JSON. Because it's readable and there are a lot of json visualization tools on web.

flink-ml/flink-ml-api/src/main/java/org/apache/flink/ml/util/persist/PipelineStageFactory.java
```diff
@@ -0,0 +1,67 @@
...
+	 * stageClassName, and class of stage has a non-parameter constructor.
+	 *
+	 * @param json the stage json to restore a PipelineStage from
+	 * @return the restored PipelineStage
+	 */
+	@SuppressWarnings("unchecked")
+	public static <T extends Persistable> T createFromJson(String json) {
+		Gson gson = new Gson();
+		JsonObject jobj = gson.fromJson(json, JsonObject.class);
+		String className = jobj.get("stageClassName").getAsString();
```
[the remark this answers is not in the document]
c4emmmm: Thanks, i'll take it

flink-ml/pom.xml
c4emmmm: There's another module consisting of some Estimators and Transformers, which will be proposed after this PR. The module will be named 'flink-ml-lib', like the api is 'flink-ml-api'. If flink-library/flink-ml is renamed to flink-mllib, I'm afraid it'll be quite confusing.

c4emmmm:
Test cases are added, including the behavior of pipeline and params, serializing and restoring a pipeline, and the ExtractParamInfoUtils.

flink-ml/flink-ml-api/src/main/java/org/apache/flink/ml/api/core/Pipeline.java
```diff
@@ -0,0 +1,217 @@
...
+	 *     </li>
+	 * </ul>
+	 *
+	 * After all the {@link Estimator}s are trained to fit their input tables, a new pipeline will be created
+	 * with the same stages in this pipeline, except that all the Estimators in the new pipeline are replaced
+	 * with their corresponding Models generated in the above process.
+	 * <p>If there is no {@link Estimator} in the pipeline, the method returns a copy of this pipeline.
+	 */
+	public Pipeline fit(Table input) {
+		List<PipelineStage> transformStages = new ArrayList<>();
```
ex00: I suggest to set `stages.size();` as initial capacity for `transformStages` list

flink-ml/flink-ml-api/src/main/java/org/apache/flink/ml/util/param/ExtractParamInfosUtil.java
```diff
@@ -0,0 +1,69 @@
...
+import org.apache.flink.ml.api.misc.param.WithParams;
+
+import java.lang.reflect.Field;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Utility to extract all ParamInfos defined in a WithParams, mainly used in persistence.
+ */
+public class ExtractParamInfosUtil {
```
ex00: I assume you can make it as final class.

flink-ml/flink-ml-api/src/main/java/org/apache/flink/ml/util/param/ExtractParamInfosUtil.java
```diff
@@ -0,0 +1,69 @@
...
+		}
+
+		Field[] fields = clz.getDeclaredFields();
+		for (Field f : fields) {
+			f.setAccessible(true);
+			if (ParamInfo.class.isAssignableFrom(f.getType())) {
+				try {
+					result.add((ParamInfo) f.get(s));
+				} catch (IllegalAccessException e) {
+					System.err.println("Failed to extract field:" + f.getName() + ", ignore it.");
```
ex00: I suppose to use logger for logging error here.

flink-ml/flink-ml-api/src/test/java/org/apache/flink/ml/api/core/PipelineTest.java
```diff
@@ -0,0 +1,181 @@
...
+			}
+			res.append(((SelfDescribe) s).describe());
+		}
+		return res.toString();
+	}
+
+	/**
+	 * Interface to describe a class with a string, only for pipeline test.
+	 */
+	public interface SelfDescribe {
```
ex00: I think, that all implementations of this interface are stubs - not mock, because you have implemented some logic for check behavior. I may be mistaken, please correct me

flink-ml/flink-ml-api/src/main/java/org/apache/flink/ml/api/core/Pipeline.java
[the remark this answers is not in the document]
ex00: Thank you for the explanation.

flink-ml/flink-ml-api/src/test/java/org/apache/flink/ml/api/core/PipelineTest.java
c4emmmm: I'm not sure about the difference between stub and mock. Let me find it out first.

Other advices seem reasonable. I'll take them. Thanks.

flink-ml/flink-ml-api/src/main/java/org/apache/flink/ml/api/core/PipelineStage.java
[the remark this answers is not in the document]
walterddr: Yes, this is also my confusion I think. let me rephrase it differently:
1. if this is only for human readability, why do we need a `toJson` and `fromJson`? shouldn't the `toJson` be enough? 
2. I can see that parameter editing is a very tunable part for ML users. I am not expert but do we usually package param literals (e.g. the values) and models together? can we load them differently? e.g. load the param json and the model json separately. 
3. with my personal preference, if it can be a choice for user (json or yaml, or other) that would be even better.

I think one large question I have is whether we use this for serving purpose. if we are not using this for serving purpose. I am actually fine with Json or Yaml.

flink-ml/flink-ml-api/src/main/java/org/apache/flink/ml/api/misc/persist/Persistable.java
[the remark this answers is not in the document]
walterddr: +1 thanks for the explanation @c4emmmm.

flink-ml/pom.xml
walterddr: backward compatibility and version supporting seems reasonable. should we mark the current FlinkML as deprecated first before we merge? 
also @wuchong makes a statement that I think we might need to do if we go down this path, we might also want to send a pooling mail to user/dev to make sure the community is OK for this.

flink-ml/flink-ml-api/src/main/java/org/apache/flink/ml/api/core/PipelineStage.java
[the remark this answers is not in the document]
ex00: +1 to rename methods for serialization/deserialization model stages.
If we would want to support other formats for e.g. PMML for serialization/deserialization, then toJson/fromJson doesn't match to logic in this case.

I suggest move serialization/deserialization implementation of logic to separate object from direct implementation in this interface - and inject serializer/deserializer to pipeline object.
I mean next
```java
class ExampleTransformer implements Transformer<ExampleTransformer>{
   .... 
   public ExampleTransformer(PipelineSerializer implementationForSpecificFormat){
            this.serializer=implementationForSpecificFormat;
....
    }
....
   public String serialization() {    // renamed version of toJson()
            return this.serializer.serialize(this);
   }
..... //some approach with deserializer
}
```
or move out serialization and deserialization methods from Pipeline objects at all.
```java
Transformer example = new ExampleTransformer();
//actions with ExampleTransformer
String serializedPipeline = new PipelineSerializer("supported_format_name_like_json").serialize(example);
```
I think this approach is more flexible for support different formats for serialization\deserialization pipelines and this the changes we can do in the next PR if community will approve this propolsal.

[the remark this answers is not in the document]
c4emmmm: @Rongrong @ex00 @blublinsky
Thank you all for so much concern.

I think most of your questions can be concluded as three points:
1.Why use JSON rather than standard model languages like PMML? What if users want to use other format?
2.How to serve a model?

In fact the Persistable is designed to be used inside FlinkML only. It's not only the storage format but also a part of user interface, like SQL. We tried to design a SQL-like language to describe pipeline stages, but discarded since it's requiring users learn new things, which may be a disadvantage at the beginning of FlinkML. Though the json may be hard to write from zero, it is friendly enough to read and edit. We tried to support these separately but found that if the representation is defined reversible, it can meet both requirements.

It's also planned to support standard model languages. Let's take PMML as the example. Since it is not reversible once a model is converted to PMML, and may be hard to describe all kinds of pipeline stages including pipelines, we didn't choose it as the unified storage format, but to provide a PMMLExportable interface. Models can implement it then users could convert their trained model to a PMML-format string or file.
As for serving, it's supposed that Model.transform() does exactly what serving does. This means a user can easily serve their model trained by FlinkML, just save the model json then use it in his serving job. It's also possible to use the trained model outside Flink if the model implements PMMLExportable. We can also provide PMMLPredictor(extends Transformer) to serve an external PMML model. I also tried implement FLIP-23 proposed by blublinsky on top of this API. I put metadata of the updating model queue(like kafka) in the params then acquire the source in Model.transform(). Though the implementation is quite complex, it's universal and easy to use for users. We can do more to integrate FLIP-39 and FLIP-23 and provide better experience for users.

This is what we thought about, and I hope it solve your problems. I would appreciate you letting me know if you have any question or think there's a mistake.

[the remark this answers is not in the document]
blublinsky: There are a few questions here:
1. I think ML and serving will require 2 separate instances of Flink. ML is mostly batch, while Serving can be either batch or streaming. I think such separation should be very explicit in the overall design. Although theoretically it is possible to use ML to update the model directly, such a use case is not going to be a dominant one.
2. If we agree on this, there are 2 reasons to export the model - here model is really a pipeline. It is used for continue learning and serving. Continue learning, in this case means that I have new data and I want to start from where I left off parameters wise. For this, current persistence implementation might be fine. For serving, it is probably not. You want to to make sure that serving can be done either in Flink or in a standalone Java app, for example. Tensorflow is trying to use the same format for both, but I do not see a big problem separating the two. For serving you might want to try to stick with one of the existing standards to avoid writing your own evaluator.
3. For export, I like the separation of the exporter from pipelines (see https://openscoring.io/blog/2018/07/09/converting_sparkml_pipeline_pmml/). In this case you can roll out additional exporters without modifying the existing code. Actually using the same approach for persistence allows for more flexibility as well

flink-ml/flink-ml-api/src/main/java/org/apache/flink/ml/api/misc/persist/Persistable.java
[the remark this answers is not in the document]
blublinsky: Human readability argument was used many times and it never quite work. Its fine to read a small model, but a large one?
I like Spark pipelines, which provides me with an API which I can use to get exactly the info that I need instead of sifting through a huge JSON string.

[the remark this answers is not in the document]
c4emmmm: @blublinsky
Sorry that I went through SparkML but didn't find exactly the api you mean. Since I never really used Spark ML, I'm afraid I missed something. Could you please tell me what you are indicating?

flink-ml/flink-ml-api/src/main/java/org/apache/flink/ml/api/core/PipelineStage.java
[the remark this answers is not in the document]
c4emmmm: @blublinsky
Thanks for your feedback. Your raised a few good points and suggestions, on persistence and model serving. Here are my thoughts.

First of all, I'd like to highlight that the current persistence system is mainly used inside Flink only, and export will be provided in the future as a utility interface.

--------------
I suppose that the persistence system should use a reversible format that can fully express a PipelineStage(including Pipeline) and can be used to restore the stage, whatever the stage is. Saved model can be used for serving or as an initial model of another estimator. Saved estimator may be used in tuning, where users only want to modify some params and rerun the same estimator.

Some 3rd-party system may also use the persistence result. The jpmml-sparkml you mentioned reads a saved PipelineModel and convert it to a PMML, where the saved PipelineModel is consist of saved stages files, and how these files are written is completely defined by its stage class itself. Exporting a pipeline as PMML and saving a pipeline are independent in this case.

In my point of view, the persistence is neither export nor checkpoint. It's more like a JobGraph or SQL, which describes the procedure of pipeline. Besides that, we also tried to find a way like SQL to describe the pipeline, but it's hard to design a new SQL-like language that everyone understands and accepts. JSON is finally chosen.

It's true that the JSON may be too complex to read if everything in a large pipeline put in it. In fact we suppose that the JSON contains only the params set by users, and large models should be saved separately and only the meta is in the JSON. Since where to save the model is also one of the required configurations of estimators, which is usually used to create a Sink, it should be quite natural to put in the params. Of course, utility interfaces and methods like "WithExternalModelFile" would be provided in future to make the implementation easier.

The persisted model JSON can be used to restore the pipeline and serve in Flink. All these should be really easy for a Flink user. If one want to use the trained model in an external serving system that support only standard model format like PMML, he should "export" the model as PMML before using it, and it's not the job persistence should do. For instance, we can implement serving transformers such as PMMLServingTransformer to serve external standard models.

I hope the above has explained “persistence". Next I will try to explain “export", which is not initially included in this PR.

---------------
I really appreciated your 3rd point. Initially, We planned to provide "Exportable" interfaces like PMMLExportable. Model may extend it and provide exportPMML() method to users. It's hard to give a unified implementation for every models, so the interface will be implemented by each model. Exportable requires the model implement it, but it would be hard to extend a model to support exporting without modifying the code, and would be even worst if multiple “export” is going to be supported.

Your 3rd suggestion is a perfect solution to address the above issue, we can implement an exporter utility, which is completely independent from the model. We may first introduce a basic Exporter<M extends Model> interface. And library developers or any one who wants to extend a library can implement it. Users need to create a new exporter instance then call exporter.export(model) to use it as needed. This should be more flexible and also easy to use. 

We can add “exporter” interface in the next PRs after the api and some initial estimators/transformers are merged. I would really appreciate if you can join the discussion when we design it at that time.

-----------------

Thanks again for your feedback. Please let me know if I have addressed all your comments.

flink-ml/flink-ml-api/src/main/java/org/apache/flink/ml/api/misc/persist/Persistable.java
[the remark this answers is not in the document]
blublinsky: The overall proposal is very similar to https://spark.apache.org/docs/latest/ml-pipeline.html.
and https://spark.apache.org/docs/latest/ml-pipeline.html#how-it-works
There is also a set of predefined, commonly used transformers https://spark.apache.org/docs/latest/ml-features.html

What I was referring to is something like this:

    val treeModel = model.stages(2).asInstanceOf[DecisionTreeClassificationModel]
    println("Learned classification tree model:\n" + treeModel.toDebugString)

c4emmmm:
As mentioned in the mail "[DISCUSS] FLIP-39: Flink ML pipeline and ML libs" in flink-dev, we have to add a TableEnvironment as input argument in Estimator.fit() and Transformer.transform().

This is because it's necessary to register a function, and may use to judge whether the pipeline is running in a batch or streaming environment since some algorithm only support batch training. While it is impossible to get the TableEnvironment via the only input Table without a flink-table-planner dependency, which should not be touched by users, so we have to add this as input.

Another way is to get the TableEnvironment via a static method, but it is not safe since users may create a TableEnvironment without using this method, and cause unknown problems.

Adding the input argument should not change much the experience of users and library developers. Please let me know if anyone has questions about this.

flink-ml/flink-ml-api/src/main/java/org/apache/flink/ml/api/misc/persist/Persistable.java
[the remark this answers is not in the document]
c4emmmm: Thanks for your explain. I found this example code in SparkML. It’s mainly use toDebugString method to acquire the info. But since I read the code, I found that the toDebugString method is provided only by a few models, most of which are tree models.

Here are all models I found that has this function or something like it:
DecisionTreeClassificationModel,
DecisionTreeRegressionModel,
RandomForest,
DecisionTreeModel,
GBTClassificationModel,
GBTRegressionModel

and an example debug string return by this method is like this:
RandomForestClassificationModel (uid=rfc_6c4ceb92ba78) with 20 trees
  Tree 0 (weight 1.0):
    If (feature 0 <= 3="" 10="" 1.0)="" if="" (feature="" <="0.0)" predict:="" 0.0="" else=""> 6.0)
       Predict: 0.0
     Else (feature 10 > 0.0)
      If (feature 12 <= 12="" 63.0)="" predict:="" 0.0="" else="" (feature=""> 63.0)
       Predict: 0.0
    Else (feature 0 > 1.0)
     If (feature 13 <= 3="" 1.0)="" if="" (feature="" <="3.0)" predict:="" 0.0="" else=""> 3.0)
       Predict: 1.0
     Else (feature 13 > 1.0)
      If (feature 7 <= 7="" 1.0)="" predict:="" 0.0="" else="" (feature=""> 1.0)
       Predict: 0.0
  Tree 1 ... (repeat count of trees times with different weights)

The requirement to acquire needed information from the pipelines is indeed valuable. If estimators or models could rephrase their params with a more friendly description, it would be easier to understand. I think this is the value that the toDebugString mainly provides. Maybe we should stipulate that all models have friendly and well-defined toString() method, or even add a describe() method in the basic interfaces like PipelineStage.

But it's not what we should do with the json we discussing. The json is designed to completely describe a pipeline, and should be able to generate from and convert to a pipeline without information loss, which allows it as a storage format. The readability is mostly for users editing a pipeline without editing the code and recompiling it. This makes it more convenient to tune or reuse pipeline. 

Another question is about initial transformers. It's already planned that next PR is some initial estimators, models and transformers after the api is accepted. They would have well defined toString method to provide enough information a user needs. More will be added since then, and hopefully even more libraries could be contributed to enrich the Flink ML.

c4emmmm:
> Thanks for contributing this PR, and great to see new machine learning effort arriving at Flink!
> 
> We also need to create a licensing NOTICE file to declare bundled libs. According to my experience, the community prefers to have it at the first place in case we forget it which may delay the later release process. Please include that in this PR, or open a PR as immediate followup. You can take a look at https://cwiki.apache.org/confluence/display/FLINK/Licensing and use [flink-connector-hive's NOTICE file](https://github.com/apache/flink/blob/master/flink-connectors/flink-connector-hive/src/main/resources/META-INF/NOTICE) as an example

@bowenli86 
Thanks for your notice. According to the advice to replace gson with flink-shaded-jackson, I have refactored the code and changed the dependency. Since now flink-ml-api only depends on "flink-table-api-java" and "flink-shaded-jackson", I think we no longer need the NOTICE file. Please let me know if I'm mistaken.

flink-ml/pom.xml
shaoxuan-wang: I started a [discussion](http://apache-flink-mailing-list-archive.1008284.n3.nabble.com/SURVEY-Usage-of-flink-ml-and-DISCUSS-Delete-flink-ml-td29057.html) in dev/user-ml to collect the feedback. @zentol pointed out that we can just remove the legacy flink-ml package, we do not expect any further active development on this package.  I think he is right. I will create another PR and delete it.

flink-ml/flink-ml-api/src/main/java/org/apache/flink/ml/api/core/PipelineStage.java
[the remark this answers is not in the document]
shaoxuan-wang: Thanks for all the comments. It triggers a broad and deep thinking on this interface and the future plan to export/load the model in different formats. I have offline discussed this with @c4emmmm. The final plan looks good to me. @blublinsky, @walterddr, @ex00 , please let us know if you have any further question on the solution proposed by @c4emmmm.

shaoxuan-wang:
Thanks for the contribution, @c4emmmm. The entire design looks good to me. I have gone over all the comments. It seems most of them have been addressed. There is one major comment about the way to persist and reload the mode. I noticed that you have responded to the reviewers, while they have not yet confirmed if they are happy with your solution or not. 
In general I am +1 on this PR. Will merge it in the next 1-2 days if there is no further comments/concerns coming out.

flink-ml/flink-ml-api/src/main/java/org/apache/flink/ml/api/core/PipelineStage.java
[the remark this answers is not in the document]
walterddr: +1 to expose more extendable interface in the next PR. 

I am still not fully convince that "toJson" addresses all of the concerns that we raise, especially once @blublinsky raised. but I also have a strong feeling that this can be address in future PR. 
One fundamental supporting factor I am on the "toJson" interface is that: a readable/editable format where a model can be bi-directionally converted in and out of FlinkML is 100% a need in multiple scenarios. for this PR purpose I don't think we need to focus on addressing "all" scenarios. what do you guys think?

IMO the follow up can be on "export", "serving", or even "standardizing" the model

flink-ml/pom.xml
walterddr: +1 to remove if no response in the user-ml. thanks for taking care of this @shaoxuan-wang .

flink-ml/flink-ml-api/src/main/java/org/apache/flink/ml/api/core/PipelineStage.java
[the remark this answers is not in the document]
shaoxuan-wang: Yes @walterddr. "toJson" interface makes it easier to read/edit a persisted pipeline. We can continue to discuss/work on the design of model export/load in the next PR. And it will make the discussion much easier when there is a PR for model export ready for review.
I can understand why the persistable interface introduces a lot confusions here. I talked with @c4emmmm offline again, and here are our decisions:
1. delete persistable interface (it is a common interface, but currently only intend for the persistence of pipelineStage, which does not quite make sense), and leave the implementation of `toJson` and `loadJson` just in PipelineStage.
2. delete the PipelineStageFactory interface as well as the `stageMap` in `toJson` and `loadJson`, as this contract is not ensured from interface.

flink-ml/pom.xml
zentol: I'd propose to name this module `flink-ml-parent`; I always find it odd if the same artifactId is used for completely different modules in different versions of a project. I believe this is an unnecessary source for confusion.

flink-ml/flink-ml-api/src/main/java/org/apache/flink/ml/api/core/PipelineStage.java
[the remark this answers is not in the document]
ex00: Thanks guys, this plan sounds good for me. Could you create subtask for it in FLINK-12470?

sunjincheng121:

sunjincheng121:
The CI was failed, I restarted it!

flink-ml/pom.xml
shaoxuan-wang: @zentol  agree, `flink-ml-parent` sounds good.

flink-ml-parent/flink-ml-api/pom.xml
```diff
@@ -0,0 +1,45 @@
...
+under the License.
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+		 xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+		 xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+	<modelVersion>4.0.0</modelVersion>
+
+	<parent>
+		<groupId>org.apache.flink</groupId>
+		<artifactId>flink-ml-parent</artifactId>
```
sunjincheng121: How about named `flink-machineLearning` or `flink-ml`, due to we already want to delete the `libraries/flink-ml` in this discussion thread?  http://mail-archives.apache.org/mod_mbox/flink-dev/201905.mbox/%3CCAAjCYUXy61SEouWVpX5nTJ6G2fwT76RAVtAOOf2XSVMHtzb_SQ@mail.gmail.com%3E

flink-ml-parent/flink-ml-api/src/test/java/org/apache/flink/ml/api/core/PipelineTest.java
```diff
@@ -0,0 +1,181 @@
...
+import org.apache.flink.ml.api.misc.param.Params;
+import org.apache.flink.table.api.Table;
+import org.apache.flink.table.api.TableEnvironment;
+
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+
+/**
+ * Tests for pipeline behavior.
```
sunjincheng121: Tests the behavior of {@link Pipeline}.

flink-ml-parent/flink-ml-api/src/test/java/org/apache/flink/ml/api/misc/ParamsTest.java
```diff
@@ -0,0 +1,72 @@
...
+import org.apache.flink.ml.api.misc.param.Params;
+
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+
+/**
+ * Test for the behavior and validator of Params.
+ */
+public class ParamsTest {
```
sunjincheng121: Test for the behavior and validator of {@link Params}.

flink-ml-parent/flink-ml-api/src/test/java/org/apache/flink/ml/api/misc/ParamsTest.java
```diff
@@ -0,0 +1,72 @@
...
+import org.apache.flink.ml.api.misc.param.ParamInfo;
+import org.apache.flink.ml.api.misc.param.ParamInfoFactory;
+import org.apache.flink.ml.api.misc.param.Params;
+
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+
+/**
+ * Test for the behavior and validator of Params.
```
sunjincheng121: Test for the behavior and validator of {@link Params}.

flink-ml-parent/flink-ml-api/src/test/java/org/apache/flink/ml/util/param/ExtractParamInfosUtilTest.java
```diff
@@ -0,0 +1,104 @@
...
+import org.apache.flink.ml.api.misc.param.Params;
+import org.apache.flink.ml.api.misc.param.WithParams;
+
+import org.junit.Test;
+
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * Test for ExtractParamInfosUtil.
```
sunjincheng121: Test for {@link ExtractParamInfosUtil}.

flink-ml-parent/flink-ml-api/src/test/java/org/apache/flink/ml/util/param/ExtractParamInfosUtilTest.java
```diff
@@ -0,0 +1,104 @@
...
+		}
+	}
+
+	/**
+	 * Mock interface extending WithParams with one ParamInfo. Only for test.
+	 * @param <T> implementation class of InterfaceWithParamInfo
+	 */
+	public interface InterfaceWithParamInfo<T extends InterfaceWithParamInfo>
+		extends WithParams<T> {
+		ParamInfo<String> KI = ParamInfoFactory.createParamInfo("KI", String.class)
```
sunjincheng121: `KI` is never used, right? can we remove it?

flink-ml-parent/flink-ml-api/src/main/java/org/apache/flink/ml/api/misc/param/ParamInfoFactory.java
```diff
@@ -0,0 +1,129 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.ml.api.misc.param;
+
+/**
+ * Factory to create ParamInfo, all ParamInfos should be create via this class.
```
sunjincheng121: ` be create` -> ` be created`?

c4emmmm:
@sunjincheng121 
Thanks for your review and the reminder to add flink-ml into flink-dist, I'll do it in later PRs. And I'll take the suggestions in javaDocs.
As for the module name, I think it's correct that we should avoid confusing users with using same name indicating completely different modules even in different versions.  But I'm also afraid that 'flink-ml-parent' is not a proper name. I'm not sure but maybe we should have only one 'parent' in flink. I'd like to ask more ideas before I change the module name again.

flink-ml-parent/flink-ml-api/src/test/java/org/apache/flink/ml/util/param/ExtractParamInfosUtilTest.java
c4emmmm: KI is defined in an interface, which means it's 'public static final' implicitly, and final variables must have initial value, so the assignment can not be omitted.
And the util is used to find out all ParamInfos defined in a PipelineStage and its superclass and implemented interfaces. The KI is defined here to help testing the acquisition from interfaces so itself can not be removed as well.

shaoxuan-wang:
@sunjincheng121 @c4emmmm , I think it is OK to use parent as suffix. We already have `flink-test-utils-parent` module.

shaoxuan-wang:
@flinkbot approve all

flink-ml/flink-ml-api/src/main/java/org/apache/flink/ml/api/core/PipelineStage.java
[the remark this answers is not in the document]
c4emmmm: @ex00 @walterddr @blublinsky
You may have noticed that we send a mail to flink-dev bringing up a discussion about export and import. Since we have discussed here in the pr, I think maybe you will be interested in the topic. We summarized some thoughts on this in this doc together with some examples:

https://docs.google.com/document/d/1B84b-1CvOXtwWQ6_tQyiaHwnSeiRqh-V96Or8uHqCp8/edit?usp=sharing

Please take a look if you have time.

PR #8214, opened 2019-04-18
[FLINK-11476] [table] Create CatalogManager to manage multiple catalogs and encapsulate Calcite schema

bowenli86:
## What is the purpose of the change

This PR introduces CatalogManager API and implementation to manage all registered ReadableCatalog instances. It encapsulates catalogs and their interrelations with Calcite.

It currently does not connect to planner yet, and has a couple followup TODOs:

- [FLINK-12257] Convert CatalogBaseTable to org.apache.calcite.schema.Table so that planner can use unified catalog APIs
- [FLINK-12258] Decouple CatalogManager with Calcite	

## Brief change log

- Created CatalogManager to mange all catalogs
- Created CatalogCalciteSchema and DatabaseCalciteSchema to model a strict two-level reference structure for Flink in Calcite, the full name of tables and views will be of format `<catalog_name>.<db_name>.<meta-object_name>`
- Introduced a concept of "current catalog" and "current database" to CatalogManager to help resolve meta-objects paths in SQL. E.g. with this function, "select * from tbl" can be properly interpreted as "select * from <current_catalog>.<current_db>.tbl"

## Verifying this change

This change added tests and can be verified as follows:

- Added FlinkCatalogManagerTest

## Does this pull request potentially affect one of the following parts:

  - Dependencies (does it add or upgrade a dependency): (no)
  - The public API, i.e., is any changed class annotated with `@Public(Evolving)`: (yes)
  - The serializers: (no)
  - The runtime per-record code paths (performance sensitive): (no)
  - Anything that affects deployment or recovery: JobManager (and its components), Checkpointing, Yarn/Mesos, ZooKeeper: (no)
  - The S3 file system connector: (no)

## Documentation

  - Does this pull request introduce a new feature? (yes)
  - If yes, how is the feature documented? (JavaDocs)

Documentation will be added later on when the feature completes.

PR #8404, opened 2019-05-10
[FLINK-11476][table] Create CatalogManager to manage multiple catalogs

dawidwys:

This PR adds a `CatalogManager` that stores tables and resolves paths accordingly. It subsumes #8214 

## Brief change log

  - added `CatalogManager` implementation with logic for resolving tables using also the default path
  - all table registrations and lookups in `TableEnvironment` go through `CatalogManager`
  - added Calcite's `Schema` structure corresponding to the `CatalogManager`. This enables Calcite to lookup for Tables through `CatalogManager`
  - **Added `Catalog` related APIs to `TableEnvironment`**

## Verifying this change

  - All existent tests should pass
  - `org.apache.flink.table.catalog.CatalogManagerTest`
  - `org.apache.flink.table.catalog.CatalogManagerPathResolutionTest`

## Documentation


twalthr:

dawidwys:
fyi: @bowenli86 @xuefuz

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/TableEnvironment.java
```diff
@@ -280,6 +316,55 @@
 	 */
 	void sqlUpdate(String stmt, QueryConfig config);
 
+	/**
+	 * Gets the current default catalog name of the current session.
+	 *
+	 * @return the current default catalog that is used for path resolution
+	 * @see TableEnvironment#setCurrentCatalog(String)
+	 */
+	String getCurrentCatalogName();
```
xuefuz: Nit: I wonder if we just name it getCurrentCatalog(), which returns the name of the current catalog. (Same for get current database name. I understand that the given name is more explicit, but getCurrentCatalog() is a mirror of the set method below and is more consistent with  catalog APIs in which, every get (table, database) is named getXxx() instead of getXxxName().

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/TableEnvironment.java
```diff
@@ -280,6 +316,55 @@
...
+	 * @see TableEnvironment#setCurrentDatabase(String, String)
+	 */
+	String getCurrentDatabaseName();
+
+	/**
+	 * Sets the current default catalog and database. That path will be used as the default one
+	 * when looking for unqualified object names.
+	 *
+	 * <p>This is used during resolution of object paths. The default path is constructed as
+	 * {@code [current-catalog].[current.database]}. During the resolution, first we try to look for
```
xuefuz: should it be [current-database]?

flink-table/flink-table-planner/src/main/java/org/apache/flink/table/catalog/CalciteCatalogTable.java
```diff
@@ -0,0 +1,87 @@
...
+		this.typeFactory = typeFactory;
+	}
+
+	public Table getTable() {
+		return table;
+	}
+
+	@Override
+	public Map<String, String> getProperties() {
+		throw new UnsupportedOperationException("Calcite table cannot be expressed as a map of properties.");
```
xuefuz: It might be better just to return an empty map instead of throwing an exception. By the definition, getProperties() return any additional properties a table might have. It doesn't mean the property form of the table.

flink-table/flink-table-planner/src/main/java/org/apache/flink/table/catalog/CalciteCatalogTable.java
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	@Override
+	public String getComment() {
+		return null;
+	}
+
+	@Override
+	public CatalogBaseTable copy() {
+		return this;
```
xuefuz: Maybe we throw an exception here because we are not able to provide a copy. Return just this might be dangerous because caller assumes a copy and make changes to the returned object.

flink-table/flink-table-planner/src/main/java/org/apache/flink/table/catalog/CatalogCalciteSchema.java
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	@Override
+	public Set<String> getTableNames() {
+		return new HashSet<>();
+	}
+
+	@Override
+	public RelProtoDataType getType(String name) {
+		return null;
```
xuefuz: Returning "null" means no type with the given name is found. I'm not sure if that's desired, but feel free to add an TODO item for this if necessary.

flink-table/flink-table-planner/src/main/java/org/apache/flink/table/catalog/CatalogManager.java
```diff
@@ -0,0 +1,315 @@
...
+		return externalCatalogs.keySet();
+	}
+
+	/**
+	 * Gets the current default catalog that will be used when resolving table path.
+	 *
+	 * @return the current default catalog
+	 * @see CatalogManager#resolveTable(String...)
+	 */
+	public String getCurrentCatalogName() {
```
xuefuz: maybe just getCurrentCatalog()?

flink-table/flink-table-planner/src/main/java/org/apache/flink/table/catalog/CatalogManager.java
```diff
@@ -0,0 +1,315 @@
...
+		}
+	}
+
+	/**
+	 * Gets the current default database name that will be used when resolving table path.
+	 *
+	 * @return the current default database
+	 * @see CatalogManager#resolveTable(String...)
+	 */
+	public String getCurrentDatabaseName() {
```
xuefuz: Same as above.

flink-table/flink-table-planner/src/main/java/org/apache/flink/table/catalog/DatabaseCalciteSchema.java
```diff
@@ -0,0 +1,126 @@
...
+			if (!catalog.tableExists(tablePath)) {
+				return null;
+			}
+
+			CatalogBaseTable table = catalog.getTable(tablePath);
+
+			if (table instanceof CalciteCatalogTable) {
+				return ((CalciteCatalogTable) table).getTable();
+			} else {
+				throw new TableException("Unsupported table type: " + table);
```
xuefuz: what about CatalogTable and CatalogView types?

flink-table/flink-table-planner/src/main/java/org/apache/flink/table/catalog/DatabaseCalciteSchema.java
```diff
@@ -0,0 +1,126 @@
...
+		try {
+			return new HashSet<>(catalog.listTables(dbName));
+		} catch (DatabaseNotExistException e) {
+			throw new CatalogException(e);
+		}
+	}
+
+	@Override
+	public RelProtoDataType getType(String name) {
+		return null;
```
xuefuz: Same as above.

flink-table/flink-table-planner/src/main/scala/org/apache/flink/table/catalog/ExternalTableUtil.scala
```diff
@@ -40,63 +40,67 @@ object ExternalTableUtil extends Logging {
...
-
-    case _: StreamTableEnvImpl if externalTable.isStreamTable =>
+    } else if (!isBatch && externalTable.isStreamTable) {
       val source = TableFactoryUtil.findAndCreateTableSource(externalTable)
       new StreamTableSourceTable[T](source.asInstanceOf[StreamTableSource[T]], statistics)
-
-    case _ =>
+    } else {
       throw new ValidationException(
         "External catalog table does not support the current environment for a table source.")
```
xuefuz: This message might need to change because it covers two cases: isBatch==true && externalTable.isStreamTable, and isBatch==false && externalTable.isBatchTable.

flink-table/flink-table-planner/src/main/java/org/apache/flink/table/catalog/CatalogManager.java
```diff
@@ -0,0 +1,315 @@
...
+	/**
+	 * Tries to resolve a table path to a {@link CatalogTableOperation}. First it tries to look for
+	 * {@code [default-path].[table-path]} if no table is found assumes the table path is a fully qualified one
+	 * and looks for {@code [table-path]}.
+	 *
+	 * @param tablePath table path to look for
+	 * @return {@link CatalogTableOperation} containing both fully qualified table identifier and its
+	 * {@link TableSchema}.
+	 */
+	public Optional<CatalogTableOperation> resolveTable(String... tablePath) {
```
xuefuz: This method seems resolving tables registered either in Catalogs or ExternalCatalogs. What about those registered in neither of them, such as those registered by registerTable()?

flink-table/flink-table-planner/src/main/scala/org/apache/flink/table/api/TableEnvImpl.scala
```diff
@@ -69,45 +66,19 @@ import _root_.scala.collection.mutable
...
-    .sqlToRelConverterConfig(getSqlToRelConverterConfig)
-    // the converter is needed when calling temporal table functions from SQL, because
-    // they reference a history table represented with a tree of table operations
-    .context(Contexts.of(
-      new TableOperationConverter.ToRelConverterSupplier(expressionBridge)
-    ))
-    // set the executor to evaluate constant expressions
-    .executor(new ExpressionReducer(config))
-    .build
+  private val BUILTIN_CATALOG_NAME = "builtin"
```
bowenli86: does TableEnvironment need to know the default catalog's name in CatalogManager?

flink-table/flink-table-planner/src/main/java/org/apache/flink/table/catalog/CatalogManagerSchema.java
```diff
@@ -0,0 +1,122 @@
...
+ * Bridge between {@link CatalogManager} and {@link Schema}. This way we can query Flink's specific catalogs
+ * from Calcite.
+ *
+ * <p>The mapping for {@link Catalog}s is modeled as a strict two-level reference structure for Flink in Calcite,
+ * the full path of tables and views is of format [catalog_name].[db_name].[meta-object_name].
+ *
+ * <p>It also supports {@link ExternalCatalog}s. An external catalog maps 1:1 to Calcite's schema.
+ */
+@Internal
+public class CatalogManagerSchema implements Schema {
```
bowenli86: can we use a better name for this class? Not sure if we should support ExternalCatalog yet, but even if we do, its naming has been a bit confusing

flink-table/flink-table-planner/src/main/java/org/apache/flink/table/planner/PlanningConfigurationBuilder.java
```diff
@@ -0,0 +1,202 @@
...
+		RelOptSchema relOptSchema = new CalciteCatalogReader(
+			rootSchema,
+			defaultSchema,
+			TYPE_FACTORY,
+			CalciteConfig.connectionConfig(getSqlParserConfig(calciteConfig(tableConfig))));
+
+		return new FlinkRelBuilder(context, cluster, relOptSchema, expressionBridge);
+	}
+
+	/** Returns the Calcite [[org.apache.calcite.plan.RelOptPlanner]] of this TableEnvironment. */
```
twalthr: Update comment here and below to Java.

flink-table/flink-sql-client/src/test/java/org/apache/flink/table/client/gateway/local/LocalExecutorITCase.java
```diff
@@ -229,9 +229,9 @@ public void testCompleteStatement() throws Exception {
 		final SessionContext session = new SessionContext("test-session", new Environment());
 
 		final List<String> expectedTableHints = Arrays.asList(
-			"TableNumber1",
-			"TableNumber2",
-			"TableSourceSink");
+			"builtin.default.TableNumber1",
+			"builtin.default.TableNumber2",
+			"builtin.default.TableSourceSink");
```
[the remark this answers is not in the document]
twalthr: One additional comment: could we rename the default names to something more meaningful such as: `default-catalog.default-database.table`

flink-table/flink-table-planner/src/main/java/org/apache/flink/table/catalog/CalciteCatalogTable.java
dawidwys: I agree this would be better. Unfortunately this makes it impossible to register the table in the catalog as it creates a copy on either `createTable` or `getTable`.

flink-table/flink-table-planner/src/main/java/org/apache/flink/table/catalog/CatalogCalciteSchema.java
dawidwys: AFAIK we do not support user defined types to be stored in Catalog, do we? That's why I think it is desired to return null here until we do.

flink-table/flink-table-planner/src/main/java/org/apache/flink/table/catalog/CalciteCatalogTable.java
dawidwys: I thought this method represents a serialized (toProperties) representation of the `Table`, but apparently I was wrong. I will fix this.

flink-table/flink-table-planner/src/main/java/org/apache/flink/table/catalog/CatalogManagerSchema.java
```diff
@@ -0,0 +1,122 @@
...
+
+	private final CatalogManager catalogManager;
+	private Boolean isBatch;
```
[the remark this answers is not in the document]
dawidwys: Sorry about that. :( That's how working in Scala & Java at the same time ends.

flink-table/flink-table-planner/src/main/scala/org/apache/flink/table/catalog/ExternalTableUtil.scala
dawidwys: Yes, but I think this message applies to both of those. BTW the logic has not changed, just the type of the flag has changed.

flink-table/flink-table-planner/src/main/java/org/apache/flink/table/catalog/CatalogManager.java
dawidwys: As Timo said, all objects in the end should be registered in one of the Catalogs. Right now we maintain the `ExternalCatalogs` just for backwards compatibility.

This method is used whenever we have to look up any object exclusively in the API module. This means e.g. in the `TableEnvironment#scan` method.

flink-table/flink-table-planner/src/main/java/org/apache/flink/table/catalog/DatabaseCalciteSchema.java
```diff
@@ -0,0 +1,126 @@
...
+
+			CatalogBaseTable table = catalog.getTable(tablePath);
+
+			if (table instanceof CalciteCatalogTable) {
+				return ((CalciteCatalogTable) table).getTable();
+			} else {
+				throw new TableException("Unsupported table type: " + table);
+			}
+		} catch (Exception e) {
+			throw new TableException("Could not find table: " + tableName, e);
```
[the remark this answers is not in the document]
dawidwys: Actually we should never end up here with `TableNotExistException` as we are checking `catalog.tableExist()` before. That's why I went with `TableExceptin` here.

I added appropriate comment there.

flink-table/flink-table-planner/src/test/scala/org/apache/flink/table/utils/TableTestBase.scala
```diff
@@ -118,8 +118,12 @@ abstract class TableTestUtil {
         }
         else if (expectedLine == TableTestUtil.ANY_SUBTREE) {
           break
-        }
-        else if (expectedLine != actualLine) {
+        } else if (!verifyCatalogPath && actualLine.contains("table=[[")) {
```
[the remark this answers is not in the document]
dawidwys: I did not want change all of the tests. That's why I added an option to strip the default catalog & database during validation.

dawidwys:
@bowenli86 The goal is to have a common `TableEnvironment` for both blink planner and legacy/flink planner. The entry point to either of them will be `Planner` interface, but the `CatalogManager` will not be part of it. The `CatalogManager` will entirely reside in the api module in `TableEnvironment`. The only fully functional `TableEnvironment` that exists in the master branch is the flink's one, thus this is the one we should work with. The one in `blink-planner` is just the minimal version to be able to run some of the tests.

flink-table/flink-table-planner/src/main/java/org/apache/flink/table/catalog/DatabaseCalciteSchema.java
dawidwys: For the `CatalogTable` we need to agree how do we retrieve/create `TableSourceFactory`, shall we get it from the catalog or via service discovery.

For a CatalogView we can add it already, but in this PR I wanted to focus on just the existing functionalities.

flink-table/flink-table-planner/src/main/scala/org/apache/flink/table/api/TableEnvImpl.scala
dawidwys: Does `CatalogManager` needs to know about it? This is just a initial value for the current catalog, so I think it fits better to the `TableEnvironment` rather than `CatalogManager`.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/TableEnvironment.java
```diff
@@ -280,6 +317,54 @@
...
+	 *
+	 * <p>This is used during the resolution of object paths. The default path is constructed as
+	 * {@code [current-catalog].[current-database]}. During the resolution, first we try to look for
+	 * {@code [default-path].[object-path]} if no object is found we assume the object path is a fully
+	 * qualified one and we look under {@code [object-path]}.
+	 *
+	 * @param catalogName The name of the catalog to set as the current default catalog.
+	 * @throws CatalogException thrown if a catalog with given name could not be set as the default one
+	 */
+	void useCatalog(String catalogName);
```
bowenli86: nit: can we add the exception (even though it's runtime exception) to signature to inform users this may throw exception, we've done that for catalog APIs

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/TableEnvironment.java
```diff
@@ -280,6 +317,54 @@
...
+	 * <p>This is used during the resolution of object paths. The default path is constructed as
+	 * {@code [current-catalog].[current.database]}. During the resolution, first we try to look for
+	 * {@code [default-path].[object-path]} if no object is found we assume the object path is a fully
+	 * qualified one and we look for {@code [object-path]}.
+	 *
+	 * @param catalogName The name of the catalog to set as the current catalog.
+	 * @param databaseName The name of the database to set as the current database.
+	 * @throws CatalogException thrown if the given catalog and database could not be set as the default ones
+	 */
+	void useDatabase(
```
bowenli86: nit: can we add the exception (even though it's runtime exception) to signature to inform users this may throw exception, we've done that for catalog APIs

bowenli86:
Can we add to the deprecated APIs some comments of which version they can be dropped? e.g. "Can be dropped after release 1.10"

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/TableEnvironment.java
```diff
@@ -280,6 +317,54 @@
...
+	 * database to the catalog's default one. To assign both catalog and database explicitly
+	 * see {@link TableEnvironment#useDatabase(String, String)}.
+	 *
+	 * <p>This is used during the resolution of object paths. The default path is constructed as
+	 * {@code [current-catalog].[current-database]}. During the resolution, first we try to look for
+	 * {@code [default-path].[object-path]} if no object is found we assume the object path is a fully
+	 * qualified one and we look under {@code [object-path]}.
+	 *
+	 * @param catalogName The name of the catalog to set as the current default catalog.
+	 * @throws CatalogException thrown if a catalog with given name could not be set as the default one
```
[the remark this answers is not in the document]
dawidwys: Connection, Authorization ...

flink-table/flink-table-planner/src/main/java/org/apache/flink/table/catalog/CatalogManager.java
```diff
@@ -0,0 +1,316 @@
...
+		checkArgument(tablePath != null && tablePath.length != 0, "Table path must not be null or empty.");
+
+		List<String> defaultPath = new ArrayList<>();
```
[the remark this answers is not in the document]
dawidwys: Yes, I also spotted that and I'm currently fixing it.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/TableEnvironment.java
dawidwys: This is against majority (if not all) code style guidelines. The `throws` cause should declare only those exception that we expect the user to handle.

BTW, me & @twalthr had a chat few times before and we fill we should revisit the exception structure in the catalog API anyway.

flink-table/flink-table-planner/src/main/java/org/apache/flink/table/catalog/CatalogManager.java
```diff
@@ -0,0 +1,316 @@
...
+			if (!potentialTable.isPresent()) {
+				potentialTable = lookupExternalTable(path);
+			}
+			return potentialTable.map(schema -> new CatalogTableOperation(path, schema));
+		} catch (TableNotExistException e) {
+			return Optional.empty();
+		}
+	}
+
+	private Optional<TableSchema> lookupCatalogTable(List<String> path) throws TableNotExistException {
```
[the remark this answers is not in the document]
dawidwys: In `currentCatalog.getTable(objectPath)`

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/TableEnvironment.java
bowenli86: Adding CatalogException to signature is actually suggested by Timo at https://github.com/apache/flink/pull/8007#discussion_r275425557 .

I agree we should revisit the exceptions as I too find it inconsistent and problematic

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/CatalogManager.java
```diff
@@ -19,33 +19,35 @@
...
 
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 
+import static java.lang.String.format;
```
bowenli86: can we revert the newly added static import besides Preconditions? I think we should avoid static import as much as possible as they make the code not noticeably shorter but less readable and maintainable.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/CatalogManager.java
```diff
@@ -250,56 +258,64 @@ public void setCurrentDatabase(String databaseName) throws DatabaseNotExistExcep
...
 	}
 
-	private Optional<CatalogTableOperation> lookupPath(List<String> path) {
+	private Optional<CatalogTableOperation> lookupPath(List<String> prefix, List<String> userPath) {
 		try {
-			Optional<TableSchema> potentialTable = lookupCatalogTable(path);
+			List<String> path = new ArrayList<>(prefix);
+			path.addAll(userPath);
+
+			Optional<CatalogTableOperation> potentialTable = lookupCatalogTable(path);
```
bowenli86: you can check whether the path is of size 3 before calling `lookupCatalogTable()` because we are fully sure that catalog table's full path must be of size 3.

And in `lookupCatalogTable()`, we can change the check from `if (path.size() == 3)` to `checkArgument(path.size() == 3)` and add a comment to let callers be aware that they should make sure their input args are valid

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/CatalogManager.java
```diff
@@ -250,56 +258,64 @@ public void setCurrentDatabase(String databaseName) throws DatabaseNotExistExcep
...
-			return potentialTable.map(schema -> new CatalogTableOperation(path, schema));
+			return potentialTable;
 		} catch (TableNotExistException e) {
 			return Optional.empty();
 		}
 	}
 
-	private Optional<TableSchema> lookupCatalogTable(List<String> path) throws TableNotExistException {
-		if (path.size() >= 3) {
+	private Optional<CatalogTableOperation> lookupCatalogTable(List<String> path) throws TableNotExistException {
```
bowenli86: the signature has been confusing. Returning an `Optional` implies that it will be empty if the operation doesn't find the target, but the method is also throwing `TableNotExistException` when the target table is not found.

I'd suggest try-catch `TableNotExistException` in `lookupCatalogTable()` and, in case of it, return an empty `Optional`. And remove the try-catch in `lookupPath()`

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/CatalogManager.java
```diff
@@ -250,56 +258,64 @@ public void setCurrentDatabase(String databaseName) throws DatabaseNotExistExcep
...
+		if (path.size() == 3) {
 			Catalog currentCatalog = catalogs.get(path.get(0));
 			String currentDatabaseName = path.get(1);
 			String tableName = String.join(".", path.subList(2, path.size()));
```
bowenli86: given that the path will be of size 3, tableName should just be `path.get(2)`

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/CatalogManager.java
dawidwys: External catalogs support arbitrary nesting.

dawidwys: External catalogs support arbitrary nesting.

dawidwys: External catalogs support arbitrary nesting.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/CatalogManager.java
dawidwys: Good catch!

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/CatalogManager.java
dawidwys: Good catch!

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/CatalogManager.java
dawidwys: I tend to disagree. My experience is, that it is usually the other way round(with some rare exceptions). Static imports make the code more readable and maintainable.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/CatalogManager.java
bowenli86: which part we are not sure about? isn't table lookup for external catalogs in `lookupExternalTable` rather than `lookupCatalogTable`?

bowenli86: what I mean is, `lookupPath()` can be written as 

private Optional<CatalogTableOperation> lookupPath(...) {
	...
	Optional<CatalogTableOperation> potentialTable 
       // if the path is not of size 3, we don't even need to bother with looking it up in catalogs
	if (path.size() == 3) {
		potentialTable = lookupCatalogTable(path);
	}

	if (!potentialTable.isPresent()) {
		potentialTable = lookupExternalTable(path);
	}
	return potentialTable
}

private Optional<CatalogTableOperation> lookupCatalogTable(List<String> path) {
	checkArgument(path.size() == 3, "xxx")
	try {

		...
	} catch () {}
}

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/CatalogManager.java
bowenli86: I believe we should only allow two cases to use static import: a) static constants b) extraordinarily common static helpers that don't cause confusions at all for any java developers in some specific context, like assertXxxx in junit tests, and checkArgument/checkNotNull for validating input args.

When I read this code, it indeed confused me that I had to look around for where the `format()` and `asList()` come from.

dawidwys: I think we should apply common sense here. In my experience it makes it much cleaner when we statically import well known methods or constants (like e.g. `String.format()`/`Assert.assertTrue`\`Collectors.toList()` etc.). There is very little chance anyone will confuse 
```
CatalogException(format(
	"A database with name [%s] does not exist in the catalog: [%s].",
	databaseName,
	currentCatalogName))
```
with any other `format`.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/CatalogManager.java
dawidwys: Ok, I misunderstood you. 

I don't like this approach as this way both `lookupPath` and `lookupCatalogTable` validate structure of the `Catalog` path. In my approach the whole logic is enclosed in a single lookup**CatalogTable** method.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/CatalogManager.java
dawidwys: I rephrased my previous comment and I think we agree in the principal. I guess the only difference is if we assume `String#format` and `Arrays#asList` to be well known by majority of java developers.

They really broadly used in Flink's code base(and not only in Flink's) and I truly believe they are rarely confused.

dawidwys: Anyway, after a quick scan to be remain consistent with other places, I will change it.

flink-table/flink-table-planner/src/main/java/org/apache/flink/table/catalog/CalciteCatalogTable.java
xuefuz: Never mind then.

flink-table/flink-table-planner/src/main/java/org/apache/flink/table/catalog/CatalogCalciteSchema.java
xuefuz: I thought this was to resolve any type in the type system and I saw in Blink something other than null was returned. Anyway, I wasn't clear what this API is for.

flink-table/flink-table-planner/src/main/java/org/apache/flink/table/catalog/DatabaseCalciteSchema.java
xuefuz: Okay.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/CatalogManager.java
bowenli86: the validation is fine. 

> This is not a performance critical code.

I think this will be critical to any benchmarks we run. I just heard from my colleague who is doing Flink/Blink TPCH now that, even how the `CatalogReader` runs is hearting the benchmark quite a bit because it needs to [exam three schema paths](https://github.com/apache/flink/pull/8404/files#diff-0bb30406df09433cce9b3617f910db9bR47) to find the table in a query. So any perf improvement in query planning phase would help us in the long run. Hope the background helps explain the necessity :)

bowenli86: the validation is fine. 

> This is not a performance critical code.

Just in general I think query planning is critical to any benchmarks we run. I just heard from my colleague who is doing Flink/Blink TPCH now that, even how the `CatalogReader` runs is hearting the benchmark quite a bit because it needs to [exam three schema paths](https://github.com/apache/flink/pull/8404/files#diff-0bb30406df09433cce9b3617f910db9bR47) to find the table. Hope the background helps explain the necessity. Though I think this probably won't hurt benchmarks run with catalog table and blink planner because a catalog table will be returned without going to look up external tables, any perf improvement in query planning phase would help us in the long run. 

I will leave this issue up to you.

bowenli86: I'll leave up to you

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/TableConfig.java
```diff
@@ -57,6 +57,18 @@
...
+	 * Specifies the name of the initial catalog to be created when instantiating
+	 * TableEnvironment.
+	 */
+	private String bultinCatalogName = "default-catalog";
+
+	/**
+	 * Specifies the name of the default database in the initial catalog to be created when instantiating
+	 * TableEnvironment.
+	 */
+	private String bultinDatabaseName = "default-database";
```
xuefuz: Having '-' in the names might  confuse the parser as it can be read as a numeric minus operation.

flink-table/flink-table-planner/src/main/java/org/apache/flink/table/planner/PlanningConfigurationBuilder.java
```diff
@@ -129,16 +142,48 @@ public Context getContext() {
...
+			caseSensitive = false;
+		} else {
+			caseSensitive = sqlParserConfig.caseSensitive();
+		}
+
+		SqlParser.Config parserConfig = SqlParser.configBuilder(sqlParserConfig)
+			.setCaseSensitive(caseSensitive)
+			.build();
+
+		return new CatalogReader(
```
xuefuz: Maybe this is taken care of somewhere else, but I'm wondering how we tell Calcite to look up in new catalogs that are registered later on in user's session.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/CatalogManager.java
```diff
@@ -0,0 +1,352 @@
...
+
+	/**
+	 * Registers a catalog under the given name. The catalog name must be unique across both
+	 * {@link Catalog}s and {@link ExternalCatalog}s.
+	 *
+	 * @param catalogName name under which to register the given catalog
+	 * @param catalog catalog to register
+	 * @throws CatalogException if the registration of the catalog under the given name failed
+	 */
+	public void registerCatalog(String catalogName, Catalog catalog) {
```
bowenli86: I might miss it somewhere: how is the catalog's corresponding `CatalogCalciteSchema` registered to Calcite's root schema?

flink-table/flink-table-planner/src/main/java/org/apache/flink/table/planner/PlanningConfigurationBuilder.java
dawidwys: I would say this is the core difference from the previous PR.

We do not create a separate Calcite schema, but we use the `CatalogManagerCalciteSchema` as the root one (that's why we had to create the `CalciteSchemaBuilder`). At the same time `CatalogManagerCalciteSchema` is just a thin adapter around `CatalogManager`. That means any changes to `CatalogManager` are directly visible by Calcite.

bowenli86:
Just passed it with another round. LGTM +1 to merge

flink-table/flink-table-planner/src/main/java/org/apache/flink/table/planner/PlanningConfigurationBuilder.java
xuefuz: Great!

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/CatalogManager.java
```diff
@@ -0,0 +1,352 @@
...
+		if (potentialCurrentCatalog == null) {
+			throw new CatalogException(String.format("A catalog with name [%s] does not exist.", catalogName));
+		}
+
+		if (!currentCatalogName.equals(catalogName)) {
+			currentCatalogName = catalogName;
+			currentDatabaseName = potentialCurrentCatalog.getDefaultDatabase();
+
+			LOG.info(
+				"Sets the current default catalog as [{}] and the current default database as [{}].",
```
twalthr: nit: `Set the`

flink-table/flink-table-planner/src/test/java/org/apache/flink/table/catalog/TestExternalTableSourceFactory.java
```diff
@@ -0,0 +1,93 @@
...
+import java.util.List;
+import java.util.Map;
+
+import static org.apache.flink.table.descriptors.ConnectorDescriptorValidator.CONNECTOR_TYPE;
+
+/**
+ * Table source factory for testing. It creates a dummy {@link TableSource}
+ * that returns an empty {@link TableSchema}.
+ */
+public class TestExternalTableSourceFactory implements TableSourceFactory<Row> {
```
twalthr: I think I added this comment before. But why not using the existing testing table source factory?

flink-table/flink-table-planner/src/test/scala/org/apache/flink/table/utils/TableTestBase.scala
```diff
@@ -106,8 +108,16 @@ abstract class TableTestUtil {
     // depends on the native machine (Little/Big Endian)
     val actualNoCharset = actual.replace("_UTF-16LE'", "'").replace("_UTF-16BE'", "'")
 
+    // majority of tests did not assume existence of Catalog API.
+    // this enables disabling catalog path verification
+    val acutalWithAdjustedPath = if (!verifyCatalogPath) {
```
twalthr: nit: typo

flink-table/flink-table-api-java-bridge/src/main/java/org/apache/flink/table/api/java/BatchTableEnvironment.java
```diff
@@ -283,9 +285,13 @@ static BatchTableEnvironment create(ExecutionEnvironment executionEnvironment) {
 	 */
 	static BatchTableEnvironment create(ExecutionEnvironment executionEnvironment, TableConfig tableConfig) {
 		try {
-			Class clazz = Class.forName("org.apache.flink.table.api.java.BatchTableEnvImpl");
-			Constructor con = clazz.getConstructor(ExecutionEnvironment.class, TableConfig.class);
-			return (BatchTableEnvironment) con.newInstance(executionEnvironment, tableConfig);
+			Class<?> clazz = Class.forName("org.apache.flink.table.api.java.BatchTableEnvImpl");
+			Constructor con = clazz.getConstructor(ExecutionEnvironment.class, TableConfig.class, CatalogManager.class);
+			CatalogManager catalogManager = new CatalogManager(
+				tableConfig.getBultinCatalogName(),
```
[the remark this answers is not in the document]
dawidwys: I think both spellings are actually correct, but I will change it.

flink-table/flink-table-planner/src/test/java/org/apache/flink/table/catalog/TestExternalTableSourceFactory.java
dawidwys: None of the existing factories actually create a `TableSource`, whereas it is needed to get a `TableSchema`. This is required for the path resolution, as the output of a path resolution is the full resolved path and the `TableSchema` of the entry.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/TableEnvironment.java
```diff
@@ -280,6 +315,142 @@
...
+	 *         </tr>
+	 *     </tbody>
+	 * </table>
+	 *
+	 * @param catalogName The name of the catalog to set as the current catalog.
+	 * @param databaseName The name of the database to set as the current database.
+	 * @throws CatalogException thrown if the given catalog and database could not be set as the default ones
+	 */
+	void useDatabase(
+		String catalogName,
```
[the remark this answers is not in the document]
dawidwys: It is, sort of. Underneath they translate to:
```
useCatalog = CatalogManager.setCatalog
useDatabase = CatalogManager.setCatalog + CatalogManager.setDatabase
```

I think it makes sense to leave it as it is, as the alternative would be to provide just `useDatabase(databaseName)`. I find this method error prone though. As the catalog name would be implicit.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/TableEnvironment.java
bowenli86: My comment can be a bit misunderstanding. Yeah, my point is we just need to be consistent with how it's exposed. An example is that now, contrary to `useCatalog()`, `Catalog`'s APIs not only document `CatalogException` in javadoc but also `throws CatalogException` in their signatures.

Seems that we all agree upon the javadoc part. What do you guys think of the signature part? If we all feel it's not necessary to expose it to users in the API, I feel it's better to remove `throws CatalogException` from `Catalog`'s API signatures to be consistent.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/TableEnvironment.java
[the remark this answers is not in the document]
xuefuz: This form is useful, as our DDL will have "use catalogName.databaseName" as well as "use databaseName". Additionally, it might make sense also provide useDatabase(databaseName), without which, user has to figure out the catalog name in order to switch databases in the current catalog.

flink-python/pyflink/table/table.py
```diff
@@ -512,18 +512,24 @@ def drop_columns(self, fields):
         """
         return Table(self._j_table.dropColumns(fields))
 
-    def insert_into(self, table_name):
+    def insert_into(self, table_path, *table_path_continued):
         """
         Writes the :class:`Table` to a :class:`TableSink` that was registered under
-        the specified name.
+        the specified name. For the path resolution algorithm :func:`~TableEnvironment.useDatabase`.
```
twalthr: For the path resolution algorithm see :func:`~TableEnvironment.use_database`

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/TableEnvironment.java
twalthr: I don't have a strong opinion here. Personally, I would add it to the signature if it is an important interface to highlight the exception. I mean there is a reason why Java also support declaring runtime exceptions explicitly. But I would definitely not bubble the runtime exception up in the stack of signatures.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/TableEnvironment.java
```diff
@@ -443,13 +443,11 @@
 	 *     </tbody>
 	 * </table>
 	 *
-	 * @param catalogName The name of the catalog to set as the current catalog.
 	 * @param databaseName The name of the database to set as the current database.
 	 * @throws CatalogException thrown if the given catalog and database could not be set as the default ones
```
twalthr: remove this `@throws`?

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/TableEnvironment.java
```diff
@@ -175,7 +175,7 @@
 	 * @return The resulting {@link Table}.
 	 * @throws ValidationException if no table is found using the given table path.
```
twalthr: I don't know if we should document these exceptions. Usually, there is a lot that can go wrong under the hood.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/TableEnvironment.java
twalthr: But I think the root cause here is that `CatalogException` is not a checked exception. I suggested it to handle communication failures similar to a `IOException`.

PR #8124, opened 2019-04-09
[FLINK-11877] Implement the runtime handling of the InputSelectable interface

sunhaibotb:
## What is the purpose of the change

**This pull request is only about the commits starting from a certain commit [86e62bb](https://github.com/apache/flink/pull/8124/commits/86e62bb30a0bd24d903fb04d62e361cd5f11cc42),  and all commits before are from [PR#7959](https://github.com/apache/flink/pull/7959) (this pull requests depends on it).**

This pull request implements the runtime handling of the `InputSelectable` interface, and this will enable the stream operators read their input selectively. The pull does not cover to support checkpointing of selective reading, and there is no strong demand on it at this moment, so we think it can be deferred to support.

## Brief change log

  - Introduces a new interface `Input` to represent the logical input of operators.
  - Introduces a new class `NetworkInput` which implements `Input` to represent the network input. It encapsulates deserializer, network buffers, input gate, etc.
  - Introduces a new class `StreamTwoInputSelectableProcessor` to handle selective reading.
  - Introduces a new class `TwoInputSelectableStreamTask` to execute the two-input operators implementing `InputSelectable`.
  - Changes UnionInputGate to support polling the next element. 
  - Changes the `addCoOperator()` method in `StreamGraph` that set the vertex class to  `TwoInputSelectableStreamTask` when the operator is an `InputSelectable` instance.
  - Adds a new method `boolean moreAvailable()` to the `InputGate` interface for the implementation of `NetworkInput`.
  - Adds micro-benchmark code for `StreamTwoInputProcessor` and `StreamTwoInputSelectableProcessor` .

## Verifying this change

This change added tests and can be verified as follows:
  - Added test for selective reading of  the operators.
  - Added integration tests for selective reading of  the operators.
  - Extended test for the watermark and streamStatus forwarding of `StreamTwoInputSelectableProcessor`.
  - Added test for the `boolean moreAvailable()` method of  `SingleInputGate` and `UnionInputGate`. 

## Micro Benchmarks
   - Measurement: **Throughput** in the `ANY` reading mode
   - Benchmark Executor:  JMH
       * @Fork value: 3
       * Number of warmup iterations: 10
       * Number of measurement iterations: 10
       * Records per forked invocation: 100,000,000
   - Environment
       * JDK: OracleJDK 1.8.0_202
       * OS: Linux 3.10.0
       * CPU: 96 cores  Intel(R) Xeon(R) Platinum 8163 CPU @ 2.50GHz
[...]

PR #8409, opened 2019-05-10
[FLINK-12478] Decompose monolithic run-loops in StreamTask implementa…

StefanRRichter:
…tions into step-wise calls

## Brief change log

Split processing of network events in stream tasks into step-wise granularity as a first step to migrate the threading model in stream tasks to a mailbox-based, single threaded approach.

See https://docs.google.com/document/d/1eDpsUKv2FqwZiS1Pm6gYO5eFHScBHfULKmH1-ZEWB4g

## Verifying this change

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/OneInputStreamTask.java
+++ b/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/OneInputStreamTask.java
@@ -39,8 +39,6 @@ public class OneInputStreamTask<IN, OUT> extends StreamTask<OUT, OneInputStreamO
 
 	private StreamInputProcessor<IN> inputProcessor;
 
-	private volatile boolean running = true;
-
 	private final WatermarkGauge inputWatermarkGauge = new WatermarkGauge();
 
 	/**
@@ -98,13 +96,8 @@ public class OneInputStreamTask<IN, OUT> extends StreamTask<OUT, OneInputStreamO
 	}
 
 	@Override
-	protected void run() throws Exception {
-		// cache processor reference on the stack, to make the code more JIT friendly
-		final StreamInputProcessor<IN> inputProcessor = this.inputProcessor;
-
-		while (running && inputProcessor.processInput()) {
-			// all the work happens in the "processInput" method
-		}
+	protected boolean performDefaultAction() throws Exception {
+		return inputProcessor.processInput();
 	}
 
 	@Override
@@ -116,6 +109,5 @@ public class OneInputStreamTask<IN, OUT> extends StreamTask<OUT, OneInputStreamO
 
 	@Override
 	protected void cancelTask() {
-		running = false;
 	}
 }
--- a/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamIterationHead.java
+++ b/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamIterationHead.java
@@ -24,6 +24,7 @@ import org.apache.flink.streaming.api.watermark.Watermark;
 import org.apache.flink.streaming.runtime.io.BlockingQueueBroker;
 import org.apache.flink.streaming.runtime.io.RecordWriterOutput;
 import org.apache.flink.streaming.runtime.streamrecord.StreamRecord;
+import org.apache.flink.util.FlinkRuntimeException;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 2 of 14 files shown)


flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
```diff
@@ -222,12 +222,24 @@ protected StreamTask(
...
 
+	/**
+	 * This method implements the default action of the task (e.g. processing one event from the input). Implementations
+	 * should (in general) be non-blocking.
+	 *
+	 * @return <code>true</code> if there is more work to perform as default action for this task and <code>false</code>
+	 * if the task is ready to finish.
+	 * @throws Exception on any problems in the action.
+	 */
+	protected abstract boolean performDefaultAction() throws Exception;
```
pnowojski: I think you should either squash this commit with next one that introduces `ActionContext` and reverts/replaces those changes, or you should extract `ActionContext` and move it from 2nd to 1st commit.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/OneInputStreamTask.java
```diff
@@ -98,13 +96,8 @@ public void init() throws Exception {
...
 
 	@Override
-	protected void run() throws Exception {
-		// cache processor reference on the stack, to make the code more JIT friendly
-		final StreamInputProcessor<IN> inputProcessor = this.inputProcessor;
-
-		while (running && inputProcessor.processInput()) {
-			// all the work happens in the "processInput" method
-		}
+	protected boolean performDefaultAction() throws Exception {
```
pnowojski: I'm worried about the possible performance degradation caused by moving the `while (true)` loop to the outside level and adding one more virtual call. @sunhaibotb was reporting that this is causing performance degradations. 

Before merging this, can we measure this? Maybe this is visible in `benchmarkCount` or we could write a micro benchmark for the `StreamTask` or `OneInputStreamTask`?

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamIterationHead.java
```diff
@@ -41,88 +42,73 @@
...
+		final String iterationId = getConfiguration().getIterationId();
+		if (iterationId == null || iterationId.length() == 0) {
+			throw new FlinkRuntimeException("Missing iteration ID in the task configuration");
+		}
+
+		this.dataChannel = new ArrayBlockingQueue<>(1);
+		this.brokerID = createBrokerIdString(getEnvironment().getJobID(), iterationId,
+			getEnvironment().getTaskInfo().getIndexOfThisSubtask());
+		this.iterationWaitTime = getConfiguration().getIterationWaitTime();
+		this.shouldWait = iterationWaitTime > 0;
```
1u0: I'd move this whole new code block into `init` method. This way it would be similar to `StreamIterationTail` and also avoid unneeded initialization in not task-execution contexts.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
StefanRRichter: I would disagree with that because the context without the mailbox makes no sense and without the boolean return value and no mailbox this would not work (there may be ways to make it work, but those would then also have to be changed in a later commit, so no win). I think it is ok two have two commits that apply the change incrementally and can stand by themselves.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/OneInputStreamTask.java
StefanRRichter: Maybe there is an impact, maybe not. But I also don't see any more performant alternative if we want to change the threading model. Besides, I executed at least the `ManualSpeedTest` for now and at least there was no regression.

pnowojski: `ManualWindowSpeedITCase` is not necessarily a good benchmark and definitely it's obsolete. Please check out for regressions in our existing `flink-benchmarks` results (once they will be fixed), since I'm pretty sure they supersede this one and/or add an additional one for better `OneInputStreamTask` coverage. We will need those benchmarks sooner or later before we complete this story. The sooner we have them, the better trends will be visible once we finally optimise something (or cause some regression).

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
pnowojski: Your changes are modifying the same lines of code twice, which just add reviewing noise (it took me some time to realise that you are overwriting those changes here later and I have already written a comment that something is missing with ` boolean performDefaultAction`)  and the same will be true in the future, if someone will want to look at the git history. 

Thus either we should come up with a sequence of changes that do not modify the same code twice too much (for example leaving `ActionContext` not implemented/empty/no-op in the first commit?) or the commits should be squashed.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/OneInputStreamTask.java
sunhaibotb: === TwoInputXStreamTask.java ===
@Override
protected void run() throws Exception {
	// cache processor reference on the stack, to make the code more JIT friendly
	final StreamTwoInputXProcessor<IN1, IN2> inputProcessor = this.inputProcessor;
	...
	inputProcessor.processInput();
	...
}


=== StreamTwoInputXProcessor.java ===
public boolean processInput() throws Exception {
	while (running) {
		...
		if (element.isRecord()) {
			continue;
		}
		...
	}
	return true;
}


=== TwoInputXStreamTask.java ===

@Override
protected void run() throws Exception {
	// cache processor reference on the stack, to make the code more JIT friendly
	final StreamTwoInputXProcessor<IN1, IN2> inputProcessor = this.inputProcessor;
	...
	while (running && inputProcessor.processInput()) {
		// all the work happens in the "processInput" method
	}
	...
}

@Override
protected void cancelTask() {
	running = false;
}

=== StreamTwoInputXProcessor.java ===
public boolean processInput() throws Exception {
	while (true) {
		...
		if (element.isRecord()) {
			return true;
		}
		...
	}
}

sunhaibotb: === TwoInputXStreamTask.java ===
@Override
protected void run() throws Exception {
	// cache processor reference on the stack, to make the code more JIT friendly
	final StreamTwoInputXProcessor<IN1, IN2> inputProcessor = this.inputProcessor;
	...
	inputProcessor.processInput();
	...
}


=== StreamTwoInputXProcessor.java ===
public boolean processInput() throws Exception {
	while (running) {
		...
		if (element.isRecord()) {
			continue;
		}
		...
	}
	return true;
}


=== TwoInputXStreamTask.java ===

@Override
protected void run() throws Exception {
	// cache processor reference on the stack, to make the code more JIT friendly
	final StreamTwoInputXProcessor<IN1, IN2> inputProcessor = this.inputProcessor;
	...
	while (running && inputProcessor.processInput()) {
		// all the work happens in the "processInput" method
	}
	...
}

@Override
protected void cancelTask() {
	running = false;
}

=== StreamTwoInputXProcessor.java ===
public boolean processInput() throws Exception {
	while (true) {
		...
		if (element.isRecord()) {
			return true;
		}
		...
	}
}

sunhaibotb: I only tested the following two cases,  Case 1 is much faster than Case 2. I can not sure if the change here has a performance regression.

Case 1:
```
=== TwoInputXStreamTask.java ===
@Override
protected void run() throws Exception {
	// cache processor reference on the stack, to make the code more JIT friendly
	final StreamTwoInputXProcessor<IN1, IN2> inputProcessor = this.inputProcessor;

	inputProcessor.processInput();
}


=== StreamTwoInputXProcessor.java ===
public boolean processInput() throws Exception {
	while (running) {
		...
		if (element.isRecord()) {
			continue;
		}
		...
	}
	return true;
}
```

Case 2:
```
=== TwoInputXStreamTask.java ===

@Override
protected void run() throws Exception {
	// cache processor reference on the stack, to make the code more JIT friendly
	final StreamTwoInputXProcessor<IN1, IN2> inputProcessor = this.inputProcessor;

	while (running && inputProcessor.processInput()) {
		// all the work happens in the "processInput" method
	}
}

=== StreamTwoInputXProcessor.java ===
public boolean processInput() throws Exception {
	while (true) {
		...
		if (element.isRecord()) {
			return true;
		}
		...
	}
}
```

StefanRRichter: I ran the following:
```
		env.setParallelism(1);
		source
			.map(new MultiplyByTwo())
			.addSink(new DiscardingSink<>());

master:
Benchmark                          Mode  Cnt      Score     Error   Units
SumLongsBenchmark.benchmarkCount  thrpt   30  16240.866 ± 311.845  ops/ms

master + #8442:
Benchmark                          Mode  Cnt      Score     Error   Units
SumLongsBenchmark.benchmarkCount  thrpt   30  16540.537 ± 295.100  ops/ms
```

```
		env.setParallelism(1);
		source
			.map(new MultiplyByTwo())
			.rebalance()
			.map((x) -> x)
			.addSink(new DiscardingSink<>());

master:
Benchmark                          Mode  Cnt     Score     Error   Units
SumLongsBenchmark.benchmarkCount  thrpt   30  9715.334 ± 170.632  ops/ms

master + #8442:
Benchmark                          Mode  Cnt     Score     Error   Units
SumLongsBenchmark.benchmarkCount  thrpt   30  9932.752 ± 124.494  ops/ms
```

I would conclude the changes run as fast or a notch faster than master on my local machine.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamIterationHead.java
StefanRRichter: In general not a bad idea, but how can I assign to `final` fields from an init method?

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
StefanRRichter: It changes the line twice because it is incrementally changed. I see no problem with that, it is work for different PRs with different purposes. If we would really decide to merge them, then we should also merge the JIRA tasks, but I think they are valuable as separated steps.

Even if i had a dummy context, the boolean is still required and the suggestion would not really help to solve the problem.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/OneInputStreamTask.java
sunhaibotb: In theory, I don't think it's possible to get faster because I don't see the optimization that might get faster in this change, at most it is the same as before. Feeling some kind of interference makes benchmark inaccurate, or this benchmarking can not accurately measure their performance. In my experience, on the contrary, it will make one more virtual call than before (as @pnowojski  said), because JIT do not inline `performDefaultAction()` which  has more than one implementations. Of course, the accurate benchmark results should be the ultimately criterion.

StefanRRichter: @sunhaibotb if you are in doubt, feel free to checkout the PR and double check the measurement :-) Does your theory also include that the code does a couple more things differently, e.g. it is not checking the `isRunning` volatile read in each iteration? Whether or not inlining or other optimizations for low cardinality polymorphism happend is also unclear, and last but not least the method call could be insignificant in the larger context.

StefanRRichter: I am not even claiming it is faster, probably it is just the same ballpark plus some variance, but at least seemingly no regression to worry about.

StefanRRichter: Btw I am also talking about the performance of #8442, not just this PR.

sunhaibotb: > last but not least the method call could be insignificant in the larger context.

I quite agree with that.

If we narrow down the context and just start to measure from the loop (like the benchmark of [PR-8124](https://github.com/apache/flink/pull/8124)), the difference should show up. But considering in the context of a job,  especially for most production jobs，this difference may have only a slight impact and can be ignored. So it's fine to me. @pnowojski, what do you think?

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
pnowojski: In that case maybe let's squash them?

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamIterationHead.java
StefanRRichter: I was thinking about this, but it seems like that field is supplied in some lazy way as of now, so it would not just work without deeper changes.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
pnowojski: Shall we close this PR and supersede it with https://github.com/apache/flink/pull/8431?

PR #8410, opened 2019-05-10
[FLINK-11159] Allow configuration whether to fall back to savepoints for restore

yanghua:

*This pull request provides a checkpoint config option `isPreferCheckpointForRecovery` to allow configuration whether to fall back to savepoints for restore*

## Brief change log

  - *Provide a checkpoint config option `isPreferCheckpointForRecovery`* 
  - *Allow configuration whether to fall back to savepoints for restore*
  - *Add two test cases*

## Verifying this change

  - *testRestoreLatestCheckpointWhenPreferCheckpoint*
  - *testRestoreLatestCheckpointWhenPreferSavepoint*

## Documentation

### DIFF
--- a/docs/dev/stream/state/checkpointing.md
+++ b/docs/dev/stream/state/checkpointing.md
@@ -76,6 +76,8 @@ Other parameters for checkpointing include:
 
   - *fail/continue task on checkpoint errors*: This determines if a task will be failed if an error occurs in the execution of the task's checkpoint procedure. This is the default behaviour. Alternatively, when this is disabled, the task will simply decline the checkpoint to the checkpoint coordinator and continue running.
 
+  - *prefer checkpoint for recovery*: This determines if a job will fallback to latest checkpoint even when there are more recent savepoints available to potentially reduce recovery time.
+
 <div class="codetabs" markdown="1">
 <div data-lang="java" markdown="1">
 {% highlight java %}
@@ -100,6 +102,9 @@ env.getCheckpointConfig().setMaxConcurrentCheckpoints(1);
 
 // enable externalized checkpoints which are retained after job cancellation
 env.getCheckpointConfig().enableExternalizedCheckpoints(ExternalizedCheckpointCleanup.RETAIN_ON_CANCELLATION);
+
+// allow job recovery fallback to checkpoint when there is a more recent savepoint
+env.getCheckpointConfig().setPreferCheckpointForRecovery(true);
 {% endhighlight %}
 </div>
 <div data-lang="scala" markdown="1">
--- a/docs/dev/stream/state/checkpointing.zh.md
+++ b/docs/dev/stream/state/checkpointing.zh.md
@@ -76,6 +76,8 @@ Other parameters for checkpointing include:
 
   - *fail/continue task on checkpoint errors*: This determines if a task will be failed if an error occurs in the execution of the task's checkpoint procedure. This is the default behaviour. Alternatively, when this is disabled, the task will simply decline the checkpoint to the checkpoint coordinator and continue running.
 
+  - *prefer checkpoint for recovery*: This determines if a job will fallback to latest checkpoint even when there are more recent savepoints available to potentially reduce recovery time.
+
 <div class="codetabs" markdown="1">
 <div data-lang="java" markdown="1">
 {% highlight java %}
@@ -125,6 +127,9 @@ env.getCheckpointConfig.setFailTasksOnCheckpointingErrors(false)
 
 // allow only one checkpoint to be in progress at the same time
 env.getCheckpointConfig.setMaxConcurrentCheckpoints(1)
+
+// allow job recovery fallback to checkpoint when there is a more recent savepoint
+env.getCheckpointConfig().setPreferCheckpointForRecovery(true)
 {% endhighlight %}
 </div>
 </div>
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/checkpoint/CheckpointCoordinator.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/checkpoint/CheckpointCoordinator.java
@@ -183,6 +183,8 @@ public class CheckpointCoordinator {
 	/** Registry that tracks state which is shared across (incremental) checkpoints. */
 	private SharedStateRegistry sharedStateRegistry;
 
+	private boolean isPreferCheckpointForRecovery;
+
 	// --------------------------------------------------------------------------------------------
 
 	public CheckpointCoordinator(
@@ -199,7 +201,8 @@ public class CheckpointCoordinator {
 			CompletedCheckpointStore completedCheckpointStore,
 			StateBackend checkpointStateBackend,
 			Executor executor,
-			SharedStateRegistryFactory sharedStateRegistryFactory) {
+			SharedStateRegistryFactory sharedStateRegistryFactory,
+			boolean isPreferCheckpointForRecovery) {
 
 		// sanity checks
 		checkNotNull(checkpointStateBackend);
@@ -233,6 +236,7 @@ public class CheckpointCoordinator {
 		this.executor = checkNotNull(executor);
 		this.sharedStateRegistryFactory = checkNotNull(sharedStateRegistryFactory);
 		this.sharedStateRegistry = sharedStateRegistryFactory.create(executor);
+		this.isPreferCheckpointForRecovery = isPreferCheckpointForRecovery;
 
 		this.recentPendingCheckpoints = new ArrayDeque<>(NUM_GHOST_CHECKPOINT_IDS);
 		this.masterHooks = new HashMap<>();
@@ -1080,7 +1084,7 @@ public class CheckpointCoordinator {
 			LOG.debug("Status of the shared state registry of job {} after restore: {}.", job, sharedStateRegistry);
 
 			// Restore from the latest checkpoint
-			CompletedCheckpoint latest = completedCheckpointStore.getLatestCheckpoint();
+			CompletedCheckpoint latest = completedCheckpointStore.getLatestCheckpoint(isPreferCheckpointForRecovery);
 
 			if (latest == null) {
 				if (errorIfNoCheckpoint) {
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/checkpoint/StandaloneCompletedCheckpointStore.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/checkpoint/StandaloneCompletedCheckpointStore.java
@@ -77,11 +77,6 @@ public class StandaloneCompletedCheckpointStore implements CompletedCheckpointSt
 	}
 
 	@Override
-	public CompletedCheckpoint getLatestCheckpoint() {
-		return checkpoints.isEmpty() ? null : checkpoints.getLast();
-	}
-
-	@Override
 	public List<CompletedCheckpoint> getAllCheckpoints() {
 		return new ArrayList<>(checkpoints);
 	}
--- a/flink-streaming-java/src/main/java/org/apache/flink/streaming/api/environment/CheckpointConfig.java
+++ b/flink-streaming-java/src/main/java/org/apache/flink/streaming/api/environment/CheckpointConfig.java
@@ -72,6 +72,9 @@ public class CheckpointConfig implements java.io.Serializable {
 	/** Determines if a tasks are failed or not if there is an error in their checkpointing. Default: true */
 	private boolean failOnCheckpointingErrors = true;
 
+	/** Determines if a job will fallback to checkpoint when there is a more recent savepoint. **/
+	private boolean preferCheckpointForRecovery = false;
+
 	// ------------------------------------------------------------------------
 
 	/**
@@ -283,6 +286,24 @@ public class CheckpointConfig implements java.io.Serializable {
 	@PublicEvolving
 	public boolean isExternalizedCheckpointsEnabled() {
 		return externalizedCheckpointCleanup != null;
+	}
+
+	/**
+	 * Returns whether a job recovery should fallback to checkpoint when there is a more recent savepoint.
+	 *
+	 * @return <code>true</code> if a job recovery should fallback to checkpoint.
+	 */
+	@PublicEvolving
+	public boolean isPreferCheckpointForRecovery() {
+		return preferCheckpointForRecovery;
+	}
+
+	/**
+	 * Sets whether a job recovery should fallback to checkpoint when there is a more recent savepoint.
+	 */
+	@PublicEvolving
+	public void setPreferCheckpointForRecovery(boolean preferCheckpointForRecovery) {
+		this.preferCheckpointForRecovery = preferCheckpointForRecovery;
 	}
 
 	/**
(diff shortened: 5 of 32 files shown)


yanghua:
@gyfora If you have time, can you review this PR for me? thanks.

flink-runtime/src/main/java/org/apache/flink/runtime/checkpoint/CheckpointCoordinator.java
```diff
@@ -183,6 +184,8 @@
 	/** Registry that tracks state which is shared across (incremental) checkpoints. */
 	private SharedStateRegistry sharedStateRegistry;
 
+	private boolean isPerferCheckpointForRecovery;
```
gyfora: There is a typo in basically every use of the word "preferCheckpointForRecovery" it should be prefer not "perfer". I think the easiest would be to search the changes and correct it in every place.

flink-runtime/src/main/java/org/apache/flink/runtime/checkpoint/CheckpointCoordinator.java
```diff
@@ -1093,6 +1098,23 @@ public boolean restoreLatestCheckpointedState(
 				}
 			}
 
+			CompletedCheckpoint candidate = null;
+			if (isPerferCheckpointForRecovery && completedCheckpointStore.getAllCheckpoints().size() > 1) {
```
gyfora: It feels like this part basically reimplements the getLatestCheckpoint() logic of the completed checkpoint store. Maybe instead of having this hear the getLatestCheckpoint( ) could take the preferCheckpoints boolean parameter for better encapsulation

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/environment/CheckpointConfig.java
```diff
@@ -72,6 +72,9 @@
 	/** Determines if a tasks are failed or not if there is an error in their checkpointing. Default: true */
 	private boolean failOnCheckpointingErrors = true;
 
+	/** Determines if a job will fallback to checkpoint when there is a savepoint later than checkpoint. **/
+	private boolean perferCheckpointForRecovery = false;
```
gyfora: I am personally not a big fan of configuring checkpoint parameters in the CheckpointConfig class, I actually wish that all these options would be exposed as general flink configuration parameters (settable in flink-conf.yaml).

What do you think @StefanRRichter ?

flink-runtime/src/main/java/org/apache/flink/runtime/checkpoint/CheckpointCoordinator.java
yanghua: Agree.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/environment/CheckpointConfig.java
yanghua: I think this config option belongs to checkpoint field. So provide it in `CheckpointConfig` sounds more reasonable. What's more, it is a job level config option, right? However, `flink-conf.yaml` is a cluster-level configuration file(if users deploy Flink with standalone mode, it will affect all the jobs).

flink-runtime/src/main/java/org/apache/flink/runtime/checkpoint/CheckpointCoordinator.java
yanghua: Sorry for the typo, will fix it soon.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/environment/CheckpointConfig.java
```diff
@@ -285,6 +288,16 @@ public boolean isExternalizedCheckpointsEnabled() {
 		return externalizedCheckpointCleanup != null;
 	}
 
+	@PublicEvolving
+	public boolean isPreferCheckpointForRecovery() {
+		return perferCheckpointForRecovery;
```
gyfora: There are still a few typos if you search for "perfer"

flink-runtime/src/main/java/org/apache/flink/runtime/checkpoint/StandaloneCompletedCheckpointStore.java
```diff
@@ -77,8 +79,32 @@ public void addCheckpoint(CompletedCheckpoint checkpoint) throws Exception {
...
 
 	@Override
-	public CompletedCheckpoint getLatestCheckpoint() {
-		return checkpoints.isEmpty() ? null : checkpoints.getLast();
+	public CompletedCheckpoint getLatestCheckpoint(boolean isPreferCheckpointForRecovery) {
+		if (checkpoints.isEmpty()) {
+			return null;
+		}
+
+		CompletedCheckpoint candidate = checkpoints.getLast();
```
gyfora: Do you think we could somehow avoid having pretty much exactly the same logic duplicated twice (ZK/Standalone)? Maybe the whole logic could be implemented in the CompletedCheckpointStore interface based on getAllCheckpoints()

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/environment/CheckpointConfig.java
yanghua: Sorry, will fix them.

flink-runtime/src/main/java/org/apache/flink/runtime/checkpoint/StandaloneCompletedCheckpointStore.java
yanghua: Agree, actually, I found the similar code snippet when I was implementing. The difference between the original implementation of the `getLatestCheckpoint` is the usage of `peekLast` and `getLast`. I was afraid to destroy the original semantics. If it does not matter, I'd like to do this refactor.

gyfora: we should preserve the current behaviour, but the difference between getLast and peekLast is minor, just check for empty list I guess

yanghua: I will give a default implementation in the `CompletedCheckpointStore `:

```java
default CompletedCheckpoint getLatestCheckpoint(boolean isPreferCheckpointForRecovery) throws Exception {
		if (getAllCheckpoints().isEmpty()) {
			return null;
		}

		CompletedCheckpoint candidate = getAllCheckpoints().get(getAllCheckpoints().size() - 1);
		if (isPreferCheckpointForRecovery && getAllCheckpoints().size() > 1) {
			List<CompletedCheckpoint> allCheckpoints;
			try {
				allCheckpoints = getAllCheckpoints();
				ListIterator<CompletedCheckpoint> listIterator = allCheckpoints.listIterator(allCheckpoints.size() - 1);
				while (listIterator.hasPrevious()) {
					CompletedCheckpoint prev = listIterator.previous();
					if (!prev.getProperties().isSavepoint()) {
						candidate = prev;
						LOG.info("Found a completed checkpoint before the latest savepoint, will use it to recover!");
						break;
					}
				}
			} catch (Exception e) {
				LOG.error("Method getAllCheckpoints caused exception : ", e);
				throw new FlinkRuntimeException(e);
			}
		}

		return candidate;
	}
```

gyfora: Sounds good, as long as we can be sure getAllCheckpoints() returns checkpoint in order.

yanghua:
@gyfora What do you think about the new version of this PR?

yanghua:
It's very strange, the Travis reported compile error :

```
16:40:05.120 [ERROR] /home/travis/build/apache/flink/flink-runtime/src/test/java/org/apache/flink/runtime/jobgraph/JobGraphTest.java:[343,95] constructor CheckpointCoordinatorConfiguration in class org.apache.flink.runtime.jobgraph.tasks.CheckpointCoordinatorConfiguration cannot be applied to given types;
16:40:05.120 [ERROR] required: long,long,long,int,org.apache.flink.runtime.checkpoint.CheckpointRetentionPolicy,boolean,boolean
16:40:05.120 [ERROR] found: long,long,long,int,org.apache.flink.runtime.checkpoint.CheckpointRetentionPolicy,boolean
```

But the class `JobGraphTest` does not use `CheckpointCoordinatorConfiguration`.

yanghua:
cc @gyfora

gyfora:
What do you mean it doesnt use it? you can see it here: https://github.com/apache/flink/blob/master/flink-runtime/src/test/java/org/apache/flink/runtime/jobgraph/JobGraphTest.java#L343

yanghua:
Oh @gyfora Sorry, yes, I have found it after I gave that comment. The reason is that I did not rebase the latest master code. Now, it works fine. I ping you to do the final check.

gyfora:
The changes look good, last thing that would be great to have is some documentation for the new feature :)

yanghua:
@gyfora OK. Will supply the docs.

yanghua:
@gyfora I have added java doc and flink doc. Please review.

docs/dev/stream/state/checkpointing.md
```diff
@@ -76,6 +76,8 @@ Other parameters for checkpointing include:
 
   - *fail/continue task on checkpoint errors*: This determines if a task will be failed if an error occurs in the execution of the task's checkpoint procedure. This is the default behaviour. Alternatively, when this is disabled, the task will simply decline the checkpoint to the checkpoint coordinator and continue running.
 
+  - *is prefer checkpoint for recovery*: This determines if a job will fallback to checkpoint when there is a savepoint later than checkpoint.
```
gyfora: I would change the wording slightly here to be more grammatically correct:

*prefer checkpoint for recovery*: This determines if a job will fallback to latest checkpoint even when there are more recent savepoints available to potentially reduce recovery time

docs/dev/stream/state/checkpointing.zh.md
```diff
@@ -76,6 +76,8 @@ Other parameters for checkpointing include:
 
   - *fail/continue task on checkpoint errors*: This determines if a task will be failed if an error occurs in the execution of the task's checkpoint procedure. This is the default behaviour. Alternatively, when this is disabled, the task will simply decline the checkpoint to the checkpoint coordinator and continue running.
 
+  - *is prefer checkpoint for recovery*: This determines if a job will fallback to checkpoint when there is a savepoint later than checkpoint.
```
gyfora: I dont know what's the difference between the checkpointing.zh.md and checkpointing.md files, do we need to change both?

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/environment/CheckpointConfig.java
```diff
@@ -285,6 +288,24 @@ public boolean isExternalizedCheckpointsEnabled() {
 		return externalizedCheckpointCleanup != null;
 	}
 
+	/**
+	 * Returns whether a job recovery should fallback to checkpoint when there is a savepoint later than checkpoint.
```
gyfora: I would replace in all the docs : " when there is a savepoint later than checkpoint" with "when there is a more recent savepoint"

gyfora:
I had a few smaller doc comments, otherwise the changes look pretty good. I dont know how the different doc files are handled so once thats cleared up and we squash the commits I think this is good to go :)

docs/dev/stream/state/checkpointing.zh.md
yanghua: The files with pattern `xxx.zh.md` is copied for translating purpose. They will be translated into Chinese. So for the correctness, I think change both would be a good solution.

yanghua:
@gyfora Thanks for correcting some representation problems. I will update the documentation. And I want to make sure that if I need to squash these commits or you would do it?

gyfora:
Once the content is correct you either squash it or I will do it myself when I merge, its up to you :) I think we can merge this today

yanghua:
@gyfora Thanks. I have refactored the java doc and flink doc based on your suggestion. In addition, I have also squashed commits and rebased the branch.

flink-runtime/src/main/java/org/apache/flink/runtime/checkpoint/StandaloneCompletedCheckpointStore.java
R-CodeBoy: Does it mean that the latest checkpoint is be surely  in the last position in list? thanks

yanghua: @R-CodeBoy Yes, the `getAllCheckpoints ` returns an ArrayList which wrapped an ArrayDeque.

R-CodeBoy: @yanghua Do you compare the cost time of recovery between checkpoint and savepoint,In my options ,sometimes it depends on the checkpoint interval when considering the recovery speed or reprocessing the data.

yanghua: Actually, it is a config option which depends on users themselves production scene. Some scene, the cost of rolling offset back is higher than recovery time, however, some scene is converse. To guess the scene is difficult to a general compute engine.

StephanEwen:
This PR unfortunately changes a few sensitive core components  (CheckpointCoordinator and CompletedCheckpointStore) and introduces a new configuration while pretty much violating all best practices and code style of Flink :-(

Just some examples from a quick look:
  - constant fields should be final in class
  - exception rethrown as RuntimeExceptions (when signature even allows exceptions)
  - mockito for testing
  - operational config in ExecutionConfig rather than in flink-conf.yaml
  - bad logging `LOG.error("Method getAllCheckpoints caused exception : ", e);`
  - default method used to misuse interface as abstract class
  - interface level logger instance (side effect of the above)
  - and various other things in the details...

I am inclined to push for reverting this commit and reopen the issue for another implementation attempt. The amount of code issues are too large for a patchup here, in my opinion.

Also, as a personal take: I think we need to put an emphasis on setting good examples for the community to improve the code quality in Flink. If we don't set good examples and gradually improve the code quality, Flink will be harder and harder to maintain in the future.

zentol:
The CompletedCheckpointStore is indeed quite sketchy; there are better ways to share logic ...

The mocking is _less_ problematic in my view given that the test files we're modifying are already filled to the brim with mocking (and IIRC the checkpoint coordinator is notoriously difficult to interact with in tests without mocking).
Naturally I'd prefer if we'd rework these tests entirely, but we can't just put this burden on the next guy to touch these tests. That said there are quite a few places though where we could mocking though this in simple ways. (e.g., there's no reason to mock a StreamStatehandle)

It also seems like the documentation is rather light on potential downsides to using this feature, which were discussed in the JIRA.

There's quite a bit of cleanup to be done here, as such I generally wouldn't oppose reverting it. However, there's a time constraint in play with the feature freeze coming up. Would we be fine for delaying this feature? If not, should we find a committer first to either review a cleaned up version, or even do the cleanup itself, before reverting it?

gyfora:
@StephanEwen @zentol I agree with most comments, and as far as the logging, exception handling goes it should have been reviewed better.

As for the comment related to operational config in ExecutionConfig. I think the main problem is the existence of CheckpointConfig. Most of the checkpointing related settings can only be configured there at the moment so it doesnt make sense to put something new only in flink-conf. That would be extremely confusing. I think we need to rework the whole CheckpointConfig logic to allow every single thing to be override from flink-conf.yaml.

I was not aware that we are trying to avoid interface default methods in the Flink codebase, I am sorry. I think they can make code much cleaner by removing the need to introduce abstract classes everywhere.

Overall I agree that we need to clean this up one way or the other, I am not opposed to reverting this before the 1.9.0 release.

PR #8412, opened 2019-05-10
[FLINK-12111][tests] Harden AbstractTaskManagerProcessFailureRecoveryTest

zentol:

Assortment of changes to improve/harden the `AbstractTaskManagerProcessFailureRecoveryTest`

## Brief change log

* removed unused field
* no longer sets `taskManagerProcess1` to null so that the process output is printed on failure
* wait until destroyed process has actually shut down
Prevents theoretical scenarios where the job can finish because the destroy() command takes a while to take effect.
* reduce number of initial TMs to 1,
The batch test could still succeed (if ExecutionMode == BATCH) even if the new TM was never used.
Reduce the number of initial TMs to 1 so that once that TM crashes all tasks MUST be moved to the new TM.
Doubled number of slots to compensate the loss of a TM.
* allow 2 restarts
For some reason this test could fail multiple times, instead of just once.

## Verifying this change

The issue with the BATCH execution mode could be reproduced easily (just skip the start of the third TM), and the change should fix this in an obvious way.

The restart fix is basically a shot in the dark + band-aid; ideally we would find the underlying cause.

### DIFF
--- a/flink-tests/src/test/java/org/apache/flink/test/recovery/AbstractTaskManagerProcessFailureRecoveryTest.java
+++ b/flink-tests/src/test/java/org/apache/flink/test/recovery/AbstractTaskManagerProcessFailureRecoveryTest.java
@@ -171,6 +171,7 @@ public abstract class AbstractTaskManagerProcessFailureRecoveryTest extends Test
 
 			// kill one of the previous TaskManagers, triggering a failure and recovery
 			taskManagerProcess1.destroy();
+			waitForShutdown("TaskManager 1", taskManagerProcess1);
 
 			// we create the marker file which signals the program functions tasks that they can complete
 			touchFile(new File(coordinateTempDir, PROCEED_MARKER_FILE));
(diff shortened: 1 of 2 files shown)


zentol:
It weakens the test in some regards (as this kind of weird timing issues is no longer covered) but strengthens it in other areas (BATCH mode being _actually_ tested).

I tried to reproduce the scenario you described by upping the heartbeat timeout, so that the network stack always fails first. This however didn't work; the restart was delayed since all tasks on the TM that timed out were stuck in a CANCELING state. Only once the heartbeat actually timed out could the restart proceed. As such I'm no longer sure whether this scenario can actually occur.

Upping the restart delay should work, will implement that right away.

tillrohrmann:
After an offline discussion with @zentol, we concluded that the following is happening most likely:

Killing the first TM process `TM1` does not happen synchronously. Therefore, it can happen that `TM1` sees the proceed marker file and finishes the computation. Next, the reducer is started and consumes the data from `TM1`. Now `TM1` is killed and the network stack signals the connection loss. Since the reducer is running with a parallelism of `1` and is deployed on `TM2`, the `ExecutionGraph` can be quickly restarted without waiting on the heartbeat to time out (because the `rpcTimeout` is set to `100 s` and the cancel attempts won't fail). Restarting the `ExecutionGraph` will result in deploying some tasks to `TM1` whose heartbeat hasn't been timed out. Last the heartbeat of `TM1` times out and the job fails a second time.

The proposed solution is the following:

* Wait for `TM1` to be terminated before creating the proceed marker file. That way the mappers running on `TM1` should never complete.
* Set the number of allowed restarts to `2` in order to allow for two job restarts.

PR #8413, opened 2019-05-10
[FLINK-12488][metrics] Pass Status group to NetworkEnvironment

zentol:

Pass the TaskManagers `Status` metric group to the NetworkEnvironment to retain API compatibility.

## Verifying this change

Run `MetricsAvailabilityITCase`.

### DIFF
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/taskexecutor/TaskManagerRunner.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/taskexecutor/TaskManagerRunner.java
@@ -20,12 +20,14 @@ package org.apache.flink.runtime.taskexecutor;
 
 import org.apache.flink.annotation.VisibleForTesting;
 import org.apache.flink.api.common.time.Time;
+import org.apache.flink.api.java.tuple.Tuple2;
 import org.apache.flink.configuration.Configuration;
 import org.apache.flink.configuration.ConfigurationUtils;
 import org.apache.flink.configuration.GlobalConfiguration;
 import org.apache.flink.configuration.TaskManagerOptions;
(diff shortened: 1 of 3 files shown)


flink-runtime/src/main/java/org/apache/flink/runtime/taskexecutor/TaskManagerRunner.java
```diff
@@ -358,15 +360,15 @@ public static TaskExecutor startTaskManager(
 				remoteAddress,
 				localCommunicationOnly);
 
-		TaskManagerMetricGroup taskManagerMetricGroup = MetricUtils.instantiateTaskManagerMetricGroup(
+		Tuple2<TaskManagerMetricGroup, MetricGroup> taskManagerMetricGroup = MetricUtils.instantiateTaskManagerMetricGroup(
```
azagrebin: I would consider moving:
```
MetricGroup statusGroup = taskManagerMetricGroup.addGroup(METRIC_GROUP_STATUS_NAME);
// Initialize the TM metrics			// Initialize the TM metrics
instantiateStatusMetrics(statusGroup);			instantiateStatusMetrics(statusGroup);
```
out of `MetricUtils.instantiateTaskManagerMetricGroup` into this method here
instead of having `Tuple2` return type, imho, it will be more readable.

zentol: This may be something we have to live with. The `MetricUtils` exists so that JM/TM are roughly in sync with each other in what metrics they expose and how the scope works.

Moving the status group initialization of the TM out of here makes this a lot less obvious.

PR #8416, opened 2019-05-10
[FLINK-12331] Introduce partition/gate setup to decouple task registration with NetworkEnvironment

azagrebin:

Introduce partition/gate setup to decouple task registration with NetworkEnvironment.
The PR also does some preparation refactoring.

## Brief change log

  - Introduce factories for ResultPartion and SingleInputgates
  - Introduce buffer pool factories for ResultPartion and SingleInputgates
  - Introduce MemorySegmentProvider for RemoteInputChannel to assign exclusive segments
  - Refactor NetworkEnvironment#setupXXX() to ResultPartitionWriter#setup() and InputGate#setup()

## Verifying this change

existing unit tests.

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-core/src/main/java/org/apache/flink/core/memory/MemorySegmentProvider.java
+++ b/flink-core/src/main/java/org/apache/flink/core/memory/MemorySegmentProvider.java
@@ -1,0 +1,31 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.core.memory;
+
+import java.io.IOException;
+import java.util.Collection;
+
+/**
+ * The provider used for requesting and releasing batch of memory segments.
+ */
+public interface MemorySegmentProvider {
+	Collection<MemorySegment> requestMemorySegments() throws IOException;
+
+	void recycleMemorySegments(Collection<MemorySegment> segments) throws IOException;
+}
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NetworkEnvironment.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NetworkEnvironment.java
@@ -40,6 +40,7 @@ import org.apache.flink.runtime.io.network.netty.NettyConfig;
 import org.apache.flink.runtime.io.network.netty.NettyConnectionManager;
 import org.apache.flink.runtime.io.network.partition.ResultPartition;
 import org.apache.flink.runtime.io.network.partition.ResultPartitionConsumableNotifier;
+import org.apache.flink.runtime.io.network.partition.ResultPartitionFactory;
 import org.apache.flink.runtime.io.network.partition.ResultPartitionID;
 import org.apache.flink.runtime.io.network.partition.ResultPartitionManager;
 import org.apache.flink.runtime.io.network.partition.consumer.SingleInputGate;
@@ -88,7 +89,7 @@ public class NetworkEnvironment {
 
 	private final TaskEventPublisher taskEventPublisher;
 
-	private final IOManager ioManager;
+	private final ResultPartitionFactory resultPartitionFactory;
 
 	private boolean isShutdown;
 
@@ -98,14 +99,14 @@ public class NetworkEnvironment {
 			ConnectionManager connectionManager,
 			ResultPartitionManager resultPartitionManager,
 			TaskEventPublisher taskEventPublisher,
-			IOManager ioManager) {
+			ResultPartitionFactory resultPartitionFactory) {
 		this.config = config;
 		this.networkBufferPool = networkBufferPool;
 		this.connectionManager = connectionManager;
 		this.resultPartitionManager = resultPartitionManager;
 		this.taskEventPublisher = taskEventPublisher;
-		this.ioManager = ioManager;
 		this.isShutdown = false;
+		this.resultPartitionFactory = resultPartitionFactory;
 	}
 
 	public static NetworkEnvironment create(
@@ -128,6 +129,8 @@ public class NetworkEnvironment {
 		registerNetworkMetrics(metricGroup, networkBufferPool);
 
 		ResultPartitionManager resultPartitionManager = new ResultPartitionManager();
+		ResultPartitionFactory resultPartitionFactory =
+			new ResultPartitionFactory(resultPartitionManager, ioManager);
 
 		return new NetworkEnvironment(
 			config,
@@ -135,7 +138,7 @@ public class NetworkEnvironment {
 			connectionManager,
 			resultPartitionManager,
 			taskEventPublisher,
-			ioManager);
+			resultPartitionFactory);
 	}
 
 	private static void registerNetworkMetrics(MetricGroup metricGroup, NetworkBufferPool networkBufferPool) {
@@ -283,18 +286,8 @@ public class NetworkEnvironment {
 			ResultPartition[] resultPartitions = new ResultPartition[resultPartitionDeploymentDescriptors.size()];
 			int counter = 0;
 			for (ResultPartitionDeploymentDescriptor rpdd : resultPartitionDeploymentDescriptors) {
-				resultPartitions[counter++] = new ResultPartition(
-					taskName,
-					taskActions,
-					jobId,
-					new ResultPartitionID(rpdd.getPartitionId(), executionId),
-					rpdd.getPartitionType(),
-					rpdd.getNumberOfSubpartitions(),
-					rpdd.getMaxParallelism(),
-					resultPartitionManager,
-					partitionConsumableNotifier,
-					ioManager,
-					rpdd.sendScheduleOrUpdateConsumersMessage());
+				resultPartitions[counter++] = resultPartitionFactory.create(
+					taskName, taskActions, jobId, executionId, rpdd, partitionConsumableNotifier);
 			}
 
 			registerOutputMetrics(outputGroup, buffersGroup, resultPartitions);
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NetworkEnvironment.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NetworkEnvironment.java
@@ -138,8 +138,8 @@ public class NetworkEnvironment {
 		ResultPartitionFactory resultPartitionFactory =
 			new ResultPartitionFactory(resultPartitionManager, ioManager);
 
-		SingleInputGateFactory singleInputGateFactory =
-			new SingleInputGateFactory(config, connectionManager, resultPartitionManager, taskEventPublisher);
+		SingleInputGateFactory singleInputGateFactory = new SingleInputGateFactory(
+			config, connectionManager, resultPartitionManager, taskEventPublisher, networkBufferPool);
 
 		return new NetworkEnvironment(
 			config,
@@ -247,7 +247,7 @@ public class NetworkEnvironment {
 					config.floatingNetworkBuffersPerGate() : Integer.MAX_VALUE;
 
 				// assign exclusive buffers to input channels directly and use the rest for floating buffers
-				gate.assignExclusiveSegments(networkBufferPool);
+				gate.assignExclusiveSegments();
 				bufferPool = networkBufferPool.createBufferPool(0, maxNumberOfMemorySegments);
 			} else {
 				maxNumberOfMemorySegments = gate.getConsumedPartitionType().isBounded() ?
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NetworkEnvironment.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NetworkEnvironment.java
@@ -129,7 +129,8 @@ public class NetworkEnvironment {
 
 		NetworkBufferPool networkBufferPool = new NetworkBufferPool(
 			config.numNetworkBuffers(),
-			config.networkBufferSize());
+			config.networkBufferSize(),
+			config.networkBuffersPerChannel());
 
 		registerNetworkMetrics(metricGroup, networkBufferPool);
 
@@ -246,7 +247,7 @@ public class NetworkEnvironment {
 					config.floatingNetworkBuffersPerGate() : Integer.MAX_VALUE;
 
 				// assign exclusive buffers to input channels directly and use the rest for floating buffers
-				gate.assignExclusiveSegments(networkBufferPool, config.networkBuffersPerChannel());
+				gate.assignExclusiveSegments(networkBufferPool);
 				bufferPool = networkBufferPool.createBufferPool(0, maxNumberOfMemorySegments);
 			} else {
 				maxNumberOfMemorySegments = gate.getConsumedPartitionType().isBounded() ?
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NetworkEnvironment.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NetworkEnvironment.java
@@ -57,7 +57,6 @@ import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.util.Collection;
-import java.util.Optional;
 
 import static org.apache.flink.util.Preconditions.checkNotNull;
 
@@ -135,11 +134,19 @@ public class NetworkEnvironment {
 		registerNetworkMetrics(metricGroup, networkBufferPool);
 
 		ResultPartitionManager resultPartitionManager = new ResultPartitionManager();
-		ResultPartitionFactory resultPartitionFactory =
-			new ResultPartitionFactory(resultPartitionManager, ioManager);
+		ResultPartitionFactory resultPartitionFactory = new ResultPartitionFactory(
+			resultPartitionManager,
+			ioManager,
+			networkBufferPool,
+			config.networkBuffersPerChannel(),
+			config.floatingNetworkBuffersPerGate());
 
 		SingleInputGateFactory singleInputGateFactory = new SingleInputGateFactory(
-			config, connectionManager, resultPartitionManager, taskEventPublisher, networkBufferPool);
+			config,
+			connectionManager,
+			resultPartitionManager,
+			taskEventPublisher,
+			networkBufferPool);
 
 		return new NetworkEnvironment(
 			config,
@@ -196,43 +203,13 @@ public class NetworkEnvironment {
 			}
 
 			for (final ResultPartition partition : producedPartitions) {
-				setupPartition(partition);
+				partition.setup();
 			}
 
 			// Setup the buffer pool for each buffer reader
 			final SingleInputGate[] inputGates = task.getAllInputGates();
 			for (SingleInputGate gate : inputGates) {
 				setupInputGate(gate);
-			}
-		}
-	}
-
-	@VisibleForTesting
-	public void setupPartition(ResultPartition partition) throws IOException {
-		BufferPool bufferPool = null;
-
-		try {
-			int maxNumberOfMemorySegments = partition.getPartitionType().isBounded() ?
-				partition.getNumberOfSubpartitions() * config.networkBuffersPerChannel() +
-					config.floatingNetworkBuffersPerGate() : Integer.MAX_VALUE;
-			// If the partition type is back pressure-free, we register with the buffer pool for
-			// callbacks to release memory.
-			bufferPool = networkBufferPool.createBufferPool(partition.getNumberOfSubpartitions(),
-				maxNumberOfMemorySegments,
-				partition.getPartitionType().hasBackPressure() ? Optional.empty() : Optional.of(partition));
-
-			partition.registerBufferPool(bufferPool);
-
-			resultPartitionManager.registerResultPartition(partition);
-		} catch (Throwable t) {
-			if (bufferPool != null) {
-				bufferPool.lazyDestroy();
-			}
-
-			if (t instanceof IOException) {
-				throw (IOException) t;
-			} else {
-				throw new IOException(t.getMessage(), t);
 			}
 		}
 	}
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NetworkEnvironment.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NetworkEnvironment.java
@@ -27,7 +27,6 @@ import org.apache.flink.runtime.deployment.InputGateDeploymentDescriptor;
 import org.apache.flink.runtime.deployment.ResultPartitionDeploymentDescriptor;
 import org.apache.flink.runtime.executiongraph.ExecutionAttemptID;
 import org.apache.flink.runtime.io.disk.iomanager.IOManager;
-import org.apache.flink.runtime.io.network.buffer.BufferPool;
 import org.apache.flink.runtime.io.network.buffer.NetworkBufferPool;
 import org.apache.flink.runtime.io.network.metrics.InputBufferPoolUsageGauge;
 import org.apache.flink.runtime.io.network.metrics.InputBuffersGauge;
@@ -47,9 +46,7 @@ import org.apache.flink.runtime.io.network.partition.consumer.SingleInputGate;
 import org.apache.flink.runtime.io.network.partition.consumer.SingleInputGateFactory;
 import org.apache.flink.runtime.taskexecutor.TaskExecutor;
 import org.apache.flink.runtime.taskmanager.NetworkEnvironmentConfiguration;
-import org.apache.flink.runtime.taskmanager.Task;
 import org.apache.flink.runtime.taskmanager.TaskActions;
-import org.apache.flink.util.ExceptionUtils;
 import org.apache.flink.util.Preconditions;
 
 import org.slf4j.Logger;
@@ -188,60 +185,6 @@ public class NetworkEnvironment {
 	@VisibleForTesting
 	public NetworkEnvironmentConfiguration getConfiguration() {
 		return config;
-	}
-
-	// --------------------------------------------------------------------------------------------
-	//  Task operations
-	// --------------------------------------------------------------------------------------------
-
-	public void registerTask(Task task) throws IOException {
-		final ResultPartition[] producedPartitions = task.getProducedPartitions();
-
-		synchronized (lock) {
-			if (isShutdown) {
-				throw new IllegalStateException("NetworkEnvironment is shut down");
-			}
-
-			for (final ResultPartition partition : producedPartitions) {
-				partition.setup();
-			}
-
-			// Setup the buffer pool for each buffer reader
-			final SingleInputGate[] inputGates = task.getAllInputGates();
-			for (SingleInputGate gate : inputGates) {
-				setupInputGate(gate);
-			}
-		}
-	}
-
-	@VisibleForTesting
-	public void setupInputGate(SingleInputGate gate) throws IOException {
-		BufferPool bufferPool = null;
-		int maxNumberOfMemorySegments;
-		try {
-			if (config.isCreditBased()) {
-				maxNumberOfMemorySegments = gate.getConsumedPartitionType().isBounded() ?
-					config.floatingNetworkBuffersPerGate() : Integer.MAX_VALUE;
-
-				// assign exclusive buffers to input channels directly and use the rest for floating buffers
-				gate.assignExclusiveSegments();
-				bufferPool = networkBufferPool.createBufferPool(0, maxNumberOfMemorySegments);
-			} else {
-				maxNumberOfMemorySegments = gate.getConsumedPartitionType().isBounded() ?
-					gate.getNumberOfInputChannels() * config.networkBuffersPerChannel() +
-						config.floatingNetworkBuffersPerGate() : Integer.MAX_VALUE;
-
-				bufferPool = networkBufferPool.createBufferPool(gate.getNumberOfInputChannels(),
-					maxNumberOfMemorySegments);
-			}
-			gate.setBufferPool(bufferPool);
-		} catch (Throwable t) {
-			if (bufferPool != null) {
-				bufferPool.lazyDestroy();
-			}
-
-			ExceptionUtils.rethrowIOException(t);
-		}
 	}
 
 	/**
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NetworkEnvironment.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NetworkEnvironment.java
@@ -44,6 +44,7 @@ import org.apache.flink.runtime.io.network.partition.ResultPartitionFactory;
 import org.apache.flink.runtime.io.network.partition.ResultPartitionID;
 import org.apache.flink.runtime.io.network.partition.ResultPartitionManager;
 import org.apache.flink.runtime.io.network.partition.consumer.SingleInputGate;
+import org.apache.flink.runtime.io.network.partition.consumer.SingleInputGateFactory;
 import org.apache.flink.runtime.taskexecutor.TaskExecutor;
 import org.apache.flink.runtime.taskmanager.NetworkEnvironmentConfiguration;
 import org.apache.flink.runtime.taskmanager.Task;
@@ -91,6 +92,8 @@ public class NetworkEnvironment {
 
 	private final ResultPartitionFactory resultPartitionFactory;
 
+	private final SingleInputGateFactory singleInputGateFactory;
+
 	private boolean isShutdown;
 
 	private NetworkEnvironment(
@@ -99,14 +102,16 @@ public class NetworkEnvironment {
 			ConnectionManager connectionManager,
 			ResultPartitionManager resultPartitionManager,
 			TaskEventPublisher taskEventPublisher,
-			ResultPartitionFactory resultPartitionFactory) {
+			ResultPartitionFactory resultPartitionFactory,
+			SingleInputGateFactory singleInputGateFactory) {
 		this.config = config;
 		this.networkBufferPool = networkBufferPool;
 		this.connectionManager = connectionManager;
 		this.resultPartitionManager = resultPartitionManager;
 		this.taskEventPublisher = taskEventPublisher;
-		this.isShutdown = false;
 		this.resultPartitionFactory = resultPartitionFactory;
+		this.singleInputGateFactory = singleInputGateFactory;
+		this.isShutdown = false;
 	}
 
 	public static NetworkEnvironment create(
@@ -132,13 +137,17 @@ public class NetworkEnvironment {
 		ResultPartitionFactory resultPartitionFactory =
 			new ResultPartitionFactory(resultPartitionManager, ioManager);
 
+		SingleInputGateFactory singleInputGateFactory =
+			new SingleInputGateFactory(config, connectionManager, resultPartitionManager, taskEventPublisher);
+
 		return new NetworkEnvironment(
 			config,
 			networkBufferPool,
 			connectionManager,
 			resultPartitionManager,
 			taskEventPublisher,
-			resultPartitionFactory);
+			resultPartitionFactory,
+			singleInputGateFactory);
 	}
 
 	private static void registerNetworkMetrics(MetricGroup metricGroup, NetworkBufferPool networkBufferPool) {
@@ -168,6 +177,7 @@ public class NetworkEnvironment {
 		return networkBufferPool;
 	}
 
+	@VisibleForTesting
 	public NetworkEnvironmentConfiguration getConfiguration() {
 		return config;
 	}
@@ -311,12 +321,10 @@ public class NetworkEnvironment {
 			SingleInputGate[] inputGates = new SingleInputGate[inputGateDeploymentDescriptors.size()];
 			int counter = 0;
 			for (InputGateDeploymentDescriptor igdd : inputGateDeploymentDescriptors) {
-				inputGates[counter++] = SingleInputGate.create(
+				inputGates[counter++] = singleInputGateFactory.create(
 					taskName,
 					jobId,
 					igdd,
-					this,
-					taskEventPublisher,
 					taskActions,
 					inputChannelMetrics,
 					numBytesInCounter);
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NetworkEnvironment.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NetworkEnvironment.java
@@ -92,31 +92,50 @@ public class NetworkEnvironment {
 
 	private boolean isShutdown;
 
-	public NetworkEnvironment(
+	private NetworkEnvironment(
+			NetworkEnvironmentConfiguration config,
+			NetworkBufferPool networkBufferPool,
+			ConnectionManager connectionManager,
+			ResultPartitionManager resultPartitionManager,
+			TaskEventPublisher taskEventPublisher,
+			IOManager ioManager) {
+		this.config = config;
+		this.networkBufferPool = networkBufferPool;
+		this.connectionManager = connectionManager;
+		this.resultPartitionManager = resultPartitionManager;
+		this.taskEventPublisher = taskEventPublisher;
+		this.ioManager = ioManager;
+		this.isShutdown = false;
+	}
+
+	public static NetworkEnvironment create(
 			NetworkEnvironmentConfiguration config,
 			TaskEventPublisher taskEventPublisher,
 			MetricGroup metricGroup,
 			IOManager ioManager) {
-		this.config = checkNotNull(config);
-
-		this.networkBufferPool = new NetworkBufferPool(config.numNetworkBuffers(), config.networkBufferSize());
+		checkNotNull(ioManager);
+		checkNotNull(taskEventPublisher);
+		checkNotNull(config);
 
 		NettyConfig nettyConfig = config.nettyConfig();
-		if (nettyConfig != null) {
-			this.connectionManager = new NettyConnectionManager(nettyConfig, config.isCreditBased());
-		} else {
-			this.connectionManager = new LocalConnectionManager();
-		}
+		ConnectionManager connectionManager = nettyConfig != null ?
+			new NettyConnectionManager(nettyConfig, config.isCreditBased()) : new LocalConnectionManager();
 
-		this.resultPartitionManager = new ResultPartitionManager();
-
-		this.taskEventPublisher = checkNotNull(taskEventPublisher);
+		NetworkBufferPool networkBufferPool = new NetworkBufferPool(
+			config.numNetworkBuffers(),
+			config.networkBufferSize());
 
 		registerNetworkMetrics(metricGroup, networkBufferPool);
 
-		this.ioManager = checkNotNull(ioManager);
+		ResultPartitionManager resultPartitionManager = new ResultPartitionManager();
 
-		isShutdown = false;
+		return new NetworkEnvironment(
+			config,
+			networkBufferPool,
+			connectionManager,
+			resultPartitionManager,
+			taskEventPublisher,
+			ioManager);
 	}
 
 	private static void registerNetworkMetrics(MetricGroup metricGroup, NetworkBufferPool networkBufferPool) {
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/buffer/NetworkBufferPool.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/buffer/NetworkBufferPool.java
@@ -21,6 +21,7 @@ package org.apache.flink.runtime.io.network.buffer;
 import org.apache.flink.configuration.TaskManagerOptions;
 import org.apache.flink.core.memory.MemorySegment;
 import org.apache.flink.core.memory.MemorySegmentFactory;
+import org.apache.flink.core.memory.MemorySegmentProvider;
 import org.apache.flink.util.ExceptionUtils;
 import org.apache.flink.util.MathUtils;
 
@@ -31,6 +32,7 @@ import javax.annotation.Nullable;
 
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Optional;
@@ -49,7 +51,7 @@ import static org.apache.flink.util.Preconditions.checkNotNull;
  * the buffers for the network data transfer. When new local buffer pools are created, the
  * NetworkBufferPool dynamically redistributes the buffers between the pools.
  */
-public class NetworkBufferPool implements BufferPoolFactory {
+public class NetworkBufferPool implements BufferPoolFactory, MemorySegmentProvider {
 
 	private static final Logger LOG = LoggerFactory.getLogger(NetworkBufferPool.class);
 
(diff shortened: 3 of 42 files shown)


azagrebin:

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NetworkEnvironment.java
```diff
@@ -98,15 +99,15 @@ private NetworkEnvironment(
...
+		ResultPartitionFactory resultPartitionFactory) {
 
 		this.config = config;
 		this.networkBufferPool = networkBufferPool;
 		this.connectionManager = connectionManager;
 		this.resultPartitionManager = resultPartitionManager;
 		this.taskEventPublisher = taskEventPublisher;
-		this.ioManager = ioManager;
 		this.isShutdown = false;
+		this.resultPartitionFactory = resultPartitionFactory;
```
zhijiangW: `this.resultPartitionFactory = resultPartitionFactory` in front of `this.isShutdown = false`?

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NetworkEnvironment.java
```diff
@@ -108,6 +112,7 @@ private NetworkEnvironment(
 		this.taskEventPublisher = taskEventPublisher;
 		this.isShutdown = false;
```
zhijiangW: Put `isShutdown=false` at the end of this constructor?

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NetworkEnvironment.java
```diff
@@ -128,13 +133,17 @@ public static NetworkEnvironment create(
 		ResultPartitionFactory resultPartitionFactory =
 			new ResultPartitionFactory(resultPartitionManager, checkNotNull(ioManager));
 
+		SingleInputGateFactory singleInputGateFactory =
+			new SingleInputGateFactory(config, connectionManager, resultPartitionManager);
```
zhijiangW: `TaskEventDispatcher` could also be included in `singleInputGateFactory`?

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/consumer/SingleInputGateTest.java
```diff
@@ -415,16 +395,19 @@ public void testRequestBuffersWithRemoteInputChannel() throws Exception {
...
 
 		try {
-			final ResultPartitionID resultPartitionId = new ResultPartitionID();
-			final ConnectionID connectionId = new ConnectionID(new InetSocketAddress("localhost", 5000), 0);
-			addRemoteInputChannel(network, inputGate, connectionId, resultPartitionId, 0);
+			RemoteInputChannel remoteFromUnknown =
+				InputChannelBuilder.newBuilder()
+					.setupFromNetworkEnvironment(network)
+					.buildUnknown(inputGate)
+					.toRemoteInputChannel(InputChannelBuilder.STUB_CONNECTION_ID);
```
zhijiangW: why not build remote directly?

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/consumer/SingleInputGateTest.java
```diff
@@ -415,16 +395,19 @@ public void testRequestBuffersWithRemoteInputChannel() throws Exception {
...
+			IntermediateResultPartitionID resultPartitionId = remoteFromUnknown.getPartitionId().getPartitionId();
+			inputGate.setInputChannel(resultPartitionId, remoteFromUnknown);
```
zhijiangW: `setInputChannel` could be removed?

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/consumer/SingleInputGateTest.java
```diff
@@ -415,16 +395,19 @@ public void testRequestBuffersWithRemoteInputChannel() throws Exception {
...
 
 			network.setupInputGate(inputGate);
 
 			NetworkBufferPool bufferPool = network.getNetworkBufferPool();
 			if (enableCreditBasedFlowControl) {
-				RemoteInputChannel remote = (RemoteInputChannel) inputGate.getInputChannels()
-					.get(resultPartitionId.getPartitionId());
+				RemoteInputChannel remote = (RemoteInputChannel) inputGate.getInputChannels().get(resultPartitionId);
```
zhijiangW: Use the previous `remoteFromUnknown` directly, so `IntermediateResultPartitionID resultPartitionId = remoteFromUnknown.getPartitionId().getPartitionId()` could also be removed.

flink-core/src/main/java/org/apache/flink/core/memory/MemorySegmentProvider.java
```diff
@@ -0,0 +1,34 @@
...
+import java.util.Collection;
+
+/**
+ * The provider used for requesting and releasing batch of memory segments.
+ */
+public interface MemorySegmentProvider {
+	/** Requests default number of memory segments. */
+	Collection<MemorySegment> requestMemorySegments() throws IOException;
+
+	Collection<MemorySegment> requestMemorySegments(int numRequiredBuffers) throws IOException;
```
zhijiangW: If we only consider the semantics of `MemorySegmentProvider` interface, the `requestMemorySegments()` is a bit redundant, which could be replaced by `requestMemorySegments(int numRequiredBuffers)`. The only reason we keep both methods is that it seems not elegant to maintain `networkBuffersPerChannel` in `SingleInputGate`.

TBH, I do not think it is worth defining this new interface not very clean to accommodate the implementation. Because `requestMemorySegments()` seems blurry for the usage which relies on the default value, and we should also consider if the default value is 0, whether the returned collection is null or empty size. 
In addition, `requestMemorySegments(int numRequiredBuffers)` is only used for tests ATM.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/consumer/SingleInputGate.java
```diff
@@ -155,8 +155,8 @@
 	 */
 	private BufferPool bufferPool;
 
-	/** Global network buffer pool to request and recycle exclusive buffers (only for credit-based). */
-	private NetworkBufferPool networkBufferPool;
+	/** Global global memory segment provider to request and recycle exclusive buffers (only for credit-based). */
```
zhijiangW: repeated `global`

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/consumer/SingleInputGate.java
```diff
@@ -535,7 +510,7 @@ public void requestPartitions() throws IOException, InterruptedException {
 			throws IOException, InterruptedException {
 		while (true) {
 			synchronized (inputChannelsWithData) {
-				while (inputChannelsWithData.size() == 0) {
+				while (inputChannelsWithData.isEmpty()) {
```
zhijiangW: why make this change?

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/consumer/SingleInputGateFactory.java
```diff
@@ -121,8 +127,8 @@ else if (partitionLocation.isRemote()) {
 					connectionManager,
 					partitionRequestInitialBackoff,
 					partitionRequestMaxBackoff,
-					metrics
-				);
+					metrics,
+					memorySegmentProvider);
```
zhijiangW: Whether to separate line for last `);` which keeps the consistent with `local` and `unknown` cases?

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/netty/PartitionRequestClientHandlerTest.java
```diff
@@ -203,14 +205,20 @@ public void testCancelBeforeActive() throws Exception {
...
 	 *
 	 * @param inputGate The input gate owns the created input channel.
 	 * @return The new created remote input channel.
 	 */
-	static RemoteInputChannel createRemoteInputChannel(SingleInputGate inputGate) throws Exception {
-		return createRemoteInputChannel(inputGate, mock(PartitionRequestClient.class));
+	static RemoteInputChannel createRemoteInputChannel(
+		SingleInputGate inputGate, MemorySegmentProvider memorySegmentProvider) throws Exception {
+
+		return createRemoteInputChannel(inputGate, mock(PartitionRequestClient.class), 0, 0, memorySegmentProvider);
```
zhijiangW: It is better to avoid mock `PartitionRequestClient` here.

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/netty/PartitionRequestClientHandlerTest.java
```diff
@@ -237,7 +254,8 @@ static RemoteInputChannel createRemoteInputChannel(
 			SingleInputGate inputGate,
 			PartitionRequestClient client,
 			int initialBackoff,
-			int maxBackoff) throws Exception {
+			int maxBackoff,
+			MemorySegmentProvider memorySegmentProvider) throws Exception {
 		final ConnectionManager connectionManager = mock(ConnectionManager.class);
 		when(connectionManager.createPartitionRequestClient(any(ConnectionID.class)))
```
zhijiangW: There are already too many utils here to construct the `RemoteInputChannel`, but we still miss the units of `{SingleInputGate, PartitionRequestClient, MemorySegmentProvider}` which could avoid pass `createRemoteInputChannel(inputGate, client, 0, 0, networkBufferPool)` in `CreditBasedPartitionRequestClientHandlerTest`. 

Maybe we need a separate hotfix for refactoring these parts to make `connectionManager.createPartitionRequestClient` call in separate tests if necessary, then we only need one type to construct `RemoteInputChannel`.

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/consumer/InputChannelBuilder.java
```diff
@@ -46,6 +48,7 @@
 	private int maxBackoff = 0;
 	private InputChannelMetrics metrics =
 		new InputChannelMetrics(UnregisteredMetricGroups.createUnregisteredTaskMetricGroup());
```
zhijiangW: Here we could use `InputChannelTestUtils.newUnregisteredInputChannelMetrics`

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/consumer/RemoteInputChannelTest.java
```diff
@@ -1010,8 +1032,9 @@ private RemoteInputChannel createRemoteInputChannel(
 
 		return InputChannelBuilder.newBuilder()
 			.setConnectionManager(connectionManager)
-			.setInitialBackoff(initialBackoff)
+			.setInitialBackoff(initialBackOff)
```
zhijiangW: No need change?

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/consumer/RemoteInputChannelTest.java
```diff
@@ -992,16 +993,37 @@ public void testConcurrentRecycleAndRelease2() throws Exception {
...
+		throws IOException, InterruptedException {
+
+		return createRemoteInputChannel(
+				inputGate, mock(PartitionRequestClient.class), 0, 0, memorySegmentProvider);
+	}
+
+	private RemoteInputChannel createRemoteInputChannel(
+		SingleInputGate inputGate,
+		PartitionRequestClient partitionRequestClient,
+		int initialBackOff,
```
zhijiangW: lowercase `off` to keep consistent with maxBackoff?

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/consumer/RemoteInputChannelTest.java
```diff
@@ -992,16 +993,37 @@ public void testConcurrentRecycleAndRelease2() throws Exception {
...
+	}
 
-		return createRemoteInputChannel(inputGate, mock(PartitionRequestClient.class), 0, 0);
+	private RemoteInputChannel createRemoteInputChannel(
+		SingleInputGate inputGate,
+		MemorySegmentProvider memorySegmentProvider)
+		throws IOException, InterruptedException {
+
+		return createRemoteInputChannel(
+				inputGate, mock(PartitionRequestClient.class), 0, 0, memorySegmentProvider);
```
zhijiangW: better avoid mock `PartitionRequestClient`.

The same issue for creating `RemoteInputChannel` in various units, if we do not need unify the `createPartitionRequestClient` with the constructing of `RemoteInputChannel`, all these `createRemoteInputChannel` could be covered by one `InputChannelBuilder`.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/ResultPartition.java
```diff
@@ -181,13 +188,16 @@ public ResultPartition(
...
-	public void registerBufferPool(BufferPool bufferPool) {
+	@Override
+	public void setup() throws IOException {
+		BufferPool bufferPool = resultPartitionBufferPoolFactory.create(
+			getNumberOfSubpartitions(), getPartitionType(), this);
 		checkArgument(bufferPool.getNumberOfRequiredMemorySegments() >= getNumberOfSubpartitions(),
-				"Bug in result partition setup logic: Buffer pool has not enough guaranteed buffers for this result partition.");
+			"Bug in result partition setup logic: Buffer pool has not enough guaranteed buffers for this result partition.");
 
 		checkState(this.bufferPool == null, "Bug in result partition setup logic: Already registered buffer pool.");
```
zhijiangW: this check should be put before `resultPartitionBufferPoolFactory.create`

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/Task.java
```diff
@@ -836,6 +833,21 @@ else if (transitionState(current, ExecutionState.FAILED, t)) {
 		}
 	}
 
+	private void setupPartionsAndGates() throws IOException {
```
zhijiangW: why not remove this method to use the following static directly?

flink-runtime/src/test/java/org/apache/flink/runtime/taskmanager/TaskTest.java
```diff
@@ -262,7 +268,29 @@ public void testExecutionFailsInBlobsMissing() throws Exception {
 	}
 
 	@Test
-	public void testExecutionFailsInNetworkRegistration() throws Exception {
+	public void testExecutionFailsInNetworkRegistrationFprPartitions() throws Exception {
```
zhijiangW: Fpr->For

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/PartitionBufferPoolFactory.java
```diff
@@ -0,0 +1,31 @@
...
+
+package org.apache.flink.runtime.io.network;
+
+import org.apache.flink.runtime.io.network.buffer.BufferPool;
+import org.apache.flink.runtime.io.network.buffer.BufferPoolOwner;
+import org.apache.flink.runtime.io.network.partition.ResultPartitionType;
+
+import java.io.IOException;
+
+/** factory of {@link BufferPool} for network partitions or channels. */
```
zhijiangW: This comment annotation /** */ is used for variable, and the class should be like following
/**
* 
*/
I am not sure of it.

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/consumer/SingleInputGateTest.java
azagrebin: I tried to preserve the previous way of building remote but looking at it more, true, it does not seem to be relevant.

flink-core/src/main/java/org/apache/flink/core/memory/MemorySegmentProvider.java
azagrebin: I think `MemorySegmentProvider` makes sense from point of interface segregation principle. This interface encapsulates segment management concern which is the only thing that remote channel needs atm. `NetworkBufferPool` looks like implementing 2 separate concerns: `BufferPoolFactory` and `MemorySegmentProvider`.

As for `requestMemorySegments(int numRequiredBuffers)`, you are right, it is more clear to remove it from the interface and adjust the tests with explicit default value, e.g. 1.

azagrebin:
Thanks for the review @zhijiangW !
I addressed the smaller comments. The order of commits seems to have improved now :)

- As for `PartitionBufferPoolFactory`, it is specific for partitions and `BufferPoolFactory` is a general one. `PartitionBufferPoolFactory` can be also seen as wrapping `BufferPoolFactory` with the partition concern. I do not think we should include partition type or something to the `BufferPoolFactory`. The `PartitionBufferPoolFactory.create` basically transforms partition type and network configuration into general parameters of `BufferPoolFactory` so I think they belong to different level of abstractions. Alternatively, we could just add more config parameters to partition/gate constructors (config and `BufferPoolFactory`) and keep `PartitionBufferPoolFactory` logic inside but this would explode the constructors and partition/gate classes even more. It should not hurt to have each class to address as less concerns as possible.

- Commits for `PartitionBufferPoolFactory` factories address a certain concern, independent from setup commits. They can be compiled and tested independently as well. Not sure how temporary they are. At least, it should simplify review. We can of course squash them before merge.

- `ResultPartitionFactory` encapsulates all components which are needed to create any partition. `ResultPartitionFactory.create` accepts only partition specific parameters. Overall I think it should reduce creation of partition in place and should not hurt to have this concern separately. I will also move creation of subpartitions into the factory.

zhijiangW:
@azagrebin thanks for the confirmation.

As for `PartitionBufferPoolFactory`:

- The current `BufferPoolFactory` exists only for partition/gate, and no other parts would need `BufferPool` in flink stack atm. The new introduced `PartitionBufferPoolFactory` is also for partition/gate, but has no relationship with `BufferPoolFactory` in the architecture form. So I think it is not very specific for existing two parallel interfaces related with `BufferPool`.

- We could also pass the `ResultPartitionType` in constructor of `AbstractPartitionBufferPoolFactory`, then the new proposed interface method is shown as `create(int size, BufferPoolOwner owner)`, so the only difference with previous methods in `BufferPoolFactory` is we do not pass the `maxUsedBuffers` which is calculated based on `ResultPartitionType`.

- The key motivation of proposing new interface is we could only pass one parameter factory in constructor of `ResultPartition`, otherwise we need two parameters of `NetworkEnvironmentConfiguration` and `BufferPoolFactory`. But for the constructor of `SingleInputGate`, it seems the same because we could use `NetworkEnvironmentConfiguration` to replace the current `isCreditBased` parameter. 

- Another possible advantage of new interface is we extract the logic of creating `BufferPool` to another place, then it makes simple for `SingleInputGate`,`ResultPartition` and `NetworkEnvironment`.

- From the aspect of extended architecture, it seems no problem if we did nothing. Because different `ShuffleService` implementations might need different `InputGate` and `ResultPartitionWriter` implementations. Even we might extract the current `BlockingSubpartition` as a separate implementation of `ResultPartitionWriter` in future, and it does not need `BufferPoolFactory` any more.

In summary I do not think it is very necessary to introduce new interface here, or not very worthing.  If we want to make constructor of partition/gate simple, we could only pass `Supplier<BufferPool>` as you suggested before, and it could be explained by `NetworkEnvironment` itself. If the way of `Supplier` seems not good, I have another option:

- Add new method `createBufferPool(int numRequiredBuffers, BufferPoolOwner owner)` in existing `BufferPoolFactory`.

- Make `AbstractPartitionBufferPoolFactory` implements `BufferPoolFactory` and we could also pass `ResultPartitionType` into constructor of `AbstractPartitionBufferPoolFactory`.

- `ResultPartitionBufferPoolFactory/InputGateBufferPoolFactory` extends `AbstractPartitionBufferPoolFactory` and they only need to implement the new method `createBufferPool(int numRequiredBuffers, BufferPoolOwner owner)`.

To do so we extend and reuse the general `BufferPoolFactory` and only abstract/distinguish the implementation of calculating `maxPoolSize` in partition/gate. What do you think?

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NetworkEnvironment.java
```diff
@@ -151,7 +159,7 @@ public static NetworkEnvironment create(
 			checkNotNull(taskEventPublisher),
 			resultPartitionFactory,
 			singleInputGateFactory,
-			resultPartitionBufferPoolFactory);
+			resultPartitionBufferPoolFactory, inputGateBufferPoolFactory);
```
zhijiangW: inputGateBufferPoolFactory should be in separate line for the commit of `[hotfix][network] Introduce InputGateBufferPoolFactory`

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/PartitionBufferPoolFactory.java
```diff
@@ -0,0 +1,33 @@
...
+import org.apache.flink.runtime.io.network.buffer.BufferPoolOwner;
+import org.apache.flink.runtime.io.network.partition.ResultPartitionType;
+
+import java.io.IOException;
+
+/**
+ * Factory of {@link BufferPool} for network partitions or channels.
+ */
+public interface PartitionBufferPoolFactory {
+	BufferPool create(int size, ResultPartitionType type, BufferPoolOwner owner) throws IOException;
```
zhijiangW: empty line before `create` method?

zhijiangW:
Commits for `PartitionBufferPoolFactory`: In my previous review I saw it was not in the constructor of `NetworkEnvironment` in that commit, and then changed into the constructor in the following commit. So I mean the first introduced commit should be squashed with following commit to make sure it is in the constructor when introduced. Now I see it is already in right state, maybe some UI issue for commit sequence you confirmed before. :)

zhijiangW:
As for `ResultPartitionFactory`: The `ShuffleService` already takes the role of factory for creating partition/gate. Of course we could further extract a specific factory component from `ShuffleService`. But I think the factory might be different for different `ShuffleService` implementations, and it is regarded as a private tool inside specific `ShuffleService` implementation. 

E.g. for `FSShuffleService#FSResultPartition`, we might do not need the institute of `subpartition`, only need the logic `subpartition` index. Also for interactive queries, if it wants to reuse the `ResultPartitionWriter`, it also does not need to construct `subpartition` inside specific instance.

flink-core/src/main/java/org/apache/flink/core/memory/MemorySegmentProvider.java
zhijiangW: We are indeed in the lack of semantics of requesting/releasing batch of buffers before, and I think it is reasonable to extend these methods. 

- `SegmentProvider` is not really needed because `RemoteInputChannel` still needs construct the `Buffer` internally after requesting `MemorySegment`.

- We could extend the general `BufferProvider` interface to provide method of `requestBuffers(int numRequiredBuffers, BufferRecycler)`,  and `RemoteInputChannel` already implements `BufferRecycler` interface.

- We could also extend the general `BufferRecycler`  interface to provide method of `recycle(Collection<MemorySegment>)`. 

To do so we make use of the existing semantics and architecture. But the only concern is we would affect the previous implementations to bring more changes.

azagrebin:
Thanks for more review @zhijiangW !

I removed `PartitionBufferPoolFactory` commits and changed them to function suppliers. Normal partition/gate factories are now responsible for buffer pool creation logic. We can reintroduce `PartitionBufferPoolFactory` in future if normal factories get too many responsibilities.

True,`ShuffleService` will be a factory for partitions/gates. I just do not want it to take too many responsibilities. I think delegating further creation process to concrete factories already makes sense atm. Ideally, any class should not take more than one responsibility among couple of abstraction layers but it also depends practically on how it is already big or not, of course. Also class constructors should ideally not have any complex logic to see real class dependencies and simplify dependency injection, e.g. for testing. We can just have `BasePartitionFactory` and `FSResultPartitionFactory` for `FSResultPartition` then or see how it goes.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/ResultPartition.java
```diff
@@ -154,13 +161,16 @@
...
-
+	@Override
+	public void setup() throws IOException {
 		checkState(this.bufferPool == null, "Bug in result partition setup logic: Already registered buffer pool.");
 
+		BufferPool bufferPool = bufferPoolFactory.apply(this);
+		checkArgument(bufferPool.getNumberOfRequiredMemorySegments() >= getNumberOfSubpartitions(),
+			"Bug in result partition setup logic: Buffer pool has not enough guaranteed buffers for this result partition.");
+
 		this.bufferPool = checkNotNull(bufferPool);
```
zentol: null checks is ineffective

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/consumer/LocalInputChannelTest.java
```diff
@@ -114,11 +114,12 @@ public void testConcurrentConsumeMultiplePartitions() throws Exception {
 				.setNumTargetKeyGroups(parallelism)
 				.setResultPartitionManager(partitionManager)
 				.setSendScheduleOrUpdateConsumersMessage(true)
+				.setBufferPoolFactory(p ->
+					networkBuffers.createBufferPool(producerBufferPoolSize, producerBufferPoolSize))
 				.build();
 
 			// Create a buffer pool for this partition
```
zentol: comment seems a bit out-dated, I'd probably just remove it

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NetworkEnvironment.java
```diff
@@ -88,35 +86,62 @@
...
+
+		SingleInputGateFactory singleInputGateFactory = new SingleInputGateFactory(
+			config, connectionManager, resultPartitionManager, taskEventPublisher, networkBufferPool);
+
+		return new NetworkEnvironment(
+			config,
+			networkBufferPool,
+			connectionManager,
+			resultPartitionManager,
+			checkNotNull(taskEventPublisher),
```
zhijiangW: `checkNotNull` should be done in first reference with creation `SingleInputGateFactory` above.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NetworkEnvironment.java
```diff
@@ -88,35 +86,62 @@
...
 
 		registerNetworkMetrics(metricGroup, networkBufferPool);
 
-		this.ioManager = checkNotNull(ioManager);
-
-		isShutdown = false;
+		ResultPartitionManager resultPartitionManager = new ResultPartitionManager();
+		ResultPartitionFactory resultPartitionFactory = new ResultPartitionFactory(
+			resultPartitionManager, checkNotNull(ioManager), networkBufferPool,
+			config.networkBuffersPerChannel(), config.floatingNetworkBuffersPerGate());
```
zhijiangW: I am not sure whether it is worth passing `config` here instead of two separate parameters like `SingleInputGateFactory`, which could avoid changing the structor of `ResultPartitionFactory` if there are more infos needed from `config` future.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/buffer/NetworkBufferPool.java
```diff
@@ -69,14 +71,19 @@
 
 	private int numTotalRequiredBuffers;
 
+	private final int numberOfRequestedBuffers;
```
zhijiangW: numberOfRequestedBuffers -> numberOfRequiredBuffers?

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/ResultPartitionFactory.java
```diff
@@ -0,0 +1,199 @@
...
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.annotation.Nonnull;
+
+import java.io.IOException;
+import java.util.Optional;
+
+/** Factory for {@link ResultPartition} to use in {@link org.apache.flink.runtime.io.network.NetworkEnvironment}. */
```
zhijiangW:
```
/**
*
**/
```

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/ResultPartition.java
```diff
@@ -125,52 +124,33 @@
...
+		boolean sendScheduleOrUpdateConsumersMessage,
+		@Nonnull FunctionWithException<BufferPoolOwner, BufferPool, IOException> bufferPoolFactory) {
 
 		this.owningTaskName = checkNotNull(owningTaskName);
 		this.taskActions = checkNotNull(taskActions);
 		this.jobId = checkNotNull(jobId);
 		this.partitionId = checkNotNull(partitionId);
 		this.partitionType = checkNotNull(partitionType);
-		this.subpartitions = new ResultSubpartition[numberOfSubpartitions];
+		this.subpartitions = subpartitions;
```
zhijiangW: checkNotNull(subpartitions)?

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/consumer/RemoteInputChannel.java
```diff
@@ -121,24 +117,29 @@ public RemoteInputChannel(
...
+		this.memorySegmentProvider = memorySegmentProvider;
 	}
 
 	/**
 	 * Assigns exclusive buffers to this input channel, and this method should be called only once
 	 * after this input channel is created.
 	 */
-	void assignExclusiveSegments(List<MemorySegment> segments) {
+	void assignExclusiveSegments() throws IOException {
+		Collection<MemorySegment> segments = memorySegmentProvider.requestMemorySegments();
```
zhijiangW: request should be after the check of `this.initialCredit == 0` below.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/consumer/SingleInputGate.java
```diff
@@ -219,6 +211,17 @@ public SingleInputGate(
 		this.isCreditBased = isCreditBased;
 	}
 
+	@Override
+	public void setup() throws IOException {
```
zhijiangW: In the setup, we should also `checkState(this.bufferPool == null)` firstly, then it could void bring potential trouble of releasing the related buffers if setup is called twice.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/consumer/SingleInputGate.java
```diff
@@ -297,37 +300,18 @@ public void setBufferPool(BufferPool bufferPool) {
 
 	/**
 	 * Assign the exclusive buffers to all remote input channels directly for credit-based mode.
-	 *
-	 * @param networkBufferPool The global pool to request and recycle exclusive buffers
-	 * @param networkBuffersPerChannel The number of exclusive buffers for each channel
 	 */
-	public void assignExclusiveSegments(NetworkBufferPool networkBufferPool, int networkBuffersPerChannel) throws IOException {
+	public void assignExclusiveSegments() throws IOException {
```
zhijiangW: should add @VisibleForTesting in this method.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/consumer/SingleInputGateFactory.java
```diff
@@ -0,0 +1,209 @@
...
+		@Nonnull NetworkBufferPool networkBufferPool) {
+
+		this.isCreditBased = networkConfig.isCreditBased();
+		this.partitionRequestInitialBackoff = networkConfig.partitionRequestInitialBackoff();
+		this.partitionRequestMaxBackoff = networkConfig.partitionRequestMaxBackoff();
+		this.connectionManager = connectionManager;
+		this.partitionManager = partitionManager;
+		this.taskEventPublisher = taskEventPublisher;
+		this.networkBufferPool = networkBufferPool;
+		this.networkBuffersPerChannel = networkConfig.networkBuffersPerChannel();
```
zhijiangW: It might be better to put `this.networkBuffersPerChannel` together with above `this.partitionRequestMaxBackoff`, because these infos are all form `networkConfig`.

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/NetworkEnvironmentTest.java
```diff
@@ -63,35 +64,31 @@
...
 	/**
-	 * Verifies that {@link NetworkEnvironment#registerTask(Task)} sets up (un)bounded buffer pool
+	 * Verifies that {@link Task#setupPartionsAndGates(ResultPartitionWriter[], InputGate[])}} sets up (un)bounded buffer pool
 	 * instances for various types of input and output channels.
 	 */
 	@Test
 	public void testRegisterTaskUsesBoundedBuffers() throws Exception {
 		final NetworkEnvironment network = new NetworkEnvironmentBuilder()
 			.setIsCreditBased(enableCreditBasedFlowControl)
+			.setNetworkBuffersPerChannel(2)
```
zhijiangW: It is not necessary because the default value is 2 in `NetworkEnvironmentBuilder`

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/buffer/LocalBufferPoolDestroyTest.java
```diff
@@ -48,7 +48,7 @@ public void testDestroyWhileBlockingRequest() throws Exception {
 		LocalBufferPool localBufferPool = null;
 
 		try {
-			networkBufferPool = new NetworkBufferPool(1, 4096);
+			networkBufferPool = new NetworkBufferPool(1, 4096, 1);
```
zhijiangW: Maybe it is better to keep the previous constructor to avoid change many tests.
```
NetworkBufferPool(int numberOfSegmentsToAllocate, int segmentSize) {
    this (numberOfSegmentsToAllocate, segmentSize, 2)
}
```

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/InputChannelTestUtils.java
```diff
@@ -133,31 +103,64 @@ public static LocalInputChannel createLocalInputChannel(
...
+		PartitionRequestClient client,
+		MemorySegmentProvider memorySegmentProvider) {
+
+		return InputChannelBuilder.newBuilder()
+			.setConnectionManager(mockConnectionManagerWithPartitionRequestClient(client))
+			.setMemorySegmentProvider(memorySegmentProvider)
+			.buildRemote(inputGate);
+	}
+
+	public static ConnectionManager mockConnectionManagerWithPartitionRequestClient(PartitionRequestClient client) {
```
zhijiangW: Maybe it is better to construct the real `PartitionRequestClient` and return in `createPartitionRequestClient`, not pass it here because most of the usages are stilling mocking the `ConnectionID` and `PartitionRequestClientFactory` while constructing `PartitionRequestClient`.

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/PartitionTestUtils.java
```diff
@@ -40,35 +38,38 @@ public static ResultPartition createPartition(ResultPartitionType type) {
...
 	public static ResultPartition createPartition(
 			ResultPartitionConsumableNotifier notifier,
 			ResultPartitionType type,
 			boolean sendScheduleOrUpdateConsumersMessage) {
 
-		return createPartition(notifier, type, 1, sendScheduleOrUpdateConsumersMessage);
+		return createPartition(null, notifier, type, 1, sendScheduleOrUpdateConsumersMessage);
 	}
 
 	public static ResultPartition createPartition(
```
zhijiangW: Since we already have the `ResultPartitionBuilder`, this `PartitionTestUtils` seems not very necessary because it seems trouble to maintain the different units of parameters.

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/ResultPartitionTest.java
```diff
@@ -215,10 +215,9 @@ private void testReleaseMemory(final ResultPartitionType resultPartitionType) th
 		final int numAllBuffers = 10;
 		final NetworkEnvironment network = new NetworkEnvironmentBuilder()
 			.setNumNetworkBuffers(numAllBuffers).build();
-		final ResultPartitionConsumableNotifier notifier = new NoOpResultPartitionConsumableNotifier();
-		final ResultPartition resultPartition = createPartition(notifier, resultPartitionType, false);
+		final ResultPartition resultPartition = createPartition(network, resultPartitionType, 1);
```
zhijiangW: Actually the default value 1 is not needed, just because we could not give all the possible units of parameters in `PartitionTestUtils`. So it might be better to use `ResultPartitionBuilder` directly to delete intermediate `PartitionTestUtils` as I mentioned above.

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/consumer/SingleInputGateBuilder.java
```diff
@@ -0,0 +1,102 @@
...
+	}
+
+	public SingleInputGateBuilder setupBufferPoolFactory(NetworkEnvironment environment) {
+		NetworkEnvironmentConfiguration config = environment.getConfiguration();
+		this.bufferPoolFactory = SingleInputGateFactory.createBufferPoolFactory(
+			environment.getNetworkBufferPool(),
+			config.isCreditBased(),
+			config.networkBuffersPerChannel(),
+			config.floatingNetworkBuffersPerGate(),
+			numberOfChannels, partitionType);
```
zhijiangW: separate line for `numberOfChannels` and `partitionType`?

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/ResultPartitionBuilder.java
```diff
@@ -0,0 +1,165 @@
...
+import org.apache.flink.util.function.FunctionWithException;
+
+import java.io.IOException;
+
+/**
+ * Utility class to encapsulate the logic of building a {@link ResultPartition} instance.
+ */
+public class ResultPartitionBuilder {
+
+	private JobID jobId = new JobID();
```
zhijiangW: should be final for these vars.

zhijiangW:
Thanks for the updates @azagrebin ! It seems much better now.

I have other two concerns:

- It is really nice now to introduce some builders for constructing `ResultPartition`, `SingleInputGate` and `InputChannel`. But there still exits various utils in tests to support various of parameter units for creating these items. It might bring unnecessary changes every time when make changes for partition/gate/channel. Even these scattered utils make hard to maintain and still could not provide everything. So I want to remove all these unnecessary utils in tests to unified use the proposed builder, maybe not in the scope of this PR.

- I was ever a bit confused of the proposed `MemorySegmentProvider` interface, because I think the new introduced methods should be extended in existing `BufferProvider` and `BufferRecycler` to support the batch of buffer requesting/recycle. But now I think it seems make sense to do as now, because it is very simple to let `NetworkBufferPool` implement the new interface and `NetworkBufferPool` actually has the corresponding functions internally before.

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/InputGateConcurrentTest.java
```diff
@@ -88,7 +90,9 @@ public void testConsumptionWithRemoteChannels() throws Exception {
 		final ConnectionManager connManager = createDummyConnectionManager();
 		final Source[] sources = new Source[numberOfChannels];
 
-		final SingleInputGate gate = createSingleInputGate(numberOfChannels);
```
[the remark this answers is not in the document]
azagrebin: True, actually I do not mind short-hands. Though, I would remove one of short-hands with 3 parameters. I think builder is more readable in case of many parameters.

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/consumer/LocalInputChannelTest.java
azagrebin: well, this is still what actually happens atm, why is it out-dated?

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/consumer/SingleInputGateBuilder.java
```diff
@@ -0,0 +1,82 @@
...
+import org.apache.flink.runtime.taskmanager.TaskActions;
+
+/**
+ * Utility class to encapsulate the logic of building a {@link SingleInputGate} instance.
+ */
+public class SingleInputGateBuilder {
+
+	private JobID jobId = new JobID();
+
+	private IntermediateDataSetID intermediateDataSetID = new IntermediateDataSetID();
```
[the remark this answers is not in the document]
azagrebin: though builder is not supposed to be immutable in general, we can keep it final for now and change if needed

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/consumer/InputChannelBuilder.java
```diff
@@ -0,0 +1,145 @@
...
+		LocalInputChannel channel = new LocalInputChannel(
+			inputGate,
+			channelIndex,
+			partitionId,
+			partitionManager,
+			taskEventPublisher,
+			initialBackoff,
+			maxBackoff,
+			metrics);
+		inputGate.setInputChannel(partitionId.getPartitionId(), channel);
```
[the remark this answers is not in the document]
azagrebin: true, I can add it explicitly to the method name but it is how it is used in almost all tests, so I would leave it.

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/netty/PartitionRequestClientTest.java
```diff
@@ -106,15 +106,15 @@ public void testDoublePartitionRequest() throws Exception {
 		final PartitionRequestClient client = new PartitionRequestClient(
 			channel, handler, mock(ConnectionID.class), mock(PartitionRequestClientFactory.class));
 
-		final NetworkBufferPool networkBufferPool = new NetworkBufferPool(10, 32);
+		final int numExclusiveBuffers = 2;
```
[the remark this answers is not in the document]
azagrebin: I left it where it was used somewhere else, like in assertions, for readability

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/buffer/NetworkBufferPool.java
azagrebin: I would name it `numberOfSegmentsToRequest`

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/buffer/LocalBufferPoolDestroyTest.java
azagrebin: I did not want to explode number of constructors for tests, especially for one, not very important parameters. I think it does not make too long to add it in tests.

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/InputChannelTestUtils.java
azagrebin: we could consider it, but this sounds like another refactoring :) for a later follow up

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NetworkEnvironment.java
azagrebin: In general, I would prefer to keep classes/functions to have as least dependencies as possible if full `config` is not needed, it is easier to see the real dependencies and use in other places, like avoid mocking full config in tests.

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/PartitionTestUtils.java
azagrebin: Well, certain short-cuts are still shorter if used in many places.. Not sure, I'll reduce it a bit more but atm I do not see it strictly being necessary.

azagrebin:
Thanks for the reviews @zentol @zhijiangW , I addressed the comments

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NetworkEnvironment.java
```diff
@@ -115,7 +115,11 @@ public static NetworkEnvironment create(
 			TaskEventPublisher taskEventPublisher,
 			MetricGroup metricGroup,
 			IOManager ioManager) {
-		NettyConfig nettyConfig = checkNotNull(config).nettyConfig();
+		checkNotNull(ioManager);
```
zentol: the check at L 136 is now reedundant

azagrebin: true, done

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/InputGateFairnessTest.java
```diff
@@ -327,7 +327,8 @@ public FairnessVerifyingInputGate(
 				boolean isCreditBased) {
 
 			super(owningTaskName, jobId, consumedResultId, ResultPartitionType.PIPELINED,
-				consumedSubpartitionIndex, numberOfInputChannels, taskActions, new SimpleCounter(), isCreditBased);
+				consumedSubpartitionIndex, numberOfInputChannels, taskActions, new SimpleCounter(),
+				isCreditBased, () -> null);
```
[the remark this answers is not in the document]
azagrebin: I would prefer `throw new UnsupportedOperationException()`. There is no need for `Optional` in production code.

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/ResultPartitionBuilder.java
```diff
@@ -48,6 +55,14 @@
 
 	private boolean sendScheduleOrUpdateConsumersMessage = false;
 
+	private NetworkBufferPool networkBufferPool = new NetworkBufferPool(1, 1,1);
+
+	private int networkBuffersPerChannel = 1;
+
+	private int floatingNetworkBuffersPerGate = 1;
+
+	private FunctionWithException<BufferPoolOwner, BufferPool, IOException> bufferPoolFactory;
```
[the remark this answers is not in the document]
azagrebin: Is it related to the previous comment about `UnsupportedOperationException`?

azagrebin:
Thanks for the review @pnowojski, I've addressed comments and left a comment about `Optional`.

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/ResultPartitionBuilder.java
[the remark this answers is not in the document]
pnowojski: I think not. This builder is later performing `!= null` check on this field

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/InputGateFairnessTest.java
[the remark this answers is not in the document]
pnowojski: In that case `throw new UnsupportedOperationException` is of course better :)

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/ResultPartitionBuilder.java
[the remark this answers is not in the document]
azagrebin: ok, I see, done

PR #8417, opened 2019-05-11
[hotfix][javadocs] Correct the example of select for Interface Table.

wenhuitang:
The example of select in Table.java is "tab.select("key, value.avg + ' The average' as average")" which can causes ValidationException "Cannot resolve field [key], input field list:[TMP_0]". So correct the example.

### DIFF
--- a/flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/Table.java
+++ b/flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/Table.java
@@ -97,9 +97,22 @@ public interface Table {
 	 *
 	 * <pre>
 	 * {@code
-	 *   tab.select("key, value.avg + ' The average' as average")
+	 *   tab.select("key")
 	 * }
 	 * </pre>
+	 *
+	 * <pre>
+	 * {@code
+	 *   tab.select("value.avg + ' The average' as average")
+	 * }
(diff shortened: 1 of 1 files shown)


dawidwys:
Hi @wenhuitang. To be honest I don't think this change improves the javadoc. The result of the original example depends on how the orignal `tab` Table was defined. So it would work if e.g. the schema of tab was `(key: String, ` The average`: Integer)`. Your example has the same problem/shortcoming that it does not have the context. So your examples also can fail if a wrong Table is used.

To sum up I'm -1 for this change.

wenhuitang:
> Hi @wenhuitang. To be honest I don't think this change improves the javadoc. The result of the original example depends on how the orignal `tab` Table was defined. So it would work if e.g. the schema of tab was `(key: String, ` The average`: Integer)`. Your example has the same problem/shortcoming that it does not have the context. So your examples also can fail if a wrong Table is used.
> 
> To sum up I'm -1 for this change.

Hi @dawidwys  ， thanks for your review. I thought this example was wrong cause this example select a field 'key with a aggregation without using groupBy. It seems that the example " tab.select('key, 'value.avg + " The average" as 'average)" still causes ValidationException  even the schema of tab was (key: String, value: Integer, The average: Integer). 
and the example is kind of misleading, maybe we can found a better example.

PR #8418, opened 2019-05-11
[FLINK-12491][docs][configuration] Fix incorrect javadoc for path sep…

kezhuw:
…arators of CoreOptions.TMP_DIRS

Fix incorrect javadoc for path separators part of  `CoreOptions.TMP_DIRS` and `ConfigConstants.TASK_MANAGER_TMP_DIR_KEY`

## Brief change log
* Fix incorrect javadoc for path separators part of  `CoreOptions.TMP_DIRS` and `ConfigConstants.TASK_MANAGER_TMP_DIR_KEY`.
* Refactor `ConfigurationUtils.splitPaths` so we can test both unix and windows path separators without demand on ci test matrix.

## Verifying this change
* Added test for documented path separators.

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-core/src/main/java/org/apache/flink/configuration/CoreOptions.java
+++ b/flink-core/src/main/java/org/apache/flink/configuration/CoreOptions.java
@@ -320,8 +320,8 @@ public class CoreOptions {
     // ------------------------------------------------------------------------
 
     /**
-     * The config parameter defining the directories for temporary files, separated by ",", "|", or
-     * the system's {@link java.io.File#pathSeparator}.
+     * The config parameter defining the directories for temporary files, separated by ",", or the
+     * system's {@link java.io.File#pathSeparator}.
      */
     @Documentation.OverrideDefault(
             "'LOCAL_DIRS' on Yarn. '_FLINK_TMP_DIR' on Mesos. System.getProperty(\"java.io.tmpdir\") in standalone.")
@@ -331,7 +331,7 @@ public class CoreOptions {
                     .defaultValue(System.getProperty("java.io.tmpdir"))
                     .withDeprecatedKeys("taskmanager.tmp.dirs")
                     .withDescription(
-                            "Directories for temporary files, separated by\",\", \"|\", or the system's java.io.File.pathSeparator.");
+                            "Directories for temporary files, separated by\",\", or the system's java.io.File.pathSeparator.");
 
     // ------------------------------------------------------------------------
     //  program
--- a/docs/_includes/generated/common_miscellaneous_section.html
+++ b/docs/_includes/generated/common_miscellaneous_section.html
@@ -24,7 +24,7 @@
             <td><h5>io.tmp.dirs</h5></td>
             <td style="word-wrap: break-word;">'LOCAL_DIRS' on Yarn. '_FLINK_TMP_DIR' on Mesos. System.getProperty("java.io.tmpdir") in standalone.</td>
             <td>String</td>
-            <td>Directories for temporary files, separated by",", "|", or the system's java.io.File.pathSeparator.</td>
+            <td>Directories for temporary files, separated by",", or the system's java.io.File.pathSeparator.</td>
         </tr>
     </tbody>
 </table>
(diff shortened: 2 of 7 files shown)


dawidwys:
Hi @kezhuw,
Thank you for your contribution. I think your correct that there is a problem in `org.apache.flink.configuration.ConfigurationUtils#splitPaths`, but rather than adjusting the javadocs (and docs) to the broken code, how about we fix the method?

dawidwys:
@flinkbot approve description

kezhuw:
@dawidwys After dug deeper, I found the history of used path separators:

* Initially, the separator was solely ":", it was changed to ˋFile.pathSeparatorˋ in commit  https://github.com/apache/flink/commit/80d6c586c495f742da6e06874e1f3b2f678d7231#diff-0637c4d13b3c0f4301a62a47bedae39e ( Improved compatibility with Windows )
* Then the commit https://github.com/apache/flink/commit/68100e5c9414da6826ab04fc1d40f2ea6e8744c2#diff-c13f0f23fb00495ef0df245fec2bbbb2 (Allow both comma and double-colon as path separator) change it to ˋ",|"+File.pathSeparatorˋ. The documentation https://github.com/apache/flink/commit/68100e5c9414da6826ab04fc1d40f2ea6e8744c2#diff-32da23d69ac63fdaae0193ad7de283e3 says that:
   > (colon ':' on unix) or a comma character ','. An example would be /data1/tmp:/data/tmp:/data3/tmp or /data1/tmp,/data2/tmp,/data3/tmp
* After that commit, documentation added in https://github.com/apache/flink/commit/a7c407ace4f6cbfbde3e247071cee5a755ae66db#diff-125ce56ea60a577549026859d5d9a4e0 and https://github.com/apache/flink/commit/76abcaa55d0d6ab704b7ab8164718e8e2dcae2c4#diff-c4482d965ffc8a83e8a099e0c7f07a22 say that ˋ|ˋ is supported also which is not true.

I look at ˋflink-conf.yamlˋ in master (518616e9f3527b2564a640b38c80db3eb54bef37 https://github.com/apache/flink/blob/master/flink-dist/src/main/resources/flink-conf.yaml#L153), it says that:
> Add a delimited list for multiple directories, using the system directory delimiter (colon ':' on unix) or a comma, e.g.:   /data1/tmp:/data2/tmp:/data3/tmp

So I think it is a mistake in documentation not code.

Besides the history, I think it is distractive to support more separators.

What do you think ?

dawidwys:
What do you think @aljoscha? Which separators should we support?

kezhuw:
@dawidwys @aljoscha I think current the two path separators serve their "purposes" if there is any:
* ˋFile.pathSeparatorˋ is same as os's path separator, it fits target os's convention.
* ˋ,ˋ provides os-agnostic path separator, so we can configure separated file paths cross os.

A new os-agnostic path separator introduces nothing but distraction in my opinion.

aljoscha:
Isn't the problem that someone interpreted the `|` as an actual separator and not as the regex syntax that it is?

kezhuw:
@aljoscha I think this is the case when that javadoc was introduced in commit https://github.com/apache/flink/commit/a7c407ace4f6cbfbde3e247071cee5a755ae66db#diff-125ce56ea60a577549026859d5d9a4e0 .

kezhuw:
@dawidwys @aljoscha Hi, is there any consensus on this issue ?

flink-core/src/main/java/org/apache/flink/configuration/CoreOptions.java
```diff
@@ -210,7 +210,7 @@
 
 	/**
 	 * The config parameter defining the directories for temporary files, separated by
-	 * ",", "|", or the system's {@link java.io.File#pathSeparator}.
+	 * "," or the system's {@link java.io.File#pathSeparator}.
```
dawidwys: Update also the description in `TMP_DIRS.withDescription` few lines below.

kezhuw: Have fixed.

kezhuw:
@flinkbot attention @dawidwys

kezhuw:
@dawidwys Could you take time to review this pull request ? It has been more that one month past since your last review comment.

kezhuw:
@flinkbot run azure

kezhuw:
@dawidwys @aljoscha Could you have time to review this pr ? We have forgotten it for a long time.

kezhuw:
@flinkbot run azure

kezhuw:
@flinkbot run azure

PR #8420, opened 2019-05-12
[FLINK-12408][python] Allow to define the data types in Python

dianfu:

*This pull request adds the basic types in Python to allow to define data types in Python*

## Brief change log

  - *Adds BooleanType, TinyIntType, SmallIntType,  intType, BigIntType, FloatType, DoubleType, DecimalType, DateType, TimeType, TimestampType, RowType, MapType, MultisetType, ArrayType and UserDefinedType in Python*
  - *Adds type infer utilities in Python to infer the data type from python data*
  - *Adds schema parser utilities in Java to parse the JSON defined data type*

The following things should be noted:
 - Some data types such as `TimestampWithTimeXXX`, `DistinctType`, etc are still not supported as there is no corresponding TypeInformation for these types yet.
 - *`AnyType`, `TypeInformationAnyType`are still not supported as we need to find an proper way to support them*

## Verifying this change

  - *Added tests in test_types.py*

  - The serializers: (no)

## Documentation


sunjincheng121:
@flinkbot attention @twalthr

flink-python/pyflink/serializers.py
```diff
@@ -0,0 +1,192 @@
...
+    END_OF_DATA_SECTION = -1
+
+
+class Serializer(object):
+
+    def dump_stream(self, iterator, stream):
+        """
+        Serializes an iterator of objects to the output stream.
+        """
+        raise NotImplementedError
```
WeiZhong94: How about using `abc.ABCMeta` to mark the abstract method instead of `NotImplementedError `? I think using `abc.ABCMeta` is more semantically clear.

flink-python/pyflink/table/types.py
```diff
@@ -15,172 +15,1803 @@
...
+            if data_type.nullable:
+                return True
+            else:
+                raise ValueError(new_msg("This field is not nullable, but got None"))
+        else:
+            return False
+
+    _type = type(data_type)
+
+    def assert_acceptable_types(obj):
```
WeiZhong94: Maybe we can check the data type before verifier is created? The input `obj` is unnecessary for the core logic `assert _type in _acceptable_types`. We needn't run this function every time.

twalthr:
Thanks @dianfu. I would like to take a look soon. We should make the types consistent with the overall type strategy. I already saw differences in the documentation. But will do a full pass shortly.

flink-python/pyflink/table/types.py
```diff
@@ -15,172 +15,1803 @@
...
+
+    def __new__(self, *args, **kwargs):
+        if args and kwargs:
+            raise ValueError("Can not use both args "
+                             "and kwargs to create Row")
+        if kwargs:
+            # create row objects
+            names = sorted(kwargs.keys())
+            row = tuple.__new__(self, [kwargs[n] for n in names])
+            row.__fields__ = names
```
WeiZhong94: The identifiers that double underscore before and after a name are special method names used by Python. I think we shouldn't use the special-method-looking name as our method name or variable name.

dianfu:
@twalthr Thanks in advance. I noticed that you are working on the rework of SQL & Table API type system. Definitely agree that we should make sure the Python types consistent with the overall type strategy. You comments would be very helpful to achieve it. Looking forward to your feedback!

flink-table/flink-table-planner/src/main/scala/org/apache/flink/table/api/Types.scala
```diff
@@ -196,4 +200,145 @@ object Types {
   def MULTISET[E](elementType: TypeInformation[E]): TypeInformation[util.Map[E, lang.Integer]] = {
     new MultisetTypeInfo(elementType)
   }
+
+  def fromJson(json: String): TypeInformation[_] = {
```
sunjincheng121: Should we add this method? It's better to only use the py4j's ability for creating the Java Type. then we can get the benefits as follows:
- Reduce the dependency of  `table-planner module`, i.e. remove the dependency: org.json4s
- We do not need to maintain a new method `fromJson` when adding a new type in Java.
- Do not need to add `serializers` in python.

Can we try these efforts? And please let me know what do you think?

flink-python/pyflink/table/types.py
```diff
@@ -15,172 +15,1803 @@
...
+
+    return convert_row
+
+
+def _to_java_type(data_type):
+    """
+    Converts Python type to Java type.
+    """
+    gateway = get_gateway()
+    return gateway.jvm.org.apache.flink.table.api.Types.fromJson(data_type.json())
```
sunjincheng121: Serializing all types to JSON format can reduce the communication of many JVMs, which is a good idea. But I think the other issue is we should add the `fromJson()`, i.e. add a new dependency in java module, and introducing maintenance costs on the Java side, which I left the text in Java section.  :)

flink-python/pyflink/table/types.py
dianfu: Good catch. +1.

flink-python/pyflink/table/types.py
dianfu: Make sense. Changed to _fields, using the same name as namedtuple.

flink-python/pyflink/table/types.py
dianfu: Good idea. I have removed method fromJson and updated method `_to_java_type` to create Java TypeInformation using Py4J.

flink-table/flink-table-planner/src/main/scala/org/apache/flink/table/api/Types.scala
dianfu: Makes sense. Have updated the PR and addressed these issues.

dianfu:
@sunjincheng121 @WeiZhong94 Thanks a lot for your review. I have updated the PR according to your comments.

dianfu:
@twalthr Any feedback? Very appreciated if you could take a look. Thanks in advance.

sunjincheng121:

flink-python/pyflink/table/types.py
```diff
@@ -16,171 +16,1811 @@
...
+        elif isinstance(key, slice):
+            return RowType(self.fields[key])
+        else:
+            raise TypeError('RowType keys should be strings, integers or slices')
+
+    def __repr__(self):
+        return "RowType(%s)" % ",".join(str(field) for field in self)
+
+    @classmethod
+    def from_json(cls, json):
```
sunjincheng121: Add the convenient method is a pretty good idea, but for now, we do not add the `to_json`for pair, and this JIRA is the main goal is to align with JAVA API, So It's better to add it when suitable time. What do you think?

flink-python/pyflink/table/types.py
```diff
@@ -16,171 +16,1811 @@
...
+_static_length_type_mappings = dict((t.type_name(), t) for t in _static_length_types)
+
+_var_length_types = [BinaryType, VarBinaryType, CharType, VarCharType]
+_var_length_type_mappings = dict((t.type_name(), t) for t in _var_length_types)
+
+_complex_type_mappings = dict(
+    (t.type_name(), t) for t in [ArrayType, MapType, MultisetType, RowType])
+
+
+def _parse_datatype_json_value(json_value):
```
sunjincheng121: Same as the `from_json` comment.

flink-python/pyflink/table/tests/test_types.py
```diff
@@ -0,0 +1,747 @@
...
+            ValueError,
+            "test_name",
+            lambda: _create_type_verifier(
+                DataTypes.VARCHAR(nullable=False), name="test_name")(None))
+
+        schema = DataTypes.ROW(
+            [DataTypes.FIELD('a', DataTypes.ROW([DataTypes.FIELD('b', DataTypes.INT())]))])
+        self.assertRaisesRegexp(
+            TypeError,
+            "field b in field a",
```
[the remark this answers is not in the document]
dianfu: Make sense to me. +1

flink-python/pyflink/table/types.py
```diff
@@ -16,171 +16,1811 @@
...
+    basestring = unicode = str
+
+__all__ = ['DataTypes', 'UserDefinedType', 'Row']
 
 
 class DataType(object):
     """
     Base class for data types.
+
+    :param nullable: boolean, whether the type can be null (None) or not.
```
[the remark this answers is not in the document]
dianfu: +1

flink-python/pyflink/table/types.py
dianfu: Agree. We can remove `from_json` for now and add it when necessary.

dianfu:
@sunjincheng121 Thanks a lot for your review. Have updated the PR and addressed your comments.

sunjincheng121:
Thanks for the quick update @dianfu!
one more suggestion is could you open a new JIRA for you mentioned in ` should be noted` section? so that when the JAVA type system is complete, we can absolutely align the JAVA.

Thanks,
Jincheng

dianfu:
@sunjincheng121 Make sense to me. Have created an JIRA https://issues.apache.org/jira/browse/FLINK-12609 to track this.

sunjincheng121:
LGTM. +1 to merged.
@flinkbot approve all

sunjincheng121:
Merging...

PR #8421, opened 2019-05-12
[FLINK-12492]Minor optimize the cep operator by avoiding unnecessary copy

Aitozi:

When use with rocksdbbackend, we can skip the copy of input events. But to make this more constraint, i add a method in the KeyedStateBackend interface to make this still work id there is added a new backend.

## Documentation

### DIFF
--- a/flink-libraries/flink-cep/src/main/java/org/apache/flink/cep/operator/CepOperator.java
+++ b/flink-libraries/flink-cep/src/main/java/org/apache/flink/cep/operator/CepOperator.java
@@ -292,12 +292,7 @@ public class CepOperator<IN, KEY, OUT>
 			elementsForTimestamp = new ArrayList<>();
 		}
 
-		if (getExecutionConfig().isObjectReuseEnabled()) {
-			// copy the StreamRecord so that it cannot be changed
-			elementsForTimestamp.add(inputSerializer.copy(event));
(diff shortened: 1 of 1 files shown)


Aitozi:
@flinkbot attention @dawidwys

dawidwys:
Hi @Aitozi,
We definitely shouldn't change the `KeyedStateBackend` interface for this change. So -1 for this change in the current state.

On the other hand I think the reason behind this change is correct. I just wonder if it wouldn't be ok to just remove the copy at all for all state backends. As far as I understand it, it was introduced to maintain the original order, but it sort of violates the contract of the `objectReuseEnabled`, which says that we should not perform additional copies. The contract of object reuse clearly states that some careless changes to objects might result in wrong results. To sum up I think I would be ok with removing this copy at all.

Aitozi:
Hi @dawidwys ,Thanks for your review. Completely agree you on that this copy is a little against the contract of the `objectReuseEnabled`. I'm +1 to remove this copy.  I will repush the code.

Aitozi:
ping @dawidwys . I have update the PR.  By the way, i notice this issue when tracking the problem mentioned in [FLINK-12351](https://issues.apache.org/jira/browse/FLINK-12351). I also think it should not be fixed..

PR #8426, opened 2019-05-13
[FLINK-12301] Fix ScalaCaseClassSerializer to support value types

aljoscha:
We now use Scala reflection because it correctly deals with Scala
language features.


aljoscha:
@flinkbot attention @igalshilman

aljoscha:
Merged

PR #8430, opened 2019-05-13
[FLINK-12068] [runtime] Backtrack failover regions if intermediate results are unavailable

zhuzhurk:

*In region failover, when a region fails due to unavailable input result partitions, it needs to backtrack the failover regions to recover the failed tasks as well as the unavailable result partitions. The detailed design is at https://docs.google.com/document/d/1YHOpMLdC-dtgjcM-EDn6v-oXgsEQKXSoMjqRcYVbJA8/edit.*

## Brief change log

  - *RestartPipelinedRegionStrategy(based on next generation interface) handles DataConsumptionException, proposing to restart regions producing needed but unavailable result partitions as well as all its consumer regions.*
  - *Add a ResultPartitionAvailabilityChecker interface and implement it for querying result partition availability*
  - *Calculate and cache region inputs and consumers in region building phase. This helps to speed up the failover handling significantly, at the time cost of slows down the region building significantly and space cost for 2 edge scale caches. See verification part below.*

## Verifying this change

  - *Added tests in flip1 RestartPipelinedRegionStrategyTest to verify the correctness*
  - *Performance tests are manually conducted. Currently for a job with 16 million edges it takes < 200ms to calculate the tasks to restart. The region building time in this case increased from 600ms to ~6s to build some helper caches.*

  - The serializers: (no)

## Documentation

### DIFF
--- a/docs/ops/config.zh.md
+++ b/docs/ops/config.zh.md
@@ -100,11 +100,15 @@ The default fraction for managed memory can be adjusted using the taskmanager.me
 
 {% include generated/security_configuration.html %}
 
+### Network Environment
+
+{% include generated/network_environment_configuration.html %}
+
 ### Network communication (via Netty)
 
 These parameters allow for advanced tuning. The default values are sufficient when running concurrent high-throughput jobs on a large cluster.
 
-{% include generated/netty_configuration.html %}
+{% include generated/network_netty_configuration.html %}
 
 ### Web Frontend
 


flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/failover/flip1/RestartPipelinedRegionStrategyTest.java
```diff
@@ -30,37 +38,245 @@
 public class RestartPipelinedRegionStrategyTest extends TestLogger {
```
zentol: What I'm missing here are a few _minimal_ tests for the key backtracking behaviors, ideally with similar topologies to reduce cognitive load.

For example,

```
v1 |-\
      -> v3	  
v2 |-/

v1/v2  fails -> v1&v3 is restarted, v2 optionally depending on partition state
v3 fails, v3 is restarted, v1/v2 optionally dependending on partition state
```

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/RestartPipelinedRegionStrategy.java
```diff
@@ -47,8 +50,14 @@
...
 
+	/** All failover regions. */
+	private final IdentityHashMap<FailoverRegion, Object> regions;
+
 	/** Maps execution vertex id to failover region. */
-	private final Map<ExecutionVertexID, FailoverRegion> regions;
+	private final Map<ExecutionVertexID, FailoverRegion> vertexToRegionMap;
+
+	/** The checker helps to query result partition availability. */
+	private RegionFailoverResultPartitionAvailabilityChecker resultPartitionAvailabilityChecker;
```
zentol: can be final

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/RestartPipelinedRegionStrategy.java
```diff
@@ -165,6 +181,33 @@ private void buildOneRegionForAllVertices() {
...
+		buildRegionInputsAndOutputs();
+	}
+
+	private void buildRegionInputsAndOutputs() {
+		for (FailoverRegion region : regions.keySet()) {
+			IdentityHashMap<FailoverRegion, Object> consumers = new IdentityHashMap<>();
+			Set<IntermediateResultPartitionID> inputs = new HashSet<>();
+			Set<ExecutionVertexID> consumerVertices = new HashSet<>();
+			Set<FailoverVertex> regionVertices = region.getAllExecutionVertices();
+			regionVertices.forEach(v -> {
```
zentol: related: https://github.com/apache/flink/pull/8430#discussion_r284194311

The idea here is fine. I'm wondering if we could merge this logic into the initial region building; we already iterate over all inputs there anyway. I think we can even detect region boundaries relatively easy (if an input edge is not pipelined you have a boundary, and you even know the (temporary) regions)

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/FailoverRegion.java
```diff
@@ -33,18 +30,21 @@
  */
 public class FailoverRegion {
 
+	/** All vertex IDs in this region. */
+	private final Set<ExecutionVertexID> executionVertexIDs;
+
 	/** All vertices in this region. */
-	private final Map<ExecutionVertexID, FailoverVertex> executionVertices;
+	private final Set<FailoverVertex> executionVertices;
```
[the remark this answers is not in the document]
zhuzhurk: This is to avoid the cost to convert `Collection<FailoverVertex>` to `Set<FailoverVertex>`. 
We need a `Set<ExecutionVertexID>` and `Set<FailoverVertex>`. But the map provides `Set<ExecutionVertexID>` and `Collection<FailoverVertex>`.

Though I think this may not make a big difference as it is vertex scale complexity.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/RestartPipelinedRegionStrategy.java
zhuzhurk: Yes.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/RestartPipelinedRegionStrategy.java
```diff
@@ -162,40 +201,90 @@ private void buildOneRegionForAllVertices() {
...
+		// so that the failover process will try to recover it
+		if (cause instanceof DataConsumptionException) {
+			resultPartitionAvailabilityChecker.markResultPartitionFailed(
+				((DataConsumptionException) cause).getPartitionId().getPartitionId());
+		}
 
-		return getRegionsToRestart(failedRegion).stream().flatMap(
+		// calculate the tasks to restart based on the result of regions to restart
+		Set<FailoverRegion> regionsToRestart = getRegionsToRestart(failedRegion);
+		Set<ExecutionVertexID> tasksToRestart = regionsToRestart.stream().flatMap(
```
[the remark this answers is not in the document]
zhuzhurk: Yes we can.
It may not make a big difference though, as it is vertex scale complexity(N), small compared to edge scale operations(N^2).

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/RestartPipelinedRegionStrategy.java
```diff
@@ -162,40 +201,90 @@ private void buildOneRegionForAllVertices() {
...
+							FailoverRegion producerRegion = vertexToRegionMap.get(inEdge.getSourceVertex().getExecutionVertexID());
+							if (!visitedRegions.containsKey(producerRegion)) {
+								regionsToVisit.add(producerRegion);
+							}
+						}
+					}
+				}
+
+				// all consumer regions of an involved region should be involved
+				for (FailoverRegion consumerRegion : regionConsumers.get(regionToRestart)) {
```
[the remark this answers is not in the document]
zhuzhurk: not applicable to the latest version since I pushed down the visited region check.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/RestartPipelinedRegionStrategy.java
```diff
@@ -162,40 +176,96 @@ private void buildOneRegionForAllVertices() {
...
+		while (!regionsToVisit.isEmpty()) {
+			FailoverRegion regionToRestart = regionsToVisit.poll();
+
+			// an involved region should be restarted
+			regionsToRestart.put(regionToRestart, null);
 
-		// TODO: implement backtracking logic
+			// if a needed input result partition is not available, its producer region is involved
+			for (FailoverVertex vertex : regionToRestart.getAllExecutionVertices()) {
+				for (FailoverEdge inEdge : vertex.getInputEdges()) {
```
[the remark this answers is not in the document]
zhuzhurk: It is similar to pre-calculating the region input result partitions and consumer regions.
Having boundary vertices, we still need to calculate the input result partitions and consumer regions, through iterating edges of the boundary vertices. 

The pre-calculation saves time in 2 aspects:
 * loop count(edges to iterate), by finding out the outer boundaries, especially for large regions
 * time cost each loop, by doing some lookup ahead, such as vertexToRegionMap.get(...)

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/RestartPipelinedRegionStrategy.java
zhuzhurk: I tried the merged logic and find the region building becomes even slower.
I think this is due to the merged logic saved some iteration overhead, but added some more work, including building some legacy relations and later deprecating them in region merging.

PR #8431, opened 2019-05-13
[FLINK-12480] Introduce mailbox to StreamTask main-loop

StefanRRichter:
This PR sits on top of #8409 and introduces the mailbox approach to the `StreamTask`. It is the next sub-task in our effort to change the threading model of `StreamTask`

## Brief change log

- New class `MailboxImpl` with interfaces.
- Rewritten `StreamTask#run` that spins around default action and mailbox.

## Verifying this change
Added unit tests for the mailbox, everything else is covered by existing tests.

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/OneInputStreamTask.java
+++ b/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/OneInputStreamTask.java
@@ -39,8 +39,6 @@ public class OneInputStreamTask<IN, OUT> extends StreamTask<OUT, OneInputStreamO
 
 	private StreamInputProcessor<IN> inputProcessor;
 
-	private volatile boolean running = true;
-
 	private final WatermarkGauge inputWatermarkGauge = new WatermarkGauge();
 
 	/**
@@ -98,13 +96,8 @@ public class OneInputStreamTask<IN, OUT> extends StreamTask<OUT, OneInputStreamO
 	}
 
 	@Override
-	protected void run() throws Exception {
-		// cache processor reference on the stack, to make the code more JIT friendly
-		final StreamInputProcessor<IN> inputProcessor = this.inputProcessor;
-
-		while (running && inputProcessor.processInput()) {
-			// all the work happens in the "processInput" method
-		}
+	protected boolean performDefaultAction() throws Exception {
+		return inputProcessor.processInput();
 	}
 
 	@Override
@@ -116,6 +109,5 @@ public class OneInputStreamTask<IN, OUT> extends StreamTask<OUT, OneInputStreamO
 
 	@Override
 	protected void cancelTask() {
-		running = false;
 	}
 }
--- a/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/OneInputStreamTask.java
+++ b/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/OneInputStreamTask.java
@@ -96,8 +96,10 @@ public class OneInputStreamTask<IN, OUT> extends StreamTask<OUT, OneInputStreamO
 	}
 
 	@Override
-	protected boolean performDefaultAction() throws Exception {
-		return inputProcessor.processInput();
+	protected void performDefaultAction(ActionContext context) throws Exception {
+		if (!inputProcessor.processInput()) {
+			context.allActionsCompleted();
+		}
 	}
 
 	@Override
--- a/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
+++ b/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
@@ -222,11 +222,23 @@ public abstract class StreamTask<OUT, OP extends StreamOperator<OUT>>
 
 	protected abstract void init() throws Exception;
 
-	protected abstract void run() throws Exception;
-
 	protected abstract void cleanup() throws Exception;
 
 	protected abstract void cancelTask() throws Exception;
+
+	/**
+	 * This method implements the default action of the task (e.g. processing one event from the input). Implementations
+	 * should (in general) be non-blocking.
+	 *
+	 * @return <code>true</code> if there is more work to perform as default action for this task and <code>false</code>
+	 * if the task is ready to finish.
+	 * @throws Exception on any problems in the action.
+	 */
+	protected abstract boolean performDefaultAction() throws Exception;
+
+	protected void run() throws Exception {
+		while (isRunning && performDefaultAction()) {}
+	}
 
 	/**
 	 * Emits the {@link org.apache.flink.streaming.api.watermark.Watermark#MAX_WATERMARK MAX_WATERMARK}
--- a/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
+++ b/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
@@ -55,8 +55,11 @@ import org.apache.flink.streaming.runtime.partitioner.ConfigurableStreamPartitio
 import org.apache.flink.streaming.runtime.partitioner.StreamPartitioner;
 import org.apache.flink.streaming.runtime.streamrecord.StreamRecord;
 import org.apache.flink.streaming.runtime.streamstatus.StreamStatusMaintainer;
+import org.apache.flink.streaming.runtime.tasks.mailbox.Mailbox;
+import org.apache.flink.streaming.runtime.tasks.mailbox.MailboxImpl;
 import org.apache.flink.util.ExceptionUtils;
 import org.apache.flink.util.Preconditions;
+import org.apache.flink.util.function.ThrowingRunnable;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -124,6 +127,12 @@ public abstract class StreamTask<OUT, OP extends StreamOperator<OUT>>
 	/** The logger used by the StreamTask and its subclasses. */
 	private static final Logger LOG = LoggerFactory.getLogger(StreamTask.class);
 
+	/** Special value, letter that terminates the mailbox loop. */
+	private static final Runnable POISON_LETTER = () -> {};
+
+	/** Special value, letter that "wakes up" a waiting mailbox loop. */
+	private static final Runnable DEFAULT_ACTION_AVAILABLE = () -> {};
+
 	// ------------------------------------------------------------------------
 
 	/**
@@ -182,6 +191,8 @@ public abstract class StreamTask<OUT, OP extends StreamOperator<OUT>>
 
 	private final SynchronousSavepointLatch syncSavepointLatch;
 
+	protected final Mailbox mailbox;
+
 	// ------------------------------------------------------------------------
 
 	/**
@@ -214,6 +225,7 @@ public abstract class StreamTask<OUT, OP extends StreamOperator<OUT>>
 		this.accumulatorMap = getEnvironment().getAccumulatorRegistry().getUserMap();
 		this.recordWriters = createRecordWriters(configuration, environment);
 		this.syncSavepointLatch = new SynchronousSavepointLatch();
+		this.mailbox = new MailboxImpl();
 	}
 
 	// ------------------------------------------------------------------------
@@ -230,14 +242,29 @@ public abstract class StreamTask<OUT, OP extends StreamOperator<OUT>>
 	 * This method implements the default action of the task (e.g. processing one event from the input). Implementations
 	 * should (in general) be non-blocking.
 	 *
-	 * @return <code>true</code> if there is more work to perform as default action for this task and <code>false</code>
-	 * if the task is ready to finish.
+	 * @param context context object for collaborative interaction between the action and the stream task.
 	 * @throws Exception on any problems in the action.
 	 */
-	protected abstract boolean performDefaultAction() throws Exception;
+	protected abstract void performDefaultAction(ActionContext context) throws Exception;
 
-	protected void run() throws Exception {
-		while (isRunning && performDefaultAction()) {}
+	/**
+	 * Runs the stream-tasks main processing loop.
+	 */
+	private void run() throws Exception {
+		final ActionContext actionContext = new ActionContext();
+		while (true) {
+			if (mailbox.hasMail()) {
+				Runnable letter;
+				while ((letter = mailbox.tryTakeMail()) != null) {
+					if (letter == POISON_LETTER) {
+						return;
+					}
+					letter.run();
+				}
+			}
+
+			performDefaultAction(actionContext);
+		}
 	}
 
 	/**
@@ -438,6 +465,7 @@ public abstract class StreamTask<OUT, OP extends StreamOperator<OUT>>
 
 	@Override
 	public final void cancel() throws Exception {
+		mailbox.clearAndPut(POISON_LETTER);
 		isRunning = false;
 		canceled = true;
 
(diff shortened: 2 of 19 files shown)


StefanRRichter:
CC @1u0

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
```diff
@@ -1280,4 +1320,40 @@ public void tryHandleCheckpointException(CheckpointMetaData checkpointMetaData,
...
+
+		/**
+		 * This method must be called to end the stream task when all actions for the tasks have been performed.
+		 */
+		public void allActionsCompleted() {
+			mailbox.shutDown(POISON_LETTER);
+		}
+
+		/**
+		 * Calling this method signals that the mailbox-tread should continue invoking the default action, e.g. because
```
Aitozi: mailbox-thread?

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/MailboxImpl.java
```diff
@@ -0,0 +1,233 @@
...
+		notFull.signal();
+		return letter;
+	}
+
+	private int increaseIndexWithWrapAround(int old) {
+		return (old + 1) & moduloMask;
+	}
+
+	private boolean isFull() {
+		return count >= ringBuffer.length;
```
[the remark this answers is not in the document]
StefanRRichter: No, it can't. Just a habbit to include this case still if there is no extra cost to it.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/Mailbox.java
```diff
@@ -0,0 +1,36 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.streaming.runtime.tasks.mailbox;
+
+import javax.annotation.Nonnull;
+
+/**
+ * A mailbox is basically a blocking queue for inter-thread message exchange in form of {@link Runnable} object between
```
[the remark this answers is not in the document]
StefanRRichter: the meaning is "between threads"

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/MailboxImpl.java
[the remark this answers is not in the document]
Aitozi: ok

StefanRRichter:
@Aitozi  thanks for taking a look. As I said in the description, this PR is based on #8409, which is the PR that only contains the changes for FLINK-12478. But this work builds upon the previous PR, so that is why you can also see the commit here.

StefanRRichter:
@Aitozi about the travis ci failure, I think it is unrelated to this PR because I can see that it happens for a batch job and we did not change batch. However, I it looks like this could be cause by FLINK-12434. What do you think @pnowojski ?

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/Mailbox.java
```diff
@@ -0,0 +1,36 @@
...
+
+package org.apache.flink.streaming.runtime.tasks.mailbox;
+
+import javax.annotation.Nonnull;
+
+/**
+ * A mailbox is basically a blocking queue for inter-thread message exchange in form of {@link Runnable} objects between
+ * multiple producer threads and a single consumer.
+ */
+public interface Mailbox extends MailboxReader, MailboxWriter {
```
1u0: I propose to abstract the message type.
Instead of using `Runnable`, make it a generic parameter (`Mailbox<Message>, MailboxReader<Message>, MailboxWriter<Message>`).

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
```diff
@@ -124,6 +127,12 @@
 	/** The logger used by the StreamTask and its subclasses. */
 	private static final Logger LOG = LoggerFactory.getLogger(StreamTask.class);
 
+	/** Special value, letter that terminates the mailbox loop. */
+	private static final Runnable POISON_LETTER = () -> {};
+
+	/** Special value, letter that "wakes up" a waiting mailbox loop. */
+	private static final Runnable DEFAULT_ACTION_AVAILABLE = () -> {};
```
1u0: Will this (and `ActionContext.actionsAvailable()`) be used later?

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/Mailbox.java
StefanRRichter: Why?

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
StefanRRichter: Yes, once we have non-blocking inputs.

StefanRRichter:
@1u0 thanks for the review. I think about renaming to sender/receiver, it might really better fit in the context.

About shutdown, yes it is required, yes it is optimized and I purposefully did not put it to the writer interface because this method should not be visible to all writers, only the object owner (here the stream task) should be allowed to invoke this method, that is also why it can be "out of style" with the client code interfaces.

Also I see no need for generifying the mailbox without a purpose. It is clearly scoped to runnables, and we avoid bloated code and casting. So what excatly would be the benefits?

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
```diff
@@ -1280,4 +1320,40 @@ public void tryHandleCheckpointException(CheckpointMetaData checkpointMetaData,
...
+		 * new input became available for processing.
+		 *
+		 * @throws InterruptedException on interruption.
+		 */
+		public void actionsAvailable() throws InterruptedException {
+			mailbox.putMail(DEFAULT_ACTION_AVAILABLE);
+		}
+
+		/**
+		 * Calling this method signals that the mailbox-tread should (temporarily) stop invoking the default action,
```
1u0: Left typo.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/Mailbox.java
1u0: It eases reasoning about `Mailbox` abstraction. So far we don't have any contracts/expectation for message type.
I hope, that using generic type here would not impact performance.

1u0: Generalizing it should not bloat `Mailbox` api and implementation at all, imo.

Here you are using `Runnable` as "message", it's rather implementation detail choice.
Technically, `StreamTask` could have been using plain data objects as messages (using the same `Mailbox` abstraction). Potentially, such approach simplifies reasoning about control flow.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/OneInputStreamTask.java
```diff
@@ -98,12 +96,9 @@ public void init() throws Exception {
...
 
 	@Override
-	protected void run() throws Exception {
-		// cache processor reference on the stack, to make the code more JIT friendly
-		final StreamInputProcessor<IN> inputProcessor = this.inputProcessor;
-
-		while (running && inputProcessor.processInput()) {
-			// all the work happens in the "processInput" method
+	protected void performDefaultAction(ActionContext context) throws Exception {
+		if (!inputProcessor.processInput()) {
```
kezhuw: This is a blocking operation, it will block until one `StreamRecord` processed. In the meantime, other letters in mailbox such as processing timer are delayed.

StefanRRichter: I am aware of that and that is why we have also planned FLINK-12479 to make the inputs non-blocking.
For the scope of this PR it should not matter and we don't have to wait for FLINK-12479 to complete until we work on the subtasks that use the mailbox for checkpoints etc.

kezhuw: Thanks, I did notice that.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/Mailbox.java
```diff
@@ -0,0 +1,36 @@
...
+
+package org.apache.flink.streaming.runtime.tasks.mailbox;
+
+import javax.annotation.Nonnull;
+
+/**
+ * A mailbox is basically a blocking queue for inter-thread message exchange in form of {@link Runnable} objects between
+ * multiple producer threads and a single consumer.
+ */
+public interface Mailbox extends MailboxReceiver, MailboxSender {
```
kezhuw: Is it necessary to mark all these interfaces `@Internal` ?

StefanRRichter: Necessary is relative because the default is IMO `@Internal`. Furthermore, `MailboxSender` might eventually become publicly exposed for user code to schedule tasks into the task's main thread, e.g. through the context of source functions, similar to the checkpoint lock to enable atomic operations. I would then mark it accordingly later, when it is exposed.

kezhuw:
@StefanRRichter After read through https://docs.google.com/document/d/1eDpsUKv2FqwZiS1Pm6gYO5eFHScBHfULKmH1-ZEWB4g, I have some doubts about this feature which may relate to your discussion with @1u0 about message abstraction, I use word "letter" in following description to avoid conflicting with @1u0's point:
* Will `StreamElement` handling, checkpoint handling be represented as letters in mailbox ?
* Is there any cases that sender of letter do not know how to handle that letter, eg. do not know how to wrap it as `Runnable` ?

If it is true to second question, I think it is the responsibility of mailbox, which has more knowledges than senders, to wrap not runnable letters as runnable letters if we don't want casting in processing of letters. Even in this case, I think overloading methods is a more appropriate approach than introducing generic message type as the latter is intrusive to type signature and may incur to unnecessary wrapper classes.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/MailboxImpl.java
```diff
@@ -0,0 +1,233 @@
...
+		return count == 0;
+	}
+
+	@Override
+	public void clearAndPut(@Nonnull Runnable shutdownAction) {
+		lock.lock();
+		try {
+			int localCount = count;
+			while (localCount > 0) {
+				ringBuffer[headIndex] = null;
```
kezhuw: What about `AutoCloseable` letters ?

StefanRRichter:
@kezhuw the letters will always be just `Runnable`, and right now there are two cases: checkpoint trigger and processing timers. Stream elements are processed as part of the default action, they don't go through the mailbox for performance reasons. Imagine the default method as the often-called, hot action of the task. The letters are non-frequent request from other threads to run some code for them in the task's main thread to avoid concurrency. Thats all.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/MailboxImpl.java
StefanRRichter: I don't understand what you mean and why a letter would be `AutoCloseable`. Just `Runnable` is enough.

StefanRRichter:
Maybe to extend the explanation a bit, the `MailboxSender` interface is in function not unlike the `Executor` interface. Just we want a bit more control over when letters are accepted in the queue.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/MailboxImpl.java
kezhuw: Previously, I thought there are maybe `Runnable` objects that wrap underlying resource, such as `Buffer`. After your explanation in main discussion thread, I think this is not true for checkpoint trigger and processing timers. I was somewhat paranoid in this case. Aside, thanks for your explanation in main thread.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/MailboxImpl.java
```diff
@@ -0,0 +1,233 @@
...
+	}
+
+	//------------------------------------------------------------------------------------------------------------------
+
+	@Override
+	public boolean tryPutMail(@Nonnull Runnable letter) {
+		final ReentrantLock lock = this.lock;
+		lock.lock();
+		try {
+			if (isFull()) {
```
Aitozi: A little doubt raise here that: as i know the mail box is enqueued by the timer or checkpoint action right? if this `enqueue` action should check the capacity of the queue, does this means that the timer trigger of checkpoint trigger may be put failed and have to retry with blocking the main thread?

pnowojski: `tryPutMail` is currently a dead unused method ([checkout this discussion](https://github.com/apache/flink/pull/8442#discussion_r284575775)). For checkpointing we will be using blocking variant of this method, which will have exactly the same semantic as before - one thing that we will need to double check is old behaviour vs current in case if we enqueue multiple checkpoint triggers.

StefanRRichter: @Aitozi thanks for thanking this through, here is how I concluded that this is not problematic:
- If you check the code, checkpoints and timers are even currently not invoked from the "main thread", but via executors.
- Right now both can also block on their attempt to acquire the checkpointing lock, which is very similar in effect to blocking on the mailbox.
- We have non-blocking ways of emitting letters to the mailbox, if the need arises.
Does this resolve your concerns?

Aitozi: Yes, my focus is on that do we need to block the enqueue operation if the queue is full, As before there is no check for the `backoff` (like the capacity here) of the `to do list` like timer and checkpoint. Although the default capacity of mailbox(64) is enough I think,  Because only the pending checkpoint and the first timer poll from timerservice should be enqueued to this queue IMO, but once this queue is full, do we need to block the main thread?

Aitozi: > @Aitozi thanks for thanking this through, here is how I concluded that this is not problematic:
> 
> * If you check the code, checkpoints and timers are even currently not invoked from the "main thread", but via executors.
> * Right now both can also block on their attempt to acquire the checkpointing lock, which is very similar in effect to blocking on the mailbox.
> * We have non-blocking ways of emitting letters to the mailbox, if the need arises.
> 
> Does this resolve your concerns?

sorry i do not refresh to see your reply just now, I will check the code again.

StefanRRichter: If you make the comparison, the current behavior of blocking directly on the checkpointing lock is essentially the same having a queue of size 1. I was thinking through this for a bit and there should no downside from this change for responsiveness, but if you think there is I would like to hear about it.

PR #8435, opened 2019-05-14
[FLINK-12443][table-planner-blink] Replace InternalType with LogicalType in blink

JingsongLi:

Use new LogicalType in blink planner.
NOTE: Use only LogicalType to override the previous InternalType functionality. Not implementing the new functionality of LogicalType.

LogicalType and InternalType:
1.LogicalType have nullable, so it has no singleton object.
2.YearMonthIntervalType and DayTimeIntervalType is an independent type.
3.MultisetType is not a MapType.

## Verifying this change

ut

  - The serializers: (no)

## Documentation


flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/functions/aggfunctions/DeclarativeAggregateFunction.java
```diff
@@ -62,7 +62,7 @@
 	/**
 	 * All types of the aggregate buffer.
 	 */
-	public abstract InternalType[] getAggBufferTypes();
+	public abstract LogicalType[] getAggBufferTypes();
```
wuchong: Regarding to `DeclarativeAggregateFunction`, I think it is an API expose to "users" or out of table system. Maybe we should use `DataType` for `getAggBufferTypes` and `getResultType`.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/logical/RowType.java
```diff
@@ -253,4 +257,12 @@ private static void validateFields(List<RowField> fields) {
 				String.format("Field names must be unique. Found duplicates: %s", duplicates));
 		}
 	}
+
+	public static RowType of(LogicalType... types) {
```
twalthr: Please extract changes that are unrelated to the issue into separate commits.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/logical/TimestampType.java
```diff
@@ -152,12 +152,14 @@ public boolean equals(Object o) {
 		if (!super.equals(o)) {
 			return false;
 		}
+
 		TimestampType that = (TimestampType) o;
-		return precision == that.precision;
+
+		return precision == that.precision && kind == that.kind;
```
twalthr: This was on purpose as mentioned in the comments of the constructor.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/logical/TypeInformationAnyType.java
```diff
@@ -68,6 +68,10 @@ public TypeInformationAnyType(TypeInformation<T> typeInfo) {
 		this(true, typeInfo);
 	}
 
+	public TypeInformationAnyType(Class<T> clazz) {
```
twalthr: This does not belong into a logical type but into some utility functions. It calls the type extractor and can potentially throw exceptions that need to be catched.

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/functions/aggfunctions/DeclarativeAggregateFunction.java
twalthr: Yes, DataType should be used at the edges.

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/functions/aggfunctions/RankLikeAggFunctionBase.java
```diff
@@ -98,30 +108,30 @@ protected Expression orderKeyEqualsExpression() {
 		return ret.orElseGet(() -> literal(true));
 	}
 
-	protected Expression generateInitLiteral(InternalType orderType) {
-		if (orderType.equals(InternalTypes.BOOLEAN)) {
+	protected Expression generateInitLiteral(LogicalType orderType) {
+		if (orderType instanceof BooleanType) {
```
twalthr: Leverage the logical type root instead of instanceof checks.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/calcite/RelTimeIndicatorConverter.scala
```diff
@@ -602,7 +602,7 @@ class RexTimeIndicatorMaterializerUtils(rexBuilder: RexBuilder) {
   private def timestamp(isNullable: Boolean): RelDataType = rexBuilder
     .getTypeFactory
     .asInstanceOf[FlinkTypeFactory]
-    .createTypeFromInternalType(InternalTypes.TIMESTAMP, isNullable = isNullable)
+    .createTypeFromLogicalType(new TimestampType(), isNullable = isNullable)
```
twalthr: Shouldn't the nullability only set in the logical type now and not be a separate flag in this method?

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/functions/aggfunctions/DeclarativeAggregateFunction.java
JingsongLi: I have different thought.
The `DeclarativeAggregateFunction` does not expose physical representation. It is the interface to description logic computing for user.

>The DataType class has two responsibilities: declaring a logical type and giving hints about the physical representation of data to the optimizer. While the logical type is mandatory, hints are optional but useful at the edges to other APIs.

So I think we should use `LogicalType` for `getAggBufferTypes` and `getResultType`.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/codegen/LongHashJoinGenerator.scala
```diff
@@ -44,11 +44,12 @@ object LongHashJoinGenerator {
...
         joinType == HashJoinType.PROBE_OUTER) &&
         filterNulls.forall(b => b) &&
-        keyType.getFieldTypes.length == 1 && {
-      val t = keyType.getTypeAt(0)
-      t == InternalTypes.LONG || t == InternalTypes.INT || t == InternalTypes.SHORT ||
-          t == InternalTypes.BYTE || t == InternalTypes.FLOAT || t == InternalTypes.DOUBLE ||
-          t.isInstanceOf[DateType] || t.isInstanceOf[TimestampType] || t == InternalTypes.TIME
+        keyType.getFieldCount == 1 && {
+      keyType.getTypeAt(0) match {
+        case _: BigIntType | _: IntType | _: SmallIntType | _: TinyIntType |
```
twalthr: Use type families instead?

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/types/logical/LogicalTypeUtils.java
```diff
@@ -0,0 +1,84 @@
...
+	public static boolean isPrimitive(LogicalType type) {
+		return type instanceof BooleanType || type instanceof TinyIntType || type instanceof SmallIntType ||
+				type instanceof IntType || type instanceof BigIntType || type instanceof FloatType ||
+				type instanceof DoubleType;
+	}
+
+	public static boolean isAtomic(LogicalType type) {
+		return isPrimitive(type) || type instanceof DecimalType || type instanceof VarCharType ||
+				type instanceof TimeType || type instanceof TimestampType || type instanceof DateType ||
+				type instanceof BinaryType || type instanceof YearMonthIntervalType || type instanceof DayTimeIntervalType;
```
twalthr: VarBinary is missing here.

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/types/logical/LogicalTypeUtils.java
```diff
@@ -0,0 +1,84 @@
...
+	}
+
+	/**
+	 * Returns whether this type should be automatically casted to
+	 * the target type in an arithmetic operation.
+	 */
+	public static boolean shouldAutoCastTo(LogicalType from, LogicalType to) {
+		if (from instanceof TinyIntType) {
```
twalthr: use logical type roots instead of instanceof checks

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/functions/aggfunctions/DeclarativeAggregateFunction.java
twalthr: The user should not be confronted to two different type stacks. That's why the description in `DataType` calls it "hints", a user should always use DataType. If he adds a hint where it is not appropriate, we just don't need to use those hints.

JingsongLi: Yes, I got it, it is better to let user just know one single type stack.

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/functions/aggfunctions/RankLikeAggFunctionBase.java
wuchong: Agree, `InternalType` is more similar to `LogicalTypeRoot`.

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/types/logical/LogicalTypeUtils.java
JingsongLi: I'll remove this useless method.

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/functions/aggfunctions/RankLikeAggFunctionBase.java
JingsongLi: Yeah, and Java can use `switch case`.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/logical/TimestampType.java
JingsongLi: Yes, `TimeIndicatorTypeInfo` and `SqlTimeTypeInfo.TIMESTAMP` also have same behavior.
But why are the kind does not affect equality and hashCode?
Is it for `Timestamp(6) of ROWTIME` and `Timestamp(6) of REGULAR` to interoperate?

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/functions/aggfunctions/RankLikeAggFunctionBase.java
JingsongLi: For simple codes or codes in Java, I prefer use the logical type root.
But for complex type codes in Scala, I prefer use type to check directly. Just like:
```
t match {
  case rowType: RowType => // use rowType do some thing
  case arrayType: ArrayType => // use arrayType do some thing
}
```
I think it is better than:
```
t.getTypeRoot match {
  case LogicalTypeRoot.ROW => t.asInstanceOf[RowType] // use rowType do some thing
  case LogicalTypeRoot.ARRAY => t.asInstanceOf[ArrayType] // use rowType do some thing
}
```
What do you think?

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/calcite/RelTimeIndicatorConverter.scala
JingsongLi: I think this PR is big enough. My idea is to override the functionality of `InternalType` first, and then to extend the functionality(like nullable) of `LogicalType` in the next PR.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/logical/RowType.java
JingsongLi: OK

flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/logical/TypeInformationAnyType.java
JingsongLi: Good catch, I'll take it out.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/codegen/LongHashJoinGenerator.scala
JingsongLi: Decimal is not supported here, so type families cannot be used. (Just because we haven't supported it yet)
I think I can use type families in `TypeCheckUtils`.

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/functions/aggfunctions/RankLikeAggFunctionBase.java
twalthr: I would recommend to use the provided visitors instead. I also have a default visitor in my FLINK-12254 branch. FLINK-12254 will add a lot of tooling around types as well. Maybe this PR was a bit too early, but this is no problem.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/logical/TimestampType.java
twalthr: Yes, we don't want to have special path and checks for time attributes throughout the stack. If we want to check, we can use `getKind()` but the type fully behaves like a timestamp otherwise. Also, the serialized string representation is always just timestamp, so serialization and deserialization leads to objects that are `equal`.

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/functions/aggfunctions/RankLikeAggFunctionBase.java
JingsongLi: Glad to see new visitors in FLINK-12254.
I looked at the code again and found that use type to check directly is not so strong.
The below code is not bad:
```
t.getTypeRoot match {
      case BOOLEAN => s"$rowTerm.getBoolean($indexTerm)"
      case TINYINT => s"$rowTerm.getByte($indexTerm)"
      case SMALLINT => s"$rowTerm.getShort($indexTerm)"
      ......
      case ROW => s"$rowTerm.getRow($indexTerm, ${t.asInstanceOf[RowType].getFieldCount})"
}
```
So I change all LogicalType instance match to the logical type root match.

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/dataformat/BinaryArray.java
```diff
@@ -55,20 +62,22 @@ public static int calculateHeaderInBytes(int numFields) {
 	 * It store real value when type is primitive.
 	 * It store the length and offset of variable-length part when type is string, map, etc.
 	 */
-	public static int calculateFixLengthPartSize(InternalType type) {
-		if (type.equals(InternalTypes.BOOLEAN)) {
+	public static int calculateFixLengthPartSize(LogicalType type) {
+		if (type instanceof BooleanType) {
```
KurtYoung: check `LogicalTypeRoot` instead?

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/dataformat/BinaryRow.java
```diff
@@ -94,16 +69,32 @@ public static int calculateFixPartSizeInBytes(int arity) {
 	 * If it is a fixed-length field, we can call this BinaryRow's setXX method for in-place updates.
 	 * If it is variable-length field, can't use this method, because the underlying data is stored continuously.
 	 */
-	public static boolean isInFixedLengthPart(InternalType type) {
+	public static boolean isInFixedLengthPart(LogicalType type) {
 		if (type instanceof DecimalType) {
-			return ((DecimalType) type).precision() <= DecimalType.MAX_COMPACT_PRECISION;
+			return ((DecimalType) type).getPrecision() <= Decimal.MAX_COMPACT_PRECISION;
```
KurtYoung: you can move decimal logic into the switch

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/runtime/rank/AbstractTopNFunction.java
```diff
@@ -153,12 +155,12 @@ public void open(Configuration parameters) throws Exception {
 
 		// initialize rankEndFetcher
 		if (!isConstantRankEnd) {
-			InternalType rankEndIdxType = inputRowType.getInternalTypes()[rankEndIndex];
-			if (rankEndIdxType.equals(InternalTypes.LONG)) {
+			LogicalType rankEndIdxType = inputRowType.getLogicalTypes()[rankEndIndex];
+			if (rankEndIdxType instanceof BigIntType) {
```
KurtYoung: check `LogicalTypeRoot` instead?

JingsongLi:
> I try to start reviewing from runtime package, and noticed that the old internal types haven't yet deleted, do you want to delete them in this PR or as a follow up issue?

I prefer delete InternalTypes in follow up PR, If we delete `InternalType`, we need to have a conversion of `LogicalType` and `TypeInformation`.

aljoscha:
I can't really comment too much, also because I think this is probably too big to have a sensible review.

One thing I noticed, though, is that this adds conversion logic in the Blink Planner code. @twalthr recently added code for that (https://github.com/apache/flink/blob/master/flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/utils/TypeConversions.java) and I think more conversion code is coming up. We shouldn't have code duplication here.

Another thing I noticed (you're probably already aware of this), is that in some places you convert logical types to `TypeInformation` only to then convert this back to logical types. For example `AggFunctionFactory.scala` has:

```
case DECIMAL =>
  val d = argTypes(0).asInstanceOf[DecimalType]
  val decimalTypeInfo = DecimalTypeInfo.of(d.getPrecision, d.getScale)
  new AvgAggFunction.DecimalAvgAggFunction(decimalTypeInfo)
```

and then `DecimalAvgAggFunction` converts this back:

```
@Override
public DataType getResultType() {
	DecimalType t = FlinkTypeSystem.inferAggAvgType(type.scale());
	return DataTypes.DECIMAL(t.getPrecision(), t.getScale());
}
```

This is already in the existing code, not introduced in this PR. I just noticed it now, though.

JingsongLi:
Hi @aljoscha 
>I can't really comment too much, also because I think this is probably too big to have a sensible review.

Sorry to be such a big PR. I tried to split it up, but it brought a lot of transformations of `InternalType` and `LogicalType` to ensure that every commit was compiled and passed, and finally I chose one PR.

>One thing I noticed, though, is that this adds conversion logic in the Blink Planner code. @twalthr recently added code for that (https://github.com/apache/flink/blob/master/flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/utils/TypeConversions.java) and I think more conversion code is coming up. We shouldn't have code duplication here.

Thank you for reminding me about the `TypeConvertions`. I took a quick look to it, But I think it's different from Blink's conversion right now.
1.Blink should deal with not nullable. (`TypeConvertions` now just deal with nullable true)
2.Blink's conversion will lose some information. Just like PojoTypeInfo, Blink just convert it to RowType, but `TypeConvertions` use `LegacyTypeInformationType` to support it.
Yeah, we shouldn't have code duplication here. I will reuse `TypeConvertions`, but in order to avoid further enlarged changes, I need to make some additional convertions based on it.

>Another thing I noticed (you're probably already aware of this), is that in some places you convert logical types to TypeInformation only to then convert this back to logical types. For example AggFunctionFactory.scala has:

Good catch, there are still many places where we shouldn't use `TypeInformation`, and I should change them all to use `LogicalType` in follow up PRs.

KurtYoung:
I would suggest to make this PR as clean as possible, by making the relationship with `TypeInformation`, `LogicalType` and calcite's `RelDataType` clean. I also observed issues similar with aljoscha's comment, and from my experience, it's very unlikely to be cleaned up in the future if we let this PR slip into master.

JingsongLi:
> I would suggest to make this PR as clean as possible, by making the relationship with `TypeInformation`, `LogicalType` and calcite's `RelDataType` clean. I also observed issues similar with aljoscha's comment, and from my experience, it's very unlikely to be cleaned up in the future if we let this PR slip into master.

Agree, we can track them from `LogicalTypeConverters`, but a lot of `TypeInformation` still block by `Expression` and `TableEnv` and `Udx`.

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/types/PlannerTypeConversions.java
```diff
@@ -0,0 +1,245 @@
...
+import java.time.LocalDateTime;
+import java.util.Optional;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+/**
+ * Converters for {@link LogicalType} and {@link DataType}.
+ */
+public class PlannerTypeConversions {
```
KurtYoung: this class mixed conversion between `LogicalType`, `DataType` and `TypeInformation`. It's better to split them into dedicated classes, and for each type of conversion, explain what scenarios would this conversion will happen. For example, when we need to convert `LogicalType` to `TypeInformation`

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/types/PlannerTypeUtils.java
```diff
@@ -0,0 +1,330 @@
...
+import static org.apache.flink.table.types.logical.LogicalTypeRoot.BIGINT;
+import static org.apache.flink.table.types.logical.LogicalTypeRoot.DOUBLE;
+import static org.apache.flink.table.types.logical.LogicalTypeRoot.FLOAT;
+import static org.apache.flink.table.types.logical.LogicalTypeRoot.INTEGER;
+import static org.apache.flink.table.types.logical.LogicalTypeRoot.SMALLINT;
+
+/**
+ * Utilities for {@link LogicalType} and {@link DataType}..
+ */
+public class PlannerTypeUtils {
```
KurtYoung: Also a mix of multiple convert methods with unclean purpose. We can at least strip off conversion between `LogicalType` and `Class` to another dedicate class and explain when this kind of conversion will happen.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/calcite/FlinkTypeFactory.scala
```diff
@@ -41,87 +40,112 @@ import scala.collection.JavaConverters._
...
   // NOTE: for future data types it might be necessary to
   // override more methods of RelDataTypeFactoryImpl
 
-  private val seenTypes = mutable.HashMap[(InternalType, Boolean), RelDataType]()
+  private val seenTypes = mutable.HashMap[LogicalType, RelDataType]()
 
-  def createTypeFromInternalType(
-      tp: InternalType,
-      isNullable: Boolean): RelDataType = {
+  def createTypeFromLogicalType(t: LogicalType, isNullable: Boolean): RelDataType = {
```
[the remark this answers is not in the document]
JingsongLi: OK, I'll replace it with `createTypeFromLogicalType(t.copy(nullable))`

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/functions/aggfunctions/AvgAggFunction.java
```diff
@@ -47,7 +44,7 @@
 	private UnresolvedReferenceExpression sum = new UnresolvedReferenceExpression("sum");
 	private UnresolvedReferenceExpression count = new UnresolvedReferenceExpression("count");
 
-	public abstract TypeInformation getSumType();
+	public abstract DataType getSumType();
```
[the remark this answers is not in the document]
JingsongLi: Look at @twalthr 's response.

>The user should not be confronted to two different type stacks. That's why the description in DataType calls it "hints", a user should always use DataType. If he adds a hint where it is not appropriate, we just don't need to use those hints.

In @wuchong 's first thread.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/calcite/FlinkTypeFactory.scala
```diff
@@ -406,78 +416,76 @@ object FlinkTypeFactory {
...
 
   def toInternalRowType(logicalRowType: RelDataType): RowType = {
-    // convert to InternalType
-    val logicalFieldTypes = logicalRowType.getFieldList.asScala map {
-      relDataType => FlinkTypeFactory.toInternalType(relDataType.getType)
-    }
-    // field names
-    val logicalFieldNames = logicalRowType.getFieldNames.asScala
-    new RowType(logicalFieldTypes.toArray[InternalType], logicalFieldNames.toArray)
+    new RowSchema(logicalRowType).internalType
```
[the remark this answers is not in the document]
JingsongLi: I'll delete it.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/calcite/FlinkTypeFactory.scala
```diff
@@ -189,35 +213,21 @@ class FlinkTypeFactory(typeSystem: RelDataTypeSystem) extends JavaTypeFactoryImp
...
-          InternalTypes.TIMESTAMP
-        case tpe: TypeInformation[_] => createInternalTypeFromTypeInfo(tpe)
+      tableSchema.getFieldDataTypes.map(_.getLogicalType).map {
+        case t: TimestampType if isStreaming.isDefined && !isStreaming.get =>
+          if (t.getKind == TimestampKind.REGULAR) t else new TimestampType(3)
+        case t => t
       })
   }
 
   def buildRelDataType(
```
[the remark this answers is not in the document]
JingsongLi: I'll change name to `buildLogicalRowType`

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/calcite/FlinkTypeFactory.scala
```diff
@@ -189,35 +213,21 @@ class FlinkTypeFactory(typeSystem: RelDataTypeSystem) extends JavaTypeFactoryImp
   def buildLogicalRowType(tableSchema: TableSchema, isStreaming: Option[Boolean]): RelDataType = {
```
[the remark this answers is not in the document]
JingsongLi: The original intention here is to convert all time indicators to regular `TimestampType` when it is batch mode.
You mean this conversion should be outside?

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/plan/schema/ArrayRelDataType.scala
```diff
@@ -18,7 +18,7 @@
 
 package org.apache.flink.table.plan.schema
 
-import org.apache.flink.table.`type`.ArrayType
+import org.apache.flink.table.types.logical.ArrayType
```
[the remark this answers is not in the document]
JingsongLi: Nice catch, Yes, Previously, the main purpose was to distinguish `BasicArrayTypeInfo`, `ObjectArrayTypeInfo`, `PojoTypeInfo`, `RowTypeInfo` and etc... Now the `LogicalType` are sql type. so I think we don't need `Array/Map/Multiset RelDataType`.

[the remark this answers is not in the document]
JingsongLi: NOTE: we should use `RelRecordType` with `StructKind.PEEK_FIELDS_NO_EXPAND` to replace `RowRelDataType`.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/calcite/FlinkTypeFactory.scala
```diff
@@ -147,12 +171,12 @@ class FlinkTypeFactory(typeSystem: RelDataTypeSystem) extends JavaTypeFactoryImp
     * Creates a struct type with the input fieldNames and input fieldTypes using FlinkTypeFactory
     *
     * @param fieldNames field names
-    * @param fieldTypes field types, every element is Flink's [[InternalType]]
+    * @param fieldTypes field types, every element is Flink's [[LogicalType]]
     * @return a struct type with the input fieldNames, input fieldTypes, and system fields
     */
   def buildLogicalRowType(
```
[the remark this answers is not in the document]
JingsongLi: The `buildLogicalRowType` return row type with `StructKind.FULLY_QUALIFIED`.
The `createTypeFromLogicalType` return row type with `StructKind.PEEK_FIELDS_NO_EXPAND`.
Them has different semantics.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/calcite/FlinkTypeFactory.scala
```diff
@@ -164,19 +188,19 @@ class FlinkTypeFactory(typeSystem: RelDataTypeSystem) extends JavaTypeFactoryImp
...
-    * @param fieldTypes     field types, every element is Flink's [[InternalType]]
+    * @param fieldTypes     field types, every element is Flink's [[LogicalType]]
     * @param fieldNullables field nullable properties
     * @return a struct type with the input fieldNames, input fieldTypes, and system fields
     */
   def buildLogicalRowType(
       fieldNames: Seq[String],
-      fieldTypes: Seq[InternalType],
+      fieldTypes: Seq[LogicalType],
       fieldNullables: Seq[Boolean]): RelDataType = {
```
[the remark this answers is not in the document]
JingsongLi: OK

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/calcite/FlinkTypeFactory.scala
```diff
@@ -406,78 +416,76 @@ object FlinkTypeFactory {
...
+  def toLogicalType(relDataType: RelDataType): LogicalType = {
+    val logicalType = relDataType.getSqlTypeName match {
+      case BOOLEAN => new BooleanType()
+      case TINYINT => new TinyIntType()
+      case SMALLINT => new SmallIntType()
+      case INTEGER => new IntType()
+      case BIGINT => new BigIntType()
+      case FLOAT => new FloatType()
+      case DOUBLE => new DoubleType()
+      case VARCHAR | CHAR => new VarCharType(VarCharType.MAX_LENGTH)
```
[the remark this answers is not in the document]
JingsongLi: >I remember you want disable all CHAR support

I have remove Java CHAR support, not sql ChAR.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/calcite/FlinkTypeFactory.scala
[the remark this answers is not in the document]
JingsongLi: I'll remove it.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/calcite/FlinkTypeFactory.scala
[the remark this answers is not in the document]
KurtYoung: OMG, how the caller suppose to tell this difference just seeing these two API?

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/calcite/FlinkTypeFactory.scala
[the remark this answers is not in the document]
KurtYoung: It's better to be outside, `FlinkTypeFactory` is just responsible to translate between flink's type and calcite's type. It's weird to depend on `TableSchema`

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/calcite/FlinkTypeFactory.scala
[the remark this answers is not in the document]
JingsongLi: `buildLogicalRowType` is for sql row from a `table`.
`createTypeFromLogicalType` just is for field types, it is a nested row when type is `RowType`.
I think the biggest difference should be the parameters.
Do you have any good idea to distinguish them.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/calcite/FlinkTypeFactory.scala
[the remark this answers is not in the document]
JingsongLi: OK, But we need modify `ExprCodeGenerator.generateResultExpression` to support assign `VARCHAR(A size)` to `VARCHAR(B size)`.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/calcite/FlinkTypeFactory.scala
```diff
@@ -406,78 +416,76 @@ object FlinkTypeFactory {
...
+      case VARBINARY | BINARY => new VarBinaryType(VarBinaryType.MAX_LENGTH)
```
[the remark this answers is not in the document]
JingsongLi: I'll keep precision.

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/types/PlannerTypeConversions.java
JingsongLi: OK

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/types/PlannerTypeUtils.java
JingsongLi: ok

JingsongLi:
Fixed comments and rebase. Please take a look.

KurtYoung:
Some tests failed

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/plan/schema/GenericRelDataType.scala
```diff
@@ -55,4 +57,8 @@ class GenericRelDataType(
   override def hashCode(): Int = {
     genericType.hashCode()
   }
+
+  override def generateTypeString(sb: lang.StringBuilder, withDetail: Boolean): Unit = {
```
JingsongLi: Now we delete `ArrayRelDataType` and use `ArraySqlType` and etc...
NOTE: `equals` of `ArraySqlType` is:
```
@Override public boolean equals(Object obj) {
    if (obj instanceof RelDataTypeImpl) {
      final RelDataTypeImpl that = (RelDataTypeImpl) obj;
      return this.digest.equals(that.digest);
    }
    return false;
  }
```
So we must need override `generateTypeString` all calcite extended type.

KurtYoung: Put the reason to the method comment

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/calcite/FlinkTypeFactory.scala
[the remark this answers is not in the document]
JingsongLi: @KurtYoung I have done some refactor to this in d5f0cb4 , what do you think?

PR #8436, opened 2019-05-14
[FLINK-12507][table-runtime-blink] Fix AsyncLookupJoin doesn't close all generated ResultFutures

wuchong:

The `AsyncLookupJoin` doesn't close all the generated ResultFutures (which contains UDFs in it). Because currently we iterate on the BlockingQueue to close. But when the job is failing (we are using `failingSource` to trigger restore),  some ResultFutures may not be callback and not in the BlockingQueue.

## Brief change log

  - Put all the ResultFutures to an ArrayList, and iterate on the list to close ResultFutures.

  - The serializers: (no)

## Documentation


KurtYoung:
+1

PR #8438, opened 2019-05-14
[FLINK-12152] Make the vcore that Application Master used configurable for Flink on YARN

yanghua:

*This pull request makes the vcore that Application Master used configurable for Flink on YARN*

## Brief change log

  - *Add a config option `vc` for Flink on YARN*

## Verifying this change

This change is already covered by existing tests, such as *testCommandLineClusterSpecification*.

## Documentation

### DIFF
--- a/flink-yarn/src/main/java/org/apache/flink/yarn/AbstractYarnClusterDescriptor.java
+++ b/flink-yarn/src/main/java/org/apache/flink/yarn/AbstractYarnClusterDescriptor.java
@@ -242,6 +242,14 @@ public abstract class AbstractYarnClusterDescriptor implements ClusterDescriptor
 			throw new YarnDeploymentException("Couldn't get cluster description, please check on the YarnConfiguration", e);
 		}
 
+		int configuredAmVcores = flinkConfiguration.getInteger(YarnConfigOptions.APP_MASTER_VCORES);
+		if (configuredAmVcores > numYarnMaxVcores) {
+			throw new IllegalConfigurationException(
+				String.format("The number of requested virtual cores for application master %d" +
+						" exceeds the maximum number of virtual cores %d available in the Yarn Cluster.",
+					configuredAmVcores, numYarnMaxVcores));
+		}
+
 		int configuredVcores = flinkConfiguration.getInteger(YarnConfigOptions.VCORES, clusterSpecification.getSlotsPerTaskManager());
 		// don't configure more than the maximum configured number of vcores
 		if (configuredVcores > numYarnMaxVcores) {
@@ -971,7 +979,7 @@ public abstract class AbstractYarnClusterDescriptor implements ClusterDescriptor
 		// Set up resource type requirements for ApplicationMaster
 		Resource capability = Records.newRecord(Resource.class);
 		capability.setMemory(clusterSpecification.getMasterMemoryMB());
-		capability.setVirtualCores(1);
+		capability.setVirtualCores(flinkConfiguration.getInteger(YarnConfigOptions.APP_MASTER_VCORES));
 
 		final String customApplicationName = customName != null ? customName : applicationName;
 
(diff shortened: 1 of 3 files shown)


yanghua:
@tillrohrmann The short config option name may need to be discussed. Currently, I used `vc`(vcore), it can also be `ac`(application master vcore) or `jc`(jobmanager vcore). What's your opinion?

yanghua:
@tillrohrmann Do you have any suggestions about the abbreviations of the parameters?

tillrohrmann:
What about only exposing this feature via the `flink-conf.yaml`? We could name the configuration parameter `yarn.appmaster.vcores` which is similar to `yarn.containers.vcores` which can be used to configure the vcores for the containers.

yanghua:
Actually, IMO, we should not expose this config option via the `flink-conf.yaml`. Because it's not a general purpose config option, it's used for YARN( a special cluster deploy mode). So I think to expose it as a flink-on-YARN config option make sense. WDYT?

tillrohrmann:
We already expose some deployment specific configuration options via the `flink-conf.yaml`. See https://ci.apache.org/projects/flink/flink-docs-stable/ops/config.html#yarn for example. Thus, this is ok to do.

yanghua:
@tillrohrmann It seems you are right, we did not use this mode so I have not paid attention. Sorry.

yanghua:
@tillrohrmann What about the new changes?

flink-yarn/src/main/java/org/apache/flink/yarn/AbstractYarnClusterDescriptor.java
```diff
@@ -229,6 +229,13 @@ private void isReadyForDeployment(ClusterSpecification clusterSpecification) thr
 			throw new YarnDeploymentException("Flink configuration object has not been set");
 		}
 
+		int configuredAmVcores = flinkConfiguration.getInteger(YarnConfigOptions.APP_MASTER_VCORES);
+		if (configuredAmVcores < YarnConfigOptions.APP_MASTER_VCORES.defaultValue()) {
```
tillrohrmann: I think we should compare `configuredAmVcores` against `numYarnMaxVcores` which we calculate a bit further down. I think we don't have to check that we request fewer cores than `1` because it won't take any effect.

flink-yarn/src/main/java/org/apache/flink/yarn/configuration/YarnConfigOptions.java
```diff
@@ -49,6 +49,14 @@
 			.defaultValue(-1)
 			.withDescription("The port where the application master RPC system is listening.");
 
+	/**
+	 * The vcores used by YARN application master.
+	 */
+	public static final ConfigOption<Integer> APP_MASTER_VCORES =
+		key("yarn.appmaster.vcores")
+		.defaultValue(1)
+		.withDescription("The number of virtual cores (vcores) used by YARN application master.");
```
tillrohrmann: I guess we need to regenerate the configuration documentation because we added a new config option: `mvn package -Dgenerate-config-docs -pl flink-docs`.

flink-yarn/src/test/java/org/apache/flink/yarn/YarnClusterDescriptorTest.java
```diff
@@ -147,6 +147,41 @@ public void testFailIfTaskSlotsHigherThanMaxVcores() throws ClusterDeploymentExc
...
+			fail("The deploy call should have failed.");
+		} catch (ClusterDeploymentException e) {
+			// we expect the cause to be an IllegalConfigurationException
+			if (!(e.getCause() instanceof IllegalConfigurationException)) {
+				throw e;
+			}
+		} finally {
+			clusterDescriptor.close();
+		}
+	}
```
tillrohrmann: I think we don't need this test here.

flink-yarn/src/main/java/org/apache/flink/yarn/configuration/YarnConfigOptions.java
yanghua: I have tried, but it does not take effect. In addition, it seems all the generated HTML documentation are not tracked by git.

tillrohrmann: You first need to install `flink-yarn`. I'll do it while merging.

PR #8442, opened 2019-05-14
[FLINK-12483] Support (legacy) SourceFunction as special case in the mailbox model for stream tasks

StefanRRichter:

This PR sits on top of #8431 and introduces compatibility code in `SourceStreamTask` so that sources can run with the mailbox model.

## Brief change log

- Only changes `SourceStreamTask` to run the operator in a separate thread, in parallel to the mailbox thread.

## Verifying this change

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SourceStreamTask.java
+++ b/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SourceStreamTask.java
@@ -45,6 +45,8 @@ import org.apache.flink.util.FlinkException;
 public class SourceStreamTask<OUT, SRC extends SourceFunction<OUT>, OP extends StreamSource<OUT, SRC>>
 	extends StreamTask<OUT, OP> {
 
+	private static final Runnable SOURCE_POISON_LETTER = () -> {};
+
 	private volatile boolean externallyInducedCheckpoints;
 
 	public SourceStreamTask(Environment env) {
@@ -101,12 +103,43 @@ public class SourceStreamTask<OUT, SRC extends SourceFunction<OUT>, OP extends S
 	protected void performDefaultAction(ActionContext context) throws Exception {
 		// Against the usual contract of this method, this implementation is not step-wise but blocking instead for
 		// compatibility reasons with the current source interface (source functions run as a loop, not in steps).
-		headOperator.run(getCheckpointLock(), getStreamStatusMaintainer());
+		final LegacySourceFunctionThread sourceThread = new LegacySourceFunctionThread();
+		sourceThread.start();
+
+		// We run an alternative mailbox loop that does not involve default actions and synchronizes around actions.
+		try {
+			runAlternativeMailboxLoop();
+		} catch (Exception mailboxEx) {
+			// We cancel the source function if some runtime exception escaped the mailbox.
+			if (!isCanceled()) {
+				cancelTask();
+			}
+			throw mailboxEx;
+		}
+
+		sourceThread.join();
+		sourceThread.checkThrowSourceExecutionException();
+
 		context.allActionsCompleted();
 	}
 
+	private void runAlternativeMailboxLoop() throws InterruptedException {
+
+		while (true) {
+
+			Runnable letter = mailbox.takeMail();
+			if (letter == SOURCE_POISON_LETTER) {
+				break;
+			}
+
+			synchronized (getCheckpointLock()) {
+				letter.run();
+			}
+		}
+	}
+
 	@Override
-	protected void cancelTask() throws Exception {
+	protected void cancelTask() {
 		if (headOperator != null) {
 			headOperator.cancel();
 		}
@@ -130,6 +163,35 @@ public class SourceStreamTask<OUT, SRC extends SourceFunction<OUT>, OP extends S
 			// we do not trigger checkpoints here, we simply state whether we can trigger them
 			synchronized (getCheckpointLock()) {
 				return isRunning();
+			}
+		}
+	}
+
+	/**
+	 * Runnable that executes the the source function in the head operator.
+	 */
+	private class LegacySourceFunctionThread extends Thread {
+
+		private Throwable sourceExecutionThrowable;
+
+		LegacySourceFunctionThread() {
+			this.sourceExecutionThrowable = null;
+		}
+
+		@Override
+		public void run() {
+			try {
+				headOperator.run(getCheckpointLock(), getStreamStatusMaintainer());
+			} catch (Throwable t) {
+				sourceExecutionThrowable = t;
+			} finally {
+				mailbox.clearAndPut(SOURCE_POISON_LETTER);
+			}
+		}
+
+		void checkThrowSourceExecutionException() throws Exception {
+			if (sourceExecutionThrowable != null) {
+				throw new Exception(sourceExecutionThrowable);
 			}
 		}
 	}
--- a/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SourceStreamTask.java
+++ b/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SourceStreamTask.java
@@ -98,8 +98,11 @@ public class SourceStreamTask<OUT, SRC extends SourceFunction<OUT>, OP extends S
 	}
 
 	@Override
-	protected void run() throws Exception {
+	protected void performDefaultAction(ActionContext context) throws Exception {
+		// Against the usual contract of this method, this implementation is not step-wise but blocking instead for
+		// compatibility reasons with the current source interface (source functions run as a loop, not in steps).
 		headOperator.run(getCheckpointLock(), getStreamStatusMaintainer());
+		context.allActionsCompleted();
 	}
 
 	@Override
--- a/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamIterationHead.java
+++ b/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamIterationHead.java
@@ -24,6 +24,7 @@ import org.apache.flink.streaming.api.watermark.Watermark;
 import org.apache.flink.streaming.runtime.io.BlockingQueueBroker;
 import org.apache.flink.streaming.runtime.io.RecordWriterOutput;
 import org.apache.flink.streaming.runtime.streamrecord.StreamRecord;
+import org.apache.flink.util.FlinkRuntimeException;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -41,88 +42,72 @@ public class StreamIterationHead<OUT> extends OneInputStreamTask<OUT, OUT> {
 
 	private static final Logger LOG = LoggerFactory.getLogger(StreamIterationHead.class);
 
-	private volatile boolean running = true;
+	private RecordWriterOutput<OUT>[] streamOutputs;
+
+	private final BlockingQueue<StreamRecord<OUT>> dataChannel;
+	private final String brokerID;
+	private final long iterationWaitTime;
+	private final boolean shouldWait;
 
 	public StreamIterationHead(Environment env) {
 		super(env);
+		final String iterationId = getConfiguration().getIterationId();
+		if (iterationId == null || iterationId.length() == 0) {
+			throw new FlinkRuntimeException("Missing iteration ID in the task configuration");
+		}
+
+		this.dataChannel = new ArrayBlockingQueue<>(1);
+		this.brokerID = createBrokerIdString(getEnvironment().getJobID(), iterationId,
+			getEnvironment().getTaskInfo().getIndexOfThisSubtask());
+		this.iterationWaitTime = getConfiguration().getIterationWaitTime();
+		this.shouldWait = iterationWaitTime > 0;
 	}
 
 	// ------------------------------------------------------------------------
 
 	@Override
-	protected void run() throws Exception {
+	protected void performDefaultAction(ActionContext context) throws Exception {
+		StreamRecord<OUT> nextRecord = shouldWait ?
+			dataChannel.poll(iterationWaitTime, TimeUnit.MILLISECONDS) :
+			dataChannel.take();
 
-		final String iterationId = getConfiguration().getIterationId();
-		if (iterationId == null || iterationId.length() == 0) {
-			throw new Exception("Missing iteration ID in the task configuration");
+		if (nextRecord != null) {
+			synchronized (getCheckpointLock()) {
+				for (RecordWriterOutput<OUT> output : streamOutputs) {
+					output.collect(nextRecord);
+				}
+			}
+		} else {
+			context.allActionsCompleted();
 		}
+	}
 
-		final String brokerID = createBrokerIdString(getEnvironment().getJobID(), iterationId ,
-				getEnvironment().getTaskInfo().getIndexOfThisSubtask());
+	// ------------------------------------------------------------------------
 
-		final long iterationWaitTime = getConfiguration().getIterationWaitTime();
-		final boolean shouldWait = iterationWaitTime > 0;
-
-		final BlockingQueue<StreamRecord<OUT>> dataChannel = new ArrayBlockingQueue<StreamRecord<OUT>>(1);
-
+	@SuppressWarnings("unchecked")
+	@Override
+	public void init() {
 		// offer the queue for the tail
 		BlockingQueueBroker.INSTANCE.handIn(brokerID, dataChannel);
 		LOG.info("Iteration head {} added feedback queue under {}", getName(), brokerID);
 
-		// do the work
-		try {
-			@SuppressWarnings("unchecked")
-			RecordWriterOutput<OUT>[] outputs = (RecordWriterOutput<OUT>[]) getStreamOutputs();
+		this.streamOutputs = (RecordWriterOutput<OUT>[]) getStreamOutputs();
 
-			// If timestamps are enabled we make sure to remove cyclic watermark dependencies
-			if (isSerializingTimestamps()) {
-				synchronized (getCheckpointLock()) {
-					for (RecordWriterOutput<OUT> output : outputs) {
-						output.emitWatermark(new Watermark(Long.MAX_VALUE));
-					}
-				}
-			}
-
-			while (running) {
-				StreamRecord<OUT> nextRecord = shouldWait ?
-					dataChannel.poll(iterationWaitTime, TimeUnit.MILLISECONDS) :
-					dataChannel.take();
-
-				if (nextRecord != null) {
-					synchronized (getCheckpointLock()) {
-						for (RecordWriterOutput<OUT> output : outputs) {
-							output.collect(nextRecord);
-						}
-					}
-				}
-				else {
-					// done
-					break;
+		// If timestamps are enabled we make sure to remove cyclic watermark dependencies
+		if (isSerializingTimestamps()) {
+			synchronized (getCheckpointLock()) {
+				for (RecordWriterOutput<OUT> output : streamOutputs) {
+					output.emitWatermark(new Watermark(Long.MAX_VALUE));
 				}
 			}
 		}
-		finally {
-			// make sure that we remove the queue from the broker, to prevent a resource leak
-			BlockingQueueBroker.INSTANCE.remove(brokerID);
-			LOG.info("Iteration head {} removed feedback queue under {}", getName(), brokerID);
-		}
 	}
 
 	@Override
-	protected void cancelTask() {
-		running = false;
-	}
-
-	// ------------------------------------------------------------------------
-
-	@Override
-	public void init() {
-		// does not hold any resources, no initialization necessary
-	}
-
-	@Override
-	protected void cleanup() throws Exception {
-		// does not hold any resources, no cleanup necessary
+	protected void cleanup() {
+		// make sure that we remove the queue from the broker, to prevent a resource leak
+		BlockingQueueBroker.INSTANCE.remove(brokerID);
+		LOG.info("Iteration head {} removed feedback queue under {}", getName(), brokerID);
 	}
 
 	// ------------------------------------------------------------------------
--- a/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
+++ b/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
@@ -55,8 +55,11 @@ import org.apache.flink.streaming.runtime.partitioner.ConfigurableStreamPartitio
 import org.apache.flink.streaming.runtime.partitioner.StreamPartitioner;
 import org.apache.flink.streaming.runtime.streamrecord.StreamRecord;
 import org.apache.flink.streaming.runtime.streamstatus.StreamStatusMaintainer;
+import org.apache.flink.streaming.runtime.tasks.mailbox.Mailbox;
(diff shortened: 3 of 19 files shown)


flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SourceStreamTask.java
```diff
@@ -45,6 +45,8 @@
 public class SourceStreamTask<OUT, SRC extends SourceFunction<OUT>, OP extends StreamSource<OUT, SRC>>
 	extends StreamTask<OUT, OP> {
 
+	private static final Runnable SOURCE_POISON_LETTER = () -> {};
```
Aitozi: Can these special mail instances put in a central place in `MailboxImpl.java` to be more clear?

StefanRRichter: I think that would be to wrong place: this letter is not special to the mailbox, it is only special to this task.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SourceStreamTask.java
```diff
@@ -130,4 +164,33 @@ public boolean triggerCheckpoint(CheckpointMetaData checkpointMetaData, Checkpoi
 			}
 		}
 	}
+
+	/**
+	 * Runnable that executes the the source function in the head operator.
+	 */
+	private class LegacySourceFunctionThread extends Thread {
+
+		private Throwable sourceExecutionThrowable;
```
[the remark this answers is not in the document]
StefanRRichter: It is not required to be volatile because we only evaluate it from a different thread after calling `join()` which guarantees visibility.

StefanRRichter:
@Aitozi Yes, this PR is only about the last commit. This is how we typically do Flink PRs that build upon other PRs. Saying, the PR is based on some other PR means: for the review, start only with the new commits that were not in the base-PR. Otherwise, how could not open followup PRs for work that is not yet merged.

Aitozi:
Thanks for explanation @StefanRRichter , +1 from my side.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SourceStreamTask.java
```diff
@@ -98,12 +100,44 @@ protected void cleanup() {
 	}
 
 	@Override
-	protected void run() throws Exception {
-		headOperator.run(getCheckpointLock(), getStreamStatusMaintainer());
+	protected void performDefaultAction(ActionContext context) throws Exception {
+		// Against the usual contract of this method, this implementation is not step-wise but blocking instead for
```
pnowojski: Wouldn't it be better to not hijack the execution from mailbox here? It's kind of fishy this way, when one task behaves in a completely different way. Maybe it would be better to run the mailbox loop in `StreamTask` always and this `performDefaultAction` would just spawn a source thread, without joining on it?

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamIterationHead.java
```diff
@@ -66,21 +66,20 @@ public StreamIterationHead(Environment env) {
...
 			dataChannel.take();
 
-		if (nextRecord == null) {
-			return false;
-		}
-
-		synchronized (getCheckpointLock()) {
-			for (RecordWriterOutput<OUT> output : streamOutputs) {
-				output.collect(nextRecord);
+		if (nextRecord != null) {
```
pnowojski: nit: invert if/else branches (shorter branch should go first)?

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
```diff
@@ -230,14 +242,29 @@ protected StreamTask(
...
 	 */
-	protected abstract boolean performDefaultAction() throws Exception;
+	protected abstract void performDefaultAction(ActionContext context) throws Exception;
 
-	protected void run() throws Exception {
-		while (isRunning && performDefaultAction()) {}
+	/**
+	 * Runs the stream-tasks main processing loop.
+	 */
+	private void run() throws Exception {
```
pnowojski: Could you benchmark the impact of those changes? It would be nice to measure `OneInputStreamTask` somehow in an isolation, unless we have good coverage of it in window benchmarks (`globalWindow`? `keyBy`?). Probably we need to either enable checkpointing for those benchmarks or add a separate benchmark(s) with checkpointing.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/MailboxReceiver.java
```diff
@@ -0,0 +1,59 @@
...
+
+	/**
+	 * Returns the oldest letter from the mailbox (head of queue) if the mailbox is not empty or
+	 * <code>null</code> otherwise.
+	 *
+	 * @return the oldest letter from the mailbox (head of queue) if the mailbox is not empty or
+	 * <code>null</code> otherwise.
+	 */
+	@Nullable
+	Runnable tryTakeMail();
```
pnowojski: `Optional`?

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/MailboxImpl.java
```diff
@@ -0,0 +1,233 @@
...
+import javax.annotation.concurrent.GuardedBy;
+
+import java.util.concurrent.locks.Condition;
+import java.util.concurrent.locks.ReentrantLock;
+
+/**
+ * Implementation of {@link Mailbox} inspired by {@link java.util.concurrent.ArrayBlockingQueue} and tailored towards
+ * our use case with multiple writers, single reader and volatile reads instead of lock & read on {@link #count}.
+ */
+public class MailboxImpl implements Mailbox {
```
pnowojski: `@Threadsafe`

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/MailboxReceiver.java
```diff
@@ -0,0 +1,59 @@
...
+
+	/**
+	 * This method returns the oldest letter from the mailbox (head of queue) or blocks until a letter is available.
+	 *
+	 * @return the oldest letter from the mailbox (head of queue).
+	 * @throws InterruptedException on interruption.
+	 */
+	@Nonnull
+	Runnable takeMail() throws InterruptedException;
```
pnowojski: Why do we need blocking variant? Isn't it a deadcode?

edit: ok, now I've found the usage. It should have been introduced in the latter commit.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/MailboxSender.java
```diff
@@ -0,0 +1,52 @@
...
+public interface MailboxSender {
+
+	/**
+	 * Enqueues the given letter to the mailbox, if capacity is available. On success, this returns <code>true</code>
+	 * and <code>false</code> if the mailbox was already full.
+	 *
+	 * @param letter the letter to enqueue.
+	 * @return <code>true</code> iff successful.
+	 */
+	boolean tryPutMail(@Nonnull Runnable letter);
```
pnowojski: Do we need non blocking variant here?

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/MailboxSender.java
```diff
@@ -0,0 +1,52 @@
...
+	 * @throws InterruptedException on interruption.
+	 */
+	void putMail(@Nonnull Runnable letter) throws InterruptedException;
+
+	/**
+	 * This method blocks until the mailbox has again capacity to enqueue new letters.
+	 *
+	 * @throws InterruptedException on interruption.
+	 */
+	void waitUntilHasCapacity() throws InterruptedException;
```
pnowojski: Ditto: do we need it?

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/Mailbox.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+public interface Mailbox extends MailboxReceiver, MailboxSender {
+
+	/**
+	 * The effect of this is that all pending letters are dropped and the given priorityAction
+	 * is enqueued to the head of the mailbox.
+	 *
+	 * @param priorityAction action to enqueue atomically after the mailbox was cleared.
+	 */
+	void clearAndPut(@Nonnull Runnable priorityAction);
```
pnowojski: I think `@Nonnull` is redundant and we should implicitly assume that parameters are not nullable unless stated differently.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/MailboxImpl.java
```diff
@@ -0,0 +1,233 @@
...
+		Preconditions.checkState(capacity > 0);
+		this.moduloMask = capacity - 1;
+		this.ringBuffer = new Runnable[capacity];
+		this.lock = new ReentrantLock();
+		this.notEmpty = lock.newCondition();
+		this.notFull = lock.newCondition();
+	}
+
+	@Override
+	public boolean hasMail() {
```
pnowojski: `isEmpty()`?

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SourceStreamTask.java
StefanRRichter: I was thinking about that, but then 1) how the you add the synchronization part around `run()` and 2) it would still loop around the defaultAction, but starting the thread should then be ignored? I think this is actually cleaner and more performant.

pnowojski: looping around (2nd and later executions) the default action should be a no-op. I guess it could even be an empty method and starting source thread could be done in another method/place. 

> 1) how the you add the synchronization part around run()

What do you mean?

StefanRRichter: That suggestion would need an `if` in the default action that triggers for the first invocation. Starting in another method, the question is which one and we also probably require to join back in another place then.

What I mean is, the alternative loop does two things different:
- no invocation of default action.
- running all letters inside a `synchronized` block.

So the question that remains is, how you plan to add the synchronized block to the existing mailbox loop? Personally I feel like this approach might be the cleanest and most performant, but I would be open to suggestions for a complete alternative if you have something in mind.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
StefanRRichter: Please see my comment here:
https://github.com/apache/flink/pull/8409#discussion_r284663084

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/MailboxSender.java
StefanRRichter: I think it can be helpful if client code that wants to run code can be non-blocking if required.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/MailboxSender.java
StefanRRichter: Same: For clients that want to be non-blocking w.r.t. the queue.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/MailboxImpl.java
StefanRRichter: Is empty would always be used inverted, so I think the positive phrase is ok.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/MailboxReceiver.java
StefanRRichter: I would prefer to keep this as returning `null` because it is a data structure and therefore in line with what JDK queues etc do.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamIterationHead.java
StefanRRichter: I don't think there is a general rule like this, also not in the wip styleguide. I mean it is as good as asking to put the "likely" case first, which this does.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/Mailbox.java
StefanRRichter: There is an additional benefit from adding this, which is that the compiler introduces an assertion for `!= null` if you run the code with `-ea`.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamIterationHead.java
pnowojski: It's a quite common code style convention.
```
if (condition) {
  instruction_1;
  instruction_2;
  ...
  instruction_15;
}
else {
  single_instruction;
}
```
Once you finally get to `else branch`, you might not longer remember what's this if all about. If reversed, you can quickly digest and forget the shorter branch, before moving to the else, still having fresh in mind what was before the `if` and inside the `condition`.

On the other hand, "likely" "unlikely" are:
- often a fuzzy concepts, that are subject to change and hard to evaluate
- often it doesn't matter to the reader, since you need to go through both branches anyway

But this is a nit, so if you are not convinced do as you are pleased.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SourceStreamTask.java
pnowojski: Ok, If I understand this correctly we have two choices:
1. (proposed in this PR) violate the contract of this method, hijack the mailbox execution and add checkpoint lock handling here
2. keep the contract of this method, but spill the hopefully soon to be deprecated checkpointing lock handling into the `StreamTask`

?

In that case you might be right that 1. sounds better.

One last question, have you considered how does this play with an idea of dropping `SourceTasks` and just using `OneInputStreamTask` to handle them? Will it be only possible for a connectors using new source interface and we would still keep this `SourceStreamTask` as a compatibility layer? Or could we somehow move this code to `OneInputStreamTask` as well?

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/Mailbox.java
pnowojski: Hmmm, how useful is that? Shouldn't we use `checkState`/`checkArgument`/`checkNotNull`/`requireNonNull` anyway? If not, shouldn't we add this to our code style?

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/MailboxImpl.java
pnowojski: ok, but in that case
1. please synchronise naming convention `isEmpty` + `isNonEmpty` or `hasMail` + `hasNoMail`
2. call one from the another (avoiding minor code duplication)

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/MailboxReceiver.java
pnowojski: Same reasoning should apply to `InputGate` for example and this is the same discussion that we had before. `@Nullable` is unfortunately worthless in our code base :( Since performance doesn't matter here, I would prefer `Optional` far more compared to maintaining a similarity with a Java collections - unless you want to implement their interfaces.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/MailboxSender.java
pnowojski: Yes sure, it can be helpful, but do we need it now? Why should we maintain and test a now unused code? We can always introduce it later and the later we introduce it, the fewer refactorings/bug fixing/maintenance time we spend on it.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SourceStreamTask.java
StefanRRichter: I was thinking through the alternatives and they have all more downsides in my opinion, performance wise and from the required modifications.

Now the "nice" thing about this compatibility layer is that we can play kind of the same trick when we remove the `SourceStreamTask` - we should always be able to just send runnables that fireup the source thread and hijack the mailbox. Question is then where and when to emit those in a nice way, but I think technially it is possible.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/MailboxReceiver.java
StefanRRichter: Even outside the question about how JDK does it, this class is just a data-structure, at the lowest level of abstraction and I feel like there is not truly any benefit a "forced protection" on this level. I can see benefits further up the stack, where semantics of interfaces start to be less clear about what to expect. Or would you retroactively like to add `Optional` returns on classes like `ArrayList` if you could?

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/Mailbox.java
StefanRRichter: Hmm, I would suggest to do it so because it has 3 benefits over the other options:
- Shorter to write and part of the signature makes it more visible
- IDE can show warnings in many cases when you try to violate it.
- Can be de/activated in production code.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/MailboxImpl.java
StefanRRichter: I don't understand exactly what the proposal is about. Right now there is no public `hasNoMail` so are you suggesting to introduce it? Or was there a wrong assumption that such a method exists?

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/MailboxSender.java
StefanRRichter: I was asking myself the same question, and here are my arguments why I decided to introduce it:
- most importantly, from all I can forsee, this interface will at some point be exposed to user code, e.g. via some source context. I think therefore it is ok to introduce things proactively if user can find a benefit in them instead of waiting for them to ask for it. And I can think of use cases where non-blocking support can be beneficial.
- symmetry in the interface between input and output.
- the test effort it took me is like adding 3 lines, the method has a low complexity. Opportunity cost to introduce it here is therefore reasonably low.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/Mailbox.java
pnowojski: Ok. I think it would be better to have some one common policy in this regard for the future, so maybe we will need to follow up on this at some point.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/MailboxImpl.java
pnowojski: There is `private boolean isEmpty()` :) That's why it was a bit confusing for me to have both of them and that assuming we want to keep both of them, that they are named with different conventions.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/MailboxReceiver.java
pnowojski: Retroactively I would enforce compile time checks for `@Nullable` :) But aside that, I wouldn't add `Optional` to java collections, because they are more generic and there are use cases where performance overhead of `Optional` matter. Here, that's not the case.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/MailboxSender.java
pnowojski: > most importantly, from all I can forsee, this interface will at some point be exposed to user code

if that's the case I agree. Having said that, are we planning to expose it?

StefanRRichter: Here is my line of thought: user code currently is using the checkpointing lock to execute atomic operations w.r.t. event processing. If we are planning to remove the lock, we should give users a proper substitute for the mechanism. Executing one letter can also be regarded an atomic operation w.r.t. event processing. The exposed interface would then be the `MailboxSender` interface, probably at least through whatever will be the future source contexts, maybe also in other places (sinks come to my mind). It does not take much creativity to think of cases where non-blocking options can help there as well.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/MailboxImpl.java
StefanRRichter: Oh yes, this method has a different semantic. I only introduced it to write to code in a more readable way instead of repeating the check code all over the place. Will think about it and maybe suffix those private methods with `...Internal`

StefanRRichter:
@pnowojski I think all points should be addressed in the latest version. Can you take a final look and approve if everything is ok?

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/MailboxImpl.java
pnowojski: nit: there is still this small inconsistency of naming `hasMail()` vs `isEmpty()`, but it barely matters

StefanRRichter:
Thanks for the reviews @pnowojski @Aitozi - merging.

PR #8443, opened 2019-05-14
[FLINK-12510][network] Fix deadlock in InputGates

pnowojski:
Because recursive calls reading from SingleInputGate or InputChannel can potentially trigger some
notifications to happen, it's better to not execute those calls under the locks in SingleInputGate
and UnionInputGate.

This change is covered by various existing ITCases.

## Documentation

  - If yes, how is the feature documented? (**not applicable /** docs / JavaDocs / not documented)

### DIFF
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/consumer/UnionInputGate.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/consumer/UnionInputGate.java
@@ -199,33 +199,22 @@ public class UnionInputGate extends InputGate {
 	private Optional<InputWithData<InputGate, BufferOrEvent>> waitAndGetNextData(boolean blocking)
 			throws IOException, InterruptedException {
 		while (true) {
-			InputGate inputGate;
-			synchronized (inputGatesWithData) {
-				while (inputGatesWithData.size() == 0) {
-					if (blocking) {
-						inputGatesWithData.wait();
-					} else {
-						resetIsAvailable();
-						return Optional.empty();
-					}
-				}
-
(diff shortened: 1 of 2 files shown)


flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/consumer/UnionInputGate.java
```diff
@@ -210,12 +211,16 @@ public void requestPartitions() throws IOException, InterruptedException {
...
+			}
 
-				// In case of inputGatesWithData being inaccurate do not block on an empty inputGate, but just poll the data.
-				Optional<BufferOrEvent> bufferOrEvent = inputGate.pollNextBufferOrEvent();
+			// In case of inputGatesWithData being inaccurate do not block on an empty inputGate, but just poll the data.
+			// Do not poll the gate under inputGatesWithData lock, since this can trigger notifications
+			// that could deadlock because of wrong locks taking order.
+			Optional<BufferOrEvent> bufferOrEvent = inputGate.pollNextBufferOrEvent();
 
+			synchronized (inputGatesWithData) {
```
StefanRRichter: I can see that this can fix the deadlock, but it also splits up a previously atomic operation into two parts. I could not spot a concrete problem with it, just wanted to make sure that we don't miss a potential problem here.

pnowojski: This was the old behaviour. I have concatenated those two critical sections together in a hotfix/refactoring commit 😳 

There should be no issue with this, since there are only two actors locking on {{inputGatesWithData}}: one reader of the input gate, and the guy/guys notifying that new data has arrived.

PR #8445, opened 2019-05-15
[FLINK-12127][network,config] Move network related options form TaskManagerOptions and NettyConfig into NetworkEnvironmentOptions

zhijiangW:

*The introduced `NetworkEnvironmentOptions` could be used for different shuffle services in future, not only for current task manager. So it is better to extract network related parameters as independent options.*

## Brief change log

  - *Extract network related options from `TaskManagerOptions` into `NetworkEnvironmentOptions*
  - *Migrate network related options from `NettyConfig` into `NetworkEnvironmentOptions`*
  - *Modify the `config.md` for generating new `network_configuration.html`*
  - *Adjust the related tests*

## Verifying this change

  - The serializers: (no)

## Documentation


zhijiangW:

zentol:
```
04:15:52.982 [INFO] There are 3 errors reported by Checkstyle 8.12 with /tools/maven/checkstyle.xml ruleset.
04:15:52.984 [ERROR] src/test/java/org/apache/flink/test/recovery/AbstractTaskManagerProcessFailureRecoveryTest.java:[31] (imports) ImportOrder: Import org.apache.flink.configuration.NetworkEnvironmentOptions appears after other imports that it should precede
04:15:52.986 [ERROR] src/test/java/org/apache/flink/test/cancelling/CancelingTestBase.java:[32] (imports) ImportOrder: Import org.apache.flink.configuration.NetworkEnvironmentOptions appears after other imports that it should precede
04:15:52.987 [ERROR] src/test/java/org/apache/flink/test/checkpointing/EventTimeWindowCheckpointingITCase.java:[35] (imports) ImportOrder: Import org.apache.flink.configuration.NetworkEnvironmentOptions appears after other imports that it should precede
```

zhijiangW:
Thanks for the notification @zentol !
I already updated the codes for the checkstyle issue.

flink-core/src/main/java/org/apache/flink/configuration/NetworkEnvironmentOptions.java
```diff
@@ -0,0 +1,220 @@
...
+import org.apache.flink.annotation.docs.ConfigGroups;
+
+import static org.apache.flink.configuration.ConfigOptions.key;
+
+/**
+ * The set of configuration options relating to network stack.
+ */
+@PublicEvolving
+@ConfigGroups(groups = @ConfigGroup(name = "NetworkEnvironment", keyPrefix = "network.environment"))
+public class NetworkEnvironmentOptions {
```
azagrebin: As I understood, we could move this class also to `org.apache.flink.runtime.io.network` if `org.apache.flink.docs.configuration.ConfigOptionsDocGenerator#LOCATIONS` is configured properly like it was for `NettyConfig` by including `org.apache.flink.runtime.io.network.netty` into `LOCATIONS` (which should be removed now).

azagrebin:
@flinkbot approve all

flink-core/src/main/java/org/apache/flink/configuration/TaskManagerOptions.java
```diff
@@ -97,6 +97,21 @@
 				" different TaskManagers need different values for this option, usually it is specified in an" +
 				" additional non-shared TaskManager-specific config file.");
 
+	/**
```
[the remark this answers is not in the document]
zhijiangW: Ok, I could revert this change. 
Just because it was in the network section, and after I migrated the network section out, it seems exist lonely.  :)

flink-core/src/main/java/org/apache/flink/configuration/NetworkEnvironmentOptions.java
zhijiangW: If we regard `NetworkEnvironmentOptions` as general one which could be used for other shuffle services future, then it makes sense to put it under configuration package like other options. And the `org.apache.flink.runtime.io.network.netty` could be removed. 

If we think of it as `NetworkEnvironment` private, it could be put under `network` package. And `org.apache.flink.runtime.io.network.netty` could be changed to `org.apache.flink.runtime.io.network`.

What do you think?

zhijiangW:
Thanks for the reviews @azagrebin !

I agree with your suggestion of metric naming for `ShuffleService` future. We could do that then.

flink-core/src/main/java/org/apache/flink/configuration/NetworkEnvironmentOptions.java
azagrebin: ok, looking more how other options are added, if we say that the network environment shuffle is the default one, it can also stay in the core package, agreed :)

zhijiangW:
FYI, I reverted the first hotfix commit and removed the unused `LOCATIONS`. @azagrebin

flink-core/src/main/java/org/apache/flink/configuration/NetworkEnvironmentOptions.java
```diff
@@ -0,0 +1,220 @@
...
+import org.apache.flink.annotation.docs.ConfigGroup;
+import org.apache.flink.annotation.docs.ConfigGroups;
+
+import static org.apache.flink.configuration.ConfigOptions.key;
+
+/**
+ * The set of configuration options relating to network stack.
+ */
+@PublicEvolving
+@ConfigGroups(groups = @ConfigGroup(name = "NetworkEnvironment", keyPrefix = "network.environment"))
```
zentol: I'm reasonable certain that this doesn't work properly, and that you haven't regenerated the config docs through the generator.
Did you modify the `.html` files manually?

I'm asking since the name of the `.html` isn't correct; it is derived from the class names and config group names. For both this class and group it will be `network_environment_options.html`.

This ConfigGroup has a few issues too. `network.environment` as a `keyPrefix` at this time does not make sense since there is no matching option. Additionally, as described above the name clashes with the name of the class, which likely only works since there's no option actually matching the group.

Finally, it seems sensible to extract the netty options into a separate file to keep related options closer to each other. The current layout would be
```
<some generic network options (like taskmanager.network.memory)>
<netty options>
<some more generic network options LIke taskmanager.network.request-backoff)>
```

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/netty/NettyConfig.java
```diff
@@ -38,53 +36,6 @@
 
 	private static final Logger LOG = LoggerFactory.getLogger(NettyConfig.class);
 
-	// - Config keys ----------------------------------------------------------
-
-	public static final ConfigOption<Integer> NUM_ARENAS = ConfigOptions
```
zentol: I'm not sure what the benefit of moving these is. The `NetworkEnvironment` _itself_ isn't even using them (and I supposed never will since they are details of the netty implementation).
I'd keep them here to prevent `NetworkEnvironmentOptions` from becoming a hodge-podge of options for various shuffle-service implementations as well as generic ones.

This is in a way quite similar to how we handle YARN/Mesos/etc.; the `ResourceManagerOptions` contain generic options for all container environments, but we still have distinct `*Options` classes for each.

azagrebin: `NetworkEnvironment` relies on `NetworkEnvironmentConfiguration` which is basically a Java parsed version of `NetworkEnvironmentOptions`. `NettyConfig` is part of `NetworkEnvironmentConfiguration`, also parsed and intended for `NetworkEnvironment`. It is all about netty used for shuffling. We also consider netty shuffle implementation to be the default one as it is basically now. I am not sure about splitting its config.  Other shuffle services might reuse these options, might not. Their additional options will be separate anyways, either own class or even outside of Flink.

flink-core/src/main/java/org/apache/flink/configuration/NetworkEnvironmentOptions.java
zhijiangW: Yes I modified the html file manually. Sorry for not fully understanding the config mechanism before. I would double check and adjust it.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/netty/NettyConfig.java
zhijiangW: Thanks for sharing the experience in `ResourceManagerOptions` @zentol !
The `NettyConfig` is actually used in current `NetworkEnvironment` which should start `NettyClient` and `NettyServer` based on `NettyConfig`. So all the options in `NetworkEnvironmentOptions` are used by `NetworkEnvironment` atm. But TBH I am also not sure the other shuffle services would reuse it or not, and it might rely on which kind of shuffle service we want to implement future. So the current `NetworkEnvironmentOptions` might be a special one for the default implementation or a general one for some implementations.

zhijiangW:
@zentol  I have updated the codes for addressing the issues. :)

zhijiangW:
Thanks for the further review @zentol . It is green. :)

PR #8446, opened 2019-05-15
[FLINK-12414] [runtime] Implement ExecutionGraph to SchedulingTopology

eaglewatcherwb:
adapter

Change-Id: I7ee2cbba0e2a7a94a0d6d19fc8851dde7661c4d7

*This pull request implements the ExecutionGraph to SchedulingTopology adapter*

## Brief change log

  - *Implement ExecutionGraph to SchedulingTopology adapter*
  - *Implement DefaultExecutionVertex of SchedulingExecutionVertex*
  - *Implement DefaultResultPartition of SchedulingResultPartition*

## Verifying this change

  - *Add unit test in ExecutionGraphToSchedulingTopologyAdapterTest*
  - *Add unit test in DefaultResultPartitionTest*
  - *Add unit test in DefaultExecutionVertexTest*

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/adapter/DefaultExecutionVertex.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/adapter/DefaultExecutionVertex.java
@@ -20,11 +20,11 @@ package org.apache.flink.runtime.scheduler.adapter;
 
 import org.apache.flink.api.common.InputDependencyConstraint;
 import org.apache.flink.runtime.execution.ExecutionState;
-import org.apache.flink.runtime.executiongraph.ExecutionVertex;
-import org.apache.flink.runtime.executiongraph.IntermediateResultPartition;
 import org.apache.flink.runtime.scheduler.strategy.ExecutionVertexID;
 import org.apache.flink.runtime.scheduler.strategy.SchedulingExecutionVertex;
 import org.apache.flink.runtime.scheduler.strategy.SchedulingResultPartition;
+
+import javax.xml.ws.Provider;
 
 import java.util.ArrayList;
 import java.util.Collection;
@@ -38,27 +38,26 @@ import static org.apache.flink.util.Preconditions.checkNotNull;
  */
 public class DefaultExecutionVertex implements SchedulingExecutionVertex {
 
-	private final ExecutionVertex executionVertex;
-
 	private final ExecutionVertexID executionVertexId;
 
 	private final List<SchedulingResultPartition> consumedPartitions;
 
 	private final List<SchedulingResultPartition> producedPartitions;
 
-	private InputDependencyConstraint inputDependencyConstraint;
+	private final InputDependencyConstraint inputDependencyConstraint;
 
-	public DefaultExecutionVertex(ExecutionVertex vertex) {
-		this.executionVertex = checkNotNull(vertex);
-		this.executionVertexId = new ExecutionVertexID(vertex.getJobvertexId(), vertex.getParallelSubtaskIndex());
-		this.inputDependencyConstraint = vertex.getJobVertex().getInputDependencyConstraint();
+	private final Provider<ExecutionState> stateProvider;
 
+	public DefaultExecutionVertex(
+			ExecutionVertexID executionVertexId,
+			List<SchedulingResultPartition> producedPartitions,
+			InputDependencyConstraint dependencyConstraint,
+			Provider<ExecutionState> stateProvider) {
+		this.executionVertexId = checkNotNull(executionVertexId);
+		this.inputDependencyConstraint = checkNotNull(dependencyConstraint);
 		this.consumedPartitions = new ArrayList<>();
-		Collection<IntermediateResultPartition> partitions = vertex.getProducedPartitions().values();
-		this.producedPartitions = new ArrayList<>(partitions.size());
-		for (IntermediateResultPartition irp : partitions) {
-			producedPartitions.add(new DefaultResultPartition(irp, this));
-		}
+		this.stateProvider = checkNotNull(stateProvider);
+		this.producedPartitions = checkNotNull(producedPartitions);
 	}
 
 	@Override
@@ -68,7 +67,7 @@ public class DefaultExecutionVertex implements SchedulingExecutionVertex {
 
 	@Override
 	public ExecutionState getState() {
-		return executionVertex.getExecutionState();
+		return stateProvider.invoke(null);
 	}
 
 	@Override
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/adapter/DefaultExecutionVertex.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/adapter/DefaultExecutionVertex.java
@@ -24,12 +24,11 @@ import org.apache.flink.runtime.scheduler.strategy.ExecutionVertexID;
 import org.apache.flink.runtime.scheduler.strategy.SchedulingExecutionVertex;
 import org.apache.flink.runtime.scheduler.strategy.SchedulingResultPartition;
 
-import javax.xml.ws.Provider;
-
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
+import java.util.function.Supplier;
 
 import static org.apache.flink.util.Preconditions.checkNotNull;
 
@@ -46,17 +45,17 @@ public class DefaultExecutionVertex implements SchedulingExecutionVertex {
 
 	private final InputDependencyConstraint inputDependencyConstraint;
 
-	private final Provider<ExecutionState> stateProvider;
+	private final Supplier<ExecutionState> stateSupplier;
 
 	public DefaultExecutionVertex(
 			ExecutionVertexID executionVertexId,
 			List<SchedulingResultPartition> producedPartitions,
 			InputDependencyConstraint dependencyConstraint,
-			Provider<ExecutionState> stateProvider) {
+			Supplier<ExecutionState> stateSupplier) {
 		this.executionVertexId = checkNotNull(executionVertexId);
 		this.inputDependencyConstraint = checkNotNull(dependencyConstraint);
 		this.consumedPartitions = new ArrayList<>();
-		this.stateProvider = checkNotNull(stateProvider);
+		this.stateSupplier = checkNotNull(stateSupplier);
 		this.producedPartitions = checkNotNull(producedPartitions);
 	}
 
@@ -67,7 +66,7 @@ public class DefaultExecutionVertex implements SchedulingExecutionVertex {
 
 	@Override
 	public ExecutionState getState() {
-		return stateProvider.invoke(null);
+		return stateSupplier.get();
 	}
 
 	@Override
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/adapter/DefaultExecutionVertex.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/adapter/DefaultExecutionVertex.java
@@ -18,7 +18,6 @@
 
 package org.apache.flink.runtime.scheduler.adapter;
 
-import org.apache.flink.api.common.InputDependencyConstraint;
 import org.apache.flink.runtime.execution.ExecutionState;
 import org.apache.flink.runtime.scheduler.strategy.ExecutionVertexID;
 import org.apache.flink.runtime.scheduler.strategy.SchedulingExecutionVertex;
@@ -41,19 +40,15 @@ public class DefaultExecutionVertex implements SchedulingExecutionVertex {
 
 	private final List<SchedulingResultPartition> consumedPartitions;
 
-	private final List<SchedulingResultPartition> producedPartitions;
-
-	private final InputDependencyConstraint inputDependencyConstraint;
+	private final List<? extends SchedulingResultPartition> producedPartitions;
 
 	private final Supplier<ExecutionState> stateSupplier;
 
 	public DefaultExecutionVertex(
 			ExecutionVertexID executionVertexId,
-			List<SchedulingResultPartition> producedPartitions,
-			InputDependencyConstraint dependencyConstraint,
+			List<? extends SchedulingResultPartition> producedPartitions,
 			Supplier<ExecutionState> stateSupplier) {
 		this.executionVertexId = checkNotNull(executionVertexId);
-		this.inputDependencyConstraint = checkNotNull(dependencyConstraint);
 		this.consumedPartitions = new ArrayList<>();
 		this.stateSupplier = checkNotNull(stateSupplier);
 		this.producedPartitions = checkNotNull(producedPartitions);
@@ -79,12 +74,7 @@ public class DefaultExecutionVertex implements SchedulingExecutionVertex {
 		return Collections.unmodifiableCollection(producedPartitions);
 	}
 
-	@Override
-	public InputDependencyConstraint getInputDependencyConstraint() {
-		return inputDependencyConstraint;
-	}
-
-	void addConsumedPartition(SchedulingResultPartition partition) {
+	<X extends SchedulingResultPartition> void addConsumedPartition(X partition) {
 		consumedPartitions.add(partition);
 	}
 }
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/adapter/DefaultExecutionVertex.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/adapter/DefaultExecutionVertex.java
@@ -1,0 +1,92 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.runtime.scheduler.adapter;
+
+import org.apache.flink.api.common.InputDependencyConstraint;
+import org.apache.flink.runtime.execution.ExecutionState;
+import org.apache.flink.runtime.executiongraph.ExecutionVertex;
+import org.apache.flink.runtime.executiongraph.IntermediateResultPartition;
+import org.apache.flink.runtime.scheduler.strategy.ExecutionVertexID;
+import org.apache.flink.runtime.scheduler.strategy.SchedulingExecutionVertex;
+import org.apache.flink.runtime.scheduler.strategy.SchedulingResultPartition;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+
+import static org.apache.flink.util.Preconditions.checkNotNull;
+
+/**
+ * Default implementation of {@link SchedulingExecutionVertex}.
+ */
+public class DefaultExecutionVertex implements SchedulingExecutionVertex {
+
+	private final ExecutionVertex executionVertex;
+
+	private final ExecutionVertexID executionVertexId;
+
+	private final List<SchedulingResultPartition> consumedPartitions;
+
+	private final List<SchedulingResultPartition> producedPartitions;
+
+	private InputDependencyConstraint inputDependencyConstraint;
+
+	public DefaultExecutionVertex(ExecutionVertex vertex) {
+		this.executionVertex = checkNotNull(vertex);
+		this.executionVertexId = new ExecutionVertexID(vertex.getJobvertexId(), vertex.getParallelSubtaskIndex());
+		this.inputDependencyConstraint = vertex.getJobVertex().getInputDependencyConstraint();
+
+		this.consumedPartitions = new ArrayList<>();
+		Collection<IntermediateResultPartition> partitions = vertex.getProducedPartitions().values();
+		this.producedPartitions = new ArrayList<>(partitions.size());
+		for (IntermediateResultPartition irp : partitions) {
+			producedPartitions.add(new DefaultResultPartition(irp, this));
+		}
+	}
+
+	@Override
+	public ExecutionVertexID getId() {
+		return executionVertexId;
+	}
+
+	@Override
+	public ExecutionState getState() {
+		return executionVertex.getExecutionState();
+	}
+
+	@Override
+	public Collection<SchedulingResultPartition> getConsumedResultPartitions() {
+		return Collections.unmodifiableCollection(consumedPartitions);
+	}
+
+	@Override
+	public Collection<SchedulingResultPartition> getProducedResultPartitions() {
+		return Collections.unmodifiableCollection(producedPartitions);
+	}
+
+	@Override
+	public InputDependencyConstraint getInputDependencyConstraint() {
+		return inputDependencyConstraint;
+	}
+
+	void addConsumedPartition(SchedulingResultPartition partition) {
+		consumedPartitions.add(partition);
+	}
+}
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/adapter/DefaultResultPartition.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/adapter/DefaultResultPartition.java
@@ -18,7 +18,6 @@
 
 package org.apache.flink.runtime.scheduler.adapter;
 
-import org.apache.flink.runtime.executiongraph.IntermediateResultPartition;
 import org.apache.flink.runtime.io.network.partition.ResultPartitionType;
 import org.apache.flink.runtime.jobgraph.IntermediateDataSetID;
 import org.apache.flink.runtime.jobgraph.IntermediateResultPartitionID;
@@ -41,31 +40,39 @@ import static org.apache.flink.util.Preconditions.checkNotNull;
  */
 public class DefaultResultPartition implements SchedulingResultPartition {
 
-	private final IntermediateResultPartition resultPartition;
+	private final IntermediateResultPartitionID resultPartitionId;
 
-	private final SchedulingExecutionVertex producer;
+	private final IntermediateDataSetID intermediateDataSetId;
+
+	private final ResultPartitionType partitionType;
+
+	private SchedulingExecutionVertex producer;
 
 	private final List<SchedulingExecutionVertex> consumers;
 
-	DefaultResultPartition(IntermediateResultPartition partition, SchedulingExecutionVertex producer) {
-		this.resultPartition = checkNotNull(partition);
-		this.producer = checkNotNull(producer);
+	DefaultResultPartition(
+			IntermediateResultPartitionID partitionId,
+			IntermediateDataSetID intermediateDataSetId,
+			ResultPartitionType partitionType) {
+		this.resultPartitionId = checkNotNull(partitionId);
+		this.intermediateDataSetId = checkNotNull(intermediateDataSetId);
+		this.partitionType = checkNotNull(partitionType);
 		this.consumers = new ArrayList<>();
 	}
 
 	@Override
 	public IntermediateResultPartitionID getId() {
-		return resultPartition.getPartitionId();
+		return resultPartitionId;
 	}
 
 	@Override
 	public IntermediateDataSetID getResultId() {
-		return resultPartition.getIntermediateResult().getId();
+		return intermediateDataSetId;
 	}
 
 	@Override
 	public ResultPartitionType getPartitionType() {
-		return resultPartition.getResultType();
+		return partitionType;
 	}
 
 	@Override
@@ -96,5 +103,9 @@ public class DefaultResultPartition implements SchedulingResultPartition {
 
 	void addConsumer(SchedulingExecutionVertex vertex) {
 		consumers.add(checkNotNull(vertex));
+	}
+
+	void setProducer(SchedulingExecutionVertex vertex) {
+		producer = checkNotNull(vertex);
 	}
 }
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/adapter/DefaultResultPartition.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/adapter/DefaultResultPartition.java
@@ -1,0 +1,100 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.runtime.scheduler.adapter;
+
+import org.apache.flink.runtime.executiongraph.IntermediateResultPartition;
+import org.apache.flink.runtime.io.network.partition.ResultPartitionType;
+import org.apache.flink.runtime.jobgraph.IntermediateDataSetID;
+import org.apache.flink.runtime.jobgraph.IntermediateResultPartitionID;
+import org.apache.flink.runtime.scheduler.strategy.SchedulingExecutionVertex;
+import org.apache.flink.runtime.scheduler.strategy.SchedulingResultPartition;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+
+import static org.apache.flink.runtime.scheduler.strategy.SchedulingResultPartition.ResultPartitionState.DONE;
+import static org.apache.flink.runtime.scheduler.strategy.SchedulingResultPartition.ResultPartitionState.EMPTY;
+import static org.apache.flink.runtime.scheduler.strategy.SchedulingResultPartition.ResultPartitionState.PRODUCING;
+import static org.apache.flink.runtime.scheduler.strategy.SchedulingResultPartition.ResultPartitionState.RELEASED;
+import static org.apache.flink.util.Preconditions.checkNotNull;
+
+/**
+ * Default implementation of {@link SchedulingResultPartition}.
+ */
+public class DefaultResultPartition implements SchedulingResultPartition {
+
+	private final IntermediateResultPartition resultPartition;
+
+	private final SchedulingExecutionVertex producer;
+
+	private final List<SchedulingExecutionVertex> consumers;
+
+	DefaultResultPartition(IntermediateResultPartition partition, SchedulingExecutionVertex producer) {
+		this.resultPartition = checkNotNull(partition);
+		this.producer = checkNotNull(producer);
+		this.consumers = new ArrayList<>();
+	}
+
+	@Override
+	public IntermediateResultPartitionID getId() {
+		return resultPartition.getPartitionId();
+	}
+
+	@Override
+	public IntermediateDataSetID getResultId() {
+		return resultPartition.getIntermediateResult().getId();
+	}
+
+	@Override
+	public ResultPartitionType getPartitionType() {
+		return resultPartition.getResultType();
+	}
+
+	@Override
+	public ResultPartitionState getState() {
+		switch (producer.getState()) {
+			case CREATED:
+			case SCHEDULED:
+			case DEPLOYING:
+				return EMPTY;
+			case RUNNING:
+				return PRODUCING;
+			case FINISHED:
+				return DONE;
+			default:
+				return RELEASED;
+		}
+	}
+
+	@Override
+	public SchedulingExecutionVertex getProducer() {
+		return producer;
+	}
+
+	@Override
+	public Collection<SchedulingExecutionVertex> getConsumers() {
+		return Collections.unmodifiableCollection(consumers);
+	}
+
+	void addConsumer(SchedulingExecutionVertex vertex) {
+		consumers.add(checkNotNull(vertex));
+	}
+}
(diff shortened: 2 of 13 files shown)


eaglewatcherwb:
Hi @tillrohrmann @GJL 
I have submitted the firs version of the PR. When I implemented this, both memory overhead and performance are considered and two plans are optional. The first is to only store `ExecutionGraph` in the adapter, construct and return a temporary variable when `getVertex` or `getResultPartition`. The advantage is that it involves the least memory overhead. But I am wondering that the performance may be poor, even constructing a single `SchedulingExecutionVertex` or `SchedulingResultPartition`, the producers and consumers will be recursively constructed, which will contain all the ancestor and descendant nodes, may be all the nodes of the graph in the worst case. Since `getVertex` and `getResultPartition` are frequently called when partition/execution state change, the performance will be critical. What's more, too many temporary variables will also lead to GC regularly, which is another time-consuming process.


Thus, in current PR, I adopt the second plan, which stores all the `SchedulingExecutionVertex` and `SchedulingResultPartition` in the adapter as member variables, which is of good performance but involves additional memory overhead to store the information.

What do you think? Any comment is highly appreciated.

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/adapter/DefaultExecutionVertexTest.java
```diff
@@ -0,0 +1,148 @@
...
+import static org.apache.flink.runtime.io.network.partition.ResultPartitionType.BLOCKING;
+import static org.apache.flink.runtime.jobgraph.DistributionPattern.ALL_TO_ALL;
+import static org.hamcrest.Matchers.containsInAnyOrder;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertThat;
+
+/**
+ * Unit tests for {@link DefaultExecutionVertex}.
+ */
+public class DefaultExecutionVertexTest {
```
tillrohrmann: `extends TestLogger` is missing

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/adapter/DefaultResultPartitionTest.java
```diff
@@ -0,0 +1,183 @@
...
+import static org.apache.flink.runtime.scheduler.strategy.SchedulingResultPartition.ResultPartitionState.PRODUCING;
+import static org.apache.flink.runtime.scheduler.strategy.SchedulingResultPartition.ResultPartitionState.RELEASED;
+import static org.hamcrest.Matchers.containsInAnyOrder;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertThat;
+
+/**
+ * Unit tests for {@link DefaultResultPartition}.
+ */
+public class DefaultResultPartitionTest {
```
tillrohrmann: Same here

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/adapter/DefaultResultPartition.java
```diff
@@ -0,0 +1,100 @@
...
+	public SchedulingExecutionVertex getProducer() {
+		return producer;
+	}
+
+	@Override
+	public Collection<SchedulingExecutionVertex> getConsumers() {
+		return Collections.unmodifiableCollection(consumers);
+	}
+
+	void addConsumer(SchedulingExecutionVertex vertex) {
```
tillrohrmann: Can't we assign the `consumers` in the constructor?

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/adapter/ExecutionGraphToSchedulingTopologyAdapter.java
```diff
@@ -0,0 +1,97 @@
...
+					scheduleVertex.addConsumedPartition(partition);
+					partition.addConsumer(scheduleVertex);
+				}
+			}
+		}
+	}
+
+	@Override
+	public Iterable<SchedulingExecutionVertex> getVertices() {
+		return Collections.unmodifiableList(executionVerticesList);
```
zentol: Make `executionVerticesList` an unmodifiable list instead.

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/adapter/ExecutionGraphToSchedulingTopologyAdapter.java
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	@Override
+	public Optional<SchedulingExecutionVertex> getVertex(ExecutionVertexID executionVertexId) {
+		return Optional.of(executionVertices.get(executionVertexId));
+	}
+
+	@Override
+	public Optional<SchedulingResultPartition> getResultPartition(IntermediateResultPartitionID intermediateResultPartitionId) {
+		return Optional.of(resultPartitions.get(intermediateResultPartitionId));
```
zentol: you likely want to use `Optional#ofNullable` here and in `#getVertex´, `Optional#of` fails with an NPE when applied to `null`.

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/adapter/ExecutionGraphToSchedulingTopologyAdapter.java
```diff
@@ -0,0 +1,97 @@
...
+	private final Map<IntermediateResultPartitionID, DefaultResultPartition> resultPartitions;
+
+	public ExecutionGraphToSchedulingTopologyAdapter(ExecutionGraph graph) {
+		Preconditions.checkNotNull(graph, "execution graph can not be null");
+
+		this.executionVertices = new HashMap<>();
+		this.resultPartitions = new HashMap<>();
+		this.executionVerticesList = new ArrayList<>();
+
+		Map<ExecutionVertex, DefaultExecutionVertex> executionVertexMap = new HashMap<>();
```
zentol: For all these collections, `graph.getTotalNumberOfVertices() * some_constant` is a better initial size.

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/adapter/DefaultExecutionVertexTest.java
```diff
@@ -0,0 +1,148 @@
...
+
+		schedulingExecutionVertices = new ArrayList<>();
+		adapter.getVertices().forEach(vertex -> schedulingExecutionVertices.add(vertex));
+		executionVertices = new ArrayList<>();
+		executionGraph.getAllExecutionVertices().forEach(vertex -> executionVertices.add(vertex));
+	}
+
+	@Test
+	public void testGetId() {
+		for (int idx = 0; idx < schedulingExecutionVertices.size(); idx++){
```
zentol: missing space before `{`

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/adapter/DefaultExecutionVertexTest.java
```diff
@@ -0,0 +1,148 @@
...
+			assertEquals(schedulingExecutionVertices.get(idx).getId().getJobVertexId(),
```
zentol: this would be more readable if both arguments were on separate lines

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/adapter/DefaultExecutionVertexTest.java
```diff
@@ -0,0 +1,148 @@
...
+				executionVertices.get(idx).getJobvertexId());
+			assertEquals(schedulingExecutionVertices.get(idx).getId().getSubtaskIndex(),
+				executionVertices.get(idx).getParallelSubtaskIndex());
+		}
+	}
+
+	@Test
+	public void testGetExecutionState() {
+		final ExecutionState[] states = ExecutionState.values();
+		Random random = new Random();
```
zentol: The random element here seems unnecessary. You could instead iterate over all states and set them once on a single vertex.

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/adapter/DefaultResultPartitionTest.java
```diff
@@ -0,0 +1,183 @@
...
+	}
+
+	@Test
+	public void testGetPartitionState() {
+		List<SchedulingExecutionVertex> schedulingExecutionVertices = new ArrayList<>();
+		executionGraph.getAllExecutionVertices().forEach(
+			vertex -> schedulingExecutionVertices.add(new DefaultExecutionVertex(vertex)));
+
+		final ExecutionState[] states = ExecutionState.values();
+		Random random = new Random();
```
zentol: same as in the other test, the random element is not necessary imo

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/adapter/DefaultResultPartitionTest.java
```diff
@@ -0,0 +1,183 @@
...
+	private ExecutionGraph executionGraph;
+
+	private ExecutionGraphToSchedulingTopologyAdapter adapter;
+
+	private List<IntermediateResultPartition> intermediateResultPartitions;
+
+	private List<SchedulingResultPartition> schedulingResultPartitions;
+
+	@Before
+	public void setUp() throws Exception {
```
zentol: This test is too complicated. It should not be necessary to create an entire ExecutionGraph or to use the adapter at all.

Just create a single partition, create a well-defined SchedulingExecutionVertex implementation, create **1** `DefaultResultPartiton`, and test it.

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/adapter/DefaultExecutionVertexTest.java
```diff
@@ -0,0 +1,148 @@
...
+		jobVertices[1] = createNoOpVertex(parallelism);
+		jobVertices[1].connectNewDataSetAsInput(jobVertices[0], ALL_TO_ALL, BLOCKING);
+		jobVertices[0].setInputDependencyConstraint(ALL);
+		jobVertices[1].setInputDependencyConstraint(ANY);
+		ExecutionGraph executionGraph = createSimpleTestGraph(
+			new JobID(),
+			taskManagerGateway,
+			triggeredRestartStrategy,
+			jobVertices);
+		ExecutionGraphToSchedulingTopologyAdapter adapter = new ExecutionGraphToSchedulingTopologyAdapter(executionGraph);
```
zentol: we should test this in isolation without the adapter being involved. You are introducing an indirection that could invalidate the tests at a later point.
Instead, create the `DefaultExecutioVertices` manually.

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/strategy/TestingSchedulingExecutionVertex.java
```diff
@@ -31,6 +32,8 @@
 
 	private final ExecutionVertexID executionVertexId;
 
+	private InputDependencyConstraint inputDependencyConstraint = InputDependencyConstraint.ANY;
```
zentol: can be final

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/adapter/DefaultExecutionVertexTest.java
```diff
@@ -0,0 +1,148 @@
...
+				.collect(Collectors.toList());
+			Set<IntermediateResultPartitionID> partitionIds2 = executionVertices
+				.get(idx).getProducedPartitions().keySet();
+			assertThat(partitionIds1, containsInAnyOrder(partitionIds2.toArray()));
+		}
+	}
+
+	@Test
+	public void testGetProducedResultPartitions() {
+		for (int idx = parallelism; idx < schedulingExecutionVertices.size(); idx++) {
```
zentol: should start at 0?

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/adapter/DefaultExecutionVertexTest.java
```diff
@@ -0,0 +1,148 @@
...
+/**
+ * Unit tests for {@link DefaultExecutionVertex}.
+ */
+public class DefaultExecutionVertexTest {
+
+	private final SimpleAckingTaskManagerGateway taskManagerGateway = new SimpleAckingTaskManagerGateway();
+
+	private final TestRestartStrategy triggeredRestartStrategy = TestRestartStrategy.manuallyTriggered();
+
+	private final int parallelism = 3;
```
zentol: the above 3 fields are unnecessary and could be in-lined into `#setup`

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/adapter/DefaultResultPartitionTest.java
```diff
@@ -0,0 +1,183 @@
...
+import static org.junit.Assert.assertThat;
+
+/**
+ * Unit tests for {@link DefaultResultPartition}.
+ */
+public class DefaultResultPartitionTest {
+
+	private final SimpleAckingTaskManagerGateway taskManagerGateway = new SimpleAckingTaskManagerGateway();
+
+	private final TestRestartStrategy triggeredRestartStrategy = TestRestartStrategy.manuallyTriggered();
```
zentol: the above2  fields are unnecessary and could be in-lined into `#setup`

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/adapter/DefaultExecutionVertexTest.java
eaglewatcherwb: OK

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/adapter/DefaultResultPartitionTest.java
eaglewatcherwb: OK

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/adapter/DefaultResultPartition.java
eaglewatcherwb: In current implementation, I would like to initialize `DefaultResultPartition` by the producedPartitions when initializing `DefaultExecutionVertex`, and the consumer `DefaultExecutionVertex` has not been initialized, thus not easy to assign the `consumers` in the constructor.

Another consideration of using `addConsumer` is that I would like to initialize the information of `DefaultExecutionVertex#consumedPartitions` and `DefaultResultPartition#consumers` in one step to avoid creating contradicting topologies.

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/adapter/ExecutionGraphToSchedulingTopologyAdapter.java
eaglewatcherwb: OK

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/adapter/ExecutionGraphToSchedulingTopologyAdapter.java
eaglewatcherwb: OK

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/adapter/ExecutionGraphToSchedulingTopologyAdapter.java
eaglewatcherwb: OK

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/adapter/DefaultExecutionVertexTest.java
eaglewatcherwb: OK

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/adapter/DefaultExecutionVertexTest.java
eaglewatcherwb: OK

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/adapter/DefaultExecutionVertexTest.java
eaglewatcherwb: OK

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/adapter/DefaultResultPartitionTest.java
eaglewatcherwb: OK

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/adapter/DefaultResultPartitionTest.java
eaglewatcherwb: OK, I will refine the code.

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/adapter/DefaultExecutionVertexTest.java
eaglewatcherwb: OK

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/strategy/TestingSchedulingExecutionVertex.java
eaglewatcherwb: OK

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/adapter/DefaultExecutionVertexTest.java
eaglewatcherwb: Sorry for the confusion of wrong method name, actually this was testing `ConsumedResultPartitions`, since only consumer jobVertex#2 has `InputEdge`, thus I start from `parallelism`.

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/adapter/DefaultExecutionVertexTest.java
eaglewatcherwb: `parallelism` would be used in `testGetConsumedResultPartitions`, I will move the other 2 fields into `#setup`

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/adapter/DefaultResultPartitionTest.java
eaglewatcherwb: OK

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/adapter/DefaultResultPartitionTest.java
eaglewatcherwb: OK

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/adapter/DefaultExecutionVertexTest.java
```diff
@@ -18,131 +18,112 @@
...
 	public void testGetExecutionState() {
 		final ExecutionState[] states = ExecutionState.values();
-		Random random = new Random();
-		for (ExecutionVertex executionVertex: executionVertices) {
-			setVertexState(executionVertex, states[Math.abs(random.nextInt()) % states.length]);
-		}
-
-		for (int idx = 0; idx < schedulingExecutionVertices.size(); idx++) {
-			assertEquals(schedulingExecutionVertices.get(idx).getState(), executionVertices.get(idx).getExecutionState());
+		for (ExecutionState state : states) {
```
zentol: could inline `ExecutionState.values()` to `for (ExecutionState state : ExecutionState.values()) {`

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/adapter/DefaultExecutionVertexTest.java
```diff
@@ -18,131 +18,112 @@
...
+			for (SchedulingExecutionVertex srp : schedulingExecutionVertices) {
```
zentol: it should be sufficient to test this for 1 vertex

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/adapter/DefaultResultPartitionTest.java
```diff
@@ -62,121 +49,73 @@
...
-	public void testBasicInfo() {
-		for (int idx = 0; idx < intermediateResultPartitions.size(); idx++) {
-			final IntermediateResultPartition partition = intermediateResultPartitions.get(idx);
-			final SchedulingResultPartition schedulingResultPartition = schedulingResultPartitions.get(idx);
-			assertEquals(partition.getPartitionId(), schedulingResultPartition.getId());
-			assertEquals(partition.getIntermediateResult().getId(), schedulingResultPartition.getResultId());
-			assertEquals(partition.getResultType(), schedulingResultPartition.getPartitionType());
-		}
+	public void setUp() {
+		schedulingExecutionVertices = new ArrayList<>(2);
```
zentol: we never iterate over the array, so we could just have 2 fields (producer and consumer), is a bit easier to read

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/adapter/DefaultResultPartitionTest.java
```diff
@@ -62,121 +49,73 @@
...
+		assertEquals(resultPartition.getProducer().getId(), schedulingExecutionVertices.get(0).getId());
 	}
 
 	@Test
 	public void testGetPartitionState() {
-		List<SchedulingExecutionVertex> schedulingExecutionVertices = new ArrayList<>();
-		executionGraph.getAllExecutionVertices().forEach(
-			vertex -> schedulingExecutionVertices.add(new DefaultExecutionVertex(vertex)));
-
 		final ExecutionState[] states = ExecutionState.values();
```
zentol: in-line, as above

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/adapter/DefaultExecutionVertex.java
```diff
@@ -0,0 +1,91 @@
...
+
+	private final ExecutionVertexID executionVertexId;
+
+	private final List<SchedulingResultPartition> consumedPartitions;
+
+	private final List<SchedulingResultPartition> producedPartitions;
+
+	private final InputDependencyConstraint inputDependencyConstraint;
+
+	private final Provider<ExecutionState> stateProvider;
```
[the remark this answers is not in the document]
eaglewatcherwb: OK,I will use `Supplier`.

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/adapter/DefaultExecutionVertexTest.java
```diff
@@ -18,131 +18,112 @@
...
-	@Test
-	public void testGetProducedResultPartitions() {
-		for (int idx = parallelism; idx < schedulingExecutionVertices.size(); idx++) {
-			Collection<IntermediateResultPartitionID> partitionIds1 = schedulingExecutionVertices
-				.get(idx).getConsumedResultPartitions().stream().map(SchedulingResultPartition::getId)
-				.collect(Collectors.toList());
+	/**
+	 * A simple implementation of {@link Provider} for testing.
+	 */
+	public static class ExecutionStateProviderTest implements Provider<ExecutionState> {
```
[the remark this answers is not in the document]
eaglewatcherwb: package private is OK.

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/adapter/DefaultExecutionVertexTest.java
```diff
@@ -18,131 +18,112 @@
...
 
 	@Test
-	public void testGetInputDependencyConstraint() {
-		for (int idx = 0; idx < schedulingExecutionVertices.size(); idx++) {
-			assertEquals(schedulingExecutionVertices.get(idx).getInputDependencyConstraint(),
-				executionVertices.get(idx).getJobVertex().getInputDependencyConstraint());
-		}
+	public void testGetProducedResultPartitions() {
+		Collection<IntermediateResultPartitionID> partitionIds1 =  schedulingExecutionVertices
+			.get(0).getProducedResultPartitions().stream().map(SchedulingResultPartition::getId)
```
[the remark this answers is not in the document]
eaglewatcherwb: OK

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/adapter/DefaultExecutionVertexTest.java
```diff
@@ -18,131 +18,112 @@
...
+		assertThat(partitionIds1, containsInAnyOrder(partitionIds2.toArray()));
 	}
 
 	@Test
 	public void testGetConsumedResultPartitions() {
-		for (int idx = 0; idx < schedulingExecutionVertices.size(); idx++) {
-			Collection<IntermediateResultPartitionID> partitionIds1 =  schedulingExecutionVertices
-				.get(idx).getProducedResultPartitions().stream().map(SchedulingResultPartition::getId)
+			Collection<IntermediateResultPartitionID> partitionIds1 = schedulingExecutionVertices
+				.get(1).getConsumedResultPartitions().stream().map(SchedulingResultPartition::getId)
```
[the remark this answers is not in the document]
eaglewatcherwb: OK

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/adapter/DefaultExecutionVertexTest.java
```diff
@@ -18,131 +18,112 @@
...
 
-	@Test
-	public void testGetId() {
-		for (int idx = 0; idx < schedulingExecutionVertices.size(); idx++){
-			assertEquals(schedulingExecutionVertices.get(idx).getId().getJobVertexId(),
-				executionVertices.get(idx).getJobvertexId());
-			assertEquals(schedulingExecutionVertices.get(idx).getId().getSubtaskIndex(),
-				executionVertices.get(idx).getParallelSubtaskIndex());
-		}
+		schedulingExecutionVertices = new ArrayList<>(2);
```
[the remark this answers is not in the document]
eaglewatcherwb: OK

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/adapter/DefaultExecutionVertexTest.java
eaglewatcherwb: OK

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/adapter/DefaultExecutionVertexTest.java
eaglewatcherwb: OK

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/adapter/DefaultResultPartitionTest.java
eaglewatcherwb: OK

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/adapter/DefaultResultPartitionTest.java
eaglewatcherwb: OK

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/adapter/DefaultResultPartition.java
zentol: I'm wondering whether this circular dependency between the partitions and vertices isn't a design flaw in the interfaces. (Also applies to the Failover interfaces)

Similarly to the existing `Execution*` classes, we create initially independent nodes and then retroactively bake a graph structure into them by setting partitions/producers respectively.

Maybe we could instead have the vertices only be these really independent nodes (with an ID and state), setup the partitions afterwards as the graph structure (which, now that I think about it, are awfully close to `ExecutionEdges`, the differentiation becomes quite blurry), and use the vertices to lookup partitions in the graph. (something like `SchedulingTopology#getProcudedResultPartitions(ExecutionVertexID id)`)

@tillrohrmann @GJL WDYT?

tillrohrmann: I think you are right that in order to break up the cyclic dependency we would need to decouple the `SchedulingResultPartition` from the `SchedulingExecutionVertex` (via the `ExecutionVertexID` for example).

I'm wondering how bad this cyclic dependency and the need for mutable state is, though. From a user's perspective, the existing interfaces are a bit easier and more convenient to use. On the down side, it makes the implementation a bit harder and harder to test in isolation. However, do we want to test the `Scheduling*` implementations in isolation? Moreover, you always have this problem in graph structures which have bidirectional edges or loops.

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/adapter/ExecutionGraphToSchedulingTopologyAdapterTest.java
```diff
@@ -0,0 +1,188 @@
...
+
+			assertPartitionsEquals(originalConsumedPartitions, adaptedConsumedPartitions);
+
+			Collection<IntermediateResultPartition> originalProducedPartitions = originalVertex.getProducedPartitions().values();
+			Collection<SchedulingResultPartition> adaptedProducedPartitions = adaptedVertex.getProducedResultPartitions();
+
+			assertPartitionsEquals(originalProducedPartitions, adaptedProducedPartitions);
+		}
+		if (adaptedVertices.hasNext()) {
+			fail("Number of adapted vertices exceeds number of original vertices.");
```
GJL: if block can be replaced with

```suggestion
assertFalse("Number of adapted vertices exceeds number of original vertices.", adaptedVertices.hasNext());
```

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/adapter/DefaultResultPartition.java
```diff
@@ -0,0 +1,111 @@
...
+import static org.apache.flink.runtime.scheduler.strategy.SchedulingResultPartition.ResultPartitionState.DONE;
+import static org.apache.flink.runtime.scheduler.strategy.SchedulingResultPartition.ResultPartitionState.EMPTY;
+import static org.apache.flink.runtime.scheduler.strategy.SchedulingResultPartition.ResultPartitionState.PRODUCING;
+import static org.apache.flink.runtime.scheduler.strategy.SchedulingResultPartition.ResultPartitionState.RELEASED;
+import static org.apache.flink.util.Preconditions.checkNotNull;
+
+/**
+ * Default implementation of {@link SchedulingResultPartition}.
+ */
+public class DefaultResultPartition implements SchedulingResultPartition {
```
GJL: Does it have to be `public`?

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/adapter/DefaultExecutionVertex.java
```diff
@@ -0,0 +1,80 @@
...
+import java.util.Collections;
+import java.util.List;
+import java.util.function.Supplier;
+
+import static org.apache.flink.util.Preconditions.checkNotNull;
+
+/**
+ * Default implementation of {@link SchedulingExecutionVertex}.
+ */
+public class DefaultExecutionVertex implements SchedulingExecutionVertex {
```
GJL: Does it have to be `public`?

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/adapter/DefaultResultPartition.java
GJL: Did we consider naming this class `DefaultSchedulingResultPartition` ? Same with the `SchedulingExecutionVertex` implementation. Would that be too verbose?

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/adapter/DefaultResultPartition.java
```diff
@@ -0,0 +1,111 @@
...
+			case CREATED:
+			case SCHEDULED:
+			case DEPLOYING:
+				return EMPTY;
+			case RUNNING:
+				return PRODUCING;
+			case FINISHED:
+				return DONE;
+			default:
+				return RELEASED;
```
GJL: Was it a conscious decision to return `RELEASED` as the default? Since `RELEASED` is currently not implemented, I'd rather not see it in production code.

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/adapter/ExecutionGraphToSchedulingTopologyAdapter.java
```diff
@@ -0,0 +1,156 @@
...
+	public ExecutionGraphToSchedulingTopologyAdapter(ExecutionGraph graph) {
+		checkNotNull(graph, "execution graph can not be null");
+
+		final int totalVertexCnt = graph.getTotalNumberOfVertices();
+		int totalPartitionCnt = 0;
+		for (ExecutionJobVertex jobVertex : graph.getAllVertices().values()) {
+			totalPartitionCnt += jobVertex.getProducedDataSets().length * jobVertex.getParallelism();
+		}
+
+		this.executionVerticesById = new HashMap<>(totalVertexCnt);
```
GJL: I'd suggest to use the constructor with no arguments because:
- It is easier to read
- The default load factor is 0.75, i.e., after inserting approx. 0.75 * totalVertexCnt, elements will be re-hashed anyways
- Using `ceil(totalVertexCnt/0.75)` would be harder to understand
- In this code path there won't be a noticeable performance hit due to re-hashing

Also see https://stackoverflow.com/questions/564899/choosing-the-initial-capacity-of-a-hashset-with-an-expected-number-of-unique-val

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/adapter/ExecutionGraphToSchedulingTopologyAdapter.java
```diff
@@ -0,0 +1,156 @@
...
+		this.executionVerticesList = new ArrayList<>(totalVertexCnt);
+		Map<IntermediateResultPartitionID, DefaultResultPartition> tmpResultPartitionsById = new HashMap<>(totalPartitionCnt);
```
GJL: See my other comment.

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/adapter/ExecutionGraphToSchedulingTopologyAdapter.java
```diff
@@ -0,0 +1,156 @@
...
+		Map<ExecutionVertex, DefaultExecutionVertex> executionVertexMap = new HashMap<>(totalVertexCnt);
```
GJL: See my other comment.

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/adapter/ExecutionGraphToSchedulingTopologyAdapterTest.java
eaglewatcherwb: OK

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/adapter/DefaultExecutionVertex.java
eaglewatcherwb: OK, package-private is enough.

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/adapter/DefaultResultPartition.java
eaglewatcherwb: Emm, `DefaultSchedulingResultPartition` and `DefaultSchedulingExecutionVertex` should be better, once I also thought `DefaultSchedulingExecutionVertex` may be a little confused whether the default implementation of  `ExecutionVertex` or `SchedulingExecutionVertex`. Although a little more words, it is worthy of clear names.

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/adapter/DefaultResultPartition.java
eaglewatcherwb: The annotation of `ResultPartitionState#EMPTY` is 
> Producer is not yet running.

I found that `RELEASED` is more proper than other enum value for `CANCELING`, `CANCELED`, `FAILED`, `RECONCILING`, so I return it.

What's your opinion of the mapping of the `ExecutionState` mentioned above ? Shall we return `ResultPartitionState#EMPTY` and modify the annotation to
> Producer is not yet running or in abnormal state.

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/adapter/ExecutionGraphToSchedulingTopologyAdapter.java
eaglewatcherwb: OK, the reason is reasonable.

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/adapter/ExecutionGraphToSchedulingTopologyAdapter.java
eaglewatcherwb: OK

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/adapter/ExecutionGraphToSchedulingTopologyAdapter.java
eaglewatcherwb: OK

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/adapter/DefaultResultPartition.java
GJL: Javadoc for `EMPTY` should be rewritten – can be done in a follow up. I would say that `RELEASED` is a terminal state which is transitioned into as a result of a conscious decision to release the partition. Task failure should not result in a _"released"_ partition.

cc: @tillrohrmann

tillrohrmann: I guess that in case of a failure we should return the partition state to the initial state (e.g. `EMPTY`).

eaglewatcherwb: OK, in the latest PR, I return `EMPTY` for the failure state.

PR #5399, opened 2018-02-02
[FLINK-6206] [runtime] Use LOG.error() when logging failure state changes

casidiablo:
It's very inconvenient to have these logged with `INFO`. It makes it hard to detect errors when inspecting logs with, say, Kibana.

PR #8447, opened 2019-05-15
[hotfix][runtime] Set log level to error when job transits to erroneous states

link3280:

When a job/task fails and transits to an erroneous state, we should log it on error level instead of info level.

## Brief change log

* Set log level to error when job or task transits states with an exception 

## Verifying this change

  - The serializers: no

## Documentation

  - If yes, how is the feature documented?


zentol:
Duplicate of #5399.

PR #490, opened 2015-03-17
[FLINK-1714]Fix the bug of logger class loader

matadorhong:
For log in class SecurityUtils, the  LoggerFactory.getLogger() function should be for "SecurityUtils" not "MapPartitionDriver". It is a bug I think

PR #472, opened 2015-03-10
[FLINK-1594] [streaming] [wip] DataStream supports self-connection

gaborhermann:
* Added StreamEdge abstraction, so more than one edge can be defined between stream operators.
* Added OutputSelectorWrapper to avoid directed emission spreading over all the code.
* Added factories for reader and writer creation

Travis fails on StreamVertexTest.coTest() (it gets stuck). Locally it passes.

PR #470, opened 2015-03-10
[FLINK-1576] [gelly] improvements to the gelly examples.

vasia:
Updated GraphMetrics, MusicProfiles and PageRank to run with and without parameters.
Added input descriptions to LabelPropagation and SSSP.
Fixed some minor issues in the SSSP example.
Fixed a bug in MusicProfiles that wasn't generating the user-user graph properly.
Changed the PageRank library method to initialize the vertex ranks.

PR #468, opened 2015-03-09
 [FLINK-1629][FLINK-1630][FLINK-1547] Rework Flink on YARN

rmetzger:
The main change here is a reworked container scheduling logic in the YARN ApplicationMaster.

This commit is changing:
[FLINK-1629]: users can now "fire and forget" jobs to YARN or YARN sessions to there. (Detached mode)
[FLINK-1630]: YARN is now reallocating failed YARN containers during the lifetime of a YARN session.
[FLINK-1547]: Users can now specify if they want the ApplicationMaster (= the JobManager = the entire YARN session) to restart on failure, and how often. After the first restart, the session will behave like a detached session. There is now backup of state between the old and the new AM.

The whole resource negotiation process between the RM and the AM has been reworked.
 Flink is now much more flexible when requesting new containers and also giving back uneeded containers.

A new test case is testing the container restart. It is also verifying that the web frontend is proplery started, that the logfile access is possible and that the configuration values the user specifies when starting the YARN session are visible in the web frontend.

I've tested the change on a cluster (Hadoop 2.4.1) and on the Hortonworks Sandbox 2.2 (Hadoop 2.6.0)

PR #8449, opened 2019-05-15
[FLINK-12235][hive] Support Hive partition in HiveCatalog

lirui-apache:

To implement partition related operations in `HiveCatalogBase`.

## Brief change log

  - Introduced `HiveCatalogPartition` to represent a partition that can be handled by `HiveCatalog`.
  - Implemented partition-related operations in `HiveCatalogBase`. Although we intend to let `HiveCatalog` and `GenericHiveMetastoreCatalog` share the implementations, this PR only enables/tests these operations for `HiveCatalog`.
 - Moved partition related tests from `GenericInMemoryCatalogTest` to `CatalogTestBase`, so that `GenericInMemoryCatalogTest` and `HiveCatalogTest` can share these test cases.

## Verifying this change

This PR is tested using partition related test cases in `CatalogTestBase`.

  - The serializers: no

## Documentation


lirui-apache:
@xuefuz @bowenli86 please take a look. Thanks.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalog.java
```diff
@@ -200,44 +200,56 @@ protected Table createHiveTable(ObjectPath tablePath, CatalogBaseTable table) {
...
+		List<String> partValues = getFullPartitionValues(new ObjectPath(hiveTable.getDbName(), hiveTable.getTableName()),
+			partitionSpec, partCols);
+		// validate partition values
+		for (int i = 0; i < partCols.size(); i++) {
+			if (StringUtils.isNullOrWhitespaceOnly(partValues.get(i))) {
+				throw new PartitionSpecInvalidException(catalogName, partCols,
+					new ObjectPath(hiveTable.getDbName(), hiveTable.getTableName()), partitionSpec);
+			}
+		}
+		HiveCatalogPartition hiveCatalogPartition = (HiveCatalogPartition) catalogPartition;
```
xuefuz: Do we need to validate the casting is valid?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
```diff
@@ -407,4 +443,201 @@ private Table getHiveTable(ObjectPath tablePath) throws TableNotExistException {
...
+						partitionSpec, tablePath), e);
+			} catch (TableNotPartitionedException | TableNotExistException | PartitionSpecInvalidException e) {
+				throw new PartitionNotExistException(catalogName, tablePath, partitionSpec, e);
+			}
+		} else if (!ignoreIfNotExists) {
+			throw new PartitionNotExistException(catalogName, tablePath, partitionSpec);
+		}
+	}
+
+	private void ensurePartitionedTable(ObjectPath tablePath, Table hiveTable) throws TableNotPartitionedException {
```
xuefuz: 1. can this be static?
2. Can this be combined with another method of the same name?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
```diff
@@ -407,4 +443,201 @@ private Table getHiveTable(ObjectPath tablePath) throws TableNotExistException {
...
+		}
+	}
+
+	@Override
+	public void dropPartition(ObjectPath tablePath, CatalogPartitionSpec partitionSpec, boolean ignoreIfNotExists)
+		throws PartitionNotExistException, CatalogException {
+		try {
+			Table hiveTable = getHiveTable(tablePath);
+			ensurePartitionedTable(tablePath, hiveTable);
+			client.dropPartition(tablePath.getDatabaseName(), tablePath.getObjectName(),
```
xuefuz: Can we skip line #490 - 491 and let client.dropPartition() to take care of exceptions? This may avoid additional trips to Hive metastore.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
```diff
@@ -407,4 +443,201 @@ private Table getHiveTable(ObjectPath tablePath) throws TableNotExistException {
...
+		throws TableNotExistException, TableNotPartitionedException, PartitionSpecInvalidException, PartitionAlreadyExistsException, CatalogException {
+
+		validateCatalogPartition(partition);
+
+		Table hiveTable = getHiveTable(tablePath);
+
+		ensurePartitionedTable(tablePath, hiveTable);
+
+		try {
+			client.add_partition(createHivePartition(hiveTable, partitionSpec, partition));
```
xuefuz: Wouldn't this line take care of exceptions that might happen in line #468, #470, and #472?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
```diff
@@ -407,4 +443,201 @@ private Table getHiveTable(ObjectPath tablePath) throws TableNotExistException {
...
+			throw new PartitionNotExistException(catalogName, tablePath, partitionSpec, e);
+		}
+	}
+
+	@Override
+	public List<CatalogPartitionSpec> listPartitions(ObjectPath tablePath)
+		throws TableNotExistException, TableNotPartitionedException, CatalogException {
+		Table hiveTable = getHiveTable(tablePath);
+
+		ensurePartitionedTable(tablePath, hiveTable);
```
xuefuz: Similar question as above.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
```diff
@@ -407,4 +443,201 @@ private Table getHiveTable(ObjectPath tablePath) throws TableNotExistException {
...
+	public List<CatalogPartitionSpec> listPartitions(ObjectPath tablePath, CatalogPartitionSpec partitionSpec)
+		throws TableNotExistException, TableNotPartitionedException, CatalogException {
+		Table hiveTable = getHiveTable(tablePath);
+
+		ensurePartitionedTable(tablePath, hiveTable);
+
+		try {
+			// partition spec can be partial
+			List<String> partialVals = MetaStoreUtils.getPvals(hiveTable.getPartitionKeys(), partitionSpec.getPartitionSpec());
+			return client.listPartitionNames(tablePath.getDatabaseName(), tablePath.getObjectName(), partialVals,
```
xuefuz: Similar question as above.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
```diff
@@ -407,4 +443,201 @@ private Table getHiveTable(ObjectPath tablePath) throws TableNotExistException {
...
+		}
+	}
+
+	@Override
+	public CatalogPartition getPartition(ObjectPath tablePath, CatalogPartitionSpec partitionSpec)
+		throws PartitionNotExistException, CatalogException {
+		try {
+			Table hiveTable = getHiveTable(tablePath);
+			ensurePartitionedTable(tablePath, hiveTable);
+			Partition hivePartition = client.getPartition(tablePath.getDatabaseName(), tablePath.getObjectName(),
```
xuefuz: Same.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
```diff
@@ -407,4 +443,201 @@ private Table getHiveTable(ObjectPath tablePath) throws TableNotExistException {
...
+	@Override
+	public void alterPartition(ObjectPath tablePath, CatalogPartitionSpec partitionSpec, CatalogPartition newPartition, boolean ignoreIfNotExists)
+		throws PartitionNotExistException, CatalogException {
+		// Explicitly check if the partition exists or not
+		// because alter_partition() doesn't throw NoSuchObjectException like dropPartition() when the target doesn't exist
+		if (partitionExists(tablePath, partitionSpec)) {
+			try {
+				Table hiveTable = getHiveTable(tablePath);
+				ensurePartitionedTable(tablePath, hiveTable);
+				client.alter_partition(
```
xuefuz: Same.

flink-connectors/flink-connector-hive/src/test/java/org/apache/flink/table/catalog/hive/HiveCatalogTest.java
```diff
@@ -154,4 +161,13 @@ protected void checkEquals(CatalogView v1, CatalogView v2) {
 		// thus properties of Hive view is a super set of those in its corresponding Flink view
 		assertTrue(v2.getProperties().entrySet().containsAll(v1.getProperties().entrySet()));
 	}
+
+	@Override
+	protected void checkEquals(CatalogPartition expected, CatalogPartition actual) {
+		HiveCatalogPartition hivePartition1 = (HiveCatalogPartition) expected;
```
xuefuz: We probably want to make sure the casting is valid before we do so. Better yet, we need to make sure the classes of the inputs are also equal.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/GenericInMemoryCatalog.java
```diff
@@ -412,31 +418,22 @@ public void createPartition(ObjectPath tablePath, CatalogPartitionSpec partition
...
-				tablePath, partitionSpec);
-		}
-
 		partitions.get(tablePath).put(partitionSpec, partition.copy());
 	}
 
 	@Override
 	public void dropPartition(ObjectPath tablePath, CatalogPartitionSpec partitionSpec, boolean ignoreIfNotExists)
-			throws PartitionNotExistException, CatalogException {
+		throws PartitionNotExistException, CatalogException {
```
xuefuz: Nit: Additional tab is needed before throws.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/GenericInMemoryCatalog.java
```diff
@@ -540,25 +536,18 @@ private boolean isPartitionSpecValid(ObjectPath tablePath, CatalogPartitionSpec
...
+		CatalogTable table = (CatalogTable) baseTable;
 		List<String> partitionKeys = table.getPartitionKeys();
 		Map<String, String> spec = partitionSpec.getPartitionSpec();
 
 		// The size of partition spec should not exceed the size of partition keys
-		if (partitionKeys.size() < spec.size()) {
-			return false;
-		} else {
-			int size = spec.size();
+		return partitionKeys.size() == spec.size() && spec.keySet().containsAll(partitionKeys);
```
xuefuz: This is different in semantics from the original code.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalog.java
```diff
@@ -200,44 +200,56 @@ protected Table createHiveTable(ObjectPath tablePath, CatalogBaseTable table) {
...
+		partition.setValues(partValues);
+		partition.setDbName(hiveTable.getDbName());
+		partition.setTableName(hiveTable.getTableName());
+		partition.setCreateTime((int) (System.currentTimeMillis() / 1000));
+		partition.setParameters(hiveCatalogPartition.getProperties());
+		partition.setSd(hiveTable.getSd().deepCopy());
+
+		String location = hiveCatalogPartition.getLocation();
+		if (null == location) {
```
bowenli86: what happens if we don't set the partition location? Won't hive metastore automatically run the following logic for us, like how it does for Table?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
```diff
@@ -146,6 +159,29 @@ protected abstract void validateCatalogBaseTable(CatalogBaseTable catalogBaseTab
...
+	/**
+	 * Creates a Hive Partition from a CatalogPartition.
+	 *
+	 * @param hiveTable        the Hive table to which the partition belongs.
+	 * @param catalogPartition the CatalogPartition from which the Hive partition will be created.
+	 * @param partitionSpec    the corresponding CatalogPartitionSpec
+	 * @return a Hive Partition
+	 */
+	protected abstract Partition createHivePartition(Table hiveTable, CatalogPartitionSpec partitionSpec,
+													CatalogPartition catalogPartition) throws PartitionSpecInvalidException;
```
bowenli86: nit: can we reformat it to be more readable? like:

```
protected abstract Partition createHivePartition(
	Table hiveTable,
	CatalogPartitionSpec partitionSpec,
	CatalogPartition catalogPartition) throws PartitionSpecInvalidException;
```

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
```diff
@@ -407,4 +443,201 @@ private Table getHiveTable(ObjectPath tablePath) throws TableNotExistException {
...
+		for (String keyVal : partKeyVals) {
+			String[] kv = keyVal.split("=");
+			spec.put(kv[0], kv[1]);
+		}
+		return new CatalogPartitionSpec(spec);
+	}
+
+	/**
+	 * Get a list of ordered partition values by re-arranging them based on the given list of partition keys.
+	 *
```
bowenli86: minor: missing a param for `tablePath`

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
```diff
@@ -407,4 +443,201 @@ private Table getHiveTable(ObjectPath tablePath) throws TableNotExistException {
...
+	 * @param partitionSpec a partition spec.
+	 * @param partitionKeys a list of partition keys.
+	 * @return A list of partition values ordered according to partitionKeys.
+	 * @throws PartitionSpecInvalidException thrown if partitionSpec and partitionKeys have different sizes,
+	 *                                       or any key in partitionKeys doesn't exist in partitionSpec.
+	 */
+	List<String> getFullPartitionValues(ObjectPath tablePath, CatalogPartitionSpec partitionSpec, List<String> partitionKeys)
```
bowenli86: what does the "full" mean here? probably rename to "getOrdered/ArrangedPartitionValues" to conform to its logic?

nit: can we move `tablePath` to the end as the last param given it's here to only build a message without any real effect.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/GenericInMemoryCatalog.java
```diff
@@ -412,31 +418,22 @@ public void createPartition(ObjectPath tablePath, CatalogPartitionSpec partition
 		checkNotNull(partitionSpec);
 		checkNotNull(partition);
 
-		if (!tableExists(tablePath)) {
```
bowenli86: I think it may not worth extract these to their own methods like `ensureTableExists` since it's such a simple logic and very clear to readers what happens.  When I read this code now, I have to find and check where the `ensureXxx` are and what they do.

flink-table/flink-table-common/src/test/java/org/apache/flink/table/catalog/CatalogTestBase.java
```diff
@@ -683,4 +1058,8 @@ protected void checkEquals(CatalogView v1, CatalogView v2) {
 		assertEquals(v1.getOriginalQuery(), v2.getOriginalQuery());
 		assertEquals(v1.getExpandedQuery(), v2.getExpandedQuery());
 	}
+
+	protected void checkEquals(CatalogPartition expected, CatalogPartition actual) {
```
bowenli86: we can just move `CatalogTestUtil.checkEquals(catalogpartitions)` to here and remove it from CatalogTestUtil

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
lirui-apache: Yes, but that means we need to parse the exceptions returned by HMS client to figure out the root cause. I think it's cleaner to check explicitly by ourselves.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
lirui-apache: Yeah I guess we can do this for `dropPartition`, because `dropPartition` doesn't require a precise exception like `createPartition`. So it's probably OK to wrap anything we get in a `PartitionNotExistException`

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
lirui-apache: Same as `createPartition`, this API requires more specific exception. Better not to rely on HMS client for it.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
lirui-apache: Same reason for `listPartitions(ObjectPath tablePath)`

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
lirui-apache: Will make the change for this one since it's similar to `dropPartition`.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
lirui-apache: Same as `dropPartition`. Will make the change

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
lirui-apache: It can be static as long as we pass all the needed (non-static) fields to it as parameters. The other method is in `GenericInMemoryCatalog`. Shall we have a class for different catalogs to share util methods?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
lirui-apache: nice catch :)

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
lirui-apache: It means a full spec, which contains values for all partition keys. And a partial spec can only contain values for a subset of partition keys. Operations like `createPartition`, `dropPartition` require a full spec. Operations like `listPartitions` can accept partial spec. Let me rename to to getOrderedFullPartitionValues, to indicate it orders the values and requires a full spec.

flink-connectors/flink-connector-hive/src/test/java/org/apache/flink/table/catalog/hive/HiveCatalogTest.java
lirui-apache: OK

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/GenericInMemoryCatalog.java
lirui-apache: It's intended to reduce code duplication. And makes things easier if we want to, say, change exception message in the future.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/GenericInMemoryCatalog.java
lirui-apache: Thanks for pointing out, will do

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/GenericInMemoryCatalog.java
lirui-apache: Yes, that's intended change

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalog.java
lirui-apache: Unfortunately, removing this will cause problems in `alterPartition`, because HMS won't generate path when altering a partition. Let's keep it as is, for now.

bowenli86: I see. I think we can address it by using location of old partition if the new partition doesn't have it.  We already retrieved old partition from HMS in `partitionExist()` (as a result we may need to refactor how alterPartition() checks if a partition exists or not). It's similar to how `alterTable()` works.

bowenli86: Another thought: the root cause of missing location seems to be that we assign the table's StorageDescriptor to the new partition rather than the old partition's, in line 230. Should we use the old partition's StorageDescriptor there?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
```diff
@@ -429,4 +467,199 @@ private Table getHiveTable(ObjectPath tablePath) throws TableNotExistException {
...
+	@Override
+	public void dropPartition(ObjectPath tablePath, CatalogPartitionSpec partitionSpec, boolean ignoreIfNotExists)
+		throws PartitionNotExistException, CatalogException {
+		try {
+			Table hiveTable = getHiveTable(tablePath);
+			client.dropPartition(tablePath.getDatabaseName(), tablePath.getObjectName(),
+				getOrderedFullPartitionValues(partitionSpec, getFieldNames(hiveTable.getPartitionKeys()), tablePath), true);
+		} catch (NoSuchObjectException e) {
+			if (!ignoreIfNotExists) {
+				throw new PartitionNotExistException(catalogName, tablePath, partitionSpec);
```
bowenli86: include ‘e’ in the  new PartitionNotExistException?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
```diff
@@ -429,4 +467,199 @@ private Table getHiveTable(ObjectPath tablePath) throws TableNotExistException {
...
+	@Override
+	public CatalogPartition getPartition(ObjectPath tablePath, CatalogPartitionSpec partitionSpec)
+		throws PartitionNotExistException, CatalogException {
+		try {
+			Table hiveTable = getHiveTable(tablePath);
+			Partition hivePartition = client.getPartition(tablePath.getDatabaseName(), tablePath.getObjectName(),
+				getOrderedFullPartitionValues(partitionSpec, getFieldNames(hiveTable.getPartitionKeys()), tablePath));
+			return createCatalogPartition(hivePartition);
+		} catch (NoSuchObjectException e) {
+			throw new PartitionNotExistException(catalogName, tablePath, partitionSpec);
```
bowenli86: include ‘e’ in the new PartitionNotExistException?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalog.java
lirui-apache: Makes sense. I'll have a try

lirui-apache:
@bowenli86 @xuefuz thanks for your comments. Please take a look at the updated PR. Thanks.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
xuefuz: I'd say making a method static as long as it doesn't need to access instance members. For utils, I don't see why not if there is a case for sharing.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
xuefuz: However, additional trips to HMS should be avoided if we can. I guess getHiveTable() requires round trip.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
xuefuz: getHiveTable() requires a round trip. Compared to precise exception msg, I'd say we minimize the number of trips.

lirui-apache: We need to get the hive table in order to get the partition columns (which in turn is used to get ordered partition values). So I don't think we can get rid of the trip to HMS

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
lirui-apache: Same as `dropPartition` -- we need the hive table to figure out partition columns.

bowenli86:
@xuefuz  I'd rather get https://github.com/apache/flink/pull/8480 in first and rebase this PR since 8480 is a much larger change set and not easy to rebase. Same as https://github.com/apache/flink/pull/8477

lirui-apache:
Rebased.
@xuefuz @bowenli86 please take another look. Thanks.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/GenericInMemoryCatalog.java
```diff
@@ -531,11 +520,18 @@ public boolean partitionExists(ObjectPath tablePath, CatalogPartitionSpec partit
 		return partitions.containsKey(tablePath) && partitions.get(tablePath).containsKey(partitionSpec);
 	}
 
+	private void ensureFullPartitionSpec(ObjectPath tablePath, CatalogPartitionSpec partitionSpec)
```
xuefuz: TableNotExistException doesn't seem being thrown anywhere from this method.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/GenericInMemoryCatalog.java
```diff
@@ -492,15 +484,12 @@ public void alterPartition(ObjectPath tablePath, CatalogPartitionSpec partitionS
 		checkNotNull(tablePath);
 		checkNotNull(partitionSpec);
 
-		if (!tableExists(tablePath)) {
```
xuefuz: Don't we need this check any more? The method signature specifies throwing TableNotExistException.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalog.java
```diff
@@ -607,44 +609,231 @@ private  static Table instantiateHiveTable(ObjectPath tablePath, CatalogBaseTabl
...
+		partition.setTableName(hiveTable.getTableName());
+		partition.setCreateTime((int) (System.currentTimeMillis() / 1000));
+		partition.setParameters(hiveCatalogPartition.getProperties());
+		partition.setSd(hiveTable.getSd().deepCopy());
+		partition.getSd().setLocation(hiveCatalogPartition.getLocation());
+
+		return partition;
+	}
+
+	private CatalogPartition createCatalogPartition(Partition hivePartition) {
```
xuefuz: Static?

In fact, many of the private util method might be candidates as static.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalog.java
```diff
@@ -607,44 +609,231 @@ private  static Table instantiateHiveTable(ObjectPath tablePath, CatalogBaseTabl
...
+	public void alterPartition(ObjectPath tablePath, CatalogPartitionSpec partitionSpec, CatalogPartition newPartition, boolean ignoreIfNotExists)
+		throws PartitionNotExistException, CatalogException {
+		// Explicitly check if the partition exists or not
+		// because alter_partition() doesn't throw NoSuchObjectException like dropPartition() when the target doesn't exist
+		if (partitionExists(tablePath, partitionSpec)) {
+			try {
+				Table hiveTable = getHiveTable(tablePath);
+				Partition newHivePartition = createHivePartition(hiveTable, partitionSpec, newPartition);
+				if (newHivePartition.getSd().getLocation() == null) {
+					Partition oldHivePartition = getHivePartition(tablePath, partitionSpec);
```
xuefuz: Minor: this piece of code makes several trips to HMS. In fact, partitionExists() already gets the partition, getHiveTable() makes another visit. I understand code is cleaner when being organized this way, but maybe we can optimize a bit toward performance.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalog.java
```diff
@@ -607,44 +609,231 @@ private  static Table instantiateHiveTable(ObjectPath tablePath, CatalogBaseTabl
...
 	public List<CatalogPartitionSpec> listPartitions(ObjectPath tablePath)
-			throws TableNotExistException, TableNotPartitionedException, CatalogException {
-		throw new UnsupportedOperationException();
+		throws TableNotExistException, TableNotPartitionedException, CatalogException {
+		Table hiveTable = getHiveTable(tablePath);
+
+		ensurePartitionedTable(tablePath, hiveTable);
+
+		try {
+			return client.listPartitionNames(tablePath.getDatabaseName(), tablePath.getObjectName(), (short) -1).stream()
```
bowenli86: nit: can we document what does the `-1` mean just for reference?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalog.java
```diff
@@ -607,44 +609,231 @@ private  static Table instantiateHiveTable(ObjectPath tablePath, CatalogBaseTabl
...
+				throw new CatalogException(
+					String.format("Failed to alter existing partition with new partition %s of table %s",
+						partitionSpec, tablePath), e);
+			}
+		} else if (!ignoreIfNotExists) {
+			throw new PartitionNotExistException(catalogName, tablePath, partitionSpec);
+		}
+	}
+
+	private Partition createHivePartition(Table hiveTable, CatalogPartitionSpec partitionSpec, CatalogPartition catalogPartition)
```
bowenli86: minor: after unification of the two catalogs, all such methods are renamed to `instantiateXxx()`. Shall we rename this method to be consistent with others?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalog.java
```diff
@@ -607,44 +609,231 @@ private  static Table instantiateHiveTable(ObjectPath tablePath, CatalogBaseTabl
...
+		throws PartitionSpecInvalidException {
```
bowenli86: nit: one more tab here

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalog.java
```diff
@@ -607,44 +609,231 @@ private  static Table instantiateHiveTable(ObjectPath tablePath, CatalogBaseTabl
 	// ------ partitions ------
 
 	@Override
-	public void createPartition(ObjectPath tablePath, CatalogPartitionSpec partitionSpec, CatalogPartition partition, boolean ignoreIfExists)
-			throws TableNotExistException, TableNotPartitionedException, PartitionSpecInvalidException, PartitionAlreadyExistsException, CatalogException {
-		throw new UnsupportedOperationException();
+	public boolean partitionExists(ObjectPath tablePath, CatalogPartitionSpec partitionSpec)
+		throws CatalogException {
```
bowenli86: nit: one more tab here. Same to all other methods below

flink-table/flink-table-common/src/test/java/org/apache/flink/table/catalog/CatalogTestBase.java
```diff
@@ -716,4 +1091,8 @@ protected void checkEquals(CatalogView v1, CatalogView v2) {
...
+		assertEquals(expected.getProperties(), actual.getProperties());
```
bowenli86: also check partition comment?

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/GenericInMemoryCatalog.java
lirui-apache: It's thrown from getTable(), but that doesn't seem optimal. I'll make isFullPartitionSpec throw TableNotExistException.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalog.java
lirui-apache: We need to contact HMS at least three times, i.e. get table, get existing partition, alter partition. I'll make sure no extra trips to HMS than these three.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalog.java
lirui-apache: OK

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/GenericInMemoryCatalog.java
lirui-apache: We don't have to. Because the following call to getTable will take care of that.

flink-table/flink-table-common/src/test/java/org/apache/flink/table/catalog/CatalogTestBase.java
lirui-apache: CatalogPartition doesn't have comment. I'll check descriptions instead.

bowenli86: oh, then probably don't need to do so, we don't check descriptions so far yet

lirui-apache:
Rebased and addressed comments.
@bowenli86 @xuefuz please take another look. Thanks.

lirui-apache:
> @lirui-apache Thank you very much for the update!
> 
> I just spotted that, currently how several APIs impl work like this: 1) get a raw hive table 2) parse part of the raw table. The latter step actually duplicate with logic in `instantiateHiveCatalogTable()`. E.g. `ensureTableAndPartitionMatch()` parses `FLINK_PROPERTY_IS_GENERIC`, `instantiateHivePartition()` parses partition keys, `ensurePartitionedTable()` parses the raw table's partition key size, all of which we can get by just parsing the raw table to a `CatalogTable` thru `instantiateHiveCatalogTable()` in advance. The current duplication also means if we change some general logic in parsing a hive table, we need to change two places. Thus I wonder if it makes sense to just parse the raw table as whole at the beginning rather than having scattered places each parsing only part of it themselves. And we can remove util methods such as `getFieldNames()` which is only used to get the partition keys which is already available in `CatalogTable`.
> 
> For example, change
> 
> ```
> public void createPartition(...) {
>   Table hiveTable = getHiveTable(tablePath);
>   ensureTableAndPartitionMatch(hiveTable, partition);
>   ensurePartitionedTable(tablePath, hiveTable);
>   try {
>     client.add_partition(instantiateHivePartition(hiveTable, partitionSpec, partition));
>   } ...
> }
> ```
> 
> to something like:
> 
> ```
> public void createPartition(...) {
>   Table hiveTable = getHiveTable(tablePath);
>   CatalogBaseTable catalogTable = instantiateHiveCatalogTable(hiveTable);
>   ... check whether catalogTabe and catalogPartition type matches would be much easier here ...
>   ... check whether catalogTable is partitioned would be easier here ...
>   try {
>     client.add_partition(
>       instantiateHivePartition(catalogTable, partitionSpec, partition, hiveTable.getSd()));
>   } ...
> }
> ```

I'm not sure how much benefit this can bring us. It might make `ensureTableAndPartitionMatch` a little easier -- we can check the type of CatalogBaseTable instead of parsing a property. But I don't think we should take the same approach for `ensurePartitionedTable`. `ensurePartitionedTable` is already simple enough. And for APIs like `listPartitions`, creating a CatalogBaseTable just to get num of partition cols seems an overkill to me. And the same applies to `getFieldNames`. E.g. I think it'll be an overkill to create a CatalogBaseTable to get partition col names in `dropPartition`.
Maybe an alternative approach to the problem you mentioned is to have more util methods, in order to avoid duplication. For example, we should have a util method to decide whether a Hive table is generic. And all the APIs needing this logic can call the util method. What do you think?

bowenli86:
ok, let's leave that part as it for now

lirui-apache:
Rebase and address comments.
@bowenli86 @xuefuz let me know if I missed anything. Thanks.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalog.java
```diff
@@ -608,44 +611,283 @@ private  static Table instantiateHiveTable(ObjectPath tablePath, CatalogBaseTabl
...
 
 	@Override
-	public void createPartition(ObjectPath tablePath, CatalogPartitionSpec partitionSpec, CatalogPartition partition, boolean ignoreIfExists)
-			throws TableNotExistException, TableNotPartitionedException, PartitionSpecInvalidException, PartitionAlreadyExistsException, CatalogException {
-		throw new UnsupportedOperationException();
+	public boolean partitionExists(ObjectPath tablePath, CatalogPartitionSpec partitionSpec)
+			throws CatalogException {
+		checkNotNull(tablePath, "Table path cannot be null");
+		checkNotNull(partitionSpec, "CatalogPartitionSpec cannot be null");
+		try {
```
xuefuz: Nit: keeping an empty line after checks would be good.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalog.java
```diff
@@ -608,44 +611,283 @@ private  static Table instantiateHiveTable(ObjectPath tablePath, CatalogBaseTabl
...
 	}
 
 	@Override
-	public void alterPartition(ObjectPath tablePath, CatalogPartitionSpec partitionSpec, CatalogPartition newPartition, boolean ignoreIfNotExists)
+	public void dropPartition(ObjectPath tablePath, CatalogPartitionSpec partitionSpec, boolean ignoreIfNotExists)
 			throws PartitionNotExistException, CatalogException {
-		throw new UnsupportedOperationException();
+		checkNotNull(tablePath, "Table path cannot be null");
+		checkNotNull(partitionSpec, "CatalogPartitionSpec cannot be null");
+		try {
```
xuefuz: Same as above.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalog.java
```diff
@@ -608,44 +611,283 @@ private  static Table instantiateHiveTable(ObjectPath tablePath, CatalogBaseTabl
...
+			throw new PartitionNotExistException(catalogName, tablePath, partitionSpec, e);
+		} catch (TException e) {
+			throw new CatalogException(
+				String.format("Failed to alter existing partition with new partition %s of table %s",
+					partitionSpec, tablePath), e);
+		}
+	}
+
+	// make sure both table and partition are generic, or neither is
+	private void ensureTableAndPartitionMatch(Table hiveTable, CatalogPartition catalogPartition) {
```
xuefuz: static?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalog.java
```diff
@@ -608,44 +611,283 @@ private  static Table instantiateHiveTable(ObjectPath tablePath, CatalogBaseTabl
...
+		boolean isGeneric = Boolean.valueOf(hiveTable.getParameters().get(FLINK_PROPERTY_IS_GENERIC));
+		if ((isGeneric && catalogPartition instanceof HiveCatalogPartition) ||
+			(!isGeneric && catalogPartition instanceof GenericCatalogPartition)) {
+			throw new CatalogException(String.format("Cannot handle %s partition for %s table",
+				catalogPartition.getClass().getName(), isGeneric ? "generic" : "non-generic"));
+		}
+	}
+
+	private Partition instantiateHivePartition(Table hiveTable, CatalogPartitionSpec partitionSpec, CatalogPartition catalogPartition)
```
xuefuz: static?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalog.java
```diff
@@ -608,44 +611,283 @@ private  static Table instantiateHiveTable(ObjectPath tablePath, CatalogBaseTabl
...
+
+		return partition;
+	}
+
+	private static CatalogPartition instantiateCatalogPartition(Partition hivePartition) {
+		// TODO: create GenericCatalogPartition for GenericCatalogTable
+		return new HiveCatalogPartition(hivePartition.getParameters(), hivePartition.getSd().getLocation());
+	}
+
+	private void ensurePartitionedTable(ObjectPath tablePath, Table hiveTable) throws TableNotPartitionedException {
```
xuefuz: static

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalog.java
```diff
@@ -608,44 +611,283 @@ private  static Table instantiateHiveTable(ObjectPath tablePath, CatalogBaseTabl
...
+	 * Get a list of ordered partition values by re-arranging them based on the given list of partition keys.
+	 *
+	 * @param partitionSpec a partition spec.
+	 * @param partitionKeys a list of partition keys.
+	 * @param tablePath path of the table to which the partition belongs.
+	 * @return A list of partition values ordered according to partitionKeys.
+	 * @throws PartitionSpecInvalidException thrown if partitionSpec and partitionKeys have different sizes,
+	 *                                       or any key in partitionKeys doesn't exist in partitionSpec.
+	 */
+	private List<String> getOrderedFullPartitionValues(CatalogPartitionSpec partitionSpec, List<String> partitionKeys, ObjectPath tablePath)
```
xuefuz: static?

lirui-apache:
Made `ensureTableAndPartitionMatch` static. Several other methods can't be static, unless we pass the needed non-static field as parameter. @xuefuz shall we do that?

xuefuz:
> Made `ensureTableAndPartitionMatch` static. Several other methods can't be static, unless we pass the needed non-static field as parameter. @xuefuz shall we do that?

No. I think we only need to make what is static in nature  static in their current forms.

bowenli86:
Thanks @lirui-apache very much for the PR. LGTM, merging

PR #8453, opened 2019-05-15
[FLINK-12254][table-common] Preparation for using the new type system

twalthr:
## Brief change log

This PR is a preparation for FLINK-12254. It fixes a couple of shortcomings when working with the new type system and adds the first utility classes that might useful for other contributions as well.

## Verifying this change

See commit messages.

  - The serializers: no

## Documentation

### DIFF
--- a/flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/DataTypeVisitor.java
+++ b/flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/DataTypeVisitor.java
@@ -1,0 +1,37 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.table.types;
+
+import org.apache.flink.annotation.PublicEvolving;
+
+/**
+ * The visitor definition of {@link DataType}. The visitor transforms a data type into
+ * instances of {@code R}.
+ */
+@PublicEvolving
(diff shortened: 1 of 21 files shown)


flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/logical/VarCharType.java
```diff
@@ -86,6 +89,9 @@ public LogicalType copy(boolean isNullable) {
 
 	@Override
 	public String asSerializableString() {
+		if (length == MAX_LENGTH) {
```
dawidwys: Could we have that only in the `summaryString`? Personally I'd prefer to serialize only the actual types and not the shortcuts.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/utils/ClassDataTypeConverter.java
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	/**
+	 * Returns the clearly identifiable data type if possible. For example, {@link Long} can be
+	 * expressed as {@link DataTypes#BIGINT()}. However, for example, {@link Row} cannot be extracted
+	 * as information about the fields is missing. Or {@link BigDecimal} needs to be mapped from a
+	 * variable precision/scale to constant ones.
+	 */
+	public static Optional<DataType> extractDataType(Class<?> clazz) {
+		// byte arrays have higher priority that regular arrays
```
dawidwys:
```suggestion
		// byte arrays have higher priority than regular arrays
```

flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/DataTypeVisitor.java
```diff
@@ -0,0 +1,37 @@
...
+import org.apache.flink.annotation.PublicEvolving;
+
+/**
+ * The visitor definition of {@link DataType}. The visitor transforms a data type into
+ * instances of {@code R}.
+ */
+@PublicEvolving
+public interface DataTypeVisitor<R> {
+
+	R visit(AtomicDataType atomicDataType);
```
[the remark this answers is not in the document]
twalthr: Makes sense to me. Will update the visitor methods.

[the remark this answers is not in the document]
twalthr: We also not do it for the LogicalTypeVisitor and it's also redundant. Let's keep it as it is.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/logical/VarCharType.java
twalthr: I was also unsure about this change. Summary strings should be sufficient, I agree.

PR #8457, opened 2019-05-15
[FLINK-12511][table][hive] make variable 'comment' in all catalog metadata classes final

bowenli86:

Because of historical reasons, currently the variable "comment" in all catalog metadata classes are not final yet, it has a default value and can be overwritten in constructor. It creates problems like overloaded constructors are built in a weird, if not wrong, way. (See code below).

This PR removes the default value of "comment" and make it final that can only be assigned value upon construction, and ends the not-so-well-designed overloaded constructors for catalog metadata classes.

## Brief change log

- made 'comment' in catalog metadata classes and removed its default value
- rebuilt constructors for catalog metadata classes

## Verifying this change

This change is already covered by existing tests, such as *CatalogTestBase*.

  - The serializers: (no)

## Documentation


bowenli86:
merging...

PR #8458, opened 2019-05-15
[FLINK-12452][table][hive] alterTable() should ensure existing base table and the new one are of the same type

bowenli86:

Currently all catalogs doesn't check if existing base table and the new one are of the same type in alterTable(), e.g. existing table is a view but user tries to replace it with a table. This PR adds such checks to all catalogs

## Brief change log

- added check in in-memory catalog to make sure existing and new table are of the same classes since in-memory catalog theoretically can store any table and view implementations 
- added checks in `HiveCatalogBase` to make sure 1) existing and new table are of the same type 2) not alter existing Hive tables in types that Flink cannot create for now, including EXTERNAL_TABLE, INDEX_TABLE and MATERIALZED_VIEW. Flink of course cannot alter existing ones if it cannot create them.

Note that though Flink cannot create them, it's able to read most parts of their metadata, like schema and properties, thru existing mechanism, but not their unique metadata. 

Also note that add capability of creating EXTERNAL_TABLE should be fairly easy to do, e.g. with a table property "table_type=external", we just haven't done it yet.  INDEX_TABLE and MATERIALZED_VIEW need more research.

## Verifying this change

Added new unit tests in `GenericInMemoryCatalogTest` and `CatalogTestBase`.

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
+++ b/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
@@ -319,6 +319,7 @@ public abstract class HiveCatalogBase implements Catalog {
 				if (!ignoreIfNotExists) {
 					throw new TableNotExistException(catalogName, tablePath);
 				}
+<<<<<<< HEAD
 				return;
 			}
 
--- a/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
+++ b/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
@@ -325,12 +325,7 @@ public abstract class HiveCatalogBase implements Catalog {
 			Table oldTable = getHiveTable(tablePath);
 			TableType oldTableType = TableType.valueOf(oldTable.getTableType());
 
-			if (oldTableType == TableType.EXTERNAL_TABLE
-					|| oldTableType == TableType.INDEX_TABLE
-					|| oldTableType == TableType.MATERIALIZED_VIEW) {
-				throw new CatalogException(
-					String.format("The existing Hive table is of type '%s', and HiveCatalogBase cannot handle it", oldTableType.name()));
-			} else if (oldTableType == TableType.VIRTUAL_VIEW) {
+			if (oldTableType == TableType.VIRTUAL_VIEW) {
 				if (!(newCatalogTable instanceof CatalogView)) {
 					throw new CatalogException(
 						String.format("Table types don't match. The existing table is a view, but the new catalog base table is not."));
@@ -344,7 +339,8 @@ public abstract class HiveCatalogBase implements Catalog {
 				// Else, do nothing
 			} else {
 				throw new CatalogException(
-					String.format("The existing table is a table, but the new catalog base table is not."));
+					String.format("Hive table type '%s' is not supported yet.",
+						oldTableType.name()));
 			}
 
 			Table newTable = createHiveTable(tablePath, newCatalogTable);
--- a/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
+++ b/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
@@ -319,7 +319,6 @@ public abstract class HiveCatalogBase implements Catalog {
 				if (!ignoreIfNotExists) {
 					throw new TableNotExistException(catalogName, tablePath);
 				}
-<<<<<<< HEAD
 				return;
 			}
 
@@ -334,18 +333,23 @@ public abstract class HiveCatalogBase implements Catalog {
 			} else if (oldTableType == TableType.VIRTUAL_VIEW) {
 				if (!(newCatalogTable instanceof CatalogView)) {
 					throw new CatalogException(
-						String.format("The existing table is a view, but the new catalog base table is not."));
+						String.format("HiveCatalogBase does not support type '%s' of existing Hive table yet.", oldTableType.name()));
+				}
+			} else if (oldTableType == TableType.VIRTUAL_VIEW) {
+				if (!(newCatalogTable instanceof CatalogView)) {
+					throw new CatalogException(
+						String.format("Table types don't match. The existing table is a view, but the new catalog base table is not."));
 				}
 				// Else, do nothing
 			} else if ((oldTableType == TableType.MANAGED_TABLE)) {
 				if (!(newCatalogTable instanceof CatalogTable)) {
 					throw new CatalogException(
-						String.format("The existing table is a table, but the new catalog base table is not."));
+						String.format("Table types don't match. The existing table is a table, but the new catalog base table is not."));
 				}
 				// Else, do nothing
 			} else {
 				throw new CatalogException(
-					String.format("HiveCatalogBase does not recognize Hive table type '%s'", oldTableType.name()));
+					String.format("The existing table is a table, but the new catalog base table is not."));
 			}
 
 			Table newTable = createHiveTable(tablePath, newCatalogTable);
--- a/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
+++ b/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
@@ -21,6 +21,8 @@ package org.apache.flink.table.catalog.hive;
 import org.apache.flink.table.catalog.Catalog;
 import org.apache.flink.table.catalog.CatalogBaseTable;
 import org.apache.flink.table.catalog.CatalogDatabase;
+import org.apache.flink.table.catalog.CatalogTable;
+import org.apache.flink.table.catalog.CatalogView;
 import org.apache.flink.table.catalog.ObjectPath;
 import org.apache.flink.table.catalog.exceptions.CatalogException;
 import org.apache.flink.table.catalog.exceptions.DatabaseAlreadyExistException;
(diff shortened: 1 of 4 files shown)


flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
```diff
@@ -318,13 +320,44 @@ public void alterTable(ObjectPath tablePath, CatalogBaseTable newCatalogTable, b
 					throw new TableNotExistException(catalogName, tablePath);
 				}
 			} else {
```
xuefuz: Code will be easier to read if we add return in the if branch and remove else branch.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
```diff
@@ -318,13 +320,44 @@ public void alterTable(ObjectPath tablePath, CatalogBaseTable newCatalogTable, b
...
+
+				// Flink cannot alter existing Hive tables in types that Flink cannot create for now,
+				// including EXTERNAL_TABLE, INDEX_TABLE and MATERIALZED_VIEW.
+				// Note that though Flink cannot create them, it's able to read most parts of their metadata,
+				// like schema and properties, thru existing mechanism, but not their unique metadata.
+				// Also note that add capability of creating EXTERNAL_TABLE should be fairly easy to do, e.g. with a table property
+				// "table_type=external", we just haven't done it yet.
+				// INDEX_TABLE and MATERIALZED_VIEW need more research.
+
+				if (oldTableType == TableType.EXTERNAL_TABLE
```
xuefuz: Instead of listing types that we don't support, it's probably better to check if the type is what we do support.

Also event though Flink is not able to create external table yet, that's probably due to the limitation of the DLL and I don't see why we should disallow user to modify it.

It seems okay if we say we don't support other types yet.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
```diff
@@ -318,13 +320,44 @@ public void alterTable(ObjectPath tablePath, CatalogBaseTable newCatalogTable, b
...
+						|| oldTableType == TableType.INDEX_TABLE
+						|| oldTableType == TableType.MATERIALIZED_VIEW) {
+					throw new CatalogException(
+						String.format("The existing Hive table is of type '%s', and HiveCatalogBase cannot handle it", oldTableType.name()));
```
xuefuz: Instead of saying "HiveCatalogBase cannot handle it", it's probably better to say "xxx type is currently not supported".

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
```diff
@@ -318,13 +320,44 @@ public void alterTable(ObjectPath tablePath, CatalogBaseTable newCatalogTable, b
...
+				} else if (oldTableType == TableType.VIRTUAL_VIEW) {
+					if (!(newCatalogTable instanceof CatalogView)) {
+						throw new CatalogException(
+							String.format("The existing table is a view, but the new catalog base table is not."));
+					}
+					// Else, do nothing
+				} else if ((oldTableType == TableType.MANAGED_TABLE)) {
+					if (!(newCatalogTable instanceof CatalogTable)) {
+						throw new CatalogException(
+							String.format("The existing table is a table, but the new catalog base table is not."));
```
xuefuz: Nit: maybe start the message with "Table type doesn't match. ...."

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/GenericInMemoryCatalog.java
```diff
@@ -209,11 +209,17 @@ public void alterTable(ObjectPath tablePath, CatalogBaseTable newTable, boolean
...
 
-		// TODO: validate the new and old CatalogBaseTable must be of the same type. For example, this doesn't
-		//		allow alter a regular table to partitioned table, or alter a view to a table, and vice versa.
-		//		And also add unit tests.
-
 		if (tableExists(tablePath)) {
+			CatalogBaseTable oldTable = tables.get(tablePath);
+
+			// Theoretically, in-memory catalog can store any tables and views implementations
+			// thus, we need to check they are of the same class
```
xuefuz: Nit: The comment doesn't seem explaining well why we want to put the check. Instead of "thus ...", we can probably say ". However, to be consistent with other catalogs, we check if...".

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/GenericInMemoryCatalog.java
```diff
@@ -209,11 +209,17 @@ public void alterTable(ObjectPath tablePath, CatalogBaseTable newTable, boolean
...
+			if (oldTable.getClass() != newTable.getClass()) {
+				throw new CatalogException(
+					String.format("Existing table is '%s' and new table is '%s'. They should be of the same class.",
```
xuefuz: Same as above. We can probably say "Table types doesn't match. ..."

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
bowenli86: Actually I listed both here just to make it more clear on what we do and do not support right now.

Yeah, as mentioned in the comment, external table is not supported only because we haven't done it yet, not because DDL limitation. Enabling external table will be on its own PR.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/GenericInMemoryCatalog.java
bowenli86: there's a subtle difference between checks in in-memory catalog and HiveCatalogBase.

HiveCatalogBase can only handle HiveCatalogTable/View (HiveCatalog) or GenericCatalogTable/View (GenericHiveMetastoreCatalog), and user input is already validated in validateCatalogBaseTable(). Thus, we only need to check if the input is CatalogTable or CatalogView.

However, in-memory catalog can host any implementations of table and view, thus we need to check here that they are of the same impl class, not just whether it's a table or view.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/GenericInMemoryCatalog.java
bowenli86: see above

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
bowenli86: agree. Shall we do it separately since it's orthogonal to what this PR is for? I will refactor that in a [hotfix] PR

bowenli86:
@xuefuz thanks for your review, can you take another look?

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/GenericInMemoryCatalog.java
xuefuz: I guess you didn't get what I meant. I was saying the comment doesn't seem making much sense. It doesn't explain why we put checks  and why we are mentioning in-memory catalog here. There is a "thus", but I don't understand the cause/result relationship that you're trying to explain.

bowenli86: I removed the comment as the code actually explains itself

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
xuefuz: 1. Saying in the comment is fine. I meant when we check in the code, it's better to check against something we support rather than checking against something we don't . The difference is that there might be additional things coming in the future that we don't support. We can avoid code changes if we don't plan to support it.

2. I understand external table cannot be created in Flink. The reason is the limitation of DDL. However, altering an external table should already work. I don't see why we need to exclude it now. Let me know if my judgement is inaccurate.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
xuefuz: It's probably better to fix it now while it's still warm.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
bowenli86: 1. I think the current code is better since it checks what we've supported, what we plan to support, what we don't know (the "additional things coming in the future" you mentioned)
2. we don't support it yet. external table has additional properties we need to parse and check

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogBase.java
bowenli86: Ok. I fixed in source repo and rebased this PR to master, and the rebase took quite some time even though it's a small change.

In the future I think we should follow [Apache Flink Code Style and Quality Guide](https://docs.google.com/document/d/1JGRVK6udBFa8xbyv7ipPQiLEhxebA64f8KKcaAH0UPY/edit#heading=h.i11eha8b46sj). In section ""Separate Refactoring, Cleanup and Independent Changes" it clarifies that such change should be on its own.

bowenli86:
thanks for your review @xuefuz , can you take another look?

PR #8459, opened 2019-05-16
[FLINK-12476] [State TTL] Consider setting a default background cleanup strategy in StateTtlConfig

yanghua:

*This pull request sets a default background cleanup strategy in StateTtlConfig*

## Brief change log

  - *Added a general config method `cleanupInBackground`*
  - *Changed document*

## Verifying this change

## Documentation

  - If yes, how is the feature documented? (not applicable / **docs** / **JavaDocs** / not documented)

### DIFF
--- a/docs/dev/stream/state/state.md
+++ b/docs/dev/stream/state/state.md
@@ -400,6 +400,35 @@ This option is not applicable for the incremental checkpointing in the RocksDB s
 - For existing jobs, this cleanup strategy can be activated or deactivated anytime in `StateTtlConfig`, 
 e.g. after restart from savepoint.
 
+#### Cleanup in background
+
+Besides cleanup in full snapshot, you can also activate the cleanup in background. The following option 
+will activate a default background cleanup in StateTtlConfig if it is supported for the used backend:
+
+<div class="codetabs" markdown="1">
+<div data-lang="java" markdown="1">
+{% highlight java %}
+import org.apache.flink.api.common.state.StateTtlConfig;
+StateTtlConfig ttlConfig = StateTtlConfig
+    .newBuilder(Time.seconds(1))
+    .cleanupInBackground()
+    .build();
+{% endhighlight %}
+</div>
+ <div data-lang="scala" markdown="1">
+{% highlight scala %}
+import org.apache.flink.api.common.state.StateTtlConfig
+val ttlConfig = StateTtlConfig
+    .newBuilder(Time.seconds(1))
+    .cleanupInBackground
+    .build
+{% endhighlight %}
+</div>
+</div>
+
+For more fine-grained control over some special cleanup in background, you can configure it separately as described below.
+Currently, heap state backend relies on incremental cleanup and RocksDB backend uses compaction filter for background cleanup.
+
 ##### Incremental cleanup
 
 Another option is to trigger cleanup of some state entries incrementally.
@@ -435,6 +464,7 @@ val ttlConfig = StateTtlConfig
 This strategy has two parameters. The first one is number of checked state entries per each cleanup triggering.
 If enabled, it is always triggered per each state access.
 The second parameter defines whether to trigger cleanup additionally per each record processing.
+If you enable the default background cleanup then this strategy will be activated for heap backend with 5 checked entries and without cleanup per record processing.
 
 **Notes:**
 - If no access happens to the state or no records are processed, expired state will persist.
@@ -465,7 +495,7 @@ import org.apache.flink.api.common.state.StateTtlConfig;
 
 StateTtlConfig ttlConfig = StateTtlConfig
     .newBuilder(Time.seconds(1))
-    .cleanupInRocksdbCompactFilter()
+    .cleanupInRocksdbCompactFilter(1000)
     .build();
 {% endhighlight %}
 </div>
@@ -476,18 +506,19 @@ import org.apache.flink.api.common.state.StateTtlConfig
 
 val ttlConfig = StateTtlConfig
     .newBuilder(Time.seconds(1))
-    .cleanupInRocksdbCompactFilter
+    .cleanupInRocksdbCompactFilter(1000)
     .build
 {% endhighlight %}
 </div>
 </div>
 
 RocksDB compaction filter will query current timestamp, used to check expiration, from Flink every time 
-after processing certain number of state entries. This number is 1000 by default. 
-You can optionally change it and pass a custom value to 
+after processing certain number of state entries.
+You can change it and pass a custom value to 
 `StateTtlConfig.newBuilder(...).cleanupInRocksdbCompactFilter(long queryTimeAfterNumEntries)` method. 
 Updating the timestamp more often can improve cleanup speed 
 but it decreases compaction performance because it uses JNI call from native code.
+If you enable the default background cleanup then this strategy will be activated for RocksDB backend and the current timestamp will be queried each time 1000 entries have been processed.
 
 You can activate debug logs from the native code of RocksDB filter 
 by activating debug level for `FlinkCompactionFilter`:
--- a/flink-core/src/main/java/org/apache/flink/api/common/state/StateTtlConfig.java
+++ b/flink-core/src/main/java/org/apache/flink/api/common/state/StateTtlConfig.java
@@ -29,6 +29,9 @@ import javax.annotation.Nullable;
 import java.io.Serializable;
 import java.util.EnumMap;
 
+import static org.apache.flink.api.common.state.StateTtlConfig.CleanupStrategies.EMPTY_STRATEGY;
+import static org.apache.flink.api.common.state.StateTtlConfig.IncrementalCleanupStrategy.DEFAULT_INCREMENTAL_CLEANUP_STRATEGY;
+import static org.apache.flink.api.common.state.StateTtlConfig.RocksdbCompactFilterCleanupStrategy.DEFAULT_ROCKSDB_COMPACT_FILTER_CLEANUP_STRATEGY;
 import static org.apache.flink.api.common.state.StateTtlConfig.StateVisibility.NeverReturnExpired;
 import static org.apache.flink.api.common.state.StateTtlConfig.TtlTimeCharacteristic.ProcessingTime;
 import static org.apache.flink.api.common.state.StateTtlConfig.UpdateType.OnCreateAndWrite;
@@ -166,7 +169,9 @@ public class StateTtlConfig implements Serializable {
 		private StateVisibility stateVisibility = NeverReturnExpired;
 		private TtlTimeCharacteristic ttlTimeCharacteristic = ProcessingTime;
 		private Time ttl;
-		private CleanupStrategies cleanupStrategies = new CleanupStrategies();
+		private boolean isCleanupInBackground = false;
+		private final EnumMap<CleanupStrategies.Strategies, CleanupStrategies.CleanupStrategy> strategies =
+			new EnumMap<>(CleanupStrategies.Strategies.class);
 
 		public Builder(@Nonnull Time ttl) {
 			this.ttl = ttl;
@@ -249,7 +254,7 @@ public class StateTtlConfig implements Serializable {
 		/** Cleanup expired state in full snapshot on checkpoint. */
 		@Nonnull
 		public Builder cleanupFullSnapshot() {
-			cleanupStrategies.activate(CleanupStrategies.Strategies.FULL_STATE_SCAN_SNAPSHOT);
+			strategies.put(CleanupStrategies.Strategies.FULL_STATE_SCAN_SNAPSHOT, EMPTY_STRATEGY);
 			return this;
 		}
 
@@ -284,7 +289,7 @@ public class StateTtlConfig implements Serializable {
 		public Builder cleanupIncrementally(
 			@Nonnegative int cleanupSize,
 			boolean runCleanupForEveryRecord) {
-			cleanupStrategies.activate(
+			strategies.put(
 				CleanupStrategies.Strategies.INCREMENTAL_CLEANUP,
 				new IncrementalCleanupStrategy(cleanupSize, runCleanupForEveryRecord));
 			return this;
@@ -304,8 +309,10 @@ public class StateTtlConfig implements Serializable {
 		 * https://github.com/facebook/rocksdb/blob/master/include/rocksdb/compaction_filter.h#L69
 		 * It means that the TTL filter should be tested for List state taking into account this caveat.
 		 *
+		 * @deprecated Use more general configuration method {@link #cleanupInBackground()} instead
 		 */
 		@Nonnull
+		@Deprecated
 		public Builder cleanupInRocksdbCompactFilter() {
 			return cleanupInRocksdbCompactFilter(1000L);
 		}
@@ -322,8 +329,20 @@ public class StateTtlConfig implements Serializable {
 		 */
 		@Nonnull
 		public Builder cleanupInRocksdbCompactFilter(long queryTimeAfterNumEntries) {
-			cleanupStrategies.activate(CleanupStrategies.Strategies.ROCKSDB_COMPACTION_FILTER,
+			strategies.put(
+				CleanupStrategies.Strategies.ROCKSDB_COMPACTION_FILTER,
 				new RocksdbCompactFilterCleanupStrategy(queryTimeAfterNumEntries));
+			return this;
+		}
+
+		/**
+		 * Enable cleanup of expired state in background.
+		 *
+		 * <p>Depending on actually used backend, the corresponding cleanup will kick in if supported.
+		 */
+		@Nonnull
+		public Builder cleanupInBackground() {
+			isCleanupInBackground = true;
 			return this;
 		}
 
@@ -344,7 +363,7 @@ public class StateTtlConfig implements Serializable {
 				stateVisibility,
 				ttlTimeCharacteristic,
 				ttl,
-				cleanupStrategies);
+				new CleanupStrategies(strategies, isCleanupInBackground));
 		}
 	}
 
@@ -358,7 +377,11 @@ public class StateTtlConfig implements Serializable {
 	public static class CleanupStrategies implements Serializable {
 		private static final long serialVersionUID = -1617740467277313524L;
 
-		private static final CleanupStrategy EMPTY_STRATEGY = new EmptyCleanupStrategy();
+		static final CleanupStrategy EMPTY_STRATEGY = new EmptyCleanupStrategy();
+
+		private final boolean isCleanupInBackground;
+
+		private final EnumMap<Strategies, CleanupStrategy> strategies;
 
 		/** Fixed strategies ordinals in {@code strategies} config field. */
 		enum Strategies {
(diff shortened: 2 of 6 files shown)


yanghua:
@azagrebin Can you review this PR?

flink-core/src/main/java/org/apache/flink/api/common/state/StateTtlConfig.java
```diff
@@ -327,6 +329,14 @@ public Builder cleanupInRocksdbCompactFilter(long queryTimeAfterNumEntries) {
 			return this;
 		}
 
+		/**
+		 * Cleanup expired state by activating both strategies. Depending on actually used backend, the corresponding cleanup will kick in.
+		 */
+		@Nonnull
+		public Builder cleanupInBackground() {
+			return cleanupIncrementally(10, true).cleanupInRocksdbCompactFilter(1000L);
```
aljoscha: I don't think the setting should delegate to the specific options. This will mean that this code will have to be adapted in the future if other backends or other cleanup strategies are added.

The state backends should notice that "cleanup in background" is enabled and then react accordingly.

yanghua: @aljoscha so you mean we do not need this `cleanupInBackground` and stay the same?

azagrebin: @aljoscha true, we can move concrete decision about default configuration to backends.

@yanghua the method `cleanupInBackground` should be  still added but rather as a boolean flag in StateTtlConfig. Backends should check this flag where they check configuration of their corresponding cleanup strategies and use default ones.

aljoscha: exactly! 👌

yanghua: Hi @aljoscha and @azagrebin I have refactored the code and give a **non-final** version of PR for discussing. If we let specific statebackend react for the special strategy, we may change some class's access modifier. WDYT?

flink-state-backends/flink-statebackend-rocksdb/src/main/java/org/apache/flink/contrib/streaming/state/ttl/RocksDbTtlCompactFiltersManager.java
```diff
@@ -108,7 +110,16 @@ public void configCompactFilter(
...
+
+		//if ttl is enable and cleanup in background and user did not specify the cleanup strategy
+		//then build the compact filter cleanup strategy and apply the default value
+		if (ttlConfig.isEnabled() && ttlConfig.isCleanupInBackground() &&
+			ttlConfig.getCleanupStrategies().getRocksdbCompactFilterCleanupStrategy() == null) {
+			ttlConfig.getCleanupStrategies().activate(ROCKSDB_COMPACTION_FILTER,
+				new StateTtlConfig.RocksdbCompactFilterCleanupStrategy(1000L));
+		}
+
+		if (ttlConfig.getCleanupStrategies().inRocksdbCompactFilter()) {
```
klion26: Maybe we need to add `ttlConfig.isEnabled()` here?

yanghua: Oh... you are right, I forgot.

flink-runtime/src/main/java/org/apache/flink/runtime/state/ttl/TtlStateFactory.java
```diff
@@ -202,6 +204,15 @@ private IS createReducingState() throws Exception {
 	}
 
 	private TtlIncrementalCleanup<K, N, TTLSV> getTtlIncrementalCleanup() {
+		//if cleanup in background and user did not specify the cleanup strategy,
+		//then build the incremental cleanup strategy and apply the default value
+		if (ttlConfig.isCleanupInBackground() &&
+			ttlConfig.getCleanupStrategies().getIncrementalCleanupStrategy() == null) {
+			ttlConfig.getCleanupStrategies().activate(
```
azagrebin: I would be against modifying `ttlConfig` in any way. It is supposed to be immutable, once created by its builder. This protects against implicit changes of `ttlConfig` which is created by user. We should actually make `CleanupStrategies.strategies/activate` private or better even move `CleanupStrategies.strategies` to builder (instead of having there `CleanupStrategies`) and remove `CleanupStrategies.activate`. The builder should then also build immutable `CleanupStrategies` for `StateTtlConfig` at the end.

If we move `isCleanupInBackground` to `CleanupStrategies` with its getter (where it rather belongs anyways) we can actually just modify `CleanupStrategies.getIncrementalCleanupStrategy/getRocksdbCompactFilterCleanupStrategy` to return default strategies as `strategies.getOrDefault` if they are not set but `isCleanupInBackground` is set. The default strategy singletons can be static final constants in respective classes. The backends and private field modifiers can also stay untouched but any strategy and backend can still independently decide what to use if `isCleanupInBackground` is set. `CleanupStrategies.inFullSnapshot/inRocksdbCompactFilter` should use the respective getters != null then, not `CleanupStrategies.strategies` directly.

docs/dev/stream/state/state.md
```diff
@@ -400,6 +400,17 @@ This option is not applicable for the incremental checkpointing in the RocksDB s
...
+
+Besides cleanup in full snapshot, you also can activate the cleanup in background. The option will adapt
+all the cleanup strategies with default values. Users who do not need to think about details or used backend.
+Currently, Flink provide two kind of cleanup strategies in background and the default config value list below:
+
+* Incremental cleanup(cleanup size: 10, run cleanup for every record: true)
+* Cleanup during RocksDB compaction(query time after how many entries: 1000)
+
+If users who want to get more fine-grained control about the special cleanup in background can use it directly.
+
```
azagrebin: We also need a code example here:
```
Besides cleanup in full snapshot, you can also activate the cleanup in background. This option will activate
a strategy configured by default if the background cleanup is supported for the used backend. 

This feature can be activated in `StateTtlConfig`:
<<<<<< ... code example is needed like in other chapters .... >>>>>

For more fine-grained control over some special cleanup in background, you can configure it separately as described below.
```

This should go to the respective chapters where the parameter values are discussed:
```
 * Incremental cleanup(cleanup size: 10, run cleanup for every record: true)
 * Cleanup during RocksDB compaction(query time after how many entries: 1000)
```

yanghua:
@azagrebin What do you think about the new changes?

docs/dev/stream/state/state.md
```diff
@@ -400,6 +400,37 @@ This option is not applicable for the incremental checkpointing in the RocksDB s
 - For existing jobs, this cleanup strategy can be activated or deactivated anytime in `StateTtlConfig`, 
 e.g. after restart from savepoint.
 
+#### Cleanup in background
+
+Besides cleanup in full snapshot, you also can activate the cleanup in background. The option will adapt
```
azagrebin: `Besides cleanup in full snapshot, you can also activate the cleanup in background. The following option will activate a default background cleanup in `StateTtlConfig` if it is supported for the used backend:`

then code example

docs/dev/stream/state/state.md
```diff
@@ -400,6 +400,37 @@ This option is not applicable for the incremental checkpointing in the RocksDB s
...
+all the cleanup strategies with default values. Users who do not need to think about details or used backend.
+Currently, Flink provide two kind of cleanup strategies in background and the default config value list below:
+
+This feature can be activated in `StateTtlConfig`:
+
+<div class="codetabs" markdown="1">
+<div data-lang="java" markdown="1">
+{% highlight java %}
+import org.apache.flink.api.common.state.StateTtlConfig;
+ StateTtlConfig ttlConfig = StateTtlConfig
```
azagrebin: space at the beginning of line

flink-core/src/main/java/org/apache/flink/api/common/state/StateTtlConfig.java
```diff
@@ -304,8 +306,10 @@ public Builder cleanupIncrementally(
 		 * https://github.com/facebook/rocksdb/blob/master/include/rocksdb/compaction_filter.h#L69
 		 * It means that the TTL filter should be tested for List state taking into account this caveat.
 		 *
+		 * @deprecated Use more general default configurate method {@link #cleanupInBackground()} instead
```
azagrebin: `default configurate` -> `configuration`

flink-core/src/main/java/org/apache/flink/api/common/state/StateTtlConfig.java
```diff
@@ -327,6 +331,15 @@ public Builder cleanupInRocksdbCompactFilter(long queryTimeAfterNumEntries) {
 			return this;
 		}
 
+		/**
+		 * Specify cleanup expired state in background. Depending on actually used backend, the corresponding cleanup will kick in.
```
azagrebin:
```
Enable cleanup of expired state in background.

<p>Depending on actually used backend, the corresponding cleanup will kick in if supported.
```

flink-core/src/main/java/org/apache/flink/api/common/state/StateTtlConfig.java
```diff
@@ -376,39 +391,57 @@ public StateTtlConfig build() {
...
+				return (RocksdbCompactFilterCleanupStrategy) strategies.get(Strategies.ROCKSDB_COMPACTION_FILTER);
+			}
 		}
 	}
 
 	/** Configuration of cleanup strategy while taking the full snapshot.  */
 	public static class IncrementalCleanupStrategy implements CleanupStrategies.CleanupStrategy {
 		private static final long serialVersionUID = 3109278696501988780L;
 
+		public static final IncrementalCleanupStrategy DEFAULT_INCREMENTAL_CLEANUP_STRATEGY = new IncrementalCleanupStrategy(10, true);
```
azagrebin: why are the default strategies public?

flink-core/src/main/java/org/apache/flink/api/common/state/StateTtlConfig.java
```diff
@@ -376,39 +391,57 @@ public StateTtlConfig build() {
...
 			activate(strategy, EMPTY_STRATEGY);
 		}
 
-		public void activate(Strategies strategy, CleanupStrategy config) {
+		private void activate(Strategies strategy, CleanupStrategy config) {
 			strategies.put(strategy, config);
 		}
 
 		public boolean inFullSnapshot() {
 			return strategies.containsKey(Strategies.FULL_STATE_SCAN_SNAPSHOT);
```
azagrebin: `getIncrementalCleanupStrategy() != null`

flink-core/src/main/java/org/apache/flink/api/common/state/StateTtlConfig.java
```diff
@@ -376,39 +391,57 @@ public StateTtlConfig build() {
...
 		}
 
+		public boolean isCleanupInBackground() {
+			return isCleanupInBackground;
+		}
+
+		private void setCleanupInBackground(boolean cleanupInBackground) {
+			isCleanupInBackground = cleanupInBackground;
```
azagrebin: should be set only in builder

flink-core/src/main/java/org/apache/flink/api/common/state/StateTtlConfig.java
```diff
@@ -360,6 +373,8 @@ public StateTtlConfig build() {
 
 		private static final CleanupStrategy EMPTY_STRATEGY = new EmptyCleanupStrategy();
 
+		private boolean isCleanupInBackground = false;
```
azagrebin: Let's make it final. If we move `isCleanupInBackground` and `strategies` to builder, we can make `CleanupStrategies` immutable by passing `isCleanupInBackground` and `strategies` to `CleanupStrategies` constructor in `builder.build()`. `CleanupStrategies` field can be removed from builder then. `activate` and `setCleanupInBackground` are also not needed in `CleanupStrategies`.

docs/dev/stream/state/state.md
```diff
@@ -400,6 +400,37 @@ This option is not applicable for the incremental checkpointing in the RocksDB s
...
+import org.apache.flink.api.common.state.StateTtlConfig
+val ttlConfig = StateTtlCon fig
+    .newBuilder(Time.seconds(1))
+    .cleanupInBackground()
+    .build
+{% endhighlight %}
+</div>
+</div>
+
+For more fine-grained control over some special cleanup in background, you can configure it separately as described below.
```
azagrebin: Let's also add:
`Currently, heap state backend relies on incremental cleanup and RocksDB backend uses compaction filter for background cleanup.`

flink-core/src/main/java/org/apache/flink/api/common/state/StateTtlConfig.java
azagrebin: the per record cleanup should be disabled by default, it can cause unpredictable performance problems.

azagrebin: I would be more conservative and set default checked entries to 5.

docs/dev/stream/state/state.md
```diff
@@ -432,9 +463,9 @@ val ttlConfig = StateTtlCon fig
 </div>
 </div>
 
-This strategy has two parameters. The first one is number of checked state entries per each cleanup triggering.
+This strategy has two parameters. The first one is number of checked state entries per each cleanup triggering (default: 10).
 If enabled, it is always triggered per each state access.
-The second parameter defines whether to trigger cleanup additionally per each record processing.
+The second parameter defines whether to trigger cleanup additionally per each record processing (default: true).
```
azagrebin: I would rather add an explicit sentence:
`If you enable the default background cleanup then this strategy will be activated for heap backend with 5 checked entries and without cleanup per record processing.`
Similar for rocks.

flink-core/src/main/java/org/apache/flink/api/common/state/StateTtlConfig.java
```diff
@@ -376,39 +391,57 @@ public StateTtlConfig build() {
...
+		}
+
+		private void setCleanupInBackground(boolean cleanupInBackground) {
+			isCleanupInBackground = cleanupInBackground;
+		}
+
 		@Nullable
 		public IncrementalCleanupStrategy getIncrementalCleanupStrategy() {
-			return (IncrementalCleanupStrategy) strategies.get(Strategies.INCREMENTAL_CLEANUP);
+			if (isCleanupInBackground()) {
```
azagrebin:
```
IncrementalCleanupStrategy defaultStrategy = isCleanupInBackground ?
    Strategies.INCREMENTAL_CLEANUP, DEFAULT_INCREMENTAL_CLEANUP_STRATEGY : null;
return (IncrementalCleanupStrategy) strategies.getOrDefault(defaultStrategy);
```

flink-core/src/main/java/org/apache/flink/api/common/state/StateTtlConfig.java
yanghua: OK, but the `EMPTY_STRATEGY` must be public because it must be used in `cleanupFullSnapshot`. If it is not public, there is a grammar error.

flink-core/src/main/java/org/apache/flink/api/common/state/StateTtlConfig.java
yanghua: > why are the default strategies public?

Otherwise, there is a grammar error.

flink-core/src/main/java/org/apache/flink/api/common/state/StateTtlConfig.java
yanghua: I think this change will cause bugs(full snapshot and incremental cleanup has their own keys), there are many test error, such as:

```
[ERROR]   HeapAsyncSnapshotTtlStateTest>TtlStateTestBase.testMultipleKeysWithSnapshotCleanup:325->TtlStateTestBase.testMultipleStateIds:388 Expired state should be unavailable expected:<null> but was:<9>
04:50:53.740 [ERROR]   HeapAsyncSnapshotTtlStateTest>TtlStateTestBase.testMultipleKeysWithSnapshotCleanup:325->TtlStateTestBase.testMultipleStateIds:388 Expired state should be unavailable expected:<[]> but was:<[1, 4]>
04:50:53.740 [ERROR]   HeapAsyncSnapshotTtlStateTest>TtlStateTestBase.testMultipleKeysWithSnapshotCleanup:325->TtlStateTestBase.testMultipleStateIds:388 Expired state should be unavailable expected:<null> but was:<7>
04:50:53.740 [ERROR]   HeapAsyncSnapshotTtlStateTest>TtlStateTestBase.testMultipleKeysWithSnapshotCleanup:325->TtlStateTestBase.testMultipleStateIds:388 Expired state should be unavailable expected:<[]> but was:<[4=4, 25=null, 15=15]>
04:50:53.740 [ERROR]   HeapAsyncSnapshotTtlStateTest>TtlStateTestBase.testMultipleKeysWithSnapshotCleanup:325->TtlStateTestBase.testMultipleStateIds:388 Expired state should be unavailable expected:<null> but was:<test value3>
04:50:53.740 [ERROR]   HeapAsyncSnapshotTtlStateTest>TtlStateTestBase.testMultipleKeysWithSnapshotCleanup:325->TtlStateTestBase.testMultipleStateIds:388 Expired state should be unavailable expected:<[]> but was:<[1, 4]>
```

log details: https://travis-ci.org/apache/flink/jobs/535110884

yanghua:
Hi @azagrebin thanks for your suggestion, I have accepted most of them, there are few suggestions I left some comments. Please review again.

yanghua:
@aljoscha and @azagrebin Any more suggestions?

flink-core/src/main/java/org/apache/flink/api/common/state/StateTtlConfig.java
azagrebin: it can be at least package private.

yanghua:
@azagrebin Thanks and any further suggestions?

flink-core/src/main/java/org/apache/flink/api/common/state/StateTtlConfig.java
azagrebin: true, I thought about another method, sorry for confusion.

flink-core/src/main/java/org/apache/flink/api/common/state/StateTtlConfig.java
```diff
@@ -322,11 +329,22 @@ public Builder cleanupInRocksdbCompactFilter() {
...
 			return this;
 		}
 
+		/**
+		 * Enable cleanup of expired state in background.
+		 *
+		 * <p>Depending on actually used backend, the corresponding cleanup will kick in if supported.
+		 */
+		@Nonnull
+		public Builder cleanupInBackground() {
```
azagrebin: I suggest we add a simple unit test for this method where `StateTtlConfig` is built like in the doc example with `cleanupInBackground` and test checks that `StateTtlConfig.getCleanupStrategies().isCleanupInBackground()` is `true`, `getIncrementalCleanupStrategy()` is not null, `getIncrementalCleanupStrategy()getCleanupSize()` is 5 etc.

yanghua:
@azagrebin thanks for your suggestion, I have fixed issues.

flink-core/src/main/java/org/apache/flink/api/common/state/StateTtlConfig.java
```diff
@@ -322,11 +329,22 @@ public Builder cleanupInRocksdbCompactFilter() {
 		 */
 		@Nonnull
 		public Builder cleanupInRocksdbCompactFilter(long queryTimeAfterNumEntries) {
-			cleanupStrategies.activate(CleanupStrategies.Strategies.ROCKSDB_COMPACTION_FILTER,
+			strategies.put(CleanupStrategies.Strategies.ROCKSDB_COMPACTION_FILTER,
```
azagrebin: Could you move to the new line `CleanupStrategies.Strategies.ROCKSDB_COMPACTION_FILTER,`?
According to the code style, we move all arguments to the new line if the line is broken.

docs/dev/stream/state/state.md
```diff
@@ -400,6 +400,35 @@ This option is not applicable for the incremental checkpointing in the RocksDB s
...
+StateTtlConfig ttlConfig = StateTtlConfig
+    .newBuilder(Time.seconds(1))
+    .cleanupInBackground()
+    .build();
+{% endhighlight %}
+</div>
+ <div data-lang="scala" markdown="1">
+{% highlight scala %}
+import org.apache.flink.api.common.state.StateTtlConfig
+val ttlConfig = StateTtlCon fig
```
azagrebin: remove space `StateTtlCon fig` -> `StateTtlConfig`

docs/dev/stream/state/state.md
```diff
@@ -400,6 +400,35 @@ This option is not applicable for the incremental checkpointing in the RocksDB s
...
+    .newBuilder(Time.seconds(1))
+    .cleanupInBackground()
```
azagrebin: `()` can be removed for `.cleanupInBackground` in scala

yanghua:
@azagrebin Really thanks for the review! cc @aljoscha

azagrebin:
@yanghua could you also squash the commits and add the description from jira issue as a detailed description in the main commit?

yanghua:
Thanks @azagrebin , have done.

GJL:
Merging

PR #8460, opened 2019-05-16
[FLINK-12524] [table-planner-blink] Introduce planner rules about rank

godfreyhe:

*Introduce planner rules about rank*

## Brief change log

  - *added CalcRankTransposeRule that transposes FlinkLogicalCalc past FlinkLogicalRank
to reduce rank input fields*
  - *added RankNumberColumnRemoveRule that emoves the output column of rank number iff there is a equality condition for the rank column*

## Verifying this change

  - *Added CalcRankTransposeRuleTest that validates CalcRankTransposeRule*
  - *Added RankNumberColumnRemoveRuleTest that validates RankNumberColumnRemoveRule*

## Documentation

### DIFF
--- a/flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/plan/rules/logical/CalcRankTransposeRule.scala
+++ b/flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/plan/rules/logical/CalcRankTransposeRule.scala
@@ -1,0 +1,197 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.flink.table.plan.rules.logical
+
+import org.apache.flink.table.api.TableException
+import org.apache.flink.table.plan.nodes.logical._
+import org.apache.flink.table.plan.util.{FlinkRexUtil, RankUtil}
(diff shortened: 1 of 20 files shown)


flink-table/flink-table-planner-blink/src/test/scala/org/apache/flink/table/plan/rules/logical/RankNumberColumnRemoveRuleTest.scala
```diff
@@ -0,0 +1,95 @@
...
+        |      ROW_NUMBER() OVER (PARTITION BY a ORDER BY rowtime DESC) as rank_num
+        |  FROM MyTable)
+        |WHERE rank_num >= 1 AND rank_num < 3
+      """.stripMargin
+
+    util.verifyPlan(sql)
+  }
+
+  @Test
+  def testCannotRemoveRankNumberColumn3(): Unit = {
```
KurtYoung: Looks like this query actually removed row number?

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/plan/rules/logical/CalcRankTransposeRule.scala
```diff
@@ -0,0 +1,197 @@
...
+import org.apache.calcite.rex.{RexBuilder, RexInputRef, RexProgram}
+import org.apache.calcite.util.ImmutableBitSet
+
+import scala.collection.JavaConversions._
+
+/**
+  * Planner rule that transposes [[FlinkLogicalCalc]] past [[FlinkLogicalRank]]
+  * to reduce rank input fields.
+  */
+class CalcRankTransposeRule
```
KurtYoung: How many benefits we can gain from this rule? Is it necessary for some kinds of optimizations?

flink-table/flink-table-planner-blink/src/test/scala/org/apache/flink/table/plan/rules/logical/RankNumberColumnRemoveRuleTest.scala
godfreyhe: the original Rank does not output row number, so RankNumberColumnRemoveRule will not be matched(it requires `isRowNumber && constantRowNumber && rank.outputRankNumber` is true)

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/plan/rules/logical/CalcRankTransposeRule.scala
godfreyhe: It's very useful for stream job to reduce Rank state.

PR #8462, opened 2019-05-16
[FLINK-12496][table-planner-blink] Support translation from StreamExecGroupWindowAggregate to StreamTransformation.

beyond1920:

The pull request aims to support translation from StreamExecGroupWindowAggregate to StreamTransformation.

## Brief change log

  - Update translateToPlanInternal in StreamExecGroupWindowAggregate. 
  - Add WindowAggregateITCase. 
  - Delete TimeIndicatorTypeInfo in blink planner module, Use TimeIndicatorTypeInfo in table-common module.
  - Fix Bug in TypeConverters when convert RowTimeIndicator or PROCTIME_INDICATOR
  - Fix Bug in StreamExecSink when inputRowType contains RowTimeIndicator field.
  - Fix Bug in StateListView and StateMapView to using correct namespace when read or write from underlying Statebackend.
  - Fix Bug in createKeyedState of PerWindowStateDataViewStore.
  - Fix Bug in TestingUpsertTableSink to set same parallism of Map and sink with input DataStream.
  - Fix minor bug in DistinctAggCodeGen.
  - Fix bug when calls equals to ROWTIME_INDICATOR or PROCTIME_INDICATOR in TimeIndicatorTypeInfo.

## Verifying this change

ITcase

  - The serializers: (no)

## Documentation


flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/calcite/FlinkTypeFactory.scala
```diff
@@ -190,11 +190,11 @@ class FlinkTypeFactory(typeSystem: RelDataTypeSystem) extends JavaTypeFactoryImp
...
       tableSchema.getFieldTypes map {
-        case TimeIndicatorTypeInfo.PROCTIME_INDICATOR
-          if isStreaming.isDefined && !isStreaming.get =>
+        case tp: TimeIndicatorTypeInfo
+          if !tp.isEventTime && isStreaming.isDefined && !isStreaming.get =>
           InternalTypes.TIMESTAMP
-        case TimeIndicatorTypeInfo.ROWTIME_INDICATOR
-          if isStreaming.isDefined && !isStreaming.get =>
+        case tp: TimeIndicatorTypeInfo
+          if tp.isEventTime && isStreaming.isDefined && !isStreaming.get =>
```
wuchong: Combine the above two case in to `case tp: TimeIndicatorTypeInfo if isStreaming.isDefined && !isStreaming.get` ?

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/plan/nodes/physical/stream/StreamExecGroupWindowAggregate.scala
```diff
@@ -122,7 +135,236 @@ class StreamExecGroupWindowAggregate(
...
+      outputRowType,
+      aggCalls,
+      namedProperties)
+
+    val timeIdx = if (isRowtimeIndicatorType(window.timeAttribute.getResultType)) {
+      if (inputTimestampIndex < 0) {
+        throw new TableException(
+          "Group Window Agg: Time attribute could not be found. \n" +
+          "Time attribute could not be found. This is a bug.\n" +
+            "please contact customer support for this"
```
wuchong: Please polish the exception message. For example: 

"Group window aggregate must defined on a time attribute, but the time attribute can't be found. This should never happen. Please file an issue."

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/plan/util/AggregateUtil.scala
```diff
@@ -695,15 +699,43 @@ object AggregateUtil extends Enumeration {
     (propPos._1, propPos._2, propPos._3)
   }
 
-  def isRowtimeIndicatorType(fieldType: TypeInformation[_]): Boolean = {
-    TimeIndicatorTypeInfo.ROWTIME_INDICATOR == fieldType
+  def isRowtimeIndicatorType(fieldType: TypeInformation[_]): Boolean = fieldType match {
```
wuchong: What do you think about to move `isRowtimeIndicatorType`, `isProctimeIndicatorType`, `isTimeIntervalType` and `isRowIntervalType` to FlinkTypeFactory? It is wired that they appears in `AggregateUtil`.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/plan/util/WindowEmitStrategy.scala
```diff
@@ -115,7 +115,10 @@ class WindowEmitStrategy(
 
 object WindowEmitStrategy {
   def apply(tableConfig: TableConfig, window: LogicalWindow): WindowEmitStrategy = {
-    val isEventTime = window.timeAttribute.getResultType == TimeIndicatorTypeInfo.ROWTIME_INDICATOR
+    val isEventTime = window.timeAttribute.getResultType match {
```
wuchong: Can we replace this with `isRowtimeIndicatorType` util function here?

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/plan/util/AggregateUtil.scala
beyond1920: FlinkTypeFactory has already referenced InternalType, RelDataType, and provide convert conversion functions between these two types.  If we move `isRowtimeIndicatorType`,` isProctimeIndicatorType`, `isTimeIntervalType` and `isRowIntervalType` to FlinkTypeFactory, then FlinkTypeFactory will reference three types, which is not good.

flink-table/flink-table-planner-blink/src/main/java/org/apache/calcite/avatica/util/DateTimeUtils.java
```diff
@@ -309,6 +309,28 @@ public static SimpleDateFormat newDateFormat(String format) {
 		return sdf;
 	}
 
+	/** Helper for CAST({timestamp} Or {date} or {time} AS VARCHAR(n)). */
+	public static String unixDateTimeToString(Object o, TimeZone tz) {
```
wuchong: This method is only used in `TestSinkUtil`, can we move it to `TestSinkUtil`? 

We want to avoid to maintain files copied from Calcite.

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/dataview/PerWindowStateDataViewStore.java
```diff
@@ -66,10 +66,7 @@ public PerWindowStateDataViewStore(
 		checkNotNull(keyedStateBackend.getKeySerializer(),
```
wuchong: This method is no longer needed and the javadoc doesn't match this method implementation. 
We can use `keyedStateBackend.getOrCreateKeyedState(windowSerializer, stateDescriptor);` instead of calling this method.

flink-table/flink-table-planner-blink/src/main/java/org/apache/calcite/avatica/util/DateTimeUtils.java
beyond1920: Sure, I will move the method to `TestSinkUtil`.
But almost all method in `DateTimeUtils` are copied from Calcite……

wuchong:
LGTM

PR #8463, opened 2019-05-16
[FLINK-12530][network] Move Task.inputGatesById to NetworkEnvironment

azagrebin:

`Task.inputGatesById` indexes `SingleInputGate`s by id. The end user of this indexing is `NetworkEnvironment` for two cases:

 - `SingleInputGate` triggers producer partition readiness check and
 then the successful result of check is dispatched back to this `SingleInputGate` by id.
 We can just return a future from `TaskActions.triggerPartitionProducerStateCheck`.
 `SingleInputGate` could use the future to react with re-triggering of the
 partition request if the producer is ready. Then `inputGatesById` is not needed for dispatching.

 - `TaskExecutor.updatePartitions` uses `inputGatesById` to dispatch `PartitionInfo` update to the right `SingleInputGate`. If `inputGatesById` is moved to `NetworkEnvironment`, which should be a better place for gate management, and `NetworkEnvironment.updatePartitionInfo` is added then
 `TaskExecutor.updatePartitions` could directly call `NetworkEnvironment.updatePartitionInfo`.

Additional refactoring:

 - `TaskActions.triggerPartitionProducerStateCheck` is
 separated into another interface `PartitionProducerStateProvider`.
 `TaskActions` is too broad interface used also for other purposes.
 Shuffle API needs only `PartitionProducerStateProvider `.

- `PartitionProducerStateProvider` returns future with the `ResponseHandle` which contains the producer state and accepts callbacks to cancel or fail consumption as a result of state check.

- `Task.triggerPartitionProducerStateCheck` is also refactored into a `RemoteChannelStateChecker` which becomes internal detail of `NetworkEnvironment`. `RemoteChannelStateChecker` accepts `ResponseHandle`, checks whether producer is ready for consumption or aborts consumption using `ResponseHandle.cancelConsumption` or `ResponseHandle.failConsumption`.

## Brief change log

  - Change TaskActions.triggerPartitionProducerStateCheck to react on future instead of inputGatesById
  - add NetworkEnviroment.updatePartitionInfo
  - use NetworkEnviroment.updatePartitionInfo in TaskExecutor.updatePartitions instead of Task.inputGatesById
  - move Move Task.inputGatesById to NetworkEnvironment
  - add SingleInputGate.close future and add a callback for it to remove gate from NetworkEnvironment.inputGatesById
- Move `TaskActions.triggerPartitionProducerStateCheck` to a
 separate interface `PartitionProducerChecker`
 - Refactor `Task.triggerPartitionProducerStateCheck` into `RemoteChannelStateChecker`

## Verifying this change

the change is a refactoring 

  - The serializers: (no)

## Documentation


azagrebin:

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NetworkEnvironment.java
```diff
@@ -110,6 +117,8 @@ public NetworkEnvironment(
 
 		this.resultPartitionManager = new ResultPartitionManager();
 
+		this.inputGatesById = new ConcurrentHashMap<>();
```
tillrohrmann: Which threads do access this structure concurrently?

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NetworkEnvironment.java
```diff
@@ -150,6 +159,11 @@ public NetworkEnvironmentConfiguration getConfiguration() {
 		return config;
 	}
 
+	@VisibleForTesting
+	public Map<IntermediateDataSetID, SingleInputGate> getInputGatesById() {
```
tillrohrmann: Do we need to expose the map or would it be enough to have `getInputGate(IntermediateDataSetID)`?

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NetworkEnvironment.java
```diff
@@ -331,6 +346,24 @@ private void registerInputMetrics(MetricGroup inputGroup, MetricGroup buffersGro
...
+	/**
+	 * Update consuming gate with newly available partition.
+	 *
+	 * @param partitionInfo telling where the partition can be retrieved from
+	 * @throws IOException IO problem by the update
+	 * @throws InterruptedException potentially blocking operation was interrupted
+	 * @throws PartitionException the input gate with the id from the partitionInfo is not found
+	 */
+	public void updatePartitionInfo(PartitionInfo partitionInfo)
+		throws IOException, InterruptedException, PartitionException {
```
tillrohrmann: Personally I'd prefer to put `throws` on the same line as `)`. The problem is that it now looks as if the `throws` line belongs to the body. Just personal taste, though.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/consumer/SingleInputGate.java
```diff
@@ -634,7 +633,8 @@ void notifyChannelNonEmpty(InputChannel channel) {
 	}
 
 	void triggerPartitionStateCheck(ResultPartitionID partitionId) {
-		taskActions.triggerPartitionProducerStateCheck(jobId, consumedResultId, partitionId);
+		taskActions.triggerPartitionProducerStateCheck(consumedResultId, partitionId,
+			() -> retriggerPartitionRequest(partitionId.getPartitionId()));
```
tillrohrmann: Line breaking

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/consumer/SingleInputGateTest.java
```diff
@@ -542,6 +543,29 @@ public void testUpdateUnknownInputChannel() throws Exception {
...
+			new InputChannelDeploymentDescriptor(
+				new ResultPartitionID(),
+				ResultPartitionLocation.createLocal())};
+		IntermediateDataSetID id = new IntermediateDataSetID();
+		InputGateDeploymentDescriptor igdd = new InputGateDeploymentDescriptor(id,
+			ResultPartitionType.PIPELINED, 0, channelDescs);
+		SingleInputGate[] inputGates = network.createInputGates("",
+			new NoOpTaskActions(), Collections.singletonList(igdd),
+			new UnregisteredMetricsGroup(), new UnregisteredMetricsGroup(), new UnregisteredMetricsGroup(),
+			new SimpleCounter());
```
tillrohrmann: line breaks

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/consumer/SingleInputGateTest.java
```diff
@@ -542,6 +543,29 @@ public void testUpdateUnknownInputChannel() throws Exception {
...
+		assertEquals(1, inputGates.length);
+		assertEquals(1, network.getInputGatesById().size());
+		assertTrue(network.getInputGatesById().containsKey(id));
+		inputGates[0].close();
+		assertEquals(0, network.getInputGatesById().size());
+		assertFalse(network.getInputGatesById().containsKey(id));
```
tillrohrmann: I'd suggest to use Hamcrest for new tests

tillrohrmann: The two last assertions seem to be redundant.

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/TaskActions.java
```diff
@@ -30,15 +30,15 @@
...
 	 * @param resultPartitionId ID of the result partition to check. This
 	 * identifies the producing execution and partition.
+	 * @param producerReadyCallback callback of producer is in one of consumable states.
 	 */
 	void triggerPartitionProducerStateCheck(
-		JobID jobId,
 		IntermediateDataSetID intermediateDataSetId,
-		ResultPartitionID resultPartitionId);
+		ResultPartitionID resultPartitionId,
+		ThrowingRunnable<Exception> producerReadyCallback);
```
tillrohrmann: Instead of adding this callback, we could also let this method return a `CompletableFuture<Void>` which if completed indicates to retrigger the partition request. That way we would not have to pass in the callback which is forwarded to some other place.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/consumer/SingleInputGate.java
```diff
@@ -185,22 +181,24 @@
...
-		JobID jobId,
 		IntermediateDataSetID consumedResultId,
 		final ResultPartitionType consumedPartitionType,
 		int consumedSubpartitionIndex,
 		int numberOfInputChannels,
 		TaskActions taskActions,
 		Counter numBytesIn,
-		boolean isCreditBased) {
+		boolean isCreditBased,
+		Runnable closeListener) {
```
tillrohrmann: For the close listener we could maybe apply a similar trick as with the partition request retriggering. We could for example add a termination future to the `SingleInputGate` which is completed once the gate gets closed. On this future we can register the removal from the `inputGatesById` map. WDYT?

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NetworkEnvironment.java
```diff
@@ -331,6 +346,24 @@ private void registerInputMetrics(MetricGroup inputGroup, MetricGroup buffersGro
...
+		IntermediateDataSetID intermediateResultPartitionID = partitionInfo.getIntermediateDataSetID();
+		SingleInputGate inputGate = inputGatesById.get(intermediateResultPartitionID);
+		if (inputGate == null) {
+			throw new PartitionException("No reader with ID " + intermediateResultPartitionID + " was found.");
```
tillrohrmann: Should this maybe be an `IllegalStateException` because this should actually not happen?

flink-runtime/src/main/java/org/apache/flink/runtime/taskexecutor/TaskExecutor.java
```diff
@@ -614,34 +612,23 @@ private void stopTaskExecutorServices() throws Exception {
...
-								log.error("Could not update input data location for task {}. Trying to fail task.", task.getTaskInfo().getTaskName(), e);
-
-								try {
-									task.failExternally(e);
-								} catch (RuntimeException re) {
-									// TODO: Check whether we need this or make exception in failExtenally checked
-									log.error("Failed canceling task with execution ID {} after task update failure.", executionAttemptID, re);
-								}
+								task.failExternally(e);
+							} catch (RuntimeException re) {
```
tillrohrmann: I would suggest to remove this catch block and instead do the following:
```
  CompletableFuture.runAsync(() -> ..., getRpcService()));
```

flink-runtime/src/main/java/org/apache/flink/runtime/taskexecutor/TaskExecutor.java
```diff
@@ -614,34 +612,23 @@ private void stopTaskExecutorServices() throws Exception {
...
 			for (final PartitionInfo partitionInfo: partitionInfos) {
-				IntermediateDataSetID intermediateResultPartitionID = partitionInfo.getIntermediateDataSetID();
-
-				final SingleInputGate singleInputGate = task.getInputGateById(intermediateResultPartitionID);
+				// Run asynchronously because it might be blocking
+				getRpcService().execute(
+					() -> {
+						try {
+							networkEnvironment.updatePartitionInfo(partitionInfo);
+						} catch (IOException | PartitionException | InterruptedException e) {
```
zhijiangW: We changed some previous behavior here but seems not covered by related tests before.
The `PartitionException` would trigger cancel task before, so the task final state might be canceled. But now we fail the task directly, so the task final state might be failed. Does it need to supplement a unit test?

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NetworkEnvironment.java
zhijiangW: Canceler thread might close `SingleInputGate` while rpc thread is creating `SingleInputGate`.

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/consumer/SingleInputGateTest.java
```diff
@@ -542,6 +543,29 @@ public void testUpdateUnknownInputChannel() throws Exception {
...
+		InputGateDeploymentDescriptor igdd = new InputGateDeploymentDescriptor(id,
+			ResultPartitionType.PIPELINED, 0, channelDescs);
+		SingleInputGate[] inputGates = network.createInputGates("",
+			new NoOpTaskActions(), Collections.singletonList(igdd),
+			new UnregisteredMetricsGroup(), new UnregisteredMetricsGroup(), new UnregisteredMetricsGroup(),
+			new SimpleCounter());
+		assertEquals(1, inputGates.length);
+		assertEquals(1, network.getInputGatesById().size());
+		assertTrue(network.getInputGatesById().containsKey(id));
+		inputGates[0].close();
```
zhijiangW: close could be put in finally part, including `network.shutdown`

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/consumer/SingleInputGateTest.java
```diff
@@ -542,6 +543,29 @@ public void testUpdateUnknownInputChannel() throws Exception {
 		}
 	}
 
+	@Test
+	public void checkInputGateRemoveInNetworkEnvironment() throws IOException {
+		NetworkEnvironment network = createNetworkEnvironment();
```
zhijiangW: Do we need `network.shutdown()` in finally part?

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/consumer/SingleInputGateTest.java
```diff
@@ -542,6 +543,29 @@ public void testUpdateUnknownInputChannel() throws Exception {
...
+		InputChannelDeploymentDescriptor[] channelDescs = new InputChannelDeploymentDescriptor[]{
+			// Local
+			new InputChannelDeploymentDescriptor(
+				new ResultPartitionID(),
+				ResultPartitionLocation.createLocal())};
+		IntermediateDataSetID id = new IntermediateDataSetID();
+		InputGateDeploymentDescriptor igdd = new InputGateDeploymentDescriptor(id,
+			ResultPartitionType.PIPELINED, 0, channelDescs);
+		SingleInputGate[] inputGates = network.createInputGates("",
```
zhijiangW: For the array case, it seems better to make size more than 1, such as 2? I am not very sure.

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/consumer/SingleInputGateTest.java
```diff
@@ -542,6 +543,29 @@ public void testUpdateUnknownInputChannel() throws Exception {
 		}
 	}
 
+	@Test
```
zhijiangW: In addition, do you think we should also cover the current case for closing `SingleInputGate` and `creatingInputGates` in different threads?

flink-runtime/src/main/java/org/apache/flink/runtime/taskexecutor/TaskExecutor.java
tillrohrmann: I think before it would send `PartitionException` back to the `JobMaster` and which would then fail the `Execution`. This should have the same effect. Verifying whether this is guarded by a test makes sense, though.

zhijiangW: Yes, I agree with the same effect, only concern the unit test. If we ever had the unit test for verifying the task is in `CANCELED` state after `PartitionException`, then this change should make the previous test failure because the task would be in `FAILED` state after change.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NetworkEnvironment.java
azagrebin: We should actually inject it as dependency to check in tests

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/consumer/SingleInputGateTest.java
azagrebin: I will rewrite a bit the test to make it more generic

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NetworkEnvironment.java
azagrebin: This is how it was reported before.. but I agree `IllegalStateException` fits better

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/consumer/SingleInputGateTest.java
azagrebin: I think we can rely on ConcurrentHashMap provided in constructor (I changed the approach a bit), then there is no need to test ConcurrentHashMap.

flink-runtime/src/main/java/org/apache/flink/runtime/taskexecutor/TaskExecutor.java
azagrebin: Does it not fail the execution currently (just on JM side)?
From what I see, `Execution.sendUpdatePartitionInfoRpcCall` reacts on failed RPC by failing itself.

azagrebin:
Thanks for the review @tillrohrmann @zhijiangW !
I have addressed the comments

azagrebin:
I have also pushed one more hotfix to separate `PartitionProducerChecker` interface from `TaskActions` because it is the only thing which InputGate requires for shuffle service API.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NetworkEnvironment.java
```diff
@@ -119,11 +118,21 @@ private NetworkEnvironment(
 		this.isShutdown = false;
 	}
 
+	public static NetworkEnvironment create(
+		NetworkEnvironmentConfiguration config,
```
zhijiangW: Keep the same indentation as the below `create` method?

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/consumer/SingleInputGateTest.java
```diff
@@ -522,30 +525,72 @@ public void testUpdateUnknownInputChannel() throws Exception {
...
-			new NoOpTaskActions(), Collections.singletonList(igdd),
-			new UnregisteredMetricsGroup(), new UnregisteredMetricsGroup(), new UnregisteredMetricsGroup(),
-			new SimpleCounter());
-		assertEquals(1, inputGates.length);
-		assertEquals(1, network.getInputGatesById().size());
-		assertTrue(network.getInputGatesById().containsKey(id));
-		inputGates[0].close();
-		assertEquals(0, network.getInputGatesById().size());
-		assertFalse(network.getInputGatesById().containsKey(id));
+	public void checkInputGateRemoveInNetworkEnvironment() throws Exception {
```
zhijiangW: checkInputGateRemoveInNetworkEnvironment -> testInputGateRemoveInNetworkEnvironment ?

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NetworkEnvironment.java
```diff
@@ -86,37 +89,90 @@
 
 	private final ResultPartitionManager resultPartitionManager;
 
+	private final Map<IntermediateDataSetID, SingleInputGate> inputGatesById;
```
zhijiangW: Maybe it is more clearly to define `ConcurrentHashMap` instead of `Map` here.

zhijiangW:
Thanks for the updates @azagrebin!

I left several minor comments, only one concern of new proposed `PartitionProducerChecker` which seems to provide the similar function as current `PartitionProducerStateChecker`. The only difference is the parameter `JobID` in `PartitionProducerStateChecker#requestPartitionProducerState`. And the `JobID` is not used in current process, so we might remove this parameter and then no need to introduce the new interface `PartitionProducerChecker`. WDYT?

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NetworkEnvironment.java
```diff
@@ -97,13 +104,15 @@ private NetworkEnvironment(
 			NetworkBufferPool networkBufferPool,
 			ConnectionManager connectionManager,
 			ResultPartitionManager resultPartitionManager,
+			Map<IntermediateDataSetID, SingleInputGate> inputGatesById,
```
tillrohrmann: Passing the map for storing the input gates into `NetworkEnvironment` exposes in my opinion a bit too many implementation details of this structure. I basically says that the object needs to store the `SingleInputGates` in this structure. Otherwise the test will fail. I think it would be better to have a `getInputGate(IntermediateDataSetID)` and let the way the gates are stored internally be an implementation detail.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NetworkEnvironment.java
azagrebin: Let's keep `Map` because `NetworkEnviroment` does not depend on `ConcurrentHashMap` methods  at the moment.

azagrebin:
Thanks for the further review @tillrohrmann. I have addressed comments.

azagrebin:
Thanks for the review @zhijiangW, I've addressed comments.
I think `PartitionProducerStateChecker` and `PartitionProducerChecker` have different levels of abstraction. The state checker is used by Task to query the execution state of the producer from JM, basically their RPC interface. Then task makes further decisions based on it. Shuffle API (remote channel) is basically interested in only whether the partition producer is producing or not at the moment. I think, there is a bit coupling between Task decisions and netty implementation at the moment. We might later further rethink it and maybe move Task logic completely into netty implementation based on just the producer state.

zhijiangW:
Thanks for the explanation @azagrebin !

If not caring about the detail implementation, we should think through the partition checker logic and make clear the scope owner of it. The below sharing is just my personal thought, maybe not very correct:

- During requesting partition, `RemoteInputChannel/InputGate` might receive `PartitionNotFoundException`, so `RemoteInputChannel/InputGate` should decide how to handle this exception. It could throw this exception to the outside directly to cause task fail. Or it wants to further query partition's state to make the final decision.

- The checker/query should be targeting the partition's state, not producer's state. If the producer state is `FINISHED` but the partition state might be `RELEASED`, then only the partition's state could give the right decision.

- `ShuffleMaster` could provide the ability for querying partition's state future, just like `ShuffleMaster` would communicate with `ShuffleService` for releasing partition. For simple implementation, we could make use of the RPC between TM/JM for the communication.

If the current partition checker refactor in this PR might not be the final way/direction to go, it is better not to touch it now, since it is not very related to the scope of moving `inputGatesById`. Or we could forward step by step and keep the current refactor in this PR.

azagrebin:
Thanks for the thoughts @zhijiangW !

True, this state might be not final. Further steps in this topic are probably more related to the partition lifecycle management and out of scope of this PR.

The main motivation of introducing the `PartitionProducerChecker` now was to separate it from `TaskActions`. `TaskActions` is very broad interface used in other components for unrelated purposes. Shuffle API does not need it in the scope of the first refactoring. Introducing `PartitionProducerChecker` improves Shuffle API decoupling because it reflects only what is needed at the moment.  I can move this hotfix into another PR if it feels like this but it is small and has been already reviewed. Also the related code is touched anyway in this PR.

zhijiangW:
Thanks for the confirmation, @azagrebin !
I am not caring about keeping that hotfix in this PR, no need to submit it separately. :)

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/PartitionStateChecker.java
```diff
@@ -0,0 +1,63 @@
...
+	 */
+	CompletableFuture<? extends CheckResult> triggerPartitionProducerCheck(
+		IntermediateDataSetID intermediateDataSetId,
+		ResultPartitionID resultPartitionId);
+
+	/**
+	 * Result of partition state check, accepts check callbacks.
+	 */
+	interface CheckResult {
+		ExecutionState getConsumerExecutionState();
```
tillrohrmann: Do we really need the consumer's execution state here?

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/consumer/InputGateID.java
```diff
@@ -0,0 +1,85 @@
...
+
+import org.apache.flink.runtime.executiongraph.ExecutionAttemptID;
+import org.apache.flink.runtime.jobgraph.IntermediateDataSetID;
+
+import java.io.Serializable;
+
+/**
+ * Runtime identifier of a consumed {@link org.apache.flink.runtime.executiongraph.IntermediateResult}.
+ *
+ * <p>In runtime the {@link org.apache.flink.runtime.jobgraph.IntermediateDataSetID} is not enough to uniquely
```
tillrohrmann: nit: At runtime

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/consumer/RemoteChannelStateChecker.java
```diff
@@ -0,0 +1,125 @@
...
+			handleFailedCheckResult(checkResult);
+		}
+		return false;
+	}
+
+	private boolean isProducerConsumerReady(CheckResult checkResult) {
+		ExecutionState consumerState = checkResult.getConsumerExecutionState();
+		Either<ExecutionState, Throwable> result = checkResult.getProducerExecutionState();
+		ExecutionState producerState = result.isLeft() ? result.left() : ExecutionState.RUNNING;
+		return consumerState == ExecutionState.RUNNING &&
```
tillrohrmann: Can we check the `consumerState` outside of the `RemoteChannelStateChecker`? If the consumer (==this) is not running, then we should simply ignore the update message.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/consumer/RemoteChannelStateChecker.java
```diff
@@ -0,0 +1,125 @@
...
+		} else {
+			handleFailedCheckResult(checkResult);
+		}
+		return false;
+	}
+
+	private boolean isProducerConsumerReady(CheckResult checkResult) {
+		ExecutionState consumerState = checkResult.getConsumerExecutionState();
+		Either<ExecutionState, Throwable> result = checkResult.getProducerExecutionState();
+		ExecutionState producerState = result.isLeft() ? result.left() : ExecutionState.RUNNING;
```
tillrohrmann: This logic is duplicated here and in `abortConsumptionOrIgnoreCheckResult`. Would be good to deduplicate it.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/consumer/RemoteChannelStateChecker.java
```diff
@@ -0,0 +1,125 @@
...
+				// see a producer in this state, something went wrong with
+				// scheduling in topological order.
+				final String msg = String.format("Producer with attempt ID %s of partition %s in unexpected state %s.",
+					resultPartitionId.getProducerId(),
+					resultPartitionId.getPartitionId(),
+					producerState);
+
+				checkResult.failConsumption(new IllegalStateException(msg));
+			}
+		} else {
```
tillrohrmann: The else branch would not be needed if we check the consumer state before calling into the `RemoteChannelStateChecker`.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/consumer/RemoteChannelStateChecker.java
```diff
@@ -0,0 +1,125 @@
...
+	private final ResultPartitionID resultPartitionId;
+
+	private final String taskNameWithSubtask;
+
+	public RemoteChannelStateChecker(ResultPartitionID resultPartitionId, String taskNameWithSubtask) {
+		this.resultPartitionId = resultPartitionId;
+		this.taskNameWithSubtask = taskNameWithSubtask;
+	}
+
+	public boolean isProducerConsumerReadyOrAbortConsumption(CheckResult checkResult) {
```
tillrohrmann: Not super important and out of scope for this PR, but this method either triggers some action (fail or cancel) or returns a decision (trigger new partition check or not). I think it would be more symmetric if this class would not trigger any action but only return a decision what to do:
```
enum Action {
  FAIL(Throwable cause),
  CANCEL(String msg),
  TRIGGER_PARTITION_CHECK,
  NOOP
}
```
Then the caller would be responsible for making the action. That way this class would only need access to `checkResult.getProducerExecutionState()` and not `checkResult` itself.

azagrebin: True, I would suggest we consider this refactoring as another issue

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/PartitionStateChecker.java
azagrebin: At the moment we use it as a shortcut if consumer is done before the check has been accomplished then we skip the check. Ideally we should use closeFuture in SingleInputGate instead of ConsumerExecutionState for this. Although this shortcut is the best effort because the state change or gate close can happen concurrently right after this check, I suggest we consider it as a separate refactoring as it would change a bit current behaviour: https://issues.apache.org/jira/browse/FLINK-12672

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/consumer/RemoteChannelStateChecker.java
azagrebin: https://issues.apache.org/jira/browse/FLINK-12672

azagrebin:
Thanks for the reviews @tillrohrmann  @zhijiangW !
I squashed the commits and adjusted the comments.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/consumer/RemoteChannelStateChecker.java
azagrebin: https://issues.apache.org/jira/browse/FLINK-13638

PR #8467, opened 2019-05-16
[FLINK-12535][network] Make CheckpointBarrierHandler non-blocking

pnowojski:
This change makes `CheckpointBarrierHandler` a non blocking class with similar interface as `InputGate`. This is one step towards a mailbox model.

## Documentation

### DIFF
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/io/AsyncDataInput.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/io/AsyncDataInput.java
@@ -1,0 +1,61 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.runtime.io;
+
+import org.apache.flink.annotation.Internal;
+
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * Interface defining couple of essential methods for asynchronous and non blocking data polling.
+ *
+ * <p>For the most efficient usage, user of this class is supposed to call {@link #pollNext()}
+ * until it returns that no more elements are available. If that happens, he should check if
+ * input {@link #isFinished()}. If not, he should wait for {@link #isAvailable()}
+ * {@link CompletableFuture} to be completed. For example:
+ *
+ * <pre>
+ * {@code
+ *	AsyncDataInput<T> input = ...;
+ *	while (!input.isFinished()) {
+ *		Optional<T> next;
+ *
+ *		while (true) {
+ *			next = input.pollNext();
+ *			if (!next.isPresent()) {
+ *				break;
+ *			}
+ *			// do something with next
+ *		}
+ *
+ *		input.isAvailable().get();
+ *	}
+ * }
+ * </pre>
+ */
+@Internal
+public interface AsyncDataInput<T> extends AvailabilityListener {
+	/**
+	 * Poll the next element. This method should be non blocking.
+	 *
+	 * @return {@code Optional.empty()} will be returned if there is no data to return or
+	 * if {@link #isFinished()} returns true. Otherwise {@code Optional.of(element)}.
+	 */
+	Optional<T> pollNext() throws Exception;
+}
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/io/AvailabilityListener.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/io/AvailabilityListener.java
@@ -1,0 +1,61 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.runtime.io;
+
+import org.apache.flink.annotation.Internal;
+
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * Interface defining couple of essential methods for listening on data availability using
+ * {@link CompletableFuture}. For usage check out for example {@link AsyncDataInput}.
+ */
+@Internal
+public interface AvailabilityListener {
+	/**
+	 * Constant that allows to avoid volatile checks {@link CompletableFuture#isDone()}. Check
+	 * {@link #isAvailable()} for more explanation.
+	 */
+	CompletableFuture<?> AVAILABLE = CompletableFuture.completedFuture(null);
+
+	/**
+	 * @return true if is finished and for example end of input was reached, false otherwise.
+	 */
+	boolean isFinished();
+
+	/**
+	 * Check if this instance is available for further processing.
+	 *
+	 * <p>When hot looping to avoid volatile access in {@link CompletableFuture#isDone()} user of
+	 * this method should do the following check:
+	 * <pre>
+	 * {@code
+	 *	AvailabilityListener input = ...;
+	 *	if (input.isAvailable() == AvailabilityListener.AVAILABLE || input.isAvailable().isDone()) {
+	 *		// do something;
+	 *	}
+	 * }
+	 * </pre>
+	 *
+	 * @return a future that is completed if there are more records available. If there are more
+	 * records available immediately, {@link #AVAILABLE} should be returned. Previously returned
+	 * not completed futures should become completed once there is more input available or if
+	 * the input {@link #isFinished()}.
+	 */
+	CompletableFuture<?> isAvailable();
+}
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/io/NullableAsyncDataInput.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/io/NullableAsyncDataInput.java
@@ -1,0 +1,39 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.runtime.io;
+
+import org.apache.flink.annotation.Internal;
+
+import javax.annotation.Nullable;
+
+/**
+ * The variant of {@link AsyncDataInput} that for performance reasons returns {@code null} from
+ * {@link #pollNextNullable()} instead returning {@code Optional.empty()} from
+ * {@link AsyncDataInput#pollNext()}.
+ */
+@Internal
+public interface NullableAsyncDataInput<T> extends AvailabilityListener {
+	/**
+	 * Poll the next element. This method should be non blocking.
+	 *
+	 * @return {@code null} will be returned if there is no data to return or
+	 * if {@link #isFinished()} returns true. Otherwise return {@code element}.
+	 */
+	@Nullable
+	T pollNextNullable() throws Exception;
+}
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/consumer/InputGate.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/consumer/InputGate.java
@@ -19,6 +19,7 @@
 package org.apache.flink.runtime.io.network.partition.consumer;
 
 import org.apache.flink.runtime.event.TaskEvent;
+import org.apache.flink.runtime.io.AsyncDataInput;
 
 import java.io.IOException;
 import java.util.Optional;
@@ -68,9 +69,7 @@ import static org.apache.flink.util.Preconditions.checkNotNull;
  * will have an input gate attached to it. This will provide its input, which will consist of one
  * subpartition from each partition of the intermediate result.
  */
-public abstract class InputGate implements AutoCloseable {
-
-	public static final CompletableFuture<?> AVAILABLE = CompletableFuture.completedFuture(null);
+public abstract class InputGate implements AsyncDataInput<BufferOrEvent>, AutoCloseable {
 
 	protected CompletableFuture<?> isAvailable = new CompletableFuture<>();
 
@@ -85,14 +84,14 @@ public abstract class InputGate implements AutoCloseable {
 	 *
 	 * @return {@code Optional.empty()} if {@link #isFinished()} returns true.
 	 */
-	public abstract Optional<BufferOrEvent> getNextBufferOrEvent() throws IOException, InterruptedException;
+	public abstract Optional<BufferOrEvent> getNext() throws IOException, InterruptedException;
 
 	/**
 	 * Poll the {@link BufferOrEvent}.
 	 *
 	 * @return {@code Optional.empty()} if there is no data to return or if {@link #isFinished()} returns true.
 	 */
-	public abstract Optional<BufferOrEvent> pollNextBufferOrEvent() throws IOException, InterruptedException;
+	public abstract Optional<BufferOrEvent> pollNext() throws IOException, InterruptedException;
 
 	public abstract void sendTaskEvent(TaskEvent event) throws IOException;
 
@@ -103,6 +102,7 @@ public abstract class InputGate implements AutoCloseable {
 	 * records available immediately, {@link #AVAILABLE} should be returned. Previously returned
 	 * not completed futures should become completed once there are more records available.
 	 */
+	@Override
 	public CompletableFuture<?> isAvailable() {
 		return isAvailable;
 	}
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/consumer/InputGate.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/consumer/InputGate.java
@@ -99,8 +99,9 @@ public abstract class InputGate implements AutoCloseable {
 	public abstract int getPageSize();
 
 	/**
-	 * @return a future that is completed if there are more records available. If there more records
-	 * available immediately, {@link #AVAILABLE} should be returned.
+	 * @return a future that is completed if there are more records available. If there are more
+	 * records available immediately, {@link #AVAILABLE} should be returned. Previously returned
+	 * not completed futures should become completed once there are more records available.
 	 */
 	public CompletableFuture<?> isAvailable() {
 		return isAvailable;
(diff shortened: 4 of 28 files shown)


pnowojski:
Test failure is unrelated and caused by: https://issues.apache.org/jira/browse/FLINK-12512

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/consumer/InputGate.java
```diff
@@ -78,32 +79,32 @@
...
-
 	public abstract void requestPartitions() throws IOException, InterruptedException;
 
 	/**
 	 * Blocking call waiting for next {@link BufferOrEvent}.
 	 *
 	 * @return {@code Optional.empty()} if {@link #isFinished()} returns true.
 	 */
-	public abstract Optional<BufferOrEvent> getNextBufferOrEvent() throws IOException, InterruptedException;
+	public abstract Optional<BufferOrEvent> getNext() throws IOException, InterruptedException;
```
AjaxXu: Even through get a buffer or event without blocking, it still can be named with getNextBufferOrEvent(). What do you think?

flink-runtime/src/main/java/org/apache/flink/runtime/io/AsyncDataInput.java
```diff
@@ -0,0 +1,50 @@
...
+ * Interface defining couple of essential methods for asynchronous and non blocking data polling.
+ */
+public interface AsyncDataInput<T> {
+	CompletableFuture<?> AVAILABLE = CompletableFuture.completedFuture(null);
+
+	/**
+	 * Poll the next element.
+	 *
+	 * @return If the {@link T} is an {@code Optional}:
+	 * 1. return {@code Optional.of(element)} - next returned element. {@code Optional.empty()} will
```
StefanRRichter: Feel like the 1. & 2. is on he wrong level which confused me a little. I would rather do 1. Optional ... 2. Otherwise and in that case the enumeration could also be dropped.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/consumer/InputGate.java
```diff
@@ -68,7 +69,7 @@
  * will have an input gate attached to it. This will provide its input, which will consist of one
  * subpartition from each partition of the intermediate result.
  */
-public abstract class InputGate implements AutoCloseable {
+public abstract class InputGate implements AsyncDataInput<Optional<BufferOrEvent>>, AutoCloseable {
 
 	public static final CompletableFuture<?> AVAILABLE = CompletableFuture.completedFuture(null);
```
sunhaibotb: The `InputGate#AVAILABLE`  should be removed because the new `AsyncDataInput` interface already has `AVAILABLE`.

flink-runtime/src/main/java/org/apache/flink/runtime/io/AsyncDataInput.java
```diff
@@ -0,0 +1,50 @@
...
+	 * be returned if there is no data to return or if {@link #isFinished()} returns true.
+	 * otherwise:
+	 * 2. next element or null if there is no data to return or if {@link #isFinished()} returns true.
+	 */
+	T pollNext() throws Exception;
```
StefanRRichter: Minor suggestion: personal taste, I usually like to have such methods named like `tryNext()`, but I know that `poll` is also very common. You can adapt it if you find that slightly more clear as well.

StefanRRichter: Exception is not mentioned in the doc, even if it is rather clear what that is about.

flink-runtime/src/main/java/org/apache/flink/runtime/io/AsyncDataInput.java
```diff
@@ -0,0 +1,50 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.runtime.io;
+
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * Interface defining couple of essential methods for asynchronous and non blocking data polling.
```
StefanRRichter: I would suggest to outline the intended usage here in the class level doc:
Consume via poll until you get some representation of null or empty, check if the data is exhausted and if not wait on the future. This might be helpful, especially because otherwise this looks pretty similar to an iterator, but can be used in a slightly different and more efficient way w.r.t. `isFinished()/hasNext()`

flink-runtime/src/main/java/org/apache/flink/runtime/io/AsyncDataInput.java
```diff
@@ -0,0 +1,50 @@
...
+
+	/**
+	 * @return true if end of input was reached, false otherwise.
+	 */
+	boolean isFinished();
+
+	/**
+	 * @return a future that is completed if there are more records available. If there are more
+	 * records available immediately, {@link #AVAILABLE} should be returned. Previously returned
+	 * not completed futures should become completed once there are more records available.
```
StefanRRichter: The interface does never talk about `records`, so I would rather go with `more input available`

flink-runtime/src/main/java/org/apache/flink/runtime/io/AsyncDataInput.java
```diff
@@ -0,0 +1,50 @@
...
+ */
+
+package org.apache.flink.runtime.io;
+
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * Interface defining couple of essential methods for asynchronous and non blocking data polling.
+ */
+public interface AsyncDataInput<T> {
```
StefanRRichter: @Internal?

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/BarrierBuffer.java
```diff
@@ -152,37 +155,37 @@ public BarrierBuffer(InputGate inputGate, BufferBlocker bufferBlocker, long maxB
 		this.queuedBuffered = new ArrayDeque<BufferOrEventSequence>();
 	}
 
+	@Override
+	public CompletableFuture<?> isAvailable() {
+		if (currentBuffered == null) {
+			return inputGate.isAvailable();
```
StefanRRichter: I think this is not ok, `InputGate#AVAILABLE` is currently not the same as `AsyncDataInput#AVAILABLE`, which messes up the checks for referencial equality. I guess you want to replace the `InputGate#AVAILABLE` with the instance from the interface.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/BarrierBuffer.java
```diff
@@ -207,11 +210,28 @@ else if (bufferOrEvent.getEvent().getClass() == CancelCheckpointMarker.class) {
...
 			}
 		}
 	}
 
+	private Optional<BufferOrEvent> handleEmptyBuffer() throws Exception {
+		if (!inputGate.isFinished()) {
+			return Optional.empty();
+		}
+
+		if (!endOfStream) {
```
StefanRRichter: Does not follow your advice to put the simple case first and even has to negate the predicate to do so. You could move the else branch first and out.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/BarrierBuffer.java
```diff
@@ -207,11 +210,28 @@ else if (bufferOrEvent.getEvent().getClass() == CancelCheckpointMarker.class) {
...
+			// end of input stream. stream continues with the buffered data
```
StefanRRichter: This comment and interaction is not very straight forward IMO. Could we pull out the code that continues with the buffered data into a dedicated method that is a) used as part of `pollNext()` for that case and b) is used here instead of `pollNext()`? Because we already know we will end up in a sub-case of `pollNext`.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/CheckpointBarrierHandler.java
```diff
@@ -19,34 +19,31 @@
...
-	 * @throws InterruptedException Thrown if the thread is interrupted while blocking during
-	 *                              waiting for the next BufferOrEvent to become available.
-	 * @throws Exception Thrown in case that a checkpoint fails that is started as the result of receiving
-	 *                   the last checkpoint barrier
+	 * Blocking version of {@link #pollNext()}.
 	 */
-	BufferOrEvent getNextNonBlocked() throws Exception;
+	@Deprecated
+	default BufferOrEvent getNextNonBlocked() throws Exception {
+		Optional<BufferOrEvent> bufferOrEvent = pollNext();
```
StefanRRichter: I am a bit confused: this does not look like a blocking implementation (e.g. using `isAvailable()`) but rather like it could actually violate the conditions on the next line.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamInputProcessor.java
```diff
@@ -206,27 +207,35 @@ public boolean processInput() throws Exception {
...
-					// Event received
-					final AbstractEvent event = bufferOrEvent.getEvent();
-					if (event.getClass() != EndOfPartitionEvent.class) {
-						throw new IOException("Unexpected event: " + event);
+			final Optional<BufferOrEvent> bufferOrEvent = barrierHandler.pollNext();
+			if (bufferOrEvent.isPresent()) {
+				processBufferOrEvent(bufferOrEvent.get());
+			} else {
+				if (!barrierHandler.isFinished()) {
+					barrierHandler.isAvailable().get();
```
StefanRRichter: This whole method is bit of a mess for my taste. Still wonder if we would not rewrite this part a little so that `processBufferOrEvent` is still called in the same iteration when availability is signaled and not after another roundtrip?

flink-runtime/src/main/java/org/apache/flink/runtime/io/AsyncDataInput.java
StefanRRichter: Right, you cannot add a lot of information about that, reason is more that you can see declared exceptions in the docs if you add them. Just a visibility thing.

StefanRRichter: Both are established and ok, I just found that poll still at times makes me slightly think more which of the two it was :D

flink-runtime/src/main/java/org/apache/flink/runtime/io/AsyncDataInput.java
```diff
@@ -0,0 +1,84 @@
...
+ *
+ * <p>For the most efficient usage, user of this class is suppose to call {@link #pollNext()}
+ * until it returns that no more elements are available. If that happens, he should check if
+ * input {@link #isFinished()}. If not, he should wait for {@link #isAvailable()}
+ * {@link CompletableFuture} to be completed. For example:
+ *
+ * <pre>
+ * {@code
+ *	while (!input.isFinished()) {
+ *		AsyncDataInput<Optional<T>> input = ...;
```
1u0: Move this assignment before the outer `while` loop?

flink-runtime/src/main/java/org/apache/flink/runtime/io/AsyncDataInput.java
```diff
@@ -0,0 +1,84 @@
...
+	 *         next returned element. {@code Optional.empty()} will be returned if there is no data
+	 *         to return or if {@link #isFinished()} returns true.
+	 *     </li>
+	 *     <li>
+	 *         If next element or {@code null} if there is no data to return or if
+	 *         {@link #isFinished()} returns true.
+	 *     </li>
+	 * </ol>
+	 */
+	T pollNext() throws Exception;
```
1u0: Instead of using `AsyncDataInput<Optional<T>>`, would it make sense to make this method explicitly return `Optional<T>`? This way, the contract in your java doc would be "enforced" by method signature explicitly.

flink-runtime/src/main/java/org/apache/flink/runtime/io/AsyncDataInput.java
```diff
@@ -0,0 +1,84 @@
...
+	 * @return
+	 * <ol>
+	 *     <li>
+	 *         If the {@link T} is an {@code Optional}: return {@code Optional.of(element)} -
+	 *         next returned element. {@code Optional.empty()} will be returned if there is no data
+	 *         to return or if {@link #isFinished()} returns true.
+	 *     </li>
+	 *     <li>
+	 *         If next element or {@code null} if there is no data to return or if
+	 *         {@link #isFinished()} returns true.
```
1u0: I personally find this documentation vague. Too many `if`s, hard to understand what is what.
Could you try to simplify this docs?

flink-runtime/src/main/java/org/apache/flink/runtime/io/AsyncDataInput.java
1u0: The `Input` interface in #8476 PR uses `AsyncDataInput<StreamElement>` and uses `null` as sentinel (the interface doesn't use `Optional<StreamElement>` due to performance concerns).

We have discussed offline with @pnowojski and there were some alternative ideas how the interface of `AsyncDataInput` may be more explicit (so the contract  of `pollNext()` method is not overloaded by `Optional.empty()` or `null` return value as marker) and more cleaner (at least, from my point of view).

I'll just leave one of the ideas that I find fine (but not enforced).
The `AsyncDataInput<T>` could have been extended to have two `poll*` methods, with the same behavior regarding traversal of input, but with just different caller conventions. So it's up to the caller to decide which flavor they prefer: 
```java
interface AsyncDataInput<T> {
    ...
    Optional<T> pollNext();
    
    @Nullable
    default T pollNextOrNull() {
         return pollNext().orElse(null);
    }
}
```

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/BarrierBuffer.java
```diff
@@ -112,6 +113,8 @@
 	/** Flag to indicate whether we have drawn all available input. */
 	private boolean endOfStream;
 
+	private boolean isFinished;
```
Aitozi: No comment for this field, I think this should be comment because it has a different behavior with `BufferTracker.java` about `finish`.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/consumer/SingleInputGate.java
```diff
@@ -494,12 +486,12 @@ public void requestPartitions() throws IOException, InterruptedException {
 	// ------------------------------------------------------------------------
 
 	@Override
-	public Optional<BufferOrEvent> getNextBufferOrEvent() throws IOException, InterruptedException {
+	public Optional<BufferOrEvent> getNext() throws IOException, InterruptedException {
 		return getNextBufferOrEvent(true);
```
Aitozi: I notice here still have a choice for blocking or non-blocking, I check the code path, does this only used for the batch task?(just a question)

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamInputProcessor.java
Aitozi: Does this means we will waste `the time of waiting for new data` delaying execute mail for the this roundtrip?

flink-runtime/src/main/java/org/apache/flink/runtime/io/AsyncDataInput.java
pnowojski: Do you have any suggestion how?

StefanRRichter:
@pnowojski reminder that I think this is good to merge, you might have a look into other comments one more time.

flink-runtime/src/main/java/org/apache/flink/runtime/io/AsyncDataInput.java
pnowojski: I have split this into `AsyncDataInput` and `NullableAsyncDataInput`.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/consumer/SingleInputGate.java
pnowojski: I think yes. Blocking method looks like it is only used in `AbstractRecordReader` and/or in tests.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/BarrierBuffer.java
```diff
@@ -152,37 +155,37 @@ public BarrierBuffer(InputGate inputGate, BufferBlocker bufferBlocker, long maxB
...
+		return AVAILABLE;
+	}
+
 	// ------------------------------------------------------------------------
 	//  Buffer and barrier handling
 	// ------------------------------------------------------------------------
 
 	@Override
-	public BufferOrEvent getNextNonBlocked() throws Exception {
+	public Optional<BufferOrEvent> pollNext() throws Exception {
```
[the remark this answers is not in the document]
pnowojski: To make it really hard for a user to forget about checking for an empty result. Alternate/equivalent solution would be to mark it `@Nullable` and enable compile time errors about mishandling `null` (but that's almost impossible to do in our code base with so much legacy code).

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamInputProcessor.java
pnowojski: @Aitozi at the moment yes. To make it work correctly with mailbox thread, this change of non blocking `BarrierHandler` has to be integrated with pending @StefanRRichter 's PR.

pnowojski:
I've addressed the other comments and pushed the changes.

Aitozi:
LGTM +1.

flink-runtime/src/main/java/org/apache/flink/runtime/io/AvailabilityListener.java
```diff
@@ -0,0 +1,61 @@
...
+	 *	}
+	 * }
+	 * </pre>
+	 *
+	 * @return a future that is completed if there are more records available. If there are more
+	 * records available immediately, {@link #AVAILABLE} should be returned. Previously returned
+	 * not completed futures should become completed once there is more input available or if
+	 * the input {@link #isFinished()}.
+	 */
+	CompletableFuture<?> isAvailable();
```
1u0: Maybe you can simplify the contract of this interface, by having only one method that returns 3 possible states of an underlying data stream? In nutshell, implementations of this interface have three situations:
 * stream has finished;
 * stream has not finished, and there is an immediate data ready to be processed;
 * stream has not finished, but the caller needs to wait when data is ready.

One way, you can apply the same trick as with `CompletableFuture<?> AVAILABLE = CompletableFuture.completedFuture(null);`.
Basically, having pre-defined different constant dummy `CompletableFuture` that would be marker of `isFinished`:

```
public interface AvailabilityListener {
    CompletableFuture<?> FINISHED = CompletableFuture.completedFuture(null);
    CompletableFuture<?> AVAILABLE = CompletableFuture.completedFuture(null);

    CompletableFuture<?> getStatusFuture();
}
```

pnowojski: At first I liked the idea, however:
1. I think it is actually more difficult for the class user. Having `isFinished()` method is pretty self explanatory, while you would have to read about `FINISHED` constant in the java doc. Yes, it solves the connection between `isFinished()` and `isAvailable()`, but the issue that I mentioned IMO kind of balances this out
2. It would be actually quite a bit harder to implement and could cause new potential issues/bugs, like someone switched `getStatusFuture()` to `FINISHED` and then accidentally back to `AVAILABLE`.

I think there is another solution, with exposing to the implementator some context class similar to the one used in mailbox, while keeping the outer interface named differently.

However same comment as I wrote to Stefan, I think this will have to be kept in sync with `SourceReader` and we should discuss it there.

StefanRRichter: There might be even another way that we can discuss here. IIRC one core problem in this PR with `CompletableFuture` is that it is a class and not an interface, so the implementation that is already completed cannot be more effcient. Did we ever consider returning a `CompletionStage` instead, which could be backed by a completable future in the cases that require waiting, and with some efficient completed implementations? All return an enum that indicates either `AVAILABLE` or `FINISHED`.

StefanRRichter: Something ala `CompletionStage<Status> inputStatus()`.

flink-runtime/src/main/java/org/apache/flink/runtime/io/NullableAsyncDataInput.java
```diff
@@ -0,0 +1,39 @@
...
+
+import javax.annotation.Nullable;
+
+/**
+ * The variant of {@link AsyncDataInput} that for performance reasons returns {@code null} from
+ * {@link #pollNextNullable()} instead returning {@code Optional.empty()} from
+ * {@link AsyncDataInput#pollNext()}.
+ */
+@Internal
+public interface NullableAsyncDataInput<T> extends AvailabilityListener {
```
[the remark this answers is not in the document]
pnowojski: Yes, I agree method name is not clear in this regard, but I think the javadoc of the method clarifies this enough.

flink-runtime/src/main/java/org/apache/flink/runtime/io/AvailabilityListener.java
pnowojski: No, we haven't thought about this. This also might be a good idea. I'll merge this as it is now, and as I wrote before, let's revisit it later after depending PRs on this are merged with context of propose `SourceReader` in mind.

PR #8468, opened 2019-05-16
[FLINK-12399][table][table-planner] Fix FilterableTableSource does not change after applyPredicate

walterddr:

This PR fixes the problem: FilterableTableSource does not generate a new digest after the predicate push down unless `explainSource()` API is explicitly override and return different digest comparing with the original table source.

## Brief change log

  - Adding a PushDownTableSource API that all predicate pushdown implementation should extend and an override for the `explainPushDown` is required.
  - Changed the `explainTerm` API from `FlinkLogicalTableSourceScan` and `PhysicalTableSourceScan` to include both `explainSource` and `explainPushDown`
  - Changed the `TestFilterableTableSource` accordingly 
  - Changed the corresponding dependent APIs such as `ProjectableTableSource` and `NestedProjectableTableSource`, `OrcTableSource` and `HBaseTableSource`

## Verifying this change
  - Covered for current tests.

  - The serializers: no

## Documentation
  - If yes, how is the feature documented? n/a

### DIFF
--- a/flink-connectors/flink-orc/src/main/java/org/apache/flink/orc/OrcTableSource.java
+++ b/flink-connectors/flink-orc/src/main/java/org/apache/flink/orc/OrcTableSource.java
@@ -213,7 +213,8 @@ public class OrcTableSource
 
 	@Override
 	public String explainSource() {
-		return "OrcFile[path=" + path + ", schema=" + orcSchema + ", filter=" + predicateString() + "]";
+		return "OrcFile[path=" + path + ", schema=" + orcSchema + ", filter=" + predicateString()
+			+ ", selectedFields=" + Arrays.toString(selectedFields) + "]";
 	}
 
 	private String predicateString() {
--- a/flink-connectors/flink-orc/src/main/java/org/apache/flink/orc/OrcTableSource.java
+++ b/flink-connectors/flink-orc/src/main/java/org/apache/flink/orc/OrcTableSource.java
@@ -218,10 +218,10 @@ public class OrcTableSource
 	}
 
 	private String predicateString() {
-		if (predicates != null) {
-			return "AND(" + Arrays.toString(predicates) + ")";
-		} else {
+		if (predicates == null || predicates.length == 0) {
 			return "TRUE";
+		} else {
+			return "AND(" + Arrays.toString(predicates) + ")";
 		}
 	}
 
--- a/flink-formats/flink-parquet/src/main/java/org/apache/flink/formats/parquet/ParquetTableSource.java
+++ b/flink-formats/flink-parquet/src/main/java/org/apache/flink/formats/parquet/ParquetTableSource.java
@@ -66,6 +66,7 @@ import org.slf4j.LoggerFactory;
 import javax.annotation.Nullable;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 
 /**
@@ -223,7 +224,8 @@ public class ParquetTableSource
 	@Override
 	public String explainSource() {
 		return "ParquetFile[path=" + path + ", schema=" + parquetSchema + ", filter=" + predicateString()
-			+ ", typeInfo=" + typeInfo + "]";
+			+ ", typeInfo=" + typeInfo + ", selectedFields=" + Arrays.toString(selectedFields)
+			+ ", pushDownStatus=" + isFilterPushedDown + ", predicates=" + predicate + "]";
 	}
 
 	private String predicateString() {
--- a/flink-formats/flink-parquet/src/main/java/org/apache/flink/formats/parquet/ParquetTableSource.java
+++ b/flink-formats/flink-parquet/src/main/java/org/apache/flink/formats/parquet/ParquetTableSource.java
@@ -225,7 +225,7 @@ public class ParquetTableSource
 	public String explainSource() {
 		return "ParquetFile[path=" + path + ", schema=" + parquetSchema + ", filter=" + predicateString()
 			+ ", typeInfo=" + typeInfo + ", selectedFields=" + Arrays.toString(selectedFields)
-			+ ", pushDownStatus=" + isFilterPushedDown + ", predicates=" + predicate + "]";
+			+ ", pushDownStatus=" + isFilterPushedDown + "]";
 	}
 
 	private String predicateString() {
--- a/flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/plan/rules/logical/PushFilterIntoTableSourceScanRule.scala
+++ b/flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/plan/rules/logical/PushFilterIntoTableSourceScanRule.scala
@@ -19,6 +19,7 @@
 package org.apache.flink.table.planner.plan.rules.logical
 
 import org.apache.flink.table.api.config.OptimizerConfigOptions
+import org.apache.flink.table.api.TableException
 import org.apache.flink.table.expressions.Expression
 import org.apache.flink.table.planner.calcite.FlinkContext
 import org.apache.flink.table.planner.expressions.converter.ExpressionConverter
@@ -103,7 +104,18 @@ class PushFilterIntoTableSourceScanRule extends RelOptRule(
     val remainingPredicates = new util.LinkedList[Expression]()
     predicates.foreach(e => remainingPredicates.add(e))
 
-    val newRelOptTable = applyPredicate(remainingPredicates, relOptTable, relBuilder.getTypeFactory)
+    val newRelOptTable: FlinkRelOptTable = 
+      applyPredicate(remainingPredicates, relOptTable, relBuilder.getTypeFactory)
+    val newTableSource = newRelOptTable.unwrap(classOf[TableSourceTable[_]]).tableSource
+    val oldTableSource = relOptTable.unwrap(classOf[TableSourceTable[_]]).tableSource
+
+    if (remainingPredicates.size() > 0
+      && newTableSource.asInstanceOf[FilterableTableSource[_]].isFilterPushedDown
+      && newTableSource.explainSource().equals(oldTableSource.explainSource)) {
+      throw new TableException("Failed to push filter into table source! "
+        + "table source with pushdown capability must override and change "
+        + "explainSource() API to explain the pushdown applied!")
+    }
 
     val newScan = new LogicalTableScan(scan.getCluster, scan.getTraitSet, newRelOptTable)
 
--- a/flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/plan/rules/logical/PushFilterIntoTableSourceScanRule.scala
+++ b/flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/plan/rules/logical/PushFilterIntoTableSourceScanRule.scala
@@ -109,8 +109,7 @@ class PushFilterIntoTableSourceScanRule extends RelOptRule(
     val newTableSource = newRelOptTable.unwrap(classOf[TableSourceTable[_]]).tableSource
     val oldTableSource = relOptTable.unwrap(classOf[TableSourceTable[_]]).tableSource
 
-    if (remainingPredicates.size() > 0
-      && newTableSource.asInstanceOf[FilterableTableSource[_]].isFilterPushedDown
+    if (newTableSource.asInstanceOf[FilterableTableSource[_]].isFilterPushedDown
       && newTableSource.explainSource().equals(oldTableSource.explainSource)) {
       throw new TableException("Failed to push filter into table source! "
         + "table source with pushdown capability must override and change "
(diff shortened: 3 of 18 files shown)


walterddr:
@flinkbot attention @KurtYoung @godfreyhe @beyond1920 please kindly take a look at the API changes. this is only my personal view so any feedback and suggestions are highly appreciated.

walterddr:
@twalthr @fhueske can you please help take a look at this PR and see if this can be the right solution? So far we've seen some reporting related to this issue in ML as well as the [JIRA comments](https://issues.apache.org/jira/browse/FLINK-12399). Thanks -Rong

flink-table/flink-table-common/src/main/java/org/apache/flink/table/sources/PushDownTableSource.java
```diff
@@ -0,0 +1,36 @@
...
+ * @param <T> The return type of the {@link TableSource}.
+ */
+public interface PushDownTableSource<T> {
+
+	/**
+	 * Describes the push down rule for this specific table source.
+	 *
+	 * @return A String explaining the {@link PushDownTableSource}.
+	 */
+	String explainPushDown();
```
walterddr: I figure I might be able to use some sort of default here but it will be hard without understanding what exactly is the pushdown is. open to any suggestion.

However I think having this API here alone prevents user from not overriding the `pushdown` and prevents the problematic pushdown rule application

KurtYoung:
Looks like this approach will break backward compatibilities

walterddr:
Hi @KurtYoung yes. the purpose of this PR is to enforce user to explicitly override this API. since there's a `default String explainSource()` method in `TableSource`. it is likely that user will forget to override the method.

Some alternative approach I can think of is
1. generalize the "predicate" definition and put them in `ProjectableTableSource`: this allows a `default String explainPushdown()` similar to `TableSource` (thus backward compatible), but this require some additional changes (and deprecate marking of the current applyFilter/applyProject APIs)
2. solution in https://github.com/apache/flink/pull/8389, in which I think the discussion lead to this PR :-)

Do you think we need the additional API change? If not I think solution in #8389 also works.

KurtYoung:
Another possible and simple solution might be:
1. we add some more documents to `ProjectableTableSource` and `FilterableTableSource` to let user know he should overwrite `explainSource`, and add required information to the returning string. Like for `ProjectableTableSource`, even if no projection had been pushdown, we should still add some digest to `explainSource` like `projectPushdown=true`, to distinguish with the table source who never tried to push project into it. 
2. If #1 is strictly applied, we can add some sanity check in project and filter pushdown rule, to make sure the digest have been changed once the rule has been applied. Letting the program failed at compile stage if he provided a wrong implementation of `ProjectableTableSource` or `FilterableTableSource` .

JingsongLi:
Minor: another source is PartitionableTableSource?

walterddr:
Thanks for the suggestions @KurtYoung @JingsongLi . actually yeah, that's pretty much what I was intended to do in #8389. But I think the addition of (1) documentations and (2) sanity checks should be a good idea.
My concern also lies with the point @godfreyhe had in the original PR: if we have other types of pushdown-able table sources. we will have a lengthy match/case clause in the TableScan rules to handle each of these separately (and like @JingsongLi mentioned, it is easy to missed one or two table source when modifying code)

However to get this move forward, I will try to separate that concern into another JIRA to discuss how do we improve this in the future. For now I will restore the PR in the original approach. What do you guys think?

KurtYoung:
@walterddr sounds good to me.

walterddr:
thanks folks for the help and review :-) I will update the PR and incorporate the comments soon

walterddr:
@KurtYoung can you help take another look on the approach? thanks

walterddr:
Thanks for the explanation @KurtYoung. NOW I think I understood what you meant. yeah I fact I think that's a better idea to only add checks and explicitly explain the exception to users. I will try constructing a solution on that direction. sorry for the misunderstanding previously! :-)

KurtYoung:
Never mind, it's me didn't explain my thought clear enough.

walterddr:
@KurtYoung sorry for the delay. I just implemented a way to safe check to prevent users from utilizing a pushdown table source without overriding and explaining the predicator pushdown in the `explainSource()` API. please kindly take a look when you have time.

The throw exception might've been too much of a harsh term. I can change it to warning and let the program pass if necessary, please let me know what do you think of this approach. thanks

flink-connectors/flink-orc/src/main/java/org/apache/flink/orc/OrcTableSource.java
```diff
@@ -213,7 +213,8 @@ public boolean isFilterPushedDown() {
 
 	@Override
 	public String explainSource() {
-		return "OrcFile[path=" + path + ", schema=" + orcSchema + ", filter=" + predicateString() + "]";
+		return "OrcFile[path=" + path + ", schema=" + orcSchema + ", filter=" + predicateString()
```
walterddr: had to change this override since in fact the OrcTableSource does not explain the pushed down predicate.

flink-formats/flink-parquet/src/main/java/org/apache/flink/formats/parquet/ParquetTableSource.java
```diff
@@ -223,7 +224,8 @@ public TableSchema getTableSchema() {
 	@Override
 	public String explainSource() {
 		return "ParquetFile[path=" + path + ", schema=" + parquetSchema + ", filter=" + predicateString()
-			+ ", typeInfo=" + typeInfo + "]";
+			+ ", typeInfo=" + typeInfo + ", selectedFields=" + Arrays.toString(selectedFields)
```
walterddr: same here.

flink-table/flink-table-planner/src/main/scala/org/apache/flink/table/plan/rules/logical/PushFilterIntoTableSourceScanRule.scala
```diff
@@ -83,6 +84,14 @@ class PushFilterIntoTableSourceScanRule extends RelOptRule(
 
     val newTableSource = filterableSource.applyPredicate(remainingPredicates)
 
+    if (remainingPredicates.size() > 0
+      && newTableSource.asInstanceOf[FilterableTableSource[_]].isFilterPushedDown
+      && newTableSource.explainSource().equals(scan.tableSource.explainSource())) {
+      throw new TableException("Failed to push filter into table source! "
```
walterddr: throwing `TableException` here. do you think we should make this change a soft one to only generate a warning and still let the planner to go through?

walterddr:
Hi @KurtYoung any additional comments on this? please kindly let me know if there's anything I misinterpret.

KurtYoung:
Sorry for the delay due to flink forward, I will review this ASAP

flink-table/flink-table-planner/src/main/scala/org/apache/flink/table/plan/rules/logical/PushFilterIntoTableSourceScanRule.scala
KurtYoung: I think throw exception is fine, since it's not easy to see the logs which generated during optimization.

flink-table/flink-table-planner/src/main/scala/org/apache/flink/table/plan/rules/logical/PushFilterIntoTableSourceScanRule.scala
```diff
@@ -83,6 +84,14 @@ class PushFilterIntoTableSourceScanRule extends RelOptRule(
 
     val newTableSource = filterableSource.applyPredicate(remainingPredicates)
 
+    if (remainingPredicates.size() > 0
```
KurtYoung: Could you also add this to the same rule in blink-planner?

flink-table/flink-table-planner/src/main/scala/org/apache/flink/table/plan/rules/logical/PushProjectIntoTableSourceScanRule.scala
```diff
@@ -50,16 +50,23 @@ class PushProjectIntoTableSourceScanRule extends RelOptRule(
     if (!(0 until scan.getRowType.getFieldCount).toArray.sameElements(accessedLogicalFields)) {
 
       // try to push projection of physical fields to TableSource
-      val newTableSource = source match {
+      val (newTableSource, isProjectSuccess) = source match {
```
KurtYoung: Could you also add this to the same rule in blink-planner?

flink-table/flink-table-planner/src/test/scala/org/apache/flink/table/api/stream/table/validation/TableSourceValidationTest.scala
```diff
@@ -0,0 +1,72 @@
...
+import org.apache.flink.table.api.{TableException, TableSchema, Types}
+import org.apache.flink.table.utils.{TableTestBase, TestFilterableTableSourceWithoutExplainSourceOverride, TestProjectableTableSourceWithoutExplainSourceOverride}
+import org.hamcrest.Matchers
+import org.junit.Test
+
+class TableSourceValidationTest extends TableTestBase {
+
+  @Test
+  def testPushProjectTableSourceWithoutExplainSource(): Unit = {
+    expectedException.expectCause(Matchers.isA(classOf[TableException]))
```
KurtYoung: you can just change the annotation to  `@Test(expected = TableException.class)`

walterddr: I actually tried using the annotation but it doesnt work since the wrapped around exception is not `TableException`. had to unparse the cause of the top level.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/plan/rules/logical/PushFilterIntoTableSourceScanRule.scala
```diff
@@ -103,7 +104,18 @@ class PushFilterIntoTableSourceScanRule extends RelOptRule(
     val remainingPredicates = new util.LinkedList[Expression]()
     predicates.foreach(e => remainingPredicates.add(e))
 
-    val newRelOptTable = applyPredicate(remainingPredicates, relOptTable, relBuilder.getTypeFactory)
+    val newRelOptTable: FlinkRelOptTable = 
+      applyPredicate(remainingPredicates, relOptTable, relBuilder.getTypeFactory)
+    val newTableSource = newRelOptTable.unwrap(classOf[TableSourceTable[_]]).tableSource
+    val oldTableSource = relOptTable.unwrap(classOf[TableSourceTable[_]]).tableSource
+
+    if (remainingPredicates.size() > 0
```
KurtYoung: unnecessary check for remainingPredicates?

flink-connectors/flink-orc/src/main/java/org/apache/flink/orc/OrcTableSource.java
godfreyhe: in `predicateString` method, it's better return `TRUE` if predicates is null or empty, otherwise return predicates string

flink-formats/flink-parquet/src/main/java/org/apache/flink/formats/parquet/ParquetTableSource.java
godfreyhe: `filter` and `predicates` are duplicate ?

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/plan/rules/logical/PushFilterIntoTableSourceScanRule.scala
godfreyhe: I also think we can check remainingPredicates here, because a filterable table source may accept all predicates but the explainSource does not been changed.
ps, `isFilterPushedDown` should be always true here, and we should also check this.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/plan/rules/logical/PushProjectIntoTableSourceScanRule.scala
```diff
@@ -67,13 +67,23 @@ class PushProjectIntoTableSourceScanRule extends RelOptRule(
...
       case projecting: ProjectableTableSource[_] =>
-        projecting.projectFields(usedFields)
+        (projecting.projectFields(usedFields), true)
+      case nonProjecting: TableSource[_] =>
+        // projection cannot be pushed to TableSource
+        (nonProjecting, false)
+    }
+
+    if (isProjectSuccess
+      && newTableSource.explainSource().equals(oldTableSource.explainSource())) {
```
godfreyhe: there is a special case: if the query is `select count(*) from projectableTableSource`, the `usedFields` to project-pushdown is empty, and the table source should keep at least one fields to output to count aggregate. if the projectable table has only one filed, the `explainSource` does not change.

flink-connectors/flink-orc/src/main/java/org/apache/flink/orc/OrcTableSource.java
walterddr: changed accordingly. actually the check for empty was missing

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/plan/rules/logical/PushProjectIntoTableSourceScanRule.scala
walterddr: I think the special case can be put outside of the scope of this PR since it is more related to how `*` is handled. I can open another JIRA for addressing this. what do you think?

KurtYoung:
@flinkbot run travis

KurtYoung:
tests passed locally, and merging this..

PR #8471, opened 2019-05-17
[FLINK-12529][runtime] Release record-deserializer buffers timely to improve the efficiency of heap usage on taskmanager

sunhaibotb:

This pull request releases the buffer of  the corresponding record deserializer after receiving `EndOfPartitionEvent` from the input channel  to improve the efficiency of heap memory usage on taskmanager.

## Brief change log

  - Change `StreamInputProcessor` to release the buffer of the corresponding record deserializer immediately after receiving `EndOfPartitionEvent` from the input channel
  - Change `StreamTwoInputProcessor` to release the buffer of the corresponding record deserializer immediately after receiving `EndOfPartitionEvent` from the input channel

## Verifying this change

This change is already covered by existing tests, such as `OneInputStreamTaskTest`, `TwoInputStreamTaskTest`.

## Documentation

### DIFF
--- a/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamTaskNetworkInput.java
+++ b/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamTaskNetworkInput.java
@@ -19,6 +19,7 @@
 package org.apache.flink.streaming.runtime.io;
 
 import org.apache.flink.annotation.Internal;
+import org.apache.flink.annotation.VisibleForTesting;
 import org.apache.flink.api.common.typeutils.TypeSerializer;
 import org.apache.flink.runtime.event.AbstractEvent;
 import org.apache.flink.runtime.io.disk.iomanager.IOManager;
@@ -81,6 +82,21 @@ public final class StreamTaskNetworkInput implements StreamTaskInput {
 		this.inputIndex = inputIndex;
 	}
 
+	@VisibleForTesting
+	StreamTaskNetworkInput(
+		CheckpointedInputGate checkpointedInputGate,
+		TypeSerializer<?> inputSerializer,
+		IOManager ioManager,
+		int inputIndex,
+		RecordDeserializer<DeserializationDelegate<StreamElement>>[] recordDeserializers) {
+
+		this.checkpointedInputGate = checkpointedInputGate;
+		this.deserializationDelegate = new NonReusingDeserializationDelegate<>(
+			new StreamElementSerializer<>(inputSerializer));
+		this.recordDeserializers = recordDeserializers;
+		this.inputIndex = inputIndex;
+	}
+
 	@Override
 	@Nullable
 	public StreamElement pollNextNullable() throws Exception {
@@ -119,6 +135,9 @@ public final class StreamTaskNetworkInput implements StreamTaskInput {
 		if (bufferOrEvent.isBuffer()) {
 			lastChannel = bufferOrEvent.getChannelIndex();
 			currentRecordDeserializer = recordDeserializers[lastChannel];
+			checkState(currentRecordDeserializer != null,
+				"currentRecordDeserializer has already been released");
+
 			currentRecordDeserializer.setNextBuffer(bufferOrEvent.getBuffer());
 		}
 		else {
@@ -128,6 +147,10 @@ public final class StreamTaskNetworkInput implements StreamTaskInput {
 			if (event.getClass() != EndOfPartitionEvent.class) {
 				throw new IOException("Unexpected event: " + event);
 			}
+
+			// release the record deserializer immediately,
+			// which is very valuable in case of bounded stream
+			releaseDeserializer(bufferOrEvent.getChannelIndex());
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


sunhaibotb:
@flinkbot attention @pnowojski

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamTwoInputProcessor.java
```diff
@@ -307,21 +307,35 @@ private void processBufferOrEvent(BufferOrEvent bufferOrEvent) throws IOExceptio
...
+		}
+
+		// cleanup the barrier handler resources
+		barrierHandler.cleanup();
+	}
+
+	private void releaseDeserializer(int channelIndex) {
+		// recycle buffers and clear the deserializer.
+		RecordDeserializer<?> deserializer = recordDeserializers[channelIndex];
+		if (deserializer != null) {
```
pnowojski: Instead of introducing another `@Nullable` field, can not we just always clear deserializer? To me it looks like `releaseDeserializer` (with `deserializer.getCurrentBuffer()` + `deserializer.clear()`) should be already an idempotent method - if yes could you

1. just do not set `recordDeserializers[channelIndex] = null;` 
2. rename `releaseDeserializer` to `clearDeserializer`

+ the same for `StreamTaskNetworkInput` class.

sunhaibotb: The reason why setting `recordDeserializers[channelIndex] = null` is that `clear()` currently does not completely release resources .There is no java doc on `RecordDeserializer#clear()`, we don't know whether the method should ensure the complete release of resources. However, in terms of actual implementation, the following resources in `SpillingAdaptiveSpanningRecordDeserializer` have not been released or incompletely: `SpanningWrapper#buffer`, `NonSpanningWrapper#utfByteBuffer`,  `NonSpanningWrapper#utfCharBuffer`.

sunhaibotb:
Thanks for reviewing @pnowojski . I will add a test for for `StreamTaskNetworkInput`.

sunhaibotb:
I added a test for `StreamTaskNetworkInput`,  and the PR has been updated. @pnowojski

sunhaibotb:
The PR has been rebased on the latest master because the construction parameters of `StreamTaskNetworkInput` have changed in the latest master branch, adding the unit test without rebasing will cause compilation failures on travis.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamTaskNetworkInput.java
```diff
@@ -128,6 +128,10 @@ private void processBufferOrEvent(BufferOrEvent bufferOrEvent) throws IOExceptio
 			if (event.getClass() != EndOfPartitionEvent.class) {
 				throw new IOException("Unexpected event: " + event);
 			}
+
```
pnowojski: 10 lines above there is a potential NPE:
```
currentRecordDeserializer = recordDeserializers[lastChannel];
checkState(currentRecordDeserializer != null, "currentRecordDeserializer has already been released");
```

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/io/StreamTaskNetworkInputTest.java
```diff
@@ -0,0 +1,68 @@
...
+		TypeSerializer<String> inSerializer = BasicTypeInfo.STRING_TYPE_INFO.createSerializer(new ExecutionConfig());
+		StreamTestSingleInputGate inputGate = new StreamTestSingleInputGate<>(numInputChannels, 1024, inSerializer);
+
+		StreamTaskNetworkInput networkInput = new StreamTaskNetworkInput(
+			new BarrierBuffer(inputGate.getInputGate(), new CachedBufferBlocker(512)),
+			inSerializer,
+			new IOManagerAsync(),
+			0);
+
+		RecordDeserializer<?>[] recordDeserializers = (RecordDeserializer<?>[]) PowerMockito
```
pnowojski: instead of using mockito to check for private implementation details, you can inject test no-op/wrapping implementations `recordDeserializers`, `checkpointedInputGate` and check whether respected cleaning methods were called.

Relaying on private field access is fragile and basically just duplicates the production code.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamTwoInputProcessor.java
pnowojski: Ok. In that case according to https://stackoverflow.com/questions/29011893/how-to-annotate-an-array-nonnull :
```
private final RecordDeserializer<DeserializationDelegate<StreamElement>> @Nullable[] recordDeserializers;
```
and please add missing null handling as per warnings.

sunhaibotb: I tried to add the `@Nullable` annotate as you said, but the grammar didn't support it, which reported the compilation error "'`@Nullable`' not applicable to type use". In addition, that question on stackoverflow uses `org.eclipse.jdt.annotation.NonNull` instead of the native annotate of JDK.

sunhaibotb: If we add the state check as another comment said (https://github.com/apache/flink/pull/8471#discussion_r302958200) , I don't think it matter whether it's marked or not here.

sunhaibotb:
The code has been updated and the travis is green (flinkbot doesn't seem to have updated the status on this page and still displays `PENDING`, but it has actually succeeded). Because of the conflict with the latest master branch, I rebased on the latest master and forcibly pushed.  @StefanRRichter  @pnowojski

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamTwoInputProcessor.java
pnowojski: Ok, let's drop this, since this seems to be supported only for eclipse...

based on https://stackoverflow.com/questions/32327134/where-does-a-nullable-annotation-refer-to-in-case-of-a-varargs-parameter

> When you are using a real Java 8 type annotation, i.e. an annotation with @Target(ElementType.TYPE_USE), the outcome in best explained by using the ordinary array declaration:

And based on:

https://stackoverflow.com/questions/4963300/which-notnull-java-annotation-should-i-use

only Eclipse's `@Nulalble` support ` @Target(ElementType.TYPE_USE) `.

pnowojski:
One more note, please also try to include better descriptions in the commit messages, for example by copying the same thing that you include in the PR description (it makes easier understanding why a change was implemented when looking at the git history).

PR #8479, opened 2019-05-19
[FLINK-11193][State Backends]Use user passed configuration overriding default configuration loading from file

Aitozi:

As described in the [jira](https://issues.apache.org/jira/browse/FLINK-11193), User's customize configuration which is configured by `backend.configure()` method will be override by the configuration loading from flink-conf.yaml. I think the config in the code should has a higher priority than the default file configuration.

## Brief change log

Merge configuration before create the new state backend.

## Verifying this change

Adding a test case to verify.

*(example:)*

## Documentation

### DIFF
--- a/flink-state-backends/flink-statebackend-rocksdb/src/main/java/org/apache/flink/contrib/streaming/state/RocksDBStateBackend.java
+++ b/flink-state-backends/flink-statebackend-rocksdb/src/main/java/org/apache/flink/contrib/streaming/state/RocksDBStateBackend.java
@@ -153,7 +153,8 @@ public class RocksDBStateBackend extends AbstractStateBackend implements Configu
 	private final RocksDBMemoryConfiguration memoryConfiguration;
 
 	/** This determines the type of priority queue state. */
-	private final PriorityQueueStateType priorityQueueStateType;
+	@Nullable
+	private PriorityQueueStateType priorityQueueStateType;
 
 	/** The default rocksdb metrics options. */
 	private final RocksDBNativeMetricOptions defaultMetricOptions;
@@ -274,8 +275,6 @@ public class RocksDBStateBackend extends AbstractStateBackend implements Configu
 		this.checkpointStreamBackend = checkNotNull(checkpointStreamBackend);
 		this.enableIncrementalCheckpointing = enableIncrementalCheckpointing;
 		this.numberOfTransferThreads = UNDEFINED_NUMBER_OF_TRANSFER_THREADS;
-		// use RocksDB-based implementation as default from FLINK-15637
-		this.priorityQueueStateType = PriorityQueueStateType.ROCKSDB;
 		this.defaultMetricOptions = new RocksDBNativeMetricOptions();
 		this.enableTtlCompactionFilter = TernaryBoolean.UNDEFINED;
 		this.memoryConfiguration = new RocksDBMemoryConfiguration();
@@ -333,10 +332,11 @@ public class RocksDBStateBackend extends AbstractStateBackend implements Configu
 		this.memoryConfiguration = RocksDBMemoryConfiguration.fromOtherAndConfiguration(original.memoryConfiguration, config);
 		this.memoryConfiguration.validate();
 
-		final String priorityQueueTypeString = config.getString(TIMER_SERVICE_FACTORY);
-
-		this.priorityQueueStateType = priorityQueueTypeString.length() > 0 ?
-			PriorityQueueStateType.valueOf(priorityQueueTypeString.toUpperCase()) : original.priorityQueueStateType;
+		if (null == original.priorityQueueStateType) {
+			this.priorityQueueStateType = config.getEnum(PriorityQueueStateType.class, TIMER_SERVICE_FACTORY);
+		} else {
+			this.priorityQueueStateType = original.priorityQueueStateType;
+		}
 
 		// configure local directories
 		if (original.localRocksDbDirectories != null) {
@@ -533,7 +533,7 @@ public class RocksDBStateBackend extends AbstractStateBackend implements Configu
 			keyGroupRange,
 			executionConfig,
 			localRecoveryConfig,
-			priorityQueueStateType,
+			getPriorityQueueStateType(),
 			ttlTimeProvider,
 			metricGroup,
 			stateHandles,
@@ -768,6 +768,22 @@ public class RocksDBStateBackend extends AbstractStateBackend implements Configu
 	@Deprecated
 	public void disableTtlCompactionFilter() {
 		enableTtlCompactionFilter = TernaryBoolean.FALSE;
+	}
+
+	/**
+	 * Gets the type of the priority queue state. It will fallback to the default value, if it is not explicitly set.
+	 * @return The type of the priority queue state.
+	 */
+	public PriorityQueueStateType getPriorityQueueStateType() {
+		return priorityQueueStateType == null ?
+			PriorityQueueStateType.valueOf(TIMER_SERVICE_FACTORY.defaultValue()) : priorityQueueStateType;
+	}
+
+	/**
+	 * Sets the type of the priority queue state. It will fallback to the default value, if it is not explicitly set.
+	 */
+	public void setPriorityQueueStateType(PriorityQueueStateType priorityQueueStateType) {
+		this.priorityQueueStateType = checkNotNull(priorityQueueStateType);
 	}
 
 	// ------------------------------------------------------------------------
--- a/flink-state-backends/flink-statebackend-rocksdb/src/test/java/org/apache/flink/contrib/streaming/state/RocksDBStateBackendConfigTest.java
+++ b/flink-state-backends/flink-statebackend-rocksdb/src/test/java/org/apache/flink/contrib/streaming/state/RocksDBStateBackendConfigTest.java
@@ -31,6 +31,7 @@ import org.apache.flink.runtime.execution.Environment;
 import org.apache.flink.runtime.io.disk.iomanager.IOManager;
 import org.apache.flink.runtime.jobgraph.JobVertexID;
 import org.apache.flink.runtime.operators.testutils.MockEnvironment;
+import org.apache.flink.runtime.operators.testutils.MockEnvironmentBuilder;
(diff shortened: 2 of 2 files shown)


Aitozi:
Hi, @StefanRRichter can you help take a look on this PR when you have time, thx.

Aitozi:
The travis-ci failure seems unrelated

flink-state-backends/flink-statebackend-rocksdb/src/main/java/org/apache/flink/contrib/streaming/state/RocksDBStateBackend.java
```diff
@@ -295,6 +297,7 @@ public RocksDBStateBackend(AbstractStateBackend checkpointStreamBackend, boolean
 	 * @param classLoader The class loader.
 	 */
 	private RocksDBStateBackend(RocksDBStateBackend original, Configuration config, ClassLoader classLoader) {
+		this.configuration = config;
```
klion26: When do we use RocksDBStateBackend, seems we always use the **original** statebackend's conf first?

Aitozi: Thanks @klion26 for your review. I'm not sure whether i quite get you meaning. But IMO this construct is only for creating a re-configured copy of the original state backend. So the configure passed in will be used first ? And also should keep this configure to a member field to reuse it to config the statebackend loaded from `StateBackendLoader#fromApplicationOrConfigOrDefault`.

klion26: But the code below(start from line 302) seems to use the **original backend's** conf first, so I'm not sure you change now is on the right directory(maybe the change you've made can't solve the problem on the issue)?

Aitozi: Hi, @klion26 I check the code again, only the `priorityQueueStateType` is configured from configuration, and the type from configuration will be used first see 
```
final String priorityQueueTypeString = config.getString(TIMER_SERVICE_FACTORY);

this.priorityQueueStateType = priorityQueueTypeString.length() > 0 ?
             PriorityQueueStateType.valueOf(priorityQueueTypeString.toUpperCase()) : original.priorityQueueStateType;

```
And other parameters are all passed by the constructor, so it  can work for the issue i think. but it looks strange indeed(some follow the original backend, some follow the config first), I will think how to make it better.

Aitozi:
Hi @klion26 I refactor the test case to verify this fix work explicitly. @StefanRRichter could you help review this bug fix, too.

flink-state-backends/flink-statebackend-rocksdb/src/main/java/org/apache/flink/contrib/streaming/state/RocksDBStateBackend.java
klion26: How about the other configs? Do we need to do the same thing here for the other configs?

Aitozi: other config are all follow this manner: check the original config whether set or undefined, if it is not defined then it will look into the config , at last fallback to the default config which is consistent with the doc of the `config()` 

> Creates a copy of this state backend that uses the values defined in the configuration for fields where that **were not yet specified** in this state backend.

But for the `priorityQueueStateType` there is no undefined type, it only has two: *rocksdb* and *heap*, if we have to follow the manner I think we just have to add an undefined value for `priorityQueueStateType`, but i am not sure do we need to do this.

What i am do now is just overriding the config loading from file with the original statebackend config to solve this problem.

klion26: Do you think we need to check the config first, then fallback the original config, and at last fallback to the default config?

Aitozi: I think we don't have to, I prefer to keep the former style, but only use the config in user code with higher priority than the default configuration loading from the file.

Aitozi:
Thanks for  your suggestion @azagrebin , I will look into this .

Aitozi:
Hi @azagrebin , I have done the fix following your comments, please help review when you are free.

Aitozi:
Do you know why this push do not trigger travis ci ? How can i trigger it manually @azagrebin ?

azagrebin:
@aitozi it triggered the CI, the setup has been recently changed, you can see CI results in the special [build comment](https://github.com/apache/flink/pull/8479#issuecomment-526521132) in this PR.

Aitozi:
@azagrebin I check the CI log just now, I think it failed due to 

> PackagesNotFoundError: The following packages are not available from current channels:
>  - python=3.3

seems unrelated to this PR ?

flink-state-backends/flink-statebackend-rocksdb/src/test/java/org/apache/flink/contrib/streaming/state/RocksDBStateBackendConfigTest.java
```diff
@@ -172,6 +173,40 @@ public void testConfigureTimerService() throws Exception {
...
+			RocksDBStateBackend.PriorityQueueStateType.HEAP.toString());
+
+		StateBackend loadedBackend = StateBackendLoader.fromApplicationOrConfigOrDefault(backend, configFromConfFile,
+			Thread.currentThread().getContextClassLoader(), null);
+
+		assertTrue(loadedBackend instanceof RocksDBStateBackend);
+		final RocksDBStateBackend rocksDBStateBackend = (RocksDBStateBackend) loadedBackend;
+
+		RocksDBKeyedStateBackend<Integer> keyedBackend2 = createKeyedStateBackend(rocksDBStateBackend, env);
+		Assert.assertEquals(RocksDBPriorityQueueSetFactory.class, keyedBackend2.getPriorityQueueFactory().getClass());
```
azagrebin: `keyedBackend(2).dispose();` at the end

flink-state-backends/flink-statebackend-rocksdb/src/test/java/org/apache/flink/contrib/streaming/state/RocksDBStateBackendConfigTest.java
```diff
@@ -172,6 +173,40 @@ public void testConfigureTimerService() throws Exception {
...
+		RocksDBStateBackend backend = new RocksDBStateBackend(tempFolder.newFolder().toURI().toString());
+		Configuration config = new Configuration();
+		config.setString(
+			RocksDBOptions.TIMER_SERVICE_FACTORY,
+			RocksDBStateBackend.PriorityQueueStateType.ROCKSDB.toString());
+		backend = backend.configure(config, Thread.currentThread().getContextClassLoader());
+
+		RocksDBKeyedStateBackend<Integer> keyedBackend = createKeyedStateBackend(backend, env);
+
+		Assert.assertEquals(RocksDBPriorityQueueSetFactory.class, keyedBackend.getPriorityQueueFactory().getClass());
```
azagrebin: In these two lines we duplicate check from `testConfigureTimerService`.
I would suggest we test only one thing in a test at a time.

azagrebin:
yes, I think it is unrelated
AFAIK it was fixed, you could rebase the PR on the current master and should go away.

flink-state-backends/flink-statebackend-rocksdb/src/main/java/org/apache/flink/contrib/streaming/state/RocksDBStateBackend.java
```diff
@@ -314,10 +315,12 @@ private RocksDBStateBackend(RocksDBStateBackend original, Configuration config,
 		this.enableTtlCompactionFilter = original.enableTtlCompactionFilter
 			.resolveUndefined(config.getBoolean(TTL_COMPACT_FILTER_ENABLED));
 
-		final String priorityQueueTypeString = config.getString(TIMER_SERVICE_FACTORY);
-
-		this.priorityQueueStateType = priorityQueueTypeString.length() > 0 ?
-			PriorityQueueStateType.valueOf(priorityQueueTypeString.toUpperCase()) : original.priorityQueueStateType;
+		if (original.priorityQueueStateType == PriorityQueueStateType.UNDEFINED) {
+			String priorityQueueTypeString = config.getString(TIMER_SERVICE_FACTORY);
+			this.priorityQueueStateType = PriorityQueueStateType.valueOf(priorityQueueTypeString.toUpperCase());
```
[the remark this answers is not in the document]
Aitozi: agree to use `null` directly.

Aitozi:
Please take a look again when you have time, I have fixed following your comments. @azagrebin

Aitozi:
ping @azagrebin

Aitozi:
OK, I have addressed your comments @azagrebin .

Aitozi:
ping @azagrebin

Aitozi:
Thanks @carp84  for help review this PR , and I have rebased the master solved the conflicts.

azagrebin:
@Aitozi @carp84 
I rebased the PR on the current master and introduced a setter for the queue to use in application code instead of `configure`. This is more aligned with the other options. I also modified the test accordingly.
Could you have a look? If there are no objections, I will merge this into master, 1.10 and maybe 1.9 depending on the effort.

carp84:
> @Aitozi @carp84
> I rebased the PR on the current master and introduced a setter for the queue to use in application code instead of `configure`. This is more aligned with the other options. I also modified the test accordingly.
> Could you have a look? If there are no objections, I will merge this into master, 1.10 and maybe 1.9 depending on the effort.

+1 (belated) on introducing a setter instead of `configure`. Sorry for the late response, just noticed...

I'd like to see this bug fix also go into release-1.10, please let me know if you need a hand on the backport @azagrebin . Thanks.

PR #8480, opened 2019-05-19
[FLINK-12552][table] Combine HiveCatalog and GenericHiveMetastoreCat…

xuefuz:
…alog

This PR is to combine GenericHiveMetastoreCatalog and HiveCatalog into one single class. The reasons are: 1. they both talk to Hive metastore to store/retrieve metadata stored in Hive and the implementation is very similar. 2. User needs two catalogs in order to store both generic metadata and Hive metadata. If the two points to the same Hive metastore instance, each will see the same list of meta objects (such as tables) but may not understand if the objects are not created by itself. This creates a lot of confusion to the users. If the two catalog doesn't share Hive metastore, then it's a deployment overhead. Using a single catalog avoids both. 

To differentiate the two types of meta objects in a single catalog, Generic meta objects will have a property named "is_generic".

## Brief change log

*(for example:)*
  - Combined and merged code in GenericHiveMetastoreCatalog and HiveCatalogBase into HiveCatalog
  - Removed GenericHiveMetastoreCatalog and HiveCatalogBase class
  - Added abstract classes for tables/views/functions/partitions/databases for both type of meta objects
  - Adapted existing tests

## Verifying this change

This change is already covered by existing tests, such as GenericInMemoryCatalogTest, GenericHiveMetastoreCatalogTest, and HiveCatalogTest.

  - The serializers: (No)
  - Anything that affects deployment or recovery: JobManager (and its components), Checkpointing, Yarn/Mesos, ZooKeeper: (No)
  - The S3 file system connector: (No)

## Documentation


xuefuz:
cc: @bowenli86 @twalthr

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalog.java
```diff
@@ -54,117 +75,494 @@
...
+		} else {
+			throw new CatalogException(String.format("Unsupported catalog database type %s", database.getClass()), null);
+		}
 
-		return new Database(
-			databaseName,
-			catalogDatabase.getComment(),
-			hiveCatalogDatabase.getLocation(),
-			hiveCatalogDatabase.getProperties());
+		createHiveDatabase(hiveDatabase, ignoreIfExists);
```
bowenli86: same as `alterHiveDatabase()`, seems no need to be on its own as nowhere else is using it

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalog.java
```diff
@@ -54,117 +75,494 @@
...
+	}
+
+	private static CatalogBaseTable instantiateHiveCatalogTable(Table hiveTable) {
+		boolean isView = TableType.valueOf(hiveTable.getTableType()) == TableType.VIRTUAL_VIEW;
 
 		// Table properties
 		Map<String, String> properties = hiveTable.getParameters();
-
-		// Table comment
+		boolean isGeneric = Boolean.valueOf(properties.get(GENERC_META_PROPERTY_KEY));
```
bowenli86: remove the key from properties before returning to user?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalog.java
```diff
@@ -54,117 +75,494 @@
...
 
-	@Override
-	protected Table createHiveTable(ObjectPath tablePath, CatalogBaseTable table) {
-		Map<String, String> properties = new HashMap<>(table.getProperties());
-
-		// Table comment
-		properties.put(HiveTableConfig.TABLE_COMMENT, table.getComment());
-
+	private  static Table instantiateHiveTable(ObjectPath tablePath, CatalogBaseTable table) {
 		Table hiveTable = new Table();
```
bowenli86: shall we verify `table` is of types among `GenericCatalogTable/View` and `HiveCatalogTable/View`, similar to how database is verified?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalog.java
```diff
@@ -54,117 +75,494 @@
...
 		hiveTable.setDbName(tablePath.getDatabaseName());
 		hiveTable.setTableName(tablePath.getObjectName());
 		hiveTable.setCreateTime((int) (System.currentTimeMillis() / 1000));
 
+		Map<String, String> properties = new HashMap<>(table.getProperties());
+		// Table comment
+		properties.put(HiveTableConfig.TABLE_COMMENT, table.getComment());
+		if (table instanceof GenericCatalogTable || table instanceof GenericCatalogView) {
+			properties = maskFlinkProperties(properties);
```
bowenli86: do we need to add the "flink.is_generic" key?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalog.java
```diff
@@ -192,46 +590,62 @@ protected Table createHiveTable(ObjectPath tablePath, CatalogBaseTable table) {
...
+		return properties.entrySet().stream()
+			.filter(e -> e.getKey() != null && e.getValue() != null)
+			.collect(Collectors.toMap(e -> FLINK_PROPERTY_PREFIX + e.getKey(), e -> e.getValue()));
+	}
+
 	// ------ partitions ------
 
 	@Override
 	public void createPartition(ObjectPath tablePath, CatalogPartitionSpec partitionSpec, CatalogPartition partition, boolean ignoreIfExists)
-			throws TableNotExistException, TableNotPartitionedException, PartitionSpecInvalidException, PartitionAlreadyExistsException, CatalogException {
```
bowenli86: revert tab change for partitions APIs all below

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/GenericInMemoryCatalog.java
```diff
@@ -48,6 +48,9 @@
  * A generic catalog implementation that holds all meta objects in memory.
  */
 public class GenericInMemoryCatalog implements Catalog {
+	public static final String FLINK_META_PROPERTY_KEY = "is_generic";
```
bowenli86: seems duplicate to HiveCatalog.GENERIC_META_PROPERTY_KEY. 

We can either use 'is_generic' and `mask/retrieveFlinkProperties()` to deal with the `flink.` prefix, or just use 'flink.is_generic' after/before calls to `mask/retrieveFlinkProperties()`. Either way, we need to be consistent

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalog.java
```diff
@@ -54,117 +75,494 @@
...
-	protected CatalogDatabase createCatalogDatabase(Database hiveDatabase) {
-		return new HiveCatalogDatabase(
-			hiveDatabase.getParameters(),
-			hiveDatabase.getLocationUri(),
-			hiveDatabase.getDescription());
+	public CatalogDatabase getDatabase(String databaseName) throws DatabaseNotExistException, CatalogException {
+		Database hiveDatabase = getHiveDatabase(databaseName);
+
+		Map<String, String> properties = hiveDatabase.getParameters();
+		boolean isGeneric = Boolean.valueOf(properties.get(GENERC_META_PROPERTY_KEY));
```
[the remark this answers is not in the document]
xuefuz: This property is useful for users as well and I think we should keep it for user to see.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalog.java
```diff
@@ -54,117 +75,494 @@
...
-	protected void validateCatalogBaseTable(CatalogBaseTable table)
-			throws CatalogException {
-		if (!(table instanceof HiveCatalogTable) && !(table instanceof HiveCatalogView)) {
+	public void alterDatabase(String databaseName, CatalogDatabase newDatabase, boolean ignoreIfNotExists)
+			throws DatabaseNotExistException, CatalogException {
+		checkArgument(!StringUtils.isNullOrWhitespaceOnly(databaseName), "databaseName cannot be null or empty");
+		checkNotNull(newDatabase, "newDatabase cannot be null");
+
+		if (newDatabase instanceof HiveCatalogDatabase) {
+			alterHiveDatabase(databaseName, instantiateHiveDatabase(databaseName, (HiveCatalogDatabase) newDatabase), ignoreIfNotExists);
```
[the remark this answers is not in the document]
xuefuz: OK

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalog.java
```diff
@@ -54,117 +75,494 @@
...
-	// ------ tables and views------
+	private static Database instantiateHiveDatabase(String databaseName, HiveCatalogDatabase database) {
+		return new Database(databaseName,
+			database.getComment(),
+			database.getLocation(),
+			database.getProperties());
+	}
+
+	private static Database instantiateHiveDatabase(String databaseName, GenericCatalogDatabase database) {
+		// Add a property to make it as a generic catalog database
```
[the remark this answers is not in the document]
xuefuz: Removed.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalog.java
```diff
@@ -54,117 +75,494 @@
 import java.util.Map;
 import java.util.stream.Collectors;
 
+import static org.apache.flink.util.Preconditions.checkArgument;
+import static org.apache.flink.util.Preconditions.checkNotNull;
+
 /**
- * A catalog implementation for Hive.
+ * Base class for catalogs backed by Hive metastore.
```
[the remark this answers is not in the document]
xuefuz: Updated

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalog.java
xuefuz: It's useful for users.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalog.java
xuefuz: The checks are done below on AbstractCatalogTable/View.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalog.java
xuefuz: ok

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalog.java
xuefuz: THE key is added in each of meta class (view, tablle, etc).

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/GenericInMemoryCatalog.java
xuefuz: This is key is added automatically as part of metaobject instantiation. It's subject to mask/retrieve operations.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalog.java
```diff
@@ -54,117 +75,494 @@
...
+public class HiveCatalog implements Catalog {
 	private static final Logger LOG = LoggerFactory.getLogger(HiveCatalog.class);
+	private static final String DEFAULT_DB = "default";
 
-	public HiveCatalog(String catalogName, String hivemetastoreURI) {
-		super(catalogName, hivemetastoreURI);
+	// Prefix used to distinguish properties created by Hive and Flink,
+	// as Hive metastore has its own properties created upon table creation and migration between different versions of metastore.
+	private static final String FLINK_PROPERTY_PREFIX = "flink.";
+	private static final String GENERC_META_PROPERTY_KEY = "flink.is_generic";
```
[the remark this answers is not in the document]
xuefuz: Updated

xuefuz:
PR is updated based on review feedback. Re. the new files shouldn't be part of the first commit but I forgot to include them. However, they cannot be in a separate JIRA as the first commit needed it and it cannot stand on its own. When we merge the PR, we can combine all commits into the first one, right? I'm not sure what issue it's. However, I can recommit a PR and do a forced push.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalog.java
```diff
@@ -54,117 +76,491 @@
...
+		Database hiveDatabase;
+		if (database instanceof HiveCatalogDatabase) {
+			hiveDatabase = instantiateHiveDatabase(databaseName, (HiveCatalogDatabase) database);
+		} else if (database instanceof GenericCatalogDatabase) {
+			hiveDatabase = instantiateHiveDatabase(databaseName, (GenericCatalogDatabase) database);
+		} else {
+			throw new CatalogException(String.format("Unsupported catalog database type %s", database.getClass()), null);
+		}
+
+		createHiveDatabase(hiveDatabase, ignoreIfExists);
```
bowenli86: can we remove `createHiveDatabase()` and put its logic here given that no one else is using `createHiveDatabase()`?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalog.java
```diff
@@ -54,117 +76,491 @@
...
+		if (table instanceof GenericCatalogTable || table instanceof GenericCatalogView) {
+			properties = maskFlinkProperties(properties);
+		}
 		// Table properties
 		hiveTable.setParameters(properties);
 
 		// Hive table's StorageDescriptor
 		// TODO: This is very basic Hive table.
-		//  [FLINK-11479] Add input/output format and SerDeLib information for Hive tables in HiveCatalogUtil#createHiveTable
+		//  [FLINK-11479] Add input/output format and SerDeLib information for Hive tables in HiveCatalogUtil#instantiateHiveTable
```
bowenli86: minor: we don't have `HiveCatalogUtil` class anymore

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/AbstractCatalogFunction.java
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * An abstract catalog function implementation.
+ */
+public abstract class AbstractCatalogFunction implements CatalogFunction {
+	private final String className; // Fully qualified class name of the function
+	private final Map<String, String> properties;
+
+	public AbstractCatalogFunction(String className, Map<String, String> properties) {
+		this.className = checkNotNull(className, "className cannot be null");
```
bowenli86: check if it's empty too?

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/AbstractCatalogTable.java
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	@Override
+	public Map<String, String> getProperties() {
+		return properties;
+	}
+
+	@Override
+	public TableSchema getSchema() {
+		return this.tableSchema;
```
bowenli86: minor: I observed that the getters in `AbstractCatalogTable/VIew` sometimes uses `return this.xxx` and sometimes `return xxx`. I'm fine with either. Just that can we be a bit more consistent to use only one of them?

xuefuz: This is not necessarily introduced by this PR. AbstractXXX classes are mostly extracted from original concrete classes. Nevertheless. I make the changes.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/AbstractCatalogFunction.java
xuefuz: ok

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalog.java
xuefuz: This is a private method and has its unit functionality. While being called by only one place, I don't see any harm to keep it this way.

xuefuz:
Thanks for the review. Those abstract classes are part of this PR, and they weren't planned as a separate PR. Thus, I don't see a good reason to have a separate JIRA/PR for them.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/AbstractCatalogFunction.java
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * An abstract catalog function implementation.
+ */
+public abstract class AbstractCatalogFunction implements CatalogFunction {
+	private final String className; // Fully qualified class name of the function
+	private final Map<String, String> properties;
+
+	public AbstractCatalogFunction(String className, Map<String, String> properties) {
+		checkArgument(!StringUtils.isNullOrWhitespaceOnly(className));
```
bowenli86: missing error message. I will add for you when merging code.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/AbstractCatalogView.java
```diff
@@ -0,0 +1,77 @@
...
+	// Expanded query text takes care of the this, as an example.
+	private final String expandedQuery;
+
+	private final TableSchema schema;
+	private final Map<String, String> properties;
+	private final String comment;
+
+	public AbstractCatalogView(String originalQuery, String expandedQuery, TableSchema schema,
+			Map<String, String> properties, String comment) {
+		this.originalQuery = checkNotNull(originalQuery, "originalQuery cannot be null");
```
bowenli86: should also check whether the string is empty. I'll change for you when merging code

bowenli86:
> Those abstract classes are part of this PR, and they weren't planned as a separate PR. Thus, I don't see a good reason to have a separate JIRA/PR for them.

First of all, I want to clarify that, IMO, adding abstract classes is a good change :) and my suggestion is not to diminish this effort.

But I disagree here because, from what I can tell, it doesn't have much to do with unifying two catalogs, and can happen independently. It doesn't need to be on its own PR. However, I believe it's better for us to create its own JIRA though it's not planned in advance, and have it in its own commit in this PR. This is part of the "Separate Refactoring, Cleanup and Independent Changes" requirement in the [guide](https://docs.google.com/document/d/1JGRVK6udBFa8xbyv7ipPQiLEhxebA64f8KKcaAH0UPY).  The isolation of changes makes review, merge, track much easier.

------

LGTM besides a few missing parts that I'll take care of for you. Will merge once CI passes.

PR #8485, opened 2019-05-20
[FLINK-12555] Introduce an encapsulated metric group layout for shuffle API

azagrebin:

At the moment, partition/gate create methods in NetworkEnvironment have a lot of metrics arguments to maintain original layout for metric groups. This approach is not quite encapsulated and clean for shuffle API. We can have just one parent group for shuffle metrics. The old layout can be still maintained in parallel and deprecated. At the moment we can do it with a couple of casts (if shuffle implementation is NetworkEnvironment) and adding an additional legacy metric registration which can be removed later.

## Brief change log

  - Change `NetworkEnvironment.createResultPartitionWriters/createInputGates` to have only one parent metric group argument.
  - Add InputChannelMetricsWithLegacy to increment input metrics from the legacy group as well
  - Move legacy metric group creation to deprecated `NetworkEnvironment.registerLegacyNetworkMetrics`
  - clean `Task` of legacy code but add call to `NetworkEnvironment.registerLegacyNetworkMetrics` after creation of partitions and gates (later needs instanceOf check for actual ShuffleService)

## Verifying this change

simple refactoring, existing tests

  - The serializers: (no)

## Documentation

### DIFF
--- a/docs/monitoring/metrics.md
+++ b/docs/monitoring/metrics.md
@@ -1000,7 +1000,8 @@ Thus, in order to infer the metric identifier:
   </tbody>
 </table>
 
-### Network
+
+### Network (Deprecated: use [Default shuffle service metrics]({{ site.baseurl }}/monitoring/metrics.html#default-shuffle-service))
 <table class="table table-bordered">
   <thead>
     <tr>
@@ -1044,10 +1045,10 @@ Thus, in order to infer the metric identifier:
     <tr>
       <td>outPoolUsage</td>
       <td>An estimate of the output buffers usage.</td>
-      <td>Gauge</td>      
+      <td>Gauge</td>
     </tr>
     <tr>
-      <td rowspan="4">Network.&lt;Input|Output&gt;.&lt;gate&gt;<br />
+      <td rowspan="4">Network.&lt;Input|Output&gt;.&lt;gate|partition&gt;<br />
         <strong>(only available if <tt>taskmanager.net.detailed-metrics</tt> config option is set)</strong></td>
       <td>totalQueueLen</td>
       <td>Total number of queued buffers in all input/output channels.</td>
@@ -1067,6 +1068,123 @@ Thus, in order to infer the metric identifier:
       <td>avgQueueLen</td>
       <td>Average number of queued buffers in all input/output channels.</td>
       <td>Gauge</td>
+    </tr>
+  </tbody>
+</table>
+
+### Default shuffle service
+
+Metrics related to data exchange between task executors using netty network communication.
+
+<table class="table table-bordered">
+  <thead>
+    <tr>
+      <th class="text-left" style="width: 18%">Scope</th>
+      <th class="text-left" style="width: 22%">Infix</th>
+      <th class="text-left" style="width: 22%">Metrics</th>
+      <th class="text-left" style="width: 30%">Description</th>
+      <th class="text-left" style="width: 8%">Type</th>
+    </tr>
+  </thead>
+  <tbody>
+    <tr>
+      <th rowspan="2"><strong>TaskManager</strong></th>
+      <td rowspan="2">Status.Shuffle.Netty</td>
+      <td>AvailableMemorySegments</td>
+      <td>The number of unused memory segments.</td>
+      <td>Gauge</td>
+    </tr>
+    <tr>
+      <td>TotalMemorySegments</td>
+      <td>The number of allocated memory segments.</td>
+      <td>Gauge</td>
+    </tr>
+    <tr>
+      <th rowspan="8">Task</th>
+      <td rowspan="2">Shuffle.Netty.Input.Buffers</td>
+      <td>inputQueueLength</td>
+      <td>The number of queued input buffers.</td>
+      <td>Gauge</td>
+    </tr>
+    <tr>
+      <td>inPoolUsage</td>
+      <td>An estimate of the input buffers usage.</td>
+      <td>Gauge</td>
+    </tr>
+    <tr>
+      <td rowspan="2">Shuffle.Netty.Output.Buffers</td>
+      <td>outputQueueLength</td>
+      <td>The number of queued output buffers.</td>
+      <td>Gauge</td>
+    </tr>
+    <tr>
+      <td>outPoolUsage</td>
+      <td>An estimate of the output buffers usage.</td>
+      <td>Gauge</td>
+    </tr>
+    <tr>
+      <td rowspan="4">Shuffle.Netty.&lt;Input|Output&gt;.&lt;gate|partition&gt;<br />
+        <strong>(only available if <tt>taskmanager.net.detailed-metrics</tt> config option is set)</strong></td>
+      <td>totalQueueLen</td>
+      <td>Total number of queued buffers in all input/output channels.</td>
+      <td>Gauge</td>
+    </tr>
+    <tr>
+      <td>minQueueLen</td>
+      <td>Minimum number of queued buffers in all input/output channels.</td>
+      <td>Gauge</td>
+    </tr>
+    <tr>
+      <td>maxQueueLen</td>
+      <td>Maximum number of queued buffers in all input/output channels.</td>
+      <td>Gauge</td>
+    </tr>
+    <tr>
+      <td>avgQueueLen</td>
+      <td>Average number of queued buffers in all input/output channels.</td>
+      <td>Gauge</td>
+    </tr>
+    <tr>
+      <th rowspan="8"><strong>Task</strong></th>
+      <td rowspan="8">Shuffle.Netty.Input</td>
+      <td>numBytesInLocal</td>
+      <td>The total number of bytes this task has read from a local source.</td>
+      <td>Counter</td>
+    </tr>
+    <tr>
+      <td>numBytesInLocalPerSecond</td>
+      <td>The number of bytes this task reads from a local source per second.</td>
+      <td>Meter</td>
+    </tr>
+    <tr>
+      <td>numBytesInRemote</td>
+      <td>The total number of bytes this task has read from a remote source.</td>
+      <td>Counter</td>
+    </tr>
+    <tr>
+      <td>numBytesInRemotePerSecond</td>
+      <td>The number of bytes this task reads from a remote source per second.</td>
+      <td>Meter</td>
+    </tr>
+    <tr>
+      <td>numBuffersInLocal</td>
+      <td>The total number of network buffers this task has read from a local source.</td>
+      <td>Counter</td>
+    </tr>
+    <tr>
+      <td>numBuffersInLocalPerSecond</td>
+      <td>The number of network buffers this task reads from a local source per second.</td>
+      <td>Meter</td>
+    </tr>
+    <tr>
+      <td>numBuffersInRemote</td>
+      <td>The total number of network buffers this task has read from a remote source.</td>
+      <td>Counter</td>
+    </tr>
+    <tr>
+      <td>numBuffersInRemotePerSecond</td>
+      <td>The number of network buffers this task reads from a remote source per second.</td>
+      <td>Meter</td>
     </tr>
   </tbody>
 </table>
@@ -1236,42 +1354,42 @@ Certain RocksDB native metrics are available but disabled by default, you can fi
     <tr>
       <th rowspan="12"><strong>Task</strong></th>
       <td>numBytesInLocal</td>
-      <td>The total number of bytes this task has read from a local source.</td>
+      <td><span class="label label-danger">Attention:</span> deprecated, use <a href="{{ site.baseurl }}/monitoring/metrics.html#default-shuffle-service">Default shuffle service metrics</a>.</td>
       <td>Counter</td>
     </tr>
     <tr>
       <td>numBytesInLocalPerSecond</td>
-      <td>The number of bytes this task reads from a local source per second.</td>
+      <td><span class="label label-danger">Attention:</span> deprecated, use <a href="{{ site.baseurl }}/monitoring/metrics.html#default-shuffle-service">Default shuffle service metrics</a>.</td>
       <td>Meter</td>
     </tr>
     <tr>
       <td>numBytesInRemote</td>
-      <td>The total number of bytes this task has read from a remote source.</td>
+      <td><span class="label label-danger">Attention:</span> deprecated, use <a href="{{ site.baseurl }}/monitoring/metrics.html#default-shuffle-service">Default shuffle service metrics</a>.</td>
       <td>Counter</td>
     </tr>
     <tr>
       <td>numBytesInRemotePerSecond</td>
-      <td>The number of bytes this task reads from a remote source per second.</td>
+      <td><span class="label label-danger">Attention:</span> deprecated, use <a href="{{ site.baseurl }}/monitoring/metrics.html#default-shuffle-service">Default shuffle service metrics</a>.</td>
       <td>Meter</td>
     </tr>
     <tr>
       <td>numBuffersInLocal</td>
-      <td>The total number of network buffers this task has read from a local source.</td>
+      <td><span class="label label-danger">Attention:</span> deprecated, use <a href="{{ site.baseurl }}/monitoring/metrics.html#default-shuffle-service">Default shuffle service metrics</a>.</td>
       <td>Counter</td>
     </tr>
     <tr>
       <td>numBuffersInLocalPerSecond</td>
-      <td>The number of network buffers this task reads from a local source per second.</td>
+      <td><span class="label label-danger">Attention:</span> deprecated, use <a href="{{ site.baseurl }}/monitoring/metrics.html#default-shuffle-service">Default shuffle service metrics</a>.</td>
       <td>Meter</td>
     </tr>
     <tr>
       <td>numBuffersInRemote</td>
-      <td>The total number of network buffers this task has read from a remote source.</td>
+      <td><span class="label label-danger">Attention:</span> deprecated, use <a href="{{ site.baseurl }}/monitoring/metrics.html#default-shuffle-service">Default shuffle service metrics</a>.</td>
       <td>Counter</td>
     </tr>
     <tr>
       <td>numBuffersInRemotePerSecond</td>
-      <td>The number of network buffers this task reads from a remote source per second.</td>
+      <td><span class="label label-danger">Attention:</span> deprecated, use <a href="{{ site.baseurl }}/monitoring/metrics.html#default-shuffle-service">Default shuffle service metrics</a>.</td>
       <td>Meter</td>
     </tr>
     <tr>
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NettyShuffleEnvironment.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NettyShuffleEnvironment.java
@@ -25,6 +25,7 @@ import org.apache.flink.runtime.deployment.InputGateDeploymentDescriptor;
 import org.apache.flink.runtime.deployment.ResultPartitionDeploymentDescriptor;
 import org.apache.flink.runtime.executiongraph.ExecutionAttemptID;
 import org.apache.flink.runtime.executiongraph.PartitionInfo;
+import org.apache.flink.runtime.io.network.api.writer.ResultPartitionWriter;
 import org.apache.flink.runtime.io.network.buffer.NetworkBufferPool;
 import org.apache.flink.runtime.io.network.metrics.InputBufferPoolUsageGauge;
 import org.apache.flink.runtime.io.network.metrics.InputBuffersGauge;
@@ -46,6 +47,7 @@ import org.apache.flink.runtime.jobgraph.IntermediateDataSetID;
 import org.apache.flink.runtime.shuffle.NettyShuffleDescriptor;
 import org.apache.flink.runtime.shuffle.ShuffleDescriptor;
 import org.apache.flink.runtime.shuffle.ShuffleEnvironment;
+import org.apache.flink.runtime.shuffle.ShuffleIOOwnerContext;
 import org.apache.flink.runtime.taskmanager.NettyShuffleEnvironmentConfiguration;
 import org.apache.flink.util.Preconditions;
 
@@ -60,6 +62,7 @@ import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
 
 import static org.apache.flink.util.Preconditions.checkArgument;
+import static org.apache.flink.util.Preconditions.checkNotNull;
 
 /**
  * The implementation of {@link ShuffleEnvironment} based on netty network communication, local memory and disk files.
@@ -70,8 +73,30 @@ public class NettyShuffleEnvironment implements ShuffleEnvironment<ResultPartiti
 
 	private static final Logger LOG = LoggerFactory.getLogger(NettyShuffleEnvironment.class);
 
+	// deprecated metric groups
+
+	@SuppressWarnings("DeprecatedIsStillUsed")
+	@Deprecated
+	static final String METRIC_GROUP_NETWORK_DEPRECATED = "Network";
+	@SuppressWarnings("DeprecatedIsStillUsed")
+	@Deprecated
+	private static final String METRIC_GROUP_BUFFERS_DEPRECATED = "buffers";
+
+	// task level metric group structure: Shuffle.Netty.<Input|Output>.Buffers
+
+	static final String METRIC_GROUP_SHUFFLE = "Shuffle";
+	static final String METRIC_GROUP_NETTY = "Netty";
+	private static final String METRIC_GROUP_OUTPUT = "Output";
+	private static final String METRIC_GROUP_INPUT = "Input";
+	private static final String METRIC_GROUP_BUFFERS = "Buffers";
+
+	// task level output metrics: Shuffle.Netty.Output.*
+
 	private static final String METRIC_OUTPUT_QUEUE_LENGTH = "outputQueueLength";
 	private static final String METRIC_OUTPUT_POOL_USAGE = "outPoolUsage";
+
+	// task level input metrics: Shuffle.Netty.Input.*
+
 	private static final String METRIC_INPUT_QUEUE_LENGTH = "inputQueueLength";
 	private static final String METRIC_INPUT_POOL_USAGE = "inPoolUsage";
 
@@ -166,54 +191,68 @@ public class NettyShuffleEnvironment implements ShuffleEnvironment<ResultPartiti
 	// --------------------------------------------------------------------------------------------
 
 	@Override
-	public Collection<ResultPartition> createResultPartitionWriters(
+	public ShuffleIOOwnerContext createShuffleIOOwnerContext(
 			String ownerName,
 			ExecutionAttemptID executionAttemptID,
-			Collection<ResultPartitionDeploymentDescriptor> resultPartitionDeploymentDescriptors,
-			MetricGroup outputGroup,
-			MetricGroup buffersGroup) {
+			MetricGroup parentGroup) {
+		MetricGroup nettyGroup = checkNotNull(parentGroup).addGroup(METRIC_GROUP_SHUFFLE).addGroup(METRIC_GROUP_NETTY);
+		return new ShuffleIOOwnerContext(
+			checkNotNull(ownerName),
+			checkNotNull(executionAttemptID),
+			parentGroup,
+			nettyGroup.addGroup(METRIC_GROUP_INPUT),
+			nettyGroup.addGroup(METRIC_GROUP_OUTPUT));
+	}
+
+	@Override
+	public Collection<ResultPartition> createResultPartitionWriters(
+			ShuffleIOOwnerContext ownerContext,
+			Collection<ResultPartitionDeploymentDescriptor> resultPartitionDeploymentDescriptors) {
 		synchronized (lock) {
 			Preconditions.checkState(!isClosed, "The NettyShuffleEnvironment has already been shut down.");
 
 			ResultPartition[] resultPartitions = new ResultPartition[resultPartitionDeploymentDescriptors.size()];
 			int counter = 0;
 			for (ResultPartitionDeploymentDescriptor rpdd : resultPartitionDeploymentDescriptors) {
-				resultPartitions[counter++] = resultPartitionFactory.create(ownerName, executionAttemptID, rpdd);
+				resultPartitions[counter++] = resultPartitionFactory.create(
+					ownerContext.getOwnerName(),
+					ownerContext.getExecutionAttemptID(),
+					rpdd);
 			}
 
-			registerOutputMetrics(outputGroup, buffersGroup, resultPartitions);
-			return Arrays.asList(resultPartitions);
+			MetricGroup outputMetricGroup = ownerContext.getOutputGroup();
+			registerOutputMetrics(outputMetricGroup, outputMetricGroup.addGroup(METRIC_GROUP_BUFFERS), resultPartitions);
+			return  Arrays.asList(resultPartitions);
 		}
 	}
 
 	@Override
 	public Collection<SingleInputGate> createInputGates(
-			String ownerName,
-			ExecutionAttemptID executionAttemptID,
+			ShuffleIOOwnerContext ownerContext,
 			PartitionProducerStateProvider partitionProducerStateProvider,
-			Collection<InputGateDeploymentDescriptor> inputGateDeploymentDescriptors,
-			MetricGroup parentGroup,
-			MetricGroup inputGroup,
-			MetricGroup buffersGroup) {
+			Collection<InputGateDeploymentDescriptor> inputGateDeploymentDescriptors) {
 		synchronized (lock) {
 			Preconditions.checkState(!isClosed, "The NettyShuffleEnvironment has already been shut down.");
 
-			InputChannelMetrics inputChannelMetrics = new InputChannelMetrics(parentGroup);
+			MetricGroup networkInputGroup = ownerContext.getInputGroup();
+			@SuppressWarnings("deprecation")
+			InputChannelMetrics inputChannelMetrics = new InputChannelMetrics(networkInputGroup, ownerContext.getParentGroup());
+
 			SingleInputGate[] inputGates = new SingleInputGate[inputGateDeploymentDescriptors.size()];
 			int counter = 0;
 			for (InputGateDeploymentDescriptor igdd : inputGateDeploymentDescriptors) {
 				SingleInputGate inputGate = singleInputGateFactory.create(
-					ownerName,
+					ownerContext.getOwnerName(),
 					igdd,
 					partitionProducerStateProvider,
 					inputChannelMetrics);
-				InputGateID id = new InputGateID(igdd.getConsumedResultId(), executionAttemptID);
+				InputGateID id = new InputGateID(igdd.getConsumedResultId(), ownerContext.getExecutionAttemptID());
 				inputGatesById.put(id, inputGate);
 				inputGate.getCloseFuture().thenRun(() -> inputGatesById.remove(id));
 				inputGates[counter++] = inputGate;
 			}
 
-			registerInputMetrics(inputGroup, buffersGroup, inputGates);
+			registerInputMetrics(networkInputGroup, networkInputGroup.addGroup(METRIC_GROUP_BUFFERS), inputGates);
 			return Arrays.asList(inputGates);
 		}
 	}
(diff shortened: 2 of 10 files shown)


azagrebin:
The new metric group layout is still to discuss.
We also need to agree on how we deprecate the old groups in docs.

azagrebin:
@flinkbot attention @zentol @zhijiangW

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/Task.java
```diff
@@ -388,10 +378,12 @@ public Task(
 			this,
 			inputGateDeploymentDescriptors,
 			metrics.getIOMetricGroup(),
-			inputGroup,
-			buffersGroup,
 			metrics.getIOMetricGroup().getNumBytesInCounter());
 
+		// we will have to check type of shuffle service later whether it is NetworkEnvironment
+		//noinspection deprecation
```
zentol: inconsistent with other parts of the PR that use `@SuppressWarnings("deprecation")`

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/metrics/InputChannelMetrics.java
```diff
@@ -32,44 +32,79 @@
...
+		private final InputChannelMetrics legacyMetrics;
+
+		public InputChannelMetricsWithLegacy(MetricGroup parent, MetricGroup legacyParent) {
+			super(parent);
+			legacyMetrics = new InputChannelMetrics(legacyParent);
+		}
+
+		@Override
+		public void incNumBytesInLocalCounter(long inc) {
+			super.incNumBytesInLocalCounter(inc);
```
zentol: it's quite easy to encapsulate this logic in a `Counter` implementation, which would require significantly less overall changes and would no longer introduce a new special case on how metrics are being used.

```
public static class CounterWrapper implements Counter {
	private final Counter counter1;
	private final Counter counter2;

	private CounterWrapper(Counter counter1, Counter counter2) {
		this.counter1 = counter1;
		this.counter2 = counter2;
	}

	@Override
	public void inc() {
		counter1.inc();
		counter2.inc();
	}

	@Override
	public void inc(long n) {
		counter1.inc(n);
		counter2.inc(n);

	}

	@Override
	public void dec() {
		counter1.dec();
		counter2.dec();
	}

	@Override
	public void dec(long n) {
		counter1.dec(n);
		counter2.dec(n);
	}

	@Override
	public long getCount() {
		// assume that the counters are not accessed directly elsewhere
		return counter1.getCount();
	}
}
```

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/Task.java
azagrebin: it does not work for statements, at least in Idea. that is why Idea has this special extra comment, I guess.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/metrics/InputChannelMetrics.java
azagrebin: true, switched to the counter wrapper

azagrebin:
Thanks @zentol ! addressed comments

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/metrics/InputChannelMetrics.java
```diff
@@ -72,4 +72,47 @@ public Counter getNumBuffersInLocalCounter() {
...
+		@Override
+		public void dec(long n) {
+			for (Counter c : counters) {
+				c.dec(n);
+			}
+		}
+
+		@Override
+		public long getCount() {
+			// assume that the counters are not accessed directly elsewhere
```
zhijiangW: I am not sure whether `getCounter` would be actually used. If used we should keep the return counter as previous structure. That means calling `new InputChannelMetrics(parentGroup, networkGroup) instead?

zentol:
> And if we think there are no concerns for creating one metric group twice

There are concerns; it prints a warning to the user as it is not how the API is supposed to be used. If any components need access to a shared group then this group should be created once and passed around as needed.
Things were implemented this way on purpose to prevent components from interfering with each other by accident.

zhijiangW:
Thanks for the replies @zentol .

But in this PR `parentGroup.addGroup("Network")` is called twice in `NetworkEnvironment#createResultPartitionWriters/InputGates`. So we should change to create it still in task class and then pass it into `createResultPartitionWriters/InputGates` separately?

azagrebin:
True, thanks for noticing it @zhijiangW, it was not an intent. I will change it to create group only once. It can be called e.g. `NetworkInput`/`NetworkOutput`.

azagrebin:
@zentol
Here are the moved metrics:

|      Deprecated||||           New ||||
|:-|:-|:-|:-|:-|:-|:-:|:-|
|**Type**|**Scope**|**Infix**|**Name**|**Type**|**Scope**|**Infix**|**Name**|
|IO|Task|-|numBytesInLocal|NettyShuffle|Task|NettyShuffleInput|numBytesInLocal|
|IO|Task|-|numBytesInLocalPerSecond|NettyShuffle|Task|NettyShuffleInput|numBytesInLocalPerSecond|
|IO|Task|-|numBytesInRemote|NettyShuffle|Task|NettyShuffleInput|numBytesInRemote|
|IO|Task|-|numBytesInRemotePerSecond|NettyShuffle|Task|NettyShuffleInput|numBytesInRemotePerSecond|
|IO|Task|-|numBuffersInLocal|NettyShuffle|Task|NettyShuffleInput|numBuffersInLocal|
|IO|Task|-|numBuffersInLocalPerSecond|NettyShuffle|Task|NettyShuffleInput|numBuffersInLocalPerSecond|
|IO|Task|-|numBuffersInRemote|NettyShuffle|Task|NettyShuffleInput|numBuffersInRemote|
|IO|Task|-|numBuffersInRemotePerSecond|NettyShuffle|Task|NettyShuffleInput|numBuffersInRemotePerSecond|
|Network|Task|Network.<Input\|Output>.<gate\|partition>|totalQueueLen|NettyShuffle|Task|<NettyShuffleInput\|NettyShuffleOutput>.<gate\|partition>| totalQueueLen |
|Network|Task|Network.<Input\|Output>.<gate\|partition>| minQueueLen |NettyShuffle|Task|<NettyShuffleInput\|NettyShuffleOutput>.<gate\|partition>| minQueueLen |
|Network|Task|Network.<Input\|Output>.<gate\|partition>| maxQueueLen |NettyShuffle|Task|<NettyShuffleInput\|NettyShuffleOutput>.<gate\|partition>| maxQueueLen |
|Network|Task|Network.<Input\|Output>.<gate\|partition>| avgQueueLen |NettyShuffle|Task|<NettyShuffleInput\|NettyShuffleOutput>.<gate\|partition>| avgQueueLen |
|Network|Task| buffers | inputQueueLength |NettyShuffle|Task|NettyShuffleInput.buffers| inputQueueLength |
|Network|Task| buffers | outputQueueLength |NettyShuffle|Task|NettyShuffleOutput.buffers| outputQueueLength |
|Network|Task| buffers | inPoolUsage |NettyShuffle|Task|NettyShuffleInput.buffers| inPoolUsage |
|Network|Task| buffers | outPoolUsage |NettyShuffle|Task|NettyShuffleOutput.buffers| outPoolUsage |

zentol:
I'd prefer if the infix were `Shuffle.Netty.Input`/`Shuffle.Netty.Output`, so far we've followed a hierarchical style similarly to out config options.

The PR also needs a rebase.

docs/monitoring/metrics.md
```diff
@@ -1236,42 +1351,42 @@ Certain RocksDB native metrics are available but disabled by default, you can fi
     <tr>
       <th rowspan="12"><strong>Task</strong></th>
       <td>numBytesInLocal</td>
-      <td>The total number of bytes this task has read from a local source.</td>
+      <td>The total number of bytes this task has read from a local source. Deprecated: see NettyShuffle.</td>
```
zentol: let's remove the description entirely and put a nice `attention` icon before the deprecation notice.

docs/monitoring/metrics.md
```diff
@@ -1071,6 +1072,120 @@ Thus, in order to infer the metric identifier:
   </tbody>
 </table>
 
+### NettyShuffle
```
zentol: do users actually know what NettyShuffle is?

azagrebin: True users do not know about `NettyShuffle`, we will still have only one  at the moment.
It could be `Default network shuffle` then.

azagrebin:
@zentol  
about infix, we want to avoid multiple metric group creations. Creating outer groups `Shuffle` and `Netty` firstly for partitions then for gates would duplicate it or we have to have execution id indexing of created metrics so far which I would try to avoid.

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/Task.java
```diff
@@ -359,21 +359,14 @@ public Task(
...
 		final String taskNameWithSubtaskAndId = taskNameWithSubtask + " (" + executionId + ')';
 
-		// add metrics for buffers
-		final MetricGroup buffersGroup = metrics.getIOMetricGroup().addGroup("buffers");
-
-		// similar to MetricUtils.instantiateNetworkMetrics() but inside this IOMetricGroup
-		final MetricGroup networkGroup = metrics.getIOMetricGroup().addGroup("Network");
-		final MetricGroup outputGroup = networkGroup.addGroup("Output");
-		final MetricGroup inputGroup = networkGroup.addGroup("Input");
+		MetricGroup taskShuffleMetricGroup = networkEnvironment.createTaskMetricGroup(metrics.getIOMetricGroup());
```
zhijiangW: Does it mean `ShuffleEnvironment` interface need provide the method `createTaskMetricGroup`?

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NetworkEnvironment.java
```diff
@@ -228,7 +236,8 @@ public void releasePartitions(Collection<ResultPartitionID> partitionIds) {
 				resultPartitions[counter++] = resultPartitionFactory.create(taskName, executionId, rpdd);
 			}
 
-			registerOutputMetrics(outputGroup, buffersGroup, resultPartitions);
+			MetricGroup outputMetricGroup = taskMetricGroup.addGroup(METRIC_GROUP_OUTPUT);
+			registerOutputMetrics(outputMetricGroup, outputMetricGroup.addGroup(METRIC_GROUP_BUFFERS), resultPartitions);
```
zhijiangW: Maybe we could only pass `outputMetricGroup` here and make `outputMetricGroup.addGroup(METRIC_GROUP_BUFFERS)` inside `registerOutputMetrics`.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NetworkEnvironment.java
```diff
@@ -259,11 +269,15 @@ public void releasePartitions(Collection<ResultPartitionID> partitionIds) {
 				inputGates[counter++] = inputGate;
 			}
 
-			registerInputMetrics(inputGroup, buffersGroup, inputGates);
+			registerInputMetrics(networkInputGroup, networkInputGroup.addGroup(METRIC_GROUP_BUFFERS), inputGates);
```
zhijiangW: ditto: create buffers group inside `registerInputMetrics`

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NetworkEnvironment.java
```diff
@@ -280,6 +294,31 @@ private void registerInputMetrics(MetricGroup inputGroup, MetricGroup buffersGro
...
+		MetricGroup metricGroup,
+		ResultPartitionWriter[] producedPartitions,
+		InputGate[] inputGates) {
+		// add metrics for buffers
+		final MetricGroup buffersGroup = metricGroup.addGroup("buffers");
+
+		// similar to MetricUtils.instantiateNetworkMetrics() but inside this IOMetricGroup
+		final MetricGroup networkGroup = metricGroup.addGroup("Network");
+		final MetricGroup outputGroup = networkGroup.addGroup("Output");
+		final MetricGroup inputGroup = networkGroup.addGroup("Input");
```
zhijiangW: Use `METRIC_GROUP_OUTPUT` and `METRIC_GROUP_INPUT` instead of `Output` and `Input`, also might define a class var for `Network`.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NetworkEnvironment.java
```diff
@@ -69,7 +70,13 @@
 
 	private static final Logger LOG = LoggerFactory.getLogger(NetworkEnvironment.class);
 
-	private static final String METRIC_GROUP_NETWORK = "Network";
+	@SuppressWarnings("DeprecatedIsStillUsed")
+	@Deprecated private static final String METRIC_GROUP_NETWORK_DEPRECATED = "Network";
```
zhijiangW: Put `@Deprecated` in separate line?

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NetworkEnvironment.java
```diff
@@ -69,7 +70,13 @@
...
+	private static final String METRIC_GROUP_SHUFFLE = "Shuffle";
+	private static final String METRIC_GROUP_NETTY = "Netty";
+	private static final String METRIC_GROUP_OUTPUT = "Output";
+	private static final String METRIC_GROUP_INPUT = "Input";
+	private static final String METRIC_GROUP_BUFFERS = "Buffers";
```
zhijiangW: There are already too many metrics here and it seems hard to distinguish them if not focusing on the below implementations. Could we give some comments or other forms to clearly know the architecture of these metrics? 
At least to distinguish the task-level metrics and global metrics.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NetworkEnvironment.java
```diff
@@ -280,6 +294,31 @@ private void registerInputMetrics(MetricGroup inputGroup, MetricGroup buffersGro
...
+	@SuppressWarnings("DeprecatedIsStillUsed")
+	@Deprecated
+	public void registerLegacyNetworkMetrics(
+		MetricGroup metricGroup,
+		ResultPartitionWriter[] producedPartitions,
+		InputGate[] inputGates) {
+		// add metrics for buffers
+		final MetricGroup buffersGroup = metricGroup.addGroup("buffers");
+
+		// similar to MetricUtils.instantiateNetworkMetrics() but inside this IOMetricGroup
```
zhijiangW: It seems we could not see `IOMetricGroup` here, might need adjust the annotation.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NetworkEnvironment.java
```diff
@@ -280,6 +294,31 @@ private void registerInputMetrics(MetricGroup inputGroup, MetricGroup buffersGro
...
+	 * <p>Registers legacy metric groups if shuffle service implementation is original default one.
+	 *
+	 * @deprecated should be removed in future
+	 */
+	@SuppressWarnings("DeprecatedIsStillUsed")
+	@Deprecated
+	public void registerLegacyNetworkMetrics(
+		MetricGroup metricGroup,
+		ResultPartitionWriter[] producedPartitions,
+		InputGate[] inputGates) {
```
zhijiangW: One more indentation for parameters to keep inconsistent with other methods

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NetworkEnvironment.java
```diff
@@ -160,10 +169,10 @@ public static NetworkEnvironment create(
 			singleInputGateFactory);
 	}
 
-	private static void registerNetworkMetrics(MetricGroup metricGroup, NetworkBufferPool networkBufferPool) {
+	private static void registerNetworkMetrics(String groupName, MetricGroup metricGroup, NetworkBufferPool networkBufferPool) {
```
zhijiangW: It might be better to put this method together with other metric related methods below, avoid dispersed anywhere.

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/Task.java
```diff
@@ -388,16 +381,18 @@ public Task(
...
 
 		this.inputGates = new InputGate[gates.length];
 		int counter = 0;
 		for (InputGate gate : gates) {
 			inputGates[counter++] = new InputGateWithMetrics(gate, metrics.getIOMetricGroup().getNumBytesInCounter());
 		}
 
+		// we will have to check type of shuffle service later whether it is NetworkEnvironment
+		//noinspection deprecation
+		networkEnvironment.registerLegacyNetworkMetrics(metrics.getIOMetricGroup(), resultPartitionWriters, gates);
```
zhijiangW: ditto: it seems a bit trouble after task only sees `ShuffleEnvironment` future, otherwise we need expose this method in interface.

zhijiangW:
@azagrebin thanks for the updates and I left some other small comments.

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/Task.java
azagrebin: My idea was to expose `TaskShuffleContext ShuffleEnvironment.createTaskShuffleContext(taskName, execId, parentMetricGroup)` in `ShuffleEnvironment `. `TaskShuffleContext` would contain  taskName, execId and created in/output metric group as well. `createPartition/Gate` methods would take the `TaskShuffleContext`. This way `ShuffleEnvironment ` can create its own metric group once and other task common parameters are encapsulated.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NetworkEnvironment.java
azagrebin: the comment is still relevant but I will put `metricGroup` in parenthesis:
`.. inside this IOMetricGroup (metricGroup)`

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NetworkEnvironment.java
azagrebin: I would avoid history change in this commit, could be a followup to put all metric methods into a `NettyShuffleMetricFactory` with constants.

azagrebin:
Thanks for the review @zhijiangW, I have addressed comments

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NetworkEnvironment.java
zhijiangW: It is more clearly now 👍

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NetworkEnvironment.java
```diff
@@ -70,18 +70,35 @@
...
 	private static final String METRIC_GROUP_BUFFERS = "Buffers";
-	private static final String METRIC_TOTAL_MEMORY_SEGMENT = "TotalMemorySegments";
-	private static final String METRIC_AVAILABLE_MEMORY_SEGMENT = "AvailableMemorySegments";
+
+	// task level output metrics: Shuffle.Netty.Output.*
 
 	private static final String METRIC_OUTPUT_QUEUE_LENGTH = "outputQueueLength";
 	private static final String METRIC_OUTPUT_POOL_USAGE = "outPoolUsage";
+
+	// task level output metrics: Shuffle.Netty.Input.*
```
zhijiangW: output -> input

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/metrics/NettyShuffleMetricFactory.java
```diff
@@ -0,0 +1,148 @@
...
+	private static final String METRIC_INPUT_QUEUE_LENGTH = "inputQueueLength";
+	private static final String METRIC_INPUT_POOL_USAGE = "inPoolUsage";
+
+	private NettyShuffleMetricFactory() {
+	}
+
+	public static void registerShuffleMetrics(
+			MetricGroup metricGroup,
+			NetworkBufferPool networkBufferPool) {
+		checkNotNull(metricGroup);
```
zhijiangW: checkNotNull(networkBufferPool)?

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/metrics/NettyShuffleMetricFactory.java
```diff
@@ -0,0 +1,148 @@
...
+
+	private static final String METRIC_OUTPUT_QUEUE_LENGTH = "outputQueueLength";
+	private static final String METRIC_OUTPUT_POOL_USAGE = "outPoolUsage";
+
+	// task level output metrics: Shuffle.Netty.Input.*
+
+	private static final String METRIC_INPUT_QUEUE_LENGTH = "inputQueueLength";
+	private static final String METRIC_INPUT_POOL_USAGE = "inPoolUsage";
+
+	private NettyShuffleMetricFactory() {
```
zhijiangW: This constructor seems not necessary

zhijiangW:
Thanks for the updates @azagrebin ! 
I like the way of metric factory to make `NetworkEnvironment` simple. Only several nit comments.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/metrics/NettyShuffleMetricFactory.java
azagrebin: This is an indicator that it is a utility stateless class with only static methods which is not supposed to be instantiated.

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/Task.java
zentol: Since we intend to remove this in the next release It should be fine to use a cast here, so that we don't have to add such a short-lived method to the interface.

azagrebin: which is now happening after rebasing onto #8608

azagrebin:
Thanks for the reviews @zentol @zhijiangW, I've addressed comments
This PR is now based on #8608 which I think needs to be merged first.
While rebasing, I introduced a `ShuffleIOOwnerContext` which is now created by Task using `ShuffleEnvironment.createShuffleIOOwnerContext` before creating partitions/gates. `NettyShuffleEnvironment.createShuffleIOOwnerContext` also creates netty metric groups, only once.

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/Task.java
```diff
@@ -359,21 +360,13 @@ public Task(
...
 
-		// add metrics for buffers
-		final MetricGroup buffersGroup = metrics.getIOMetricGroup().addGroup("buffers");
-
-		// similar to MetricUtils.instantiateNetworkMetrics() but inside this IOMetricGroup
-		final MetricGroup networkGroup = metrics.getIOMetricGroup().addGroup("Network");
-		final MetricGroup outputGroup = networkGroup.addGroup("Output");
-		final MetricGroup inputGroup = networkGroup.addGroup("Input");
+		ShuffleIOOwnerContext taskShuffleContext = shuffleEnvironment
+			.createShuffleIOOwnerContext(taskNameWithSubtaskAndId, executionId, metrics.getIOMetricGroup());
```
zhijiangW: The formatting:
`shuffleEnvironment.createShuffleIOOwnerContext(
			taskNameWithSubtaskAndId,
			executionId,
			metrics.getIOMetricGroup());
`

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/Task.java
```diff
@@ -359,21 +360,13 @@ public Task(
...
 
 		// produced intermediate result partitions
 		final ResultPartitionWriter[] resultPartitionWriters = shuffleEnvironment.createResultPartitionWriters(
```
zhijiangW: remove final for `resultPartitionWriters` for consistent with other temp vars?

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/Task.java
```diff
@@ -359,21 +360,13 @@ public Task(
 
 		final String taskNameWithSubtaskAndId = taskNameWithSubtask + " (" + executionId + ')';
```
zhijiangW: Might remove this local var to use `taskNameWithSubtask + " (" + executionId + ')'` directly below.

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleEnvironment.java
```diff
@@ -92,25 +92,34 @@
 	 */
 	int start() throws IOException;
 
+	/**
+	 * Create a context of the shuffle input/output owner used to create partitions or gates belonging to the owner.
+	 *
+	 * <p>This method has to be called only once to avoid duplicated internal metric group registration.
+	 *
+	 * @param ownerName the owner name, used for logs
+	 * @param executionAttemptID execution attempt id of the producer or consumer
```
zhijiangW: of the producer or consumer -> of the owner?

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NettyShuffleEnvironment.java
```diff
@@ -228,91 +211,111 @@ public void releasePartitions(Collection<ResultPartitionID> partitionIds) {
...
+	public ShuffleIOOwnerContext createShuffleIOOwnerContext(
+			String ownerName,
+			ExecutionAttemptID executionAttemptID,
+			MetricGroup parentGroup) {
+		MetricGroup nettyGroup = createShuffleIOOwnerMetricGroup(checkNotNull(parentGroup));
+		return new ShuffleIOOwnerContext(
+			checkNotNull(ownerName),
+			checkNotNull(executionAttemptID),
+			parentGroup,
+			nettyGroup.addGroup(METRIC_GROUP_INPUT),
```
zhijiangW: It might be better to put these two `addGroup` covered by above `createShuffleIOOwnerMetricGroup` then return tuple2 to reference here because all the metrics related operations should be done inside `NettyShuffleMetricFactory`

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NettyShuffleEnvironment.java
```diff
@@ -228,91 +211,111 @@ public void releasePartitions(Collection<ResultPartitionID> partitionIds) {
...
+					rpdd);
 			}
 
-			registerOutputMetrics(outputGroup, buffersGroup, resultPartitions);
-			return resultPartitions;
+			MetricGroup outputMetricGroup = ownerContext.getOutputGroup();
+			registerOutputMetrics(
+				config.isNetworkDetailedMetrics(),
+				outputMetricGroup,
+				outputMetricGroup.addGroup(METRIC_GROUP_BUFFERS),
```
zhijiangW: Might be better to cover all the details in `NettyShuffleMetricFactory`. I mean the logic of `outputMetricGroup.addGroup(METRIC_GROUP_BUFFERS)` here.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NettyShuffleEnvironment.java
```diff
@@ -228,91 +211,111 @@ public void releasePartitions(Collection<ResultPartitionID> partitionIds) {
...
-			MetricGroup inputGroup,
-			MetricGroup buffersGroup) {
+			Collection<InputGateDeploymentDescriptor> inputGateDeploymentDescriptors) {
 		synchronized (lock) {
-			Preconditions.checkState(!isShutdown, "The NetworkEnvironment has already been shut down.");
+			Preconditions.checkState(!isClosed, "The NettyShuffleEnvironment has already been shut down.");
+
+			MetricGroup networkInputGroup = ownerContext.getInputGroup();
+			@SuppressWarnings("deprecation")
+			InputChannelMetrics inputChannelMetrics = new InputChannelMetrics(networkInputGroup, ownerContext.getParentGroup());
```
zhijiangW: Maybe we could delay the creation of `InputChannelMetrics` in the `SingleInputGateFactory#createInputChannels`, then the following would be 
`SingleInputGate inputGate = singleInputGateFactory.create(ownerContext, icdd, partitionProducerStateProvider)` instead. 
To do so we could make the current `createInputGates` simple and make all the things done when required.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NettyShuffleEnvironment.java
```diff
@@ -228,91 +211,111 @@ public void releasePartitions(Collection<ResultPartitionID> partitionIds) {
...
 				inputGate.getCloseFuture().thenRun(() -> inputGatesById.remove(id));
 				inputGates[counter++] = inputGate;
 			}
 
-			registerInputMetrics(inputGroup, buffersGroup, inputGates);
-			return inputGates;
+			registerInputMetrics(
+				config.isNetworkDetailedMetrics(),
+				networkInputGroup,
+				networkInputGroup.addGroup(METRIC_GROUP_BUFFERS),
```
zhijiangW: ditto: for the issue of `networkInputGroup.addGroup(METRIC_GROUP_BUFFERS)`

zhijiangW:
Thanks for the updates @azagrebin .
I left some final nit comments relevant with new introduced context.

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/Task.java
azagrebin: In this case we do not break the arg list, the whole function is on the new line, same as chained calls.

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/Task.java
azagrebin: this file actually looks like more using finals atm.

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/Task.java
azagrebin: I would keep it to not mix concerns in one line.

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleEnvironment.java
azagrebin: At the moment we have exec id only for tasks/executions, in this context producer or consumer. I thought about owner as a potentially any shuffle IO component within task/execution, possibly even source/sink

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NettyShuffleEnvironment.java
azagrebin: I thought about it, I think the operation is too primitive atm for an extra utility method. Besides, I think that in general `tuple2` is very adhoc and not so readable.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NettyShuffleEnvironment.java
azagrebin: The gate creation loop will create `InputChannelMetrics` multiple times, I would avoid this because `InputChannelMetrics` are shared by gates at the moment.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NettyShuffleEnvironment.java
azagrebin: I can add overloaded methods for this in `NettyShuffleMetricFactory`.

azagrebin:
Thanks @zhijiangW , I have addressed comments

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/Task.java
zhijiangW: But the vars of `taskShuffleContext` and following `gates` are not defined as final. ATM only `taskNameWithSubtaskAndId` and `resultPartitionWriters` are final. Or we could make the new added `taskShuffleContext` as final to keep consistency.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NettyShuffleEnvironment.java
zhijiangW: Actually I also do not like the way of tuple. 
But we provide the method of `createShuffleIOOwnerMetricGroup` in factory, and it is only used for adding group simple. So it might bring confused that which `addGroup` should be done inside factory and which should be done outside. The scope/rule seems not clear.

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/Task.java
zhijiangW: I could agree your way to keep it. :)

My previous concern is that `taskNameWithSubtaskAndId` was used in many places before, so it worths defining a local var. Now it is only used in one place and we could reduce one line to make task constructor short to remove it.
Another small consideration is avoiding final issue as above mentioned.

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleEnvironment.java
```diff
@@ -0,0 +1,167 @@
...
+	 * @param parentGroup parent of shuffle specific metric group
+	 * @return context of the shuffle input/output owner used to create partitions or gates belonging to the owner
+	 */
+	ShuffleIOOwnerContext createShuffleIOOwnerContext(
```
zentol: it would be useful the partition lifecycle management if this method would also accept the JobID..

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleEnvironment.java
```diff
@@ -0,0 +1,167 @@
...
+	 * Factory method for the {@link ResultPartitionWriter ResultPartitionWriters} to produce result partitions.
+	 *
+	 * <p>The order of the {@link ResultPartitionWriter ResultPartitionWriters} in the returned collection
+	 * should be the same as the iteration order of the passed {@code resultPartitionDeploymentDescriptors}.
+	 *
+	 * @param ownerContext the owner context relevant for partition creation
+	 * @param resultPartitionDeploymentDescriptors descriptors of the partition, produced by the owner
+	 * @return collection of the {@link ResultPartitionWriter ResultPartitionWriters}
+	 */
+	Collection<P> createResultPartitionWriters(
```
zentol: have you considered moving this method and `createInputGates` into the `ShuffleIOOwnerContext`, and only pass this into the task instead of the entire `ShuffleEnvironment`? The task only requires access to these 2 methods, while any other users of the `ShuffleEnvironment` doesn't need these.

azagrebin: This is an interesting idea, we can consider it as a follow-up issue.

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleEnvironment.java
azagrebin: Shuffle service does not need it at the moment, would be nice actually to keep it decoupled from the job id, I would suggest we discuss it at the corresponding partition lifecycle PR and add it if needed.

azagrebin:
Rebased on #8680

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleEnvironment.java
zentol: see this discussion then: https://github.com/apache/flink/pull/8687#discussion_r292370552

PR #8486, opened 2019-05-20
[FLINK-12372] [runtime] implement ExecutionSlotAllocator

shuai-xu:

*This pull request introduces the ExecutionSlotAllocator interface related classes and a default implementation. The implementation of ExecutionSlotAllocator will first calculate the preferred locations for the executions and then allocate slots for them. In the future, the calculation of preferred locations will be removed from here to make the ExecutionSlotAllocator more clean.*

## Verifying this change

  - *Added unit tests in DefaultExecutionSlotAllocatorTest*

  - The serializers: (no)
  - The S3 file system connector: ( no)

## Documentation

### DIFF
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/ExecutionVertex.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/ExecutionVertex.java
@@ -84,7 +84,7 @@ public class ExecutionVertex implements AccessExecutionVertex, Archiveable<Archi
 
 	private static final Logger LOG = ExecutionGraph.LOG;
 
-	private static final int MAX_DISTINCT_LOCATIONS_TO_CONSIDER = 8;
+	public static final int MAX_DISTINCT_LOCATIONS_TO_CONSIDER = 8;
 
 	// --------------------------------------------------------------------------------------------
 
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/ExecutionVertex.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/ExecutionVertex.java
@@ -86,7 +86,6 @@ public class ExecutionVertex implements AccessExecutionVertex, Archiveable<Archi
 
 	public static final int MAX_DISTINCT_LOCATIONS_TO_CONSIDER = 8;
 
-	public static final int ABC = 1;
 	// --------------------------------------------------------------------------------------------
 
 	private final ExecutionJobVertex jobVertex;
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/ExecutionVertex.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/ExecutionVertex.java
@@ -86,6 +86,7 @@ public class ExecutionVertex implements AccessExecutionVertex, Archiveable<Archi
 
 	public static final int MAX_DISTINCT_LOCATIONS_TO_CONSIDER = 8;
 
+	public static final int ABC = 1;
 	// --------------------------------------------------------------------------------------------
 
 	private final ExecutionJobVertex jobVertex;
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocator.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocator.java
@@ -1,0 +1,273 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.runtime.scheduler;
+
+import org.apache.flink.annotation.VisibleForTesting;
+import org.apache.flink.runtime.clusterframework.types.AllocationID;
+import org.apache.flink.runtime.clusterframework.types.SlotProfile;
+import org.apache.flink.runtime.concurrent.FutureUtils;
+import org.apache.flink.runtime.execution.ExecutionState;
+import org.apache.flink.runtime.executiongraph.Execution;
+import org.apache.flink.runtime.executiongraph.ExecutionEdge;
+import org.apache.flink.runtime.executiongraph.ExecutionGraph;
+import org.apache.flink.runtime.executiongraph.ExecutionJobVertex;
+import org.apache.flink.runtime.executiongraph.ExecutionVertex;
+import org.apache.flink.runtime.instance.SlotSharingGroupId;
+import org.apache.flink.runtime.jobmanager.scheduler.ScheduledUnit;
+import org.apache.flink.runtime.jobmaster.LogicalSlot;
+import org.apache.flink.runtime.jobmaster.SlotRequestId;
+import org.apache.flink.runtime.jobmaster.slotpool.Scheduler;
+import org.apache.flink.runtime.jobmaster.slotpool.SlotProvider;
+import org.apache.flink.runtime.scheduler.strategy.ExecutionVertexID;
+import org.apache.flink.runtime.taskmanager.TaskManagerLocation;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.CompletableFuture;
+
+import static org.apache.flink.runtime.executiongraph.ExecutionVertex.MAX_DISTINCT_LOCATIONS_TO_CONSIDER;
+import static org.apache.flink.util.Preconditions.checkNotNull;
+
+/**
+ * Default {@link ExecutionSlotAllocator} which will use {@link SlotProvider} to allocate slots and
+ * keep the unfulfilled requests for further cancellation.
+ */
+public class DefaultExecutionSlotAllocator implements ExecutionSlotAllocator {
+
+	static final Logger LOG = LoggerFactory.getLogger(DefaultExecutionSlotAllocator.class);
+
+	/**
+	 * Store the uncompleted slot assignments.
+	 */
+	private final Map<ExecutionVertexID, SlotExecutionVertexAssignment> pendingSlotAssignments;
+
+	private final SlotProvider slotProvider;
+
+	private final ExecutionGraph executionGraph;
+
+	public DefaultExecutionSlotAllocator(SlotProvider slotProvider, ExecutionGraph executionGraph) {
+		this.slotProvider = checkNotNull(slotProvider);
+		this.executionGraph = checkNotNull(executionGraph);
+
+		pendingSlotAssignments = new HashMap<>();
+	}
+
+	@Override
+	public Collection<SlotExecutionVertexAssignment> allocateSlotsFor(
+			Collection<ExecutionVertexSchedulingRequirements> executionVertexSchedulingRequirements) {
+
+		List<SlotExecutionVertexAssignment> slotExecutionVertexAssignments =
+				new ArrayList<>(executionVertexSchedulingRequirements.size());
+
+		Set<AllocationID> allPreviousAllocationIds =
+				computeAllPriorAllocationIdsIfRequiredByScheduling(executionVertexSchedulingRequirements);
+
+		Set<ExecutionVertexID> verticesScheduledTogether = getAllExecutionVerticesId(executionVertexSchedulingRequirements);
+
+		for (ExecutionVertexSchedulingRequirements schedulingRequirements : executionVertexSchedulingRequirements) {
+			final ExecutionVertexID executionVertexId = schedulingRequirements.getExecutionVertexId();
+			final SlotRequestId slotRequestId = new SlotRequestId();
+			final SlotSharingGroupId slotSharingGroupId = schedulingRequirements.getSlotSharingGroupId();
+
+			Execution execution = null;
+			ExecutionJobVertex ejv = executionGraph.getJobVertex(executionVertexId.getJobVertexId());
+			if (ejv != null && ejv.getParallelism() > executionVertexId.getSubtaskIndex()) {
+				execution = ejv.getTaskVertices()[executionVertexId.getSubtaskIndex()].getCurrentExecutionAttempt();
+
+			}
+			if (execution == null) {
+				throw new IllegalArgumentException("Fail to get execution for vertex id " + executionVertexId);
+			}
+
+			if (LOG.isDebugEnabled()) {
+				LOG.debug("Allocate slot with id {} for execution {}", slotRequestId, executionVertexId);
+			}
+
+			// TODO: the calculation of preferred location should be refined so we can use the preferredLocations
+			// in ExecutionVertexSchedulingRequirements instead of calculating it here
+			CompletableFuture<LogicalSlot> slotFuture =
+					calculatePreferredLocations(execution, verticesScheduledTogether).thenCompose(
+							(Collection<TaskManagerLocation> preferredLocations) ->
+								slotProvider.allocateSlot(
+									slotRequestId,
+									new ScheduledUnit(
+											null,
+											executionVertexId.getJobVertexId(),
+											slotSharingGroupId,
+											schedulingRequirements.getCoLocationConstraint()),
+									new SlotProfile(
+											schedulingRequirements.getResourceProfile(),
+											preferredLocations,
+											Arrays.asList(schedulingRequirements.getPreviousAllocationId()),
+											allPreviousAllocationIds),
+									executionGraph.isQueuedSchedulingAllowed(),
+									executionGraph.getAllocationTimeout()));
+
+			SlotExecutionVertexAssignment slotExecutionVertexAssignment =
+					new SlotExecutionVertexAssignment(executionVertexId, slotFuture);
+			// add to map first to avoid the future completed before added.
+			pendingSlotAssignments.put(executionVertexId, slotExecutionVertexAssignment);
+
+			slotFuture.whenComplete(
+					(ignored, throwable) -> {
+						pendingSlotAssignments.remove(executionVertexId);
+						if (throwable != null) {
+							slotProvider.cancelSlotRequest(slotRequestId, slotSharingGroupId, throwable);
+						}
+					});
+
+			slotExecutionVertexAssignments.add(slotExecutionVertexAssignment);
+		}
+
+		return slotExecutionVertexAssignments;
+	}
+
+	@Override
+	public void cancel(ExecutionVertexID executionVertexId) {
+		SlotExecutionVertexAssignment slotExecutionVertexAssignment = pendingSlotAssignments.get(executionVertexId);
+		if (slotExecutionVertexAssignment != null) {
+			slotExecutionVertexAssignment.getLogicalSlotFuture().cancel(false);
+		}
+	}
+
+	@Override
+	public CompletableFuture<Void> stop() {
+		List<ExecutionVertexID> executionVertexIds = new ArrayList<>(pendingSlotAssignments.keySet());
+		executionVertexIds.stream().forEach(executionVertexID -> cancel(executionVertexID));
+
+		return CompletableFuture.completedFuture(null);
+	}
+
+	private static Set<ExecutionVertexID> getAllExecutionVerticesId(
+			Collection<ExecutionVertexSchedulingRequirements> executionVertexSchedulingRequirements) {
+		Set<ExecutionVertexID> executionVertexIds = new HashSet<>(executionVertexSchedulingRequirements.size());
+		for (ExecutionVertexSchedulingRequirements schedulingRequirements : executionVertexSchedulingRequirements) {
+			executionVertexIds.add(schedulingRequirements.getExecutionVertexId());
+		}
+		return executionVertexIds;
+	}
+
+	/**
+	 * Calculates the preferred locations for an execution.
+	 * It will first try to use preferred locations based on state,
+	 * if null, will use the preferred locations based on inputs.
+	 */
+	private static CompletableFuture<Collection<TaskManagerLocation>> calculatePreferredLocations(
+			Execution execution,
+			Collection<ExecutionVertexID> verticesScheduledTogether) {
+
+		Collection<CompletableFuture<TaskManagerLocation>> preferredLocations =
+				execution.getVertex().getPreferredLocationsBasedOnState();
+
+		if (preferredLocations == null) {
+			preferredLocations = getPreferredLocationsBasedOnInputs(execution, verticesScheduledTogether);
+		}
+		return FutureUtils.combineAll(preferredLocations);
+	}
+
+	/**
+	 * Gets the location preferences of the execution, as determined by the locations
+	 * of the predecessors from which it receives input data.
+	 * If there are more than {@link MAX_DISTINCT_LOCATIONS_TO_CONSIDER} different locations of source data,
+	 * or neither the sources have not been started nor will be started with the execution together,
+	 * this method returns an empty collection to indicate no location preference.
+	 *
+	 * @return The preferred locations based in input streams, or an empty iterable,
+	 *         if there is no input-based preference.
+	 */
+	@VisibleForTesting
+	static Collection<CompletableFuture<TaskManagerLocation>> getPreferredLocationsBasedOnInputs(
+			Execution execution,
+			Collection<ExecutionVertexID> verticesScheduledTogether) {
+		Collection<CompletableFuture<TaskManagerLocation>> preferredLocations = new HashSet<>();
+
+		Set<CompletableFuture<TaskManagerLocation>> inputLocations = new HashSet<>();
+		for (int i = 0; i < execution.getVertex().getNumberOfInputs(); i++) {
+			ExecutionEdge[] inputEdges = execution.getVertex().getInputEdges(i);
+			if (inputEdges.length <= MAX_DISTINCT_LOCATIONS_TO_CONSIDER) {
+				for (ExecutionEdge inputEdge : inputEdges) {
+					ExecutionVertex producer = inputEdge.getSource().getProducer();
+					if (verticesScheduledTogether.contains(
+							new ExecutionVertexID(producer.getJobvertexId(), producer.getParallelSubtaskIndex()))
+							|| producer.getExecutionState() != ExecutionState.CREATED) {
+						inputLocations.add(producer.getCurrentTaskManagerLocationFuture());
+					} else {
+						inputLocations.clear();
+						break;
+					}
+				}
+				if (preferredLocations.isEmpty() ||
+						(!inputLocations.isEmpty() && inputLocations.size() < preferredLocations.size())) {
+					preferredLocations.clear();
+					preferredLocations.addAll(inputLocations);
+				}
+				inputLocations.clear();
+			}
+		}
+		return preferredLocations;
+	}
+
+	/**
+	 * Returns the result of {@link #computeAllPriorAllocationIds}, but only if the scheduling really requires it.
+	 * Otherwise this method simply returns an empty set.
+	 */
+	private Set<AllocationID> computeAllPriorAllocationIdsIfRequiredByScheduling(
+			Collection<ExecutionVertexSchedulingRequirements> executionVertexSchedulingRequirements) {
+		// This is a temporary optimization to avoid computing all previous allocations if not required
+		// This can go away when we progress with the implementation of the Scheduler.
+		if (slotProvider instanceof Scheduler && ((Scheduler) slotProvider).requiresPreviousExecutionGraphAllocations()) {
+			return computeAllPriorAllocationIds(executionVertexSchedulingRequirements);
+		} else {
+			return Collections.emptySet();
+		}
+	}
+
+	/**
+	 * Computes and returns a set with the prior allocation ids from all execution vertices scheduled together.
+	 *
+	 * @param executionVertexSchedulingRequirements contains the execution vertices which are scheduled together
+	 */
+	@VisibleForTesting
+	static Set<AllocationID> computeAllPriorAllocationIds(
+			Collection<ExecutionVertexSchedulingRequirements> executionVertexSchedulingRequirements) {
+		HashSet<AllocationID> allPreviousAllocationIds = new HashSet<>(executionVertexSchedulingRequirements.size());
+		for (ExecutionVertexSchedulingRequirements schedulingRequirements : executionVertexSchedulingRequirements) {
+			AllocationID priorAllocation = schedulingRequirements.getPreviousAllocationId();
+			if (priorAllocation != null) {
+				allPreviousAllocationIds.add(priorAllocation);
+			}
+		}
+		return allPreviousAllocationIds;
+	}
+
+	@VisibleForTesting
+	Map<ExecutionVertexID, SlotExecutionVertexAssignment> getPendingSlotAssignments() {
+		return Collections.unmodifiableMap(pendingSlotAssignments);
+	}
+}
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocator.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocator.java
@@ -104,7 +104,6 @@ public class DefaultExecutionSlotAllocator implements ExecutionSlotAllocator {
 			}
 
 			// TODO: the calculation of preferred location should be refined.
-			// in ExecutionVertexSchedulingRequirements instead of calculating it here
 			CompletableFuture<LogicalSlot> slotFuture = calculatePreferredLocations(
 					executionVertexId,
 					schedulingRequirements.getPreferredLocations(),
@@ -217,20 +216,17 @@ public class DefaultExecutionSlotAllocator implements ExecutionSlotAllocator {
 				);
 				// If the parallelism is large, wait for all futures coming back may cost a long time,
 				// so using the unfulfilled future number here to speed up it.
-				System.out.println(inputLocations.size() + " while " + inputLocationsFutures.size());
 				if (inputLocations.size() > MAX_DISTINCT_LOCATIONS_TO_CONSIDER ||
 						inputLocationsFutures.size() > MAX_DISTINCT_LOCATIONS_TO_CONSIDER) {
 					locationsFutures.clear();
 					break;
 				}
 			}
-			System.out.println(executionVertexId + " should wait for " + locationsFutures);
 
 			if (!locationsFutures.isEmpty()) {
 				CompletableFuture<Collection<TaskManagerLocation>> uniqueLocationsFuture =
 						FutureUtils.combineAll(locationsFutures).thenApply(
 								(locations) -> {
-									System.out.println("I am completed with size " + locations.size());
 									Set<TaskManagerLocation> uniqueLocations = new HashSet<>(locations);
 									if (uniqueLocations.size() <= MAX_DISTINCT_LOCATIONS_TO_CONSIDER) {
 										return uniqueLocations;
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocator.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocator.java
@@ -23,6 +23,7 @@ import org.apache.flink.api.common.time.Time;
 import org.apache.flink.runtime.clusterframework.types.AllocationID;
 import org.apache.flink.runtime.clusterframework.types.SlotProfile;
 import org.apache.flink.runtime.concurrent.FutureUtils;
+import org.apache.flink.runtime.executiongraph.ExecutionVertex;
 import org.apache.flink.runtime.instance.SlotSharingGroupId;
 import org.apache.flink.runtime.jobmanager.scheduler.ScheduledUnit;
 import org.apache.flink.runtime.jobmaster.LogicalSlot;
@@ -171,7 +172,7 @@ public class DefaultExecutionSlotAllocator implements ExecutionSlotAllocator {
 	/**
 	 * Gets the location preferences of the execution, as determined by the locations
 	 * of the predecessors from which it receives input data.
-	 * If there are more than {@link MAX_DISTINCT_LOCATIONS_TO_CONSIDER} different locations of source data,
+	 * If there are more than {@link ExecutionVertex#MAX_DISTINCT_LOCATIONS_TO_CONSIDER} different locations of source data,
 	 * or neither the sources have not been started nor will be started with the execution together,
 	 * this method returns an empty collection to indicate no location preference.
 	 *
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocator.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocator.java
@@ -27,7 +27,6 @@ import org.apache.flink.runtime.instance.SlotSharingGroupId;
 import org.apache.flink.runtime.jobmanager.scheduler.ScheduledUnit;
 import org.apache.flink.runtime.jobmaster.LogicalSlot;
 import org.apache.flink.runtime.jobmaster.SlotRequestId;
-import org.apache.flink.runtime.jobmaster.slotpool.Scheduler;
 import org.apache.flink.runtime.jobmaster.slotpool.SlotProvider;
 import org.apache.flink.runtime.scheduler.strategy.ExecutionVertexID;
 import org.apache.flink.runtime.taskmanager.TaskManagerLocation;
@@ -48,6 +47,7 @@ import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 
 import static org.apache.flink.runtime.executiongraph.ExecutionVertex.MAX_DISTINCT_LOCATIONS_TO_CONSIDER;
+
 import static org.apache.flink.util.Preconditions.checkNotNull;
 
 /**
@@ -56,7 +56,7 @@ import static org.apache.flink.util.Preconditions.checkNotNull;
  */
 public class DefaultExecutionSlotAllocator implements ExecutionSlotAllocator {
 
-	static final Logger LOG = LoggerFactory.getLogger(DefaultExecutionSlotAllocator.class);
+	private static final Logger LOG = LoggerFactory.getLogger(DefaultExecutionSlotAllocator.class);
 
 	/**
 	 * Store the uncompleted slot assignments.
@@ -69,17 +69,13 @@ public class DefaultExecutionSlotAllocator implements ExecutionSlotAllocator {
 
 	private final Time allocationTimeout;
 
-	private final Boolean isQueuedSchedulingAllowed;
-
 	public DefaultExecutionSlotAllocator(
 			SlotProvider slotProvider,
 			InputsLocationsRetriever inputsLocationsRetriever,
-			Time allocationTimeout,
-			boolean isQueuedSchedulingAllowed) {
+			Time allocationTimeout) {
 		this.slotProvider = checkNotNull(slotProvider);
 		this.inputsLocationsRetriever = checkNotNull(inputsLocationsRetriever);
 		this.allocationTimeout = checkNotNull(allocationTimeout);
-		this.isQueuedSchedulingAllowed = isQueuedSchedulingAllowed;
 
 		pendingSlotAssignments = new HashMap<>();
 	}
@@ -91,19 +87,15 @@ public class DefaultExecutionSlotAllocator implements ExecutionSlotAllocator {
 		List<SlotExecutionVertexAssignment> slotExecutionVertexAssignments =
 				new ArrayList<>(executionVertexSchedulingRequirements.size());
 
-		Set<AllocationID> allPreviousAllocationIds =
-				computeAllPriorAllocationIdsIfRequiredByScheduling(executionVertexSchedulingRequirements);
+		Set<AllocationID> allPreviousAllocationIds = computeAllPriorAllocationIds(executionVertexSchedulingRequirements);
 
 		for (ExecutionVertexSchedulingRequirements schedulingRequirements : executionVertexSchedulingRequirements) {
 			final ExecutionVertexID executionVertexId = schedulingRequirements.getExecutionVertexId();
 			final SlotRequestId slotRequestId = new SlotRequestId();
 			final SlotSharingGroupId slotSharingGroupId = schedulingRequirements.getSlotSharingGroupId();
 
-			if (LOG.isDebugEnabled()) {
-				LOG.debug("Allocate slot with id {} for execution {}", slotRequestId, executionVertexId);
-			}
+			LOG.debug("Allocate slot with id {} for execution {}", slotRequestId, executionVertexId);
 
-			// TODO: the calculation of preferred location should be refined.
 			CompletableFuture<LogicalSlot> slotFuture = calculatePreferredLocations(
 					executionVertexId,
 					schedulingRequirements.getPreferredLocations(),
@@ -121,7 +113,7 @@ public class DefaultExecutionSlotAllocator implements ExecutionSlotAllocator {
 													preferredLocations,
 													Arrays.asList(schedulingRequirements.getPreviousAllocationId()),
 													allPreviousAllocationIds),
-											isQueuedSchedulingAllowed,
+											true,
 											allocationTimeout));
 
 			SlotExecutionVertexAssignment slotExecutionVertexAssignment =
@@ -190,83 +182,46 @@ public class DefaultExecutionSlotAllocator implements ExecutionSlotAllocator {
 	static CompletableFuture<Collection<TaskManagerLocation>> getPreferredLocationsBasedOnInputs(
 			ExecutionVertexID executionVertexId,
 			InputsLocationsRetriever inputsLocationsRetriever) {
-		CompletableFuture<Collection<TaskManagerLocation>> preferredLocations = null;
+		CompletableFuture<Collection<TaskManagerLocation>> preferredLocations =
+				CompletableFuture.completedFuture(Collections.emptyList());
 
 		Collection<CompletableFuture<TaskManagerLocation>> locationsFutures = new ArrayList<>();
 
 		Collection<Collection<ExecutionVertexID>> allProducers =
 				inputsLocationsRetriever.getConsumedResultPartitionsProducers(executionVertexId);
 		for (Collection<ExecutionVertexID> producers : allProducers) {
-			Set<TaskManagerLocation> inputLocations = new HashSet<>();
-			List<CompletableFuture<TaskManagerLocation>> inputLocationsFutures = new ArrayList<>();
 
 			for (ExecutionVertexID producer : producers) {
 				Optional<CompletableFuture<TaskManagerLocation>> optionalLocationFuture =
 						inputsLocationsRetriever.getTaskManagerLocation(producer);
 				optionalLocationFuture.ifPresent(
-						(locationFuture) -> {
-							locationsFutures.add(locationFuture);
-							TaskManagerLocation taskManagerLocation = locationFuture.getNow(null);
-							if (taskManagerLocation != null) {
-								inputLocations.add(taskManagerLocation);
-							} else {
-								inputLocationsFutures.add(locationFuture);
-							}
-						}
-				);
-				// If the parallelism is large, wait for all futures coming back may cost a long time,
-				// so using the unfulfilled future number here to speed up it.
-				if (inputLocations.size() > MAX_DISTINCT_LOCATIONS_TO_CONSIDER ||
-						inputLocationsFutures.size() > MAX_DISTINCT_LOCATIONS_TO_CONSIDER) {
+						(locationFuture) -> locationsFutures.add(locationFuture));
+				// If the parallelism is large, wait for all futures coming back may cost a long time.
+				if (locationsFutures.size() > MAX_DISTINCT_LOCATIONS_TO_CONSIDER) {
 					locationsFutures.clear();
 					break;
 				}
 			}
 
-			if (!locationsFutures.isEmpty()) {
-				CompletableFuture<Collection<TaskManagerLocation>> uniqueLocationsFuture =
-						FutureUtils.combineAll(locationsFutures).thenApply(
-								(locations) -> {
-									Set<TaskManagerLocation> uniqueLocations = new HashSet<>(locations);
-									if (uniqueLocations.size() <= MAX_DISTINCT_LOCATIONS_TO_CONSIDER) {
-										return uniqueLocations;
-									} else {
-										return Collections.emptySet();
-									}
-								});
-				if (preferredLocations == null) {
-					preferredLocations = uniqueLocationsFuture;
-				} else {
-					preferredLocations = preferredLocations.thenCombine(
-							uniqueLocationsFuture,
-							(locationsOnOneEdge, locationsOnAnotherEdge) -> {
-								if ((!locationsOnOneEdge.isEmpty() && locationsOnAnotherEdge.size() > locationsOnOneEdge.size())
-										|| locationsOnAnotherEdge.isEmpty()) {
-									return locationsOnOneEdge;
-								} else {
-									return locationsOnAnotherEdge;
-								}
+			CompletableFuture<Collection<TaskManagerLocation>> uniqueLocationsFuture =
+					FutureUtils.combineAll(locationsFutures).thenApply(
+							(locations) -> {
+								Set<TaskManagerLocation> uniqueLocations = new HashSet<>(locations);
+								return uniqueLocations;
 							});
-				}
-				locationsFutures.clear();
-			}
+			preferredLocations = preferredLocations.thenCombine(
+					uniqueLocationsFuture,
+					(locationsOnOneEdge, locationsOnAnotherEdge) -> {
+						if ((!locationsOnOneEdge.isEmpty() && locationsOnAnotherEdge.size() > locationsOnOneEdge.size())
+								|| locationsOnAnotherEdge.isEmpty()) {
+							return locationsOnOneEdge;
+						} else {
+							return locationsOnAnotherEdge;
+						}
+					});
+			locationsFutures.clear();
 		}
-		return preferredLocations == null ? CompletableFuture.completedFuture(Collections.emptyList()) : preferredLocations;
-	}
-
-	/**
-	 * Returns the result of {@link #computeAllPriorAllocationIds}, but only if the scheduling really requires it.
-	 * Otherwise this method simply returns an empty set.
-	 */
-	private Set<AllocationID> computeAllPriorAllocationIdsIfRequiredByScheduling(
-			Collection<ExecutionVertexSchedulingRequirements> executionVertexSchedulingRequirements) {
-		// This is a temporary optimization to avoid computing all previous allocations if not required
-		// This can go away when we progress with the implementation of the Scheduler.
-		if (slotProvider instanceof Scheduler && ((Scheduler) slotProvider).requiresPreviousExecutionGraphAllocations()) {
-			return computeAllPriorAllocationIds(executionVertexSchedulingRequirements);
-		} else {
-			return Collections.emptySet();
-		}
+		return preferredLocations;
 	}
 
 	/**
@@ -277,7 +232,7 @@ public class DefaultExecutionSlotAllocator implements ExecutionSlotAllocator {
 	@VisibleForTesting
 	static Set<AllocationID> computeAllPriorAllocationIds(
 			Collection<ExecutionVertexSchedulingRequirements> executionVertexSchedulingRequirements) {
-		HashSet<AllocationID> allPreviousAllocationIds = new HashSet<>(executionVertexSchedulingRequirements.size());
+		HashSet<AllocationID> allPreviousAllocationIds = new HashSet<>();
 		for (ExecutionVertexSchedulingRequirements schedulingRequirements : executionVertexSchedulingRequirements) {
 			AllocationID priorAllocation = schedulingRequirements.getPreviousAllocationId();
 			if (priorAllocation != null) {
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocator.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocator.java
@@ -48,7 +48,6 @@ import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 
 import static org.apache.flink.runtime.executiongraph.ExecutionVertex.MAX_DISTINCT_LOCATIONS_TO_CONSIDER;
-
 import static org.apache.flink.util.Preconditions.checkNotNull;
 
 /**
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocator.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocator.java
@@ -19,15 +19,10 @@
 package org.apache.flink.runtime.scheduler;
 
 import org.apache.flink.annotation.VisibleForTesting;
+import org.apache.flink.api.common.time.Time;
 import org.apache.flink.runtime.clusterframework.types.AllocationID;
 import org.apache.flink.runtime.clusterframework.types.SlotProfile;
 import org.apache.flink.runtime.concurrent.FutureUtils;
-import org.apache.flink.runtime.execution.ExecutionState;
-import org.apache.flink.runtime.executiongraph.Execution;
-import org.apache.flink.runtime.executiongraph.ExecutionEdge;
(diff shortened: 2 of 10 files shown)


flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocator.java
```diff
@@ -0,0 +1,273 @@
...
+		Collection<CompletableFuture<TaskManagerLocation>> preferredLocations = new HashSet<>();
+
+		Set<CompletableFuture<TaskManagerLocation>> inputLocations = new HashSet<>();
+		for (int i = 0; i < execution.getVertex().getNumberOfInputs(); i++) {
+			ExecutionEdge[] inputEdges = execution.getVertex().getInputEdges(i);
+			if (inputEdges.length <= MAX_DISTINCT_LOCATIONS_TO_CONSIDER) {
+				for (ExecutionEdge inputEdge : inputEdges) {
+					ExecutionVertex producer = inputEdge.getSource().getProducer();
+					if (verticesScheduledTogether.contains(
+							new ExecutionVertexID(producer.getJobvertexId(), producer.getParallelSubtaskIndex()))
```
shuai-xu: Adding getId() to ExecutionVertex would make it more convenient.

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocator.java
```diff
@@ -0,0 +1,273 @@
...
+		return executionVertexIds;
+	}
+
+	/**
+	 * Calculates the preferred locations for an execution.
+	 * It will first try to use preferred locations based on state,
+	 * if null, will use the preferred locations based on inputs.
+	 */
+	private static CompletableFuture<Collection<TaskManagerLocation>> calculatePreferredLocations(
+			Execution execution,
```
GJL: In the original design we are aiming at not having direct dependencies on the `ExecutionGraph` from anywhere but the Scheduler itself:

`getPreferredLocationsBasedOnState()`: If I am not mistaken this can be passed via `ExecutionVertexSchedulingRequirements`

_number of input sources_ & _number of `InputEdges`_: We can pass something similar to a `SchedulingTopology` to the `ExecutionSlotAllocator` so that these information can be inferred. 

What do you think?

(this is not a review)

shuai-xu: +1 for getting rid of ExecutionGraph in ExecutionSlotAllocator. I had to add it to ExecutionSlotAllocator since we can not getCurrentTaskManagerLocationFuture from SchedulingTopology, could we add this method to SchedulingExecutionVertex?

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocator.java
```diff
@@ -0,0 +1,273 @@
...
+	@VisibleForTesting
+	static Collection<CompletableFuture<TaskManagerLocation>> getPreferredLocationsBasedOnInputs(
+			Execution execution,
+			Collection<ExecutionVertexID> verticesScheduledTogether) {
+		Collection<CompletableFuture<TaskManagerLocation>> preferredLocations = new HashSet<>();
+
+		Set<CompletableFuture<TaskManagerLocation>> inputLocations = new HashSet<>();
+		for (int i = 0; i < execution.getVertex().getNumberOfInputs(); i++) {
+			ExecutionEdge[] inputEdges = execution.getVertex().getInputEdges(i);
+			if (inputEdges.length <= MAX_DISTINCT_LOCATIONS_TO_CONSIDER) {
```
tillrohrmann: I think this part diverges from the existing code. In the existing code we disregard an input if its deployed to more than `MAX_DISTINCT_LOCATIONS_TO_CONSIDER` `TaskManagerLocations` but not if its parallelism is larger than that. Multiple sub tasks might still be deployed to the same `TaskExecutor`.

shuai-xu: Hi @tillrohrmann , the existing code adds taskManagerLocationFuture in each Execution to a set and if the size of the set is larger than MAX_DISTINCT_LOCATIONS_TO_CONSIDER, it will ignore it. As the taskManagerLocationFuture in each Execution is different, I think it is same to just judging the parallelism of the JobVertex.

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocator.java
shuai-xu: Another problem is that now when calling getConsumedResultPartitions on SchedulingExecutionVertex, we get a Collection<SchedulingResultPartition>, it can't identify how many JobEdges they belong to easily. The existing code will ignore the JobEdge which has too many different location of ExecutionEdge. So could we change Collection<SchedulingResultPartition> getConsumedResultPartitions to Collection<Collection<SchedulingResultPartition>> getConsumedResultPartitions  or that we don't distinguish different JobEdges ?

GJL: Let's think first which operations we need to implement the status quo:

```
    // gets predecessors (can group by JobVertexID to find predecessors per job vertex)
	Collection<ExecutionVertexID> getConsumedResultPartitionsProducers(ExecutionVertexID executionVertexId);

	// to lookup predecessor TM locations
	Optional<TaskManagerLocation> getTaskManagerLocation(ExecutionVertexID executionVertexId);

	// to filter by >= MAX_DISTINCT_LOCATIONS_TO_CONSIDER
	int getJobVertexParallelism(JobVertexID jobVertexId);
```

Did I miss something?

shuai-xu: It is decided by the logic we choose for the calculation. Maybe need to get the status of upstream. I create a [doc](https://docs.google.com/document/d/120L_u3C-PGa4jbiEexpFccVToZMowEI7xN-5FMhReZo/edit?usp=sharing) to record some proposal for it.

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocator.java
```diff
@@ -0,0 +1,294 @@
...
+					break;
+				}
+			}
+
+			if (!locationsFutures.isEmpty()) {
+				CompletableFuture<Collection<TaskManagerLocation>> uniqueLocationsFuture =
+						FutureUtils.combineAll(locationsFutures).thenApply(
+								(locations) -> {
+									Set<TaskManagerLocation> uniqueLocations = new HashSet<>(locations);
+									if (uniqueLocations.size() <= MAX_DISTINCT_LOCATIONS_TO_CONSIDER) {
```
GJL: If we do not change the legacy behavior (discard location futures if the number of futures is too high), this condition should never be true. Please correct me if I am wrong.

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocatorTest.java
```diff
@@ -0,0 +1,355 @@
...
+
+		Set<AllocationID> allPriorAllocationIds = executionSlotAllocator.computeAllPriorAllocationIds(testSchedulingRequirements);
+		assertThat(allPriorAllocationIds, containsInAnyOrder(expectAllocationIds.toArray()));
+	}
+
+	/**
+	 * Tests the calculation of preferred locations based on inputs for an execution.
+	 */
+	@Test
+	public void testGetPreferredLocationsBasedOnInputs() throws Exception {
```
GJL: This test is too big and violates the rule of having only a _single concept per test_ (see Clean Code by Robert C. Martin, Chapter 9, pp. 130, 131), i.e., this should be broken down to multiple tests. I would even prefer to move these into a separate class since we are testing a static method on `DefaultExecutionSlotAllocator` and not an instance of it.

I think there are (at least) two test cases needed:
- test that we do not wait for the futures if we have more than MAX_DISTINCT_LOCATIONS_TO_CONSIDER futures
- test that we prefer the TM locations of the predecessor with fewest distinct TM locations

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/ExecutionVertexSchedulingRequirements.java
```diff
@@ -0,0 +1,88 @@
...
+
+	private final ResourceProfile resourceProfile;
+
+	private final SlotSharingGroupId slotSharingGroupId;
+
+	private final CoLocationConstraint coLocationConstraint;
+
+	private final Collection<TaskManagerLocation> preferredLocations;
+
+	public ExecutionVertexSchedulingRequirements(
```
GJL: A constructor this big would benefit from the builder pattern, e.g.,
```
	/**
	 * Builder for {@link ExecutionVertexSchedulingRequirements}.
	 */
	public static class Builder {

		private ExecutionVertexID executionVertexId;

		private AllocationID previousAllocationId;

		private ResourceProfile resourceProfile = ResourceProfile.UNKNOWN;

		private SlotSharingGroupId slotSharingGroupId;

		private CoLocationConstraint coLocationConstraint;

		private Collection<TaskManagerLocation> preferredLocations = Collections.emptyList();

		public Builder withExecutionVertexId(final ExecutionVertexID executionVertexId) {
			this.executionVertexId = executionVertexId;
			return this;
		}

		public Builder withPreviousAllocationId(final AllocationID previousAllocationId) {
			this.previousAllocationId = previousAllocationId;
			return this;
		}

		public Builder withResourceProfile(final ResourceProfile resourceProfile) {
			this.resourceProfile = resourceProfile;
			return this;
		}

		public Builder withSlotSharingGroupId(final SlotSharingGroupId slotSharingGroupId) {
			this.slotSharingGroupId = slotSharingGroupId;
			return this;
		}

		public Builder withCoLocationConstraint(final CoLocationConstraint coLocationConstraint) {
			this.coLocationConstraint = coLocationConstraint;
			return this;
		}

		public Builder withPreferredLocations(final Collection<TaskManagerLocation> preferredLocations) {
			this.preferredLocations = preferredLocations;
			return this;
		}

		public ExecutionVertexSchedulingRequirements build() {
			return new ExecutionVertexSchedulingRequirements(
				executionVertexId,
				previousAllocationId,
				resourceProfile,
				slotSharingGroupId,
				coLocationConstraint,
				preferredLocations);
		}
	}
```

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocator.java
```diff
@@ -0,0 +1,294 @@
...
+									return locationsOnOneEdge;
+								} else {
+									return locationsOnAnotherEdge;
+								}
+							});
+				}
+				locationsFutures.clear();
+			}
+		}
+		return preferredLocations == null ? CompletableFuture.completedFuture(Collections.emptyList()) : preferredLocations;
```
GJL: Null check wouldn't be needed, if you initialized `preferredLocations` as follows
```
CompletableFuture<Collection<TaskManagerLocation>> preferredLocations = CompletableFuture.completedFuture(Collections.emptyList());
```

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocator.java
```diff
@@ -0,0 +1,294 @@
...
+						FutureUtils.combineAll(locationsFutures).thenApply(
+								(locations) -> {
+									Set<TaskManagerLocation> uniqueLocations = new HashSet<>(locations);
+									if (uniqueLocations.size() <= MAX_DISTINCT_LOCATIONS_TO_CONSIDER) {
+										return uniqueLocations;
+									} else {
+										return Collections.emptySet();
+									}
+								});
+				if (preferredLocations == null) {
```
GJL: null check not needed if you initialize `preferredLocations`

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocator.java
```diff
@@ -0,0 +1,294 @@
...
+				// If the parallelism is large, wait for all futures coming back may cost a long time,
+				// so using the unfulfilled future number here to speed up it.
+				if (inputLocations.size() > MAX_DISTINCT_LOCATIONS_TO_CONSIDER ||
+						inputLocationsFutures.size() > MAX_DISTINCT_LOCATIONS_TO_CONSIDER) {
+					locationsFutures.clear();
+					break;
+				}
+			}
+
+			if (!locationsFutures.isEmpty()) {
```
GJL: Logic would work without the empty check. If this is for performance optimization, I'd say it's not needed.

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocatorTest.java
```diff
@@ -0,0 +1,355 @@
...
+		JobVertex producer1 = new JobVertex("producer1");
+		producer1.setInvokableClass(NoOpInvokable.class);
```
GJL: Setting the invokable class is not needed and distracts from the actual test logic.

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocatorTest.java
```diff
@@ -0,0 +1,355 @@
...
+			executionVertexIds.add(executionVertexId);
+			slotFutures.add(new CompletableFuture<>());
+		}
+	}
+
+	/**
+	 * Tests that it will allocate slots from slot provider and remove the slot assignments when request are fulfilled.
+	 */
+	@Test
+	public void testAllocateSlotsFor() {
```
GJL: Here are some comments/suggestions:
- `TestingInputsLocationsRetriever` has a dependency to `JobVertex` – I don't think this dependency is justified. 
- Job graph always has a parallelism of 3. However, a parallelism higher than 1 is not necessary relevant for all tests.
- It is implicitly encoded in `createSimpleInputsLocationsRetriever()` that the first half of the execution vertices are producers (the other half are consumers). This makes the tests and the assertions hard to read because I have to first read `createSimpleInputsLocationsRetriever()` in `setUp()` to understand what the test asserts.
- The tests are not asserting the contract of `ExecutionSlotAllocator`: 
  - Tests should assert on the public methods of the interface, i.e., we should assert against the return type of `allocateSlotsFor` (`Collection<SlotExecutionVertexAssignment>`)
  - `@VisibleForTesting` is a code smell
- test name should describe what is being tested, e.g., the expected behavior

Find below a draft of how the tests could look like. Feel free to adapt.

```
	@Test
	public void producersAreAssignedToSlotsBeforeConsumers() {
		final ExecutionVertexID producerId = new ExecutionVertexID(new JobVertexID(), 0);
		final ExecutionVertexID consumerId = new ExecutionVertexID(new JobVertexID(), 0);

		final TestingInputLocationRetriever inputLocationRetriever = new TestingInputLocationRetriever.Builder()
			.connectConsumerToProducer(consumerId, producerId)
			.build();

		final DefaultExecutionSlotAllocator executionSlotAllocator = createExecutionSlotAllocator(inputLocationRetriever);

		inputLocationRetriever.markScheduled(producerId);
		inputLocationRetriever.markScheduled(consumerId);

		final List<ExecutionVertexSchedulingRequirements> schedulingRequirements = createSchedulingRequirements(producerId, consumerId);
		final Collection<SlotExecutionVertexAssignment> slotExecutionVertexAssignments = executionSlotAllocator.allocateSlotsFor(schedulingRequirements);

		final SlotExecutionVertexAssignment producerSlotAssignment = findSlotAssignmentByExecutionVertexId(producerId, slotExecutionVertexAssignments);
		final SlotExecutionVertexAssignment consumerSlotAssignment = findSlotAssignmentByExecutionVertexId(consumerId, slotExecutionVertexAssignments);

		assertTrue(producerSlotAssignment.getLogicalSlotFuture().isDone());
		assertFalse(consumerSlotAssignment.getLogicalSlotFuture().isDone());
	}

	@Test
	public void consumersAreAssignedToSlotsAfterProducers() {
		final ExecutionVertexID producerId = new ExecutionVertexID(new JobVertexID(), 0);
		final ExecutionVertexID consumerId = new ExecutionVertexID(new JobVertexID(), 0);

		final TestingInputLocationRetriever inputLocationRetriever = new TestingInputLocationRetriever.Builder()
			.connectConsumerToProducer(consumerId, producerId)
			.build();

		final DefaultExecutionSlotAllocator executionSlotAllocator = createExecutionSlotAllocator(inputLocationRetriever);

		inputLocationRetriever.markScheduled(producerId);
		inputLocationRetriever.markScheduled(consumerId);

		final List<ExecutionVertexSchedulingRequirements> schedulingRequirements = createSchedulingRequirements(producerId, consumerId);
		final Collection<SlotExecutionVertexAssignment> slotExecutionVertexAssignments = executionSlotAllocator.allocateSlotsFor(schedulingRequirements);
		inputLocationRetriever.assignTaskManagerLocation(producerId);

		final SlotExecutionVertexAssignment consumerSlotAssignment = findSlotAssignmentByExecutionVertexId(consumerId, slotExecutionVertexAssignments);
		assertTrue(consumerSlotAssignment.getLogicalSlotFuture().isDone());
	}
```


```
```
	class TestingInputLocationRetriever implements InputsLocationsRetriever {

		private final Map<ExecutionVertexID, List<ExecutionVertexID>> producersByVertex;

		private final Map<ExecutionVertexID, CompletableFuture<TaskManagerLocation>> taskManagerLocationsByVertex = new HashMap<>();

		TestingInputLocationRetriever(final Map<ExecutionVertexID, List<ExecutionVertexID>> producersByVertex) {
			this.producersByVertex = new HashMap<>(producersByVertex);
		}

		@Override
		public Collection<Collection<ExecutionVertexID>> getConsumedResultPartitionsProducers(final ExecutionVertexID executionVertexId) {
			final Map<JobVertexID, List<ExecutionVertexID>> executionVerticesByJobVertex =
				producersByVertex.getOrDefault(executionVertexId, Collections.emptyList())
					.stream()
					.collect(Collectors.groupingBy(ExecutionVertexID::getJobVertexId));

			return new ArrayList<>(executionVerticesByJobVertex.values());
		}

		@Override
		public Optional<CompletableFuture<TaskManagerLocation>> getTaskManagerLocation(final ExecutionVertexID executionVertexId) {
			return Optional.ofNullable(taskManagerLocationsByVertex.get(executionVertexId));
		}

		public void markScheduled(final ExecutionVertexID executionVertexId) {
			taskManagerLocationsByVertex.put(executionVertexId, new CompletableFuture<>());
		}

		public void assignTaskManagerLocation(final ExecutionVertexID executionVertexId) {
			taskManagerLocationsByVertex.compute(executionVertexId, (key, future) -> {
				if (future == null) {
					return CompletableFuture.completedFuture(new LocalTaskManagerLocation());
				}
				future.complete(new LocalTaskManagerLocation());
				return future;
			});
		}

		static class Builder {

			private final Map<ExecutionVertexID, List<ExecutionVertexID>> consumerToProducers = new HashMap<>();

			public Builder connectConsumerToProducer(final ExecutionVertexID consumer, final ExecutionVertexID producer) {
				consumerToProducers.compute(consumer, (key, producers) -> {
					if (producers == null) {
						producers = new ArrayList<>();
					}
					producers.add(producer);
					return producers;
				});
				return this;
			}

			public TestingInputLocationRetriever build() {
				return new TestingInputLocationRetriever(consumerToProducers);
			}

		}
	}

```

**Comment by GJL** [comment]
Thanks for your contribution to Apache Flink, @shuai-xu! I have left some comments that we should address prior to merging.

File: flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocator.java [hunk 12]
```diff
@@ -0,0 +1,294 @@
...
+
+	/**
+	 * Returns the result of {@link #computeAllPriorAllocationIds}, but only if the scheduling really requires it.
+	 * Otherwise this method simply returns an empty set.
+	 */
+	private Set<AllocationID> computeAllPriorAllocationIdsIfRequiredByScheduling(
+			Collection<ExecutionVertexSchedulingRequirements> executionVertexSchedulingRequirements) {
+		// This is a temporary optimization to avoid computing all previous allocations if not required
+		// This can go away when we progress with the implementation of the Scheduler.
+		if (slotProvider instanceof Scheduler && ((Scheduler) slotProvider).requiresPreviousExecutionGraphAllocations()) {
```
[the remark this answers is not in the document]
shuai-xu: it is just an optimization, we can remove it.

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/ExecutionVertexSchedulingRequirements.java
```diff
@@ -0,0 +1,88 @@
...
+		this.slotSharingGroupId = slotSharingGroupId;
+		this.coLocationConstraint = coLocationConstraint;
+		this.preferredLocations = preferredLocations;
+	}
+
+	public ExecutionVertexID getExecutionVertexId() {
+		return executionVertexId;
+	}
+
+	public AllocationID getPreviousAllocationId() {
```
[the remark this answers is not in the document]
shuai-xu: Done

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocator.java
```diff
@@ -0,0 +1,294 @@
...
+import static org.apache.flink.runtime.executiongraph.ExecutionVertex.MAX_DISTINCT_LOCATIONS_TO_CONSIDER;
+import static org.apache.flink.util.Preconditions.checkNotNull;
+
+/**
+ * Default {@link ExecutionSlotAllocator} which will use {@link SlotProvider} to allocate slots and
+ * keep the unfulfilled requests for further cancellation.
+ */
+public class DefaultExecutionSlotAllocator implements ExecutionSlotAllocator {
+
+	static final Logger LOG = LoggerFactory.getLogger(DefaultExecutionSlotAllocator.class);
```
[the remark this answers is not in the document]
shuai-xu: Done

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocator.java
```diff
@@ -0,0 +1,294 @@
...
+		Set<AllocationID> allPreviousAllocationIds =
+				computeAllPriorAllocationIdsIfRequiredByScheduling(executionVertexSchedulingRequirements);
+
+		for (ExecutionVertexSchedulingRequirements schedulingRequirements : executionVertexSchedulingRequirements) {
+			final ExecutionVertexID executionVertexId = schedulingRequirements.getExecutionVertexId();
+			final SlotRequestId slotRequestId = new SlotRequestId();
+			final SlotSharingGroupId slotSharingGroupId = schedulingRequirements.getSlotSharingGroupId();
+
+			if (LOG.isDebugEnabled()) {
+				LOG.debug("Allocate slot with id {} for execution {}", slotRequestId, executionVertexId);
```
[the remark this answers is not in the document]
shuai-xu: Done

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocator.java
```diff
@@ -0,0 +1,294 @@
...
+	 */
+	private final Map<ExecutionVertexID, SlotExecutionVertexAssignment> pendingSlotAssignments;
+
+	private final SlotProvider slotProvider;
+
+	private final InputsLocationsRetriever inputsLocationsRetriever;
+
+	private final Time allocationTimeout;
+
+	private final Boolean isQueuedSchedulingAllowed;
```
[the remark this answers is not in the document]
shuai-xu: Done

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocator.java
```diff
@@ -0,0 +1,294 @@
...
+		for (ExecutionVertexSchedulingRequirements schedulingRequirements : executionVertexSchedulingRequirements) {
+			final ExecutionVertexID executionVertexId = schedulingRequirements.getExecutionVertexId();
+			final SlotRequestId slotRequestId = new SlotRequestId();
+			final SlotSharingGroupId slotSharingGroupId = schedulingRequirements.getSlotSharingGroupId();
+
+			if (LOG.isDebugEnabled()) {
+				LOG.debug("Allocate slot with id {} for execution {}", slotRequestId, executionVertexId);
+			}
+
+			// TODO: the calculation of preferred location should be refined.
```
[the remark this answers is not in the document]
shuai-xu: OK

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/ExecutionVertexSchedulingRequirements.java
```diff
@@ -0,0 +1,88 @@
...
+
+	private final Collection<TaskManagerLocation> preferredLocations;
+
+	public ExecutionVertexSchedulingRequirements(
+			ExecutionVertexID executionVertexId,
+			AllocationID previousAllocationId,
+			ResourceProfile resourceProfile,
+			SlotSharingGroupId slotSharingGroupId,
+			CoLocationConstraint coLocationConstraint,
+			Collection<TaskManagerLocation> preferredLocations) {
```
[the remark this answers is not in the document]
shuai-xu: OK

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocator.java
shuai-xu: Done

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocator.java
shuai-xu: Done

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocator.java
shuai-xu: OK

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocator.java
shuai-xu: Done, keep it as legacy code

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocator.java
```diff
@@ -0,0 +1,294 @@
...
+							if (taskManagerLocation != null) {
+								inputLocations.add(taskManagerLocation);
+							} else {
+								inputLocationsFutures.add(locationFuture);
+							}
+						}
+				);
+				// If the parallelism is large, wait for all futures coming back may cost a long time,
+				// so using the unfulfilled future number here to speed up it.
+				if (inputLocations.size() > MAX_DISTINCT_LOCATIONS_TO_CONSIDER ||
```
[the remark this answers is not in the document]
shuai-xu: Done, keep it as legacy code

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocator.java
```diff
@@ -0,0 +1,294 @@
...
+
+	/**
+	 * Computes and returns a set with the prior allocation ids from all execution vertices scheduled together.
+	 *
+	 * @param executionVertexSchedulingRequirements contains the execution vertices which are scheduled together
+	 */
+	@VisibleForTesting
+	static Set<AllocationID> computeAllPriorAllocationIds(
+			Collection<ExecutionVertexSchedulingRequirements> executionVertexSchedulingRequirements) {
+		HashSet<AllocationID> allPreviousAllocationIds = new HashSet<>(executionVertexSchedulingRequirements.size());
```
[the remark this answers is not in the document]
shuai-xu: Got it.

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocatorTest.java
shuai-xu: Good idea

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/ExecutionVertexSchedulingRequirements.java
shuai-xu: Done

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocatorTest.java
shuai-xu: I Compared using JobVertex with ExecutionVertexID, and chose JobVertex considering it's more flexible for a complex topology. However, your suggestion is reasonable.

I appreciate the principle "One Assert per Test", but it may be a little difficult to execute because we want to check more to make sure the class works well. So I adopt the "Single Concept per Test".

I don't like VisibleForTesting too. But or else we have to implement a TestingSlotProvider to check the all previous allocation ids.

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocator.java
```diff
@@ -0,0 +1,250 @@
...
+	 * Calculates the preferred locations for an execution.
+	 * It will first try to use preferred locations based on state,
+	 * if null, will use the preferred locations based on inputs.
+	 */
+	private static CompletableFuture<Collection<TaskManagerLocation>> calculatePreferredLocations(
+			ExecutionVertexID executionVertexId,
+			Collection<TaskManagerLocation> preferredLocationsBasedOnState,
+			InputsLocationsRetriever inputsLocationsRetriever) {
+
+		if (preferredLocationsBasedOnState != null && !preferredLocationsBasedOnState.isEmpty()) {
```
GJL: I think `preferredLocationsBasedOnState` cannot be null anymore.

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocator.java
```diff
@@ -0,0 +1,250 @@
...
+			LOG.debug("Allocate slot with id {} for execution {}", slotRequestId, executionVertexId);
+
+			CompletableFuture<LogicalSlot> slotFuture = calculatePreferredLocations(
+					executionVertexId,
+					schedulingRequirements.getPreferredLocations(),
+					inputsLocationsRetriever).thenCompose(
+							(Collection<TaskManagerLocation> preferredLocations) ->
+									slotProvider.allocateSlot(
+											slotRequestId,
+											new ScheduledUnit(
```
GJL: There is a constructor that does not expect the `Execution`. You can use that one instead of passing `null`.

GJL: What would be the best way to test that we are creating the `ScheduledUnit` correctly? We can introduce a mapper (`ExecutionVertexSchedulingRequirements` -> `ScheduledUnit`) that we test separately, or intercept the objects in `TestingSlotProvider`. Same problem holds for `SlotProfile`.

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocatorTest.java
```diff
@@ -0,0 +1,257 @@
...
+
+		inputsLocationsRetriever.markScheduled(producerId);
+
+		final List<ExecutionVertexSchedulingRequirements> schedulingRequirements =
+				createSchedulingRequirements(producerId, consumerId);
+		executionSlotAllocator.allocateSlotsFor(schedulingRequirements);
+
+		executionSlotAllocator.stop().getNow(null);
+
+		assertThat(cancelledSlotRequestIds, hasSize(1));
```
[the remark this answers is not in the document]
shuai-xu: The size of receivedSlotRequestIds is not a const, so I think it's more safe to check the number.

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocatorTest.java
```diff
@@ -0,0 +1,257 @@
...
+
+		final DefaultExecutionSlotAllocator executionSlotAllocator = createExecutionSlotAllocator(inputsLocationsRetriever);
+
+		final List<ExecutionVertexSchedulingRequirements> schedulingRequirements =
+				createSchedulingRequirements(producerId);
+		executionSlotAllocator.allocateSlotsFor(schedulingRequirements);
+
+		executionSlotAllocator.cancel(producerId);
+
+		assertThat(cancelledSlotRequestIds, hasSize(1));
```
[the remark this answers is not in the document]
shuai-xu: I know it, as the size of receivedSlotRequestIds is not a const, so I think it's more safe to check the number.
For other place where the number of  expected results is known, I don't check the size.

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocator.java
```diff
@@ -0,0 +1,250 @@
...
+			AllocationID priorAllocation = schedulingRequirements.getPreviousAllocationId();
+			if (priorAllocation != null) {
+				allPreviousAllocationIds.add(priorAllocation);
+			}
+		}
+		return allPreviousAllocationIds;
+	}
+
+	@VisibleForTesting
+	Map<ExecutionVertexID, SlotExecutionVertexAssignment> getPendingSlotAssignments() {
```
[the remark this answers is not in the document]
shuai-xu: OK

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/ExecutionVertex.java
```diff
@@ -84,8 +84,9 @@
 
 	private static final Logger LOG = ExecutionGraph.LOG;
 
-	private static final int MAX_DISTINCT_LOCATIONS_TO_CONSIDER = 8;
+	public static final int MAX_DISTINCT_LOCATIONS_TO_CONSIDER = 8;
 
+	public static final int ABC = 1;
```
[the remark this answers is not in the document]
shuai-xu: Sorry, it was added by mistake.

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocatorTest.java
```diff
@@ -0,0 +1,257 @@
...
+						withPreviousAllocationId(expectAllocationIds.get(1)).
+						build(),
+				new ExecutionVertexSchedulingRequirements.Builder().
+						withExecutionVertexId(new ExecutionVertexID(new JobVertexID(), 3)).
+						build()
+		);
+
+		final TestingInputsLocationsRetriever inputsLocationsRetriever = new TestingInputsLocationsRetriever.Builder()
+				.build();
+		final DefaultExecutionSlotAllocator executionSlotAllocator = createExecutionSlotAllocator(inputsLocationsRetriever);
```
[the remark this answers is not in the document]
shuai-xu: Done

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocatorTest.java
```diff
@@ -0,0 +1,257 @@
...
+
+		Set<AllocationID> allPriorAllocationIds = executionSlotAllocator.computeAllPriorAllocationIds(testSchedulingRequirements);
+		assertThat(allPriorAllocationIds, containsInAnyOrder(expectAllocationIds.toArray()));
+	}
+
+	private DefaultExecutionSlotAllocator createExecutionSlotAllocator(InputsLocationsRetriever inputsLocationsRetriever) {
+
```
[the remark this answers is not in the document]
shuai-xu: OK

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocatorTest.java
```diff
@@ -0,0 +1,257 @@
...
+	private SlotExecutionVertexAssignment findSlotAssignmentByExecutionVertexId(
+			ExecutionVertexID executionVertexId,
+			Collection<SlotExecutionVertexAssignment> slotExecutionVertexAssignments) {
+		for (SlotExecutionVertexAssignment slotExecutionVertexAssignment : slotExecutionVertexAssignments) {
+			if (slotExecutionVertexAssignment.getExecutionVertexId().equals(executionVertexId)) {
+				return slotExecutionVertexAssignment;
+			}
+		}
+		return null;
```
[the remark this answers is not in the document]
shuai-xu: Done

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocatorTest.java
```diff
@@ -0,0 +1,257 @@
...
+	 */
+	@Test
+	public void testCancelUnFulfilledSlotRequest() {
+		slotFutures.add(new CompletableFuture<>());
+		final ExecutionVertexID producerId = new ExecutionVertexID(new JobVertexID(), 0);
+
+		final TestingInputsLocationsRetriever inputsLocationsRetriever = new TestingInputsLocationsRetriever.Builder()
+				.build();
+
+		final DefaultExecutionSlotAllocator executionSlotAllocator = createExecutionSlotAllocator(inputsLocationsRetriever);
```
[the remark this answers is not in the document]
shuai-xu: Done

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocatorTest.java
```diff
@@ -0,0 +1,257 @@
...
+		assertThat(cancelledSlotRequestIds, contains(receivedSlotRequestIds.toArray()));
+	}
+
+	/**
+	 * Tests that all unfulfilled slot requests will be cancelled when stopped.
+	 */
+	@Test
+	public void testStop() {
+		slotFutures.add(new CompletableFuture<>());
+		final ExecutionVertexID producerId = new ExecutionVertexID(new JobVertexID(), 0);
```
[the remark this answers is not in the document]
shuai-xu: OK

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocatorTest.java
```diff
@@ -0,0 +1,257 @@
...
+
+		final DefaultExecutionSlotAllocator executionSlotAllocator = createExecutionSlotAllocator(inputsLocationsRetriever);
+
+		inputsLocationsRetriever.markScheduled(producerId);
+
+		final List<ExecutionVertexSchedulingRequirements> schedulingRequirements =
+				createSchedulingRequirements(producerId, consumerId);
+		executionSlotAllocator.allocateSlotsFor(schedulingRequirements);
+
+		executionSlotAllocator.stop().getNow(null);
```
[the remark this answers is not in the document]
shuai-xu: Done, using get() need to declare `throws Exception` to the method.

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocatorTest.java
```diff
@@ -0,0 +1,257 @@
...
+	public void testCancelNonExistingExecutionVertex() {
+		final TestingInputsLocationsRetriever inputsLocationsRetriever = new TestingInputsLocationsRetriever.Builder()
+				.build();
+
+		final DefaultExecutionSlotAllocator executionSlotAllocator = createExecutionSlotAllocator(inputsLocationsRetriever);
+
+		ExecutionVertexID inValidExecutionVertexId = new ExecutionVertexID(new JobVertexID(), 0);
+		executionSlotAllocator.cancel(inValidExecutionVertexId);
+
+		assertThat(cancelledSlotRequestIds, hasSize(0));
```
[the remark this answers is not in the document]
shuai-xu: OK

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocator.java
shuai-xu: Done

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocator.java
shuai-xu: Done

shuai-xu: I think intercepting the objects in TestingSlotProvider would be better.

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/ExecutionVertexSchedulingRequirements.java
```diff
@@ -0,0 +1,157 @@
...
+
+	private final ResourceProfile resourceProfile;
+
+	private final SlotSharingGroupId slotSharingGroupId;
+
+	private final CoLocationConstraint coLocationConstraint;
+
+	private final Collection<TaskManagerLocation> preferredLocations;
+
+	public ExecutionVertexSchedulingRequirements(
```
GJL: Constructor should be now private since we have a builder.

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/PreferredLocationsCalculationTest.java
```diff
@@ -0,0 +1,109 @@
...
+		final ExecutionVertexID consumerId = new ExecutionVertexID(new JobVertexID(), 0);
+
+		TestingInputsLocationsRetriever.Builder locationRetrieverBuilder = new TestingInputsLocationsRetriever.Builder();
+		JobVertexID jobVertexID = new JobVertexID();
+		for (int i = 0; i < 9; i++) {
+			final ExecutionVertexID producerId = new ExecutionVertexID(jobVertexID, i);
+			locationRetrieverBuilder.connectConsumerToProducer(consumerId, producerId);
+		}
+
+		final TestingInputsLocationsRetriever inputsLocationsRetriever = locationRetrieverBuilder.build();
```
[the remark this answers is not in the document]
shuai-xu: Yes, the producers should be marked scheduled. But as we don't store all the vertex ids in TestingInputsLocationsRetriever, we can not have a method markAllVerticesScheduled, I mark all producers scheduled manually.

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/ExecutionVertexSchedulingRequirements.java
shuai-xu: OK

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/PreferredLocationsCalculationTest.java
```diff
@@ -0,0 +1,109 @@
...
+	/**
+	 * Tests that the input edge will be ignored if it has too many different locations.
+	 */
+	@Test
+	public void testIgnoreEdgeOfTooManyLocations() {
+		final ExecutionVertexID consumerId = new ExecutionVertexID(new JobVertexID(), 0);
+
+		TestingInputsLocationsRetriever.Builder locationRetrieverBuilder = new TestingInputsLocationsRetriever.Builder();
+		JobVertexID jobVertexID = new JobVertexID();
+		for (int i = 0; i < 9; i++) {
```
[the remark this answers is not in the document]
shuai-xu: Done

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/PreferredLocationsCalculationTest.java
```diff
@@ -0,0 +1,109 @@
...
+import java.util.concurrent.CompletableFuture;
+
+import static org.hamcrest.Matchers.containsInAnyOrder;
+import static org.hamcrest.Matchers.hasSize;
+import static org.junit.Assert.assertThat;
+
+/**
+ * Tests for the calculation of preferred locations based on inputs in {@link DefaultExecutionSlotAllocator}.
+ */
+public class PreferredLocationsCalculationTest extends TestLogger {
```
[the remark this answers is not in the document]
shuai-xu: Done

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/PreferredLocationsCalculationTest.java
```diff
@@ -0,0 +1,109 @@
...
+		}
+
+		for (int i = 0; i < parallelism2; i++) {
+			inputsLocationsRetriever.assignTaskManagerLocation(producers2.get(i));
+		}
+
+		CompletableFuture<Collection<TaskManagerLocation>> preferredLocations =
+				DefaultExecutionSlotAllocator.getPreferredLocationsBasedOnInputs(consumerId, inputsLocationsRetriever);
+
+		assertThat(preferredLocations.getNow(null), containsInAnyOrder(expectedLocations.toArray()));
```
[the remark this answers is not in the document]
shuai-xu: OK

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/PreferredLocationsCalculationTest.java
```diff
@@ -0,0 +1,109 @@
...
+			final ExecutionVertexID producerId = new ExecutionVertexID(jobVertexID, i);
+			locationRetrieverBuilder.connectConsumerToProducer(consumerId, producerId);
+		}
+
+		final TestingInputsLocationsRetriever inputsLocationsRetriever = locationRetrieverBuilder.build();
+
+		CompletableFuture<Collection<TaskManagerLocation>> preferredLocations =
+				DefaultExecutionSlotAllocator.getPreferredLocationsBasedOnInputs(consumerId, inputsLocationsRetriever);
+
+		assertThat(preferredLocations.getNow(null), hasSize(0));
```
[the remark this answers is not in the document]
shuai-xu: OK

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocatorTest.java
```diff
@@ -0,0 +1,332 @@
...
+		slotFutures.add(CompletableFuture.completedFuture(new TestingLogicalSlot()));
+		final CompletableFuture<ScheduledUnit> taskFuture = new CompletableFuture<>();
+		final CompletableFuture<SlotProfile> slotProfileFuture = new CompletableFuture<>();
+
+		slotProvider.setRequestSlotParamsCapture((slotRequestIdScheduledUnitSlotProfileTuple3) -> {
+			taskFuture.complete(slotRequestIdScheduledUnitSlotProfileTuple3.f1);
+			slotProfileFuture.complete(slotRequestIdScheduledUnitSlotProfileTuple3.f2);
+		});
+
+		final ExecutionVertexID executionVertexId1 = new ExecutionVertexID(new JobVertexID(), 0);
```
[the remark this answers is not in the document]
shuai-xu: Done

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocatorTest.java
```diff
@@ -0,0 +1,332 @@
...
+		assertThat(cancelledSlotRequestIds, hasSize(1));
+		assertThat(cancelledSlotRequestIds, contains(receivedSlotRequestIds.toArray()));
+
+		try {
+			assignments.iterator().next().getLogicalSlotFuture().get();
+			fail("Expect a CancellationException but got nothing.");
+		} catch (CancellationException e) {
+			//
+		} catch (Exception e) {
+			fail("Expect an CancellationException but got " + e);
```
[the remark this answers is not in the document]
shuai-xu: OK

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocatorTest.java
```diff
@@ -0,0 +1,332 @@
...
+import static org.junit.Assert.fail;
+
+/**
+ * Tests for {@link DefaultExecutionSlotAllocator}.
+ */
+public class DefaultExecutionSlotAllocatorTest extends TestLogger {
+
+	private ParamsCapturedTestingSlotProvider slotProvider;
+
+	private Queue<CompletableFuture<LogicalSlot>> slotFutures;
```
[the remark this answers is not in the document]
shuai-xu: It does make sense. I tried to use the existing TestingSlotProvider to avoid too many testing slot providers. Adding a AllocationToggableSlotProvider does make tests more simple.

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/TestingInputsLocationsRetriever.java
```diff
@@ -0,0 +1,98 @@
...
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+import java.util.stream.Collectors;
+
+/**
+ * A simple inputs locations retriever for testing purposes.
+ */
+class TestingInputsLocationsRetriever implements InputsLocationsRetriever {
+
+	private final Map<ExecutionVertexID, List<ExecutionVertexID>> producersByVertex;
```
GJL: `producersByConsumer` is better, I think.

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/TestingInputsLocationsRetriever.java
```diff
@@ -0,0 +1,98 @@
...
+				return CompletableFuture.completedFuture(new LocalTaskManagerLocation());
+			}
+			future.complete(new LocalTaskManagerLocation());
+			return future;
+		});
+	}
+
+	static class Builder {
+
+		private final Map<ExecutionVertexID, List<ExecutionVertexID>> consumerToProducers = new HashMap<>();
```
GJL: Let's call it `producersByConsumer` or rename the other variable.

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/TestingInputsLocationsRetriever.java
shuai-xu: OK

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocator.java
```diff
@@ -0,0 +1,248 @@
...
+		Collection<Collection<ExecutionVertexID>> allProducers =
+				inputsLocationsRetriever.getConsumedResultPartitionsProducers(executionVertexId);
+		for (Collection<ExecutionVertexID> producers : allProducers) {
+
+			for (ExecutionVertexID producer : producers) {
+				Optional<CompletableFuture<TaskManagerLocation>> optionalLocationFuture =
+						inputsLocationsRetriever.getTaskManagerLocation(producer);
+				optionalLocationFuture.ifPresent(
+						(locationFuture) -> locationsFutures.add(locationFuture));
+				// If the parallelism is large, wait for all futures coming back may cost a long time.
```
tillrohrmann: Is this main reason or is it because of the data portion which becomes smaller the larger the number of different locations becomes? Have we seen that waiting for all futures becomes a problem somewhere?

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/ExecutionSlotAllocator.java
```diff
@@ -0,0 +1,51 @@
...
+import java.util.Collection;
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * Component responsible for assigning slots to a collection of {@link Execution}.
+ */
+public interface ExecutionSlotAllocator {
+
+	/**
+	 * Allocates slots for the given executions.
```
tillrohrmann: Nit: Other JavaDocs don't use third person conjugation.

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/ExecutionVertexSchedulingRequirements.java
```diff
@@ -0,0 +1,157 @@
...
+import static org.apache.flink.util.Preconditions.checkNotNull;
+
+/**
+ * The requirements for scheduling a {@link ExecutionVertex}.
+ */
+public class ExecutionVertexSchedulingRequirements {
+
+	private final ExecutionVertexID executionVertexId;
+
+	private final AllocationID previousAllocationId;
```
tillrohrmann: `@Nullable` missing

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/ExecutionVertexSchedulingRequirements.java
```diff
@@ -0,0 +1,157 @@
...
+
+	private final SlotSharingGroupId slotSharingGroupId;
+
+	private final CoLocationConstraint coLocationConstraint;
+
+	private final Collection<TaskManagerLocation> preferredLocations;
+
+	private ExecutionVertexSchedulingRequirements(
+			ExecutionVertexID executionVertexId,
+			AllocationID previousAllocationId,
```
tillrohrmann: `@Nullable` is missing

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/ExecutionVertexSchedulingRequirements.java
```diff
@@ -0,0 +1,157 @@
...
+ */
+public class ExecutionVertexSchedulingRequirements {
+
+	private final ExecutionVertexID executionVertexId;
+
+	private final AllocationID previousAllocationId;
+
+	private final ResourceProfile resourceProfile;
+
+	private final SlotSharingGroupId slotSharingGroupId;
```
tillrohrmann: `@Nullable` missing

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/ExecutionVertexSchedulingRequirements.java
```diff
@@ -0,0 +1,157 @@
...
+
+	private final CoLocationConstraint coLocationConstraint;
```
tillrohrmann: `@Nullable` missing

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/ExecutionVertexSchedulingRequirements.java
```diff
@@ -0,0 +1,157 @@
...
+
+	private final Collection<TaskManagerLocation> preferredLocations;
+
+	private ExecutionVertexSchedulingRequirements(
+			ExecutionVertexID executionVertexId,
+			AllocationID previousAllocationId,
+			ResourceProfile resourceProfile,
+			SlotSharingGroupId slotSharingGroupId,
```
tillrohrmann: `@Nullable` missing

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/ExecutionVertexSchedulingRequirements.java
```diff
@@ -0,0 +1,157 @@
...
+			CoLocationConstraint coLocationConstraint,
```
tillrohrmann: `@Nullable` missing

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocatorPreferredLocationsTest.java
```diff
@@ -0,0 +1,116 @@
...
+import java.util.concurrent.CompletableFuture;
+
+import static org.hamcrest.Matchers.containsInAnyOrder;
+import static org.hamcrest.Matchers.hasSize;
+import static org.junit.Assert.assertThat;
+
+/**
+ * Tests for the calculation of preferred locations based on inputs in {@link DefaultExecutionSlotAllocator}.
+ */
+public class DefaultExecutionSlotAllocatorPreferredLocationsTest extends TestLogger {
```
tillrohrmann: Nice tests :-)

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocatorTest.java
```diff
@@ -0,0 +1,336 @@
...
+
+		executionSlotAllocator.cancel(producerId);
+
+		assertThat(slotProvider.getCancelledSlotRequestIds(), hasSize(1));
+		assertThat(slotProvider.getCancelledSlotRequestIds(), contains(slotProvider.getReceivedSlotRequestIds().toArray()));
+
+		try {
+			assignments.iterator().next().getLogicalSlotFuture().get();
+			fail("Expect a CancellationException but got nothing.");
+		} catch (CancellationException e) {
```
tillrohrmann: `e` could be `ignored`

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocatorTest.java
```diff
@@ -0,0 +1,336 @@
...
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertThat;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+/**
+ * Tests for {@link DefaultExecutionSlotAllocator}.
+ */
+public class DefaultExecutionSlotAllocatorTest extends TestLogger {
```
tillrohrmann: Nice tests :-)

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/ExecutionSlotAllocator.java
shuai-xu: OK

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocatorTest.java
shuai-xu: Done

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/ExecutionVertexSchedulingRequirements.java
shuai-xu: Done

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocator.java
shuai-xu: I think this is the main reason. As user will change the parallelism according to the total size of data to keep the size of each partition stable. We didn't see the case waiting for a long time since the logic is same in legacy code and for almost all jobs in alibaba, the parallelism is much larger than 8.

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/TestingInputsLocationsRetriever.java
```diff
@@ -0,0 +1,98 @@
...
+		private final Map<ExecutionVertexID, List<ExecutionVertexID>> producersByConsumer = new HashMap<>();
+
+		public Builder connectConsumerToProducer(final ExecutionVertexID consumer, final ExecutionVertexID producer) {
+			producersByConsumer.compute(consumer, (key, producers) -> {
+				if (producers == null) {
+					producers = new ArrayList<>();
+				}
+				producers.add(producer);
+				return producers;
+			});
```
eaglewatcherwb: `producersByConsumer.computeIfAbsent(consumer, (key) -> new ArrayList<>()).add(producer);` could be simple.

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocator.java
```diff
@@ -0,0 +1,248 @@
...
+		SlotExecutionVertexAssignment slotExecutionVertexAssignment = pendingSlotAssignments.get(executionVertexId);
+		if (slotExecutionVertexAssignment != null) {
+			slotExecutionVertexAssignment.getLogicalSlotFuture().cancel(false);
+		}
+	}
+
+	@Override
+	public CompletableFuture<Void> stop() {
+		List<ExecutionVertexID> executionVertexIds = new ArrayList<>(pendingSlotAssignments.keySet());
+		executionVertexIds.stream().forEach(executionVertexID -> cancel(executionVertexID));
```
eaglewatcherwb:
```suggestion
		executionVertexIds.forEach(this::cancel);
```

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocator.java
```diff
@@ -0,0 +1,248 @@
...
+					uniqueLocationsFuture,
+					(locationsOnOneEdge, locationsOnAnotherEdge) -> {
+						if ((!locationsOnOneEdge.isEmpty() && locationsOnAnotherEdge.size() > locationsOnOneEdge.size())
+								|| locationsOnAnotherEdge.isEmpty()) {
+							return locationsOnOneEdge;
+						} else {
+							return locationsOnAnotherEdge;
+						}
+					});
+			locationsFutures.clear();
```
eaglewatcherwb: why use `clear` rather than `new`, in most cases `new` is more efficient. (https://stackoverflow.com/questions/6961356/list-clear-vs-list-new-arraylistinteger)

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocator.java
```diff
@@ -0,0 +1,248 @@
...
+					locationsFutures.clear();
+					break;
+				}
+			}
+
+			CompletableFuture<Collection<TaskManagerLocation>> uniqueLocationsFuture =
+					FutureUtils.combineAll(locationsFutures).thenApply(
+							(locations) -> {
+								Set<TaskManagerLocation> uniqueLocations = new HashSet<>(locations);
+								return uniqueLocations;
```
eaglewatcherwb: lambda could be replaced with `thenApply(HashSet::new)`;

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocator.java
```diff
@@ -0,0 +1,248 @@
...
+
+		Collection<Collection<ExecutionVertexID>> allProducers =
+				inputsLocationsRetriever.getConsumedResultPartitionsProducers(executionVertexId);
+		for (Collection<ExecutionVertexID> producers : allProducers) {
+
+			for (ExecutionVertexID producer : producers) {
+				Optional<CompletableFuture<TaskManagerLocation>> optionalLocationFuture =
+						inputsLocationsRetriever.getTaskManagerLocation(producer);
+				optionalLocationFuture.ifPresent(
+						(locationFuture) -> locationsFutures.add(locationFuture));
```
eaglewatcherwb: lambda could be replaced with `locationsFutures::add`.

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocator.java
```diff
@@ -0,0 +1,248 @@
...
+	static Set<AllocationID> computeAllPriorAllocationIds(
+			Collection<ExecutionVertexSchedulingRequirements> executionVertexSchedulingRequirements) {
+		HashSet<AllocationID> allPreviousAllocationIds = new HashSet<>();
+		for (ExecutionVertexSchedulingRequirements schedulingRequirements : executionVertexSchedulingRequirements) {
+			AllocationID priorAllocation = schedulingRequirements.getPreviousAllocationId();
+			if (priorAllocation != null) {
+				allPreviousAllocationIds.add(priorAllocation);
+			}
+		}
+		return allPreviousAllocationIds;
```
eaglewatcherwb: return executionVertexSchedulingRequirements
				.stream()
				.map(ExecutionVertexSchedulingRequirements::getPreviousAllocationId)
				.filter(Objects::nonNull)
				.collect(Collectors.toSet());
This could be simple.

GJL:
Merging. Will apply remaining changes by myself.

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/DefaultExecutionSlotAllocator.java
GJL: Probably because `.clear()` is closer to the legacy code. Can be changed later.

PR #8492, opened 2019-05-21
[FLINK-12563][table-runtime-blink] Introduce vector data format in blink

JingsongLi:

Introduce vector data format to high performance parquet/orc source.

## Brief change log

Introduce `VectorizedColumnBatch`: A `VectorizedColumnBatch` is a set of rows, organized with each column as a vector.
Introduce `ColumnVector` with subclasses: `IntColumnVector`, `LongColumnVector` and etc..
Introduce `ColumnarRow`: wrap `VectorizedColumnBatch` as a `BaseRow`.

Let `IntColumnVector` and others to be a interface:
1.We can support OrcXXVector with not copy data from OrcVector
2.In the future, we can support OffHeapXXVectors
Performance: we should keep a single implement XXColumnVector in JVM. Otherwise we need to bear the overhead of virtual function calls.

## Verifying this change

ut

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/dataformat/ColumnarRow.java
+++ b/flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/dataformat/ColumnarRow.java
@@ -1,0 +1,209 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.table.dataformat;
+
+import org.apache.flink.core.memory.MemorySegment;
+import org.apache.flink.core.memory.MemorySegmentFactory;
+import org.apache.flink.table.dataformat.vector.BytesColumnVector.Bytes;
+import org.apache.flink.table.dataformat.vector.VectorizedColumnBatch;
+
+/**
+ * Source directly return columnRow in order to reduce Row convert times.
+ */
+public final class ColumnarRow implements BaseRow {
+	private byte header;
+	private VectorizedColumnBatch vectorizedColumnBatch;
+	private int rowId;
+
+	public ColumnarRow() {}
+
+	public ColumnarRow(VectorizedColumnBatch vectorizedColumnBatch) {
+		this(vectorizedColumnBatch, 0);
+	}
+
+	public ColumnarRow(VectorizedColumnBatch vectorizedColumnBatch, int rowId) {
+		this.vectorizedColumnBatch = vectorizedColumnBatch;
+		this.rowId = rowId;
+	}
+
+	public void setVectorizedColumnBatch(
+			VectorizedColumnBatch vectorizedColumnBatch) {
+		this.vectorizedColumnBatch = vectorizedColumnBatch;
+	}
+
+	public void setRowId(int rowId) {
+		this.rowId = rowId;
+	}
+
+	@Override
+	public byte getHeader() {
+		return header;
+	}
+
+	@Override
+	public void setHeader(byte header) {
+		this.header = header;
+	}
+
+	@Override
+	public int getArity() {
+		return vectorizedColumnBatch.getArity();
+	}
+
+	@Override
+	public boolean isNullAt(int ordinal) {
+		return vectorizedColumnBatch.isNullAt(rowId, ordinal);
+	}
+
+	@Override
+	public boolean getBoolean(int ordinal) {
+		return vectorizedColumnBatch.getBoolean(rowId, ordinal);
+	}
+
+	@Override
+	public byte getByte(int ordinal) {
+		return vectorizedColumnBatch.getByte(rowId, ordinal);
+	}
+
+	@Override
+	public short getShort(int ordinal) {
+		return vectorizedColumnBatch.getShort(rowId, ordinal);
+	}
+
+	@Override
+	public int getInt(int ordinal) {
+		return vectorizedColumnBatch.getInt(rowId, ordinal);
+	}
+
+	@Override
+	public long getLong(int ordinal) {
+		return vectorizedColumnBatch.getLong(rowId, ordinal);
+	}
+
+	@Override
+	public float getFloat(int ordinal) {
+		return vectorizedColumnBatch.getFloat(rowId, ordinal);
+	}
+
+	@Override
+	public double getDouble(int ordinal) {
+		return vectorizedColumnBatch.getDouble(rowId, ordinal);
+	}
+
+	@Override
+	public BinaryString getString(int ordinal) {
+		Bytes byteArray = vectorizedColumnBatch.getByteArray(rowId, ordinal);
+		MemorySegment memorySegment = MemorySegmentFactory.wrap(byteArray.data);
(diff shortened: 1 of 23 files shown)


flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/dataformat/vector/AbstractColumnVector.java
```diff
@@ -0,0 +1,54 @@
...
+/**
+ * Contains the shared structure for {@link ColumnVector}s, including NULL information and dictionary.
+ * NOTE: if there are some nulls, must set {@link #noNulls} to false.
+ */
+public abstract class AbstractColumnVector implements ColumnVector, Serializable {
+
+	private static final long serialVersionUID = 5340018531388047747L;
+
+	// If the whole column vector has no nulls, this is true, otherwise false.
+	public boolean noNulls = true;
```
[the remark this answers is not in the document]
KurtYoung: and i think this shouldn't have any default value

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/dataformat/vector/heap/HeapByteVector.java
```diff
@@ -0,0 +1,50 @@
...
+		super(len);
+		vector = new byte[len];
+	}
+
+	@Override
+	public byte getByte(int i) {
+		if (dictionary == null) {
+			return vector[i];
+		} else {
+			return (byte) dictionary.decodeToInt(dictionaryIds.vector[i]);
```
KurtYoung: Is it reasonable for byte vector to have a dictionary? the `dictionaryIds` is even larger than the actual data

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/dataformat/vector/AbstractColumnVector.java
```diff
@@ -0,0 +1,54 @@
...
+	/**
+	 * The Dictionary for this column.
+	 * If it's not null, will be used to decode the value in get().
+	 */
+	protected Dictionary dictionary;
+
+	/**
+	 * Update the dictionary.
+	 */
+	public void setDictionary(Dictionary dictionary) {
```
KurtYoung: I think `Dictionary` should be updated with `dictionaryIds`

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/dataformat/ColumnarRow.java
```diff
@@ -0,0 +1,209 @@
...
+	}
+
+	@Override
+	public BaseRow getRow(int ordinal, int numFields) {
+		throw new UnsupportedOperationException("Row is not supported.");
+	}
+
+	@Override
+	public BinaryArray getArray(int ordinal) {
+		throw new UnsupportedOperationException("Array is not supported.");
```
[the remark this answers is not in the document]
JingsongLi: The biggest dependence is that we haven't introduced BaseArray and BaseMap yet. So the conversion of Vectors and BinaryArray is heavy now.
I think we should introduce BaseArray and BaseMap first.

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/dataformat/vector/AbstractColumnVector.java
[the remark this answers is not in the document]
JingsongLi: I think it can be default false for safe.

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/dataformat/vector/AbstractColumnVector.java
JingsongLi: `dictionaryIds` maybe inconsistent with `setDictionary`.
Suppose a ColumnVector's data comes from two pages of a parquet. Perhaps one page uses a dictionary and the other page does not use a dictionary.
The first page that uses a field will have `dictionaryIds`, which requires decoding the first page. (Out batch does not support a mix of dictionary)

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/dataformat/vector/heap/HeapByteVector.java
JingsongLi: It seems that only boolean vector dose not need dictionary.
byte vector need dictionary for storage, dictionaryIds use runLengthCoding, it will be very small. In reading, we need decode it.

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/dataformat/vector/AbstractColumnVector.java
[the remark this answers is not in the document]
JingsongLi: protected instead. set through `setNullAt`.

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/dataformat/vector/BytesColumnVector.java
```diff
@@ -0,0 +1,48 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.table.dataformat.vector;
+
+/**
+ * Bytes column vector to get {@link Bytes}, it include origin data and offset and length. The data
```
liyafan82: includes original?

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/dataformat/vector/VectorizedColumnBatch.java
```diff
@@ -0,0 +1,134 @@
...
+	 * Resets the batch for writing.
+	 */
+	public void reset() {
+		for (ColumnVector column : columns) {
+			column.reset();
+		}
+		this.numRows = 0;
+	}
+
+	public void setNumRows(int numRows) {
```
KurtYoung: why not set this in constructor?

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/dataformat/vector/heap/AbstractHeapVector.java
```diff
@@ -0,0 +1,130 @@
...
+	}
+
+	/**
+	 * Returns the underlying integer column for ids of dictionary.
+	 */
+	public HeapIntVector getDictionaryIds() {
+		return dictionaryIds;
+	}
+
+	public static AbstractHeapVector[] allocateHeapVectors(InternalType[] fieldTypes, int maxRows) {
```
KurtYoung: use new type system?

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/dataformat/vector/VectorizedColumnBatch.java
JingsongLi: `VectorizedColumnBatch` is reused usually, need `setNumRows` after updating it.

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/dataformat/vector/heap/AbstractHeapVector.java
JingsongLi: ok

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/dataformat/vector/VectorizedColumnBatch.java
KurtYoung: But the `columns` is immutable, how do you plan to reuse `VectorizedColumnBatch`?

JingsongLi: No, `columns` is not immutable, we can update the data (like int[]), and reuse it. (not use `ColumnarRow` to update it, we direct copy data to int[] to speed up updating)

PR #8493, opened 2019-05-21
[FLINK-12562][table] Improve code in ExpandColumnFunctionsRule

hequn8128:

This pull requests resolve IntelliJ warnings in `ExpandColumnFunctionsRule`. Also, change TableException to ValidationException for the validation as `ValidationException` is used to indicate something wrong from users while `TableException` means unsupported stuff or some weird state for Flink. 

## Brief change log

  - Change TableException to ValidationException for validation.
  - Fix code warnings.
  - Adjust tests.

## Verifying this change

This change is already covered by existing tests, such as `ColumnFunctionsValidationTest`.

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-table/flink-table-planner/src/main/java/org/apache/flink/table/expressions/rules/ExpandColumnFunctionsRule.java
+++ b/flink-table/flink-table-planner/src/main/java/org/apache/flink/table/expressions/rules/ExpandColumnFunctionsRule.java
@@ -19,7 +19,6 @@
 package org.apache.flink.table.expressions.rules;
 
 import org.apache.flink.annotation.Internal;
-import org.apache.flink.table.api.TableException;
 import org.apache.flink.table.api.ValidationException;
 import org.apache.flink.table.expressions.ApiExpressionDefaultVisitor;
 import org.apache.flink.table.expressions.CallExpression;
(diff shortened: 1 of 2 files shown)


hequn8128:
@flinkbot attention @twalthr @sunjincheng121

flink-table/flink-table-planner/src/main/java/org/apache/flink/table/expressions/rules/ExpandColumnFunctionsRule.java
```diff
@@ -193,7 +190,7 @@ public ColumnsExpressionExpander(List<UnresolvedReferenceExpression> inputFieldR
 
 		@Override
 		protected List<UnresolvedReferenceExpression> defaultMethod(Expression expression) {
-			throw new TableException(
+			throw new ValidationException(
 				String.format("The parameters of %s() or %s() only accept column name or column " +
```
twalthr: One additional location that looks like inconsistent regarding exceptions. Is the `Preconditions.checkArgument`. Usually this method is for checking if we (the Flink developers) did everything correct how every this check is also validating user input, right?

hequn8128: Maybe `Preconditions.checkArgument` can also be used to validate user input, such as the checkArgument in `org.apache.flink.streaming.api.environment.StreamExecutionEnvironment`.  However, I think it also makes sense to change the `Preconditions.checkArgument` to ValidationException. I will update the PR according to your suggestions.

hequn8128:
@twalthr Hi, thanks a lot for your suggestions. I have updated the PR according to your comments. Sorry for the late response. 

Best, Hequn

PR #8494, opened 2019-05-21
[FLINK-12570] Switch Task from ResultPartition to ResultPartitionWriter interface

azagrebin:

This part of Shuffle API refactoring: make task not depend on the concrete implementation of ResultPartitionWriter (ResultPartition).

## Brief change log

- Move fail and finish methods to ResultPartitionWriter interface from ResultPartition
- Make Task.producedPartitions a ResultPartitionWriter array, not ResultPartition
- adjust other test implementations of ResultPartitionWriter interface

## Verifying this change

refactoring covered by existing tests.

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/api/writer/ResultPartitionWriter.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/api/writer/ResultPartitionWriter.java
@@ -22,6 +22,8 @@ import org.apache.flink.runtime.io.network.buffer.BufferConsumer;
 import org.apache.flink.runtime.io.network.buffer.BufferProvider;
 import org.apache.flink.runtime.io.network.partition.ResultPartitionID;
 
+import javax.annotation.Nullable;
+
 import java.io.IOException;
 
 /**
@@ -65,4 +67,21 @@ public interface ResultPartitionWriter extends AutoCloseable {
 	 * Manually trigger consumption from enqueued {@link BufferConsumer BufferConsumers} in one specified subpartition.
 	 */
 	void flush(int subpartitionIndex);
+
+	/**
+	 * Fail the production of the partition.
+	 *
+	 * <p>This method propagates non-{@code null} failure causes to consumers on a best-effort basis.
+	 * Closing of partition is still needed.
+	 *
+	 * @param throwable failure cause
+	 */
+	void fail(@Nullable Throwable throwable);
+
+	/**
+	 * Successfully finish the production of the partition.
+	 *
+	 * <p>Closing of partition is still needed.
+	 */
+	void finish() throws IOException;
 }
--- a/flink-runtime/src/test/java/org/apache/flink/runtime/io/network/api/writer/RecordWriterTest.java
+++ b/flink-runtime/src/test/java/org/apache/flink/runtime/io/network/api/writer/RecordWriterTest.java
@@ -55,6 +55,8 @@ import org.junit.rules.TemporaryFolder;
 import org.mockito.invocation.InvocationOnMock;
 import org.mockito.stubbing.Answer;
 
+import javax.annotation.Nullable;
+
 import java.io.IOException;
 import java.util.ArrayDeque;
 import java.util.ArrayList;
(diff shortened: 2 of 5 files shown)


azagrebin:
based on #8416 atm

azagrebin:

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/api/writer/ResultPartitionWriter.java
```diff
@@ -65,4 +67,23 @@
...
+	 * Closing of partition is still needed afterwards.
+	 *
+	 * @param throwable failure cause
+	 */
+	void fail(@Nullable Throwable throwable);
+
+	/**
+	 * Indicates successful end of partition producing.
+	 *
+	 * <p>It can be used to notify and release consumer.
```
zhijiangW: release consumer? maybe deploy consumer

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/api/writer/RecordWriterTest.java
```diff
@@ -510,6 +512,14 @@ public void flushAll() {
 		public void flush(int subpartitionIndex) {
 		}
 
+		@Override
+		public void fail(@Nullable Throwable throwable) {
```
zhijiangW: Why not keep the same way of `UnsupportedOperationException` as others?

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/api/writer/ResultPartitionWriter.java
azagrebin: True, deploy in case of blocking and it actually notifies JM which is not shuffle concern but wrapper's in Task.
For pipelined it looks more like releasing by sending EndOfPartition, I will just change to `notify consumers that the producing is done`

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/api/writer/RecordWriterTest.java
azagrebin: True, I missed it, thanks :)

azagrebin:
Thanks for the review @zhijiangW ! I've addressed comments

zentol:
Is it required that either finish() or fail() are called _before_ close()? Is their primary purpose to notify consumers?

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/api/writer/ResultPartitionWriter.java
```diff
@@ -65,4 +67,23 @@
 	 * Manually trigger consumption from enqueued {@link BufferConsumer BufferConsumers} in one specified subpartition.
 	 */
 	void flush(int subpartitionIndex);
+
+	/**
+	 * Fails partition producing.
+	 *
+	 * <p>The method propagates non-{@code null} failure cause to consumer based on best effort.
```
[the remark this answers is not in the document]
zentol: Is the propagation of errors a a requirement for the implementation, or just a detail of how netty does it?

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/api/writer/ResultPartitionWriter.java
```diff
@@ -65,4 +67,23 @@
...
+	 * Closing of partition is still needed afterwards.
+	 *
+	 * @param throwable failure cause
+	 */
+	void fail(@Nullable Throwable throwable);
+
+	/**
+	 * Indicates successful end of partition producing.
+	 *
+	 * <p>It can be used to notify the consumers that the producing is done.
```
zentol: "can be used" implies that all implementation allow this, and the user decides whether to do it or not.
If the notification is an implementation detail I'd remove it from the javadocs here.

azagrebin:
Thanks for the review @zentol ! I changed the comments

PR #8495, opened 2019-05-21
[FLINK-12556][e2e] Extend some end-to-end tests to run with custom (input) File System implementation

1u0:

This pull request extends some end-to-end tests to be independent of origin of input source (be it a local file, s3 or an artificial file system plugin), provided that they all serve the same test data content.

In particular, a new (dummy) file system plugin is introduced to check that different Flink cluster setups are working with additional (external) plugin in `lib` folder.

Later, those tests can be modified to check that new plugin system also works in different cluster setups ([FLINK-11952](https://issues.apache.org/jira/browse/FLINK-11952))

## Brief change log

  - Yarn kerberos docker test is simplified (some code reuse) without changing behavior;
  - batch wordcount test is generalized and now hosts (deleted) shaded hadoop and shaded presto tests;
  - introduced a dummy FS plugin that serves hardcoded read-only contents from memory;
  - add some test runs that use dummy FS plugin as input source.

## Verifying this change

  - The modified end-to-end tests should pass

## Documentation


flink-end-to-end-tests/flink-plugins-test/src/main/java/org/apache/flink/fs/dummy/DummyFs.java
```diff
@@ -0,0 +1,166 @@
...
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.net.URI;
+import java.nio.charset.Charset;
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.apache.flink.util.Preconditions.checkNotNull;
+
+class DummyFs extends FileSystem {
```
pnowojski: Rename to `DummyFileSystem`? One thing is that I generally do not like abbreviations another is that the pre-existing convention is to not abbreviate this.

flink-end-to-end-tests/flink-plugins-test/src/main/java/org/apache/flink/fs/dummy/DummyFs.java
```diff
@@ -0,0 +1,166 @@
...
+	}
+
+	@Override
+	public FSDataInputStream open(final Path f) throws IOException {
+		return DummyFsInputStream.create(getDataByPath(f));
+	}
+
+	@Override
+	public boolean delete(final Path path, final boolean recursive) throws IOException {
+		checkNotNull(path, "path is null");
```
pnowojski: nit: I would drop `checkNotNull` for methods that `throw new UnsupportedOperationException();`

flink-end-to-end-tests/flink-plugins-test/src/main/java/org/apache/flink/fs/dummy/DummyFs.java
```diff
@@ -0,0 +1,166 @@
...
+	}
+
+	@Override
+	public boolean rename(final Path src, final Path dst) throws IOException {
+		throw new UnsupportedOperationException("Test FS doesn't support rename operation");
+	}
+
+	@Override
+	public boolean isDistributedFS() {
+		return false;
```
pnowojski: shouldn't this be `return true`? All of the processes will see the same dummy files, right?

flink-end-to-end-tests/flink-plugins-test/src/main/java/org/apache/flink/fs/dummy/DummyFs.java
```diff
@@ -0,0 +1,166 @@
...
+
+	@Override
+	public FileSystemKind getKind() {
+		return FileSystemKind.OBJECT_STORE;
+	}
+
+	@Nullable
+	private byte[] getDataByPath(Path path) {
+		URI uri = path.toUri();
+		String resourcePath = uri.getPath().substring(1);
```
pnowojski: why do we need `substring(1)`?

flink-end-to-end-tests/flink-plugins-test/src/main/java/org/apache/flink/fs/dummy/DummyFsFactory.java
```diff
@@ -0,0 +1,52 @@
...
+import java.io.IOException;
+import java.net.URI;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * A FileSystem implementation for integration testing purposes. Supports and serves read-only content from static
+ * key value map.
+ */
+public class DummyFsFactory implements FileSystemFactory {
```
pnowojski: ditto `Fs` -> `FileSystem`.

flink-end-to-end-tests/flink-plugins-test/src/main/java/org/apache/flink/fs/dummy/DummyFs.java
```diff
@@ -0,0 +1,166 @@
...
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.net.URI;
+import java.nio.charset.Charset;
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.apache.flink.util.Preconditions.checkNotNull;
+
```
pnowojski: Should we add a java doc with a reference to `DummyFileSystemFactory`?

flink-end-to-end-tests/flink-plugins-test/src/main/java/org/apache/flink/fs/dummy/DummyFsFileStatus.java
```diff
@@ -0,0 +1,67 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.fs.dummy;
+
+import org.apache.flink.core.fs.FileStatus;
+import org.apache.flink.core.fs.Path;
+
+class DummyFsFileStatus implements FileStatus {
```
pnowojski: `DummyFileStatus`?  (pre-existing convention)

flink-end-to-end-tests/flink-plugins-test/src/main/java/org/apache/flink/fs/dummy/DummyFsInputStream.java
```diff
@@ -0,0 +1,51 @@
...
+ */
+
+package org.apache.flink.fs.dummy;
+
+import org.apache.flink.core.fs.FSDataInputStream;
+import org.apache.flink.core.memory.ByteArrayInputStreamWithPos;
+
+import java.io.IOException;
+
+class DummyFsInputStream extends FSDataInputStream {
```
pnowojski: `DummyDataInputStream`? (pre-existing convention)

flink-end-to-end-tests/run-nightly-tests.sh
```diff
@@ -88,7 +88,8 @@ run_test "Resuming Externalized Checkpoint after terminal failure (rocks, increm
 # Docker
 ################################################################################
 
-run_test "Running Kerberized YARN on Docker test " "$END_TO_END_DIR/test-scripts/test_yarn_kerberos_docker.sh"
+run_test "Running Kerberized YARN on Docker test (file input)" "$END_TO_END_DIR/test-scripts/test_yarn_kerberos_docker.sh file"
```
pnowojski: I think we could just replace `file` to `dummy-fs` in yarn docker test

flink-end-to-end-tests/run-pre-commit-tests.sh
```diff
@@ -55,6 +55,7 @@ run_test "Streaming Python Wordcount end-to-end test" "$END_TO_END_DIR/test-scri
 run_test "Wordcount end-to-end test" "$END_TO_END_DIR/test-scripts/test_batch_wordcount.sh file"
 run_test "Shaded Hadoop S3A end-to-end test" "$END_TO_END_DIR/test-scripts/test_batch_wordcount.sh hadoop"
 run_test "Shaded Presto S3 end-to-end test" "$END_TO_END_DIR/test-scripts/test_batch_wordcount.sh presto"
+run_test "Custom FS plugin end-to-end test" "$END_TO_END_DIR/test-scripts/test_batch_wordcount.sh dummy-fs"
```
pnowojski: We do not need to use `dummy-fs` here. To test plugins we can just existing `presto` and `hadoop` s3 variants.

tools/travis/splits/split_container.sh
```diff
@@ -43,8 +43,9 @@ echo "Flink distribution directory: $FLINK_DIR"
 
 # run_test "<description>" "$END_TO_END_DIR/test-scripts/<script_name>"
 
-run_test "Wordcount on Docker test" "$END_TO_END_DIR/test-scripts/test_docker_embedded_job.sh"
-run_test "Running Kerberized YARN on Docker test " "$END_TO_END_DIR/test-scripts/test_yarn_kerberos_docker.sh"
+run_test "Wordcount on Docker test (custom fs plugin)" "$END_TO_END_DIR/test-scripts/test_docker_embedded_job.sh dummy-fs"
```
pnowojski: ditto: I think we do not need to increase number of tests here

flink-end-to-end-tests/test-scripts/test_yarn_kerberos_docker.sh
```diff
@@ -47,6 +65,10 @@ function cluster_shutdown {
 trap cluster_shutdown INT
 trap cluster_shutdown EXIT
 
+#if [[ 1 ]]; then
```
pnowojski: some leftovers?

flink-end-to-end-tests/test-scripts/test_yarn_kerberos_docker.sh
```diff
@@ -92,34 +88,28 @@ do
...
-done
-
-if [[ ${CLUSTER_STARTED} -ne 0 ]]; then
+if ! retry_times $CLUSTER_SETUP_RETRIES 0 start_hadoop_cluster; then
     echo "ERROR: Could not start hadoop cluster. Aborting..."
     exit 1
 fi
 
+mkdir -p $FLINK_TARBALL_DIR
+tar czf $FLINK_TARBALL_DIR/$FLINK_TARBALL -C $(dirname $FLINK_DIR) .
```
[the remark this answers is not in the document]
1u0: They are moved here to be run later. So if the test fails early (for example on cluster spin up), it won't create unused tarball.
This change was part of the test refactoring (you can check it by reading commits one-by-one).

flink-end-to-end-tests/test-scripts/test_yarn_kerberos_docker.sh
```diff
@@ -92,34 +88,28 @@ do
...
-for (( i = 0; i < $CLUSTER_SETUP_RETRIES; i++ ))
-do
-    if start_hadoop_cluster; then
-       echo "Cluster started successfully."
-       CLUSTER_STARTED=0
-       break #continue test, cluster set up succeeded
-    fi
-
-    echo "ERROR: Could not start hadoop cluster. Retrying..."
-    docker-compose -f $END_TO_END_DIR/test-scripts/docker-hadoop-secure-cluster/docker-compose.yml down
```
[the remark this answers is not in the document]
1u0: I expect it to be handled by `cluster_shutdown`.

flink-end-to-end-tests/flink-plugins-test/src/main/java/org/apache/flink/fs/dummy/DummyFs.java
1u0: I haven't dug into semantics of this flag and how it is interpreted internally. The original implementation is heavily based on `LocalFileSystem`. But I have changed it to `true`, the batch wordcount test still pass.

flink-end-to-end-tests/test-scripts/test_yarn_kerberos_docker.sh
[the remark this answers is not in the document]
pnowojski: Yes, I was reading commits one by one and I was still confused (hence my question). I wasn't (and I'm still not) sure if all of those refactorings in the first commit are related to one another or if they are 3 independent things - if the latter one, it would be better to have them in separate commits.

flink-end-to-end-tests/flink-plugins-test/src/main/java/org/apache/flink/fs/dummy/DummyFs.java
```diff
@@ -0,0 +1,166 @@
...
+	private final URI workingDir;
+
+	private final URI homeDir;
+
+	private final Map<String, byte[]> contents;
+
+	DummyFs(Map<String, String> contents) {
+		this.workingDir = new File(System.getProperty("user.dir")).toURI();
+		this.homeDir = new File(System.getProperty("user.home")).toURI();
```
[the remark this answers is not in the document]
pnowojski: It looks like it should be guaranteed to be set. At least I couldn't google out anything that would contradict this.

flink-end-to-end-tests/flink-plugins-test/src/main/java/org/apache/flink/fs/dummy/DummyFs.java
pnowojski: I guess it should be fine either way, but to me it looks like `true` would be more accurate and potentially less confusing for future developers

flink-end-to-end-tests/flink-plugins-test/src/main/java/org/apache/flink/fs/dummy/DummyFs.java
pnowojski: There is no clear convention. I follow up a rule, that I place a comment if something is not obvious based on the class name/line of code. Here `DummyFileSystem` doesn't explain in what why is it "dummy", so it needs some explanation. The problem is that a developer can encounter this `DummyFileSystem` either through `FileSystem` or `FileSystemFactory`, so I would place an explanation in one of them, and reference it in the other.

It hardly matters if class is a test class, or not.

flink-end-to-end-tests/run-pre-commit-tests.sh
pnowojski: Hmm, testing a test class?

pnowojski:
@1u0 30 seconds is a lot for a single test :(

flink-end-to-end-tests/flink-plugins-test/src/main/java/org/apache/flink/fs/dummy/DummyFileSystemFileStatus.java
```diff
@@ -0,0 +1,67 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.fs.dummy;
+
+import org.apache.flink.core.fs.FileStatus;
+import org.apache.flink.core.fs.Path;
+
+class DummyFileSystemFileStatus implements FileStatus {
```
pnowojski: `DummyFileStatus` would follow a pre-existing convention of `HadoopFileStatus` and `LocalFileStatus`

flink-end-to-end-tests/flink-plugins-test/src/main/java/org/apache/flink/fs/dummy/DummyFileSystemInputStream.java
```diff
@@ -0,0 +1,51 @@
...
+ */
+
+package org.apache.flink.fs.dummy;
+
+import org.apache.flink.core.fs.FSDataInputStream;
+import org.apache.flink.core.memory.ByteArrayInputStreamWithPos;
+
+import java.io.IOException;
+
+class DummyFileSystemInputStream extends FSDataInputStream {
```
pnowojski: ditto: `DummyInputStream`

flink-end-to-end-tests/flink-plugins-test/src/main/java/org/apache/flink/fs/dummy/DummyFs.java
1u0: There is a description in the factory class.

pnowojski: Yes, but:
> The problem is that a developer can encounter this DummyFileSystem either through FileSystem or FileSystemFactory, so I would place an explanation in one of them, and reference it in the other.

I mean it's not obvious that if you wonder what `DummyFileSystem` is doing, that you should check a comment in `DummyFileSystemFactory` class.

flink-end-to-end-tests/flink-plugins-test/src/main/java/org/apache/flink/fs/dummy/DummyFileSystemFileStatus.java
1u0: If you insist in such change, I propose to revert back the renaming and change `DummyFileSystem` to `DummyFs` with keeping it as prefix in those classes:
 1) I'm personally neutral about using common abbreviations (id, cpu, url, db, fs and so on). And there are already implementations of `FileSystem` that use such convention;
 2) the prefix here is intended rather as a marker prefix, similar to "Hadoop", "FlinkS3", "MapR" in other `FileSystem` implementations. It doesn't mean that the corresponding `FileStatus/FSDataInputStream` implementations are dummy or mocks.

1u0:
Timings of the tests, from the CI logs:
```
[PASS] 'Running Kerberized YARN on Docker test (default input)' passed after 8 minutes and 13 seconds! Test exited with exit code 0.
[PASS] 'Running Kerberized YARN on Docker test (custom fs plugin)' passed after 4 minutes and 33 seconds! Test exited with exit code 0.

[PASS] 'Wordcount end-to-end test' passed after 0 minutes and 12 seconds! Test exited with exit code 0.
[PASS] 'Shaded Hadoop S3A end-to-end test' passed after 0 minutes and 1 seconds! Test exited with exit code 0.
[PASS] 'Shaded Presto S3 end-to-end test' passed after 0 minutes and 0 seconds! Test exited with exit code 0.
[PASS] 'Custom FS plugin end-to-end test' passed after 0 minutes and 13 seconds! Test exited with exit code 0.
```

So adding a modified Yarn kerberos test as addition, extends tests run by 4,5 minutes. He it's may look faster, because it reuses the docker build created by previous test run. Otherwise, I expect them to run about the same amount of time in equal conditions.

CC @zentol.

flink-end-to-end-tests/flink-plugins-test/src/main/java/org/apache/flink/fs/dummy/DummyFileSystemFileStatus.java
pnowojski: Ok I get your point. In that case I would vote for one of the following:
1. `DummyFileSystem`, `DumyFileStatus`, `DummyDataInputStream`
2. `DummyFSFileSystem` `DumyFSFileStatus`, `DummyFSDataInputStream` (note `FS` which would also feet the code base better compared to `Fs` and note that all files should have the same prefix)
3. Rename it altogether, for example to `StaticContent` or `PredefinedContent` (`StaticContentFileSystem`, `StaticContentFileStatus`, `StaticContentDataInputStream`, ...)

I think the 3rd would be the best, as it also describes in what way the file system is "dummy", not just the fact that it is "dummy".

flink-end-to-end-tests/test-scripts/test_yarn_kerberos_docker.sh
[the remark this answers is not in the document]
pnowojski: > I wasn't (and I'm still not) sure if all of those refactorings in the first commit are related to one another or if they are 3 independent things 

Bumping the question :)

pnowojski:
LGTM % renaming, one question and if Chesney is fine with extending the testing time (I would keep the original count of tests)

flink-end-to-end-tests/flink-plugins-test/src/main/java/org/apache/flink/fs/dummy/DummyFileSystemFileStatus.java
1u0: I'm personally against option 1. As it's misleading that `DummyFileStatus`, `DummyDataInputStream`, ... are dummy, but actually they are valid implementations for corresponding FS.
I can adjust PR for option 2., by renaming `DummyFs` to `DummyFsFileSystem`. I was not aware about "abbreviation in uppercase" convention. In particular, I was using as examples of naming `HadoopFsFactory`, `MapRFsFactory`.

Initially, I also though about option 3, but `static` can't be used as a standalone part of package name. While `staticcontent` package name feels off style with other FSs.

pnowojski: Ok, you are right about `HadoopFsFactory` and `MapRFsFactory`. I was looking mostly at different abbreviations. `DummyFsFileSystem` sounds good to me in that case.

1u0:
@pnowojski, thanks for the review again. I have updated the PR and hopefully addressed your comments (except about additional test run time overhead).

If there is strong no, the latter can be also adjusted: both (or either of them), the word count with dummy-fs test (running on PR check) and the old yarn kerberos test with internal input (running nightly) can be removed from test runs.
I'm keeping them included for now, to at least showcase that both (old and new) tests are passed.

1u0:
@pnowojski, I've rebased the PR.
The CI pass is green 🍀😃.

PR #8496, opened 2019-05-21
[FLINK-12571][network] Make NetworkEnvironment#start() return the binded data port

zhijiangW:

*`NetworkEnvironment#getConnectionManager` is currently used for getting binded data port from `ConnectionManager`. Considering the general shuffle service architecture, the internal `ConnectionManager` in `NetworkEnvironment` should not be exposed to outsides. We could make `ShuffleService#start` return the binded data port directly if exists, then for other cases it could return a default int value which seems no harm.*

## Brief change log

  - *Make `ConnectionManager#start` return binded data port.*
  - *Make `NetworkEnvironment#start` return binded data port*
  - *Return legacy getter method in `NettyServer`*
  - *Refactor the process in `TaskManagerServices`*

## Verifying this change

## Documentation

### DIFF
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/ConnectionManager.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/ConnectionManager.java
@@ -26,7 +26,12 @@ import java.io.IOException;
  */
 public interface ConnectionManager {
 
-	void start() throws IOException;
+	/**
+	 * Starts the internal related components for network connection and communication.
+	 *
+	 * @return a port to connect to the task executor for shuffle data exchange, -1 if only local connection is possible.
+	 */
+	int start() throws IOException;
 
 	/**
 	 * Creates a {@link PartitionRequestClient} instance for the given {@link ConnectionID}.
@@ -39,8 +44,6 @@ public interface ConnectionManager {
 	void closeOpenChannelConnections(ConnectionID connectionId);
 
 	int getNumberOfActiveConnections();
-
-	int getDataPort();
 
 	void shutdown() throws IOException;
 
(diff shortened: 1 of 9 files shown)


zhijiangW:

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NetworkEnvironment.java
```diff
@@ -275,15 +276,15 @@ private void registerInputMetrics(MetricGroup inputGroup, MetricGroup buffersGro
 		buffersGroup.gauge(METRIC_INPUT_POOL_USAGE, new InputBufferPoolUsageGauge(inputGates));
 	}
 
-	public void start() throws IOException {
+	public int start() throws IOException {
```
azagrebin: Let's have a doc comment for this method in particular about the return value, also in the `ConnectionManager` interface

zhijiangW:
Thanks for reviews @azagrebin !

I pushed the fixup commit for adding doc comments and also another hotfix commit for removing legacy `TaskEventDispatcher` from `NetworkEnvironment`.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/ConnectionManager.java
```diff
@@ -26,6 +26,11 @@
  */
 public interface ConnectionManager {
 
+	/**
+	 * Starts the internal related components for network connection and communication.
+	 *
+	 * @return the data port of server bind address if have.
+	 */
```
azagrebin: I would write it like this:
```
/**
   * Starts the internal related components for network connection and communication.
   *
   * @return a port to connect to the task executor for shuffle data exchange, -1 if only local connection is possible.
   */
```
I think the same comment is relevant for `NetworkEnvironment`.
We could already squash this fixup into the previous commit.

zhijiangW:
Thanks for reviews @pnowojski . It is already green.

GJL:
Merging.

PR #8497, opened 2019-05-21
[FLINK-12566][table] Remove row interval type

twalthr:

This PR simplifies the type design by removing the row interval. A count should simply be represented as a `Long`. This also allows for making row count intervals available in standard SQL.

## Brief change log

- Row interval removed

## Verifying this change

  - The serializers: no

## Documentation


flink-table/flink-table-planner/src/main/scala/org/apache/flink/table/expressions/literals.scala
```diff
@@ -60,7 +60,6 @@ case class Literal(value: Any, resultType: TypeInformation[_]) extends LeafExpre
     case _@SqlTimeTypeInfo.TIMESTAMP => value.toString + ".toTimestamp"
     case _@TimeIntervalTypeInfo.INTERVAL_MILLIS => value.toString + ".millis"
     case _@TimeIntervalTypeInfo.INTERVAL_MONTHS => value.toString + ".months"
-    case _@RowIntervalTypeInfo.INTERVAL_ROWS => value.toString + ".rows"
```
dawidwys: How about we leave this suffix, and just replace the return type with `LONG_TYPE_INFO`? This is part of the API as far as I understand. If we remove that some of the user programs might fail.

It would also be inline with the scala dsl. You didn't remove the `org.apache.flink.table.api.scala.ImplicitExpressionOperations#rows`

twalthr: The suffix expression `.rows` is still valid for expressions. But since the type does not exist anymore, this string can not be generated anymore.

dawidwys: Sorry, totally my bad. I mistaken the class.

PR #8498, opened 2019-05-21
[FLINK-12413] [runtime] Implement ExecutionFailureHandler

zhuzhurk:

*ExecutionFailureHandler is a component of new generation scheduling. See [FLINK-10429](https://issues.apache.org/jira/browse/FLINK-10429). It serves task failover handling by returning tasks to restart upon task failures. This PR is to implement ExecutionFailureHandler and related classes.*

## Brief change log

Detailed design doc can be found at [design](https://docs.google.com/document/d/1D4EP9i5rx9CTQ4-iRu6DVqVEG8KzIMu07Lk4zjxLy08/edit#)
  - *Implement ExecutionFailureHandler*
  - *Implement FailureHandlingResult*
  - *Add interface RestartBackoffTimeStrategy and RestartBackoffTimeStrategy.Factory*
  - *Add interface FailoverStrategy.Factory*

## Verifying this change

*(example:)*
  - *UT for ExecutionFailureHandler*
  - *UT for FailureHandlingResult*

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/ExecutionFailureHandler.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/ExecutionFailureHandler.java
@@ -1,0 +1,89 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.runtime.executiongraph.failover.flip1;
+
+import org.apache.flink.annotation.VisibleForTesting;
+import org.apache.flink.runtime.JobException;
+import org.apache.flink.runtime.scheduler.strategy.ExecutionVertexID;
+import org.apache.flink.runtime.throwable.ThrowableClassifier;
+import org.apache.flink.runtime.throwable.ThrowableType;
+
+import static org.apache.flink.util.Preconditions.checkNotNull;
+
+/**
+ * This handler deals with task failures to return a {@link FailureHandlingResult} which contains tasks
+ * to be restarted to recover from failures.
+ */
+public class ExecutionFailureHandler {
+
+	/** Strategy to judge which tasks should be restarted. */
+	private FailoverStrategy failoverStrategy;
+
+	/** Strategy to judge whether and when a restarting should be done. */
+	private RestartBackoffTimeStrategy restartBackoffTimeStrategy;
+
+	/**
+	 * Creates the handler to deal with task failures.
+	 *
+	 * @param topology of tasks of current job
+	 * @param failoverStrategyFactory is the factory to create the failover strategy
+	 * @param restartStrategyFactory is the factory to create the restart strategy
+	 */
+	public ExecutionFailureHandler(
+		FailoverTopology topology,
+		FailoverStrategy.Factory failoverStrategyFactory,
+		RestartBackoffTimeStrategy.Factory restartStrategyFactory) {
+
+		this.failoverStrategy = checkNotNull(failoverStrategyFactory.create(topology));
+		this.restartBackoffTimeStrategy = checkNotNull(restartStrategyFactory.create());
+	}
+
+	/**
+	 * Return result of failure handling. Can be a set of task vertices to be restarted
+	 * and a delay of the restarting. Or that the failure is not recoverable and the reason for it.
+	 *
+	 * @param failedTask is the ID of the failed task vertex
+	 * @param cause of the task failure
+	 * @return result of the failure handling
+	 */
+	public FailureHandlingResult getFailureHandlingResult(ExecutionVertexID failedTask, Throwable cause) {
+		if (ThrowableClassifier.getThrowableType(cause) == ThrowableType.NonRecoverableError) {
+			return new FailureHandlingResult(new JobException("The failure is not recoverable", cause));
+		}
+
+		restartBackoffTimeStrategy.notifyFailure(cause);
+		if (restartBackoffTimeStrategy.canRestart()) {
+			return new FailureHandlingResult(
+				failoverStrategy.getTasksNeedingRestart(failedTask, cause),
+				restartBackoffTimeStrategy.getBackoffTime());
+		} else {
+			return new FailureHandlingResult(
+				new JobException("Failed task restarting is suppressed by " + restartBackoffTimeStrategy, cause));
+		}
+	}
+
+	@VisibleForTesting
+	FailoverStrategy getFailoverStrategy() {
+		return failoverStrategy;
+	}
+
+	@VisibleForTesting
+	RestartBackoffTimeStrategy getRestartBackoffTimeStrategy() {
+		return restartBackoffTimeStrategy;
+	}
+}
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/ExecutionFailureHandler.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/ExecutionFailureHandler.java
@@ -17,7 +17,6 @@
 
 package org.apache.flink.runtime.executiongraph.failover.flip1;
 
-import org.apache.flink.annotation.VisibleForTesting;
 import org.apache.flink.runtime.JobException;
 import org.apache.flink.runtime.scheduler.strategy.ExecutionVertexID;
 import org.apache.flink.runtime.throwable.ThrowableClassifier;
@@ -32,25 +31,23 @@ import static org.apache.flink.util.Preconditions.checkNotNull;
 public class ExecutionFailureHandler {
 
 	/** Strategy to judge which tasks should be restarted. */
-	private FailoverStrategy failoverStrategy;
+	private final FailoverStrategy failoverStrategy;
 
 	/** Strategy to judge whether and when a restarting should be done. */
-	private RestartBackoffTimeStrategy restartBackoffTimeStrategy;
+	private final RestartBackoffTimeStrategy restartBackoffTimeStrategy;
 
 	/**
 	 * Creates the handler to deal with task failures.
 	 *
-	 * @param topology of tasks of current job
-	 * @param failoverStrategyFactory is the factory to create the failover strategy
-	 * @param restartStrategyFactory is the factory to create the restart strategy
+	 * @param failoverStrategy helps to decide tasks to restart on task failures
+	 * @param restartBackoffTimeStrategy helps to decide whether to restart failed tasks and the restarting delay
 	 */
 	public ExecutionFailureHandler(
-		FailoverTopology topology,
-		FailoverStrategy.Factory failoverStrategyFactory,
-		RestartBackoffTimeStrategy.Factory restartStrategyFactory) {
+		FailoverStrategy failoverStrategy,
+		RestartBackoffTimeStrategy restartBackoffTimeStrategy) {
 
-		this.failoverStrategy = checkNotNull(failoverStrategyFactory.create(topology));
-		this.restartBackoffTimeStrategy = checkNotNull(restartStrategyFactory.create());
+		this.failoverStrategy = checkNotNull(failoverStrategy);
+		this.restartBackoffTimeStrategy = checkNotNull(restartBackoffTimeStrategy);
 	}
 
 	/**
@@ -63,27 +60,17 @@ public class ExecutionFailureHandler {
 	 */
 	public FailureHandlingResult getFailureHandlingResult(ExecutionVertexID failedTask, Throwable cause) {
 		if (ThrowableClassifier.getThrowableType(cause) == ThrowableType.NonRecoverableError) {
-			return new FailureHandlingResult(new JobException("The failure is not recoverable", cause));
+			return FailureHandlingResult.unrecoverable(new JobException("The failure is not recoverable", cause));
 		}
 
 		restartBackoffTimeStrategy.notifyFailure(cause);
 		if (restartBackoffTimeStrategy.canRestart()) {
-			return new FailureHandlingResult(
+			return FailureHandlingResult.restartable(
 				failoverStrategy.getTasksNeedingRestart(failedTask, cause),
 				restartBackoffTimeStrategy.getBackoffTime());
 		} else {
-			return new FailureHandlingResult(
+			return FailureHandlingResult.unrecoverable(
 				new JobException("Failed task restarting is suppressed by " + restartBackoffTimeStrategy, cause));
 		}
-	}
-
-	@VisibleForTesting
-	FailoverStrategy getFailoverStrategy() {
-		return failoverStrategy;
-	}
-
-	@VisibleForTesting
-	RestartBackoffTimeStrategy getRestartBackoffTimeStrategy() {
-		return restartBackoffTimeStrategy;
 	}
 }
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/ExecutionFailureHandler.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/ExecutionFailureHandler.java
@@ -17,10 +17,13 @@
 
 package org.apache.flink.runtime.executiongraph.failover.flip1;
 
+import org.apache.flink.annotation.VisibleForTesting;
 import org.apache.flink.runtime.JobException;
 import org.apache.flink.runtime.scheduler.strategy.ExecutionVertexID;
 import org.apache.flink.runtime.throwable.ThrowableClassifier;
 import org.apache.flink.runtime.throwable.ThrowableType;
+
+import java.util.Optional;
 
 import static org.apache.flink.util.Preconditions.checkNotNull;
 
@@ -59,7 +62,7 @@ public class ExecutionFailureHandler {
 	 * @return result of the failure handling
 	 */
 	public FailureHandlingResult getFailureHandlingResult(ExecutionVertexID failedTask, Throwable cause) {
-		if (ThrowableClassifier.getThrowableType(cause) == ThrowableType.NonRecoverableError) {
+		if (isUnrecoverableError(cause)) {
 			return FailureHandlingResult.unrecoverable(new JobException("The failure is not recoverable", cause));
 		}
 
@@ -72,5 +75,12 @@ public class ExecutionFailureHandler {
 			return FailureHandlingResult.unrecoverable(
 				new JobException("Failed task restarting is suppressed by " + restartBackoffTimeStrategy, cause));
 		}
+	}
+
+	@VisibleForTesting
+	static boolean isUnrecoverableError(Throwable cause) {
+		Optional<Throwable> unrecoverableError = ThrowableClassifier.findThrowableOfThrowableType(
+			cause, ThrowableType.NonRecoverableError);
+		return unrecoverableError.isPresent();
 	}
 }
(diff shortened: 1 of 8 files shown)


flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/failover/flip1/ExecutionFailureHandlerTest.java
```diff
@@ -0,0 +1,170 @@
...
+
+	// ------------------------------------------------------------------------
+	//  utilities
+	// ------------------------------------------------------------------------
+
+	private ExecutionFailureHandler createExecutionFailureHandler() {
+		FailoverTopology topology = new TestFailoverTopology.Builder().build();
+
+		FailoverStrategy.Factory failoverStrategyFactory = mock(FailoverStrategy.Factory.class);
+		when(failoverStrategyFactory.create(any(FailoverTopology.class))).thenReturn(mock(FailoverStrategy.class));
```
zentol: Please create a proper implementation of the FailoverStrategy. The factory can then be a simple lambda `topology -> new MyFailoverStrategy`.

Mocking things usually just leads to problems down the line.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/ExecutionFailureHandler.java
```diff
@@ -0,0 +1,89 @@
...
+				failoverStrategy.getTasksNeedingRestart(failedTask, cause),
+				restartBackoffTimeStrategy.getBackoffTime());
+		} else {
+			return new FailureHandlingResult(
+				new JobException("Failed task restarting is suppressed by " + restartBackoffTimeStrategy, cause));
+		}
+	}
+
+	@VisibleForTesting
+	FailoverStrategy getFailoverStrategy() {
```
zentol: this method (and `getRestartBackoffTimeStrategy`) seem unnecessary; any test using this class must pass factories for the strategies, and thus can already access the actual strategy instance.
This could be as simple as:
```
FailoverStrategy failoverStrategy = new MyFailoverStrategy(failoverTopology);
RestartBackoffTimeStrategy restartStrategy = new MyRestartBackoffTimeStrategy();

ExecutionFailureHandler handler = new ExecutionFailureHandler(
	failoverTopology,
	topology -> failoverStrategy,
	() -> restartStrategy);
```

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/failover/flip1/FailureHandlingResultTest.java
```diff
@@ -0,0 +1,79 @@
...
+ */
+public class FailureHandlingResultTest extends TestLogger {
+
+	/**
+	 * Tests normal FailureHandlingResult.
+	 */
+	@Test
+	public void testNormalFailureHandlingResult() throws Exception {
+		// create a normal FailureHandlingResult
+		Set<ExecutionVertexID> tasks = new HashSet<>();
```
zentol: let's add at least 1 value to this set, this test would currently pass even if the result would always return an empty set of vertices.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/failover/flip1/ExecutionFailureHandlerTest.java
```diff
@@ -0,0 +1,170 @@
...
+	/**
+	 * Tests the case that task restarting is accepted.
+	 */
+	@Test
+	public void testNormalFailureHandling() throws Exception {
+		ExecutionFailureHandler executionFailureHandler = createExecutionFailureHandler();
+		FailoverStrategy failoverStrategy = executionFailureHandler.getFailoverStrategy();
+		RestartBackoffTimeStrategy restartStrategy = executionFailureHandler.getRestartBackoffTimeStrategy();
+
+		Set<ExecutionVertexID> tasksToBeRestarted = new HashSet<>();
```
zentol: same as below, let's add at least 1 value.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/FailureHandlingResult.java
```diff
@@ -0,0 +1,108 @@
...
+	/** Reason why the failure is not recoverable. */
+	private Throwable error;
+
+	/**
+	 * Creates a result of a set of tasks to restart to recover from the failure.
+	 *
+	 * @param verticesToRestart containing task vertices to restart to recover from the failure
+	 * @param restartDelayMS indicate a delay before conducting the restart
+	 */
+	public FailureHandlingResult(Set<ExecutionVertexID> verticesToRestart, long restartDelayMS) {
```
zentol: when you have several constructors that heavily affect the semantics of the instance it is usually more readable to have static factory methods, like
```
public static FailureHandlingResult restartable(Set<ExecutionVertexID> verticesToRestart, long restartDelayMS)
public static FailureHandlingResult unrecoverable(Throwable error)
```

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/ExecutionFailureHandler.java
```diff
@@ -0,0 +1,89 @@
...
+import static org.apache.flink.util.Preconditions.checkNotNull;
+
+/**
+ * This handler deals with task failures to return a {@link FailureHandlingResult} which contains tasks
+ * to restart to recover from failures.
+ */
+public class ExecutionFailureHandler {
+
+	/** Strategy to judge which tasks should be restarted. */
+	private FailoverStrategy failoverStrategy;
```
eaglewatcherwb: can be final

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/ExecutionFailureHandler.java
```diff
@@ -0,0 +1,89 @@
...
+
+	/** Strategy to judge whether and when a restarting should be done. */
+	private RestartBackoffTimeStrategy restartBackoffTimeStrategy;
```
eaglewatcherwb: same as above

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/ExecutionFailureHandler.java
```diff
@@ -0,0 +1,89 @@
...
+	/**
+	 * Creates the handler to deal with task failures.
+	 *
+	 * @param topology of tasks of current job
+	 * @param failoverStrategyFactory is the factory to create the failover strategy
+	 * @param restartStrategyFactory is the factory to create the restart strategy
+	 */
+	public ExecutionFailureHandler(
+		FailoverTopology topology,
+		FailoverStrategy.Factory failoverStrategyFactory,
```
zentol: I'd pass in the actual instances instead of the factories. Then we wouldn't even need to explicitly pass the topology.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/FailureHandlingResult.java
```diff
@@ -0,0 +1,108 @@
...
+public class FailureHandlingResult {
+
+	/** Task vertices to restart to recover from the failure. */
+	private Set<ExecutionVertexID> verticesToRestart;
+
+	/** Delay before the restarting can be conducted. */
+	private long restartDelayMS;
+
+	/** Reason why the failure is not recoverable. */
+	private Throwable error;
```
zentol: these fields should be final, set them to null in one of the constructors

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/FailoverStrategy.java
```diff
@@ -34,4 +34,22 @@
 	 * @return set of IDs of vertices to restart
 	 */
 	Set<ExecutionVertexID> getTasksNeedingRestart(ExecutionVertexID executionVertexId, Throwable cause);
+
+	// ------------------------------------------------------------------------
+	//  factory
+	// ------------------------------------------------------------------------
+
+	/**
+	 * The factory to Instantiate {@link FailoverStrategy}.
```
eaglewatcherwb: Factory interface for {@link FailoverStrategy}.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/FailoverStrategy.java
```diff
@@ -34,4 +34,22 @@
 	 * @return set of IDs of vertices to restart
 	 */
 	Set<ExecutionVertexID> getTasksNeedingRestart(ExecutionVertexID executionVertexId, Throwable cause);
+
+	// ------------------------------------------------------------------------
+	//  factory
+	// ------------------------------------------------------------------------
```
eaglewatcherwb: position marker is not a good choice (see Clean Code by Robert C Martin, page 67, Chapter 4).

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/FailureHandlingResult.java
```diff
@@ -0,0 +1,108 @@
...
+	}
+
+	/**
+	 * Returns the tasks to restart.
+	 *
+	 * @return the tasks to restart
+	 */
+	public Set<ExecutionVertexID> getVerticesToRestart() {
+		if (canRestart()) {
+			return verticesToRestart;
```
eaglewatcherwb: return Collections.unmodifiableSet(verticesToRestart);

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/RestartBackoffTimeStrategy.java
```diff
@@ -0,0 +1,62 @@
...
+	/**
+	 * Notify the strategy about the task failure cause.
+	 *
+	 * @param cause of the task failure
+	 */
+	void notifyFailure(Throwable cause);
+
+	// ------------------------------------------------------------------------
+	//  factory
+	// ------------------------------------------------------------------------
```
eaglewatcherwb: see the comment in FailoverStrategy.java

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/failover/flip1/ExecutionFailureHandlerTest.java
```diff
@@ -0,0 +1,170 @@
...
+/**
+ * Tests for {@link ExecutionFailureHandler}.
+ */
+public class ExecutionFailureHandlerTest extends TestLogger {
+
+	/**
+	 * Tests the case that task restarting is accepted.
+	 */
+	@Test
+	public void testNormalFailureHandling() throws Exception {
```
eaglewatcherwb: Exception is not thrown.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/failover/flip1/ExecutionFailureHandlerTest.java
```diff
@@ -0,0 +1,170 @@
...
+		assertEquals(restartDelayMs, result.getRestartDelayMS());
+		assertEquals(tasksToBeRestarted, result.getVerticesToRestart());
+		assertNull(result.getError());
+	}
+
+	/**
+	 * Tests the case that task restarting is suppressed.
+	 */
+	@Test
+	public void testRestartingSuppressedFailureHandlingResult() throws Exception {
```
eaglewatcherwb: Exception is not thrown.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/failover/flip1/ExecutionFailureHandlerTest.java
```diff
@@ -0,0 +1,170 @@
...
+		} catch (IllegalStateException ex) {
+			// expected
+		}
+	}
+
+	/**
+	 * Tests the case that the failure is non-recoverable type.
+	 */
+	@Test
+	public void testNonRecoverableFailureHandlingResult() throws Exception {
```
eaglewatcherwb: Exception is not thrown.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/failover/flip1/FailureHandlingResultTest.java
```diff
@@ -0,0 +1,79 @@
...
+/**
+ * Tests for {@link FailureHandlingResult}.
+ */
+public class FailureHandlingResultTest extends TestLogger {
+
+	/**
+	 * Tests normal FailureHandlingResult.
+	 */
+	@Test
+	public void testNormalFailureHandlingResult() throws Exception {
```
eaglewatcherwb: same as above.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/failover/flip1/FailureHandlingResultTest.java
```diff
@@ -0,0 +1,79 @@
...
+		assertEquals(delay, result.getRestartDelayMS());
+		assertEquals(tasks, result.getVerticesToRestart());
+		assertNull(result.getError());
+	}
+
+	/**
+	 * Tests FailureHandlingResult which suppresses restarts.
+	 */
+	@Test
+	public void testRestartingSuppressedFailureHandlingResult() throws Exception {
```
eaglewatcherwb: same as above.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/FailoverStrategy.java
zhuzhurk: Sorry but I don't have that book.
Let's just keep it as this marker usage is widely used in Flink.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/FailureHandlingResult.java
zhuzhurk: I think it's not necessary to add extra cost to do this conversion.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/FailureHandlingResult.java
zhuzhurk: Good idea.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/ExecutionFailureHandler.java
zhuzhurk: Fine.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/ExecutionFailureHandler.java
zhuzhurk: Ok.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/FailureHandlingResult.java
GJL: Instances of this class can be made immutable and I don't see a reason not to.

zhuzhurk: Ok. Let's make it immutable.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/failover/flip1/FailureHandlingResultTest.java
zhuzhurk: OK.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/failover/flip1/ExecutionFailureHandlerTest.java
zhuzhurk: Ok.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/FailureHandlingResult.java
zentol: nit: we could make the `verticesToRestart` field immutable instead and return it without additional wrappers.

zhuzhurk: Ok.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/failover/flip1/FailureHandlingResultTest.java
zentol: You missed this one :)

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/ExecutionFailureHandler.java
```diff
@@ -0,0 +1,76 @@
...
+	/**
+	 * Return result of failure handling. Can be a set of task vertices to restart
+	 * and a delay of the restarting. Or that the failure is not recoverable and the reason for it.
+	 *
+	 * @param failedTask is the ID of the failed task vertex
+	 * @param cause of the task failure
+	 * @return result of the failure handling
+	 */
+	public FailureHandlingResult getFailureHandlingResult(ExecutionVertexID failedTask, Throwable cause) {
+		if (ThrowableClassifier.getThrowableType(cause) == ThrowableType.NonRecoverableError) {
```
zentol: according to the [scheduler interface draft](https://docs.google.com/document/d/1fstkML72YBO1tGD_dmG2rwvd9bklhRVauh4FSsDDwXU/edit#), the `RestartBackoffTimeStrategy` is responsible for this decision.

```The delay is computed by a RestartBackoffTimeStrategy that is also responsible for deciding whether a restart should be performed at all (e.g., if the failure cause is SuppressRestartsException).```

File: flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/FailureHandlingResult.java [hunk 20]
```diff
@@ -0,0 +1,133 @@
...
+		return error == null;
+	}
+
+	/**
+	 * Returns reason why the restarting cannot be conducted.
+	 *
+	 * @return reason why the restarting cannot be conducted
+	 */
+	public Throwable getError() {
+		return error;
```
zentol: shouldn't this also throw an `IllegalStateException` if `error == null` like `getRestartDelayMS` and `getVerticesToRestart`?

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/ExecutionFailureHandler.java
zhuzhurk: Yes the design doc proposed to check the error type in the restart strategy. But that requires every strategy implementation to do the same check.

Therefore I think it's better we let the restart strategy do it's dedicated check only. 

Besides, I'm also thinking that whether we can have a composed restart strategy, which can do multiple checks for failing max limit, failing rate, etc.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/FailureHandlingResult.java
zhuzhurk: You are right.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/failover/flip1/FailureHandlingResultTest.java
zhuzhurk: My fault.

PR #8509, opened 2019-05-22
[FLINK-12458][network] Introduce PartitionConnectionException for unreachable producer

zhijiangW:

* If the consumer can not establish a connection to remote task executor while requesting remote subpartition, which might indicate the remote task executor is not reachable. We could wrap this connection exception into new proposed `PartitionConnectionException` which also extends `PartitionException`, then the job master would decide whether to restart the upstream region to re-producer partition data.*

## Brief change log

  - *Introduce new `PartitionConnectionException` extends `PartitionException.*
  - *Wrap any exceptions into `PartitionConnectionException` while creating partition request client during `RemoteInputChannel#requestSubpartition`*

## Verifying this change

  - *Added new tests in `RemoteInputChannelTest` for verifying the `PartitionConnectionException` while requesting sub partition.*

## Documentation


zhijiangW:

tillrohrmann:
The PR does not compile @zhijiangW.

zhijiangW:
Thanks for kindly reminding, it is because of unused import checkstyle. I already fixed the issue.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/consumer/RemoteInputChannel.java
```diff
@@ -159,11 +159,15 @@ void assignExclusiveSegments(List<MemorySegment> segments) {
...
 	@Override
-	public void requestSubpartition(int subpartitionIndex) throws IOException, InterruptedException {
+	public void requestSubpartition(int subpartitionIndex) throws IOException {
 		if (partitionRequestClient == null) {
 			// Create a client and request the partition
-			partitionRequestClient = connectionManager
-				.createPartitionRequestClient(connectionId);
+			try {
+				partitionRequestClient = connectionManager.createPartitionRequestClient(connectionId);
+			} catch (Throwable t) {
```
tillrohrmann: I think we should only catch `IOException` here. If a different exception is thrown, this indicates a serious Flink problem and we should best fail the `TaskExecutor` process.

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/consumer/RemoteInputChannelTest.java
```diff
@@ -1047,6 +1047,22 @@ public void testPartitionNotFoundExceptionWhileRetriggeringRequest() throws Exce
...
+	 * Tests that any exceptions thrown by {@link ConnectionManager#createPartitionRequestClient(ConnectionID)}
+	 * would be wrapped into {@link PartitionConnectionException} during
+	 * {@link RemoteInputChannel#requestSubpartition(int)}.
+	 */
+	@Test
+	public void testPartitionConnectionExceptionWhileRequestingPartition() throws Exception {
+		final RemoteInputChannel inputChannel = InputChannelTestUtils.createRemoteInputChannel(
+			createSingleInputGate(1), 0, new TestingExceptionConnectionManager());
+		try {
+			inputChannel.requestSubpartition(0);
```
tillrohrmann: fail is missing

zhijiangW: fail is not missing on my side?

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/consumer/RemoteInputChannel.java
zhijiangW: Only the `InterruptedException` might be thrown during `future.waitForChannel()`. If we only catch the `IOException`, then for other cases we could keep the previous behavior to throw the raw exception directly?

zhijiangW:
@tillrohrmann , thanks for the review!
If I understood correctly, it does not need to rebase master on my side?

tillrohrmann:
Yes, I pushed the changes to your branch. No need for any actions from your side.

PR #8510, opened 2019-05-22
 [FLINK-12254][table] Update cast() and TypeLiteralExpression to new type system

twalthr:

This PR updates `ImplicitExpressionOperations#cast` and `TypeLiteralExpression`. The Java expression API remains untouched for now.

This PR builds on top of #8500.

## Brief change log

- Deprecate old `cast()`
- Update `TypeLiteralExpression`
- Use legacy conversion where applicable

## Verifying this change

  - The serializers: no

## Documentation

### DIFF
--- a/flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/expressions/ApiExpressionUtils.java
+++ b/flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/expressions/ApiExpressionUtils.java
@@ -23,13 +23,11 @@ import org.apache.flink.api.common.typeinfo.BasicTypeInfo;
 import org.apache.flink.api.common.typeinfo.TypeInformation;
 import org.apache.flink.table.api.Table;
 import org.apache.flink.table.api.ValidationException;
+import org.apache.flink.table.types.DataType;
 import org.apache.flink.table.typeutils.TimeIntervalTypeInfo;
 
 import java.util.Arrays;
 import java.util.Optional;
-
-import static org.apache.flink.table.expressions.BuiltInFunctionDefinitions.CAST;
-import static org.apache.flink.table.expressions.BuiltInFunctionDefinitions.TIMES;
 
 /**
  * Utilities for API-specific {@link Expression}s.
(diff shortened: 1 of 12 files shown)


flink-table/flink-table-planner/src/main/scala/org/apache/flink/table/api/scala/expressionDsl.scala
```diff
@@ -262,15 +264,22 @@ trait ImplicitExpressionOperations {
...
+    * e.g. "42".cast(DataTypes.INT()) leads to 42.
     *
     * @return casted expression
     */
-  def cast(toType: TypeInformation[_]): Expression =
+  def cast(toType: DataType): Expression =
     call(CAST, expr, typeLiteral(toType))
 
+  /**
+    * @deprecated Use [[cast(DataType)]] instead.
```
dawidwys: nit: Could we add a bit longer note that it might be unsafe to mix the `DataType` and `TypeInformation` stacks?

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/expressions/ApiExpressionUtils.java
```diff
@@ -85,17 +83,7 @@ public static Expression toMonthInterval(Expression e, int multiplier) {
 		// check for constant
 		return ExpressionUtils.extractValue(e, BasicTypeInfo.INT_TYPE_INFO)
 			.map((v) -> (Expression) valueLiteral(v * multiplier, TimeIntervalTypeInfo.INTERVAL_MONTHS))
-			.orElse(
```
dawidwys: This changes the behavior of this method right? Previously e.g. `(1 + 3).toMonthInterval` or even `(1 + f0).toMonthInterval` was possible right? (Even though e.g. the window aggregation did not accept it)

twalthr: Yes, that's right but this behavior was not even tested somewhere and for simplification of the design I think it is ok to limit the values to constants. What do you think?

dawidwys: I agree, just wanted to double check.

twalthr:
Thanks for the feedback @dawidwys. I updated the PR. Will merge this once Travis is green.

dawidwys:
@flinkbot approve all

PR #8511, opened 2019-05-22
[FLINK-12319][Library/CEP]Change the logic of releasing node from recursive to non-recursive

liyafan82:

Resolve issue FlINK-12319. The CEP job throws StackOverFlowError when releasing the nodes in the SharedBuffer. 

After investigation, it shows that the reason is that the logic for releasing nodes in the SharedBuffer is a recursive algorithm. The depth of the recursion depends on the diameter of the graph, which is proportional the number of buffered events (the graph is a line graph). 
In the experiments, the number of buffered events can be as many as 100,000, which will easily cause stack overflow.

Therefore, in this PR we change the logic from recursive to non-recursive. 

## Brief change log

  - Change the SharedAccessBuffer#releaseNode to a non-recursive algorithm.
  

## Verifying this change

This change is already covered by existing tests, such as SharedBufferTest#testSharedBuffer and SharedBufferTest#testClearingSharedBufferWithMultipleEdgesBetweenEntries.

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-libraries/flink-cep/src/main/java/org/apache/flink/cep/nfa/sharedbuffer/SharedBufferAccessor.java
+++ b/flink-libraries/flink-cep/src/main/java/org/apache/flink/cep/nfa/sharedbuffer/SharedBufferAccessor.java
@@ -247,9 +247,7 @@ public class SharedBufferAccessor<V> implements AutoCloseable {
 				break;
 			}
 
-			if (!curBufferNode.release()) {
-				sharedBuffer.upsertEntry(curNode, curBufferNode);
-			} else {
+			if (curBufferNode.release()) {
 				// first release the current node
 				sharedBuffer.removeEntry(curNode);
 				releaseEvent(curNode.getEventId());
@@ -260,6 +258,8 @@ public class SharedBufferAccessor<V> implements AutoCloseable {
 						nodesToExamine.push(targetId);
 					}
 				}
+			} else {
+				sharedBuffer.upsertEntry(curNode, curBufferNode);
 			}
 		}
 	}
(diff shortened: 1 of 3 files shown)


rmetzger:
Thank you for fixing this bug.
Does it make sense to add the job triggering this bug as a test, so that we won't introduce it in the future again?

liyafan82:
> Thank you for fixing this bug.
> Does it make sense to add the job triggering this bug as a test, so that we won't introduce it in the future again?

@rmetzger thanks a lot for the good suggestion. We have added a test case for that. Please help take a look.

liyafan82:
Hi @tillrohrmann , would you please take a look at this issue? I think it deserves to be fixed.

flink-libraries/flink-cep/src/main/java/org/apache/flink/cep/nfa/sharedbuffer/SharedBufferAccessor.java
```diff
@@ -235,31 +235,32 @@ public void lockNode(final NodeId node) {
...
+		// the stack used to detect all nodes that needs to be released.
+		Stack<NodeId> nodesToExamine = new Stack<>();
+		nodesToExamine.push(node);
+
+		while (!nodesToExamine.isEmpty()) {
+			NodeId curNode = nodesToExamine.pop();
+			Lockable<SharedBufferNode> curBufferNode = sharedBuffer.getEntry(curNode);
+
+			if (curBufferNode == null) {
+				break;
```
dawidwys: What does it mean if the `currBufferNode` is null? When can it occur? Shouldn't we rather `continue` to clear all elements from `nodesToExamine`?

liyafan82: @dawidwys Good question. Thanks.

There can be duplicated attempts to remove a node. So if we do not check whether currBufferNode is null, an NPE will be thrown. 

As an example, please take a look at CEPITCase#testFlatSelectSerializationWithAnonymousClass. There are two attempts to release the node from NFA#doProcess method. 

When a node is released (and set to null), the decedent nodes have already been released too, so there is no need to  continue examining nodes in nodesToExamine.

Please note that such logic is not designed by me. The original code worked the same way. I just refactor the code so that the logic is identical to the original code.

dawidwys:
Thank you for your contribution @liyafan82 Merging it.

PR #8513, opened 2019-05-22
 [FLINK-12147] [metrics] Error for monitor kafka metrics when Use influxDB metr…

theoDiefenthal:

As InfluxDB doesn't support infinity values in float/double, we just don't send them to influx . There is no behavioral change here as InfluxDB wouldn't store those values anyways, but we are not going to pollute our logs with influxdb partial write errors about each sent infinity value on each report.

## Brief change log

When submitting metrics, check the current metric values and ignore them if they are +-infinity, so that they are not sent. 

## Verifying this change

This change is already covered by existing tests, such as InfluxDBReporter.testMetricReporting . The test was extended in order to make sure that the invalid values are filtered out correctly. 

  - The serializers: no

## Documentation

### DIFF
--- a/flink-metrics/flink-metrics-influxdb/src/main/java/org/apache/flink/metrics/influxdb/InfluxPointBuilder.java
+++ b/flink-metrics/flink-metrics-influxdb/src/main/java/org/apache/flink/metrics/influxdb/InfluxPointBuilder.java
@@ -1,3 +1,21 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package org.apache.flink.metrics.influxdb;
 
 import org.influxdb.dto.Point;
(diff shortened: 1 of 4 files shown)


1u0:
Hi @theoDiefenthal,
is it possible to delegate such filter to [`influxdb-java` client](https://github.com/influxdata/influxdb-java)?

Imo, it would be nice if the client handles such corner cases: it's specific to InfluxDB, and other projects that use the java client may face the same issue.

theoDiefenthal:
IMHO the correct place for this is here as InfluxDB is most widely used for monitoring, but not exclusively. I personally worked with InfluxDB for another purpose where I wanted those errors to arise. But one could of course avoid the roundtrip by letting the InfluxDB client directly throw such an exception... 

Further, this still is a workaround as InfluxDB dev team already tracks an issue to to support infinity values in future. [ https://github.com/influxdata/influxdb/issues/4089 , https://github.com/influxdata/flux/issues/1040 ]. The InfluxDB client should thus not be limited in terms of not sending the numbers.

1u0:
The discussion and effort in InfluxDB project only confirm that one day InfluxDB service (in such future version) may accept NaN and infinite values as valid. (I consider this situation as an argument that influxdb client should handle it).

If we land your changes here, delegating such filtering to library user, then later, we would have to add another patch to pass those values to newer InfluxDB versioned services that support those values.
This would involve InfluxDB service version checking, or dumping decision to Flink users via configuration parameter.

Alternatively:
as the client library user, I expect that the (official) library alleviates differences and feature capabilities of different server apis, so I can use the same user code against different server versions.
The `influxdb-java` client **could** have added some configuration option, that would influence api mode:
 - "best effort" mode (desired here, in Flink influxdb reporter), when client silently drops measurement points if it knows that the backend service doesn't support values;
 - "transparent" mode (current client's behavior);
 - "strict" mode, when client throws exception on client side, if it knows that the backend service doesn't support some values.

I encourage you to push such change in the `influxdb-java` client, rather than having a workaround here.
If such change is adopted there, we can just bump corresponding client version here and add configuration (in code) if needed.

Consider adding workaround here only as the last option.

theoDiefenthal:
I see your point here. I ask InfluxDB for introducing an option to just ignore infinity values on database level. ( https://github.com/influxdata/influxdb/issues/13986 ). I think this would solve the problem in an even better way as other clients could also benefit from the setting.

Crack:
`influxdb-java` 2.16 is available, after upgrade NaNs won't be sent anymore - they will just get silently dropped.

zentol:
Subsumed by #10020.

PR #8517, opened 2019-05-23
[FLINK-10921] [kinesis] Shard watermark synchronization in Kinesis consumer

tweise:

* This pull request makes adds support for source synchronization to the Kinesis consumer. Source synchronization aligns shard consumption based on the per-shard watermark.

## Brief change log

- Adds global watermark aggregation to track the current low watermark across subtasks
- Adds record emitter to prioritize per-shard queues and align the shard consumers through back pressure
- Above components are designed to be reusable for the upcoming support in the Kafka consumer.
- Integration into KinesisDataFetcher 

## Verifying this change

- This is a back port from Lyft internal codebase where it has been tested extensively and at scale.
- PR includes test for consumer and basic tests for emitter and watermark tracker.

  - The serializers: (no)

## Documentation

  - If yes, how is the feature documented? (user documentation will be added separately)

### DIFF
--- a/flink-connectors/flink-connector-kinesis/src/main/java/org/apache/flink/streaming/connectors/kinesis/util/RecordEmitter.java
+++ b/flink-connectors/flink-connector-kinesis/src/main/java/org/apache/flink/streaming/connectors/kinesis/util/RecordEmitter.java
@@ -1,0 +1,269 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.streaming.connectors.kinesis.util;
+
+import org.apache.flink.annotation.Internal;
+import org.apache.flink.streaming.runtime.operators.windowing.TimestampedValue;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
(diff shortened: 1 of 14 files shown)


flink-connectors/flink-connector-kinesis/src/main/java/org/apache/flink/streaming/connectors/kinesis/util/RecordEmitter.java
```diff
@@ -0,0 +1,268 @@
...
+	}
+
+	public void stop() {
+		running = false;
+	}
+
+	/** Emit the record. This is specific to a connector, like the Kinesis data fetcher. */
+	public abstract void emit(T record, RecordQueue<T> source);
+
+	// TODO: metrics integration
```
jgrier: Why not do this now?

flink-connectors/flink-connector-kinesis/src/main/java/org/apache/flink/streaming/connectors/kinesis/util/RecordEmitter.java
```diff
@@ -0,0 +1,268 @@
...
+			"[setCurrentWatermark] Current watermark set to {}, maxEmitTimestamp = {}",
+			watermark,
+			maxEmitTimestamp);
+	}
+
+	/**
+	 * Run loop, does not return unless {@link #stop()} was called.
+	 */
+	@Override
+	public void run() {
```
jgrier: It would be nice if the main runLoop could be refactored into smaller well-named methods just to make the code easier to follow.

jgrier:
@flinkbot approve all

flink-connectors/flink-connector-kinesis/src/main/java/org/apache/flink/streaming/connectors/kinesis/util/RecordEmitter.java
tweise: This was added for logging. There are 2 metrics available (local watermark and current global watermark). That was sufficient during testing, we can add more in the future as we learn more about the observability needs. I'm going to remove the comment since it's misleading.

flink-connectors/flink-connector-kinesis/src/main/java/org/apache/flink/streaming/connectors/kinesis/util/RecordEmitter.java
tweise: That seems possible, although given the moderate size of the run method and the need to understand it in its entirety I'm not sure how much we would gain. It would be better made as separate change anyways.

PR #8518, opened 2019-05-23
[FLINK-12564][network] Remove getBufferProvider method from ResultPartitionWriter

zhijiangW:

*`ResultPartitionWriter#getBufferProvider` seems not very general for all the writer implementations. So this method could be merged with `ResultPartitionWriter#addBufferConsumer` to refactor as `ResultPartitionWriter#requestBufferBuilder` which returns a `BufferBuilder` for writing data and the corresponding `BufferConsumer` is added into sub partition.*

## Brief change log

  - *Remove `getBufferProvider` from `ResultPartitionWriter`*
  - *Introduce `requestBufferBuilder` and `broadcastEvents` for `ResultPartitionWriter`*
  - *Refactor the process in `RecordWriter`*

## Verifying this change

covered by existing tests

## Documentation

### DIFF
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/ResultPartition.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/ResultPartition.java
@@ -18,7 +18,6 @@
 
 package org.apache.flink.runtime.io.network.partition;
 
-import org.apache.flink.annotation.VisibleForTesting;
 import org.apache.flink.api.common.JobID;
 import org.apache.flink.runtime.executiongraph.IntermediateResultPartition;
 import org.apache.flink.runtime.io.network.api.writer.ResultPartitionWriter;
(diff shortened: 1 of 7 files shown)


zhijiangW:

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/ResultPartition.java
```diff
@@ -188,7 +190,7 @@ public int getNumberOfSubpartitions() {
 		return subpartitions.length;
 	}
 
-	@Override
+	@VisibleForTesting
 	public BufferProvider getBufferProvider() {
```
zentol: can this be package private?

azagrebin: we can actually remove it and use `getBufferPool` in `LocalInputChannelTest` where it is used.

zhijiangW: It could not be package private, but we could remove it to use `getBufferPool` instead. I could submit a fixup for it.

zhijiangW:
@zentol Thanks for reviews and I pushed a fixup commit for removing `getBufferProvider` from `ResultPartition`.

PR #8520, opened 2019-05-23
[FLINK-12600] [table-planner-blink] Introduce planner rules to do deterministic rewriting on RelNode

godfreyhe:

*Introduce planner rules to do deterministic rewriting on RelNode*

## Brief change log

  - *FlinkLimit0RemoveRule, that rewrites `limit 0` to empty Values*
  - *FlinkRewriteSubQueryRule, that rewrites a Filter with condition: `(select count from T) > 0` to a Filter with condition: `exists(select * from T)`, which could be converted to SEMI Join by FlinkSubQueryRemoveRule*
  - *ReplaceIntersectWithSemiJoinRule, that rewrites distinct Intersect to a distinct Aggregate on a SEMI Join*
  - *ReplaceMinusWithAntiJoinRule, that rewrites distinct Minus to a distinct Aggregate on an ANTI Join*
  - *introduce FlinkPruneEmptyRules#JOIN_RIGHT_INSTANCE to handle ANTI join with empty right*

## Verifying this change

  - *Added FlinkLimit0RemoveRuleTest that validates the logical plan result after FlinkLimit0RemoveRule is applied, and added Limit0RemoveITCase that validates the execution result*
  - *Added FlinkRewriteSubQueryRuleTest that validates the logical plan results after FlinkRewriteSubQueryRule is applied*
  - *Added ReplaceIntersectWithSemiJoinRuleTest that validates the logical plan results after ReplaceIntersectWithSemiJoinRule is applied*
  - *Added ReplaceMinusWithAntiJoinRuleTest that validates the logical plan results after ReplaceMinusWithAntiJoinRule is applied*
  - *Added FlinkPruneEmptyRulesTest that validates the logical plan results after FlinkPruneEmptyRules is applied*
  - *Added SetOperatorsTest that validates the physical plan results after ReplaceMinusWithAntiJoinRule and ReplaceIntersectWithSemiJoinRuleTest are applied*

## Documentation


flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/plan/rules/logical/FlinkPruneEmptyRules.scala
```diff
@@ -0,0 +1,70 @@
...
+ */
+
+package org.apache.flink.table.plan.rules.logical
+
+import org.apache.calcite.plan.RelOptRule.{any, none, operand, some}
+import org.apache.calcite.plan.{RelOptRule, RelOptRuleCall}
+import org.apache.calcite.rel.RelNode
+import org.apache.calcite.rel.core.{Join, JoinRelType, Values}
+
+object FlinkPruneEmptyRules {
```
KurtYoung: you intend to introduce more than one rule in this object?

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/plan/rules/logical/FlinkRewriteSubQueryRule.scala
```diff
@@ -0,0 +1,154 @@
...
+import org.apache.calcite.sql.SqlKind
+import org.apache.calcite.sql.`type`.SqlTypeFamily
+import org.apache.calcite.sql.fun.SqlCountAggFunction
+import org.apache.calcite.tools.RelBuilderFactory
+
+import scala.collection.JavaConversions._
+
+/**
+  * Planner rule that rewrites filter condition like:
+  * `(select count(*) from T) > 0` to `exists(select * from T)`,
```
KurtYoung: why semi join is better than aggregate?
BTW, i think for query like this, a more efficient one is convert to `exists(select * from T limit 1)`

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/plan/rules/logical/FlinkPruneEmptyRules.scala
godfreyhe: `PruneEmptyRules` in Calcite contains more than one rules, I intend to keep calcite style for this rule.  Maybe later, we need to copy other rules from `PruneEmptyRules` to this file.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/plan/rules/logical/FlinkRewriteSubQueryRule.scala
godfreyhe: The estimation for SEMI/ANTI join is very inaccurate，so we intend to do deterministic rewriting on SEMI/ANTI join. And we can put this rule to CBO after we improve estimation of SEMI/ANTI join.

yes, we can do similarly rewriting for `exists(select * from T limit 1)`.

KurtYoung: I'm confused, the original query is aggregate, but you want to convert it to a semi join, knowing that the estimation of semi join is not reliable. 
Take a step back, why semi join is better than aggregate in this case?

godfreyhe: the original query is a scalar query, which will be converted to aggregate + join by Calcite rules.
for example, full sql like `SELECT * FROM x WHERE (SELECT COUNT(*) FROM y WHERE d > 10) > 0` ，the logical plan converted by Calcite rules is:

`
LogicalProject(a=[$0], b=[$1], c=[$2])
+- LogicalProject(a=[$0], b=[$1], c=[$2])
   +- LogicalFilter(condition=[>($3, 0)])
      +- LogicalJoin(condition=[true], joinType=[left])
         :- LogicalTableScan(table=[[x]])
         +- LogicalAggregate(group=[{}], EXPR$0=[COUNT()])
            +- LogicalProject($f0=[0])
               +- LogicalFilter(condition=[>($0, 10)])
                  +- LogicalTableScan(table=[[y]])
`

PR #8526, opened 2019-05-24
[FLINK-12597][ml] Remove the legacy flink-libraries/flink-ml

c4emmmm:

This pull request removes the legacy flink-ml and flink-ml-uber in flink-libraries. They have been not maintained for long time and there is also not any active commit planned. Since we've reached the consensus on the new flink-ml framework proposed in FLIP-39, it's better to remove them in case of confusion.
Existing users can still use previous versions.

## Brief change log

  - Remove flink-libraries/flink-ml, flink-libraries/flink-ml-uber, and their registration in pom.xml in flink-libraries
  - Remove dependency of flink-ml-uber in flink-dist, and its registration in opt.xml for maven-assembly-plugin
  - Docs about flink-ml are temporally retained

## Verifying this change

This change only remove existing modules, which are independent from most other modules. This change can be varified if other existing tests can still pass.

  - The serializers: (no)

## Documentation


zentol:
We have to update the documentation somehow. Either  remove it entirely, blank it if you want to re-use the pages later on, or add a note that this library is no longer contained in flink releases and has to be downloaded manually.

c4emmmm:
> We have to update the documentation somehow. Either remove it entirely, blank it if you want to re-use the pages later on, or add a note that this library is no longer contained in flink releases and has to be downloaded manually.

Sorry that I missed your feedback and reply so late.
This pr is temporally hung because the new flink-ml module has been named flink-ml-parent, and there's no more conflict now. But it's still better to remove the flink-libaraies/flink-ml since it has been unmaintained for long time and may confuse users. 
I'll continue the work of this pr later including updating the documents. Thanks for your advice.

shaoxuan-wang:
> We have to update the documentation somehow. Either remove it entirely, blank it if you want to re-use the pages later on, or add a note that this library is no longer contained in flink releases and has to be downloaded manually.

Very good point, @c4emmmm , can you please consider this suggestion and make some actions on the existing docs. Maybe you can just include the changes made in [FLINK-12928](https://github.com/apache/flink/pull/8827). We also need to open a JIRA to update flink web later.

c4emmmm:
> Very good point, @c4emmmm , can you please consider this suggestion and make some actions on the existing docs. Maybe you can just include the changes made in [FLINK-12928](https://github.com/apache/flink/pull/8827). We also need to open a JIRA to update flink web later.

Now that @sjwiesman has already proposed the PR, and it seems to be what we need to do, I suppose that we can simply reopen and accept that PR as soon as we accept this PR.
There's a problem that in [FLINK-12928](https://github.com/apache/flink/pull/8827) the descriptions and links of legacy FlinkML are only redirected rather than removed, which may confuse users. I can suggest some changes to the PR before it is merged, or take over it if @sjwiesman currently has no time to do this. What do you think? @sjwiesman

sjwiesman:
@c4emmmm Agreed, fell free to pull in my commit from #8827 or I can reopen the PR. As part of FLIP-42 we are restructuring the docs and decided to fully remove the old FlinkML documentation. If users are still using this library from an old version, they can also use an old version of the docs. I would be opposed to adding an "under construction page", I find that more confusing.

c4emmmm:
> @c4emmmm Agreed, fell free to pull in my commit from #8827 or I can reopen the PR. As part of FLIP-42 we are restructuring the docs and decided to fully remove the old FlinkML documentation. If users are still using this library from an old version, they can also use an old version of the docs. I would be opposed to adding an "under construction page", I find that more confusing.

Thanks. I also oppose that with deeper thought so I mark it with delete line.

Since the modification is almost done in #8827, I think it's better to reopen the PR and merge it. 
The only docs that mentioning FlinkML are docs/internals/components.md and components.zh.md. You can find the keyword "FlinkML" and decide whether to remove them.

@sjwiesman , would you please spend a little time doing this if @shaoxuan-wang and @zentol agree we merge this and #8827 separately? And what do you think? @shaoxuan-wang @zentol

shaoxuan-wang:
Let us merge this PR first, then reopen the #8827 .  @c4emmmm @sjwiesman 

@flinkbot approve all

PR #8527, opened 2019-05-24
[FLINK-12610] [table-planner-blink] Introduce planner rules about aggregate

godfreyhe:

*Introduce planner rules about aggregate*

## Brief change log

  - *AggregateReduceGroupingRule, that reduces unless grouping columns*
  - *PruneAggregateCallRule, that removes unreferenced AggregateCall from Aggregate*
  - *FlinkAggregateRemoveRule, that is copied from Calcite's AggregateRemoveRule, and supports SUM, MIN, MAX, AUXILIARY_GROUP functions in non-empty group aggregate*
  - *FlinkAggregateJoinTransposeRule, that is copied from Calcite's AggregateJoinTransposeRule, and supports Left/Right outer join and aggregate with AUXILIARY_GROUP*

## Verifying this change

  - *Added rule tests that validate the logical plan after the added rules are applied*
  - *Added integration tests that validate the execution result the added rules are applied*

## Documentation


flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/plan/rules/logical/AggregateCalcMergeRule.java
```diff
@@ -0,0 +1,86 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Planner rule that recognizes a {@link org.apache.calcite.rel.core.Aggregate}
+ * on top of a {@link org.apache.calcite.rel.core.Calc} and if possible
+ * aggregate through the calc or removes the calc.
+ *
+ * <p>This is only possible when no condition in calc and the grouping expressions and arguments to
```
KurtYoung: If you can't merge condition, why not change the rule match pattern to `aggregate on project`?

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/plan/rules/logical/FlinkAggregateJoinTransposeRule.java
```diff
@@ -0,0 +1,593 @@
...
+import java.util.TreeMap;
+
+import scala.Tuple2;
+import scala.collection.JavaConverters;
+import scala.collection.Seq;
+
+/**
+ * This rule is copied from Calcite's {@link org.apache.calcite.rel.rules.AggregateJoinTransposeRule}.
+ * Modification:
+ * - Do not match TemporalTableScan since it means that it is a dimension table scan currently.
```
KurtYoung: Put this rule into CBO, and we can avoid to deal with such situation.

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/plan/rules/logical/AggregateCalcMergeRule.java
godfreyhe: `AggregateProjectMergeRule` already exists and is in our rule set. The original intention of introducing this rule is to solve the hack in `RelDecorrelator` for tpch query20, however current cost model can not find best plan, so this rule has not been add to our rule set.  we can move this rule from this pr, and introduce it when needed

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/plan/rules/logical/FlinkAggregateJoinTransposeRule.java
godfreyhe: this rule is already in CBO, and CBO does not handle the unspported case (lookup table source doesn't support aggregate).

PR #8529, opened 2019-05-24
[FLINK-12603][network] Remove getOwningTaskName method from InputGate

zhijiangW:

*In order to make abstract `InputGate` simple for extending new implementations in shuffle service architecture, we could remove unnecessary methods from it.*

*Currently `InputGate#getOwningTaskName` is only used for debugging log in `BarrierBuffer` and `StreamInputProcessor`. This task name could be got from `Environment#getTaskInfo` and then be passed into the constructor of `BarrierBuffer/StreamInputProcessor` for use.*

## Brief change log

  - *Add the full task name in the `TaskInfo`*
  - *Remove `getOwningTaskName`  from `InputGate` abstract methods*
  - *Refactor the related process to get task name for `BarrierBuffer` and `StreamInputProcessor`*

## Verifying this change

covered by existing tests.

## Documentation

### DIFF
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/consumer/SingleInputGate.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/consumer/SingleInputGate.java
@@ -273,11 +273,6 @@ public class SingleInputGate extends InputGate {
 		return 0;
 	}
 
-	@Override
-	public String getOwningTaskName() {
-		return owningTaskName;
-	}
-
 	public CompletableFuture<Void> getCloseFuture() {
 		return closeFuture;
 	}
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/Task.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/Task.java
@@ -593,7 +593,7 @@ public class Task implements Runnable, TaskActions, PartitionProducerStateProvid
 
 			LOG.info("Registering task at network: {}.", this);
 
-			setupPartionsAndGates(consumableNotifyingPartitionWriters, inputGates);
+			setupPartitionsAndGates(consumableNotifyingPartitionWriters, inputGates);
 
 			for (ResultPartitionWriter partitionWriter : consumableNotifyingPartitionWriters) {
 				taskEventDispatcher.registerPartition(partitionWriter.getPartitionId());
@@ -823,7 +823,7 @@ public class Task implements Runnable, TaskActions, PartitionProducerStateProvid
 	}
 
 	@VisibleForTesting
-	public static void setupPartionsAndGates(
+	public static void setupPartitionsAndGates(
 		ResultPartitionWriter[] producedPartitions, InputGate[] inputGates) throws IOException {
 
 		for (ResultPartitionWriter partition : producedPartitions) {
--- a/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/BarrierBuffer.java
+++ b/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/BarrierBuffer.java
@@ -18,6 +18,7 @@
 package org.apache.flink.streaming.runtime.io;
 
 import org.apache.flink.annotation.Internal;
+import org.apache.flink.annotation.VisibleForTesting;
 import org.apache.flink.runtime.checkpoint.CheckpointMetaData;
 import org.apache.flink.runtime.checkpoint.CheckpointMetrics;
 import org.apache.flink.runtime.checkpoint.decline.AlignmentLimitExceededException;
@@ -79,6 +80,8 @@ public class BarrierBuffer implements CheckpointBarrierHandler {
 	 */
 	private final long maxBufferedBytes;
 
+	private final String taskName;
+
 	/**
 	 * The sequence of buffers/events that has been unblocked and must now be consumed before
 	 * requesting further data from the input gate.
@@ -119,11 +122,10 @@ public class BarrierBuffer implements CheckpointBarrierHandler {
 	 *
 	 * @param inputGate The input gate to draw the buffers and events from.
 	 * @param bufferBlocker The buffer blocker to hold the buffers and events for channels with barrier.
-	 *
-	 * @throws IOException Thrown, when the spilling to temp files cannot be initialized.
 	 */
-	public BarrierBuffer(InputGate inputGate, BufferBlocker bufferBlocker) throws IOException {
-		this (inputGate, bufferBlocker, -1);
+	@VisibleForTesting
+	BarrierBuffer(InputGate inputGate, BufferBlocker bufferBlocker) {
+		this (inputGate, bufferBlocker, -1, "Testing: No task associated");
 	}
 
 	/**
@@ -136,11 +138,9 @@ public class BarrierBuffer implements CheckpointBarrierHandler {
 	 * @param inputGate The input gate to draw the buffers and events from.
 	 * @param bufferBlocker The buffer blocker to hold the buffers and events for channels with barrier.
 	 * @param maxBufferedBytes The maximum bytes to be buffered before the checkpoint aborts.
-	 *
-	 * @throws IOException Thrown, when the spilling to temp files cannot be initialized.
+	 * @param taskName The task name for logging.
 	 */
-	public BarrierBuffer(InputGate inputGate, BufferBlocker bufferBlocker, long maxBufferedBytes)
-			throws IOException {
+	BarrierBuffer(InputGate inputGate, BufferBlocker bufferBlocker, long maxBufferedBytes, String taskName) {
 		checkArgument(maxBufferedBytes == -1 || maxBufferedBytes > 0);
 
 		this.inputGate = inputGate;
@@ -150,6 +150,8 @@ public class BarrierBuffer implements CheckpointBarrierHandler {
 
 		this.bufferBlocker = checkNotNull(bufferBlocker);
 		this.queuedBuffered = new ArrayDeque<BufferOrEventSequence>();
+
+		this.taskName = taskName;
 	}
 
 	// ------------------------------------------------------------------------
@@ -213,7 +215,7 @@ public class BarrierBuffer implements CheckpointBarrierHandler {
 	}
 
 	private void completeBufferedSequence() throws IOException {
-		LOG.debug("{}: Finished feeding back buffered data.", inputGate.getOwningTaskName());
+		LOG.debug("{}: Finished feeding back buffered data.", taskName);
 
 		currentBuffered.cleanup();
 		currentBuffered = queuedBuffered.pollFirst();
@@ -249,7 +251,7 @@ public class BarrierBuffer implements CheckpointBarrierHandler {
 				// we did not complete the current checkpoint, another started before
 				LOG.warn("{}: Received checkpoint barrier for checkpoint {} before completing current checkpoint {}. " +
 						"Skipping current checkpoint.",
-					inputGate.getOwningTaskName(),
+					taskName,
 					barrierId,
 					currentCheckpointId);
 
@@ -283,7 +285,7 @@ public class BarrierBuffer implements CheckpointBarrierHandler {
 			// actually trigger checkpoint
 			if (LOG.isDebugEnabled()) {
 				LOG.debug("{}: Received all barriers, triggering checkpoint {} at {}.",
-					inputGate.getOwningTaskName(),
+					taskName,
 					receivedBarrier.getId(),
 					receivedBarrier.getTimestamp());
 			}
@@ -314,9 +316,7 @@ public class BarrierBuffer implements CheckpointBarrierHandler {
 			if (barrierId == currentCheckpointId) {
 				// cancel this alignment
 				if (LOG.isDebugEnabled()) {
-					LOG.debug("{}: Checkpoint {} canceled, aborting alignment.",
-						inputGate.getOwningTaskName(),
-						barrierId);
+					LOG.debug("{}: Checkpoint {} canceled, aborting alignment.", taskName, barrierId);
 				}
 
 				releaseBlocksAndResetBarriers();
@@ -326,7 +326,7 @@ public class BarrierBuffer implements CheckpointBarrierHandler {
 				// we canceled the next which also cancels the current
 				LOG.warn("{}: Received cancellation barrier for checkpoint {} before completing current checkpoint {}. " +
 						"Skipping current checkpoint.",
-					inputGate.getOwningTaskName(),
+					taskName,
 					barrierId,
 					currentCheckpointId);
 
@@ -357,9 +357,7 @@ public class BarrierBuffer implements CheckpointBarrierHandler {
 			latestAlignmentDurationNanos = 0L;
 
 			if (LOG.isDebugEnabled()) {
-				LOG.debug("{}: Checkpoint {} canceled, skipping alignment.",
-					inputGate.getOwningTaskName(),
-					barrierId);
+				LOG.debug("{}: Checkpoint {} canceled, skipping alignment.", taskName, barrierId);
 			}
 
 			notifyAbortOnCancellationBarrier(barrierId);
@@ -414,7 +412,7 @@ public class BarrierBuffer implements CheckpointBarrierHandler {
 		if (maxBufferedBytes > 0 && (numQueuedBytes + bufferBlocker.getBytesBlocked()) > maxBufferedBytes) {
 			// exceeded our limit - abort this checkpoint
 			LOG.info("{}: Checkpoint {} aborted because alignment volume limit ({} bytes) exceeded.",
-				inputGate.getOwningTaskName(),
+				taskName,
 				currentCheckpointId,
 				maxBufferedBytes);
 
@@ -458,9 +456,7 @@ public class BarrierBuffer implements CheckpointBarrierHandler {
 		startOfAlignmentTimestamp = System.nanoTime();
 
 		if (LOG.isDebugEnabled()) {
-			LOG.debug("{}: Starting stream alignment for checkpoint {}.",
-				inputGate.getOwningTaskName(),
-				checkpointId);
+			LOG.debug("{}: Starting stream alignment for checkpoint {}.", taskName, checkpointId);
 		}
 	}
 
@@ -486,9 +482,7 @@ public class BarrierBuffer implements CheckpointBarrierHandler {
 			numBarriersReceived++;
 
 			if (LOG.isDebugEnabled()) {
-				LOG.debug("{}: Received barrier from channel {}.",
-					inputGate.getOwningTaskName(),
-					channelIndex);
+				LOG.debug("{}: Received barrier from channel {}.", taskName, channelIndex);
 			}
 		}
 		else {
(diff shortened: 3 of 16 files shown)


zhijiangW:

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamInputProcessor.java
```diff
@@ -121,12 +121,13 @@ public StreamInputProcessor(
...
 			TaskIOMetricGroup metrics,
-			WatermarkGauge watermarkGauge) throws IOException {
+			WatermarkGauge watermarkGauge,
+			String taskName) throws IOException {
 
 		InputGate inputGate = InputGateUtil.createInputGate(inputGates);
 
 		this.barrierHandler = InputProcessorUtil.createCheckpointBarrierHandler(
-			checkpointedTask, checkpointMode, ioManager, inputGate, taskManagerConfig);
+			checkpointedTask, checkpointMode, ioManager, inputGate, taskManagerConfig, taskName);
```
azagrebin: codestyle: let's break all function argument into separate lines if the line is broken.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamTwoInputProcessor.java
```diff
@@ -140,12 +140,13 @@ public StreamTwoInputProcessor(
...
 			WatermarkGauge input1WatermarkGauge,
-			WatermarkGauge input2WatermarkGauge) throws IOException {
+			WatermarkGauge input2WatermarkGauge,
+			String taskName) throws IOException {
 
 		final InputGate inputGate = InputGateUtil.createInputGate(inputGates1, inputGates2);
 
 		this.barrierHandler = InputProcessorUtil.createCheckpointBarrierHandler(
-			checkpointedTask, checkpointMode, ioManager, inputGate, taskManagerConfig);
+			checkpointedTask, checkpointMode, ioManager, inputGate, taskManagerConfig, taskName);
```
azagrebin: codestyle: let's break all function argument into separate lines if the line is broken.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/InputProcessorUtil.java
```diff
@@ -53,9 +54,11 @@ public static CheckpointBarrierHandler createCheckpointBarrierHandler(
 			}
 
 			if (taskManagerConfig.getBoolean(NetworkEnvironmentOptions.NETWORK_CREDIT_MODEL)) {
-				barrierHandler = new BarrierBuffer(inputGate, new CachedBufferBlocker(inputGate.getPageSize()), maxAlign);
+				barrierHandler = new BarrierBuffer(
+					inputGate, new CachedBufferBlocker(inputGate.getPageSize()), maxAlign, taskName);
```
azagrebin: codestyle: let's break all function argument into separate lines if the line is broken.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/InputProcessorUtil.java
```diff
@@ -53,9 +54,11 @@ public static CheckpointBarrierHandler createCheckpointBarrierHandler(
...
 			} else {
-				barrierHandler = new BarrierBuffer(inputGate, new BufferSpiller(ioManager, inputGate.getPageSize()), maxAlign);
+				barrierHandler = new BarrierBuffer(
+					inputGate, new BufferSpiller(ioManager, inputGate.getPageSize()), maxAlign, taskName);
```
azagrebin: codestyle: let's break all function argument into separate lines if the line is broken.

flink-core/src/main/java/org/apache/flink/api/common/TaskInfo.java
```diff
@@ -67,11 +72,12 @@ public TaskInfo(
 		checkArgument(indexOfSubtask < numberOfParallelSubtasks, "Task index must be less than parallelism.");
 		checkArgument(attemptNumber >= 0, "Attempt number must be a non-negative number.");
 		this.taskName = checkNotNull(taskName, "Task Name must not be null.");
+		this.taskNameWithSubtasks = checkNotNull(taskNameWithSubtasks);
+		this.taskNameWithSubtaskAndId = checkNotNull(taskNameWithSubtaskAndId);
 		this.maxNumberOfParallelSubtasks = maxNumberOfParallelSubtasks;
 		this.indexOfSubtask = indexOfSubtask;
 		this.numberOfParallelSubtasks = numberOfParallelSubtasks;
 		this.attemptNumber = attemptNumber;
-		this.taskNameWithSubtasks = taskName + " (" + (indexOfSubtask + 1) + '/' + numberOfParallelSubtasks + ')';
```
azagrebin: Unless we introduce at least a static factory method `TaskInfo.create(...)`,  I would keep these internal task name creation inside `TaskInfo` otherwise Task constructor contains too many details. The same for `taskNameWithSubtaskAndId`, I would rather pass the `executionAttemptID` to `TaskInfo` constructor or again add a factory method with these details.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/consumer/SingleInputGate.java
```diff
@@ -254,16 +254,6 @@ public BufferPool getBufferPool() {
 		return bufferPool;
 	}
 
-	@Override
-	public int getPageSize() {
-		if (bufferPool != null) {
-			return bufferPool.getMemorySegmentSize();
```
azagrebin: Looks like `BufferProvider.getMemorySegmentSize` can be also removed.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/InputProcessorUtil.java
```diff
@@ -53,12 +54,13 @@ public static CheckpointBarrierHandler createCheckpointBarrierHandler(
 					+ " must be positive or -1 (infinite)");
 			}
 
+			int pageSize = NetworkEnvironmentConfiguration.getPageSize(taskManagerConfig);
```
azagrebin: Are we going to have page size and `NETWORK_CREDIT_MODEL` for all shuffle services? Or is there any other better way how to abstract BarrierBuffer creation, like asking NetworkEnvironment/ShuffleService about needed parameters and pass them into Environment/StreamTasks?

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/consumer/SingleInputGate.java
zhijiangW: Yes, I would submit a hotfix commit for removing it in this PR.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/InputProcessorUtil.java
zhijiangW: Thanks for this good catch. 

The options of page size and `NETWORK_CREDIT_MODEL` might be not general for all shuffle services. `BarrierHandler` should not concern of the specific shuffle service implementation and see these parameters in theory.

Actually the `NETWORK_CREDIT_MODEL` option is deprecated. If we could remove this option completed and the page size is also not needed because `CachedBufferBlocker` could get the page size from `BufferOrEvent#getSize` after introducing the `InputGateWithMetrics` in another PR. 

There was a blocker for removing these deprecated codes before, but I think this blocker is already confirmed now. Then I would talk with Piotr about when to remove them (would be better if done in release 1.9). I think this is a right way to go to avoid extra efforts of abstracting `BarrierBuffer` creation. WDYT?

flink-core/src/main/java/org/apache/flink/api/common/TaskInfo.java
zhijiangW: Yes, the factory way would be better.

zhijiangW:
Thanks for the reviews and good suggestions @azagrebin !

Yes, we still rely on some network options in `BarrierBuffer` implementation. And I also think it is not good to do so. This PR is mainly making the interface seem more general and clean. We could further refactor the `BarrierBuffer` implementation future as I suggested inline comments.

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/Task.java
```diff
@@ -310,19 +310,20 @@ public Task(
...
-				taskInformation.getNumberOfSubtasks(),
-				attemptNumber,
-				String.valueOf(slotAllocationId));
+		this.taskInfo = TaskInfo.create(
+			taskInformation.getTaskName(),
+			taskInformation.getMaxNumberOfSubtaks(),
+			taskInformation.getNumberOfSubtasks(),
+			subtaskIndex,
+			attemptNumber,
+			String.valueOf(executionAttemptID),
```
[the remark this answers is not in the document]
zhijiangW: sure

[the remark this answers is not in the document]
zhijiangW: The `TaskInfo` class is in flink-core module which could not rely on flink-runtime module. So it could not be done as that.

zhijiangW:
I already reverted the changes for `getPage` issue which could be solved in a separate PR future in a better way.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/BarrierBuffer.java
```diff
@@ -139,8 +141,7 @@ public BarrierBuffer(InputGate inputGate, BufferBlocker bufferBlocker) throws IO
 	 *
 	 * @throws IOException Thrown, when the spilling to temp files cannot be initialized.
 	 */
```
tillrohrmann: JavaDocs have not been updated.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/BarrierBuffer.java
```diff
@@ -122,8 +124,8 @@
...
-	public BarrierBuffer(InputGate inputGate, BufferBlocker bufferBlocker) throws IOException {
-		this (inputGate, bufferBlocker, -1);
+	public BarrierBuffer(InputGate inputGate, BufferBlocker bufferBlocker) {
```
tillrohrmann: It looks as if this constructor is only used for testing. Please annotate with `@VisibleForTesting`

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/BarrierBuffer.java
```diff
@@ -122,8 +124,8 @@
...
+		this (inputGate, bufferBlocker, -1, "");
```
tillrohrmann: Give a better `taskName` like `"Testing: No task associated"`

flink-core/src/main/java/org/apache/flink/api/common/TaskInfo.java
```diff
@@ -132,11 +122,67 @@ public String getTaskNameWithSubtasks() {
...
+	//  Factory Utilities
+	// ------------------------------------------------------------------------
+
+	public static TaskInfo create(
+		String taskName,
+		int maxNumberOfParallelSubtasks,
+		int numberOfParallelSubtasks,
+		int subtaskIndex,
+		int attemptNumber,
+		String attemptIDAsString,
```
tillrohrmann: why passing a `String` and not a `ExecutionAttemptID`?

flink-core/src/main/java/org/apache/flink/api/common/TaskInfo.java
```diff
@@ -67,11 +56,12 @@ public TaskInfo(
 		checkArgument(indexOfSubtask < numberOfParallelSubtasks, "Task index must be less than parallelism.");
 		checkArgument(attemptNumber >= 0, "Attempt number must be a non-negative number.");
 		this.taskName = checkNotNull(taskName, "Task Name must not be null.");
+		this.taskNameWithSubtasks = checkNotNull(taskNameWithSubtasks);
+		this.taskNameWithSubtaskAndID = checkNotNull(taskNameWithSubtaskAndID);
```
tillrohrmann: This seems a bit odd. We are storing three times the task name. I think this is duplicated information. I think it would be better to rather store the `taskName`, its index, the number of maximum tasks and its `ExecutionAttemptID`. If `TaskInfo` contains this, then we could generate the `taskNameWithSubtaskAndID` based on this information when instantiating the `One/TwoInputStreamTask`.

flink-core/src/main/java/org/apache/flink/api/common/TaskInfo.java
```diff
@@ -132,11 +122,67 @@ public String getTaskNameWithSubtasks() {
...
+		int maxNumberOfParallelSubtasks,
+		int subtaskIndex,
+		int numberOfParallelSubtasks,
+		int attemptNumber) {
+
+		String taskNameWithSubtask = taskName + " (" + (subtaskIndex + 1) + '/' + numberOfParallelSubtasks + ')';
+		return new TaskInfo(
+			taskName,
+			taskNameWithSubtask,
+			taskNameWithSubtask,
```
tillrohrmann: this does not really fit. Why not creating a `TestingUtils#createTaskInfo` method which instantiates the `TaskInfo` with a random `ExecutionAttemptID`?

flink-core/src/main/java/org/apache/flink/api/common/TaskInfo.java
tillrohrmann: Andrey gave me the answer that `ExecutionAttemptID` is not part of `flink-core`. But then I'm wondering, whether `TaskInfo` is the right container to transport the `ExecutionAttemptID` to the `BarrierBuffer`. It feels a bit that we are abusing a class which was not intended for this purpose here.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/OneInputStreamTask.java
```diff
@@ -88,7 +88,8 @@ public void init() throws Exception {
 					getStreamStatusMaintainer(),
 					this.headOperator,
 					getEnvironment().getMetricGroup().getIOMetricGroup(),
-					inputWatermarkGauge);
+					inputWatermarkGauge,
+					getEnvironment().getTaskInfo().getTaskNameWithSubtaskAndID());
```
tillrohrmann: Maybe we could use `getEnvironment().getExecutionId()` and `getEnvironment().getTaskInfo().getTaskName()` to construct the `getTaskNameWithSubtaskAndID()`. Then we don't have to touch `TaskInfo` at all.

flink-core/src/main/java/org/apache/flink/api/common/TaskInfo.java
zhijiangW: Yes, it is because of module dependency.
I do not think we are abusing a class, because the previous usage of `allocationIDAsString` also has the same issue.
In the first version I generates the full task name with attempt id in task class and then pass this string full task name into constructor of `TaskInfo` which is not aware of how this name is generated. But Andrey suggests passing the original components as now to make task class simple logic.

flink-core/src/main/java/org/apache/flink/api/common/TaskInfo.java
zhijiangW: Yes, that is another option, but that would cause generate this task name multiple times for `BarrierBuffer` and `StreamProcessor`. The benefit of generating task name in `TaskInfo` is that it is  done only once and be reused for all the places. If we further extend to reuse this name in other components we also do not need to generate it again.

flink-core/src/main/java/org/apache/flink/api/common/TaskInfo.java
zhijiangW: Agree, then we could also avoid test related codes here.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/OneInputStreamTask.java
zhijiangW: Yes, that could also work. The only concern is we need do this multiple times in different components which need task name. Actually I think the task name with different info should belong to the scope of `TaskInfo` if we are caring about the previous implementation limit. WDYT?

flink-core/src/main/java/org/apache/flink/api/common/TaskInfo.java
tillrohrmann: I think the `allocationIDAsString` is also not right and suffers from the same problems.

flink-core/src/main/java/org/apache/flink/api/common/TaskInfo.java
zhijiangW: Ok, then I would not modify the structure of `TaskInfo` and only instantiate the task name in separate components via `RuntimeEnvironment#getExecutionID`.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/OneInputStreamTask.java
tillrohrmann: I think `TaskInfo` has the purpose to encapsulate some user facing information about the task. That's why it is located in `flink-core`. The `ExecutionAttempID` and also the `AllocationID` are runtime specific concepts which should not be part of this class imo. Thus, I would strongly suggest to not mix concerns and cleanly separate the different aspects. Which components actually need `getTaskNameWithSubtaskAndID()`? As far as I can tell, it is just the `UnionInputGate` as well as the `BarrierBuffer`.

flink-core/src/main/java/org/apache/flink/api/common/TaskInfo.java
zhijiangW: I am not sure the history consideration of introducing `allocationIDAsString` here. If we think it is not right thing to do, then I could avoid following this way. 

In addition, do you think I should refactor the way of `allocationIDAsString` in a hotfix commit?

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/OneInputStreamTask.java
zhijiangW: Agree this suggestion, and I could adjust it.

zhijiangW:
@tillrohrmann thanks for the confirmation and I have addressed the comments by squashing the commits.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/OneInputStreamTask.java
```diff
@@ -88,7 +88,8 @@ public void init() throws Exception {
 					getStreamStatusMaintainer(),
 					this.headOperator,
 					getEnvironment().getMetricGroup().getIOMetricGroup(),
-					inputWatermarkGauge);
+					inputWatermarkGauge,
+					getEnvironment().getTaskInfo().getTaskNameWithSubtasks() + " (" + getEnvironment().getExecutionId() + ')');
```
tillrohrmann: Should be factored out to avoid code duplication between this class and `TwoInputStreamTask`.

zhijiangW:
@tillrohrmann I have addressed the last comment and the travis issue is caused by another PR merged which is also solved now.

PR #8530, opened 2019-05-24
[FLINK-12611][table-planner-blink] Make time indicator nullable in blink

JingsongLi:

Make time indicator nullable in blink to support agg on TimeIndicator

## Verifying this change

ut

  - The serializers: no

## Documentation

### DIFF
--- a/flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/functions/sql/ProctimeSqlFunction.java
+++ b/flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/functions/sql/ProctimeSqlFunction.java
@@ -20,9 +20,6 @@ package org.apache.flink.table.functions.sql;
 
 import org.apache.flink.table.calcite.FlinkTypeFactory;
 
-import org.apache.calcite.rel.type.RelDataType;
-import org.apache.calcite.rel.type.RelDataTypeFactory;
-import org.apache.calcite.rel.type.RelProtoDataType;
 import org.apache.calcite.sql.SqlFunction;
 import org.apache.calcite.sql.SqlFunctionCategory;
 import org.apache.calcite.sql.SqlKind;
@@ -37,16 +34,15 @@ public class ProctimeSqlFunction extends SqlFunction {
 		super(
 				"PROCTIME",
 				SqlKind.OTHER_FUNCTION,
-				ReturnTypes.explicit(new ProctimeRelProtoDataType()),
+				ReturnTypes.explicit(factory ->
+						((FlinkTypeFactory) factory).createProctimeIndicatorType(false)),
 				null,
 				OperandTypes.NILADIC,
 				SqlFunctionCategory.TIMEDATE);
 	}
 
-	private static class ProctimeRelProtoDataType implements RelProtoDataType {
-		@Override
-		public RelDataType apply(RelDataTypeFactory factory) {
-			return ((FlinkTypeFactory) factory).createProctimeIndicatorType();
-		}
+	@Override
+	public boolean isDeterministic() {
+		return false;
 	}
 }
--- a/flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/calcite/FlinkTypeFactory.scala
+++ b/flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/calcite/FlinkTypeFactory.scala
@@ -128,8 +128,8 @@ class FlinkTypeFactory(typeSystem: RelDataTypeSystem) extends JavaTypeFactoryImp
       case LogicalTypeRoot.TIMESTAMP_WITHOUT_TIME_ZONE =>
         val timestampType = t.asInstanceOf[TimestampType]
         timestampType.getKind match {
-          case TimestampKind.PROCTIME => createProctimeIndicatorType()
-          case TimestampKind.ROWTIME => createRowtimeIndicatorType()
+          case TimestampKind.PROCTIME => createProctimeIndicatorType(true)
+          case TimestampKind.ROWTIME => createRowtimeIndicatorType(true)
           case TimestampKind.REGULAR => createSqlType(TIMESTAMP)
         }
       case _ =>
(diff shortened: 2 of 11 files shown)


JingsongLi:

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/functions/sql/ProctimeSqlFunction.java
```diff
@@ -37,16 +34,15 @@ public ProctimeSqlFunction() {
 		super(
 				"PROCTIME",
 				SqlKind.OTHER_FUNCTION,
-				ReturnTypes.explicit(new ProctimeRelProtoDataType()),
+				ReturnTypes.explicit(factory ->
```
KurtYoung: Why don't we keep align with legacy planner's ProctimeSqlFunction?

JingsongLi: Previously set to false causes sorts of inexplicable errors in `LookupJoinTest`.
The fundamental reason is that:
`TimeIndicatorRelDataType` assigns `isNullable` and overrides `equals`, but `BasicSqlType` do `computeDigest` at the constructor, so this leads to the inconsistency between `equals` and `Digest`, which leads to a variety of problems.
Fixed it in ff7c1e7
(Looks like `GenericRelDataType` has same problem.)

KurtYoung: Flink planner just use `SqlTypeName.TIMESTAMP` for return type

JingsongLi: The `ProctimeSqlFunction` of Flink planner is `ProctimeMaterializeSqlFunction` in blink.
`ProctimeSqlFunction` in blink is the "proctime()" method to access a proctime attribute.(blink introduce a new buildIn function named "proctime()")

flink-table/flink-table-planner-blink/src/test/scala/org/apache/flink/table/plan/batch/sql/join/LookupJoinTest.scala
```diff
@@ -68,8 +68,8 @@ class LookupJoinTest extends TableTestBase {
     expectExceptionThrown(
       "SELECT * FROM MyTable AS T RIGHT JOIN temporalTest " +
         "FOR SYSTEM_TIME AS OF T.proctime AS D ON T.a = D.id",
-      "Unsupported join type for semi-join RIGHT",
-      classOf[IllegalArgumentException]
+      null,
+      classOf[AssertionError]
```
wuchong: Why we need to change this?

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/calcite/FlinkTypeFactory.scala
```diff
@@ -128,8 +128,8 @@ class FlinkTypeFactory(typeSystem: RelDataTypeSystem) extends JavaTypeFactoryImp
       case LogicalTypeRoot.TIMESTAMP_WITHOUT_TIME_ZONE =>
         val timestampType = t.asInstanceOf[TimestampType]
         timestampType.getKind match {
-          case TimestampKind.PROCTIME => createProctimeIndicatorType()
-          case TimestampKind.ROWTIME => createRowtimeIndicatorType()
+          case TimestampKind.PROCTIME => createProctimeIndicatorType(true)
+          case TimestampKind.ROWTIME => createRowtimeIndicatorType(true)
```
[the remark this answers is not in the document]
JingsongLi: `createTypeWithNullability` will be called below, So there let be default true.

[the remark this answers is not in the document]
wuchong: Get

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/calcite/FlinkTypeFactory.scala
```diff
@@ -148,27 +148,27 @@ class FlinkTypeFactory(typeSystem: RelDataTypeSystem) extends JavaTypeFactoryImp
...
-        getTypeSystem,
-        originalType.asInstanceOf[BasicSqlType],
-        isEventTime = false)
-    )
+    val indicator = new TimeIndicatorRelDataType(
+      getTypeSystem,
+      originalType.asInstanceOf[BasicSqlType],
+      isNullable,
+      isEventTime = false)
+    createTypeWithNullability(indicator, isNullable)
```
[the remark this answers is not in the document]
JingsongLi: OK, let it be:
```
val originalType = createFieldTypeFromLogicalType(new TimestampType(isNullable, 3))
    canonize(new TimeIndicatorRelDataType(
      getTypeSystem,
      originalType.asInstanceOf[BasicSqlType],
      isNullable,
      isEventTime = false))
```

flink-table/flink-table-planner-blink/src/test/scala/org/apache/flink/table/plan/batch/sql/join/LookupJoinTest.scala
JingsongLi: Calcite does not allow FOR SYSTEM_TIME AS OF non-nullable left table field to Right Join.
There should be a exception:
```
java.lang.AssertionError
	at org.apache.calcite.sql2rel.SqlToRelConverter.getCorrelationUse(SqlToRelConverter.java:2517)
	at org.apache.calcite.sql2rel.SqlToRelConverter.createJoin(SqlToRelConverter.java:2426)
	at org.apache.calcite.sql2rel.SqlToRelConverter.convertFrom(SqlToRelConverter.java:2071)
	at org.apache.calcite.sql2rel.SqlToRelConverter.convertSelectImpl(SqlToRelConverter.java:646)
	at org.apache.calcite.sql2rel.SqlToRelConverter.convertSelect(SqlToRelConverter.java:627)
	at org.apache.calcite.sql2rel.SqlToRelConverter.convertQueryRecursive(SqlToRelConverter.java:3100)
	at org.apache.calcite.sql2rel.SqlToRelConverter.convertQuery(SqlToRelConverter.java:563)
	at org.apache.flink.table.calcite.FlinkPlannerImpl.rel(FlinkPlannerImpl.scala:125)
```

wuchong: Do you mean Calcite doesn't allow **nullable** left field to right join? 

However, the T.proctime field is derived from proctime() builtin method, which should be **not null**.

JingsongLi: Not nullable, Non-nullable
If this field is non-nullable and this is a right join, after join, this field will be nullable.
Calcite does not deal with this correlation, in `getCorrelationUse`, it will check the input field and output field, their nullable are not the same, so there is a AssertionError.
IMO, Calcite lack of consideration in this situation, but our lookup join don't support it too.

wuchong:
This looks good to me.
+1 to merge. 

I will rebase this and merge it.

PR #8532, opened 2019-05-24
[FLINK-12541][REST] Support to submit Python Table API jobs via REST API

dianfu:

*This pull request supports to submit Python Table API jobs via REST API.*
To support submit resource files other than jar, we need to update the REST API a bit as specified in [FLIP38](https://cwiki.apache.org/confluence/display/FLINK/FLIP-38%3A+Python+Table+API)

Current REST API | New REST API | Remarks
-- | -- | --
GET /jars | GET /artifacts | Now it lists both jar files and python package files.
POST /jars/upload | POST /artifacts/upload | Allow users to upload both jar files and python package files. Use "Content-Type" in the request header to mark the file type.
DELETE /jars/:jarid | DELETE /artifacts/:artifactid | Delete files according to “artifactid”.
GET /jars/:jarid/plan | GET /artifacts/:artifactid/plan | We will introduce a new optional parameter:pythonEntryModule.It allows specifying a python entry point module other than __main__.py.
POST /jars/:jarid/run | POST /artifacts/:artifactid/run | We also introduce the optional pythonEntryModule parameter here, as same as above.

## Brief change log

  - *[2c4d575](https://github.com/dianfu/flink/commit/2c4d5752c8e2783b5d309c8730677da555be6e7e): Changes the REST API prefix from /jars to /artifacts as we will not only support upload jar files, but also Python resource files. This commit also refactor the class name prefix from JarXXX to ArtifactXXX*
  - *[c286149](https://github.com/dianfu/flink/commit/c286149bf3bd7133bb395cd38814b29652339a46): Updates the web to use the new /artifacts REST API instead of the /jars REST API*
  - *[b516468](https://github.com/dianfu/flink/commit/b516468cb7b7957358101f57a07f1c704034a621) Supports to submit Python Table API jobs. It supports to submit Python Table API jobs via resource files such as .py, .zip and .egg.*

## Verifying this change

  - *Added tests ArtifactHandlerUtilsTest.testExtractContainedLibraries, ArtifactXXXTest*

  - The serializers: (no)

## Documentation


sunjincheng121:
Thanks for your PR. @dianfu!
I have merged the https://issues.apache.org/jira/browse/FLINK-12327, so please rebase the code :) then I'll review this PR ASAP.
Best,
Jincheng

dianfu:
@sunjincheng121 Great, I have rebased the PR.

sunjincheng121:
@flinkbot attention @tillrohrmann @zentol

tillrohrmann:
Since you are changing part of Flink's public API, how did you check that we did not break backwards compatibility? It should still be possible to submit jobs with a `1.8` client against a `1.9` cluster.

sunjincheng121:
That's a good point,  @tillrohrmann! We use the `org.apache.flink.runtime.rest.compatibility.RestAPIStabilityTest` to ensure compatibility. and in current PR should keep the old APIs there but mark Deprecated for the old APIs.

dianfu:
@tillrohrmann Thanks a lot for your reply.  As mentioned by @sunjincheng121 that the test case RestAPIStabilityTest can make sure the backwards compatibility. Besides, I have also manually checked that the old REST API still works by the following way:
1. Only apply the first [commit](https://github.com/dianfu/flink/commit/867d82e5c8a625acaed9111f64661834ed3edb1b)  which adds REST API /artifacts/ and deprecates the old REST API /jars/
2. Build the project, start up a local cluster and check that the functionalities such as uploading jars, getting plans, submitting jobs works well.
3. As only the first commit is applied, the frontend still uses the old REST API. This makes sure that the old REST API still works.

dianfu:
@zentol I think your suggestions make sense to me. Will updated the PR accordingly. Thanks a lot.

zentol:
@dianfu In your manual test, did you not include commit #3? That one modifies the API as well, and from what I can tell at a glance should also impact the legacy /jar endpoints.
Now that we separate things though this should no longer matter.

dianfu:
@zentol The manual test does not include the third commit. Definitely agree that separating the new API with the old API will be more safer for backwards compatibility. :)

dianfu:
@zentol @tillrohrmann I have updated the PR and separated the newly added API with the old ones per @zentol 's suggestion. Could you help to take a look at? Appreciated!

tillrohrmann:
`RestAPIStabilityTest.testDispatcherRestAPIStability` is failing the build @dianfu.

dianfu:
@tillrohrmann Thanks a lot for the remind and have addressed the test failure.

flink-runtime-web/web-dashboard/src/app/pages/submit/submit.component.html
```diff
@@ -33,23 +33,23 @@
         </tr>
       </thead>
       <tbody>
-        <ng-container *ngFor="let jar of listOfJar; trackBy:trackJarBy;">
-          <tr (click)="expandJar(jar)" class="clickable">
-            <td>{{jar.name}}</td>
-            <td>{{jar.uploaded | date:'yyyy-MM-dd, HH:mm:ss'}}</td>
+        <ng-container *ngFor="let artifact of listOfJar; trackBy:trackJarBy;">
```
sunjincheng121: `listOfJar` -> `listOfArtifacts`?

flink-runtime-web/web-dashboard/src/app/pages/submit/submit.component.ts
```diff
@@ -154,7 +154,7 @@ export class SubmitComponent implements OnInit, OnDestroy {
   }
 
   constructor(
-    private jarService: JarService,
+    private artifactService: ArtifactService,
```
sunjincheng121: Line 152:  `trackJarBy` -> `trackArtifactBy`?

flink-runtime-web/web-dashboard/src/app/pages/submit/submit.component.html
sunjincheng121: `trackJarBy` -> `trackArtifactBy`

flink-runtime-web/web-dashboard/src/app/pages/submit/submit.component.ts
```diff
@@ -69,45 +69,45 @@ export class SubmitComponent implements OnInit, OnDestroy {
...
 
   /**
-   * Delete jar
-   * @param jar
+   * Delete artifact
+   * @param artifact
    */
-  deleteJar(jar: JarFilesItemInterface) {
-    this.jarService.deleteJar(jar.id).subscribe(() => {
+  deleteArtifact(artifact: JarFilesItemInterface) {
```
sunjincheng121: `JarFilesItemInterface` -> `ArtifactFilesItemInterface`?

flink-runtime-web/web-dashboard/src/app/pages/submit/submit.component.ts
```diff
@@ -23,7 +23,7 @@ import { Router } from '@angular/router';
 import { JarFilesItemInterface } from 'interfaces';
```
sunjincheng121: should we rename the `JarFilesItemInterface` to `ArtifactFilesItemInterface` in `jar.js`?

flink-runtime-web/web-dashboard/src/app/services/artifact.service.ts
```diff
@@ -26,12 +26,12 @@ import { catchError, map } from 'rxjs/operators';
...
-export class JarService {
+export class ArtifactService {
   /**
-   * Get uploaded jar list
+   * Get uploaded artifact list
    */
-  loadJarList() {
-    return this.httpClient.get<JarListInterface>(`${BASE_URL}/jars`).pipe(
+  loadArtifactList() {
+    return this.httpClient.get<JarListInterface>(`${BASE_URL}/artifacts`).pipe(
```
sunjincheng121: Should we rename the `JarListInterface` to `ArtifactListInterface` in `jar.js`?

flink-runtime-web/src/main/java/org/apache/flink/runtime/webmonitor/handlers/AbstractArtifactPlanHeaders.java
```diff
@@ -0,0 +1,61 @@
...
+	}
+
+	@Override
+	public ArtifactPlanMessageParameters getUnresolvedMessageParameters() {
+		return new ArtifactPlanMessageParameters();
+	}
+
+	@Override
+	public String getTargetRestEndpointURL() {
+		return "/artifacts/:" + ArtifactIdPathParameter.KEY + "/plan";
```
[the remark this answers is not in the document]
dianfu: That makes sense to me. +1

flink-runtime-web/src/main/java/org/apache/flink/runtime/webmonitor/handlers/ArtifactRequestBody.java
```diff
@@ -0,0 +1,131 @@
...
+
+	@Nullable
+	@JsonIgnore
+	public Integer getParallelism() {
+		return parallelism;
+	}
+
+	@Nullable
+	@JsonIgnore
+	public String getDependentArtifactId() {
```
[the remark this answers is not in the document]
dianfu: This parameter allows to specify a separate jar file as the dependency. What do you think?

flink-runtime-web/web-dashboard/src/app/pages/submit/submit.component.html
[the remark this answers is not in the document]
dianfu: This PR needs to be updated to make it work as the package name of PythonDriver changes in latest commit.

flink-runtime-web/web-dashboard/src/app/services/artifact.service.ts
dianfu: I think so. +1

dianfu:
@sunjincheng121 Thanks a lot for the review.

Regarding to question 1, it's to ensure better backwards compatibility per @zentol 's suggestion.
Regarding to question 2, adding separate test cases for ArtifactXXX seems reasonable for me. We can remove the JarXXX and the corresponding test cases in the next releases together.

Will update the PR later on.

dianfu:
@sunjincheng121 Have updated the PR according to your comments. Looking forwarding to your feedback.

flink-runtime-web/src/main/java/org/apache/flink/runtime/webmonitor/handlers/ArtifactRequestBody.java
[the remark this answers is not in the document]
sunjincheng121: Yes, I know we should have a way to let the user upload their fat JAR for Python Job. Furthermore, we also need a way to let the user upload the resources of pythons files, so we need a ZIP folder structure, such as : 
zip:
  /lib ---> for JAR and python files ZIPs.
 /xx/wordcount.py ---> user code
What do you think?

dianfu:
@sunjincheng121 I think packaging the dependencies into the zip file makes much sense to me. I have updated the PR accordingly.

sunjincheng121:
I found the time out error of CI, so I have restarted the CI.

tillrohrmann:
Before diving into the review of this rather large PR, I wanted to ask whether the web based submission is a crucial feature for Flink 1.9 or not? I'm wondering whether the client based submission of Python programs wouldn't be enough as a first step.

dianfu:
@tillrohrmann I agree that client based submission can meet most requirements, while I guess there may be also users only using REST API. It's cool if we can support the submission of Python jobs for all kinds of users in 1.9.
Regarding to the large PR, actually most changes are copying the JarXXX to ArtifactXXX which are all in the first commit. The changes of the last two commits are pretty small.

sunjincheng121:
@tillrohrmann this feature not the crucial feature for Flink 1.9,  it's not must-have for Flink 1.9. but I also suggest adding this change to Flink 1.9, due to the following reasons:

-  API refactoring: this PR is pretty big like you said, but most of the change is refactoring, i.e. 
the change is: `jars` -> `artifacts`.  And I think if we ensure that the change of `jars` -> `artifacts` is the right thing, I suggest do this change in Flink 1.9.  :) 
- REST API for python: I think you are right, without REST API we can using the client to submit the Python Job, but we only need to do a small change for the support this feature(only 
300+ line change). I suggest adding this to  Flink 1.9.  :)  But I agree with you that this is nice to have, not must-have. 

image

What do you think?

sunjincheng121:
@tillrohrmann do you still think we should not merge this change into 1.9? just double check and confirm. :)

elanv:
How is it going? I want to submit a python job inside application, and it would be great if REST API supports it.

dianfu:
@elanv The REST API is still not supported as we want to add support of client based submission as the first step. However, it makes sense to consider the REST API support now as more and more functionalities have already been supported for PyFlink and more and more users are trying it out. As 1.12 is already feature freeze and so I think we could not make it in the 1.12. I will revisit this work and hope we could support it in 1.13.

ikstewa:
Is there still plans to support pyflink job submission via REST api?

PR #8533, opened 2019-05-24
[FLINK-12596][scala-shell] Unify all kinds of cluster via ClusterClient in FlinkShell

zjffdu:

In FlinkShell, we treat different kinds of cluster mode, and use scala pattern matching to handle that. 
This is not necessary and make code unreadable. We can unify them into ClusterClient which support all the cluster modes. 

## Brief change log
1. Refactoring on `FlinkShell.scala`
   * Replace `Option[Either[MiniCluster , ClusterClient[_]]]` with `ClusterClient[_]`
   * Move ClusterClient into `FlinkILoop.scala` so that let `FlinkILoop` control the shutdown of `ClusterClient` (shutdown ClusterClient in `FlinkILoop#closeInterpreter`)
   * Introduce method `startShell(config: Config, in: Option[BufferedReader], out: JPrintWriter)` to make production code and test code consistency. (Don't need to introduce `bufferedReader` for unit testing)

2. Refactoring on unit test code
   * Refactoring method `ScalaShellITCase#processInShell` to make it readable. Now each test will create a new FlinkShell instance which create a new `MiniCluster` by default. 
   * Remove `ScalaShellLocalStartupITCase` as `ScalaShellITCase` already cover this case after this PR.

## Verifying this change

```
ScalaShellITCase
```

  - The serializers: (no)

## Documentation


flink-scala-shell/pom.xml
```diff
@@ -145,6 +145,16 @@ under the License.
 				</configuration>
 			</plugin>
 
+			<plugin>
+				<artifactId>maven-surefire-plugin</artifactId>
+				<configuration>
+					<forkMode>always</forkMode>
+					<environmentVariables>
+						<FLINK_CONF_DIR>${project.build.directory}/test-classes</FLINK_CONF_DIR>
+					</environmentVariables>
```
zjffdu: Pass env `FLINK_CONF_DIR` which is needed by `ScalaShellITCase`

flink-scala-shell/src/test/scala/org/apache/flink/api/scala/ScalaShellITCase.scala
```diff
@@ -51,7 +45,7 @@ class ScalaShellITCase extends TestLogger {
 
     val output: String = processInShell(input)
 
-    Assert.assertTrue(output.contains(
+    Assert.assertTrue(output, output.contains(
```
zjffdu: Add `output` so that we can see the real output when the test fails, otherwise it is difficult to see why this kind of test fails.

zjffdu:
@tillrohrmann Could you help review it ? Thanks

flink-scala-shell/src/main/scala/org/apache/flink/api/scala/FlinkShell.scala
```diff
@@ -194,47 +200,28 @@ object FlinkShell {
...
-      cluster match {
-        case Some(Left(miniCluster)) => miniCluster.close()
-        case Some(Right(yarnCluster)) =>
-          yarnCluster.shutDownCluster()
-          yarnCluster.shutdown()
-        case _ =>
-      }
-    }
-
+    flinkILoop.process(settings)
```
zjffdu: It is not necessary to call `closeInterpreter` implicitly, as it will be called in method `process` https://github.com/scala/scala/blob/2.11.x/src/repl/scala/tools/nsc/interpreter/ILoop.scala#L998

PR #8534, opened 2019-05-24
[FLINK-12614][yarn] Refactor test to not do assertions in @After methods 

zentol:
Refactors the YARN tests to only execute assertion within `@Test` methods. Previously, assertions were also executed in `@After` methods, which tends to make the output on test failure more difficult to read.

I added a `runTests(Runnable)` method to the `YarnTestBase` which encapsulates the previous `@After` assertions. All existing tests were modified to use this method instead.

As a result the majority of changes in this PR are about indentation and exception signatures.

Additionally,
* the yarn client is shutdown in `@After`
* an unused field was removed
* a timeout was removed from 
`YARNSessionCapacitySchedulerITCase#testVCoresAreSetCorrectlyAndJobManagerHostnameAreShownInWebInterfaceAndDynamicPropertiesAndYarnApplicationNameAndTaskManagerSlots` so that the timeouts defined in test themselves cause the failure next time.

### DIFF
--- a/flink-yarn-tests/src/test/java/org/apache/flink/yarn/YarnTestBase.java
+++ b/flink-yarn-tests/src/test/java/org/apache/flink/yarn/YarnTestBase.java
@@ -26,6 +26,7 @@ import org.apache.flink.runtime.clusterframework.BootstrapTools;
 import org.apache.flink.test.util.TestBaseUtils;
 import org.apache.flink.util.Preconditions;
 import org.apache.flink.util.TestLogger;
+import org.apache.flink.util.function.RunnableWithException;
 import org.apache.flink.yarn.cli.FlinkYarnSessionCli;
 
 import org.apache.commons.io.FileUtils;
@@ -42,7 +43,6 @@ import org.apache.hadoop.yarn.api.records.ContainerId;
 import org.apache.hadoop.yarn.api.records.YarnApplicationState;
 import org.apache.hadoop.yarn.client.api.YarnClient;
 import org.apache.hadoop.yarn.conf.YarnConfiguration;
-import org.apache.hadoop.yarn.exceptions.YarnException;
 import org.apache.hadoop.yarn.server.MiniYARNCluster;
 import org.apache.hadoop.yarn.server.nodemanager.NodeManager;
 import org.apache.hadoop.yarn.server.nodemanager.containermanager.container.Container;
@@ -200,7 +200,7 @@ public abstract class YarnTestBase extends TestLogger {
 	}
 
 	@Before
-	public void checkClusterEmpty() {
+	public void setupYarnClient() {
 		if (yarnClient == null) {
 			yarnClient = YarnClient.createYarnClient();
 			yarnClient.init(getYarnConfiguration());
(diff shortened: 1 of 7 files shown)


flink-yarn-tests/src/test/java/org/apache/flink/yarn/YarnTestBase.java
```diff
@@ -212,30 +212,45 @@ public void checkClusterEmpty() {
...
+				final List<String> runningApps = yarnClient.getApplications().stream()
+					.filter(YarnTestBase::isApplicationRunning)
+					.map(app -> "App " + app.getApplicationId() + " is in state " + app.getYarnApplicationState() + '.')
+					.collect(Collectors.toList());
+				if (!runningApps.isEmpty()) {
+					Assert.fail("There is at least one application on the cluster that is not finished." + runningApps);
+				}
 			}
+		} catch (Throwable t) {
+			throw new AssertionError(ExceptionUtils.firstOrSuppressed(t, testFailure));
```
tillrohrmann: I think this won't report the `testFailure` if there is no `t`.

flink-yarn-tests/src/test/java/org/apache/flink/yarn/YarnTestBase.java
```diff
@@ -212,30 +212,45 @@ public void checkClusterEmpty() {
...
+		}
 
-		while (deadline.hasTimeLeft() && isAnyJobRunning) {
-			try {
-				Thread.sleep(500);
-			} catch (InterruptedException e) {
-				Assert.fail("Should not happen");
-			}
-			isAnyJobRunning = yarnClient.getApplications().stream()
+		try {
```
tillrohrmann: I think this block should be the `finally` block of the `try { test.run() } finally { ... }`

flink-yarn-tests/src/test/java/org/apache/flink/yarn/YarnTestBase.java
```diff
@@ -212,30 +212,45 @@ public void checkClusterEmpty() {
...
+	}
 
-		boolean isAnyJobRunning = yarnClient.getApplications().stream()
-			.anyMatch(YarnTestBase::isApplicationRunning);
+	protected void runTest(RunnableWithException test) {
+		Throwable testFailure = null;
+		try {
+			test.run();
+		} catch (Throwable t) {
+			testFailure = t;
```
tillrohrmann: No need to catch `t` if the application shut down is done in a `finally` block.

flink-yarn-tests/src/test/java/org/apache/flink/yarn/YarnTestBase.java
zentol: how can there be no `t` in this catch block?

flink-yarn-tests/src/test/java/org/apache/flink/yarn/YarnTestBase.java
zentol: What value would that add? We'd still need both try-catch blocks.

flink-yarn-tests/src/test/java/org/apache/flink/yarn/YarnTestBase.java
zentol: Since the finally block can throw exceptions, without catching the exception here you may end up swallowing the test exception entirely, which would be worse than the current state.

flink-yarn-tests/src/test/java/org/apache/flink/yarn/YarnTestBase.java
tillrohrmann: Why do we need both try-catch blocks? I think we don't need this.

flink-yarn-tests/src/test/java/org/apache/flink/yarn/YarnTestBase.java
tillrohrmann: No, this would not happen. If the finally block throws an exception it would either throw this exception if the try block has not thrown an exception or add the finally exception as a suppressed exception to the try exception.

flink-yarn-tests/src/test/java/org/apache/flink/yarn/YarnTestBase.java
tillrohrmann: Where would `t` come from? Would it rely on the fact that a failing test does not properly shuts down the Yarn application? If yes, then I think this is a brittle mechanism. As soon as we have a test which terminates the Yarn application in the failure case, it would break.

zentol: ah, now I see what you mean.
If this try block doesn't throw an exception we never re-throw the original failure.  That is indeed a defect, changing it.

flink-yarn-tests/src/test/java/org/apache/flink/yarn/YarnTestBase.java
zentol: are you saying that, given the piece of code below, exception "2" would be automatically suppressed?
```
try {
	throw new Exception("1");
} finally {
	throw new Exception("2");
}
```
According to this [SO post](https://stackoverflow.com/questions/19583374/are-exceptions-thrown-in-a-finally-block-added-to-the-list-of-suppressed-excepti) this only works when using try-with-resources.
Thus in order to suppress an exception thrown in the `finally` part we'd still have to catch the test failure exception beforehand.
Which we could of course, but we'd end up with:
```
Throwable testFailure = null;
try {
	test.run();
} catch (Throwable t) {
	testFailure = t;
} finally {
	try {
		Deadline deadline = Deadline.now().plus(Duration.ofSeconds(10));

		boolean isAnyJobRunning = yarnClient.getApplications().stream()
			.anyMatch(YarnTestBase::isApplicationRunning);

		while (deadline.hasTimeLeft() && isAnyJobRunning) {
			try {
				Thread.sleep(500);
			} catch (InterruptedException e) {
				Assert.fail("Should not happen");
			}
			isAnyJobRunning = yarnClient.getApplications().stream()
				.anyMatch(YarnTestBase::isApplicationRunning);
		}

		if (isAnyJobRunning) {
			final List<String> runningApps = yarnClient.getApplications().stream()
				.filter(YarnTestBase::isApplicationRunning)
				.map(app -> "App " + app.getApplicationId() + " is in state " + app.getYarnApplicationState() + '.')
			if (!runningApps.isEmpty()) {
				Assert.fail("There is at least one application on the cluster that is not finished." + runningApps);
			}
		}
	} catch (Throwable t) {
		throw new AssertionError(ExceptionUtils.firstOrSuppressed(t, testFailure));
	}
}

if (testFailure != null) {
	throw new AssertionError(testFailure);
}
```

This is what I meant in my other comment about having to retain both try-catch blocks. We need the outer one to catch the test exception so _can_ suppress later exceptions, and we need the inner one to _do_ the actual suppression.

tillrohrmann: You are right @zentol. What about the following idea: We encapsulate the cleanup logic in an `AutoCloseable` instance.

```
protected void runTest(RunnableWithException test) throws Exception {
		try (final CleanupYarnApplication ignored = new CleanupYarnApplication()){
			test.run();
		}
	}

private class CleanupYarnApplication implements AutoCloseable {
	@Override
	public void close() throws Exception {
		Deadline deadline = Deadline.now().plus(Duration.ofSeconds(10));

		boolean isAnyJobRunning = yarnClient.getApplications().stream()
			.anyMatch(YarnTestBase::isApplicationRunning);

		while (deadline.hasTimeLeft() && isAnyJobRunning) {
			try {
				Thread.sleep(500);
			} catch (InterruptedException e) {
				Assert.fail("Should not happen");
			}
			isAnyJobRunning = yarnClient.getApplications().stream()
				.anyMatch(YarnTestBase::isApplicationRunning);
		}

		if (isAnyJobRunning) {
			final List<String> runningApps = yarnClient.getApplications().stream()
				.filter(YarnTestBase::isApplicationRunning)
				.map(app -> "App " + app.getApplicationId() + " is in state " + app.getYarnApplicationState() + '.')
			if (!runningApps.isEmpty()) {
				Assert.fail("There is at least one application on the cluster that is not finished." + runningApps);
			}
		}
	}
}
```

PR #7782, opened 2019-02-20
[FLINK-11693] Add KafkaSerializationSchema that directly uses ProducerRecord

aljoscha:
This is a rebased and cleaned-up version of #6615 that has the Producer changes, this also has the changes of #7781.

# What is the purpose of the change

* Add a new serialization schema that directly uses `ProducerRecord`, thereby exposing all available Kafka functionality.

## Brief change log

* We introduce the new schema, along with adapters for existing schema and partitioner.
* Functionality for custom Flink partitioners that was previously hardcoded in the consumers is now isolated in the new `KafkaSerializationSchemaAdapterBase`.

## Verifying this change

* New IT cases are added to verify that headers can be written and read

PR #8535, opened 2019-05-24
[FLINK-11693] Add KafkaSerializationSchema that uses ProducerRecord

aljoscha:
This is a rebased and cleaned-up version of #7782

# What is the purpose of the change

* Add a new serialization schema that directly uses `ProducerRecord`, thereby exposing all available Kafka functionality.

## Brief change log

* We introduce the new schema, along with backwards-compatibility code in the different FlinkKafkaProducers
* This extends the existing ITCases to also use the new serializer and still use the old serializer

(side note, the Kafka tests are quite heavy and probably shouldn't be like this, but the aim of this PR is not to fix that.)

### DIFF
--- a/flink-connectors/flink-connector-kafka-base/src/main/java/org/apache/flink/streaming/connectors/kafka/KafkaSerializationSchema.java
+++ b/flink-connectors/flink-connector-kafka-base/src/main/java/org/apache/flink/streaming/connectors/kafka/KafkaSerializationSchema.java
@@ -1,0 +1,45 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.streaming.connectors.kafka;
+
+import org.apache.flink.annotation.PublicEvolving;
+
+import org.apache.kafka.clients.producer.ProducerRecord;
+
+import javax.annotation.Nullable;
+
+import java.io.Serializable;
+
+/**
+ * A {@link KafkaSerializationSchema} defines how to serialize values of type {@code T} into
+ * {@link ProducerRecord ProducerRecords}.
+ *
+ * @param <T> the type of values being serialized
+ */
+@PublicEvolving
+public interface KafkaSerializationSchema<T> extends Serializable {
+
+	/**
+	 * Serializes given element and returns it as a {@link ProducerRecord}.
+	 *
+	 * @param element element to be serialized
+	 * @param timestamp timestamp (can be null)
+	 * @return Kafka {@link ProducerRecord}
+	 */
+	ProducerRecord<byte[], byte[]> serialize(T element, @Nullable Long timestamp);
+}
--- a/flink-connectors/flink-connector-kafka-base/src/main/java/org/apache/flink/streaming/connectors/kafka/KafkaSerializationSchema.java
+++ b/flink-connectors/flink-connector-kafka-base/src/main/java/org/apache/flink/streaming/connectors/kafka/KafkaSerializationSchema.java
@@ -26,8 +26,13 @@ import javax.annotation.Nullable;
 import java.io.Serializable;
 
 /**
- * A {@link KafkaSerializationSchema} defines how to serialize values of type {@code T} into
- * {@link ProducerRecord ProducerRecords}.
+ * A {@link KafkaSerializationSchema} defines how to serialize values of type {@code T} into {@link
+ * ProducerRecord ProducerRecords}.
+ *
+ * <p>Please also implement {@link KafkaContextAware} if your serialization schema needs
+ * information
+ * about the available partitions and the number of parallel subtasks along with the subtask ID on
+ * which the Kafka Producer is running.
  *
  * @param <T> the type of values being serialized
  */
--- a/flink-connectors/flink-connector-kafka-base/src/test/java/org/apache/flink/streaming/connectors/kafka/KafkaConsumerTestBase.java
+++ b/flink-connectors/flink-connector-kafka-base/src/test/java/org/apache/flink/streaming/connectors/kafka/KafkaConsumerTestBase.java
@@ -81,11 +81,13 @@ import kafka.server.KafkaServer;
 import org.apache.commons.io.output.ByteArrayOutputStream;
 import org.apache.kafka.clients.consumer.ConsumerRecord;
 import org.apache.kafka.clients.producer.ProducerConfig;
+import org.apache.kafka.clients.producer.ProducerRecord;
 import org.apache.kafka.common.errors.TimeoutException;
 import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Rule;
 
+import javax.annotation.Nullable;
 import javax.management.MBeanServer;
 import javax.management.ObjectName;
 
@@ -1128,7 +1130,7 @@ public abstract class KafkaConsumerTestBase extends KafkaTestBaseWithFlink {
 	 * Test producing and consuming into multiple topics.
 	 * @throws Exception
 	 */
-	public void runProduceConsumeMultipleTopics() throws Exception {
+	public void runProduceConsumeMultipleTopics(boolean useLegacySchema) throws Exception {
 		final int numTopics = 5;
 		final int numElements = 20;
 
@@ -1166,12 +1168,17 @@ public abstract class KafkaConsumerTestBase extends KafkaTestBaseWithFlink {
 			}
 		});
 
-		Tuple2WithTopicSchema schema = new Tuple2WithTopicSchema(env.getConfig());
-
 		Properties props = new Properties();
 		props.putAll(standardProps);
 		props.putAll(secureProps);
-		kafkaServer.produceIntoKafka(stream, "dummy", schema, props, null);
+
+		if (useLegacySchema) {
+			Tuple2WithTopicSchema schema = new Tuple2WithTopicSchema(env.getConfig());
+			kafkaServer.produceIntoKafka(stream, "dummy", schema, props, null);
+		} else {
+			TestDeSerializer schema = new TestDeSerializer(env.getConfig());
+			kafkaServer.produceIntoKafka(stream, "dummy", schema, props);
+		}
 
 		env.execute("Write to topics");
 
@@ -1179,7 +1186,13 @@ public abstract class KafkaConsumerTestBase extends KafkaTestBaseWithFlink {
 		env = StreamExecutionEnvironment.getExecutionEnvironment();
 		env.getConfig().disableSysoutLogging();
 
-		stream = env.addSource(kafkaServer.getConsumer(topics, schema, props));
+		if (useLegacySchema) {
+			Tuple2WithTopicSchema schema = new Tuple2WithTopicSchema(env.getConfig());
+			stream = env.addSource(kafkaServer.getConsumer(topics, schema, props));
+		} else {
+			TestDeSerializer schema = new TestDeSerializer(env.getConfig());
+			stream = env.addSource(kafkaServer.getConsumer(topics, schema, props));
+		}
 
 		stream.flatMap(new FlatMapFunction<Tuple3<Integer, Integer, String>, Integer>() {
 			Map<String, Integer> countPerTopic = new HashMap<>(numTopics);
@@ -2194,12 +2207,12 @@ public abstract class KafkaConsumerTestBase extends KafkaTestBaseWithFlink {
 		}
 	}
 
-	private static class Tuple2WithTopicSchema implements KafkaDeserializationSchema<Tuple3<Integer, Integer, String>>,
-		KeyedSerializationSchema<Tuple3<Integer, Integer, String>> {
+	private abstract static class TestDeserializer implements
+			KafkaDeserializationSchema<Tuple3<Integer, Integer, String>> {
 
-		private final TypeSerializer<Tuple2<Integer, Integer>> ts;
+		protected final TypeSerializer<Tuple2<Integer, Integer>> ts;
 
-		public Tuple2WithTopicSchema(ExecutionConfig ec) {
+		public TestDeserializer(ExecutionConfig ec) {
 			ts = TypeInformation.of(new TypeHint<Tuple2<Integer, Integer>>(){}).createSerializer(ec);
 		}
 
@@ -2218,6 +2231,14 @@ public abstract class KafkaConsumerTestBase extends KafkaTestBaseWithFlink {
 		@Override
 		public TypeInformation<Tuple3<Integer, Integer, String>> getProducedType() {
 			return TypeInformation.of(new TypeHint<Tuple3<Integer, Integer, String>>(){});
+		}
+	}
+
+	private static class Tuple2WithTopicSchema extends TestDeserializer
+			implements KeyedSerializationSchema<Tuple3<Integer, Integer, String>> {
+
+		public Tuple2WithTopicSchema(ExecutionConfig ec) {
+			super(ec);
 		}
 
 		@Override
@@ -2241,5 +2262,29 @@ public abstract class KafkaConsumerTestBase extends KafkaTestBaseWithFlink {
 		public String getTargetTopic(Tuple3<Integer, Integer, String> element) {
 			return element.f2;
 		}
+	}
+
+	private static class TestDeSerializer extends TestDeserializer
+			implements KafkaSerializationSchema<Tuple3<Integer, Integer, String>> {
+
+		public TestDeSerializer(ExecutionConfig ec) {
+			super(ec);
+		}
+
+		@Override
+		public ProducerRecord<byte[], byte[]> serialize(
+				Tuple3<Integer, Integer, String> element, @Nullable Long timestamp) {
+			ByteArrayOutputStream by = new ByteArrayOutputStream();
+			DataOutputView out = new DataOutputViewStreamWrapper(by);
+			try {
+				ts.serialize(new Tuple2<>(element.f0, element.f1), out);
+			} catch (IOException e) {
+				throw new RuntimeException("Error" , e);
+			}
+			byte[] serializedValue = by.toByteArray();
+
+			return new ProducerRecord<>(element.f2, serializedValue);
+		}
+
 	}
 }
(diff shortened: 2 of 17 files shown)


aljoscha:
I don't like the backwards-compatibility story too much but I'm afraid we have to deal with that for at least the next release.

aljoscha:
Please wait on reviewing this, it still has some problems in the interactions with the new serialization schema and `FlinkKafkaPartitioner`.

aljoscha:
Also, @tweise or @jgrier is this feature relevant for you? Maybe you also have some comments/input?

tweise:
@aljoscha glad to see this change, we may be interested to use the headers with 0.11

aljoscha:
Alright, this PR is now ready for review. I decided to implement this only for the modern Kafka connector because adding it to all the old connectors was too much of a hassle. I'm open to reviewing if anyone wants to add it for the old connectors as well, though.

aljoscha:
@tweise and @jgrier Could you maybe have a look at this to see if the API is good for you like this? Also @alexeyt820?

@becketqin Could you maybe review this as well?

alexeyt820:
I was thinking why it is one to one, maybe `serialize` should return list of `ProducerRecord`. Consider use case when you want to publish data in all partitions.

flink-connectors/flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/FlinkKafkaProducer.java
```diff
@@ -495,6 +520,11 @@ public FlinkKafkaProducer(
 		this.schema = checkNotNull(serializationSchema, "serializationSchema is null");
 		this.producerConfig = checkNotNull(producerConfig, "producerConfig is null");
 		this.flinkKafkaPartitioner = checkNotNull(customPartitioner, "customPartitioner is null").orElse(null);
+		if (customPartitioner.isPresent()) {
```
becketqin: Hmm, not sure if `KafkaSerializationSchema` can fully replace the partitioner. The current API does not provide number of partitions in `KafkaSerializationSchema.serialize()`, right? So the serializer will not be able to know the total number of partitions. Should we add that to the method arguments?

becketqin:
@alexeyt820 Multicast and broadcast is an interesting use case. I am not sure if the serializer is the best place to do that. It seems the processing node should do this instead of the sink node. However, the problem here is that the processing node may not know how many partitions are there, and the sink node probably does not know which partitions should a record be sent to. 

For broadcast this might be easily resolved by having a special flag on the record, so the sink node just send the record to every partition. For multicast this is a little trickier. But in any case, putting it into the serializer seems mixing things up. I'd suggest to design multicast and broadcast mechanism in a more generic and explicit way that could be reused by all the Sink connectors.

aljoscha:
Thanks for the review, @becketqin. I tried the solution with separate constructors but couldn't get it to work, see here: https://github.com/aljoscha/flink/commits/pr/6615-simplified-two-constructor. The problem is that I don't want to duplicate the common part of the constructor. Do you maybe have a suggestion for that?

becketqin:
@aljoscha Thanks for the pointer the earlier PR. The solution in my mind was slightly different from but pretty similar to what you have in the PR. 

I was thinking that we can have a private constructor that takes both `KafkaSchemaSerializer` and `KeyedSerializationSchema` in the arguments, but one of them will always be null which is guaranteed by the public constructors. So the public constructors will always call this private constructor, but with one of the two serialization schema set to null. Would that address the problem of having duplicate code in the constructors?

aljoscha:
@becketqin yes I also thought about it like that this morning 😅. I'll update the PR

flink-connectors/flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/FlinkKafkaProducer.java
```diff
@@ -481,27 +508,140 @@ public FlinkKafkaProducer(
...
+				customPartitioner.orElse(null),
+				null, /* kafka serialization schema */
+				producerConfig,
+				semantic,
+				kafkaProducersPoolSize);
+	}
+
+	/**
+	 * Creates a {@link FlinkKafkaProducer} for a given topic. The sink produces its input to
+	 * the topic. It accepts a keyed {@link KafkaSerializationSchema} for serializing records to
```
becketqin: It accepts a ~~keyed~~ {@link KafkaSerializationSchema}...?

flink-connectors/flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/FlinkKafkaProducer.java
```diff
@@ -481,27 +508,140 @@ public FlinkKafkaProducer(
...
+				defaultTopicId,
+				serializationSchema,
+				producerConfig,
+				semantic,
+				DEFAULT_KAFKA_PRODUCERS_POOL_SIZE);
+	}
+
+	/**
+	 * Creates a FlinkKafkaProducer for a given topic. The sink produces its input to
+	 * the topic. It accepts a keyed {@link KafkaSerializationSchema} and possibly a custom {@link FlinkKafkaPartitioner}.
```
becketqin: Ditto above?

flink-connectors/flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/FlinkKafkaProducer.java
```diff
@@ -481,27 +508,140 @@ public FlinkKafkaProducer(
...
+	 * the topic. It accepts a keyed {@link KafkaSerializationSchema} for serializing records to
+	 * a {@link ProducerRecord}, including partitioning information.
+	 *
+	 * @param defaultTopicId The default topic to write data to
+	 * @param serializationSchema A serializable serialization schema for turning user objects into a kafka-consumable byte[] supporting key/value messages
+	 * @param producerConfig Configuration properties for the KafkaProducer. 'bootstrap.servers.' is the only required argument.
+	 * @param semantic Defines semantic that will be used by this producer (see {@link FlinkKafkaProducer.Semantic}).
+	 */
+	public FlinkKafkaProducer(
+			String defaultTopicId,
```
becketqin: Nit pick: just `defaultTopic`?

flink-connectors/flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/FlinkKafkaProducer.java
```diff
@@ -481,27 +508,140 @@ public FlinkKafkaProducer(
...
-		this.schema = checkNotNull(serializationSchema, "serializationSchema is null");
+
+		if (kafkaSchema != null) {
+			this.keyedSchema = null;
+			this.kafkaSchema = kafkaSchema;
+			this.flinkKafkaPartitioner = null;
+			ClosureCleaner.clean(
+					this.kafkaSchema, ExecutionConfig.ClosureCleanerLevel.RECURSIVE, true);
+
+			if (customPartitioner != null) {
```
becketqin: I had an earlier comment about this restriction. It seems the current interface of `KafkaSerializationSchema` cannot fully cover the functionality of the partitioner, because the  implementations of `KafkaSerializationSchema` do not have enough information, e.g. number of partitions, task parallelisms, etc. Should we add those information to the arguments of `KafkaSerializationSchema.serialize()`?

flink-connectors/flink-connector-kafka-base/src/test/java/org/apache/flink/streaming/connectors/kafka/KafkaConsumerTestBase.java
```diff
@@ -1217,6 +1219,99 @@ public void flatMap(Tuple3<Integer, Integer, String> value, Collector<Integer> o
 		}
 	}
 
+	/**
+	 * Test producing and consuming into multiple topics.
+	 * @throws Exception
+	 */
+	public void runProduceConsumeMultipleTopicsWithKafkaSerializer() throws Exception {
```
becketqin: Can we avoid duplicating the test code by doing the following?
1. have a private method taking schema as an argument.
2. let `runProduceConsumeMultipleTopicsWithKafkaSerializer` and `runProduceConsumeMultipleTopicsWithLegacySerializer` call that method and pass in different schema?

flink-connectors/flink-connector-kafka-base/src/test/java/org/apache/flink/streaming/connectors/kafka/KafkaConsumerTestBase.java
```diff
@@ -2242,4 +2337,47 @@ public String getTargetTopic(Tuple3<Integer, Integer, String> element) {
 			return element.f2;
 		}
 	}
+
+	private static class TestDeSerializer implements KafkaDeserializationSchema<Tuple3<Integer, Integer, String>>,
```
becketqin: Minor:
Can we have a private static deserializer class (may be abstract). And let `TestDeserializer` and `Tuple2WithTopicSchema` extends that deserializer class but implement different serializer interface?

flink-connectors/flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/FlinkKafkaProducer.java
aljoscha: 👍 I'm changing it only in the new constructors, or should I change it everywhere right away?

flink-connectors/flink-connector-kafka-base/src/test/java/org/apache/flink/streaming/connectors/kafka/KafkaConsumerTestBase.java
aljoscha: How do you mean? The two schemas don't share a common base class (except `Object`).

aljoscha:
Thanks @becketqin! I like your reviews, you are very thorough. 👌 

I fixed the smaller things. I agree on the problem that the new `KafkaSerializerSchema` cannot completely subsume the old `KeyedSerializerSchema` + partitioner. I thought about how we should do that in the interface, if we might want to add other parameters to the method in the future. I was hesitant to introduce a context object for this, but we might have to do it. What do you think?

flink-connectors/flink-connector-kafka-base/src/test/java/org/apache/flink/streaming/connectors/kafka/KafkaConsumerTestBase.java
becketqin: I was thinking of using something like `instanceOf`. Given it is private method, we don't need to worry much about the wrong argument type. We may need to add some comments to explain that. Admittedly, this is not ideal, but personally I feel it is slightly better than having a lot of duplicate code.

becketqin:
@aljoscha Thanks for updating the patch and the quick response. 

Regarding the partitioner issue, just thinking loudly, from design principle perspective, there might be two reasonable approaches:

1. Introduce a generic partitioner that is used across all the connectors that supports partitioning. Flink owns the interface and ensure unified user experience in all the connectors. So this would become a pattern for all the connector implementations.

2. Instead of introducing unified partitioner shared by all the connectors, follow the suggested approach of each connected system. E.g. in Kafka the partitioner is specified via a `partitioner.class` config. This way, user experience with the connected system is honored.

For partitioners, it seems Flink does not follow design principle 1, but following principle 2 here is a little tricky because we do not use the key and value serializer the way Kafka expected, therefore the configured Kafka partitioner will only be able to see `byte[]` instead of original objects.

Given the current status, I agree that we may as well move towards design principle 1. i.e. Introduce something like:
```
interface Partitioner<T, P, CONTEXT extends RuntimeContext> {
   P partition(T, CONTEXT extend RuntimeContext);
}
```
P is the partition id, T is the type, and CONTEXT is the context for this particular system. RuntimeContext might be an overkill. We can have a simpler one.

For Kafka it would be:
```
interface KafkaPartitioner extends Partitioner<IN, Integer, KafkaContext> {
}
```

Does this sound reasonable to you?

flink-connectors/flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/FlinkKafkaProducer.java
becketqin: Personally I feel it might be better to change it across the board to avoid unnecessary confusion.

becketqin: BTW, I just realized that according to the code style guideline, such renaming might need to be done in a separate commit.

flink-connectors/flink-connector-kafka-base/src/test/java/org/apache/flink/streaming/connectors/kafka/KafkaConsumerTestBase.java
aljoscha: done

aljoscha:
I updated the PR, PTAL (please take another look 😅)

Regarding partitioning: yes, this is a shortcoming in the current design and I didn't want to tackle it because I see the same two options that you mentioned and both require some work. In general I think the fact that we have our own serializers and always use `ByteArrayDeserializer` (and serializers) for Kafka is a bit weird and means that people can't use the same serializers that they use elsewhere with Kafka.

Regarding your option 1., I think this is tricky, because the partition that we partitioner returns has to be put into the Kafka `ProducerRecord`. This worked before because the framework code was assembling the `ProducerRecord`. When the `KafkaSerializationSchema` returns a `ProducerRecord`, this is not possible because partition (and the other fields) are `final`.

becketqin:
@aljoscha Yes, you are right. We are not able to change the partition once the `ProducerRecrod` is there. We will have to create another `ProducerRecord` in order to do that, but that will result in some overhead. Introducing a context sounds reasonable to me. This has an extra benefit of allowing context aware serialization. There might be two flavors of doing this. 

1. Pass the context as a method argument.
2. A `ContextAware` interface like `RichFunction`, with a `setContext()` and `getContext()` method.

Personally I slightly prefer flavor 2, because it avoid keeping the Context object outside of the class which actually uses it. But I don't have a strong opinion on this.

aljoscha:
@becketqin Do we want to do that in this PR or as an addition?

becketqin:
@aljoscha It seems we have to do it in this PR to avoid potential backwards incompatible interface change or functionality regression. I am worrying about the following:

1. we do one of the following:
   a. We introduce a `KafkaSerializationSchema.serialize(T element, Long timestamp, KafkaContext context)`;
   b. We simply keep the patch as is, i.e. partitioner is no longer supported.

2. Flink 1.9 releases and users start to provide implementation of `KafkaSerializationSchema`;

3. We change the API to use `ContextAware`, so `KafkaSerializationSchema.serialize()` no longer needs a KafkaContext argument. Instead the KafkaContext argument will be set using `setContext(KafkaContext)` for the `KafkaSerializationSchema`. 

If we did 1a, for backwards compatible purpose, in the next release, we will have to keep the old method signature and add a new one.
If we did 1b, Flink 1.9 users will not be able to use partitioner, which is a functionality regression.

But if step 1 does not go into release 1.9, we can do 1b with an additional patch without any problem.

flink-connectors/flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/FlinkKafkaProducer.java
```diff
@@ -625,34 +765,53 @@ public void onCompletion(RecordMetadata metadata, Exception exception) {
...
+			} else {
+				record = new ProducerRecord<>(targetTopic, null, timestamp, serializedKey, serializedValue);
+			}
 		} else {
-			record = new ProducerRecord<>(targetTopic, null, timestamp, serializedKey, serializedValue);
+			if (kafkaSchema instanceof ContextAwareSerializationSchema) {
+				ContextAwareSerializationSchema<IN> contextAwareSchema =
+						(ContextAwareSerializationSchema<IN>) kafkaSchema;
+
+				String targetTopic = contextAwareSchema.getTargetTopic(next);
```
becketqin: Do we need to check if the returned `targetTopic` is null? The logic in case of keyed schema was to use default topic instead when `getTargetTopic()` returns null.

flink-connectors/flink-connector-kafka-base/src/main/java/org/apache/flink/streaming/connectors/kafka/ContextAwareSerializationSchema.java
```diff
@@ -0,0 +1,61 @@
...
+/**
+ * An interface for {@link KafkaSerializationSchema KafkaSerializationSchemas} that need information
+ * about the context where the Kafka Producer is running along with information about the available
+ * partitions.
+ *
+ * <p>You only need to override the methods for the information that you need. However, {@link
+ * #getTargetTopic(Object)} is required because it is used to determine the available partitions.
+ */
+@PublicEvolving
+public interface ContextAwareSerializationSchema<T> {
```
becketqin: Does it make sense to name the class `KafkaContextAware`? If we do this, in the future, we can introduce a common `ContextAware` interface with setters for InstanceId, Parallelism and other common info. And the `ContextAware` may have multiple subclasses such as `HDFSContextAware`, `RedisContextAware`, etc. That change will be fully backwards compatible without any naming confusion.

flink-connectors/flink-connector-kafka-base/src/main/java/org/apache/flink/streaming/connectors/kafka/KafkaSerializationSchema.java
```diff
@@ -0,0 +1,49 @@
...
+
+import javax.annotation.Nullable;
+
+import java.io.Serializable;
+
+/**
+ * A {@link KafkaSerializationSchema} defines how to serialize values of type {@code T} into
+ * {@link ProducerRecord ProducerRecords}.
+ *
+ * <p>Use {@link ContextAwareSerializationSchema} if your serialization schema needs information
```
becketqin: Would the following doc be slightly clearer? 

"Please also implement {@link KafkaContextAware} interface if your serialization schema ..."

flink-connectors/flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/FlinkKafkaProducer.java
aljoscha: Fixing 👍

flink-connectors/flink-connector-kafka-base/src/main/java/org/apache/flink/streaming/connectors/kafka/ContextAwareSerializationSchema.java
aljoscha: Yes, I'm renaming. I wasn't happy with the name 😅

aljoscha:
I pushed changes, PTAL.

aljoscha:
Merged. Thanks for your help, @becketqin!

imperio-wxm:
Hi,
- flink-1.10

I found `FlinkKafkaProducer0xxx` unable to support  `KafkaSerializationSchema`, because of `FlinkKafkaProducer0xxx` need `KeyedSerializationSchema` or `SerializationSchema` as a serialization tool, I can't initialize like this.

```java
public class ProducerSerializationSchema implements KafkaSerializationSchema<ConsumerRecord<byte[], byte[]>> {
    @Override
    public ProducerRecord<byte[], byte[]> serialize(ConsumerRecord<byte[], byte[]> element, @Nullable Long timestamp) {
        // do my logic
        return null;
    }
}

// flink streaming
stream.addSink(new FlinkKafkaProducer011(
	"default",
	new ProducerSerializationSchema(),
	properties,
	FlinkKafkaProducer011.Semantic.EXACTLY_ONCE
));
```

PR #8536, opened 2019-05-24
[FLINK-12568][hive] Implement OutputFormat to write Hive tables

lirui-apache:

To implement OutputFormat to write Hive tables.

## Brief change log

  - Implemented `HiveTableOutputFormat` that can be used to write Hive tables.
  - `HiveTableOutputFormat` uses Hive's output formats, SerDe, and RecordWriter to write data.
  - Added `HMSClientFactory` as a unified factory to create HMS client.
  - Added `HiveTableUtil` to provide util methods to access Hive tables.

## Verifying this change

Added a unit test `HiveTableOutputFormatTest` to test HiveTableOutputFormat.

  - The serializers: no

## Documentation


lirui-apache:
@xuefuz @bowenli86 please have a look. Thanks.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/batch/connectors/hive/HiveTableSink.java
```diff
@@ -0,0 +1,142 @@
...
+	private String toStagingDir(String finalDir, Configuration conf) throws IOException {
+		String res = finalDir;
+		if (!finalDir.endsWith(Path.SEPARATOR)) {
+			res += Path.SEPARATOR;
+		}
+		// TODO: may append something more meaningful than a timestamp, like query ID
+		res += ".staging_" + System.currentTimeMillis();
+		Path path = new Path(res);
+		FileSystem fs = path.getFileSystem(conf);
+		Preconditions.checkState(fs.exists(path) || fs.mkdirs(path), "Failed to create staging dir " + path);
```
xuefuz: Not sure if checkState is a good use here. The method throws IOException, which seems more applicable than asserting a condition.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/batch/connectors/hive/HiveTableUtil.java
```diff
@@ -0,0 +1,103 @@
...
+			default:
+				throw new IOException("Unsupported Hive type category " + type.getCategory());
+		}
+	}
+
+	/**
+	 * Converts a Flink {@link TypeInformation} to corresponding Hive {@link TypeInfo}.
+	 */
+	public static TypeInfo toHiveTypeInfo(TypeInformation flinkType) {
+		if (flinkType.equals(BasicTypeInfo.STRING_TYPE_INFO)) {
```
xuefuz: switch statement instead?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalog.java
```diff
@@ -590,6 +578,17 @@ private  static Table instantiateHiveTable(ObjectPath tablePath, CatalogBaseTabl
 		return hiveTable;
 	}
 
+	private static void setStorageFormat(StorageDescriptor sd, Map<String, String> properties) {
+		// TODO: simply use text format for now
```
xuefuz: TODO is supposed to list what to do in the future. :)

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/batch/connectors/hive/HiveTableOutputFormat.java
```diff
@@ -0,0 +1,412 @@
...
+		// since our writers are transient, we don't need to do anything here
+	}
+
+	@Override
+	public void open(int taskNumber, int numTasks) throws IOException {
+		try {
+			StorageDescriptor sd = hiveTablePartition.getStorageDescriptor();
+			serializer = (AbstractSerDe) Class.forName(sd.getSerdeInfo().getSerializationLib()).newInstance();
+			ReflectionUtils.setConf(serializer, jobConf);
+			SerDeUtils.initializeSerDe(serializer, jobConf, tblProperties, null);
```
xuefuz: Nit: tell what null means.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/batch/connectors/hive/HiveTableOutputFormat.java
```diff
@@ -0,0 +1,412 @@
...
+		clonedConf.set(OUTDIR, location);
+		OutputFormat outputFormat;
+		try {
+			StorageDescriptor sd = hiveTablePartition.getStorageDescriptor();
+			Class outputFormatClz = Class.forName(sd.getOutputFormat(), true,
+				Thread.currentThread().getContextClassLoader());
+			outputFormatClz = HiveFileFormatUtils.getOutputFormatSubstitute(outputFormatClz);
+			outputFormat = (OutputFormat) outputFormatClz.newInstance();
+		} catch (InstantiationException | IllegalAccessException | ClassNotFoundException e) {
+			throw new RuntimeException("Unable to instantiate the hadoop output format", e);
```
bowenli86: throw a Flink exception?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalog.java
```diff
@@ -590,6 +578,17 @@ private  static Table instantiateHiveTable(ObjectPath tablePath, CatalogBaseTabl
...
+		String storageFormatName = DEFAULT_HIVE_TABLE_STORAGE_FORMAT;
+		StorageFormatDescriptor sfDescriptor = storageFormatFactory.get(storageFormatName);
+		checkArgument(sfDescriptor != null, "Unknown storage format " + storageFormatName);
```
bowenli86: move input checks to the top?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalog.java
```diff
@@ -590,6 +578,17 @@ private  static Table instantiateHiveTable(ObjectPath tablePath, CatalogBaseTabl
 		return hiveTable;
 	}
 
+	private static void setStorageFormat(StorageDescriptor sd, Map<String, String> properties) {
+		// TODO: simply use text format for now
+		String storageFormatName = DEFAULT_HIVE_TABLE_STORAGE_FORMAT;
+		StorageFormatDescriptor sfDescriptor = storageFormatFactory.get(storageFormatName);
```
bowenli86: nit: can we use full name for variables and not abbreviation?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/batch/connectors/hive/HiveTableUtil.java
```diff
@@ -0,0 +1,103 @@
...
+			// TODO: support complex types
+			default:
+				throw new IOException("Unsupported Hive type category " + type.getCategory());
+		}
+	}
+
+	/**
+	 * Converts a Flink {@link TypeInformation} to corresponding Hive {@link TypeInfo}.
+	 */
+	public static TypeInfo toHiveTypeInfo(TypeInformation flinkType) {
```
bowenli86: isn't this dup of `HiveTypeUtil.toHiveType()`?

flink-connectors/flink-connector-hive/src/test/java/org/apache/flink/batch/connectors/hive/HiveTableSinkTest.java
```diff
@@ -0,0 +1,117 @@
...
+	}
+
+	private BatchTableEnvironment createTableEnv(int parallelism) {
+		Configuration config = new Configuration();
+		config.setInteger(TableConfigOptions.SQL_RESOURCE_DEFAULT_PARALLELISM, 1);
+		StreamExecutionEnvironment env = StreamExecutionEnvironment.createLocalEnvironment(parallelism, config);
+		return BatchTableEnvironment.create(env);
+	}
+
+	private Table getSmall5TupleDataSet(BatchTableEnvironment env) {
```
bowenli86: nit: small**Tuple5**DataSet?

flink-connectors/flink-connector-hive/src/test/java/org/apache/flink/table/catalog/hive/HiveTestUtils.java
```diff
@@ -50,6 +54,7 @@ private static HiveConf getHiveConf() throws IOException {
 		HiveConf hiveConf = new HiveConf();
 		hiveConf.setVar(HiveConf.ConfVars.METASTOREWAREHOUSE, TEMPORARY_FOLDER.newFolder("hive_warehouse").getAbsolutePath());
 		hiveConf.setVar(HiveConf.ConfVars.METASTORECONNECTURLKEY, warehouseUri);
+		hiveConf.setVar(HiveConf.ConfVars.METASTOREURIS, "");
```
bowenli86: why set it to empty string?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/batch/connectors/hive/HiveTableSink.java
lirui-apache: It checks we're in a state that either the output dir exists, or we're able to create it. IOException also seems a reasonable choice. Will use it instead.

flink-connectors/flink-connector-hive/pom.xml
```diff
@@ -53,7 +53,21 @@ under the License.
...
 			<groupId>org.apache.flink</groupId>
-			<artifactId>flink-table-api-java</artifactId>
+			<artifactId>flink-hadoop-compatibility_${scala.binary.version}</artifactId>
+			<version>${project.version}</version>
+			<scope>provided</scope>
+		</dependency>
+
+		<dependency>
+			<groupId>org.apache.flink</groupId>
+			<artifactId>flink-table-planner-blink_${scala.binary.version}</artifactId>
```
KurtYoung: Have you discussed about this? IMO the blink planner should not be dependent.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/batch/connectors/hive/HiveTableUtil.java
lirui-apache: we can't because flink type is not enum

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalog.java
lirui-apache: It doesn't directly check input argument. It checks whether we can get a StorageFormatDescriptor based on the input argument.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/batch/connectors/hive/HiveTableUtil.java
lirui-apache: It's not. We need a `TypeInfo` here and `HiveTypeUtil.toHiveType()` gives the name of a hive type. But I guess this method can be moved to `HiveTypeUtil`

flink-connectors/flink-connector-hive/pom.xml
lirui-apache: Yes we have. I thought the blink planner will eventually replace flink's. Is there any reason why we can't depend on it?

flink-connectors/flink-connector-hive/src/test/java/org/apache/flink/table/catalog/hive/HiveTestUtils.java
lirui-apache: it's not needed anymore, will remove it

flink-connectors/flink-connector-hive/pom.xml
KurtYoung: Firstly, this will not happen very soon. Secondly, i don't see a reason that a planner should be dependent for a connector. Could you explain the reason?

JingsongLi: I think the reason is `BatchTableSink` with `BoundedStream` still in blink planner.
We need unify `TableSource` and `TableSink` interface and  move them to api.

lirui-apache: Yes, we need blink planner for `BatchTableSink ` at the moment

KurtYoung: I'm afraid the source/sink interfaces are still under discussion and the interfaces which blink planner currently used are highly possible not the final solution.

lirui-apache: Thanks for the explanation. So the suggestion is we should instead implement `BatchTableSink` in flink-table-api-java-bridge, right?

KurtYoung: I think it's also not a good idea since we are deprecating `DataSet` now. How about:
1. Decouple the implementation details for this functionality and first merge the parts which unrelated to TableSink interface, like provide an OutputFormat and test it independently.
2. Provide the TableSink implementation after we decide and merge the TableSink interface into `common` or `api` package, and add some integration tests with it.

lirui-apache:
I have removed the TableSink implementation and updated unit test accordingly.
@KurtYoung @xuefuz @bowenli86 please take another look. Thanks.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/util/HiveTypeUtil.java
```diff
@@ -132,4 +133,45 @@ private static TypeInformation toFlinkPrimitiveType(PrimitiveTypeInfo hiveType)
 					String.format("Flink doesn't support Hive primitive type %s yet", hiveType));
 		}
 	}
+
+	/**
+	 * Converts a Flink {@link TypeInformation} to corresponding Hive {@link TypeInfo}.
+	 */
+	public static TypeInfo toHiveTypeInfo(TypeInformation flinkType) {
```
bowenli86: can we try to unify `toHiveTypeInfo()` and `toHiveType()` so we have only one path to covert Flink type to Hive type?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/batch/connectors/hive/HiveTableOutputFormat.java
```diff
@@ -0,0 +1,419 @@
...
+		// since our writers are transient, we don't need to do anything here
+	}
+
+	@Override
+	public void open(int taskNumber, int numTasks) throws IOException {
+		try {
+			StorageDescriptor sd = hiveTablePartition.getStorageDescriptor();
+			serializer = (AbstractSerDe) Class.forName(sd.getSerdeInfo().getSerializationLib()).newInstance();
+			ReflectionUtils.setConf(serializer, jobConf);
+			// Pass null as partition properties, assuming they're same as table properties
```
xuefuz: I think it's possible that a Hive partition can be different from its table in terms of properties. For instance, a partition can has a different storage format. For now, I think we can add a todo item.

lirui-apache:
@bowenli86 sorry for overlooking some comments. Will update soon

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/util/HiveTypeUtil.java
lirui-apache: I'll use Hive's `TypeInfoFactory` to achieve this. Alternatively, we can maintain a mapping from type name to `TypeInfo` ourselves.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/batch/connectors/hive/HiveTableOutputFormat.java
```diff
@@ -0,0 +1,412 @@
...
+			StorageDescriptor sd = hiveTablePartition.getStorageDescriptor();
+			serializer = (AbstractSerDe) Class.forName(sd.getSerdeInfo().getSerializationLib()).newInstance();
+			ReflectionUtils.setConf(serializer, jobConf);
+			SerDeUtils.initializeSerDe(serializer, jobConf, tblProperties, null);
+			outputClass = serializer.getSerializedClass();
+		} catch (IllegalAccessException | SerDeException | InstantiationException | ClassNotFoundException e) {
+			LOG.error("Error initializing Hive serializer");
+			throw new RuntimeException(e);
+		}
+		if (Integer.toString(taskNumber).length() > 6) {
```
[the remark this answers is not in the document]
lirui-apache: The Flink runtime passes this number as the ID of the sub-task. This logic is copied from our internal Blink implementation. I'll remove it for now. We can revisit this when we hit real issues with big IDs.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/batch/connectors/hive/HiveTableOutputFormat.java
```diff
@@ -0,0 +1,412 @@
...
+	private void commitJob(String location) throws IOException {
+		jobConf.set(OUTDIR, location);
+		JobContext jobContext = new JobContextImpl(this.jobConf, new JobID());
+		OutputCommitter outputCommitter = this.jobConf.getOutputCommitter();
+		// finalize HDFS output format
+		outputCommitter.commitJob(jobContext);
+	}
+
+	// converts each field in the record to external object
+	private Object getConvertedRow(BaseRow record) {
```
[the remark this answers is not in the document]
lirui-apache: Yes it's called on every row. But is's not used to compute the partition value of a row. It's used to convert each field in a row to its external type. So we can't get rid of this overhead.

lirui-apache:
Updated to address comments and remove dependency on blink runtime.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/batch/connectors/hive/HiveTableOutputFormat.java
```diff
@@ -0,0 +1,410 @@
...
+		jobConf.set(OUTDIR, location);
+		JobContext jobContext = new JobContextImpl(this.jobConf, new JobID());
+		OutputCommitter outputCommitter = this.jobConf.getOutputCommitter();
+		// finalize HDFS output format
+		outputCommitter.commitJob(jobContext);
+	}
+
+	// converts a Row to a list so that Hive can serialize it
+	private Object getConvertedRow(Row record) {
+		int actualArity = isPartitioned ? record.getArity() - partitionCols.size() : record.getArity();
```
bowenli86: is the arity the same in the lifecycle of a `HiveTableOutputFormat` instance or can it vary? If the same, shall we make it a member variable and not compute for every record?

lirui-apache: Makes sense. PR updated.

bowenli86:
LGTM, but why do we need the commit "fix HiveTypeUtil a50ac6e"?

lirui-apache:
> LGTM, but why do we need the commit "fix HiveTypeUtil [a50ac6e](https://github.com/apache/flink/commit/a50ac6e12af3c635160fd4ee9ea22a30f8388cef)"?

Because those type info instances in `BasicTypeInfo` are not singletons. So it's not safe to check equality with `==`. This caused some problem when I developed the test case (although that test is not the final version in this PR).

lirui-apache:
Rebased. @bowenli86 please take another look. Thanks.

bowenli86:
> > LGTM, but why do we need the commit "fix HiveTypeUtil [a50ac6e](https://github.com/apache/flink/commit/a50ac6e12af3c635160fd4ee9ea22a30f8388cef)"?
> 
> Because those type info instances in `BasicTypeInfo` are not singletons. So it's not safe to check equality with `==`. This caused some problem when I developed the test case (although that test is not the final version in this PR).

Not sure about new data types, but these old ones are static final vars that are supposed to be referenced/used when it comes to data types. Here's an example in [Orc source table ](https://github.com/apache/flink/blob/5e90ed95a580aefd84b72f593954d01f4eb67f68/flink-connectors/flink-orc/src/main/java/org/apache/flink/orc/OrcTableSource.java#L397)

Can you describe the test problem? If we only use these predefined vars and not creating new ones, would it be fine? If it turns out we really need to do it, how about make changes in the later PR which adds the tests?

JingsongLi:
I think the problem maybe is:
Although they are static final vars, it will be a new instance after serialize/deserialize. (The real singleton need in: https://stackoverflow.com/questions/21361137/deserializing-a-singleton-in-java )
New data types are the same.

lirui-apache:
> > > LGTM, but why do we need the commit "fix HiveTypeUtil [a50ac6e](https://github.com/apache/flink/commit/a50ac6e12af3c635160fd4ee9ea22a30f8388cef)"?
> > 
> > 
> > Because those type info instances in `BasicTypeInfo` are not singletons. So it's not safe to check equality with `==`. This caused some problem when I developed the test case (although that test is not the final version in this PR).
> 
> Not sure about new data types, but these old ones are static final vars that are supposed to be referenced/used when it comes to data types. Here's an example in [Orc source table ](https://github.com/apache/flink/blob/5e90ed95a580aefd84b72f593954d01f4eb67f68/flink-connectors/flink-orc/src/main/java/org/apache/flink/orc/OrcTableSource.java#L397)
> 
> Can you describe the test problem? If we only use these predefined vars and not creating new ones, would it be fine? If it turns out we really need to do it, how about make changes in the later PR which adds the tests?

The problem I hit is that we got different instances of the same type info. I don't think we have the guarantee that all type info instances used are from `BasicTypeInfo`. E.g. HiveTableOutputFormat has a `RowTypeInfo` field that needs to be serialized. And after deserialization, the type info instances in this `RowTypeInfo` are different from those in `BasicTypeInfo`.

lirui-apache:
Thanks @JingsongLi , you beat me to the explanation :)
BTW, I'd like to confirm whether type info instances are supposed to be singletons? Judging from the code, it's not designed that way. So I think it's always safer to call `equals` to check equality (guess that's also why BasicTypeInfo overrides the `equals` method). Is that correct?

JingsongLi:
> Thanks @JingsongLi , you beat me to the explanation :)
> BTW, I'd like to confirm whether type info instances are supposed to be singletons? Judging from the code, it's not designed that way. So I think it's always safer to call `equals` to check equality (guess that's also why BasicTypeInfo overrides the `equals` method). Is that correct?

Yes, it's always safer to call `equals` to check equality. (Scala's "==" is equals)

lirui-apache:
Thanks @JingsongLi 
@bowenli86 do you think we can keep that change in this PR? I'm also fine to have a separate ticket to take care of it later.

bowenli86:
thanks @lirui-apache @JingsongLi for the explanation. For simplicity, let's keep it in the PR

bowenli86:
I'm re-running the failed tests in travis CI and will merge once build passes

bowenli86:
moreover, please have correct commit message in the future. E.g. the first commit's msg should be "[FLINK-12568][hive] Implement OutputFormat to write Hive tables"

lirui-apache:
> moreover, please have correct commit message in the future. E.g. the first commit's msg should be "[FLINK-12568][hive] Implement OutputFormat to write Hive tables"

Got it. Thanks.

PR #8117, opened 2019-04-05
[FLINK-12115] [filesystems]: Add support for AzureFS

piyushnarang:
## What is the purpose of the change

Add support for the Azure blob store as one of the supported flink-filesystems. 

## Brief change log

* Add Azure FS factory and file system class. 

## Verifying this change

This change added tests and can be verified as follows:

  - *Added unit tests*
  - *Added file system integration tests for HTTP / HTTPs schemes*

## Does this pull request potentially affect one of the following parts:

  - Dependencies (does it add or upgrade a dependency): yes (hadoop-azure)
  - The public API, i.e., is any changed class annotated with `@Public(Evolving)`: no
  - The serializers: no
  - The runtime per-record code paths (performance sensitive): no
  - Anything that affects deployment or recovery: JobManager (and its components), Checkpointing, Yarn/Mesos, ZooKeeper: no
  - The S3 file system connector: no

## Documentation

  - Does this pull request introduce a new feature? yes
  - If yes, how is the feature documented? Javadocs

PR #8537, opened 2019-05-24
[FLINK-12115][fs] Add support for AzureFS

tillrohrmann:
This PR is based on #8117. It decreases the dependency footprint and adds a proper NOTICE file to it.

cc @piyushnarang

### DIFF
--- a/flink-filesystems/flink-azure-fs-hadoop/pom.xml
+++ b/flink-filesystems/flink-azure-fs-hadoop/pom.xml
@@ -34,7 +34,6 @@ under the License.
 
 	<!-- need to use a release which includes this patch: https://github.com/apache/hadoop/commit/02cadbd24bf69925078d044701741e2e3fcb4b2f -->
 	<properties>
-		<fs.azure.version>2.7.0</fs.azure.version>
 		<fs.azure.sdk.version>1.16.0</fs.azure.sdk.version>
 		<fs.jackson.core.version>2.9.4</fs.jackson.core.version>
 	</properties>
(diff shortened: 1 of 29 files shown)


tillrohrmann:
@shuai-xu do you have time to test the Azure FS implementation?

piyushnarang:
Thanks for spinning this up @tillrohrmann, will review and chime in.

flink-filesystems/flink-azure-fs-hadoop/pom.xml
```diff
@@ -119,17 +111,69 @@ under the License.
...
 								<relocation>
-									<pattern>com.microsoft.azure.storage</pattern>
-									<shadedPattern>org.apache.flink.fs.shaded.com.microsoft.azure.storage</shadedPattern>
+									<pattern>com.microsoft.azure</pattern>
+									<shadedPattern>org.apache.flink.fs.azure.shaded.com.microsoft.azure</shadedPattern>
+								</relocation>
+
+								<!-- shade dependencies internally used by Azure and never exposed downstream -->
+								<relocation>
+									<pattern>org.apache.httpcomponents</pattern>
```
zentol: do we still need all this now that we have plugins?

tillrohrmann: If the plugins are used then not. If people drop the jar into `lib` it would still be beneficial. Therefore, I opted for this approach.

shuai-xu:
@tillrohrmann I have test the implementation. It works well, all libs are included. The only problem is that the file words is not uploaded to azure automatically. so the ones who run E2E tests don't know the file content and the checksum check will fail.

tillrohrmann:
Thanks for testing the Azure file system @shuai-xu. This is a valid point you are bringing up here. I think it would be best if we added a Flink Azure account similar to what we did with S3 and add a static `words` file. This would, however, still mean that a user would need to upload it manually when testing it with her own Azure account. Since this is also the problem with the S3 tests, I would address this problem as a follow up.

tillrohrmann:
Follow up for enabling the Azure FS E2E test on Travis: https://issues.apache.org/jira/browse/FLINK-12654

Merging this PR now.

piyushnarang:
Thanks @tillrohrmann :-)
ps. the words file is identical to that in the s3 tests and is the one we have in the repo.

PR #7470, opened 2019-01-11
[FLINK-11283] Accessing the key when processing connected keyed stream

yanghua:
## What is the purpose of the change

*This pull request introduces a `CoKeyedProcessFunction` for keyed connected stream*

## Brief change log

  - *Introduced a `CoKeyedProcessFunction` for keyed connected stream*
  - *Add test case and changed documentation*

## Verifying this change

This change is already covered by existing tests, such as *PureKeyedCoProcessOperatorTest and SideOutputITCase*.

## Does this pull request potentially affect one of the following parts:

  - Dependencies (does it add or upgrade a dependency): (yes / **no**)
  - The public API, i.e., is any changed class annotated with `@Public(Evolving)`: (**yes** / no)
  - The serializers: (yes / **no** / don't know)
  - The runtime per-record code paths (performance sensitive): (yes / **no** / don't know)
  - Anything that affects deployment or recovery: JobManager (and its components), Checkpointing, Yarn/Mesos, ZooKeeper: (yes / **no** / don't know)
  - The S3 file system connector: (yes / **no** / don't know)

## Documentation

  - Does this pull request introduce a new feature? (yes / **no**)
  - If yes, how is the feature documented? (not applicable / docs / JavaDocs / **not documented**)

PR #8538, opened 2019-05-24
[FLINK-11283] Accessing the key when processing connected keyed stream

aljoscha:
Cleaned up and rebased version of #7470, for running Travis tests.


yanghua:
There is a failure when downloading kafka_2.11-2.2.0.

aljoscha:
Merged

PR #8543, opened 2019-05-26
[FLINK-12101] Race condition when concurrently running uploaded jars via REST

leesf:

Fix race condition when concurrently running uploaded jars via REST.

## Brief change log

* Introduce a _ThreadLocal_ field to store _ExecutionEnvironmentFactory_ 
* When _contextEnvironmentFactory_ is null, use _ThreadLocal_ to get Factory.

## Verifying this change

  - The serializers: (no)

## Documentation

cc @tillrohrmann

### DIFF
--- a/flink-java/src/main/java/org/apache/flink/api/java/ExecutionEnvironment.java
+++ b/flink-java/src/main/java/org/apache/flink/api/java/ExecutionEnvironment.java
@@ -102,10 +102,10 @@ public abstract class ExecutionEnvironment {
 	protected static final Logger LOG = LoggerFactory.getLogger(ExecutionEnvironment.class);
 
 	/** The environment of the context (local by default, cluster if invoked through command line). */
-	private static ExecutionEnvironmentFactory contextEnvironmentFactory;
+	private static ExecutionEnvironmentFactory contextEnvironmentFactory = null;
 
 	/** The ThreadLocal used to store {@link ExecutionEnvironmentFactory}. */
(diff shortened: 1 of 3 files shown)


flink-java/src/main/java/org/apache/flink/api/java/ExecutionEnvironment.java
```diff
@@ -1061,8 +1064,12 @@ private static String getDefaultName() {
 	 * @return The execution environment of the context in which the program is executed.
 	 */
 	public static ExecutionEnvironment getExecutionEnvironment() {
+
 		return contextEnvironmentFactory == null ?
-				createLocalEnvironment() : contextEnvironmentFactory.createExecutionEnvironment();
+				(contextEnvironmentFactoryThreadLocal.get() == null ?
```
klion26: Do we need to add `contextEnvironmentFactoryThreadLocal` check in the function `areExplicitEnvironmentsAllowed`(line 1284). I can't add a review there.

leesf: Thanks for your review @klion26. Good catch and updated.

tillrohrmann: I would suggest to first check the thread local variable and then fall back to the static variable. That way, it would not be possible to override the value of an `ExecutionEnvironment` which has been set explicitly by from another thread.

leesf:
@tillrohrmann Thanks for your review. Updated this PR to address your comments.

PR #8545, opened 2019-05-26
[FLINK-12520] Support to provide fully-qualified domain host name in TaskManagerMetricGroup

Myasuka:

Inspired from [Chinese user-mail](https://lists.apache.org/thread.html/e1774a42430815b689ea792103d002b1da734d6086682d34c044ef35@%3Cuser-zh.flink.apache.org%3E) which complains that host name in metrics name could only show the first part. However, their full host name is like "ambari.host12.yy" which means the first part "ambari" cannot identify anything.

With this PR, we could support to let user record their full host name in the `TaskManagerMetricGroup` so that to identify metrics from different hosts.

## Brief change log

  - Add new option `metrics.tm.full-hostname` in `MetricOptions` to indicate whether Flink should use fully qualified host name in task manager metrics.
  - Use above metric option when `instantiateTaskManagerMetricGroup` in `TaskManagerRunner`

## Verifying this change

  - Added test that `TaskManagerLocation#getHostName(InetAddress, boolean)` could return the host name as expected.

  - The serializers: **no**

## Documentation

  - If yes, how is the feature documented? **docs**

### DIFF
--- a/flink-core/src/main/java/org/apache/flink/configuration/MetricOptions.java
+++ b/flink-core/src/main/java/org/apache/flink/configuration/MetricOptions.java
@@ -185,6 +185,14 @@ public class MetricOptions {
 				"faster updating metrics. Increase this value if the metric fetcher causes too much load. Setting this value to 0 " +
 				"disables the metric fetching completely.");
 
+	/**
+	 * Whether the host name in task manager metrics should be fully qualified domain name.
+	 */
+	public static final ConfigOption<Boolean> METRIC_FULL_HOST_NAME =
+		key("metrics.tm.full-hostname")
+			.defaultValue(false)
+			.withDescription("Flag indicating whether Flink should use fully qualified host name in task manager metrics.");
+
 	private MetricOptions() {
 	}
 }
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/TaskManagerLocation.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/TaskManagerLocation.java
@@ -191,6 +191,18 @@ public class TaskManagerLocation implements Comparable<TaskManagerLocation>, jav
 	 * @return hostname of the TaskManager
 	 */
 	public static String getHostName(InetAddress inetAddress) {
+		return getHostName(inetAddress, false);
+	}
+
+	/**
+	 * Gets the hostname of the TaskManager based on the network address. If {@code useFullHostName} is set as true,
+	 * it would return the fully qualified domain name if possible otherwise not.
+	 *
+	 * @param inetAddress the network address that the TaskManager binds its sockets to
+	 * @param useFullHostName whether to return the fully qualified domain name or take the first part.
+	 * @return hostname of the TaskManager
+	 */
+	public static String getHostName(InetAddress inetAddress, boolean useFullHostName) {
 		String hostName;
 		String fqdnHostName = getFqdnHostName(inetAddress);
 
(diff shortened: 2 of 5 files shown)


flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/TaskManagerLocation.java
```diff
@@ -202,7 +214,11 @@ public static String getHostName(InetAddress inetAddress) {
 			LOG.warn("No hostname could be resolved for the IP address {}, using IP address as host name. "
 				+ "Local input split assignment (such as for HDFS files) may be impacted.", inetAddress.getHostAddress());
 		} else {
-			hostName = NetUtils.getHostnameFromFQDN(fqdnHostName);
+			if (useFullHostName) {
+				hostName = fqdnHostName;
```
zentol: this case should most certainly not be handled in this method. Have the TaskManagerRunner call `getFqdnHostName` instead (after making it public).

flink-core/src/main/java/org/apache/flink/configuration/MetricOptions.java
```diff
@@ -185,6 +185,14 @@
 				"faster updating metrics. Increase this value if the metric fetcher causes too much load. Setting this value to 0 " +
 				"disables the metric fetching completely.");
 
+	/**
+	 * Whether the host name in task manager metrics should be fully qualified domain name.
+	 */
+	public static final ConfigOption<Boolean> METRIC_FULL_HOST_NAME =
+		key("metrics.tm.full-hostname")
```
zentol: I'd rather introduce a new variable `>tm_fqdn<` then resorting to config options.

Myasuka: @zentol Previously, I thought `<tm_fqdn>` was a good idea. However, after I looked into the code deeper and thought about the actual problem we want to fix. I think the new variable should be `<host_fqdn>`. 
There already existed a variable `<host>`, however its real meaning differs on JM and TM side. `<host>` actually means the fully qualified domain name on JM side while it means the first part of fqdn on TM side. We could introduce a new variable `<host_fqdn>` to align its meaning both on JM and TM side.
There still existed another problem if we introduce `<host_fqdn>`. If user happen to configure the `SCOPE_NAMING_TM` as `<host>.<host_fqdn>.taskmanager.<tm_id>` and since we would only pass one `hostName` parameter to `TaskManagerMetricGroup` under the condition that we do not break the API. We have two choices, one is to throw exception when initializing `ScopeFormat` with `<host>` and `<host_fqdn>` both existed. Another is introduing another configure option to control the actual `<host>` meaning which is my PR did. What do you think?

Myasuka:
BTW, @zentol I noticed that you have already self-assigned this PR. It was my fault to not update PR according to your comments in time. If you have not already spent some energy to refactor this PR in private, I think I could continue this work recently before FLINK-1.9 feature freeze. What do you think?

flink-core/src/main/java/org/apache/flink/configuration/MetricOptions.java
zentol: This should pretty easy to solve, just pass in both the host and host_fqdn into the TaskManagerMetricGroup.

I'm fine with calling it `<host_fqdn>`.

zentol:
I haven't done refactorings; I assign myself just to get a better overview of which PRs I (am/plan to) reviewing.

PR #8546, opened 2019-05-27
[FLINK-12621]. Use MiniCluster instead of JobExecutor

zjffdu:

This is a refactoring PR which remove `JobExecutor` and use `MiniCluster` instead. `JobExecutor` is specifically used for local mode, I don't think we need to introduce such kind of new class/interface for local mode, we could just use the existing `MiniCluster`.

## Brief change log

* Remove `JobExecutor` &  `JobExecutorService`
* Use `ClusterClient` to replace `JobExecutorService`

## Verifying this change

Current CI is passed, no new code is introduced, just some refactoring.

## Documentation


zjffdu:
@tillrohrmann  Could you help review it ? I notice you are original author of `JobExecutor`.

zjffdu:
@tillrohrmann  Thanks for the review. I think we should unify the flink's local client and remote client. This is related with https://github.com/apache/flink/pull/8533 which also try to unify them in scala shell. 
I don't think we should differentiate local client and remote client in api perspective. Otherwise it is very difficult for downstream project to integrate with flink. E.g. I have already hit issues when integrating flink into zeppelin notebook because of flink's api difference in local mode and remote mode. Does it make sense to you ?

tillrohrmann:
I understand the general idea @zjffdu. What I'm unsure about is where exactly the `JobExecutorService` poses a problem because it should only be used by the `LocalExecutor` internally. As far as I know, this interface should not be leaked anywhere.

I guess the best thing to do would be to create a design document which describes how Flink's client should look like in the future. Until then, we should close this PR.

TisonKun:
ref #9643 and [FLINK-13961](https://issues.apache.org/jira/browse/FLINK-13961).

kl0u:
Isn't also this the same as this https://github.com/apache/flink/pull/9643 discussion which seems to have been already resolved? This is the other duplicate jira https://issues.apache.org/jira/browse/FLINK-13961

zjffdu:
That's right @kl0u , I will close this PR

PR #8550, opened 2019-05-27
[FLINK-12401][table] Support incremental emit under AccRetract mode for non-window streaming FlatAggregate on Table API

hequn8128:

This pull request supports incremental emit under AccRetract mode for non-window streaming FlatAggregate on Table API

As described in [Flip-29](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=97552739), there are two output modes for non-window streaming flatAggregate. One is emitting with full values, the other is emitting with incremental values. [FLINK-10977](https://issues.apache.org/jira/browse/FLINK-10977) supports the former one, this jira is going to support the latter one under AccRetract mode. The reason that Acc mode has not been converted by this jira is that we can't define keys on TableAggregateFunction. Once we support define keys on the output of flatAggregate, we can support the AccMode. 

## Brief change log

There are four commits for this PR:
  - 3a39d63 adds the RetractableCollector. The collector collects a record and forwards it. The collector can output retract messages with the retract. The collector only use it in `emitRetractValueIncrementally`.
  - 03019de adjusts the codgen part. It takes the `emitRetractValueIncrementally` method into consideration and also generates the retract methods if it is a Retractable collector.
  - 611a02c This is the runtime function part. Minor changes in this part. The big difference is retractions will be generated by TableAggregateFunction manually if incremental emit value methods have been defined.
  - a9921d5 Adjusts documents about `emitRetractValueIncrementally`.

## Verifying this change

  - Add `testEmitRetractValueIncrementally()` ITCase to test the incremental emit.
  - Add `testTableAggregateEmitRetractValueIncrementally()` harness test to test the incremental emit.

  - The serializers: (no)

## Documentation


hequn8128:
@flinkbot attention @sunjincheng121 @twalthr

sunjincheng121:
@flinkbot approve description

sunjincheng121:
HI @hequn8128, Thanks for the PR.
Overall, I think I should think about the interface method again, due to in this PR there is some change(I think it's a good point) between the design doc of FLIP-29. Then review this PR.

sunjincheng121:
Since the execution mode of the Stream operator has two modes, `ACC` and `ACCRetract`, users can achieve better performance by implementing special interfaces for streaming. The table below is a quick summary.

  | emitValue | emitUpdateWithRetract | emitUpdateWithoutRetract
-- | -- | -- | --
ACC | Y | N | Y
ACCRetract | Y | Y | N

-emitValue - for batch and streaming.
-eimitUpdateWithRetract - only for streaming in ACC mode.
-emitUpdateWithoutRetract - only for streaming in ACCRetract mode(need key definition on TableAggregateFunction, [under discussion](https://docs.google.com/document/d/183qHo8PDG-xserDi_AbGP6YX9aPY0rVr80p3O3Gyz6U/edit#heading=h.evvcpnbn30wn)).

So, In this PR, change the method name from `emitRetractValueIncrementally` to `emitUpdateWithRetract` is better. What do you think?

hequn8128:
@sunjincheng121 Thanks for the great summary. I think it good to change the name to `emitUpdateWithRetract`. 

The table gives a clear picture of these methods. The `emitValue` can both be used for batch and streaming while the other two methods are used to enhance streaming cases. When we finish this PR, we can continue with the `emitUpdateWithoutRetract` once the discussion about key definition reaches a conclusion.

Best, Hequn

hequn8128:
@sunjincheng121 I have updated the PR according to your suggestions with a new commit at the top of the old ones. Thank you very much for your review.

Best, Hequn

hequn8128:
Improve examples in the document and squash some commits.

docs/dev/table/tableApi.md
```diff
@@ -2882,13 +2922,13 @@ val result = orders
       <td>
         <p>Groups and aggregates a table on a <a href="#group-windows">group window</a> and possibly one or more grouping keys. You have to close the "flatAggregate" with a select statement. And the select statement does not support aggregate functions.</p>
 {% highlight scala %}
-val tableAggFunc = new MyMinMax
+val top2 = new Top2()
```
sunjincheng121: `Top2()` -> `Top2` is enough for scala. But I am fine if want keeps the current style.

sunjincheng121:
@flinkbot approve all

docs/dev/table/tableApi.md
hequn8128: Make sense. I will remove the brackets to make it more scala style.

PR #8556, opened 2019-05-28
[FLINK-12171][Network] Do not limit the network buffer memory by heap size on the TM side

gaoyunhaii:

This pull request fixes the bug that limits the network buffer size with the heap size on the TM side. In fact, network buffer occupies a part of direct memory and is independent with the heap.

To fix this problem, The limitation on the TM side is removed. Although we may want to compare the network memory size with the total memory size on the TM side, currently we can only compute the total memory with heap + computed network memory and the computed total memory should be always larger than the computed network memory.

To remove the limitation, the max allowed memory used to check the network memory size on TM side is changed to Long.MAX_VALUE. Another option is to move the checking to the caller function on the RM side. however, it is not easy to achieve since the checking relies on the configured values of MIN and MAX, and it is not accessible outside of the current function.

## Brief change log
  - f32dde51c1183d22e8a192e19d037a4f0c48c9ae extracts common variables to simplify the tests.
  - 17f93e19b39433dc5d93c24c5b8a85feb1c37b1f fixes the exception when network buffer fraction is larger than 0.5 on TM side, and adds new tests to cover the cases that the network buffer fraction is larger than 0.5
 - 2762ce8f219b8b988d9824a8beab5c6117119403 optimizes the computation of the network memory on TM side to fix the numerical precision problem.
 - aa5c15913e411babe2bf5cc3067cdbb112d57fd0 unifies the terminology in `NettyShuffleEnvironmentConfiguration`

## Verifying this change
  - *Add tests for network buffer computation with `fraction > 0.5`*
  - *Manually verified the change by running a cluster with two task managers for both standalone and YARN mode, and test the configuration with heap = 3G/network = 2G and heap = 5G/network = 2G*.

  - The serializers: no

## Documentation

### DIFF
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/NettyShuffleEnvironmentConfiguration.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/NettyShuffleEnvironmentConfiguration.java
@@ -260,7 +260,12 @@ public class NettyShuffleEnvironmentConfiguration {
 		//              = jvmTotal - Math.min(networkBufMax, Math.max(networkBufMin, jvmHeap * netFraction)
 		// jvmTotal = jvmHeapNoNet / (1.0 - networkBufFraction)
 		float networkBufFraction = config.getFloat(NettyShuffleEnvironmentOptions.NETWORK_BUFFERS_MEMORY_FRACTION);
-		long networkBufSize = (long) (jvmHeapNoNet / (1.0 - networkBufFraction) * networkBufFraction);
+
+		// Converts to double for higher precision. Converting via string achieve higher precision for those
+		// numbers can not be represented preciously by float, like 0.4f.
+		double heapNoNetFraction = 1.0 - Double.valueOf(Float.toString(networkBufFraction));
+		long totalJavaMemorySize = (long) (jvmHeapNoNet / heapNoNetFraction);
+		long networkBufSize = (long) (totalJavaMemorySize * networkBufFraction);
 
 		// Do not need to check the maximum allowed memory since the computed total memory should always
 		// be larger than the computed network buffer memory as long as the fraction is less than 1.
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/NettyShuffleEnvironmentConfiguration.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/NettyShuffleEnvironmentConfiguration.java
@@ -236,40 +236,40 @@ public class NettyShuffleEnvironmentConfiguration {
 	public static long calculateNewNetworkBufferMemory(Configuration config, long maxJvmHeapMemory) {
 		// The maximum heap memory has been adjusted as in TaskManagerServices#calculateHeapSizeMB
 		// and we need to invert these calculations.
-		final long jvmHeapNoNet;
+		final long heapAndManagedMemory;
 		final MemoryType memoryType = ConfigurationParserUtils.getMemoryType(config);
 		if (memoryType == MemoryType.HEAP) {
-			jvmHeapNoNet = maxJvmHeapMemory;
+			heapAndManagedMemory = maxJvmHeapMemory;
 		} else if (memoryType == MemoryType.OFF_HEAP) {
 			long configuredMemory = ConfigurationParserUtils.getManagedMemorySize(config) << 20; // megabytes to bytes
 			if (configuredMemory > 0) {
 				// The maximum heap memory has been adjusted according to configuredMemory, i.e.
 				// maxJvmHeap = jvmHeapNoNet - configuredMemory
-				jvmHeapNoNet = maxJvmHeapMemory + configuredMemory;
+				heapAndManagedMemory = maxJvmHeapMemory + configuredMemory;
 			} else {
 				// The maximum heap memory has been adjusted according to the fraction, i.e.
 				// maxJvmHeap = jvmHeapNoNet - jvmHeapNoNet * managedFraction = jvmHeapNoNet * (1 - managedFraction)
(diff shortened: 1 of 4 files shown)


flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/NettyShuffleEnvironmentConfiguration.java
```diff
@@ -229,12 +229,15 @@ public static long calculateNewNetworkBufferMemory(Configuration config, long ma
...
+		// jvmHeapNoNet = jvmTotal - networkBufBytes
+		//              = jvmTotal - Math.min(networkBufMax, Math.max(networkBufMin, jvmHeap * netFraction)
+		// jvmTotal = jvmHeapNoNet / (1.0 - networkBufFraction)
 		float networkBufFraction = config.getFloat(NettyShuffleEnvironmentOptions.NETWORK_BUFFERS_MEMORY_FRACTION);
 		long networkBufSize = (long) (jvmHeapNoNet / (1.0 - networkBufFraction) * networkBufFraction);
-		return calculateNewNetworkBufferMemory(config, networkBufSize, maxJvmHeapMemory);
+
+		// Do not need to check the maximum allowed memory since the computed total memory should always
+		// be larger than the computed network buffer memory as long as the fraction is less than 1.
+		return calculateNewNetworkBufferMemory(config, networkBufSize, Long.MAX_VALUE);
```
NicoK: The fact that you can change this without a single test failing indicates no test coverage for it. Please add some tests.

gaoyunhaii:
Very thanks @NicoK for the review and sorry for missing the tests. I will address the comments and complement the tests right now.

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/NettyShuffleEnvironmentConfiguration.java
```diff
@@ -229,12 +229,15 @@ public static long calculateNewNetworkBufferMemory(Configuration config, long ma
 		}
 
 		// finally extract the network buffer memory size again from:
-		// jvmHeapNoNet = jvmHeap - networkBufBytes
-		//              = jvmHeap - Math.min(networkBufMax, Math.max(networkBufMin, jvmHeap * netFraction)
-		// jvmHeap = jvmHeapNoNet / (1.0 - networkBufFraction)
+		// jvmHeapNoNet = jvmTotal - networkBufBytes
+		//              = jvmTotal - Math.min(networkBufMax, Math.max(networkBufMin, jvmHeap * netFraction)
+		// jvmTotal = jvmHeapNoNet / (1.0 - networkBufFraction)
```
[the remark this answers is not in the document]
gaoyunhaii: Totally agree with that unified terminology should be more clear. I have modified the terminology used in `NettyShuffleEnvironmentConfiguration`.

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/NettyShuffleEnvironmentConfiguration.java
gaoyunhaii: Very sorry for missing the tests, I have complemented the tests according to the comments.

gaoyunhaii:
Hi @NicoK , very thanks for the review ! I have updated the PR to address the issues.

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/NettyShuffleEnvironmentConfiguration.java
```diff
@@ -224,32 +224,40 @@ public static NettyShuffleEnvironmentConfiguration fromConfiguration(
...
 		} else {
 			throw new RuntimeException("No supported memory type detected.");
 		}
 
 		// finally extract the network buffer memory size again from:
-		// jvmHeapNoNet = jvmHeap - networkBufBytes
-		//              = jvmHeap - Math.min(networkBufMax, Math.max(networkBufMin, jvmHeap * netFraction)
-		// jvmHeap = jvmHeapNoNet / (1.0 - networkBufFraction)
+		// heapAndManagedMemory = totalProcessMemory - networkReservedMemory
+		//              = totalProcessMemory - Math.min(networkBufMax, Math.max(networkBufMin, totalProcessMemory * networkBufFraction)
```
NicoK: nit: please fix indentation in the second line here

gaoyunhaii: Very sorry for break the indentation. I have modified this issue now.

gaoyunhaii:
Very thanks @NicoK review the codes! I have merged the two related commits, and fixed the indentation problem.

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/NettyShuffleEnvironmentConfiguration.java
NicoK: @gaoyunhaii can you merge the "fix the indent problem" into "rename the terminology" because it was introduced there and we can fix it before merging - I'll merge afterwards

gaoyunhaii: OK, I initially hesitated on whether to merge the fix and I think separate it might be easier for  reviewing. I have merged the two commits now. :)

NicoK: that's true - for reviewing it is better if you don't merge / force-push - now, this PR is ready though and we need to clean it up a bit. Thanks for that.

PR #8557, opened 2019-05-28
[FLINK-12544][network] Fix the deadlock while releasing memory triggered by multiple threads in SpillableSubpartition

zhijiangW:

*Fix the deadlock while releasing memory triggered by multiple threads in `SpillableSubpartition`.*

## Brief change log

  - *Make the call of `releaseMemory(1)` outside `synchronized` in `LocalBufferPool#requestMemorySegment`*
  - *Refactor the `NoOpTaskActions` as a separate class for using in other tests*

## Verifying this change

This change is covered by new test `SpillableSubpartitionTest#testConcurrentRequestAndReleaseMemory`*.

## Documentation

### DIFF
--- a/flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/SpillableSubpartitionTest.java
+++ b/flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/SpillableSubpartitionTest.java
@@ -18,6 +18,7 @@
 
 package org.apache.flink.runtime.io.network.partition;
 
+import org.apache.flink.api.common.JobID;
 import org.apache.flink.runtime.io.disk.iomanager.AsynchronousBufferFileWriter;
 import org.apache.flink.runtime.io.disk.iomanager.BufferFileWriter;
 import org.apache.flink.runtime.io.disk.iomanager.FileIOChannel;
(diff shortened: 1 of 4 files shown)


zhijiangW:
@pnowojski , I thought of two ways for solving the deadlock issue:

- One is making `owner.get().releaseMemory(1)` outside of `synchronized` in `LocalBufferPool#requestMemorySegment`. 

- Another is making the logic of `bufferConsumer.close()` during `SpillableSubpartition#spillFinishedBufferConsumers` outside of `synchronized`.

I take the first option in this PR and it seems more easier for fixing.

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/SpillableSubpartitionTest.java
```diff
@@ -161,6 +173,72 @@ public Void call() throws Exception {
 		blockingFinish.get();
 	}
 
+	/**
+	 * Tests a fix for FLINK-12544.
+	 *
+	 * @see <a href="https://issues.apache.org/jira/browse/FLINK-12544">FLINK-12544</a>
+	 */
+	@Test
+	public void testConcurrentRequestAndReleaseMemory() throws Exception {
```
pnowojski: Is this test failing (at least from time to time) before your change?

zhijiangW: Yes, it could be failing every time after reverting my changes, because I control the logic sequence strictly.

zhijiangW:
Thanks for the reviews and good suggestions @pnowojski !
Actually I also had the same performance concern as you when refactor the `LocalBufferPool`. The final decision is based on below considerations:

- It is buffer-level overhead which might be ignored.
- The second `synchronized` only happens during backpressure as you mentioned, so it seems no matter in this case.
-  If we adjust the `SpillableSubpartition`, it seems more safe because it is rarely used and only for batch mode. But the internal logics in `SpillableSubpartition` seem a bit messy, and we should be very careful in order not to bring other issues. In contrast, the logic in `LocalBufferPool` seems more clearly to go. And it is very obvious the owner release should not be inside `synchronized`.

In addition, I think the process of `owner.get().releaseMemory(1)` could be improved a bit. The `releaseMemory` could return a int value for indicating how many buffers are released. If the return value is more than 0, we could try get available buffer directly after that, no need to enter `wait(2000)` for blocking mode or return null directly. But it is a different topic not related to the deadlock issue. And in the latest master code, the `releaseMemory` is invalid because it would always return 0 for the new `BoundedBlockingSubpartition`. So I think it does not worth efforts for improving it.

If you still want to try another way of adjusting the `SpillableSubpartition`, I could further think and have a try.

PR #8563, opened 2019-05-29
[FLINK-12602][travis] Correct the flink pom `artifactId` config and s…

sunjincheng121:
I find a shell issue in `verify_scala_suffixes.sh`(line 145) as follows:

`grep "${module}_\d\+\.\d\+</artifactId>" "{}"`

This code want to find out all modules that the module's `artifactId`  with a `scala_binary_version` suffix. 
but the problem is our all `artifactId` value is in the pattern of `XXX_${scala.binary.version}`, such as:

`<artifactId>flink-tests_${scala.binary.version}</artifactId>`

then the result always empty, so this check did not take effect.

So, we need to correct `artifactId `of some of the modules and correct the check logic for scala-free.

## Brief change log
  - remove the scala version suffix for connector-hive and queryable-state-client-java
  - add the scala dependencies for table-api-scala and flink-sql-connectors
  - correct the scala-free check logic in `verify_scala_suffixes.sh`

NOTE:
We have two ways handling of the connector:
 1. Improve the script to check any (compile) dependencies with a scala-suffix(which we mentioned above).
 2. Add the `flink-streaming-java` dependency wich  `provided`  scope for the corresponding connectors.

For approach 1, we should add check logic:
 1. The command of `dependency:tree` should add an option: ` -Dincludes=org.apache.flink:_2.1::  ` such as `org.apache.flink:flink-streaming-java_2.11`.
  2. The command of `grep ` also need to add the logic: `E "org.scala-lang| org.apache.flink:[^:]+_2\.1[0-9]"` , also for test `grep --invert-match "org.apache.flink:[^:]_2\.1[0-9]:.:.*:test"`.

For approach 2, we should add the dependency of `link-streaming-java` for `flink-sql-connector-elasticsearch6 flink-sql-connector-kafka flink-sql-connector-kafka-0.10 flink-sql-connector-kafka-0.11 flink-sql-connector-kafka-0.9`.  

For now, I think the change logic in approach 1 is a bit complex(a lot of filtering processing logic). So, I prefer the approach 2, due to even we add some new module in the future we should well know whether we should add the scala-suffix for the `artifactId`, then manually add dependencies on for the scala in the pom. 

So, in this PR I add the dependency of `link-streaming-java` for `flink-sql-connectors`.

## Verifying this change

  - The serializers: (no)

## Documentation

    1. correct the artifactId in `queryable_state.md/queryable_state.zh.md`.

### DIFF
--- a/flink-connectors/flink-sql-connector-kafka-0.10/pom.xml
+++ b/flink-connectors/flink-sql-connector-kafka-0.10/pom.xml
@@ -41,6 +41,14 @@ under the License.
 			<artifactId>flink-connector-kafka-0.10_${scala.binary.version}</artifactId>
 			<version>${project.version}</version>
 		</dependency>
+		<!--  Due to the scope of `provided` transitive dependencies problem,
+		we should copy this dependency here for the scala suffix check.	-->
+		<dependency>
+			<groupId>org.apache.flink</groupId>
+			<artifactId>flink-streaming-java_${scala.binary.version}</artifactId>
+			<version>${project.version}</version>
+			<scope>provided</scope>
+		</dependency>
 	</dependencies>
 
 	<build>
--- a/flink-connectors/flink-sql-connector-kafka-0.10/pom.xml
+++ b/flink-connectors/flink-sql-connector-kafka-0.10/pom.xml
@@ -41,14 +41,6 @@ under the License.
 			<artifactId>flink-connector-kafka-0.10_${scala.binary.version}</artifactId>
 			<version>${project.version}</version>
 		</dependency>
-		<!--  Due to the scope of `provided` transitive dependencies problem,
-		we should copy this dependency here for the scala suffix check.	-->
-		<dependency>
-			<groupId>org.apache.flink</groupId>
-			<artifactId>flink-streaming-java_${scala.binary.version}</artifactId>
-			<version>${project.version}</version>
-			<scope>provided</scope>
-		</dependency>
 	</dependencies>
 
 	<build>
--- a/docs/dev/stream/state/queryable_state.md
+++ b/docs/dev/stream/state/queryable_state.md
@@ -174,7 +174,7 @@ jar which must be explicitly included as a dependency in the `pom.xml` of your p
 </dependency>
 <dependency>
   <groupId>org.apache.flink</groupId>
-  <artifactId>flink-queryable-state-client-java{{ site.scala_version_suffix }}</artifactId>
+  <artifactId>flink-queryable-state-client-java</artifactId>
   <version>{{ site.version }}</version>
 </dependency>
 {% endhighlight %}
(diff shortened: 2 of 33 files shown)


sunjincheng121:
@flinkbot attention @zentol

flink-connectors/flink-sql-connector-kafka-0.10/pom.xml
```diff
@@ -41,6 +41,14 @@ under the License.
 			<artifactId>flink-connector-kafka-0.10_${scala.binary.version}</artifactId>
 			<version>${project.version}</version>
 		</dependency>
+		<!--  Due to the scope of `provided` transitive dependencies problem,
```
zentol: this looks hacky; why can't we search the pom for dependencies with a scala-suffix instead? Shouldn't this be as simple as adding `*:*_2.1*:*` as an artifact pattern to `-Dincludes`?

WeiZhong94: I think current PR is a more conservative approach. Do you mean that if a flink module depends on another flink module with a scala suffix, it is also considered dependent on scala? I considered this rule, but I am not sure if it was correct. Such as a module which depends on a scala-suffix module with "provided" scope, and only uses the interfaces provided by the scala-suffix module itself, regardless of the actual scala version of the runtime. May such a module exist?

zentol: > Do you mean that if a flink module depends on another flink module with a scala suffix, it is also considered dependent on scala?

Yes. It doesn't matter whether it is provided or not.
The only exception are dependencies with a `test` scope.

sunjincheng121: Thanks @WeiZhong94 @zentol, for @zentol's suggestion also make sense to me, just like we discuss offline @WeiZhong94 . and I mentioned above, we can take `approach 1` for this PR. 
I'll update the PR soon.

sunjincheng121:
Thanks for the review @zentol @WeiZhong94  I have rebase the code and update the PR!
I appreciate if you can have a look at it again!
Best,Jincheng

zentol:
huh.

I just found this dependency in `flink-tests`:
{code}
<dependency>
	<groupId>com.typesafe.akka</groupId>
	<artifactId>akka-testkit_${scala.binary.version}</artifactId>
</dependency>
{code}

This means this module does have a scala dependency, and should have a suffix. I'm wondering why the script didn't pick that up.

sunjincheng121:
rebase code(for python CI failed)

sunjincheng121:
@zentol I have added the check of Scala suffix for all groupId, I appreciate if you have time to have look at again :)!

sunjincheng121:
Merging...

PR #8566, opened 2019-05-29
[FLINK-12673][network] Introduce NetworkEnvironment.getUnreleasedPartitions instead of using getResultPartitionManager

azagrebin:

At the moment NetworkEnvironment.getResultPartitionManager is used to query TaskExecutor whether all partitions are released. We should hide this direct access to ResultPartitionManager behind a Shuffle API method call: NetworkEnvironment.getUnreleasedPartitions.

## Brief change log

 - Refactor `ResultPartitionManager.areAllPartitionsReleased` to `getUnreleasedPartitions`
 - Add `NetworkEnvironment.getUnreleasedPartitions`

## Verifying this change

Trivial refactor covered by existing tests.

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NetworkEnvironment.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NetworkEnvironment.java
@@ -174,6 +174,7 @@ public class NetworkEnvironment {
 	//  Properties
 	// --------------------------------------------------------------------------------------------
 
+	@VisibleForTesting
 	public ResultPartitionManager getResultPartitionManager() {
 		return resultPartitionManager;
 	}
@@ -207,6 +208,16 @@ public class NetworkEnvironment {
 		for (ResultPartitionID partitionId : partitionIds) {
 			resultPartitionManager.releasePartition(partitionId, null);
 		}
+	}
+
+	/**
+	 * Report unreleased partitions.
+	 *
+	 * @return collection of partitions which still occupy some resources locally on this task executor
+	 * and have been not released yet.
+	 */
+	public Collection<ResultPartitionID> getUnreleasedPartitions() {
+		return resultPartitionManager.getUnreleasedPartitions();
 	}
 
 	// --------------------------------------------------------------------------------------------
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/ResultPartitionManager.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/ResultPartitionManager.java
@@ -22,6 +22,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
+import java.util.Collection;
 import java.util.HashMap;
 import java.util.Map;
 
@@ -120,14 +121,9 @@ public class ResultPartitionManager implements ResultPartitionProvider {
 		}
 	}
 
-	public boolean areAllPartitionsReleased() {
+	public Collection<ResultPartitionID> getUnreleasedPartitions() {
 		synchronized (registeredPartitions) {
-			for (ResultPartition partition : registeredPartitions.values()) {
-				if (!partition.isReleased()) {
-					return false;
-				}
-			}
-			return true;
+			return registeredPartitions.keySet();
 		}
 	}
 }
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/ResultPartitionManager.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/ResultPartitionManager.java
@@ -106,19 +106,17 @@ public class ResultPartitionManager implements ResultPartitionProvider {
 	// ------------------------------------------------------------------------
 
 	void onConsumedPartition(ResultPartition partition) {
-		final ResultPartition previous;
-
 		LOG.debug("Received consume notification from {}.", partition);
 
 		synchronized (registeredPartitions) {
-			previous = registeredPartitions.remove(partition.getPartitionId());
-		}
-
-		// Release the partition if it was successfully removed
-		if (partition == previous) {
-			partition.release();
-
-			LOG.debug("Released {}.", partition);
+			final ResultPartition previous = registeredPartitions.remove(partition.getPartitionId());
+			// Release the partition if it was successfully removed
+			if (partition == previous) {
+				partition.release();
+				ResultPartitionID partitionId = partition.getPartitionId();
+				LOG.debug("Released partition {} produced by {}.",
+					partitionId.getPartitionId(), partitionId.getProducerId());
+			}
 		}
 	}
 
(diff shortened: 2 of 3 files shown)


azagrebin:
@flinkbot attention @zhijiangW @zentol

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/ResultPartitionManager.java
```diff
@@ -122,14 +123,9 @@ void onConsumedPartition(ResultPartition partition) {
...
-	public boolean areAllPartitionsReleased() {
+	public Collection<ResultPartitionID> getUnreleasedPartitions() {
 		synchronized (registeredPartitions) {
-			for (ResultPartition partition : registeredPartitions.values()) {
-				if (!partition.isReleased()) {
-					return false;
-				}
-			}
-			return true;
+			return registeredPartitions.keySet();
```
zhijiangW: The behavior here is different from before. We should confirm the criteria whether based on collection size or real partition state. It exists the scenario that the partition is already removed from the collection but the state is not released yet during`ResultPartitionManager#onConsumedPartition`. That means if we release the task executor immediately, it might lose chance for releasing the partition resource gracefully. WDYT?

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NetworkEnvironment.java
```diff
@@ -201,6 +203,18 @@ public void releasePartitions(Collection<ResultPartitionID> partitionIds) {
 		}
 	}
 
+	/**
+	 * Query unreleased partitions.
+	 *
+	 * @return collection of partitions which still occupy some resources locally on this task executor
+	 * and have not been released yet. The partition can be released either with {@link ResultPartitionWriter#fail(Throwable)}
+	 * or {@link ResultPartitionWriter#finish()} and then with {@link ResultPartitionWriter#close()} after fail or finish.
```
zhijiangW: It might bring some confusing here.
`ResultPartitionWriter` has two different states `FINISHED` and `RELEASED` separately. `ResultPartitionWriter#finish()` only indicates the whole data produced and enters `FINISHED` state, but the partition is not released yet. The release could only happen in three scenarios atm

- `ResultPartitionWriter#fail` when task is failed or canceled
- `NetworkEnvironment#releasePartitions` from JM RPC call
- `ResultPartitionManager#onConsumedPartition` when partition is consumed by consumer side

In addition, it might be better to put these comments above and make `@return` short.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/ResultPartitionManager.java
azagrebin: Ok, the behaviour did not actually change but this problem was already before :) If we remove partition in `ResultPartitionManager#onConsumedPartition` and concurrently continue to release it, the old `areAllPartitionsReleased` will still skip its state check because it is simply not  in the `registeredPartitions` anymore.

I think we have to fix it here by moving release under the lock as it is already done in `ResultPartitionManager.releasePartition`.

In general, I think `registeredPartitions` should reflect the release state of them. At the end, memory occupied by partition in `registeredPartitions` is also a resource to release so being in `registeredPartitions` and unreleased should be always the case. If partition is released/unregistered, there is no point to keep in `registeredPartitions` either.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NetworkEnvironment.java
azagrebin: In this comment I was actually thinking more about how to describe which user calls of shuffle API will lead to the release. `ResultPartitionWriter#finish()` triggers end of production and will result eventually internally in `ResultPartitionManager#onConsumedPartition`. But I think you are right we should mention that the actual release can happen later e.g. when the consumption is done. I will rewrite the comment a bit.

In general, `Close` could also potentially release some resources associated with partition, like memory now, we just do not associate it with release state of partition now because of current implementation details.

azagrebin:
Thanks for the review @zhijiangW ! I addressed the comments.

As for returning a collection of partitions, it is a simple step to prepare for partition lifecycle management where we plan that the `getUnreleasedPartitions` will become part of TM/RM heartbeat where TM reports its currently being held partitions.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NetworkEnvironment.java
```diff
@@ -201,6 +203,28 @@ public void releasePartitions(Collection<ResultPartitionID> partitionIds) {
 		}
 	}
 
+	/**
+	 * Report unreleased partitions.
+	 *
+	 * <p>Partitions are released in the following cases:
```
zentol: I don't believe this javadoc belongs on this method. It contains information about implementation details of various methods, and thus should likely be placed into the class javadoc.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NetworkEnvironment.java
```diff
@@ -172,6 +173,7 @@ private static void registerNetworkMetrics(MetricGroup metricGroup, NetworkBuffe
 	//  Properties
 	// --------------------------------------------------------------------------------------------
 
+	@VisibleForTesting
 	public ResultPartitionManager getResultPartitionManager() {
```
zentol: can this be package private?

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NetworkEnvironment.java
azagrebin: This is mostly a prep step for introducing a shuffle service interface, it is not intended only for netty implementation. We can think where it belongs when the interface is introduced. I would keep it here now because it is mostly relevant here atm to understand what is unreleased.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NetworkEnvironment.java
azagrebin: No, it is used by `StreamNetworkBenchmarkEnvironment` from other package atm. It is netty specific test performance benchmark.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/ResultPartitionManager.java
zhijiangW: Yes, the `partition#release` should also put under the synchronized in `ResultPartitionManager#onConsumedPartition`, then it could solve this issue.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/ResultPartitionManager.java
```diff
@@ -106,30 +107,23 @@ public void shutdown() {
...
-		}
-
-		// Release the partition if it was successfully removed
-		if (partition == previous) {
-			partition.release();
-
-			LOG.debug("Released {}.", partition);
+			final ResultPartition previous = registeredPartitions.remove(partition.getPartitionId());
+			// Release the partition if it was successfully removed
+			if (partition == previous) {
```
zentol: this change exceeds the scope of the original issue and should be put into a separate commit.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NetworkEnvironment.java
azagrebin: Let's remove this doc comment for now in favour of more comprehensive description given in #8608.

azagrebin:
Thanks for the reviews @zhijiangW @zentol ! I have addressed the comments.

PR #8570, opened 2019-05-30
[FLINK-12688] [state] Make serializer lazy initialization thread safe in StateDescriptor

carp84:

This PR aims at making the lazy initialization of `StateDescriptor.serializer` thread safe, more details about what problem there will be w/o the fix please refer to [JIRA](https://issues.apache.org/jira/browse/FLINK-11987).

This PR supersedes [PR#8331](https://github.com/apache/flink/pull/8331) as per discussed there.

## Brief change log

Changes mainly inlcude:
* Use CAS when lazily initializing `serializer` in `StateDescriptor#initializeSerializerUnlessSet`, to make sure only one serializer will be used.
* Remove the nullify of `typeInfo` in `StateDescriptor#initializeSerializerUnlessSet` to prevent NPE as observed in `FlinkKafkaProducer011` from jira description.
* Added a new UT case in `StateDescriptorTest` to reproduce the problem and verify the fix.

## Verifying this change

Added a `testSerializerLazyInitializeInParallel` test case in `StateDescriptorTest` for verification.

     - `StateDescriptor` is annotated with `@PublicEvolving` while changes here don't involve any modification on method/interface signatures.
  - The serializers: (yes)
     - The change here assures only one single serializer instance will be wrapped after lazy initialization while previously no such assurance.

## Documentation

### DIFF
--- a/flink-core/src/main/java/org/apache/flink/api/common/state/StateDescriptor.java
+++ b/flink-core/src/main/java/org/apache/flink/api/common/state/StateDescriptor.java
@@ -20,6 +20,7 @@ package org.apache.flink.api.common.state;
 
 import org.apache.flink.annotation.Internal;
 import org.apache.flink.annotation.PublicEvolving;
+import org.apache.flink.annotation.VisibleForTesting;
 import org.apache.flink.api.common.ExecutionConfig;
 import org.apache.flink.api.common.typeinfo.TypeInformation;
 import org.apache.flink.api.common.typeutils.TypeSerializer;
@@ -27,6 +28,9 @@ import org.apache.flink.api.java.typeutils.TypeExtractor;
 import org.apache.flink.core.memory.DataInputViewStreamWrapper;
 import org.apache.flink.core.memory.DataOutputViewStreamWrapper;
 import org.apache.flink.util.Preconditions;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
@@ -37,6 +41,7 @@ import java.io.IOException;
 import java.io.ObjectInputStream;
 import java.io.ObjectOutputStream;
 import java.io.Serializable;
+import java.util.concurrent.atomic.AtomicReference;
 
 import static org.apache.flink.util.Preconditions.checkNotNull;
 import static org.apache.flink.util.Preconditions.checkState;
@@ -52,6 +57,7 @@ import static org.apache.flink.util.Preconditions.checkState;
  */
 @PublicEvolving
 public abstract class StateDescriptor<S extends State, T> implements Serializable {
+	private static final Logger LOG = LoggerFactory.getLogger(StateDescriptor.class);
 
 	/**
 	 * An enumeration of the types of supported states. Used to identify the state type
@@ -82,8 +88,7 @@ public abstract class StateDescriptor<S extends State, T> implements Serializabl
 	/** The serializer for the type. May be eagerly initialized in the constructor,
 	 * or lazily once the {@link #initializeSerializerUnlessSet(ExecutionConfig)} method
 	 * is called. */
-	@Nullable
-	protected TypeSerializer<T> serializer;
+	private final AtomicReference<TypeSerializer<T>> serializerAtomicReference = new AtomicReference<>();
 
 	/** The type information describing the value type. Only used to if the serializer
 	 * is created lazily. */
@@ -114,7 +119,7 @@ public abstract class StateDescriptor<S extends State, T> implements Serializabl
 	 */
 	protected StateDescriptor(String name, TypeSerializer<T> serializer, @Nullable T defaultValue) {
 		this.name = checkNotNull(name, "name must not be null");
-		this.serializer = checkNotNull(serializer, "serializer must not be null");
+		this.serializerAtomicReference.set(checkNotNull(serializer, "serializer must not be null"));
 		this.defaultValue = defaultValue;
 	}
 
@@ -175,6 +180,7 @@ public abstract class StateDescriptor<S extends State, T> implements Serializabl
 	 */
 	public T getDefaultValue() {
 		if (defaultValue != null) {
+			TypeSerializer<T> serializer = serializerAtomicReference.get();
 			if (serializer != null) {
 				return serializer.copy(defaultValue);
 			} else {
@@ -191,8 +197,19 @@ public abstract class StateDescriptor<S extends State, T> implements Serializabl
 	 * calling {@link #initializeSerializerUnlessSet(ExecutionConfig)}.
 	 */
 	public TypeSerializer<T> getSerializer() {
+		TypeSerializer<T> serializer = serializerAtomicReference.get();
 		if (serializer != null) {
 			return serializer.duplicate();
+		} else {
+			throw new IllegalStateException("Serializer not yet initialized.");
+		}
+	}
+
+	@VisibleForTesting
+	final TypeSerializer<T> getOriginalSerializer() {
+		TypeSerializer<T> serializer = serializerAtomicReference.get();
+		if (serializer != null) {
+			return serializer;
 		} else {
 			throw new IllegalStateException("Serializer not yet initialized.");
 		}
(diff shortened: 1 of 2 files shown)


tzulitai:
@flinkbot approve description

flink-core/src/main/java/org/apache/flink/api/common/state/StateDescriptor.java
```diff
@@ -82,8 +87,7 @@
 	/** The serializer for the type. May be eagerly initialized in the constructor,
 	 * or lazily once the {@link #initializeSerializerUnlessSet(ExecutionConfig)} method
 	 * is called. */
-	@Nullable
-	protected TypeSerializer<T> serializer;
+	private final AtomicReference<TypeSerializer<T>> serializerAtomicReference = new AtomicReference<>();
```
becketqin: Should `queryableStateName` and `ttlConfig` also be protected as they are also mutable fields?

carp84: Yes if we aim at making `StateDescriptor` thread safe but that would be a bigger project and requires more testing.

Considering the seemingly fact that reset `queryableStateName` or `ttlConfig` won't cause obvious trouble, does it make sense if we fix the serializer lazy initialization issue (which has caused a NPE problem on kafka producer) first and then check whether it worth another JIRA/PR for making the whole `StateDescriptor` thread safe? Thanks.

becketqin: Sure, sounds good to me.

carp84:
The travis build failure is due to some de/serialization issue introduced by the original patch here, will push another commit to fix it.

tzulitai:
@flinkbot approve consensus

I'm approving the consensus to fix the problem like this, before we attempt to tackle the problem that serializers can only be lazily instantiated due to lack of `ExecutionConfig`.

Just as a food for thought, there was a similar attempt in the past to make the `StateDescriptor` safe to be used across different threads w.r.t. accessing the state serializer: https://issues.apache.org/jira/browse/FLINK-6775. Apparently, that only partially solved the issue.

flink-core/src/main/java/org/apache/flink/api/common/state/StateDescriptor.java
```diff
@@ -343,6 +351,9 @@ private void writeObject(final ObjectOutputStream out) throws IOException {
 			// we have a default value
 			out.writeBoolean(true);
 
+			TypeSerializer<T> serializer = serializerAtomicReference.get();
+			checkNotNull(serializer, "Serializer not yet initialized.");
```
tzulitai: IMO, would be cleaner to move this check to the beginning of else-branch.
Moving early-returns / validations as early as possible would improve readability.

flink-core/src/main/java/org/apache/flink/api/common/state/StateDescriptor.java
```diff
@@ -376,6 +387,9 @@ private void readObject(final ObjectInputStream in) throws IOException, ClassNot
 
 			in.readFully(buffer);
 
+			TypeSerializer<T> serializer = serializerAtomicReference.get();
+			checkNotNull(serializer, "Serializer not yet initialized.");
```
tzulitai: Same here: move this to the beginning of the if-branch.

flink-core/src/test/java/org/apache/flink/api/common/state/StateDescriptorTest.java
```diff
@@ -200,6 +204,31 @@ public void testEqualsSameNameAndTypeDifferentClass() throws Exception {
...
 
+	@Test
+	public void testSerializerLazyInitializeInParallel() throws Exception {
+		final String name = "testSerializerLazyInitializeInParallel";
+		// use PojoTypeInfo which will create a new serializer when createSerializer is invoked.
+		final TestStateDescriptor<String> desc =
+			new TestStateDescriptor<>(name, new PojoTypeInfo<>(String.class, new ArrayList<>()));
+		final int threadNumber = 20;
+		final ArrayList<Thread> threads = new ArrayList<>(threadNumber);
+		final ExecutionConfig executionConfig = Mockito.mock(ExecutionConfig.class);
```
tzulitai: No need to use mockito here. `new ExecutionConfig()` would suffice.

tzulitai: As a rule of thumb in Flink, we try to avoid mocks as much as possible.

flink-core/src/main/java/org/apache/flink/api/common/state/StateDescriptor.java
```diff
@@ -82,8 +88,8 @@
 	/** The serializer for the type. May be eagerly initialized in the constructor,
 	 * or lazily once the {@link #initializeSerializerUnlessSet(ExecutionConfig)} method
 	 * is called. */
-	@Nullable
-	protected TypeSerializer<T> serializer;
+	@VisibleForTesting
+	protected final AtomicReference<TypeSerializer<T>> serializerAtomicReference = new AtomicReference<>();
```
tzulitai: It's still a bit dangerous to make this `protected`, IMO. We can't guarantee subclasses would not mess around with this.

If we want to make this visible for testing, a separate package-private final method for getting the original wrapped serializer is better.

WDYT?

flink-core/src/test/java/org/apache/flink/api/common/state/StateDescriptorTest.java
```diff
@@ -224,6 +253,15 @@ public void testEqualsSameNameAndTypeDifferentClass() throws Exception {
 		public Type getType() {
 			return Type.VALUE;
 		}
+
+		public TypeSerializer<T> getOriginalSerializer() {
```
tzulitai: This would not be necessary if we move this to the `StateDescriptor` class itself, and make the atomic reference field private. IMO, this is safer (please see my other comment above).

flink-core/src/test/java/org/apache/flink/api/common/state/StateDescriptorTest.java
```diff
@@ -200,6 +204,31 @@ public void testEqualsSameNameAndTypeDifferentClass() throws Exception {
...
+		for (int i = 0; i < threadNumber; i++) {
+			threads.add(new Thread(() -> {
+				desc.initializeSerializerUnlessSet(executionConfig);
+				TypeSerializer<String> serializer = desc.getOriginalSerializer();
+				serializers.put(System.identityHashCode(serializer), serializer);
+			}));
+		}
+		threads.forEach(Thread::start);
+		for (Thread t : threads) {
+			t.join();
```
tzulitai: We should maybe also check for exceptions in these threads. You can use Flink's `CheckedThread` for this purpose.

tzulitai:
One slight clarification regarding a comment that was mentioned in the description:
```
The change here assures only one single serializer instance will be returned by StateDescriptor.getSerializer after lazy initialization while previously no such assurance.
```
If I understood this correctly, this should be incorrect. `getSerializer` always creates a duplicate of the wrapped serializer. This is required since `TypeSerializer` isn't thread-safe.
So, the returned serializer is always a duplicate, but the one maintained within `StateDescriptor` is always a singleton.

flink-core/src/main/java/org/apache/flink/api/common/state/StateDescriptor.java
carp84: I was a little bit struggling about reducing the visibility but I agree exposing this original serializer is dangerous. So I agree to reduce the visibility to private.

And since `StateDescriptor` is annotated as `PublicEvolving`, will add a release note about this on JIRA. Thanks.

carp84:
> One slight clarification regarding a comment that was mentioned in the description:
> 
> ```
> The change here assures only one single serializer instance will be returned by StateDescriptor.getSerializer after lazy initialization while previously no such assurance.
> ```
> 
> If I understood this correctly, this should be incorrect. `getSerializer` always creates a duplicate of the wrapped serializer. This is required since `TypeSerializer` isn't thread-safe.
> So, with this change, the returned serializer is still always a duplicate, but the one maintained within `StateDescriptor` is a singleton.
> 
> @carp84 can you confirm this?

Correct, I meant the original serializer will be a singleton. Let me update the description.

tzulitai:
@flinkbot approve all

tzulitai:
Thanks for addressing comments so quickly @carp84.
LGTM, +1. Will merge after Travis gives a green light.

tzulitai:
Merging ...

carp84:
Thanks all for the careful reviews and thanks @tzulitai for helping merge the PR!

PR #8571, opened 2019-05-30
[FLINK-12682][connectors] StringWriter support custom row delimiter

wisgood:

StringWriter support custom row delimiter 

## Brief change log

Added a parameter to the constructor

### DIFF
--- a/flink-connectors/flink-connector-filesystem/src/main/java/org/apache/flink/streaming/connectors/fs/StringWriter.java
+++ b/flink-connectors/flink-connector-filesystem/src/main/java/org/apache/flink/streaming/connectors/fs/StringWriter.java
@@ -42,6 +42,8 @@ public class StringWriter<T> extends StreamWriterBase<T> {
 
 	private String rowDelimiter;
 
+	private byte[] rowDelimiterBytes;
+
 
 	/**
 	 * Creates a new {@code StringWriter} that uses {@code "UTF-8"} charset to convert
@@ -87,7 +89,10 @@ public class StringWriter<T> extends StreamWriterBase<T> {
 	public void write(T element) throws IOException {
 		FSDataOutputStream outputStream = getStream();
 		outputStream.write(element.toString().getBytes(charset));
-		outputStream.write(rowDelimiter.getBytes(charset));
+		if (rowDelimiterBytes == null) {
+			rowDelimiterBytes = rowDelimiter.getBytes(charset);
+		}
+		outputStream.write(rowDelimiterBytes);
 	}
 
 	@Override
--- a/flink-connectors/flink-connector-filesystem/src/main/java/org/apache/flink/streaming/connectors/fs/StringWriter.java
+++ b/flink-connectors/flink-connector-filesystem/src/main/java/org/apache/flink/streaming/connectors/fs/StringWriter.java
@@ -40,7 +40,7 @@ public class StringWriter<T> extends StreamWriterBase<T> {
 
 	private transient Charset charset;
 
-	private byte[] rowDelimiter ;
+	private String rowDelimiter ;
 
 
 	/**
@@ -48,7 +48,7 @@ public class StringWriter<T> extends StreamWriterBase<T> {
 	 * strings to bytes.
 	 */
 	public StringWriter() {
-		this("UTF-8",new byte[] {'\n'});
+		this("UTF-8", "\n");
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


flink-connectors/flink-connector-filesystem/src/main/java/org/apache/flink/streaming/connectors/fs/StringWriter.java
```diff
@@ -54,13 +57,15 @@ public StringWriter() {
 	 *
 	 * @param charsetName Name of the charset to be used, must be valid input for {@code Charset.forName(charsetName)}
```
Myasuka: A new param `rowDelimiter` should be added.

flink-connectors/flink-connector-filesystem/src/main/java/org/apache/flink/streaming/connectors/fs/StringWriter.java
```diff
@@ -54,13 +57,15 @@ public StringWriter() {
...
 	 */
-	public StringWriter(String charsetName) {
+	public StringWriter(String charsetName,byte[] rowDelimiter ) {
```
Myasuka: There should exist a space before `byte[] rowDelimiter`

flink-connectors/flink-connector-filesystem/src/main/java/org/apache/flink/streaming/connectors/fs/StringWriter.java
```diff
@@ -40,12 +40,15 @@
...
+	private byte[] rowDelimiter ;
+
+
 	/**
 	 * Creates a new {@code StringWriter} that uses {@code "UTF-8"} charset to convert
 	 * strings to bytes.
 	 */
 	public StringWriter() {
-		this("UTF-8");
+		this("UTF-8",new byte[] {'\n'});
```
Myasuka: There should exist a space before `new byte[] {'\n'}`

flink-connectors/flink-connector-filesystem/src/main/java/org/apache/flink/streaming/connectors/fs/StringWriter.java
Myasuka: If we have already provided the `charsetName`, why not just provided the `String` `delimiter` as parameter and the use `delimiter.getBytes(charsetName)` to get the target bytes.

flink-connectors/flink-connector-filesystem/src/main/java/org/apache/flink/streaming/connectors/fs/StringWriter.java
wisgood: good suggestion

flink-connectors/flink-connector-filesystem/src/main/java/org/apache/flink/streaming/connectors/fs/StringWriter.java
```diff
@@ -82,7 +87,7 @@ public void open(FileSystem fs, Path path) throws IOException {
 	public void write(T element) throws IOException {
 		FSDataOutputStream outputStream = getStream();
 		outputStream.write(element.toString().getBytes(charset));
-		outputStream.write('\n');
+		outputStream.write(rowDelimiter.getBytes(charset));
```
Myasuka: We could store the `delimiter` in bytes as the private field in `StringWriter.java` instead of serializing this every time.

wisgood: ok,done

wisgood: getCharsetName()  method is protected,is it better to be public  @Myasuka

wisgood: test has been passed

wisgood: Any other suggestions? @Myasuka

wisgood: I move this  to a new pr，https://github.com/apache/flink/pull/8621 ！Thanks very much！

Myasuka: Sorry for late reply, I noticed that you have addressed all my suggestions on the new PR.

PR #8573, opened 2019-05-30
[FLINK-12670][runtime] Implement FailureRateRestartBackoffTimeStrategy

eaglewatcherwb:
Change-Id: I57b9ae57ea2ac3c501e96a4c84d39346e46ab8bc

*Implement FailureRateRestartBackoffTimeStrategy*

## Brief change log

  - *Implement FailureRateRestartBackoffTimeStrategy*

## Verifying this change

  - *Added unit tests in FailureRateRestartBackoffTimeStrategyTest*

  - The serializers: (no)

## Documentation

### DIFF
--- a/docs/_includes/generated/restart_backoff_time_strategy_configuration.html
+++ b/docs/_includes/generated/restart_backoff_time_strategy_configuration.html
@@ -1,0 +1,26 @@
+<table class="table table-bordered">
+    <thead>
+        <tr>
+            <th class="text-left" style="width: 20%">Key</th>
+            <th class="text-left" style="width: 15%">Default</th>
+            <th class="text-left" style="width: 65%">Description</th>
+        </tr>
+    </thead>
+    <tbody>
+        <tr>
+            <td><h5>restart-backoff-time-strategy.failure-rate.backoff-time</h5></td>
+            <td style="word-wrap: break-word;">0</td>
+            <td>Backoff time in milliseconds between two consecutive restart attempts.</td>
+        </tr>
+        <tr>
+            <td><h5>restart-backoff-time-strategy.failure-rate.failure-rate-interval</h5></td>
+            <td style="word-wrap: break-word;">60000</td>
+            <td>Time interval in milliseconds for measuring failure rate.</td>
+        </tr>
+        <tr>
+            <td><h5>restart-backoff-time-strategy.failure-rate.max-failures-per-interval</h5></td>
+            <td style="word-wrap: break-word;">1</td>
+            <td>Maximum number of failures in given time interval before failing a job.</td>
+        </tr>
+    </tbody>
+</table>
--- a/flink-core/src/main/java/org/apache/flink/configuration/RestartBackoffTimeStrategyOptions.java
+++ b/flink-core/src/main/java/org/apache/flink/configuration/RestartBackoffTimeStrategyOptions.java
@@ -1,0 +1,56 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.configuration;
+
+import org.apache.flink.annotation.PublicEvolving;
+
+/**
+ * Configuration options for the RestartBackoffTimeStrategy.
+ */
+@PublicEvolving
+public class RestartBackoffTimeStrategyOptions {
+	/**
+	 * Maximum number of failures in given time interval {@link #RESTART_BACKOFF_TIME_STRATEGY_FAILURE_RATE_FAILURE_RATE_INTERVAL}
+	 * before failing a job in FailureRateRestartBackoffTimeStrategy.
+	 */
+	@PublicEvolving
+	public static final ConfigOption<Integer> RESTART_BACKOFF_TIME_STRATEGY_FAILURE_RATE_MAX_FAILURES_PER_INTERVAL = ConfigOptions
+		.key("restart-backoff-time-strategy.failure-rate.max-failures-per-interval")
+		.defaultValue(1)
+		.withDescription("Maximum number of failures in given time interval before failing a job.");
+
+	/**
+	 * Time interval in which greater amount of failures than {@link #RESTART_BACKOFF_TIME_STRATEGY_FAILURE_RATE_MAX_FAILURES_PER_INTERVAL}
+	 * causes job fail in FailureRateRestartBackoffTimeStrategy.
+	 */
+	@PublicEvolving
+	public static final ConfigOption<Long> RESTART_BACKOFF_TIME_STRATEGY_FAILURE_RATE_FAILURE_RATE_INTERVAL = ConfigOptions
+		.key("restart-backoff-time-strategy.failure-rate.failure-rate-interval")
+		.defaultValue(60_000L)
+		.withDescription("Time interval in milliseconds for measuring failure rate.");
+
+	/**
+	 * Backoff time between two consecutive restart attempts in FailureRateRestartBackoffTimeStrategy.
+	 */
+	@PublicEvolving
+	public static final ConfigOption<Long> RESTART_BACKOFF_TIME_STRATEGY_FAILURE_RATE_FAILURE_RATE_BACKOFF_TIME = ConfigOptions
+		.key("restart-backoff-time-strategy.failure-rate.backoff-time")
+		.defaultValue(0L)
+		.withDescription("Backoff time in milliseconds between two consecutive restart attempts.");
+}
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/FailureRateRestartBackoffTimeStrategy.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/FailureRateRestartBackoffTimeStrategy.java
@@ -1,0 +1,149 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.runtime.executiongraph.failover.flip1;
+
+import org.apache.flink.configuration.Configuration;
+import org.apache.flink.runtime.util.clock.Clock;
+import org.apache.flink.runtime.util.clock.SystemClock;
+
+import java.util.ArrayDeque;
+import java.util.Deque;
+
+import static org.apache.flink.configuration.RestartBackoffTimeStrategyOptions.RESTART_BACKOFF_TIME_STRATEGY_FAILURE_RATE_FAILURE_RATE_BACKOFF_TIME;
+import static org.apache.flink.configuration.RestartBackoffTimeStrategyOptions.RESTART_BACKOFF_TIME_STRATEGY_FAILURE_RATE_FAILURE_RATE_INTERVAL;
+import static org.apache.flink.configuration.RestartBackoffTimeStrategyOptions.RESTART_BACKOFF_TIME_STRATEGY_FAILURE_RATE_MAX_FAILURES_PER_INTERVAL;
+import static org.apache.flink.util.Preconditions.checkArgument;
+import static org.apache.flink.util.Preconditions.checkNotNull;
+
+/**
+ * Restart strategy which can restart when failure rate is not exceeded.
+ */
+public class FailureRateRestartBackoffTimeStrategy implements RestartBackoffTimeStrategy {
+
+	private final long failuresIntervalMS;
+
+	private final long backoffTimeMS;
+
+	private final int maxFailuresPerInterval;
+
+	private final Deque<Long> failureTimestamps;
+
+	private final String strategyString;
+
+	private final Clock clock;
+
+	FailureRateRestartBackoffTimeStrategy(Clock clock, int maxFailuresPerInterval, long failuresIntervalMS, long backoffTimeMS) {
+
+		checkArgument(maxFailuresPerInterval > 0, "Maximum number of restart attempts per time unit must be greater than 0.");
+		checkArgument(failuresIntervalMS > 0, "Failures interval must be greater than 0 ms.");
+		checkArgument(backoffTimeMS >= 0, "Backoff time must be at least 0 ms.");
+
+		this.failuresIntervalMS = failuresIntervalMS;
+		this.backoffTimeMS = backoffTimeMS;
+		this.maxFailuresPerInterval = maxFailuresPerInterval;
+		this.failureTimestamps = new ArrayDeque<>(maxFailuresPerInterval);
+		this.strategyString = generateStrategyString();
+		this.clock = checkNotNull(clock);
+	}
+
+	@Override
+	public boolean canRestart() {
+		if (isFailureTimestampsQueueFull()) {
+			Long now = clock.absoluteTimeMillis();
+			Long earliestFailure = failureTimestamps.peek();
+
+			return (now - earliestFailure) > failuresIntervalMS;
+		} else {
+			return true;
+		}
+	}
+
+	@Override
+	public long getBackoffTime() {
+		return backoffTimeMS;
+	}
+
+	@Override
+	public void notifyFailure(Throwable cause) {
+		if (isFailureTimestampsQueueFull()) {
+			failureTimestamps.remove();
+		}
+		failureTimestamps.add(clock.absoluteTimeMillis());
+	}
+
+	@Override
+	public String toString() {
+		return strategyString;
+	}
+
+	private boolean isFailureTimestampsQueueFull() {
+		return failureTimestamps.size() >= maxFailuresPerInterval;
+	}
+
+	private String generateStrategyString() {
+		StringBuilder str = new StringBuilder("FailureRateRestartBackoffTimeStrategy(");
+		str.append("FailureRateRestartBackoffTimeStrategy(failuresIntervalMS=");
+		str.append(failuresIntervalMS);
+		str.append(",backoffTimeMS=");
+		str.append(backoffTimeMS);
+		str.append(",maxFailuresPerInterval=");
+		str.append(maxFailuresPerInterval);
+		str.append(")");
+
+		return str.toString();
+	}
+
+	public static FailureRateRestartBackoffTimeStrategyFactory createFactory(final Configuration configuration) {
+		return new FailureRateRestartBackoffTimeStrategyFactory(
+				configuration.getInteger(RESTART_BACKOFF_TIME_STRATEGY_FAILURE_RATE_MAX_FAILURES_PER_INTERVAL),
+				configuration.getLong(RESTART_BACKOFF_TIME_STRATEGY_FAILURE_RATE_FAILURE_RATE_INTERVAL),
+				configuration.getLong(RESTART_BACKOFF_TIME_STRATEGY_FAILURE_RATE_FAILURE_RATE_BACKOFF_TIME));
+	}
+
+	/**
+	 * The factory for creating {@link FailureRateRestartBackoffTimeStrategy}.
+	 */
+	public static class FailureRateRestartBackoffTimeStrategyFactory implements RestartBackoffTimeStrategy.Factory {
+
+		private final int maxFailuresPerInterval;
+
+		private final long failuresIntervalMS;
+
+		private final long backoffTimeMS;
+
+		public FailureRateRestartBackoffTimeStrategyFactory(
+				int maxFailuresPerInterval,
+				long failuresIntervalMS,
+				long backoffTimeMS) {
+
+			this.maxFailuresPerInterval = maxFailuresPerInterval;
+			this.failuresIntervalMS = failuresIntervalMS;
+			this.backoffTimeMS = backoffTimeMS;
+		}
+
+		@Override
+		public RestartBackoffTimeStrategy create() {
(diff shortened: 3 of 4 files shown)


flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/FailureRateRestartBackoffTimeStrategy.java
```diff
@@ -0,0 +1,145 @@
...
+	}
+
+	public static FailureRateRestartBackoffTimeStrategyFactory createFactory(final Configuration configuration) {
+		return new FailureRateRestartBackoffTimeStrategyFactory(configuration);
+	}
+
+	/**
+	 * The factory for creating {@link FailureRateRestartBackoffTimeStrategy}.
+	 */
+	public static class FailureRateRestartBackoffTimeStrategyFactory implements RestartBackoffTimeStrategy.Factory {
```
klion26: seems this class will only be used in `FailureRateRestartBackoffTimeStrategy`, the Java access modifiers may be changed?

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/failover/flip1/FailureRateRestartBackoffTimeStrategyTest.java
```diff
@@ -0,0 +1,67 @@
...
+
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+/**
+ * Unit tests for {@link FailureRateRestartBackoffTimeStrategy}.
+ */
+public class FailureRateRestartBackoffTimeStrategyTest extends TestLogger {
+
+	private final Exception failure = new Exception();
```
klion26: Do we need to add a test for `backofTimeMS` and class `FailureRateRestartBackoffTimeStrategyFactory`?

flink-core/src/main/java/org/apache/flink/configuration/ConfigConstants.java
```diff
@@ -117,6 +117,30 @@
...
+	@PublicEvolving
+	public static final ConfigOption<String> RESTART_BACKOFF_TIME_STRATEGY_FAILURE_RATE_FAILURE_RATE_INTERVAL =
+		key("restart-backoff-time-strategy.failure-rate.failure-rate-interval").defaultValue("1 min");
+
+	/**
+	 * Backoff time between two consecutive restart attempts in FailureRateRestartBackoffTimeStrategy.
+	 * It can be specified using Scala's FiniteDuration notation: "1 min", "20 s".
+	 */
+	@PublicEvolving
+	public static final ConfigOption<String> RESTART_BACKOFF_TIME_STRATEGY_RESTART_BACKOFF_TIME =
```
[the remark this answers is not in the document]
eaglewatcherwb: I think RestartBackoffTimeStrategyOptions is better, since it looks more clear and is consistent with current configuration structure. It may be a little messy if we put options into corresponding strategies since we may support more strategies in the future.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/FailureRateRestartBackoffTimeStrategy.java
eaglewatcherwb: It will be used in `RestartBackoffTimeStrategyFactoryLoader`of another PR.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/failover/flip1/FailureRateRestartBackoffTimeStrategyTest.java
eaglewatcherwb: OK

flink-core/src/main/java/org/apache/flink/configuration/ConfigConstants.java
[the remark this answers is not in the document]
zhuzhurk: Adding the config options to RestartBackoffTimeStrategy will be better to support adding more strategies, otherwise the RestartBackoffTimeStrategyOptions will hosts configurations for various strategies. While Each config option is meant to serve its dedicated strategy. 
Ignoring the backward compatibility consideration, each config option should be recognized by its own strategy only.
That's why I prefer the second option.

As for now, I think both of them work, since the framework need to see config options of the fixed-delay and failure-rate strategies to do the compatibility work.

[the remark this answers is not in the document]
zhuzhurk: FYI. The restart strategy can be supported in a pluggable way, so finally the RestartBackoffTimeStrategyOptions will not know the configurations for all strategies.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/failover/flip1/FailureRateRestartBackoffTimeStrategyTest.java
```diff
@@ -0,0 +1,78 @@
...
+		final int numFailures = 3;
+		final long intervalMS = 1L;
+
+		final FailureRateRestartBackoffTimeStrategy restartStrategy =
+			new FailureRateRestartBackoffTimeStrategy(1, intervalMS, 0);
+
+		for (int failuresLeft = numFailures; failuresLeft > 0; failuresLeft--) {
+			assertTrue(restartStrategy.canRestart());
+			restartStrategy.notifyFailure(failure);
+			Thread.sleep(2 * intervalMS);
```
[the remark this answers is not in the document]
GJL: Imo sleep is not acceptable in unit tests. I think we need to be able to control the time, as such:

https://github.com/apache/flink/blob/350846507450ba7afa0159a2c574bd2f44bacaac/flink-streaming-java/src/test/java/org/apache/flink/streaming/api/functions/sink/TwoPhaseCommitSinkFunctionTest.java#L190-L214

You can use Flink's clock interface:

https://github.com/apache/flink/blob/350846507450ba7afa0159a2c574bd2f44bacaac/flink-runtime/src/main/java/org/apache/flink/runtime/util/clock/Clock.java

https://github.com/apache/flink/blob/350846507450ba7afa0159a2c574bd2f44bacaac/flink-runtime/src/test/java/org/apache/flink/runtime/util/clock/ManualClock.java

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/FailureRateRestartBackoffTimeStrategy.java
```diff
@@ -0,0 +1,145 @@
...
+
+	private final long backoffTimeMS;
+
+	private final int maxFailuresPerInterval;
+
+	private final ArrayDeque<Long> failureTimestamps;
+
+	private final String strategyString;
+
+	public FailureRateRestartBackoffTimeStrategy(int maxFailuresPerInterval, long failuresInterval, long backoffTimeMS) {
```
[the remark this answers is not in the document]
eaglewatcherwb: OK

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/FailureRateRestartBackoffTimeStrategy.java
```diff
@@ -0,0 +1,145 @@
...
+
+			return (now - earliestFailure) > failuresIntervalMS;
+		} else {
+			return true;
+		}
+	}
+
+	@Override
+	public long getBackoffTime() {
+		return backoffTimeMS;
```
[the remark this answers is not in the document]
eaglewatcherwb: `getBackoffTime` is called only when `canRestart` returns `true`, with this contract restarting will not be suppressed.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/FailureRateRestartBackoffTimeStrategy.java
```diff
@@ -0,0 +1,145 @@
...
+	 */
+	public static class FailureRateRestartBackoffTimeStrategyFactory implements RestartBackoffTimeStrategy.Factory {
+
+		private final int maxFailuresPerInterval;
+
+		private final long failuresIntervalMS;
+
+		private final long backoffTimeMS;
+
+		public FailureRateRestartBackoffTimeStrategyFactory(final Configuration configuration) {
```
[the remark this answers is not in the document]
eaglewatcherwb: The backward compatible work is done in this [jira ticket](https://issues.apache.org/jira/browse/FLINK-12709)

docs/_includes/generated/restart_backoff_time_strategy_configuration.html
```diff
@@ -0,0 +1,26 @@
+<table class="table table-bordered">
```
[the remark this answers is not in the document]
eaglewatcherwb: It is generated by the method in the doc.

flink-core/src/main/java/org/apache/flink/configuration/RestartBackoffTimeStrategyOptions.java
```diff
@@ -0,0 +1,53 @@
...
+ */
+@PublicEvolving
+public class RestartBackoffTimeStrategyOptions {
+	/**
+	 * Maximum number of failures in given time interval {@link #RESTART_BACKOFF_TIME_STRATEGY_FAILURE_RATE_FAILURE_RATE_INTERVAL}
+	 * before failing a job in FailureRateRestartBackoffTimeStrategy.
+	 */
+	@PublicEvolving
+	public static final ConfigOption<Integer> RESTART_BACKOFF_TIME_STRATEGY_FAILURE_RATE_MAX_FAILURES_PER_INTERVAL =
+		key("restart-backoff-time-strategy.failure-rate.max-failures-per-interval").defaultValue(1);
```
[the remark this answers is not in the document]
eaglewatcherwb: OK

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/failover/flip1/FailureRateRestartBackoffTimeStrategyTest.java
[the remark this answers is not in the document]
eaglewatcherwb: Thanks a lot for the example, I will do like that.

flink-core/src/main/java/org/apache/flink/configuration/RestartBackoffTimeStrategyOptions.java
```diff
@@ -0,0 +1,53 @@
...
+
+import org.apache.flink.annotation.PublicEvolving;
+
+import static org.apache.flink.configuration.ConfigOptions.key;
+
+/**
+ * Configuration options for the RestartBackoffTimeStrategy.
+ */
+@PublicEvolving
+public class RestartBackoffTimeStrategyOptions {
```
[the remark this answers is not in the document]
eaglewatcherwb: The config of both `FixedDelayRestartBackoffTimeStrategy` and `FailureRateRestartBackoffTimeStrategy` will put in this file, which are strategies for the new scheduler.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/failover/flip1/FailureRateRestartBackoffTimeStrategyTest.java
```diff
@@ -0,0 +1,84 @@
...
+		assertTrue(restartStrategy.canRestart());
+	}
+
+	@Test
+	public void testFailuresExceedingRate() {
+		final int numFailures = 3;
+		final long intervalMS = 10_000L;
+
+		final FailureRateRestartBackoffTimeStrategy restartStrategy =
+			new FailureRateRestartBackoffTimeStrategy(SystemClock.getInstance(), numFailures, intervalMS, 0);
```
zhuzhurk: It's better to use ManualClock here, as the system time elapsing is considered unstable in this case.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/failover/flip1/FailureRateRestartBackoffTimeStrategyTest.java
```diff
@@ -0,0 +1,84 @@
...
+
+		assertFalse(restartStrategy.canRestart());
+	}
+
+	@Test
+	public void testBackoffTime() {
+		final long backoffTimeMS = 10_000L;
+
+		final FailureRateRestartBackoffTimeStrategy restartStrategy =
+			new FailureRateRestartBackoffTimeStrategy(SystemClock.getInstance(), 1, 1, backoffTimeMS);
```
zhuzhurk: Same as above. It's better to use ManualClock here.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/FailureRateRestartBackoffTimeStrategy.java
```diff
@@ -0,0 +1,156 @@
...
+	private final int maxFailuresPerInterval;
+
+	private final Deque<Long> failureTimestamps;
+
+	private final String strategyString;
+
+	private final Clock clock;
+
+	FailureRateRestartBackoffTimeStrategy(Clock clock, int maxFailuresPerInterval, long failuresInterval, long backoffTimeMS) {
+
```
zhuzhurk: `failuresIntervalMS` would be better to show that it is in milliseconds format.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/failover/flip1/FailureRateRestartBackoffTimeStrategyTest.java
eaglewatcherwb: OK

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/failover/flip1/FailureRateRestartBackoffTimeStrategyTest.java
eaglewatcherwb: OK

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/FailureRateRestartBackoffTimeStrategy.java
eaglewatcherwb: OK

flink-core/src/main/java/org/apache/flink/configuration/RestartBackoffTimeStrategyOptions.java
```diff
@@ -0,0 +1,57 @@
...
+		.withDescription("Maximum number of failures in given time interval before failing a job.");
+
+	/**
+	 * Time interval in which greater amount of failures than {@link #RESTART_BACKOFF_TIME_STRATEGY_FAILURE_RATE_MAX_FAILURES_PER_INTERVAL}
+	 * causes job fail in FailureRateRestartBackoffTimeStrategy. It can be specified using Scala's FiniteDuration notation: "1 min", "20 s"
+	 */
+	@PublicEvolving
+	public static final ConfigOption<String> RESTART_BACKOFF_TIME_STRATEGY_FAILURE_RATE_FAILURE_RATE_INTERVAL = ConfigOptions
+		.key("restart-backoff-time-strategy.failure-rate.failure-rate-interval")
+		.defaultValue("1 min")
```
tillrohrmann: I'd be in favor of specifying the interval in milliseconds because in the long run we want to get rid of Scala in the runtime. Thus, at some point we must no longer rely on Scala's `FiniteDuration` for the parsing.

flink-core/src/main/java/org/apache/flink/configuration/RestartBackoffTimeStrategyOptions.java
```diff
@@ -0,0 +1,57 @@
...
+		.withDescription("Time interval for measuring failure rate.");
+
+	/**
+	 * Backoff time between two consecutive restart attempts in FailureRateRestartBackoffTimeStrategy.
+	 * It can be specified using Scala's FiniteDuration notation: "1 min", "20 s".
+	 */
+	@PublicEvolving
+	public static final ConfigOption<String> RESTART_BACKOFF_TIME_STRATEGY_FAILURE_RATE_FAILURE_RATE_BACKOFF_TIME = ConfigOptions
+		.key("restart-backoff-time-strategy.failure-rate.backoff-time")
+		.defaultValue("1 min")
```
tillrohrmann: I'd be in favor of specifying the backoff time in milliseconds because in the long run we want to get rid of Scala in the runtime. Thus, at some point we must no longer rely on Scala's `FiniteDuration` for the parsing.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/FailureRateRestartBackoffTimeStrategy.java
```diff
@@ -0,0 +1,156 @@
...
+		public FailureRateRestartBackoffTimeStrategyFactory(final Configuration configuration) {
+			this.maxFailuresPerInterval = configuration.getInteger(
+				RESTART_BACKOFF_TIME_STRATEGY_FAILURE_RATE_MAX_FAILURES_PER_INTERVAL);
+			this.failuresIntervalMS = getIntervalMSFromConfiguration(
+				configuration,
+				RESTART_BACKOFF_TIME_STRATEGY_FAILURE_RATE_FAILURE_RATE_INTERVAL);
+			this.backoffTimeMS = getIntervalMSFromConfiguration(
+				configuration,
+				RESTART_BACKOFF_TIME_STRATEGY_FAILURE_RATE_FAILURE_RATE_BACKOFF_TIME);
```
tillrohrmann: I would move the configuration parsing logic into the `createFactory` method because it now mixes concerns.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/FailureRateRestartBackoffTimeStrategy.java
```diff
@@ -0,0 +1,156 @@
...
+		this.clock = checkNotNull(clock);
+	}
+
+	@Override
+	public boolean canRestart() {
+		if (isFailureTimestampsQueueFull()) {
+			Long now = clock.absoluteTimeMillis();
+			Long earliestFailure = failureTimestamps.peek();
+
+			return (now - earliestFailure) > failuresIntervalMS;
```
tillrohrmann: How exactly is the interval defined? Assuming my interval is `2L` and I have two failures `T(0), T(2)` occurring at time `0` and `2`, then it would not restart. Thus, the boundaries of the interval `[0, 2]` are inclusive. An alternative could be to make the right boundary exclusive `[0, 2)`.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/failover/flip1/FailureRateRestartBackoffTimeStrategyTest.java
```diff
@@ -0,0 +1,83 @@
...
+
+		ManualClock clock = new ManualClock();
+
+		final FailureRateRestartBackoffTimeStrategy restartStrategy =
+			new FailureRateRestartBackoffTimeStrategy(clock, 1, intervalMS, 0);
+
+		for (int failuresLeft = numFailures; failuresLeft > 0; failuresLeft--) {
+			assertTrue(restartStrategy.canRestart());
+			restartStrategy.notifyFailure(failure);
+			clock.advanceTime(intervalMS + 1, TimeUnit.MILLISECONDS);
```
tillrohrmann: I guess the reason why we increment by `intervalMS + 1` is that the right boundary of the failure interval is inclusive, right?

flink-core/src/main/java/org/apache/flink/configuration/RestartBackoffTimeStrategyOptions.java
eaglewatcherwb: OK

flink-core/src/main/java/org/apache/flink/configuration/RestartBackoffTimeStrategyOptions.java
eaglewatcherwb: OK

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/FailureRateRestartBackoffTimeStrategy.java
eaglewatcherwb: OK

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/failover/flip1/FailureRateRestartBackoffTimeStrategyTest.java
eaglewatcherwb: Yes, you are right.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/FailureRateRestartBackoffTimeStrategy.java
eaglewatcherwb: Yes, I think both are OK. Let’s keep it as now, since it is also consistent with the definition of legacy `FailureRateRestartStrategy`.

flink-core/src/main/java/org/apache/flink/configuration/RestartBackoffTimeStrategyOptions.java
```diff
@@ -0,0 +1,56 @@
...
+		.defaultValue(60_000L)
+		.withDescription("Time interval in milliseconds for measuring failure rate.");
+
+	/**
+	 * Backoff time between two consecutive restart attempts in FailureRateRestartBackoffTimeStrategy.
+	 */
+	@PublicEvolving
+	public static final ConfigOption<Long> RESTART_BACKOFF_TIME_STRATEGY_FAILURE_RATE_FAILURE_RATE_BACKOFF_TIME = ConfigOptions
+		.key("restart-backoff-time-strategy.failure-rate.backoff-time")
+		.defaultValue(10_000L)
```
tillrohrmann: Let's set the default value to `0` because with queued scheduling there is no strict need for a delay between restarts.

docs/_includes/generated/restart_backoff_time_strategy_configuration.html
```diff
@@ -0,0 +1,26 @@
...
+            <td style="word-wrap: break-word;">"1 min"</td>
+            <td>Time interval for measuring failure rate.</td>
+        </tr>
+        <tr>
+            <td><h5>restart-backoff-time-strategy.failure-rate.max-failures-per-interval</h5></td>
+            <td style="word-wrap: break-word;">1</td>
+            <td>Maximum number of failures in given time interval before failing a job.</td>
+        </tr>
+    </tbody>
+</table>
```
tillrohrmann: This needs to be regenerated.

flink-core/src/main/java/org/apache/flink/configuration/RestartBackoffTimeStrategyOptions.java
eaglewatcherwb: OK

docs/_includes/generated/restart_backoff_time_strategy_configuration.html
eaglewatcherwb: OK

PR #8574, opened 2019-05-30
[FLINK-12642][network][metrics] Fix In/OutputBufferPoolUsageGauge failure with NullPointerException if BufferPool has not been inited yet

azagrebin:

The result partition metrics are initialised before `ResultPartitiion#setup` was called. If a reporter tries to access a In/OutputBufferPoolUsageGauge in between it will fail with an `NullPointerException` since the `BufferPool` of the partition is still `null`. Currently, the quick fix is to return zero metrics until the `BufferPool` is initialised. When we have a single-threaded access from `Task#run`, we can merge partition/gate create and setup then it should not be the case anymore.

## Brief change log

Check the partition `BufferPool` is not `null` in `In/OutputBufferPoolUsageGauge#getValue` and return zero metrics if it is `null`.

## Verifying this change

Trivial fix.

  - The serializers: (no)

## Documentation


azagrebin:
@flinkbot attention @zentol  @zhijiangW

PR #8578, opened 2019-05-31
[FLINK-12685] [table-planner-blink] Supports UNNEST query in blink planner

godfreyhe:

*Supports UNNEST query in blink planner*

## Brief change log

  - *added LogicalUnnestRule that rewrites UNNEST to explode function*
  - *added NothingType that is a placeholder internal type for Nothing type*

## Verifying this change

  - *Added test that validates the logical plan and physical plan after LogicalUnnestRule applied*
  - *Extended integration test for UNNEST queries*

## Documentation


flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/plan/optimize/program/FlinkDecorrelateProgram.scala
```diff
@@ -77,6 +81,15 @@ class FlinkDecorrelateProgram[OC <: FlinkOptimizeContext] extends FlinkOptimizeP
         join.getCondition.accept(visitor)
         super.visit(join)
       }
+
+      override def visit(other: RelNode): RelNode = {
+        other match {
+          // ignore Uncollect's inputs due to the correlate variables are from UNNEST directly,
+          // not from cases which RelDecorrelator handles
+          case r: Uncollect => r
```
JingsongLi: > not from cases which RelDecorrelator handles

I don't get it.
The real reason should be that `LogicalUnnestRule` is in the `DEFAULT_REWRITE` phase, and the `DEFAULT_REWRITE` phase is behind the `DECORRELATE` phase, so now we need not verify the Unnest correlation?
Update comments?

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/plan/rules/FlinkBatchRuleSets.scala
```diff
@@ -119,7 +119,9 @@ object FlinkBatchRuleSets {
         new CoerceInputsRule(classOf[LogicalMinus], false),
         ConvertToNotInOrInRule.INSTANCE,
         // optimize limit 0
-        FlinkLimit0RemoveRule.INSTANCE
+        FlinkLimit0RemoveRule.INSTANCE,
+        // unnest rule
+        LogicalUnnestRule.INSTANCE
```
JingsongLi: It seems that `LogicalUnnestRule` is also part of de-correlate. Why not put it in the `DECORRELATE` phase?

flink-table/flink-table-planner-blink/src/test/scala/org/apache/flink/table/plan/rules/logical/LogicalUnnestRuleTest.scala
```diff
@@ -0,0 +1,140 @@
...
+import org.apache.calcite.plan.hep.HepMatchOrder
+import org.apache.calcite.tools.RuleSets
+import org.junit.{Before, Test}
+
+import java.sql.Timestamp
+
+/**
+  * Test for [[LogicalUnnestRule]].
+  */
+class LogicalUnnestRuleTest extends TableTestBase {
```
JingsongLi: `UnnestTestBase` is not enough?

flink-table/flink-table-planner-blink/src/test/scala/org/apache/flink/table/runtime/stream/sql/UnnestITCase.scala
```diff
@@ -0,0 +1,303 @@
...
+import org.apache.flink.table.runtime.utils.{StreamingWithStateTestBase, TestData, TestingAppendSink, TestingRetractSink, TestingRetractTableSink}
+import org.apache.flink.types.Row
+
+import org.junit.Assert.assertEquals
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.junit.runners.Parameterized
+
+@RunWith(classOf[Parameterized])
+class UnnestITCase(mode: StateBackendMode) extends StreamingWithStateTestBase(mode) {
```
JingsongLi: Can you add some cases to PlanCase and ITCase like:
`select xxxx, A.field0, A.field1 from T, unnest(c) as A`

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/calcite/FlinkTypeFactory.scala
```diff
@@ -467,6 +467,9 @@ object FlinkTypeFactory {
       val mapRelDataType = relDataType.asInstanceOf[MapRelDataType]
       mapRelDataType.mapType
 
+    // CURSOR for UDTF case, whose type info will never be used, just a placeholder
+    case CURSOR => NothingType.INSTANCE
```
JingsongLi: Just use a `GenericType(NothingTypeInfo)` is ok?

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/plan/optimize/program/FlinkDecorrelateProgram.scala
godfreyhe: `RelDecorrelator` only handles correlation variables in join, project and filter, does not handle UDTF pattern:`LogicalCorrelate` on `LogicalTableFunctionScan` which will be converted to `BatchExecCorrelate` or `StreamExecCorrelate`. while correlation variables in join, project and filter are not supported in physical phase.  so the validation in `FlinkDecorrelateProgram` only need to check whether there are correlation variables in join, project and filter.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/plan/rules/FlinkBatchRuleSets.scala
godfreyhe: `decorrelate` replaces all correlated expressions in RelNode with non-correlated expressions.
while `LogicalUnnestRule` rewrites `LogicalCorrelate` on `Uncollect` pattern to `LogicalCorrelate` on `LogicalTableFunctionScan` with `explode`, which will be converted to physical Correlate in physical optimization phase. so `LogicalUnnestRule` should be put in `DEFAULT_REWRITE`.

flink-table/flink-table-planner-blink/src/test/scala/org/apache/flink/table/plan/rules/logical/LogicalUnnestRuleTest.scala
godfreyhe: why need to rename to `UnnestTestBase`?  
This class only tests `LogicalUnnestRule`.

flink-table/flink-table-planner-blink/src/test/scala/org/apache/flink/table/runtime/stream/sql/UnnestITCase.scala
godfreyhe: `@Test
  def testUnnestObjectArrayFromTableWithFilter(): Unit = {
    util.addTableSource[(Int, Array[(Int, String)])]("MyTable", 'a, 'b)
    util.verifyPlan("SELECT a, b, s, t FROM MyTable, UNNEST(MyTable.b) AS A (s, t) WHERE s > 13")
  }`
is this case what you want ？

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/calcite/FlinkTypeFactory.scala
godfreyhe: make sense.

flink-table/flink-table-planner-blink/src/test/scala/org/apache/flink/table/runtime/stream/sql/UnnestITCase.scala
JingsongLi: No, I mean `as A` and access it by `A.field0, A.field1` instead of `AS A (s, t)`

godfreyhe: i have added one

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/plan/rules/FlinkBatchRuleSets.scala
JingsongLi: Can you add these replies to code comments?

JingsongLi:
LGTM +1
Please rebase master and ensure travis green.

PR #8579, opened 2019-05-31
[FLINK-12687][table-runtime-blink] ByteHashSet is always in dense mode

liyafan82:

Improve the performance and code structure for org.apache.flink.table.runtime.util.collections.ByteHashSet.

Since there are only 256 possible byte values, the largest possible range is 255, and the condition 

range < OptimizableHashSet.DENSE_THRESHOLD

must be satisfied. So ByteHashSet must be in dense mode.

We can make use of this to improve the performance and code structure.

## Brief change log

  - Change the code so it only considers the case for dense mode.
  

## Verifying this change

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/runtime/util/collections/ByteHashSet.java
+++ b/flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/runtime/util/collections/ByteHashSet.java
@@ -17,8 +17,6 @@
 
 package org.apache.flink.table.runtime.util.collections;
 
-import org.apache.flink.table.util.MurmurHashUtil;
-
 /**
  * Byte hash set.
  */
--- a/flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/runtime/util/collections/ByteHashSet.java
+++ b/flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/runtime/util/collections/ByteHashSet.java
@@ -24,14 +24,12 @@ import org.apache.flink.table.util.MurmurHashUtil;
  */
 public class ByteHashSet extends OptimizableHashSet {
 
-	private byte[] key;
-
-	private byte min = Byte.MAX_VALUE;
-	private byte max = Byte.MIN_VALUE;
+	private final byte min = Byte.MIN_VALUE;
+	private final byte max = Byte.MAX_VALUE;
 
 	public ByteHashSet(final int expected, final float f) {
 		super(expected, f);
-		this.key = new byte[this.n + 1];
+		used = new boolean[max - min + 1];
 	}
 
 	public ByteHashSet(final int expected) {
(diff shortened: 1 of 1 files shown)


liyafan82:
Hi @JingsongLi, would you please take a look?

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/runtime/util/collections/ByteHashSet.java
```diff
@@ -43,109 +39,18 @@ public ByteHashSet() {
 	}
 
 	public boolean add(final byte k) {
-		if (k == 0) {
-			if (this.containsZero) {
-				return false;
-			}
-
-			this.containsZero = true;
+		if (used[k - min]) {
```
JingsongLi: Just simplify it: `return !used[k - min] && (used[k - min] = true);`

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/runtime/util/collections/ByteHashSet.java
```diff
@@ -17,21 +17,17 @@
...
+	private final byte min = Byte.MIN_VALUE;
+	private final byte max = Byte.MAX_VALUE;
 
 	public ByteHashSet(final int expected, final float f) {
 		super(expected, f);
-		this.key = new byte[this.n + 1];
+		used = new boolean[max - min + 1];
 	}
 
 	public ByteHashSet(final int expected) {
```
JingsongLi: Just leave behind this `public ByteHashSet(final int expected)`, remove other useless constructor.

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/runtime/util/collections/ByteHashSet.java
```diff
@@ -17,21 +17,17 @@
 
 package org.apache.flink.table.runtime.util.collections;
 
-import org.apache.flink.table.util.MurmurHashUtil;
-
 /**
  * Byte hash set.
  */
 public class ByteHashSet extends OptimizableHashSet {
```
JingsongLi: Don't extends `OptimizableHashSet`, just keep `addNull` `add` `optimize` `contains` `containsNull`

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/runtime/util/collections/ByteHashSet.java
```diff
@@ -17,21 +17,17 @@
...
 
-	private byte[] key;
-
-	private byte min = Byte.MAX_VALUE;
-	private byte max = Byte.MIN_VALUE;
+	private final byte min = Byte.MIN_VALUE;
```
JingsongLi: Remove useless field, directly use `Byte.MIN_VALUE`

JingsongLi:
Please change JIRA component and PR name to `SQL RUNTIME/Operator` and `table-runtime-blink`.

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/runtime/util/collections/ByteHashSet.java
liyafan82: Thanks for the good suggestion. Your code is definitely more efficient.

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/runtime/util/collections/ByteHashSet.java
liyafan82: I have removed all but the default constructor, because it seems the 'expected' parameter is not used in the code.

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/runtime/util/collections/ByteHashSet.java
liyafan82: Sure. I have revised accordingly.

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/runtime/util/collections/ByteHashSet.java
liyafan82: Sure. I have revised accordingly. Thanks for the suggestion.

liyafan82:
> Please change JIRA component and PR name to `SQL RUNTIME/Operator` and `table-runtime-blink`.

Revised. Thanks.

JingsongLi:
LGTM +1

wuchong:
Thanks for the effort @liyafan82 @JingsongLi . 

Merging...

wuchong:
@flinkbot approve all

PR #8585, opened 2019-05-31
 [FLINK-12690][table-api] Introduce a Planner interface 

dawidwys:

This PR introduces the `Planner` interface that should serve as a bridge between API and planner modules. 

## Brief change log

  - 9a8ae5a - this commit moves the `StreamTransformation` from flink-streaming to flink-core. This is necessary so that we do not add dependency on flink-streaming to flink-table-api
  - Added `Planner` interface that defines methods to parse sql statements and convert relational `TableOperation` program to `StreamTransformations`
  - Introduced ModiyTableOperations which definy a way to exit the Table ecosystem. This way we can provide either `TableSink` to emit to or the expected type information of the produced `StreamTransformations`

## Verifying this change

This change introduces only new interface that are not used yet, therefore no tests are added.

  - The S3 file system connector: (yes / **no /** don't know)

## Documentation


flink-core/src/main/java/org/apache/flink/streaming/api/transformations/StreamTransformation.java
```diff
@@ -100,6 +96,9 @@
 @Internal
 public abstract class StreamTransformation<T> {
```
twalthr: @aljoscha what is your opinion about this change? Should we simply move the class into the same package or should we deprecate the old location and have a new location for streaming related classes in core?

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/operations/QueryTableOperation.java
```diff
@@ -0,0 +1,45 @@
...
+import java.util.List;
+
+/**
+ * Base class for representing an operation structure behind a user-facing {@link Table} API.
+ *
+ * <p>It represents an operation that can be a node of a relational query. It has a schema, that
+ * can be used to validate a {@link QueryTableOperation} applied on top of this one.
+ */
+@Internal
+public interface QueryTableOperation extends TableOperation {
```
twalthr: nit: should we shorten the name to `QueryOperation`?

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/operations/OutputConversionTableOperation.java
```diff
@@ -0,0 +1,65 @@
...
+ * {@link QueryTableOperation}s to a {@link StreamTransformation} of given type described with
+ * {@link TypeInformation}. This is used to convert a relational query to a datastream.
+ *
+ * @param <T> expected type of {@link StreamTransformation}
+ */
+@Internal
+public class OutputConversionTableOperation<T> extends ModifyTableOperation {
+
+	private final QueryTableOperation child;
+	private final TypeInformation<T> type;
```
twalthr: Let's start using DataType.

flink-table/flink-table-planner/src/main/scala/org/apache/flink/table/catalog/ExternalCatalogSchema.scala
```diff
@@ -79,6 +79,7 @@ class ExternalCatalogSchema(
   override def getTable(name: String): Table = try {
     val externalCatalogTable = catalog.getTable(name)
     ExternalTableUtil.fromExternalCatalogTable(isBatch, externalCatalogTable)
+
```
twalthr: unrelated change

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/operations/InlineSinkTableOperation.java
```diff
@@ -0,0 +1,54 @@
...
+import org.apache.flink.table.sinks.TableSink;
+
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * DML operation that tells to write to the given sink.
+ */
+@Internal
+public class InlineSinkTableOperation<T> extends ModifyTableOperation {
```
twalthr: Maybe `UnregisteredSinkTableOperation`?

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/planner/Planner.java
```diff
@@ -0,0 +1,101 @@
...
+	 *
+	 * <p>This method accepts a list of {@link QueryTableOperation}s to allow reusing common
+	 * subtrees of multiple relational queries.
+	 *
+	 * @param tableOperations list of relational operations to plan, optimize and convert in a
+	 * single run.
+	 * @param queryConfig allows configuring runtime behavior of the resulting query
+	 * @return list of corresponding {@link StreamTransformation}s.
+	 */
+	List<StreamTransformation<?>> translate(List<ModifyTableOperation> tableOperations, QueryConfig queryConfig);
```
twalthr: `QueryConfig` looses its purpose with a unified table environment that might merge parts of a table operation tree into a common part. We should merge query config into table config. Table config can be considered a mutable session configuration that can be updated at any time. It can be passed to the constructor of a planner. Thus, we can remove the `QueryConfig` parameter from all methods. What do you think?

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/planner/Planner.java
```diff
@@ -0,0 +1,99 @@
...
+	 * {@link StreamTransformation}s.
+	 *
+	 * <p>This method accepts a list of {@link QueryOperation}s to allow reusing common
+	 * subtrees of multiple relational queries.
+	 *
+	 * @param tableOperations list of relational operations to plan, optimize and convert in a
+	 * single run.
+	 * @return list of corresponding {@link StreamTransformation}s.
+	 */
+	List<StreamTransformation<?>> translate(List<ModifyOperation> tableOperations);
```
godfreyhe: the argument and the comments are inconsistent: the argument is a list of `ModifyOperation`, while the comments describe a list of `QueryOperation`. 

btw, if this method accepts a list of `ModifyOperation`, do we need implicitly add sink or output conversion for plan test queries? like queries in `CalcTest`

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/planner/Planner.java
```diff
@@ -0,0 +1,99 @@
...
+
+	/**
+	 * Returns the AST of the specified Table API and SQL queries and the execution plan
+	 * to compute the result of the given collection of {@link QueryOperation}s.
+	 *
+	 * @param tableOperations The collection of relational queries for which the AST
+	 * and execution plan will be returned.
+	 */
+	String explain(List<QueryOperation> tableOperations);
```
godfreyhe: Blink planner add a parameter `extended` flag to indicate whether to print more details, like: estimated cost, retractTraits for stream plan, etc.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/planner/Planner.java
dawidwys: Sorry for the confusion. I updated the javadoc. Yes it should be always `ModifyOperation` as this is the way we can pass such configuration as update mode or expected output type.

I think blink does a similar thing internally with `DataStreamTableSink`.

Which `CalcTest` do you have in mind? If you are using `TableEnvironment` then it will be added there. If you have test utilities than I guess you should update it.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/planner/Planner.java
dawidwys: I added this parameter

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/planner/Planner.java
godfreyhe: I get your point.  

The `CalcTest` i mentioned above are `org.apache.flink.table.api.batch.sql.CalcTest` in flink planner and `org.apache.flink.table.plan.batch.sql.CalcTest` in blink planner.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/operations/QueryOperationVisitor.java
```diff
@@ -0,0 +1,57 @@
...
+
+	T visitSort(SortQueryOperation sort);
+
+	<U> T visitCalculatedTable(CalculatedQueryOperation<U> calculatedTable);
+
+	T visitCatalogTable(CatalogQueryOperation catalogTable);
+
+	T visitCatalogSink(CatalogSinkOperation catalogSink);
+
+	<U> T visitInlineSink(UnregisteredSinkOperation<U> inlineSink);
```
godfreyhe: I think this method name should be `visitUnregisteredSink` ?

twalthr: nit: let's simplify the method names to `visit(...)` only. I think this is pretty common.

godfreyhe:
hi @dawidwys, there is a minor question: `parse` method in `Planner` class returns validated `TableOperation`, so the `Planner` instance should hold `CatalogManager` instance i think. My question is does `Planner` need methods like `setCatalogManager` or `registerXXX` to connect `TableEnvironment` and `Planner` by service discovery. thanks~

dawidwys:
You are right, the `Planner` will need the `CatalogManager` (not only for parse, but also for Table scans). I don't want to add an `init` method though as this is rather bad design to have objects that you can construct but they are unusable unitl some method call. Therefore we should just add the `CatalogManager` to the ctor of the `Planner`.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/operations/ModifyOperation.java
```diff
@@ -0,0 +1,52 @@
...
+package org.apache.flink.table.operations;
+
+import org.apache.flink.annotation.Internal;
+import org.apache.flink.table.api.QueryConfig;
+import org.apache.flink.table.api.TableSchema;
+
+import java.util.List;
+
+/**
+ * A {@link TableOperation} that describes the DML queries such as e.g. INSERT or conversion to a
```
docete: Did we support UPDATE/MERGE/DELETE semantics？ and how?

dawidwys:
@flinkbot attention @aljoscha

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/operations/ModifyOperation.java
dawidwys: So far we did not support neither of those operations.

In the future, once a `Planner` supports any of them it should return a `DmlTableOperation` that it will understand.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/operations/ModifyOperation.java
```diff
@@ -0,0 +1,52 @@
...
+ * via {@link org.apache.flink.table.planner.Planner#translate(List, QueryConfig)}
+ *
+ * <p>It extends from {@link QueryOperation} to enable applying subsequent
+ * {@link QueryOperation}s on top. In other words it makes it possible to have an output
+ * transformation as a middle step of a relational query. This is not supported yet.
+ *
+ * @see QueryOperation
+ */
+@Internal
+public abstract class ModifyOperation extends QueryOperation {
```
KurtYoung: I think we should not let this class extends from `QueryOperation`, it makes the class hierarchy hard to understand. There are plenty of ways to achieve goal of applying subsequent `QueryOperation` on it.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/operations/QueryOperation.java
```diff
@@ -0,0 +1,107 @@
...
+import java.util.stream.Collectors;
+
+/**
+ * Base class for representing an operation structure behind a user-facing {@link Table} API.
+ *
+ * <p>It represents an operation that can be a node of a relational query. It has a schema, that
+ * can be used to validate a {@link QueryOperation} applied on top of this one.
+ */
+@Internal
+public abstract class QueryOperation implements TableOperation {
```
KurtYoung: How about we further cataloguing all the query-lated sub classes into like `LeafQueryOperation`, `UnaryQueryOperation` and `BinaryQueryOperation`, just like the original `LogicalNode` did. I think this will make user easier to understand this class family.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/operations/ModifyOperation.java
KurtYoung: Also, how about change the name to  `SinkOperation` ?

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/operations/QueryOperation.java
dawidwys: To be honest, I think introducing the `Unary` leaf etc. makes the hierarchy unnecessary more complicated. The only benefit we get are default implementations of `getChildren` method.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/operations/ModifyOperation.java
dawidwys: We wanted to be more inline with SQL terms. You have DQ(uery)L, DM(odify)L, DC(ontrol)L, and DD(efinition)L.

dawidwys: As the javadoc states, the `Modify` operation extends from `Query` operation because modify operations can also produce output and just be part of a SQL query. 

We saw a lot of requests to allow having transformations applied to `Sink`s in a `DataStream` API. This class structure prepares for such use cases.

Right now this is a forbidden operation, because the `getTableSchema` method throws `UnsupportedOperation`, so you just cannot apply subsequent `Operations` on top of this one.

KurtYoung: Keep align with SQL sounds good. 

Regarding to the extends question, I agree we can support reading from a sinked table. But conceptually, this should be separated as two things, 1) write data to a sink. 2) read data from the underlying storage of the table. 
We can provide some syntax sugar for this kind of requirements, but not with the hierarchy. If this class is align with SQL keyword, does "DML is a DQL" sounds weird to you?

dawidwys: I didn't mean `write data to a sink and then read it again`. What I meant is something as follows:
you write data to sink and the output of the operation is e.g. the number of rows stored (this is the standard SQL behavior), or e.g. id of record in an external DB.

Ad. "does DML is a DQL" actually there are quite a few different opinions. The most prominent is that SELECT(DQL) is also a DML, but if you think about a query like `SELECT * FROM INSERT INTO` it may be considered also the other way round.

dawidwys: I had a quick chat with @twalthr about that. Even though I think this is a valid thing to have at some point in time we don't need to have it for now. 

As this is an internal representation we can change that later.

flink-core/src/main/java/org/apache/flink/streaming/api/transformations/StreamTransformation.java
aljoscha: I think it's okay to keep it in the current package. An alternative would be `org.apache.flink.api.dag` or something like it.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/planner/Planner.java
```diff
@@ -0,0 +1,103 @@
...
+	 * <p><b>Note:</b>If the created {@link TableOperation} is a {@link QueryOperation}
+	 * it must be in a form that will be understood by the
+	 * {@link Planner#translate(List)} method.
+	 *
+	 * <p>The produced Operation tree should already be validated.
+	 *
+	 * @param stmt the sql statement to evaluate
+	 * @return parsed query as a tree of relational {@link TableOperation}s
+	 */
+	TableOperation parse(String stmt);
```
twalthr: Make this a list for parsing multiline statements. It is otherwise very difficult to separate by `;` properly.

twalthr: nit: `stmt` -> `statement`

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/planner/Planner.java
```diff
@@ -0,0 +1,103 @@
...
+	 * <p>This method accepts a list of {@link ModifyOperation}s to allow reusing common
+	 * subtrees of multiple relational queries. Each query's top node should be a {@link ModifyOperation}
+	 * in order to pass the expected properties of the output {@link StreamTransformation} such as
+	 * output mode (append, retract, upsert) or the expected output type.
+	 *
+	 * @param tableOperations list of relational operations to plan, optimize and convert in a
+	 * single run.
+	 * @return list of corresponding {@link StreamTransformation}s.
+	 */
+	List<StreamTransformation<?>> translate(List<ModifyOperation> tableOperations);
```
twalthr: nit: `tableOperations` -> `modifyOperations`

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/planner/Planner.java
```diff
@@ -0,0 +1,103 @@
...
+	/**
+	 * Returns the AST of the specified Table API and SQL queries and the execution plan
+	 * to compute the result of the given collection of {@link QueryOperation}s.
+	 *
+	 * @param tableOperations The collection of relational queries for which the AST
+	 * and execution plan will be returned.
+	 * @param extended if the plan should contain additional properties such as
+	 * e.g. estimated cost, traits
+	 */
+	String explain(List<QueryOperation> tableOperations, boolean extended);
```
twalthr: nit: `tableOperations` -> `queryOperations`

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/operations/UnregisteredSinkOperation.java
```diff
@@ -0,0 +1,51 @@
...
+package org.apache.flink.table.operations;
+
+import org.apache.flink.annotation.Internal;
+import org.apache.flink.table.sinks.TableSink;
+
+/**
+ * DML operation that tells to write to the given sink.
+ */
+@Internal
+public class UnregisteredSinkOperation<T> implements ModifyOperation {
```
twalthr: shall we add the `Modify` keyword as well? for `Query` we do it. I know it blows the name up but is a useful information.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/operations/TableOperation.java
```diff
@@ -18,87 +18,16 @@
...
-			stringBuilder.append(Arrays.toString((Object[]) value));
-		} else if (value instanceof Collection) {
-			stringBuilder.append(value);
-		} else {
-			stringBuilder.append("[").append(value).append("]");
-		}
-		return stringBuilder.toString();
-	}
+@PublicEvolving
+public interface TableOperation {
```
twalthr: Shall we rename to `Operation`?

twalthr: I think `asSummaryString` is still valid here.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/operations/ModifyOperation.java
KurtYoung: ok, sounds good to me

PR #8589, opened 2019-06-01
[FLINK-12625][hive][sql-client] Add descriptor, validator, and factory for HiveCatalog

bowenli86:

This PR adds descriptor, validator, and factory for HiveCatalog, and tests  in flink-sql-client.

Design doc can be [found here](https://docs.google.com/document/d/1ALxfiGZBaZ8KUNJtoT443hReoPoJEdt9Db2wiJ2LuWA/edit).

## Brief change log

- added descriptor, validator, and factory for HiveCatalog
- added unit tests in both flink-connector-hive and flnk-sql-client. Factory discovery service tests in flink-sql-client uses`TestHiveCatalog` and `TestHiveCatalogFactory`, and these classes are mostly exactly the same with `HiveCatalog` and `HiveCatalogFactory` except a few minimal differences for testing purpose
- fixed dependency convergence issues found when adding hive test dependencies to flink-sql-client by excluding conflicted libs

## Verifying this change

- added `HiveCatalogDescriptorTest` and `HiveCatalogFactoryTest` in flink-connector-hive
- added `TestHiveCatalog` and `TestHiveCatalogFactory`in flink-sql-client along with hive catalog entries in test-sql-client-catalogs.yaml for test factory discovery.

  - The serializers: (no)

## Documentation

  - Does this pull request introduce a new feature? (yes )

Docs will be added later on.


bowenli86:

flink-connectors/flink-connector-hive/pom.xml
```diff
@@ -66,13 +61,61 @@ under the License.
...
+				<exclusion>
+					<groupId>org.codehaus.jackson</groupId>
+					<artifactId>jackson-mapper-asl</artifactId>
+				</exclusion>
+			</exclusions>
 		</dependency>
 
 		<dependency>
 			<groupId>org.apache.hadoop</groupId>
 			<artifactId>hadoop-mapreduce-client-core</artifactId>
```
xuefuz: I'm wondering which part of Hive connector requires MR client lib...

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/descriptors/HiveCatalogDescriptor.java
```diff
@@ -0,0 +1,55 @@
...
+public class HiveCatalogDescriptor extends CatalogDescriptor {
+
+	private String hiveMetastoreUris;
+
+	// TODO : set default database
+	public HiveCatalogDescriptor(String hiveMetastoreUris) {
+		super(CATALOG_TYPE_VALUE_HIVE, 1);
+
+		Preconditions.checkArgument(!StringUtils.isNullOrWhitespaceOnly(hiveMetastoreUris));
+		this.hiveMetastoreUris = hiveMetastoreUris;
```
xuefuz: I think it's ok if user doens't specify hms uri, in which case, we should get hms info from the environment.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/factories/HiveCatalogFactory.java
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	@Override
+	public List<String> supportedProperties() {
+		List<String> properties = new ArrayList<>();
+
+		// default database
+		properties.add(CATALOG_DEFAULT_DATABASE);
+
+		properties.add(CATALOG_PROPERTIES);
```
xuefuz: CATALOG_PROPERTIES="properties". I'm not sure what it means adding this in the supported property list.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalog.java
```diff
@@ -124,7 +126,7 @@ public HiveCatalog(String catalogName, String defaultDatabase, HiveConf hiveConf
 		LOG.info("Created HiveCatalog '{}'", catalogName);
 	}
 
-	private static HiveConf getHiveConf(String hiveMetastoreURI) {
+	public static HiveConf getHiveConf(String hiveMetastoreURI) {
```
xuefuz: Instead of exposing this method, maybe we should move the logic to the caller instead. After all, the logic is simple enough and HiveConf instance is all HiveCatalog needs in order to connect to HMS properly. How to get HiveConf is up to the caller.

flink-table/flink-sql-client/src/test/java/org/apache/flink/table/client/gateway/local/DependencyTest.java
```diff
@@ -150,4 +154,50 @@ public TestCatalog(String name, String defaultDatabase) {
 			super(name, defaultDatabase);
 		}
 	}
+
+	/**
+	 * A testing catalog for Hivecatalog. In unit test, since there's no Hive metastore
+	 * for the catalog to connect to, we override the open and close methods to not setup any connection.
+	 */
+	public static class TestHiveCatalog extends HiveCatalog {
```
xuefuz: The test seems hacky to me. I'm wondering if we can just use embedded hms instead.

flink-connectors/flink-connector-hive/pom.xml
bowenli86: it's hive metastore requires `org.apache.hadoop.mapred.JobConf`. We can take some time to investigate why it's required later on

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/descriptors/HiveCatalogDescriptor.java
bowenli86: I'm not sure about using environment vars yet because it means the yaml config file is not the single source of truth anymore. I'll ask Timo later

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/factories/HiveCatalogFactory.java
bowenli86: removed

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/descriptors/HiveCatalogDescriptor.java
lirui-apache: I'd prefer to let user explicitly specify the URIs, especially if we allow multiple hive catalogs talking to different HMS instances in the yaml file.

bowenli86:
Rebased this PR to https://github.com/apache/flink/pull/8616

bowenli86:
@xuefuz can you take a look again?

bowenli86:
cc @xuefuz

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalog.java
```diff
@@ -132,9 +134,11 @@ public HiveCatalog(String catalogName, @Nullable String defaultDatabase, @Nullab
 		this.hiveConf = hiveConf == null ? getHiveConf(null) : hiveConf;
```
xuefuz: Is there a use case where hiveConf might be null? If not, we can assert it's not null.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalog.java
```diff
@@ -132,9 +134,11 @@ public HiveCatalog(String catalogName, @Nullable String defaultDatabase, @Nullab
...
 
 		LOG.info("Created HiveCatalog '{}'", catalogName);
+
+		HiveConf.setHiveSiteLocation(null);
```
xuefuz: I don't think we should reset this. User passes us a HiveConf instance and we use it. We don't need to update either the class or the instance.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/descriptors/HiveCatalogDescriptor.java
```diff
@@ -0,0 +1,61 @@
...
+		this.hiveSitePath = hiveSitePath;
+
+		return this;
+	}
+
+	@Override
+	protected Map<String, String> toCatalogProperties() {
+		final DescriptorProperties properties = new DescriptorProperties();
+
+		if (hiveSitePath != null) {
```
xuefuz: In the constructor we aserrted null or empty for hiveSiteUrl. How come here hiveSitePath can be null? We need to be consistent.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/factories/HiveCatalogFactory.java
```diff
@@ -0,0 +1,93 @@
...
+		properties.add(CATALOG_HIVE_SITE_PATH);
+
+		return properties;
+	}
+
+	@Override
+	public Catalog createCatalog(String name, Map<String, String> properties) {
+		final DescriptorProperties descriptorProperties = getValidatedProperties(properties);
+
+		final String defaultDatabase =
```
xuefuz: There is a TODO item above but here we're processing it. It might be better if we can be consistent.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/factories/HiveCatalogFactory.java
```diff
@@ -0,0 +1,93 @@
...
+			descriptorProperties.getOptionalString(CATALOG_DEFAULT_DATABASE)
+				.orElse(HiveCatalog.DEFAULT_DB);
+
+		final Optional<String> hiveSitePath = descriptorProperties.getOptionalString(CATALOG_HIVE_SITE_PATH);
+
+		return new HiveCatalog(name, defaultDatabase, getHiveConf(hiveSitePath.orElse(null)));
+	}
+
+	@VisibleForTesting
+	protected HiveConf getHiveConf(String hiveSitePath) {
```
xuefuz: I'm wondering where we need this method here.

bowenli86: removed as a result of the refactoring

bowenli86:
@xuefuz as we discussed offline, I kept only one public constructor and a protected constructor for testing purpose

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/descriptors/HiveCatalogDescriptor.java
xuefuz: ok.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalog.java
```diff
@@ -114,46 +114,22 @@
...
-				throw new CatalogException(
-					String.format("Failed to get hive-site.xml from the given path '%s'", filePath), e);
-			}
-		}
-
-		return url;
-	}
-
-	private static HiveConf getHiveConf(URL hiveSiteUrl) {
+	public static HiveConf getHiveConf(URL hiveSiteUrl) {
```
xuefuz: 1. This doesn't seem needing to be public
2. Maybe rename it to createHiveConf.

flink-connectors/flink-connector-hive/src/test/java/org/apache/flink/table/catalog/hive/HiveTestUtils.java
```diff
@@ -36,25 +37,34 @@
...
+	public static HiveCatalog createHiveCatalog(String catalogName) {
+		return new HiveCatalog(catalogName, null, getHiveConf());
 	}
 
 	public static HiveCatalog createHiveCatalog(HiveConf hiveConf) {
 		return new HiveCatalog(CatalogTestBase.TEST_CATALOG_NAME, null, hiveConf);
 	}
 
-	public static HiveConf getHiveConf() throws IOException {
+	public static HiveConf getHiveConf() {
```
xuefuz: rename to createHiveConf?

bowenli86:
Thanks. Merging

PR #8593, opened 2019-06-03
Distinguish the method of using NC on Linux and Windows in doc comments.

LingangJiang:
Distinguish the method of using NC on Linux and Windows in doc comments.


StephanEwen:
Look good, thanks.

Merging this...

PR #8595, opened 2019-06-03
[FLINK-12707]Close minicluster will cause memory leak when there are StreamTask closed abnormal	

liu-zhaokun:
[https://issues.apache.org/jira/browse/FLINK-12707](https://issues.apache.org/jira/browse/FLINK-12707)
There are several threads in my application,and every thread will launch a flink job with LocalStreamEnvironment/MiniCluster.But when I interrupt these threads again and again, I found there are many residue threads,and that caused a memory leak.
When I debug it,this message appears "Recipient[Actorakka://flink/user/taskmanager_0#-584606215] had already been terminated. Sender[null] sent the message of type "org.apache.flink.runtime.rpc.messages.LocalRpcInvocation". So I think
when flink close minicluster,TaskExecutor will be closed in the first place,and this operation will cause akka message which will close StreamTask abnormal.So the work thread will be more and more.
I call Thread sleep in my application to avoid this problem,any good suggestions?

### DIFF
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/minicluster/MiniCluster.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/minicluster/MiniCluster.java
@@ -459,6 +459,25 @@ public class MiniCluster implements JobExecutorService, AutoCloseableAsync {
 		}
 	}
 
+	@Override
+	public void close(){
+		try {
+			this.closeAsync();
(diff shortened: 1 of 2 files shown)


tillrohrmann:
Thanks for opening this PR @liu-zhaokun. I'm not sure whether this solves the underlying problem. I'll close this PR. Let's discuss on the JIRA first what we want to solve.

PR #8599, opened 2019-06-03
[FLINK-12650][docs] Redirect Users to Documentation Homepage if Requested Resource Does Not Exist

sjwiesman:

Currently, if you go open a documentation page, which does not exist like

https://ci.apache.org/projects/flink/flink-docs-release-1.8/dev/event_timed.html

the reply is 404. In preparation of a larger restructuring, it would be good if instead of a 404 we would redirect the user to the documentation homepage for this version of the docs.

**Important** Depending on the current configuration of the apache infra servers we may require a follow-up INFRA ticket to enable serving our 404 page. Regardless of whether or not that needs to be done, this is pre-requisite work so it does not need to be blocked on discovery. 

## Brief change log

Add a 404 redirect to the documentation

## Verifying this change

### DIFF
--- a/docs/404.md
+++ b/docs/404.md
@@ -1,0 +1,26 @@
+---
+title: "404"
+permalink: /404.html
+layout: 404_base
+---
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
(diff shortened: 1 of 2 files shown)


sjwiesman:
@flinkbot attention @rmetzger

rmetzger:
@flinkbot approve all

PR #8602, opened 2019-06-03
[FLINK-12313] Add workaround to avoid race condition in SynchronousCheckpointITCase test

1u0:

The `SynchronousCheckpointITCase` has a race condition and with some chance may fail on random tests on CI. This PR adds an additional synchronization point as a workaround to avoid the issue.

Additionally, the PR simplifies (refactors) the test, although the race condition is present in both refactored and non-refactored versions.

## Brief change log

  - The test is simplified to not use dedicated locks for general flow checks. The test run set to limited by a timeout.
  - Added a hacky synchronization point in `advanceToEndOfEventTime` method, to make sure that `triggerCheckpoint` thread has progressed far enough, before triggering `notifyCheckpointComplete`.

## Verifying this change

## Documentation

### DIFF
--- a/flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/SynchronousCheckpointITCase.java
+++ b/flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/SynchronousCheckpointITCase.java
@@ -66,7 +66,9 @@ import org.apache.flink.runtime.util.TestingTaskManagerRuntimeInfo;
 import org.apache.flink.util.SerializedValue;
 
 import org.junit.Before;
+import org.junit.Rule;
 import org.junit.Test;
+import org.junit.rules.Timeout;
 
 import java.util.Collections;
 import java.util.concurrent.Executor;
@@ -107,6 +109,9 @@ public class SynchronousCheckpointITCase {
 		synchronousCheckpointPhase.setState(CheckpointingState.NONE);
 		error.set(null);
 	}
+
+	@Rule
+	public final Timeout timeoutPerTest = Timeout.seconds(10);
 
 	@Test
 	public void taskCachedThreadPoolAllowsForSynchronousCheckpoints() throws Exception {
--- a/flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/SynchronousCheckpointITCase.java
+++ b/flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/SynchronousCheckpointITCase.java
@@ -90,6 +90,7 @@ public class SynchronousCheckpointITCase {
 	private static OneShotLatch cancellationLatch;
 	private static OneShotLatch checkpointCompletionLatch;
 	private static OneShotLatch notifyLatch;
+	private static OneShotLatch checkpointTriggered = new OneShotLatch();
 
 	private static MultiShotLatch checkpointLatch;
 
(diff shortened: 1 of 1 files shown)


1u0:
@flinkbot attention @pnowojski @StefanRRichter

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/SynchronousCheckpointITCase.java
```diff
@@ -108,6 +110,9 @@ public void initializeLatchesAndError() {
 		error.set(null);
 	}
 
+	@Rule
+	public final Timeout timeoutPerTest = Timeout.seconds(7);
```
pnowojski: Why did you pick 7 seconds? I think usually we set the timeouts to 30, 60 or 120 seconds. 7 seconds might be too fast on travis.

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/SynchronousCheckpointITCase.java
```diff
@@ -125,16 +126,24 @@ public void taskCachedThreadPoolAllowsForSynchronousCheckpoints() throws Excepti
...
+			// SynchronousCheckpointTestingTask.advanceToEndOfEventTime) and only then proceeding to
+			// notifyCheckpointComplete.
+			// Without such synchronization, the notifyCheckpointComplete execution may be executed first and leave this
+			// test in a deadlock.
 			task.triggerCheckpointBarrier(
 					42,
 					156865867234L,
 					new CheckpointOptions(CheckpointType.SYNC_SAVEPOINT, CheckpointStorageLocationReference.getDefault()),
-					false);
+					true);
```
pnowojski: Instead of that, could we return a `CompletableFuture` from the stack trace:
```
Task#triggerCheckpointBarrier ->
Task#executeAsyncCallRunnable -> {
	(...)
	try {
		return executor.submit(runnable, blocking);
	}
	(...)
```
and wait for the future to complete before calling `notifyCheckpointComplete`?

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/SynchronousCheckpointITCase.java
1u0: Doesn't it depend on a test? I see a lot of tests in Flink using 10, 5, 2 seconds as timeout values, it's hard to tell that 30, 60 or 120 is a "usual" value.

On my dev machine, the test runs takes about less than 2.5 second. 7 seconds is about x3 of the estimated run time. On some machines 5 seconds may be too little, but more than 10 seconds is too much for this test, imo.

Do you have some other value in mind?

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/SynchronousCheckpointITCase.java
1u0: Do you mean to get a future that would wait for result of `invokable.triggerCheckpoint()` call?
In that case, I think what you are proposing would deadlock the test.

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/SynchronousCheckpointITCase.java
pnowojski: Just saying from my experience: I would set the timeout to 60 seconds.

Travis can often stall for many seconds. I've never investigated the reasons why (swapping? gc? vm failover?), but the problem with travis is not that it's 2 or 3 times slower, but that it can randomly stall from time to time. For example a test might usually execute in 2 seconds, but from time to time it will take 30 seconds to complete.

On the other hand, is there any benefit of timeouting those tests in 7s instead of 60s? On travis I think there is none. Locally there is benefit only if you are debugging this particular test failure for shorter feedback loop, which is also a rare case and one can manually and temporarily lower the timeout. I think tests that are setting some lower timeouts are actively waiting on it and that creates a motivation to lower the timeout as low as possible (to speed up the tests).

(If you insist on trying this 7s timeout, I won't block it)

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/SynchronousCheckpointITCase.java
pnowojski: Yes, you are right. I forgot that the code is waiting on the latch there.

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/SynchronousCheckpointITCase.java
1u0: I don't insist on 7 seconds at all, and as mentioned before, I'm open for other values.

I just think that with too big timeouts (like 30 seconds and above), it's more probable that something wrong with the test itself, rather with the environment running it (**note:** we are discussing only this concrete test, "YMMV" for other tests). In that case I'd prefer the test to fail reasonably fast (even on CI, as it uses shared resources).

> LGTM % timeout. If you don't want to change it, let me know and I will merge it,

If you want, **I can change it to 10 seconds before merging this PR**. Otherwise, it looks like we are both not enforcing any specific value, so we can leave it as is for now.

> but in that case please monitor jira issues if someone is reporting this test as flaky.

I'll adjust the timeout (new or current) if needed and if I notice such issue, but I don't promise to vigorously monitor jira issues for that.

1u0:
@pnowojski, thank you for the review and feedback.

I've updated the test timeout to 10 seconds, and propose to stick with it for now.
If it would fail in the future due to too small time window and if I notice it, I can open a new PR to increase it.

pnowojski:
I'll merge it once travis is green (previous build has failed on some unrelated issue).

pnowojski:
Travis is green, merging.

PR #8603, opened 2019-06-04
[FLINK-12608][runtime] Add getVertexOrThrow and

eaglewatcherwb:
getResultPartitionOrThrow in SchedulingTopology to simplify code
structure

Change-Id: I33be923c3279755d360e95336199fb1bc3257f4f

*Add getVertexOrThrow and getResultPartitionOrThrow in SchedulingTopology to simplify code structure.*

## Brief change log

  - *Add getVertexOrThrow and getResultPartitionOrThrow in SchedulingTopology.*

## Verifying this change

  - Added unit tests in ExecutionGraphToSchedulingTopologyAdapterTest. 

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/strategy/SchedulingTopology.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/strategy/SchedulingTopology.java
@@ -47,7 +47,8 @@ public interface SchedulingTopology {
 	 * Looks up the {@link SchedulingExecutionVertex} for the given {@link ExecutionVertexID}.
 	 *
 	 * @param executionVertexId identifying the respective scheduling vertex
-	 * @return The respective scheduling vertex or throw {@link Exception} if the vertex does not exist
+	 * @return The respective scheduling vertex
+	 * @throws IllegalArgumentException If the vertex does not exist
 	 */
 	default SchedulingExecutionVertex getVertexOrThrow(ExecutionVertexID executionVertexId) {
 		return getVertex(executionVertexId).orElseThrow(
@@ -66,7 +67,8 @@ public interface SchedulingTopology {
 	 * Looks up the {@link SchedulingResultPartition} for the given {@link IntermediateResultPartitionID}.
 	 *
 	 * @param intermediateResultPartitionId identifying the respective scheduling result partition
-	 * @return The respective scheduling result partition or throw {@link Exception} if the partition does not exist
+	 * @return The respective scheduling result partition
+	 * @throws IllegalArgumentException If the partition does not exist
 	 */
 	default SchedulingResultPartition getResultPartitionOrThrow(IntermediateResultPartitionID intermediateResultPartitionId) {
 		return getResultPartition(intermediateResultPartitionId).orElseThrow(
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/strategy/SchedulingTopology.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/strategy/SchedulingTopology.java
@@ -49,7 +49,10 @@ public interface SchedulingTopology {
 	 * @param executionVertexId identifying the respective scheduling vertex
 	 * @return The respective scheduling vertex or throw {@link Exception} if the vertex does not exist
 	 */
-	SchedulingExecutionVertex getVertexOrThrow(ExecutionVertexID executionVertexId);
+	default SchedulingExecutionVertex getVertexOrThrow(ExecutionVertexID executionVertexId) {
+		return getVertex(executionVertexId).orElseThrow(
+				() -> new IllegalArgumentException("can not find vertex: " + executionVertexId));
+	}
 
 	/**
 	 * Looks up the {@link SchedulingResultPartition} for the given {@link IntermediateResultPartitionID}.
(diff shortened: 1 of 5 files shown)


flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/strategy/SchedulingTopology.java
```diff
@@ -43,11 +43,27 @@
 	 */
 	Optional<SchedulingExecutionVertex> getVertex(ExecutionVertexID executionVertexId);
 
+	/**
+	 * Looks up the {@link SchedulingExecutionVertex} for the given {@link ExecutionVertexID}.
+	 *
+	 * @param executionVertexId identifying the respective scheduling vertex
+	 * @return The respective scheduling vertex or throw {@link Exception} if the vertex does not exist
+	 */
+	SchedulingExecutionVertex getVertexOrThrow(ExecutionVertexID executionVertexId);
```
GJL: Did you consider using a default method? https://docs.oracle.com/javase/tutorial/java/IandI/defaultmethods.html

```
default SchedulingExecutionVertex getVertexOrThrow(ExecutionVertexID executionVertexId) {
...
}

(This is not a review)
```

eaglewatcherwb: Emm, `default` is a good way to add new method in interfaces, since in some user cases it is not easy to add corresponding method in the existing implementations. But using `default` may have the risk of `existing implementations of an interface may compile without error or warning but fail at runtime` (as Item 21 in Effective Java), thus I prefer not to use `default` as far as possible.
Anyway, thanks a lot for providing this `default` proposal.

eaglewatcherwb: Emm, it seems there is no risk of current implementation. We could use `default`.

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/strategy/TestingSchedulingTopology.java
```diff
@@ -210,8 +210,8 @@ private ProducerConsumerAllToAllConnectionBuilder(
 				for (TestingSchedulingExecutionVertex consumer : consumers) {
 					consumer.addConsumedPartition(resultPartition);
 					resultPartition.addConsumer(consumer);
-					resultPartitions.add(resultPartition);
 				}
+				resultPartitions.add(resultPartition);
```
GJL: Change should be a separate commit

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/strategy/SchedulingTopology.java
```diff
@@ -43,11 +43,33 @@
 	 */
 	Optional<SchedulingExecutionVertex> getVertex(ExecutionVertexID executionVertexId);
 
+	/**
+	 * Looks up the {@link SchedulingExecutionVertex} for the given {@link ExecutionVertexID}.
+	 *
+	 * @param executionVertexId identifying the respective scheduling vertex
+	 * @return The respective scheduling vertex or throw {@link Exception} if the vertex does not exist
```
GJL: * It is misleading to write that we potentially throw `Exception` because `IllegalArgumentException` is unchecked. 
* You can use @throws tag

```
* @return The respective scheduling vertex
* @throws IllegalArgumentException If the vertex does not exist
```

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/strategy/SchedulingTopology.java
```diff
@@ -43,11 +43,33 @@
...
 	 * @param intermediateResultPartitionId identifying the respective scheduling result partition
 	 * @return Optional containing the respective scheduling result partition or none if the partition does not exist
 	 */
 	Optional<SchedulingResultPartition> getResultPartition(IntermediateResultPartitionID intermediateResultPartitionId);
+
+	/**
+	 * Looks up the {@link SchedulingResultPartition} for the given {@link IntermediateResultPartitionID}.
+	 *
+	 * @param intermediateResultPartitionId identifying the respective scheduling result partition
+	 * @return The respective scheduling result partition or throw {@link Exception} if the partition does not exist
```
GJL: See my other comment about `Exception`.

flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/strategy/TestingSchedulingTopology.java
eaglewatcherwb: OK

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/strategy/SchedulingTopology.java
eaglewatcherwb: OK

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/strategy/SchedulingTopology.java
eaglewatcherwb: OK

GJL:
Merging.

PR #8604, opened 2019-06-04
[DOC] add  ')'  for  the doc checkpoints.md

hehuiyuan:
For `Configure for per job when constructing the state backend`, add ')'


hehuiyuan:
@rmetzger Hi

PR #8133, opened 2019-04-10
[FLINK-12146][network] Remove unregister task from NetworkEnvironment to simplify the interface of ShuffleService

zhijiangW:
## What is the purpose of the change

*`NetworkEnvironment#unregisterTask` is used for closing partition/gate and releasing partition from `ResultPartitionManager`. partition/gate close could be done in task which already maintains the arrays of them. Further we could release partition from `ResultPartitionManager` inside `ResultPartition` via introducing `ResultPartition#fail(Throwable)`. To do so, the `NetworkEnvironment#unregisterTask` could be totally replaced to remove. The benefit is simplifying the method of `NetworkEnvironment` which would be regarded as default `ShuffleService` implementation.*

## Brief change log

  - *Remove `unregisterTask` from `NetworkEnvironment`*
  - *Introduce `close(Throwable)` in `ResultPartition` for releasing*
  - *Fix the related logic in task class*

## Verifying this change

*This change is already covered by existing tests*.

## Does this pull request potentially affect one of the following parts:

  - Dependencies (does it add or upgrade a dependency): (yes / **no**)
  - The public API, i.e., is any changed class annotated with `@Public(Evolving)`: (yes / **no**)
  - The serializers: (yes / **no** / don't know)
  - The runtime per-record code paths (performance sensitive): (yes / **no** / don't know)
  - Anything that affects deployment or recovery: JobManager (and its components), Checkpointing, Yarn/Mesos, ZooKeeper: (yes / **no** / don't know)
  - The S3 file system connector: (yes / **no** / don't know)

## Documentation

  - Does this pull request introduce a new feature? (yes / **no**)
  - If yes, how is the feature documented? (**not applicable** / docs / JavaDocs / not documented)

PR #7938, opened 2019-03-07
[FLINK-10941] Keep slots which contain unconsumed result partitions (on top of #7186)

azagrebin:
## What is the purpose of the change

More changes based on #7186.

## Brief change log

The task executor gateway can use `NetworkEviroment.PartitionManager` to check still registered and unreleased partitions because it is the central point to manage partitions. It would also simplify how it is queried now going through `TaskSlot` and lingering `Tasks`. Eventually, `NetworkEviroment` becomes `ShuffleService` which could decide whether producer can be released or not this way or another.

We still make producer task executor depend on another consumer task executor, though, we do not see any problems with it now, I think it is still more safe to introduce an option as a feature flag. By default, task executor will wait for consumers, as this PR suggests, but users can always fallback to the previous behaviour using the option.

## Verifying this change

unit tests, e2e tests

## Does this pull request potentially affect one of the following parts:

  - Dependencies (does it add or upgrade a dependency): (no)
  - The public API, i.e., is any changed class annotated with `@Public(Evolving)`: (no)
  - The serializers: (no)
  - The runtime per-record code paths (performance sensitive): (no)
  - Anything that affects deployment or recovery: JobManager (and its components), Checkpointing, Yarn/Mesos, ZooKeeper: (no)
  - The S3 file system connector: (no)

## Documentation

  - Does this pull request introduce a new feature? (no)
  - If yes, how is the feature documented? (not applicable)

PR #1780, opened 2016-03-10
[FLINK-3595] [runtime] Eagerly destroy buffer pools on cancelling

uce:
When canceling a job, the Kafka 0.9 Consumer Thread may be stuck in a blocking method (output emitting) and never wakes up.

The thread as a whole cannot be simply interrupted, because of a bug in Kafka that makes the consumer freeze/hang up on interrupt.

With this PR, we destroy the buffer pools eagerly on canceling. The Kafka thread will then throw an exception if it is stuck in emitting elements and it will terminate, which is accepted in case the status is canceled.

PR #8608, opened 2019-06-04
[FLINK-11392][network] Introduce ShuffleEnvironment interface

azagrebin:

Extract `ShuffleEnvironment` interface from `NetworkEnvironment`.

## Brief change log

- Add page size config also to `TaskManagerServicesConfiguration` because this general option is also used for `MemoryManager` independently from `NetworkEnvironment`. Any other `ShuffleEnvironment` can potentially use its own page size.
- Extract `ShuffleEnvironment` interface from `NetworkEnvironment` with its public methods.
- Make `NetworkEnvironment` its default implementation
- Rename `NetworkEnvironment﻿﻿` to `﻿NettyShuffleEnvironment`.

## Verifying this change

Refactoring covered by existing tests.

  - The serializers: (no)

## Documentation


azagrebin:
The PR is based on  #8566 at the moment. Only last 3 commits are relevant.

azagrebin:

flink-runtime/src/main/java/org/apache/flink/runtime/taskexecutor/TaskManagerServicesConfiguration.java
```diff
@@ -78,33 +77,32 @@
...
-			@Nullable QueryableStateConfiguration queryableStateConfig,
-			int numberOfSlots,
-			long configuredMemory,
-			MemoryType memoryType,
-			boolean preAllocateMemory,
-			float memoryFraction,
-			long timerServiceShutdownTimeout,
-			RetryingRegistrationConfiguration retryingRegistrationConfiguration,
-			Optional<Time> systemResourceMetricsProbingInterval) {
+		InetAddress taskManagerAddress,
```
zhijiangW: Keep the previous indentation

flink-runtime/src/main/java/org/apache/flink/runtime/taskexecutor/TaskManagerServicesConfiguration.java
```diff
@@ -78,33 +77,32 @@
...
+		String[] tmpDirPaths,
+		String[] localRecoveryStateRootDirectories,
+		boolean localRecoveryEnabled,
+		@Nullable QueryableStateConfiguration queryableStateConfig,
+		int numberOfSlots,
+		long configuredMemory,
+		MemoryType memoryType,
+		boolean preAllocateMemory,
+		float memoryFraction,
+		int pageSize, long timerServiceShutdownTimeout,
```
zhijiangW: separate line for parameters

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/NetworkEnvironmentConfiguration.java
```diff
@@ -445,36 +442,14 @@ private static NettyConfig createNettyConfig(
 			final InetSocketAddress taskManagerInetSocketAddress = new InetSocketAddress(taskManagerAddress, dataport);
 
 			nettyConfig = new NettyConfig(taskManagerInetSocketAddress.getAddress(), taskManagerInetSocketAddress.getPort(),
-				getPageSize(configuration), ConfigurationParserUtils.getSlot(configuration), configuration);
+				ConfigurationParserUtils.getPageSize(configuration), ConfigurationParserUtils.getSlot(configuration), configuration);
```
zhijiangW: separate line for all the parameters

flink-runtime/src/test/java/org/apache/flink/runtime/state/TaskExecutorLocalStateStoresManagerTest.java
```diff
@@ -205,17 +205,20 @@ private void checkRootDirsClean(File[] rootDirs) {
 	private TaskManagerServicesConfiguration createTaskManagerServiceConfiguration(
 			Configuration config) throws IOException {
 		return TaskManagerServicesConfiguration.fromConfiguration(
-			config, MEM_SIZE_PARAM, InetAddress.getLocalHost(), true);
+			config, InetAddress.getLocalHost());
```
zhijiangW: separate line for every parameter or we do not need break line here because it might not be long after removing above parameters.

flink-runtime/src/test/java/org/apache/flink/runtime/taskexecutor/TaskManagerServicesConfigurationTest.java
```diff
@@ -36,7 +37,7 @@
 	private static final long MEM_SIZE_PARAM = 128L * 1024 * 1024;
 
 	/**
-	 * Verifies that {@link TaskManagerServicesConfiguration#fromConfiguration(Configuration, long, InetAddress, boolean)}
+	 * Verifies that {@link TaskManagerServicesConfiguration#fromConfiguration(Configuration, InetAddress)}
```
zhijiangW: I think this test is only for verifying the generation of `NetworkEnvironmentConfiguration`. The scope of `TaskManagerServicesConfiguration#fromConfiguration` covers many things. Also this class name `TaskManagerServicesConfigurationTest` might also be renamed to `NetworkEnvironmentConfigurationTest` or something else.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NetworkEnvironment.java
```diff
@@ -114,6 +116,22 @@ private NetworkEnvironment(
 		this.isShutdown = false;
 	}
 
+	public static NetworkEnvironment fromConfiguration(
```
zhijiangW: It might be better to put this static factory at the end of this class, also for the following `create` method.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NetworkEnvironment.java
```diff
@@ -129,11 +128,11 @@ public static NetworkEnvironment fromConfiguration(
 			maxJvmHeapMemory,
 			localTaskManagerCommunication,
 			taskManagerAddress);
-		return NetworkEnvironment.create(networkConfig, taskEventPublisher, metricGroup, ioManager);
```
zhijiangW: This should be done in previous hotfix commit while introducing `fromConfiguration` firstly.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NetworkEnvironment.java
```diff
@@ -129,11 +128,11 @@ public static NetworkEnvironment fromConfiguration(
...
+		return create(networkConfig, taskEventPublisher, metricGroup, ioManager);
 	}
 
 	public static NetworkEnvironment create(
-			NetworkEnvironmentConfiguration config,
+		NetworkEnvironmentConfiguration config,
```
zhijiangW: See the changes for commit `[FLINK-11392][network] Introduce ShuffleEnvironment interface`, this indentation should not be changed.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NetworkEnvironment.java
```diff
@@ -322,19 +302,10 @@ private void registerInputMetrics(MetricGroup inputGroup, MetricGroup buffersGro
...
-	 * @return {@code true} if the partition has been updated or {@code false} if the partition is not available anymore.
-	 * @throws IOException IO problem by the update
-	 * @throws InterruptedException potentially blocking operation was interrupted
-	 * @throws IllegalStateException the input gate with the id from the partitionInfo is not found
-	 */
+	@Override
 	public boolean updatePartitionInfo(
-			ExecutionAttemptID consumerID,
-			PartitionInfo partitionInfo) throws IOException, InterruptedException {
+		ExecutionAttemptID consumerID,
```
zhijiangW: ditto: keep previous indentation

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NetworkEnvironment.java
```diff
@@ -409,6 +382,7 @@ public void shutdown() {
 		}
 	}
 
+	@Override
```
zhijiangW: ATM this method is only used for testing, so it seems no reasonable to define an interface method only for testing. If the previous test could be refactored not relying on this method. I ever considered this issue to remove this method.

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/StreamTaskTerminationTest.java
```diff
@@ -150,7 +150,7 @@ public void testConcurrentAsyncCheckpointCannotFailFinishedStreamTask() throws E
 
 		final TaskManagerRuntimeInfo taskManagerRuntimeInfo = new TestingTaskManagerRuntimeInfo();
 
-		final NetworkEnvironment networkEnv = new NetworkEnvironmentBuilder().build();
+		final ShuffleEnvironment networkEnv = new NetworkEnvironmentBuilder().build();
```
zhijiangW: rename `shuffleEnvironment` for `networkEnv`

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/StreamTaskTest.java
```diff
@@ -902,7 +902,7 @@ public static Task createTask(
 		PartitionProducerStateChecker partitionProducerStateChecker = mock(PartitionProducerStateChecker.class);
 		Executor executor = mock(Executor.class);
 
-		NetworkEnvironment network = new NetworkEnvironmentBuilder().build();
+		ShuffleEnvironment network = new NetworkEnvironmentBuilder().build();
```
zhijiangW: ditto: rename `network` as `shuffleEnvironment`

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/SynchronousCheckpointITCase.java
```diff
@@ -295,7 +295,7 @@ private Task createTask(Class<? extends AbstractInvokable> invokableClass) throw
 				0,
 				mock(MemoryManager.class),
 				mock(IOManager.class),
-				networkEnvironment,
+			shuffleEnvironment,
```
zhijiangW: indentation issue here.

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/TaskCheckpointingBehaviourTest.java
```diff
@@ -224,7 +224,7 @@ private static Task createTask(
 				TestStreamTask.class.getName(),
 				taskConfig);
 
-		NetworkEnvironment network = new NetworkEnvironmentBuilder().build();
+		ShuffleEnvironment network = new NetworkEnvironmentBuilder().build();
```
zhijiangW: rename `network` as `shuffleEnvironment`

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/ShuffleEnvironment.java
```diff
@@ -0,0 +1,163 @@
...
+import org.apache.flink.runtime.io.network.partition.PartitionProducerStateProvider;
+import org.apache.flink.runtime.io.network.partition.ResultPartitionID;
+import org.apache.flink.runtime.io.network.partition.consumer.InputGate;
+import org.apache.flink.runtime.taskexecutor.TaskExecutor;
+
+import java.io.IOException;
+import java.util.Collection;
+
+/**
+ * Interface for the implementation of shuffle service locally on task executor.
```
zhijiangW: shuffle service -> shuffle environment

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/ShuffleEnvironment.java
```diff
@@ -0,0 +1,163 @@
...
+ *
+ * <p>Input/Output interface of local shuffle service is based on memory {@link org.apache.flink.runtime.io.network.buffer.Buffer}s.
```
zhijiangW: shuffle service -> shuffle environment

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/ShuffleEnvironment.java
```diff
@@ -0,0 +1,163 @@
...
+ * The task can request next available memory buffers from created here {@link ResultPartitionWriter}s to write shuffle data
```
zhijiangW: remove next?

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/ShuffleEnvironment.java
```diff
@@ -0,0 +1,163 @@
...
+ * and buffers from created here {@link InputGate}s to read it.
```
zhijiangW: Actually the buffers are not requested from `InputGate` by task, and it is done by network thread.
Maybe `and the task can read buffers from created here {@link InputGate}s`.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/ShuffleEnvironment.java
```diff
@@ -0,0 +1,163 @@
...
+ *
+ * <h2>Service lifecycle management.</h2>
```
zhijiangW: Remove `Service` here, only `Lifecycle management`

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/ShuffleEnvironment.java
```diff
@@ -0,0 +1,163 @@
...
+ *     <li>{@code shutdown} is called when the {@link TaskExecutor} is being stopped.</li>
+ * </ol>
+ *
+ * <h2>Shuffle Input/Output management.</h2>
+ *
+ * <h3>Result partition management.</h3>
+ *
+ * <p>The interface implements a factory of result partition writers for the task output: {@code createResultPartitionWriters}.
+ * The created writers are grouped per task and handed over to the task thread upon its startup.
+ * The task is responsible for the writers lifecycle from that moment.
```
zhijiangW: writers -> writers'

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/ShuffleEnvironment.java
```diff
@@ -0,0 +1,163 @@
...
+ *     <li>{@link ResultPartitionWriter#fail(Throwable)} and {@link ResultPartitionWriter#close()} are called
+ *     if the production has failed.</li>
+ *     <li>{@link ResultPartitionWriter#finish()} and {@link ResultPartitionWriter#close()} are called
+ *     if the production is done. The actual release can take some time
+ *     if 'the end of consumption' confirmation is being awaited implicitly
+ *     or the partition is later released by {@code releasePartitions(Collection<ResultPartitionID>)}.</li>
+ *     <li>{@code releasePartitions(Collection<ResultPartitionID>)} is called outside of the task thread,
+ *     e.g. to manage the local resource lifecycle of external partitions which outlive the task production.</li>
+ * </ol>
+ * The partitions, which currently still occupy local resources, can be queried with {@code updatePartitionInfo}.
```
zhijiangW: queried -> updated

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/ShuffleEnvironment.java
```diff
@@ -0,0 +1,163 @@
...
+ *
+ * <h3>Input gate management.</h3>
+ *
+ * <p>The interface implements a factory for the task input gates: {@code createInputGates}.
+ * The created gates are grouped per task and handed over to the task thread upon its startup.
+ * The task is responsible for the gates lifecycle from that moment.
```
zhijiangW: gates -> gates'

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/ShuffleEnvironment.java
```diff
@@ -0,0 +1,163 @@
...
+ *
+ * <p>When tha task is deployed and the input gates are created, it can happen that not all consumed partitions
+ * are known at that moment e.g. because their producers have not been started yet.
+ * Therefore, the {@link ShuffleEnvironment} provides a method {@code updatePartitionInfo} to update them
+ * externally, ouside of the task thread, when the producer becomes known.
+ */
+public interface ShuffleEnvironment {
+
+	/**
+	 * Starts the internal related services upon {@link TaskExecutor}'s startup.
```
zhijiangW: Starts -> Start to keep the same form with the following methods.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/ShuffleEnvironment.java
```diff
@@ -0,0 +1,163 @@
...
+	 * Shutdown the internal related services upon {@link TaskExecutor}'s stop.
+	 */
+	void shutdown();
+
+	boolean isShutdown();
+
+	/**
+	 * Factory method for the task's {@link ResultPartitionWriter}s.
+	 *
+	 * @param taskName string name of the task used for logs
```
zhijiangW: string name of the task -> the task name ?

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/ShuffleEnvironment.java
```diff
@@ -0,0 +1,163 @@
...
+	 * @return a port to connect to the task executor for shuffle data exchange, -1 if only local connection is possible.
+	 */
+	int start() throws IOException;
+
+	/**
+	 * Shutdown the internal related services upon {@link TaskExecutor}'s stop.
+	 */
+	void shutdown();
+
+	boolean isShutdown();
```
zhijiangW: Also add a javadoc for this method to keep consistent with others?

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/ShuffleEnvironment.java
```diff
@@ -0,0 +1,163 @@
...
+	 */
+	ResultPartitionWriter[] createResultPartitionWriters(
+		String taskName,
+		ExecutionAttemptID executionAttemptID,
+		Collection<ResultPartitionDeploymentDescriptor> resultPartitionDeploymentDescriptors,
+		MetricGroup outputGroup,
+		MetricGroup buffersGroup);
+
+	/**
+	 * Batch release local resources occupied with the given partitions.
```
zhijiangW: Release batch of partitions with local resources occupied.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/ShuffleEnvironment.java
```diff
@@ -0,0 +1,163 @@
...
+	 *
+	 * @param partitionIds partitions ids to release
```
zhijiangW: partitions -> partition?

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/ShuffleEnvironment.java
```diff
@@ -0,0 +1,163 @@
...
+	 */
+	Collection<ResultPartitionID> getUnreleasedPartitions();
+
+	/**
+	 * Factory method for the task's {@link InputGate}s.
+	 *
+	 * @param taskName string name of the task used for logs
+	 * @param executionAttemptID execution attempt id of the task
+	 * @param partitionProducerStateProvider producer state provider to query whether the producer is ready for consumption
+	 * @param inputGateDeploymentDescriptors consumed gates descriptors
```
zhijiangW: remove consumed?

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/ShuffleEnvironment.java
```diff
@@ -0,0 +1,163 @@
...
+		PartitionProducerStateProvider partitionProducerStateProvider,
+		Collection<InputGateDeploymentDescriptor> inputGateDeploymentDescriptors,
+		MetricGroup parentGroup,
+		MetricGroup inputGroup,
+		MetricGroup buffersGroup);
+
+	/**
+	 * Update consuming gate with newly available partition, previously unknown.
+	 *
+	 * @param consumerID execution id of consumer to identify belonging to it gate.
```
zhijiangW: `belonging to it gate` seems not readable. 
Maybe `execution id for finding the proper consumer to update the partition`

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/ShuffleEnvironment.java
```diff
@@ -0,0 +1,163 @@
...
+	 * @param partitionInfo partition information to be able to consume it, e.g. network location
```
zhijiangW: to consume it -> to be consumed? Or `partition information for updating`

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/StreamTaskTerminationTest.java
```diff
@@ -150,7 +150,7 @@ public void testConcurrentAsyncCheckpointCannotFailFinishedStreamTask() throws E
 
 		final TaskManagerRuntimeInfo taskManagerRuntimeInfo = new TestingTaskManagerRuntimeInfo();
 
-		final ShuffleEnvironment networkEnv = new NetworkEnvironmentBuilder().build();
+		final ShuffleEnvironment networkEnv = new NettyShuffleEnvironmentBuilder().build();
```
zhijiangW: networkEnv -> shuffleEnvironment?

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/ShuffleEnvironment.java
azagrebin: I think `requests` has similar meaning in this context because task still gets the buffers from gates (network thread is an internal detail how to give buffers to the Task) but I will rephrase it a bit.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/ShuffleEnvironment.java
azagrebin: then available as well if we do not want to provide a lot of details here

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/ShuffleEnvironment.java
azagrebin: actually the method is wrong :) should be `updatePartitionInfo` -> `getUnreleasedPartitions`

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/ShuffleEnvironment.java
azagrebin: I removed it as discussed before

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/ShuffleEnvironment.java
azagrebin: `finding` is what happens internally, but the id is for the gate identity to distinguish gates with the same id from the different consumer executions. I will change the comment.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/ShuffleEnvironment.java
azagrebin: I will change it a bit but I wanted to mention that the info enables partition consumption. I think it is already obvious that it is for the update

azagrebin:
Thanks @zhijiangW , I have addressed comments

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleEnvironment.java
```diff
@@ -0,0 +1,163 @@
...
+	 * Factory method for the task's {@link ResultPartitionWriter}s.
+	 *
+	 * @param taskName the task name, used for logs
+	 * @param executionAttemptID execution attempt id of the task
+	 * @param resultPartitionDeploymentDescriptors descriptors of the partition, produced by the task
+	 * @param outputGroup shuffle specific group for output metrics
+	 * @param buffersGroup shuffle specific group for buffer metrics
+	 * @return array of the task's {@link ResultPartitionWriter}s
+	 */
+	ResultPartitionWriter[] createResultPartitionWriters(
```
zentol: can we include the JobID in here? we may need this for the partition lifecycle tracking, and it's likely also useful for logging purposes.

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleEnvironment.java
```diff
@@ -0,0 +1,163 @@
...
+ * <p>Partitions are released in the following cases:
+ * <ol>
+ *     <li>{@link ResultPartitionWriter#fail(Throwable)} and {@link ResultPartitionWriter#close()} are called
+ *     if the production has failed.</li>
+ *     <li>{@link ResultPartitionWriter#finish()} and {@link ResultPartitionWriter#close()} are called
+ *     if the production is done. The actual release can take some time
+ *     if 'the end of consumption' confirmation is being awaited implicitly
+ *     or the partition is later released by {@code releasePartitions(Collection<ResultPartitionID>)}.</li>
+ *     <li>{@code releasePartitions(Collection<ResultPartitionID>)} is called outside of the task thread,
+ *     e.g. to manage the local resource lifecycle of external partitions which outlive the task production.</li>
```
tillrohrmann: maybe better to say in general `e.g. to manage the lifecycle of BLOCKING result partitions which can outlive their producing tasks`

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleEnvironment.java
```diff
@@ -0,0 +1,163 @@
...
+ * </ol>
+ * The partitions, which currently still occupy local resources, can be queried with {@code getUnreleasedPartitions}.
```
tillrohrmann: `{@link class#method}`

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleEnvironment.java
```diff
@@ -0,0 +1,163 @@
...
+ *
+ * <h3>Input gate management.</h3>
+ *
+ * <p>The interface implements a factory for the task input gates: {@code createInputGates}.
+ * The created gates are grouped per task and handed over to the task thread upon its startup.
+ * The task is responsible for the gates' lifecycle from that moment.
+ *
+ * <p>When tha task is deployed and the input gates are created, it can happen that not all consumed partitions
```
tillrohrmann: typo: `tha` -> `a`

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleEnvironment.java
```diff
@@ -0,0 +1,163 @@
...
+ *     if 'the end of consumption' confirmation is being awaited implicitly
+ *     or the partition is later released by {@code releasePartitions(Collection<ResultPartitionID>)}.</li>
+ *     <li>{@code releasePartitions(Collection<ResultPartitionID>)} is called outside of the task thread,
+ *     e.g. to manage the local resource lifecycle of external partitions which outlive the task production.</li>
+ * </ol>
+ * The partitions, which currently still occupy local resources, can be queried with {@code getUnreleasedPartitions}.
+ *
+ * <h3>Input gate management.</h3>
+ *
+ * <p>The interface implements a factory for the task input gates: {@code createInputGates}.
```
tillrohrmann: `@code` -> `@link`

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleEnvironment.java
```diff
@@ -0,0 +1,163 @@
...
+ * The created gates are grouped per task and handed over to the task thread upon its startup.
+ * The task is responsible for the gates' lifecycle from that moment.
+ *
+ * <p>When tha task is deployed and the input gates are created, it can happen that not all consumed partitions
+ * are known at that moment e.g. because their producers have not been started yet.
+ * Therefore, the {@link ShuffleEnvironment} provides a method {@code updatePartitionInfo} to update them
```
tillrohrmann: `@code` -> `@link`

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleEnvironment.java
```diff
@@ -0,0 +1,163 @@
...
+ * <p>Input/Output interface of local shuffle service environment is based on memory {@link Buffer}s.
+ * The task can write shuffle data into the buffers,
+ * obtained from the created here {@link ResultPartitionWriter}s
+ * and read the buffers from the created here {@link InputGate}s.
+ *
+ * <h2>Lifecycle management.</h2>
+ *
+ * <p>The interface contains method's to manage the lifecycle of the local shuffle service environment:
+ * <ol>
+ *     <li>{@code start} is called when the {@link TaskExecutor} is being started.</li>
```
tillrohrmann: Not sure whether I would include the `TaskExecutor` here. It is a concern of the `TaskExecutor` when this service is started and should not matter to the service implementation. The important bit is that `start` is being called before this service is used.

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleEnvironment.java
```diff
@@ -0,0 +1,163 @@
...
+ *     <li>{@code shutdown} is called when the {@link TaskExecutor} is being stopped.</li>
```
tillrohrmann: Same with `shutdown`

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleEnvironment.java
```diff
@@ -0,0 +1,163 @@
...
+ * <h3>Input gate management.</h3>
+ *
+ * <p>The interface implements a factory for the task input gates: {@code createInputGates}.
+ * The created gates are grouped per task and handed over to the task thread upon its startup.
+ * The task is responsible for the gates' lifecycle from that moment.
+ *
+ * <p>When tha task is deployed and the input gates are created, it can happen that not all consumed partitions
+ * are known at that moment e.g. because their producers have not been started yet.
+ * Therefore, the {@link ShuffleEnvironment} provides a method {@code updatePartitionInfo} to update them
+ * externally, ouside of the task thread, when the producer becomes known.
```
tillrohrmann: Maybe decouple from concrete `task` by saying that the `writer` has to be threadsafe since it might be used from a different thread when being updated.

flink-runtime/src/main/java/org/apache/flink/runtime/taskexecutor/TaskExecutor.java
```diff
@@ -48,7 +48,7 @@
 import org.apache.flink.runtime.highavailability.HighAvailabilityServices;
 import org.apache.flink.runtime.instance.HardwareDescription;
 import org.apache.flink.runtime.instance.InstanceID;
-import org.apache.flink.runtime.io.network.NetworkEnvironment;
+import org.apache.flink.runtime.shuffle.ShuffleEnvironment;
```
zentol: violates checkstyle import order

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleEnvironment.java
```diff
@@ -0,0 +1,163 @@
...
+	 * Start the internal related services upon {@link TaskExecutor}'s startup.
+	 *
+	 * @return a port to connect to the task executor for shuffle data exchange, -1 if only local connection is possible.
+	 */
+	int start() throws IOException;
+
+	/**
+	 * Shutdown the internal related services upon {@link TaskExecutor}'s stop.
+	 */
+	void shutdown();
```
tillrohrmann: What about letting `ShuffleEnvironment` extend `AutoCloseable`?

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleEnvironment.java
```diff
@@ -0,0 +1,163 @@
...
+ * externally, ouside of the task thread, when the producer becomes known.
+ */
+public interface ShuffleEnvironment {
+
+	/**
+	 * Start the internal related services upon {@link TaskExecutor}'s startup.
+	 *
+	 * @return a port to connect to the task executor for shuffle data exchange, -1 if only local connection is possible.
+	 */
+	int start() throws IOException;
```
tillrohrmann: The port is specific to the `NettyShuffleEnvironment` implementation. Maybe in the future, one might abstract this detail behind some `ShuffleConnectionInfo`.

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleEnvironment.java
```diff
@@ -0,0 +1,163 @@
...
+
+	/**
+	 * Shutdown the internal related services upon {@link TaskExecutor}'s stop.
+	 */
+	void shutdown();
+
+	/**
+	 * Factory method for the task's {@link ResultPartitionWriter}s.
```
tillrohrmann: No need to include the `task` here. Other components could also try to write into a `partition`.

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleEnvironment.java
```diff
@@ -0,0 +1,163 @@
...
+	 *
+	 * @param taskName the task name, used for logs
+	 * @param executionAttemptID execution attempt id of the task
+	 * @param resultPartitionDeploymentDescriptors descriptors of the partition, produced by the task
+	 * @param outputGroup shuffle specific group for output metrics
+	 * @param buffersGroup shuffle specific group for buffer metrics
+	 * @return array of the task's {@link ResultPartitionWriter}s
+	 */
+	ResultPartitionWriter[] createResultPartitionWriters(
+		String taskName,
```
tillrohrmann: maybe rename into `owner`

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleEnvironment.java
```diff
@@ -0,0 +1,163 @@
...
+		ExecutionAttemptID executionAttemptID,
+		Collection<ResultPartitionDeploymentDescriptor> resultPartitionDeploymentDescriptors,
+		MetricGroup outputGroup,
+		MetricGroup buffersGroup);
+
+	/**
+	 * Release local resources occupied with the given partitions.
```
tillrohrmann: `occupied by`

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleEnvironment.java
```diff
@@ -0,0 +1,163 @@
...
+	 *
+	 * @param partitionIds partition ids to release
```
tillrohrmann: `@param partitionIds identifying the partitions to be released`

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleEnvironment.java
```diff
@@ -0,0 +1,163 @@
...
+	 */
+	void releasePartitions(Collection<ResultPartitionID> partitionIds);
+
+	/**
+	 * Report unreleased partitions.
+	 *
+	 * @return collection of partitions which still occupy some resources locally on this task executor
```
tillrohrmann: we can remove `on this task executor`

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleEnvironment.java
```diff
@@ -0,0 +1,163 @@
...
+	 * and have been not released yet.
```
tillrohrmann: `have not been released`

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleEnvironment.java
```diff
@@ -0,0 +1,163 @@
...
+	 */
+	Collection<ResultPartitionID> getUnreleasedPartitions();
```
tillrohrmann: Maybe better to call this method `getLocalPartitions` or `getPartitionsOccupyingLocalResources`. In the case of an external shuffle service, this method would return an empty collection even though some partitions might still be unreleased but stored somewhere else.

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleEnvironment.java
```diff
@@ -0,0 +1,163 @@
...
+
+	/**
+	 * Factory method for the task's {@link InputGate}s.
```
tillrohrmann: Factory method for `{@link InputGate InputGates}` to consume result partitions.

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleEnvironment.java
```diff
@@ -0,0 +1,163 @@
...
+	 *
+	 * @param taskName the task name, used for logs
+	 * @param executionAttemptID execution attempt id of the task
+	 * @param partitionProducerStateProvider producer state provider to query whether the producer is ready for consumption
+	 * @param inputGateDeploymentDescriptors descriptors of the input gates, consumed by the task
+	 * @param parentGroup parent of shuffle specific metric group
+	 * @param inputGroup shuffle specific group for input metrics
+	 * @param buffersGroup shuffle specific group for buffer metrics
+	 * @return array of the task's {@link InputGate}s
```
tillrohrmann: `{@link A B}`

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleEnvironment.java
```diff
@@ -0,0 +1,163 @@
...
+	 */
+	InputGate[] createInputGates(
+		String taskName,
```
tillrohrmann: maybe `owner`

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleEnvironment.java
```diff
@@ -0,0 +1,163 @@
...
+	 * @param taskName the task name, used for logs
+	 * @param executionAttemptID execution attempt id of the task
+	 * @param partitionProducerStateProvider producer state provider to query whether the producer is ready for consumption
+	 * @param inputGateDeploymentDescriptors descriptors of the input gates, consumed by the task
+	 * @param parentGroup parent of shuffle specific metric group
+	 * @param inputGroup shuffle specific group for input metrics
+	 * @param buffersGroup shuffle specific group for buffer metrics
+	 * @return array of the task's {@link InputGate}s
+	 */
+	InputGate[] createInputGates(
```
tillrohrmann: Why do we return an array here instead of a `Collection<InputGate>`? Is it because of the order of the input gates? Maybe we could state that the returned collection must have the same order as the `inputGateDeploymentDescriptors` collection.

flink-runtime/src/main/java/org/apache/flink/runtime/taskexecutor/TaskExecutor.java
```diff
@@ -271,7 +271,7 @@ public TaskExecutor(
 	@Override
 	public CompletableFuture<Boolean> canBeReleased() {
 		return CompletableFuture.completedFuture(
-			taskExecutorServices.getNetworkEnvironment().getUnreleasedPartitions().isEmpty());
+			taskExecutorServices.getShuffleEnvironment().getUnreleasedPartitions().isEmpty());
```
tillrohrmann: we could directly use the `shuffleEnvironment` field here.

flink-runtime/src/main/java/org/apache/flink/runtime/taskexecutor/TaskManagerRunner.java
```diff
@@ -367,12 +365,14 @@ public static TaskExecutor startTaskManager(
...
 		TaskManagerServices taskManagerServices = TaskManagerServices.fromConfiguration(
+			configuration,
 			taskManagerServicesConfiguration,
 			taskManagerMetricGroup.f1,
 			resourceID,
 			rpcService.getExecutor(), // TODO replace this later with some dedicated executor for io.
 			EnvironmentInformation.getSizeOfFreeHeapMemoryWithDefrag(),
-			EnvironmentInformation.getMaxJvmHeapMemory());
+			EnvironmentInformation.getMaxJvmHeapMemory(),
+			localCommunicationOnly);
```
tillrohrmann: Why did you change this? The intention was that `TaskManagerServicesConfiguration` contains all configuration values necessary to instantiate the `TaskManagerServices`.

flink-runtime/src/main/java/org/apache/flink/runtime/taskexecutor/TaskManagerRunner.java
```diff
@@ -367,12 +365,14 @@ public static TaskExecutor startTaskManager(
 			taskManagerServicesConfiguration.getSystemResourceMetricsProbingInterval());
 
 		TaskManagerServices taskManagerServices = TaskManagerServices.fromConfiguration(
+			configuration,
```
tillrohrmann: I deliberately created the `TaskManagerServicesConfiguration` in order to not having to pass in the `Configuration` object to this method.

flink-runtime/src/main/java/org/apache/flink/runtime/taskexecutor/TaskManagerServices.java
```diff
@@ -221,22 +222,27 @@ public void shutDown() throws FlinkException {
...
 	 * @param taskIOExecutor executor for async IO operations
 	 * @param freeHeapMemoryWithDefrag an estimate of the size of the free heap memory
 	 * @param maxJvmHeapMemory the maximum JVM heap size
+	 * @param localCommunicationOnly True, to skip initializing the network stack.
+	 *                               Use only in cases where only one task manager runs.
 	 * @return task manager components
 	 * @throws Exception
 	 */
 	public static TaskManagerServices fromConfiguration(
+			Configuration configuration,
```
tillrohrmann: Can't we pass in the `Configuration` as part of the `TaskManagerServicesConfiguration`?

flink-runtime/src/main/java/org/apache/flink/runtime/taskexecutor/TaskManagerServices.java
```diff
@@ -221,22 +222,27 @@ public void shutDown() throws FlinkException {
...
 			TaskManagerServicesConfiguration taskManagerServicesConfiguration,
 			MetricGroup taskManagerMetricGroup,
 			ResourceID resourceID,
 			Executor taskIOExecutor,
 			long freeHeapMemoryWithDefrag,
-			long maxJvmHeapMemory) throws Exception {
+			long maxJvmHeapMemory,
+			boolean localCommunicationOnly) throws Exception {
```
tillrohrmann: I think whether it's local or not could be part of the `TaskManagerServicesConfiguration`.

flink-runtime/src/main/java/org/apache/flink/runtime/taskexecutor/TaskManagerServices.java
```diff
@@ -246,13 +252,16 @@ public static TaskManagerServices fromConfiguration(
 		// start the I/O manager, it will create some temp directories.
 		final IOManager ioManager = new IOManagerAsync(taskManagerServicesConfiguration.getTmpDirPaths());
 
-		final NetworkEnvironment network = NetworkEnvironment.create(
+		final ShuffleEnvironment shuffleEnvironment = NettyShuffleEnvironment.fromConfiguration(
```
tillrohrmann: Can't we keep the `NettyShuffleEnvironmentConfiguration` part of the `TaskManagerServicesConfiguration`? I think in the future we will need a `ShuffleEnvironmentConfiguration` which we pass to the `ShuffleEnvironmentFactory` once the `ShuffleEnvironment` is pluggable.

flink-runtime/src/main/java/org/apache/flink/runtime/util/ConfigurationParserUtils.java
```diff
@@ -131,4 +134,26 @@ public static void checkConfigParameter(boolean condition, Object parameter, Str
...
+			configuration.getString(TaskManagerOptions.MEMORY_SEGMENT_SIZE)).getBytes());
+
+		// check page size of for minimum size
+		checkConfigParameter(pageSize >= MemoryManager.MIN_PAGE_SIZE, pageSize,
+			TaskManagerOptions.MEMORY_SEGMENT_SIZE.key(),
+			"Minimum memory segment size is " + MemoryManager.MIN_PAGE_SIZE);
+		// check page size for power of two
+		checkConfigParameter(MathUtils.isPowerOf2(pageSize), pageSize,
+			TaskManagerOptions.MEMORY_SEGMENT_SIZE.key(),
+			"Memory segment size must be a power of 2.");
```
tillrohrmann: Formatting in this method seems a bit off.

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/NettyShuffleEnvironmentTest.java
```diff
@@ -47,10 +48,10 @@
 import static org.powermock.api.mockito.PowerMockito.spy;
 
 /**
- * Various tests for the {@link NetworkEnvironment} class.
+ * Various tests for the {@link NettyShuffleEnvironment} class.
  */
 @RunWith(Parameterized.class)
-public class NetworkEnvironmentTest {
+public class NettyShuffleEnvironmentTest extends TestLogger {
```
tillrohrmann: 👍

flink-runtime/src/test/java/org/apache/flink/runtime/taskexecutor/NettyShuffleEnvironmentConfigurationTest.java
```diff
@@ -0,0 +1,334 @@
...
+import java.net.InetAddress;
+import java.util.Random;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+/**
+ * Unit test for {@link TaskManagerServices}.
```
tillrohrmann: Comment seems out dated.

flink-core/src/main/java/org/apache/flink/configuration/NettyShuffleEnvironmentOptions.java
```diff
@@ -29,7 +29,7 @@
  */
 @PublicEvolving
 @ConfigGroups(groups = @ConfigGroup(name = "NetworkNetty", keyPrefix = "taskmanager.network.netty"))
-public class NetworkEnvironmentOptions {
+public class NettyShuffleEnvironmentOptions {
```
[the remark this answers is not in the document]
azagrebin: The class was introduced recently. The options belong to `TaskManagerOptions` before 1.9 and have been deprecated to become part of netty shuffle configuration group.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NettyShuffleEnvironment.java
```diff
@@ -228,6 +227,7 @@ public void releasePartitions(Collection<ResultPartitionID> partitionIds) {
 	 * @return collection of partitions which still occupy some resources locally on this task executor
 	 * and have been not released yet.
 	 */
```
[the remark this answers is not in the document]
azagrebin: true, leftover

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleEnvironment.java
```diff
@@ -0,0 +1,163 @@
...
+import org.apache.flink.runtime.taskexecutor.TaskExecutor;
+
+import java.io.IOException;
+import java.util.Collection;
+
+/**
+ * Interface for the implementation of shuffle service environment locally on task executor.
+ *
+ * <p>Input/Output interface of local shuffle service environment is based on memory {@link Buffer}s.
+ * The task can write shuffle data into the buffers,
```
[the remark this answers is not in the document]
azagrebin: It is more about keeping some complete logical statements on different lines.
I think it improves readability.

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleEnvironment.java
```diff
@@ -0,0 +1,163 @@
...
+ *
+ * <h3>Result partition management.</h3>
+ *
+ * <p>The interface implements a factory of result partition writers for the task output: {@code createResultPartitionWriters}.
+ * The created writers are grouped per task and handed over to the task thread upon its startup.
+ * The task is responsible for the writers' lifecycle from that moment.
+ *
+ * <p>Partitions are released in the following cases:
+ * <ol>
+ *     <li>{@link ResultPartitionWriter#fail(Throwable)} and {@link ResultPartitionWriter#close()} are called
```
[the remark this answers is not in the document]
azagrebin: There is a scenario where `Task.TaskCanceler` abruptly closes everything (an old fix to interrupt some Kafka thread see https://github.com/apache/flink/pull/1780/files#diff-b61470a664430bb162ac2c24c9272588R1147). When `Task.run` exits it still have to call either `fail` (to release) or `finish` and then `close`. We thought in #8133 that it is more explicit to require `close` after `fail/finish` as we need `close` separately anyways.

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleEnvironment.java
```diff
@@ -0,0 +1,163 @@
...
+ *     if the production has failed.</li>
+ *     <li>{@link ResultPartitionWriter#finish()} and {@link ResultPartitionWriter#close()} are called
+ *     if the production is done. The actual release can take some time
+ *     if 'the end of consumption' confirmation is being awaited implicitly
```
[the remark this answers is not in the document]
azagrebin: After #7938, it is the case, afair, because it needed more refactoring to send confirmation when just all the data has been received on the consumer side but not processed.

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleEnvironment.java
azagrebin: agreed

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleEnvironment.java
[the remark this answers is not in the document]
tillrohrmann: I don't fully understand your explanation. Is there a scenario in which you don't call `finish` or `fail` but `close`? Said in other words, is there a semantic difference between `fail`/`finish` and `close`? If yes, what is the difference?

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleEnvironment.java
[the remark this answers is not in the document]
tillrohrmann: I'm not convinced about this. Especially since it only a subordinate clause.

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleEnvironment.java
```diff
@@ -0,0 +1,211 @@
...
+import java.net.InetAddress;
+import java.util.Collection;
+
+import static org.apache.flink.util.Preconditions.checkNotNull;
+
+/**
+ * Interface for the implementation of shuffle service local environment.
+ *
+ * <p>Input/Output interface of local shuffle service environment is based on memory {@link Buffer Buffers}.
+ * A producer can write shuffle data into the buffers,
```
tillrohrmann: I'm not convinced that putting subordinate clauses on separate lines increase readability. If it does, then we break this convention later in the docs by putting multiple sentences on the same line.

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleEnvironment.java
```diff
@@ -0,0 +1,211 @@
...
+ * obtained from the created here {@link ResultPartitionWriter ResultPartitionWriters}
```
tillrohrmann: `here` seems wrong

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleEnvironment.java
```diff
@@ -0,0 +1,211 @@
...
+ * and a consumer read the buffers from the created here {@link InputGate InputGates}.
```
tillrohrmann: `read` -> `reads` and `here` seems wrong

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleEnvironment.java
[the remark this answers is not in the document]
tillrohrmann: Ok, now I got it. You are talking here about the `ResultPartition` and not the `ResultPartitionWriter`. This makes sense now.

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleEnvironment.java
azagrebin: true, will change to collection and add the comment about the order

flink-runtime/src/main/java/org/apache/flink/runtime/taskexecutor/TaskManagerRunner.java
azagrebin: agreed, will put all configuration data into `TaskManagerServicesConfiguration`

flink-runtime/src/main/java/org/apache/flink/runtime/taskexecutor/TaskManagerServices.java
azagrebin: `NettyShuffleEnvironmentConfiguration` is netty specific. I will add `ShuffleEnvironmentContext` to generate it from `TaskManagerServicesConfiguration` and pass to the factory.

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleEnvironment.java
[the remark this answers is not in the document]
azagrebin: At the moment, `close` is also called before `fail/finish` to abruptly trigger production cancelation from `Task.TaskCanceler.  `fail/finish` and `close` are always called on exit from `Task.run`.  `fail/finish` signal end of production and `close` (it happens so now) signals abrupt cancel and/or final cleanup.

As a followup, we can consider refactoring it to `ResultPartitionWriter.triggerCancelation` and forcing `close` to do `fail` and full release/cleanup if `fail/finish` have not been called before `close`.

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleEnvironment.java
azagrebin: ok, let's stick to keeping similar line lengths as in other classes.

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleEnvironment.java
```diff
@@ -0,0 +1,235 @@
...
+				MetricGroup parentMetricGroup,
+				IOManager ioManager) {
+			this.configuration = checkNotNull(configuration);
+			this.location = checkNotNull(location);
+			this.maxJvmHeapMemory = maxJvmHeapMemory;
+			this.localCommunicationOnly = localCommunicationOnly;
+			this.hostAddress = checkNotNull(hostAddress);
+			this.eventPublisher = eventPublisher;
+			this.parentMetricGroup = parentMetricGroup;
+			this.ioManager = ioManager;
```
tillrohrmann: Why not doing the `checkNotNull` for the last three parameters?

flink-runtime/src/main/java/org/apache/flink/runtime/taskexecutor/TaskManagerServicesConfiguration.java
```diff
@@ -91,10 +107,15 @@ public TaskManagerServicesConfiguration(
 			long timerServiceShutdownTimeout,
 			RetryingRegistrationConfiguration retryingRegistrationConfiguration,
 			Optional<Time> systemResourceMetricsProbingInterval) {
+		this.configuration = configuration;
+		this.resourceID = resourceID;
```
tillrohrmann: `checkNotNull` missing for these parameters.

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleEnvironment.java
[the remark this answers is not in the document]
azagrebin: I also added comment to `ResultPartitionWriter` to explain this in more details.

azagrebin:
Thanks for the review @tillrohrmann, I've pushed the final version with the last changes including yours.

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleEnvironmentContext.java
```diff
@@ -0,0 +1,94 @@
...
+
+	public TaskEventPublisher getEventPublisher() {
+		return eventPublisher;
+	}
+
+	public MetricGroup getParentMetricGroup() {
+		return parentMetricGroup;
+	}
+
+	public IOManager getIoManager() {
```
zhijiangW: May be `getIOManager`, upper case for both I and O

flink-runtime/src/main/java/org/apache/flink/runtime/taskexecutor/TaskManagerServicesConfiguration.java
```diff
@@ -196,17 +236,22 @@ RetryingRegistrationConfiguration getRetryingRegistrationConfiguration() {
...
 	 * @param remoteAddress identifying the IP address under which the TaskManager will be accessible
-	 * @param localCommunication True, to skip initializing the network stack.
-	 *                                      Use only in cases where only one task manager runs.
-	 * @return TaskExecutorConfiguration that wrappers InstanceConnectionInfo, NetworkEnvironmentConfiguration, etc.
+	 * @param freeHeapMemoryWithDefrag an estimate of the size of the free heap memory
+	 * @param maxJvmHeapMemory the maximum JVM heap size
+	 * @param localCommunicationOnly True if only local communication is possible.
+	 *                               Use only in cases where only one task manager runs.
+	 *
+	 * @return TaskExecutorConfiguration that wrappers InstanceConnectionInfo, etc.
```
zhijiangW: return `TaskManagerServicesConfiguration`

flink-runtime/src/main/java/org/apache/flink/runtime/taskexecutor/TaskManagerServicesConfiguration.java
```diff
@@ -160,6 +188,14 @@ MemoryType getMemoryType() {
 		return memoryType;
 	}
 
+	public long getFreeHeapMemoryWithDefrag() {
```
zhijiangW: package private

flink-runtime/src/main/java/org/apache/flink/runtime/taskexecutor/TaskManagerServicesConfiguration.java
```diff
@@ -160,6 +188,14 @@ MemoryType getMemoryType() {
...
+		return freeHeapMemoryWithDefrag;
+	}
+
+	public long getMaxJvmHeapMemory() {
```
zhijiangW: ditto: package private

flink-runtime/src/main/java/org/apache/flink/runtime/taskexecutor/TaskManagerServicesConfiguration.java
```diff
@@ -118,10 +138,22 @@ public TaskManagerServicesConfiguration(
...
+
+	public ResourceID getResourceID() {
+		return resourceID;
+	}
+
 	InetAddress getTaskManagerAddress() {
 		return taskManagerAddress;
 	}
 
+	public boolean isLocalCommunicationOnly() {
```
zhijiangW: package private

zhijiangW:
@azagrebin @tillrohrmann sorry for leaving several nit comments at the point of merging this PR.
My above comments are not critical and might be addressed conveniently future.

azagrebin:
Thanks @zhijiangW , addressed

tillrohrmann:
Thanks for the review @zhijiangW and the good work @azagrebin. Merging now.

PR #874, opened 2015-06-30
[FLINK-2297] [ml] Add threshold setting for SVM binary predictions

thvasilo:
Currently SVM outputs the raw decision function values when using the predict function.

We should have instead the ability to set a threshold above which examples are labeled as positive (1.0) and below negative (-1.0). Then the prediction function can be directly used for evaluation.

This provides that functionality, as well as the ability to provide the raw decision function values.

PR #8609, opened 2019-06-04
[FLINK-12541][container][python] Add support for Python jobs in build script

dianfu:

*This pull request adds support to build job specific docker image for Python Table API jobs.*

## Brief change log

  - *Improves the build script to support build a Python job specific docker image*

## Verifying this change

Verify this change manually.

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-container/docker/README.md
+++ b/flink-container/docker/README.md
@@ -17,25 +17,23 @@ Assume that the job jar is stored under `<COMMA_SEPARATED_PATH_TO_JOB_ARTIFACTS>
 
 If you want to build the Flink image from the version you have checked out locally run:
 
-    build.sh --from-local-dist --job-artifacts <COMMA_SEPARATED_PATH_TO_JOB_ARTIFACTS> [--with-python2|--with-python3] [--opt-jars <COMMA_SEPARATED_OPT_JARS>] --image-name <IMAGE_NAME>
+    build.sh --from-local-dist --job-artifacts <COMMA_SEPARATED_PATH_TO_JOB_ARTIFACTS> [--with-python2|--with-python3] --image-name <IMAGE_NAME>
     
 Note that you first need to call `mvn package -pl flink-dist -am` to build the Flink binaries.
 
 If you want to build the Flink image from an archive stored under `<PATH_TO_ARCHIVE>` run:
 
-    build.sh --from-archive <PATH_TO_ARCHIVE> --job-artifacts <COMMA_SEPARATED_PATH_TO_JOB_ARTIFACTS> [--with-python2|--with-python3] [--opt-jars <COMMA_SEPARATED_OPT_JARS>] --image-name <IMAGE_NAME>
+    build.sh --from-archive <PATH_TO_ARCHIVE> --job-artifacts <COMMA_SEPARATED_PATH_TO_JOB_ARTIFACTS> [--with-python2|--with-python3] --image-name <IMAGE_NAME>
 
 If you want to build the Flink image for a specific version of Flink/Hadoop/Scala run:
 
-    build.sh --from-release --flink-version 1.6.0 --hadoop-version 2.8 --scala-version 2.11 --job-artifacts <COMMA_SEPARATED_PATH_TO_JOB_ARTIFACTS> [--with-python2|--with-python3] [--opt-jars <COMMA_SEPARATED_OPT_JARS>] --image-name <IMAGE_NAME>
+    build.sh --from-release --flink-version 1.6.0 --hadoop-version 2.8 --scala-version 2.11 --job-artifacts <COMMA_SEPARATED_PATH_TO_JOB_ARTIFACTS> [--with-python2|--with-python3] --image-name <IMAGE_NAME>
 
 Please note that from Flink-1.8, hadoop version is optional and you could build the Flink image without providing any hadoop version.
     
 The script will try to download the released version from the Apache archive.
 
 The artifacts specified in <COMMA_SEPARATED_PATH_TO_JOB_ARTIFACTS> will be copied to directory /opt/artifacts of the built image.
-
-<COMMA_SEPARATED_OPT_JARS> specifies the optional jars to be used in directory <FLINK_HOME>/opt/. For example, `--opt-jars table` can be specified if the table jar is needed.
 
 ## Deploying via Docker compose
 
--- a/flink-container/docker/README.md
+++ b/flink-container/docker/README.md
@@ -13,7 +13,7 @@ Install the most recent stable version of [Docker](https://docs.docker.com/insta
 Images are based on the official Java Alpine (OpenJDK 8) image.
 
 Before building the image, one needs to build the user code jars for the job.
-Assume that the job jar is stored under `<COMMA_SEPARATED_PATH_TO_JOB_ARTIFACTS>`
+A Flink job can consist of multiple artifacts. In order to specify the required artifacts, they need to be passed to `--job-artifacts` of the build script. The individual paths are comma separated.
 
 If you want to build the Flink image from the version you have checked out locally run:
 
--- a/flink-container/docker/README.md
+++ b/flink-container/docker/README.md
@@ -13,25 +13,29 @@ Install the most recent stable version of [Docker](https://docs.docker.com/insta
 Images are based on the official Java Alpine (OpenJDK 8) image.
 
 Before building the image, one needs to build the user code jars for the job.
-Assume that the job jar is stored under `<PATH_TO_JOB_JAR>` 
+Assume that the job jar is stored under `<COMMA_SEPARATED_PATH_TO_JOB_ARTIFACTS>`
 
 If you want to build the Flink image from the version you have checked out locally run:
 
-    build.sh --from-local-dist --job-jar <PATH_TO_JOB_JAR> --image-name <IMAGE_NAME>
+    build.sh --from-local-dist --job-artifacts <COMMA_SEPARATED_PATH_TO_JOB_ARTIFACTS> [--with-python2|--with-python3] [--opt-jars <COMMA_SEPARATED_OPT_JARS>] --image-name <IMAGE_NAME>
     
 Note that you first need to call `mvn package -pl flink-dist -am` to build the Flink binaries.
 
 If you want to build the Flink image from an archive stored under `<PATH_TO_ARCHIVE>` run:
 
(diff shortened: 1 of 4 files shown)


sunjincheng121:

flink-container/src/main/java/org/apache/flink/container/entrypoint/ClassPathJobGraphRetriever.java
```diff
@@ -73,27 +76,31 @@
 			@Nonnull JobID jobId,
 			@Nonnull SavepointRestoreSettings savepointRestoreSettings,
 			@Nonnull String[] programArguments,
-			@Nullable String jobClassName) {
-		this(jobId, savepointRestoreSettings, programArguments, jobClassName, JarsOnClassPath.INSTANCE);
+			@Nullable String jobEntryPointName,
+			@Nullable String jobPythonArtifacts) {
+		this(jobId, savepointRestoreSettings, programArguments, jobEntryPointName, jobPythonArtifacts, JarsOnClassPath.INSTANCE);
```
tillrohrmann: Why do we need to touch the `ClassPathJobGraphRetriever` at all? This class should not need to know about Python. Otherwise we need to add special casing for all supported languages in the future. This does not seem right.

flink-container/src/main/java/org/apache/flink/container/entrypoint/ClassPathJobGraphRetriever.java
```diff
@@ -110,19 +117,38 @@ public JobGraph retrieveJobGraph(Configuration configuration) throws FlinkExcept
...
 		try {
 			final Class<?> mainClass = getClass().getClassLoader().loadClass(entryClass);
-			return new PackagedProgram(mainClass, programArguments);
-		} catch (ClassNotFoundException | ProgramInvocationException e) {
+			return new PackagedProgram(mainClass, getProgramArguments(isPythonProgram));
+		} catch (ProgramInvocationException e) {
 			throw new FlinkException("Could not load the provided entrypoint class.", e);
+		} catch (ClassNotFoundException e) {
+			if (!isPythonProgram && jobPythonArtifacts != null) {
+				return createPackagedProgram("org.apache.flink.python.client.PythonDriver", true);
```
tillrohrmann: This is overly complicated and error prone. Why not starting the cluster entrypoint with `o.a.f.python.client.PythonDriver` as `jobClassName`?

flink-container/src/main/java/org/apache/flink/container/entrypoint/ClassPathJobGraphRetriever.java
```diff
@@ -110,19 +117,38 @@ public JobGraph retrieveJobGraph(Configuration configuration) throws FlinkExcept
...
+
+	private String[] getProgramArguments(final boolean isPythonProgram) throws FlinkException {
+		if (isPythonProgram) {
+			final String[] args = new String[programArguments.length + 4];
+			args[0] = "pym";
+			args[1] = getJobEntryPointNameOrScanClassPath();
+			args[2] = "pyfs";
+			args[3] = jobPythonArtifacts;
+			System.arraycopy(programArguments, 0, args, 4, programArguments.length);
+			return args;
```
tillrohrmann: Why do we need to do this command line argument magic? Just because `PythonDriver` expects the arguments to be passed in a special order? Why not making the `PythonDriver` more flexible.

flink-container/src/main/java/org/apache/flink/container/entrypoint/StandaloneJobClusterConfiguration.java
```diff
@@ -69,7 +74,12 @@ JobID getJobId() {
...
 
 	@Nullable
-	String getJobClassName() {
-		return jobClassName;
+	String getJobEntryPointName() {
+		return jobEntryPointName;
+	}
+
+	@Nullable
+	String getJobPythonArtifacts() {
```
tillrohrmann: Why is this needed?

flink-container/src/main/java/org/apache/flink/container/entrypoint/StandaloneJobClusterConfigurationParserFactory.java
```diff
@@ -45,12 +45,29 @@
...
+
+	private static final Option JOB_ENTRYPOINT_NAME_OPTION = Option.builder("j")
+		.longOpt("job-entrypoint")
+		.required(false)
+		.hasArg(true)
+		.argName("job entrypoint name")
+		.desc("Entrypoint name of the job to run.")
+		.build();
+
+	private static final Option JOB_PYTHON_ARTIFACTS_OPTION = Option.builder("ja")
```
tillrohrmann: Why does this need to be handled by the `StandaloenJobClusterConfiguration` and cannot be parsed by the `PythonDriver` program?

flink-container/src/main/java/org/apache/flink/container/entrypoint/StandaloneJobClusterEntryPoint.java
```diff
@@ -53,26 +53,31 @@
 	private final String[] programArguments;
 
 	@Nullable
-	private final String jobClassName;
+	private final String jobEntryPointName;
+
+	@Nullable
+	private final String jobPythonArtifacts;
```
tillrohrmann: I think it is not a good idea to add special case logic into the `StandaloneJobClusterEntrypoint`. Why do you think this is needed?

tillrohrmann:
Why has this PR the same Flink issue assigned as #8532? Every PR should have it's own JIRA issue as stated in the [contribution guidelines](https://flink.apache.org/contribute-code.html).

sunjincheng121:
Thanks for the PR @dianfu 
I think the suggestion from @tillrohrmann makes more sense to me. we should let the PR have its own JIRA.
furthermore,  the improvement of `pyfs` also can in a new PR (with a new JIRA).
What do you think?

dianfu:
@tillrohrmann Thanks a lot for your review. 
Your suggestion makes much sense to me. I have created a dedicated JIRA [FLINK-12788](https://issues.apache.org/jira/browse/FLINK-12788) for this PR. Regarding to the changes to `StandaloneJobClusterEntrypoint`, agree that there should not be special logic for Python. I will revert that part of changes.

@sunjincheng121 Thanks a lot for your review. I have created a separate JIRA [FLINK-12787](https://issues.apache.org/jira/browse/FLINK-12787) for the PythonDriver improving. Then we can focus this PR on the build script changes for Python jobs support. 

Will updated the PR later today.

dianfu:
@tillrohrmann @sunjincheng121 I have updated the PR and the changes are only related to the build script for Python jobs support. Looking forward to your feedback.

sunjincheng121:
Thanks for the update @dianfu!

Currently, the PR looks pretty clean form my side. only one improvement I am not pretty sure that is build command option `--opt-jars`.  this way the opt jar does not need upload. but without this option, the user can build a fat jar with option jars. So this change is an improvement(not necessary change) So we may need some opinion from @tillrohrmann!
Otherwise the PR LGTM. I'll merge it when @tillrohrmann says ok!

Best,
Jincheng

sunjincheng121:
I think we can remove the `--opt-jars` due to we @StephanEwen already bring up the discussion for put the table JAR into lib. detail can be found here:  http://apache-flink-mailing-list-archive.1008284.n3.nabble.com/DISCUSS-Putting-Table-API-jars-in-lib-by-default-td29492.html
What do you think?

dianfu:
@sunjincheng121 Make sense to me as `--opt-jars` is not a must have option. Considering that the table JAR will be put into lib by default in the future, removing this option is good for me as the table JAR is the motivation for this option. Have updated the PR. Looking forward to your response.

flink-container/kubernetes/README.md
```diff
@@ -29,7 +29,7 @@ In non HA mode, you should first start the job cluster service:
 
 In order to deploy the job cluster entrypoint run:
 
-`FLINK_IMAGE_NAME=<IMAGE_NAME> FLINK_JOB=<JOB_NAME> FLINK_JOB_PARALLELISM=<PARALLELISM> envsubst < job-cluster-job.yaml.template | kubectl create -f -`
```
tillrohrmann: Why did we remove `FLINK_JOB` here?

flink-container/docker/README.md
```diff
@@ -13,26 +13,28 @@ Install the most recent stable version of [Docker](https://docs.docker.com/insta
 Images are based on the official Java Alpine (OpenJDK 8) image.
 
 Before building the image, one needs to build the user code jars for the job.
-Assume that the job jar is stored under `<PATH_TO_JOB_JAR>` 
+Assume that the job jar is stored under `<COMMA_SEPARATED_PATH_TO_JOB_ARTIFACTS>`
```
tillrohrmann: Seems not to fit. I guess the sentence needs to be changed to something like: `A Flink job can consist of multiple jars. In order to specify the required jars, they need to be passed to `--job-artifacts` of the build script. The individual paths are comma separated.

flink-container/kubernetes/README.md
dianfu: I found that FLINK_JOB has been removed in commit [753e0c6](https://github.com/dianfu/flink/commit/753e0c618a30a19954761cd75fadba4a89836ae1) and just correct the documentation here. What do you think?

flink-container/docker/README.md
dianfu: +1. Only one concern: what about changing `jars` to `artifacts`?

dianfu:
@tillrohrmann Thanks a lot for your comments. I have updated the PR and addressed one of your concerns. Regarding to FLINK_JOB, this option has been removed and the documentation has not been updated. Just correct the documentation here. Please let me know if you feel that change doesn't make sense or you prefer to address that in another PR.

flink-container/kubernetes/README.md
tillrohrmann: Good catch. Now this makes sense.

knaufk:
FYI: This change broke `test_docker_embedded_job.sh` and `test_kubernetes_embedded_job.sh` as they were still using the `--job-jar` argument. I have included a `[hotfix]` in https://github.com/apache/flink/pull/874.

dianfu:
@knaufk Thanks a lot for the hotfix. The fix makes sense to me.

PR #8610, opened 2019-06-04
[FLINK-12715][hive] Hive-1.2.1 build is broken

lirui-apache:

To fix hive-1.2.1 build. The build was broken because some util method used in FLINK-12568 is not compatible in Hive-2.3.4 and Hive-1.2.1.

## Brief change log

  - Added shim for `FileUtils.moveToTrash`.
  - Added test to cover it.

## Verifying this change

Added new test and tested for both 2.3.4 and 1.2.1 (manually).

  - The serializers: no

## Documentation

### DIFF
--- a/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/client/HiveShimV1.java
+++ b/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/client/HiveShimV1.java
@@ -20,6 +20,10 @@ package org.apache.flink.table.catalog.hive.client;
 
 import org.apache.flink.table.catalog.exceptions.CatalogException;
 
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.hive.common.FileUtils;
 import org.apache.hadoop.hive.conf.HiveConf;
 import org.apache.hadoop.hive.metastore.IMetaStoreClient;
 import org.apache.hadoop.hive.metastore.RetryingMetaStoreClient;
@@ -30,6 +34,8 @@ import org.apache.hadoop.hive.metastore.api.Table;
 import org.apache.hadoop.hive.metastore.api.UnknownDBException;
 import org.apache.thrift.TException;
 
+import java.io.IOException;
+import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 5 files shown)


lirui-apache:
cc @zjuwangg @bowenli86 @xuefuz

xuefuz:
LGTM

bowenli86:
I wonder that, since`moveToTrash()` is mainly a util, should we move it to a separate util shim class like `HiveUtilShim`? The current `HiveShim` class, as its javadoc says, is mainly for Hive Metastore. We probably should rename it to `HiveClientShim`?

LGTM otherwise.

xuefuz:
> I wonder that, since`moveToTrash()` is mainly a util, should we move it to a separate util shim class like `HiveUtilShim`? The current `HiveShim` class, as its javadoc says, is mainly for Hive Metastore. We probably should rename it to `HiveClientShim`?
> 
> LGTM otherwise.

HiveShim is quite generic in naming. It's a shim, so in theory anything can be put there. Maybe we just need to update the javadoc to reflect the new addition.

Having another shim for utils might be an overkill for now. However, we can keep an eye on this and see how we organize better when this layer gets too heavy.

lirui-apache:
I also prefer to keep HiveShim generic. Actually APIs in HiveShim like `getViews` and `getFunction` take a `IMetaStoreClient` as parameter. So in that sense it's generic in nature.
@bowenli86 what do you think?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/client/HiveShimV1.java
```diff
@@ -79,4 +85,14 @@ public Function getFunction(IMetaStoreClient client, String dbName, String funct
 			throw e;
 		}
 	}
+
+	@Override
+	public boolean moveToTrash(FileSystem fs, Path path, Configuration conf, boolean purge) throws IOException {
```
bowenli86: how is the "purge" flag taking effects?

lirui-apache: Hive-1.2.1 doesn't support auto purge for insert overwrite, see [HIVE-15880](https://issues.apache.org/jira/browse/HIVE-15880). So it doesn't have any effect for HiveShimV1.

bowenli86: I see, can we add a comment to the interface's param explaining the flag only works for hive 2 (or 2.1.x, or any version it works)?

bowenli86:
Thanks @lirui-apache for the PR and review from @xuefuz . 

LGTM. Merging

PR #8611, opened 2019-06-04
[FLINK-12693][state] Store state per key-group in CopyOnWriteStateTable

banmoy:

Extract the customized hash-map out of `CopyOnWriteStateTable`, and store state in such hash-map by key-group as same as `NestedMapsStateTable`. After that, `CopyOnWriteStateTable` and `NestedMapsStateTable` will have some common logic both for state access and snapshot, so we refactor the code to dismiss the duplication.

## Brief change log
  - Introduce `StateMap` which stores state in a key-group. The implementation includes `NestedStateMap` and `CopyOnWriteStateMap`.  `NestedStateMap` uses a nested map which provides an outer scope by namespace and an inner scope by key; `CopyOnWriteStateMap` is a customized hash-map extracted from the original `CopyOnWriteStateTable`.
 - `NestedMapsStateTable` stores state into different `NestedStateMap`s by key-group
 - `CopyOnWriteStateTable` stores state into different `CopyOnWriteStateMap`s by key-group
 - Introduce `StateMapSnapshot` which is a snapshot of `StateMap`
 - Refactor `StateTable` and `AbstractStateTableSnapshot` to dismiss the duplicated code for the nested and copy-on-write tables

## Verifying this change

  - The serializers: (no)

## Documentation


carp84:
Thanks for the patch @banmoy , could you post the result of [state benchmark](https://github.com/dataArtisans/flink-benchmarks) to make sure change here won't cause any performance regression? Thanks.

banmoy:
@carp84 I run the benchmark on 3 servers, and on each server, I run the benchmark 3 times both for the base code and this PR. The result is as follows, and I think there is no performance regression.

image

banmoy:
Hi, @StefanRRichter Can you help to review this PR? Thanks.

flink-runtime/src/main/java/org/apache/flink/runtime/state/heap/CopyOnWriteStateTable.java
```diff
@@ -219,985 +43,41 @@
...
-		}
-
-		public final void setState(@Nullable S value, int mapVersion) {
-			// naturally, we can update the state version every time we replace the old state with a different object
-			if (value != state) {
-				this.state = value;
-				this.stateVersion = mapVersion;
+	CopyOnWriteStateMapSnapshot<K, N, S>[] getStateMapSnapshotArray() {
+		CopyOnWriteStateMapSnapshot<K, N, S>[] snapshotArray =
+			new CopyOnWriteStateMapSnapshot[state.length];
```
StefanRRichter: I would suggest to suppress the warning for unchecked assignment here or, as this place is not really performance critical, rather just use array list instead of array.

flink-runtime/src/main/java/org/apache/flink/runtime/state/heap/StateTable.java
```diff
@@ -198,16 +377,118 @@ public void setMetaInfo(RegisteredKeyValueStateBackendMetaInfo<N, S> metaInfo) {
...
+		@Override
+		public Collection<StateEntry<K, N, S>> nextEntries() {
+			if (!hasNext()) {
+				return null;
+			}
+
+			Collection<StateEntry<K, N, S>> collection =
+				stateIncrementalVisitor.nextEntries();
+
+			return collection;
```
StefanRRichter: Nit: could just `return stateIncrementalVisitor.nextEntries();`

flink-runtime/src/main/java/org/apache/flink/runtime/state/heap/StateTable.java
```diff
@@ -156,10 +203,23 @@ public boolean isEmpty() {
...
-			StateTransformationFunction<S, T> transformation) throws Exception;
+			StateTransformationFunction<S, T> transformation) throws Exception {
+		K key = keyContext.getCurrentKey();
+		checkKeyNamespacePreconditions(key, namespace);
+
+		int keyGroup = keyContext.getCurrentKeyGroupIndex();
+		StateMap<K, N, S> stateMap = getMapForKeyGroup(keyGroup);
+
+		if (stateMap == null) {
+			stateMap = createStateMap();
```
StefanRRichter: I think you make a trade-off here between the attempt to save some space, and some code complexity and checking overhead for `null` in the array. Given that the keys are typically uniformly distributed across key-groups, I would assume that no map will stay `null` for long. What I would suggest is to initialize all array positions with maps at the time of array allocation and keep the invariant that all places in the array are strictly non-null. This allows you to drop all null checks on around this array.

What I can forsee is that maybe the reason to have the checks is for followup work that implements the spilling and for spilled key-groups you can have some maps removed from the array? If that is the case, then we could either keep the check or (maybe better) have a special implementation for an empty/spilled map?

StefanRRichter: If you feel like you want to keep it, you might consider writing a helper method for the task of creating and adding maps, because this is repeated code in multiple places.

flink-runtime/src/main/java/org/apache/flink/runtime/state/heap/StateTable.java
```diff
@@ -172,14 +232,133 @@ public boolean isEmpty() {
...
+	public S get(K key, N namespace) {
+		int keyGroup = KeyGroupRangeAssignment.assignToKeyGroup(key, keyContext.getNumberOfKeyGroups());
+		return get(key, keyGroup, namespace);
+	}
+
+	public Stream<K> getKeys(N namespace) {
+		return Arrays.stream(state)
+			.filter(Objects::nonNull)
+			.map(StateMap::iterator)
+			.flatMap(iter -> StreamSupport.stream(Spliterators.spliteratorUnknownSize(iter, 0), false))
```
StefanRRichter: I think if you flatmap directly from the StateMap, you could get the size of the spliterator instead of using `spliteratorUnknownSize` ?

flink-runtime/src/main/java/org/apache/flink/runtime/state/heap/AbstractStateTableSnapshot.java
```diff
@@ -19,34 +19,123 @@
...
 
 	/**
 	 * Creates a new {@link AbstractStateTableSnapshot} for and owned by the given table.
 	 *
 	 * @param owningStateTable the {@link StateTable} for which this object represents a snapshot.
 	 */
-	AbstractStateTableSnapshot(T owningStateTable) {
+	AbstractStateTableSnapshot(
+		T owningStateTable,
+		TypeSerializer<K> localKeySerializer,
```
StefanRRichter: You could also use `@Nonnull` for the parameters that go to non-null fields instead of `Preconditions`. The annotation also generates an assertion check for non-null if the code is executed with `-ea`

banmoy: > The annotation also generates an assertion check for non-null if the code is executed with -ea

I am confused about this. I use a simple code to test the assertion, but nothing happened no matter whether the `-ea` is added. The code is as follows, the version of jsr305 jar is 1.3.9 which is the same as flink used, and the version of java is 1.8.0_162.
```Java
import javax.annotation.Nonnull;

public class AnnotationTest {

	public static void main(String[] args) {
		func(null);
	}

	private static void func(@Nonnull Object object) {
		System.out.println("success");
	}
}
```
But I find Intellij has a setting `Add runtime assertions for not-null annotated methods and parameters`，and if I run the code in Intellij with this setting enabled, there will be an `IllegalArgumentException` throwed. Please let me know if I'm wrong.

flink-runtime/src/main/java/org/apache/flink/runtime/state/heap/StateTable.java
banmoy: The style of checking `null` is just following the original NestedMapsStateTable. For the followup spill implementation, we just replace the on-heap map with on-disk map, instead of removing the map from the array, so there is no need to set `null` in the array. I agree with you that no map will stay `null` for long if the number of keys is more than the number of key-groups and keys are uniformly distributed, which I think is the most case. So I will initialize all array positions with maps and remove the check.

flink-runtime/src/main/java/org/apache/flink/runtime/state/heap/StateTable.java
banmoy: Yes, we can create a spliterator with known size as you say. But in this case, we will not iterate keys in parallel, so the size will not be used. On the other hand, the size for `NestedStateMap` need some computation. So I think there is no need to get the size of the spliterator.

banmoy:
@StefanRRichter I have addressed the comments.

flink-runtime/src/main/java/org/apache/flink/runtime/state/heap/StateTable.java
```diff
@@ -172,14 +229,110 @@ public boolean isEmpty() {
...
-	public abstract Stream<K> getKeys(N namespace);
+	// ------------------------------------------------------------------------
 
-	public abstract StateIncrementalVisitor<K, N, S> getStateIncrementalVisitor(int recommendedMaxNumberOfReturnedRecords);
+	private S get(K key, int keyGroupIndex, N namespace) {
+		checkKeyNamespacePreconditions(key, namespace);
+
+		StateMap<K, N, S> stateMap = getMapForKeyGroup(keyGroupIndex);
+
+		if (stateMap == null) {
```
StefanRRichter: I think this check can go away with the changes that make all maps non-null.

flink-runtime/src/main/java/org/apache/flink/runtime/state/heap/StateMap.java
```diff
@@ -0,0 +1,185 @@
...
+ * @param <K> type of key
+ * @param <N> type of namespace
+ * @param <S> type of state
+ */
+public abstract class StateMap<K, N, S> implements Iterable<StateEntry<K, N, S>> {
+
+	/**
+	 * The serializer of the state.
+	 */
+	protected final TypeSerializer<S> stateSerializer;
```
StefanRRichter: I think that only the copy on write variant is using the serializer, so it can probably be removed from the abstract class and moved into the concrete class.

flink-runtime/src/main/java/org/apache/flink/runtime/state/heap/CopyOnWriteStateTableSnapshot.java
```diff
@@ -54,242 +37,45 @@
...
-	@Nullable
-	private final StateSnapshotTransformer<S> stateSnapshotTransformer;
-
-	/**
-	 * Result of partitioning the snapshot by key-group. This is lazily created in the process of writing this snapshot
-	 * to an output as part of checkpointing.
-	 */
-	@Nullable
-	private StateKeyGroupWriter partitionedStateTableSnapshot;
+	private final CopyOnWriteStateMapSnapshot<K, N, S>[] snapshotStateMap;
```
StefanRRichter: At least for not super performance critical code like this, I would almost always prefer `List<CopyOnWriteStateMapSnapshot<K, N, S>>` (backed by array list) over pure array for the type safety and better integration with generics.

flink-runtime/src/main/java/org/apache/flink/runtime/state/heap/StateMapSnapshot.java
```diff
@@ -0,0 +1,79 @@
...
+ * @param <K> type of key
+ * @param <N> type of namespace
+ * @param <S> type of state
+ */
+public abstract class StateMapSnapshot<K, N, S, T extends StateMap<K, N, S>> {
+
+	/**
+	 * The {@link StateMap} from which this snapshot was created.
+	 */
+	protected final T owningStateMap;
```
StefanRRichter: This is only used by the copy on write variant and therefore could probably be removed from a common abstract class. The abstract class might then just be an interface.

flink-runtime/src/main/java/org/apache/flink/runtime/state/heap/AbstractStateTableSnapshot.java
```diff
@@ -19,34 +19,106 @@
...
+
+import java.io.IOException;
+
 /**
  * Abstract base class for snapshots of a {@link StateTable}. Offers a way to serialize the snapshot (by key-group).
  * All snapshots should be released after usage.
  */
 @Internal
-abstract class AbstractStateTableSnapshot<K, N, S, T extends StateTable<K, N, S>> implements StateSnapshot {
+abstract class AbstractStateTableSnapshot<K, N, S, T extends StateTable<K, N, S>>
```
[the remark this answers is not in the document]
banmoy: I think there is no need to introduce `getStateMapSnapshotArray` for `NestedStateTable`. For `CopyOnWriteStateTable`, we should create snapshots for all `StateMap`s in the sync part of the snapshot for backend, hold them somewhere like in an array, and use them to write state to output in the async part. But for `NestedStateTable` which only supports synchronous snapshot, we can create the snapshot of `StateMap` and write state to output one by one, and there is no need to hold all `StateMap`'s snapshots firstly and then write state to output

flink-runtime/src/main/java/org/apache/flink/runtime/state/heap/StateTable.java
banmoy: This check is to pass the UT of StateBackendTestBase#testKeyGroupSnapshotRestore. The UT will get a key from the backend, but the key is in a key group which not belongs to this backend.

flink-runtime/src/main/java/org/apache/flink/runtime/state/heap/StateMapSnapshot.java
banmoy: Actually, `owningStateMap ` is also used in `NestedStateMapSnapshot#writeState`.

flink-runtime/src/main/java/org/apache/flink/runtime/state/heap/StateTable.java
```diff
@@ -39,7 +47,8 @@
  * @param <N> type of namespace
  * @param <S> type of state
  */
-public abstract class StateTable<K, N, S> implements StateSnapshotRestore {
+public abstract class StateTable<K, N, S>
```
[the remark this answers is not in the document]
banmoy: I prefer to use the inheritance for the following reasons:
1. in the followup spill implementation, we need some memory statistics about the `StateTable` to decide whether to build a on-heap map or a on-disk map. I think it's not convenient for `MapAndSnapshotFactory ` to communicate with `StateTable`
2. implementation of `StateTable` may need it's own method, such as `getStateMapSnapshotArray` in `CopyOnWriteStateTable` and more custom methods in the spill implementation
3. if we introduce `MapAndSnapshotFactory`, we should also change the way to create `StateTable` using `SnapshotStrategySynchronicityBehavior` currently, but I think this is not in the scope of this refactor

banmoy:
@StefanRRichter Thanks for the review. I have addressed the comments, and please help to review again.

flink-runtime/src/main/java/org/apache/flink/runtime/state/heap/StateTable.java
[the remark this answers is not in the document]
StefanRRichter: Ok, if your followup work has more reason to use different classes, then you should keep the inheritence. 👍

flink-runtime/src/main/java/org/apache/flink/runtime/state/heap/AbstractStateTableSnapshot.java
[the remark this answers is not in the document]
StefanRRichter: I can see your point. One could implement the structure of the snapshots in a way that is eager for copy-on-write and lazy for the nested table, so that they can still have the same interface, but I would not make that a blocker for the PR and think the current solution is also ok.

flink-runtime/src/main/java/org/apache/flink/runtime/state/heap/StateTable.java
StefanRRichter: I think that there is also a problem with the test because it is attempting to access a key on a backend that is not from a key-group that belongs to the backend. In production this would indicate a bug, so maybe it is better to have this as a precondition check on the backends and throw `IllegalStateException` and adjust the test accordingly.

flink-runtime/src/main/java/org/apache/flink/runtime/state/heap/StateMapSnapshot.java
StefanRRichter: Right, sorry I overlooked that one :)

banmoy:
@StefanRRichter Thanks for the quick reply. I agree with you  that the ut is not resonable and needs to fix, but I'm not sure whether the ut should be fixed in this PR or in another. It seems that the problem is  irrelevant to this PR, and I prefer to fix it in anther PR.  What do you think? BTW, no matter where we fix it, I think I can be responsible for the problem.

banmoy:
@StefanRRichter Thanks for your review.

banmoy:
The CI failed with checkstyle errors which seems irrelevant to this PR. I'm looking for the reason.
`00:22:30.637 [ERROR] src/main/java/org/apache/flink/table/expressions/RexNodeConverter.java:[582,25] (naming) MethodName: Name 'ConvertGet' must match pattern '^[a-z][a-zA-Z0-9]*(_[a-zA-Z0-9]+)*$'.
00:22:30.638 [ERROR] src/main/java/org/apache/flink/table/expressions/RexNodeConverter.java:[955,29] (naming) LocalVariableName: Name 'DECIMAL_PRECISION_NEEDED_FOR_LONG' must match pattern '^[a-z][a-zA-Z0-9]*$'.
 `

StefanRRichter:
Yes, it seems unrelated but also prevents us from seeing the result of the build. Can you check when the master is green, rebase to that master and force-push your commit again?

banmoy:
rebase master to fix the build error for checkstyle

banmoy:
The failed CI reported by flinkbot includes two failures:
1. `KeyedComplexChainTest. testMigrationAndRestore ` is a known unstable case, and is being solved by [FLINK-12916](https://issues.apache.org/jira/browse/FLINK-12916?jql=text%20~%20%22testMigrationAndRestore%22). Detail error log is as follows:
`17:34:30.933 [ERROR] testMigrationAndRestore[Migrate Savepoint: 1.4](org.apache.flink.test.state.operator.restore.keyed.KeyedComplexChainTest)  Time elapsed: 1.091 s  <<< ERROR!
java.util.concurrent.ExecutionException: java.util.concurrent.CompletionException: org.apache.flink.runtime.checkpoint.CheckpointException: Task received cancellation from one of its inputs
Caused by: java.util.concurrent.CompletionException: org.apache.flink.runtime.checkpoint.CheckpointException: Task received cancellation from one of its inputs
Caused by: org.apache.flink.runtime.checkpoint.CheckpointException: Task received cancellation from one of its inputs
Caused by: org.apache.flink.runtime.checkpoint.CheckpointException: Task received cancellation from one of its inputs
`

2. `SplitAggregateITCase.testMinMaxWithRetraction` failed on rocksdb, which is unrelated to this PR, and I think we should open a jira to solve it. Detail error log is as follows:
`17:37:42.984 [ERROR] testMinMaxWithRetraction[PartialAgg=ON, LocalGlobal=ON, StateBackend=ROCKSDB](org.apache.flink.table.runtime.stream.sql.SplitAggregateITCase)  Time elapsed: 2.81 s  <<< FAILURE!
java.lang.AssertionError: expected:<List(2,2,2,1, 5,1,4,2, 6,2,2,1)> but was:<List(2,1,2,1, 5,1,4,2, 6,2,2,1)>
	at org.apache.flink.table.runtime.stream.sql.SplitAggregateITCase.testMinMaxWithRetraction(SplitAggregateITCase.scala:195)`

@StefanRRichter What do you think?

StefanRRichter:
Thanks for the good work @banmoy ! Travis is green now, so I can merge this now.

banmoy:
Thanks for your review. :)  @StefanRRichter

PR #8613, opened 2019-06-04
[FLINK-12730][Runtime / Coordination] Combine BitSet implementations in flink-runtime

liyafan82:

There are two implementations for BitSet in flink-runtime ocmponent: one is org.apache.flink.runtime.operators.util.BloomFilter#BitSet, while the other is org.apache.flink.runtime.operators.util.BitSet

The two classes are quite similar in their API and implementations. The only difference is that, the former is based based on long operation while the latter is based on byte operation. This has the following consequence:

The byte based BitSet has better performance for get/set operations.
The long based BitSet has better performance for the clear operation.
We combine the two implementations and make the best of both worlds.

## Brief change log

  - Remove the BitSet class in org.apache.flink.runtime.operators.util.BloomFilter
  - Revise class org.apache.flink.runtime.operators.util.BitSet for better performance
  

## Verifying this change

This change is already covered by existing tests, such as *(BloomFilterTest)*.

  - The serializers: (no)

## Documentation


flink-runtime/src/main/java/org/apache/flink/runtime/operators/util/BitSet.java
```diff
@@ -89,8 +88,13 @@ public int bitSize() {
 	 * Clear the bit set.
 	 */
 	public void clear() {
-		for (int i = 0; i < byteLength; i++) {
-			memorySegment.put(offset + i, (byte) 0);
+		int index = 0;
+		while (index < byteLength) {
```
StefanRRichter: This looks like a potential bug to me, assuming that we can always still add 8 bytes if we are smaller than `byteLength` can go out of bounds IMO. Or in case that it could not go out of bounds because it is always a multiple of 8, the second loop that follows does not make sense and it is currently also never invoked. I would then suggest to add a test that covers clearing from a bitset with an amount of bits that is not divisible by 8.

liyafan82: @StefanRRichter , thanks a lot for your careful review. 

It is a real bug. I have fixed it. In addition, I have added test cases by making BitSetTest parameterized. So the clear method is tested against different amounts of bits.

PR #8615, opened 2019-06-04
 [FLINK-12729][state-processor-api] Add state reader for consuming non-partitioned operator state

sjwiesman:

This is the initial PR for FLIP-43 adding the functionality to read non-partitioned operator state from a state snapshot. 

## Brief changelog
- The first two commits expose internal api's for use outside of their package but mark them for internal use only. 
- The third commit adds the operator state reader functionality 

## Verifying this change
- Unit and IT tests

## Documentation

    (Full documentation coming in the keyed state reader PR)


sjwiesman:
@flinkbot attention @tzulitai

tzulitai:
@flinkbot approve description
@flinkbot approve consensus

flink-libraries/flink-state-processing-api/src/main/java/org/apache/flink/state/api/input/OperatorStateInputFormat.java
```diff
@@ -0,0 +1,170 @@
...
+			CollectionUtil.partition(splits[0].getPrioritizedManagedOperatorState().get(0).asList(), minNumSplits),
+			(state, index) ->  new OperatorStateInputSplit(new StateObjectCollection<>(new ArrayList<>(state)), index)
+		).toArray(OperatorStateInputSplit[]::new);
+	}
+
+	private OperatorStateInputSplit[] getOperatorStateInputSplits(int minNumSplits) throws IOException {
+		final OperatorState operatorState = getOperatorState();
+
+		final Map<OperatorInstanceID, List<OperatorStateHandle>> newManagedOperatorStates = new HashMap<>();
+		final Map<OperatorInstanceID, List<OperatorStateHandle>> newRawOperatorStates = new HashMap<>();
```
tzulitai: nit: maybe add an "unused" to the variable naming, to indicate that this isn't used at all.

flink-libraries/flink-state-processing-api/src/main/java/org/apache/flink/state/api/input/SavepointInputFormat.java
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	/**
+	 * Finds the {@link OperatorState} for a uid within a savepoint.
+	 *
+	 * @return A handle to the operator state in the savepoint with the provided uid.
+	 * @throws IOException If the savepoint path is invalid or the uid does not exist
+	 */
+	OperatorState getOperatorState() throws IOException {
+		final Savepoint savepoint = SavepointLoader.loadSavepoint(savepointPath);
```
tzulitai: I've been thinking whether or not this operation can be done more earlier when loading a `ExistingSavepoint`, and maintained by the `ExistingSavepoint` class. Not entirely sure though, as that would require access to the DFS on the client side; not sure how feasible that is in practice.

Not too much of a deal, as this shouldn't be a heavy workload.

flink-libraries/flink-state-processing-api/src/main/java/org/apache/flink/state/api/runtime/SavepointLoader.java
```diff
@@ -0,0 +1,54 @@
...
+	 * @return A state handle to savepoint's metadata.
+	 * @throws IOException Thrown, if the path cannot be resolved, the file system not accessed, or
+	 *     the path points to a location that does not seem to be a savepoint.
+	 */
+	public static Savepoint loadSavepoint(String savepointPath) throws IOException {
+		CompletedCheckpointStorageLocation location = AbstractFsCheckpointStorage
+			.resolveCheckpointPointer(savepointPath);
+
+		try (DataInputStream stream = new DataInputStream(location.getMetadataHandle().openInputStream())) {
+			return Checkpoints.loadCheckpointMetadata(stream, NullClassLoader.INSTANCE);
```
tzulitai: In the future, once we add type information of state into the savepoint metadata file, I don't think this `NullClassLoader` is correct anymore. We would potentially need the actual user classloader, since the metadata would contain user classes (e.g. user implemented `TypeSerializerSnapshot`s)

flink-libraries/flink-state-processing-api/src/main/java/org/apache/flink/state/api/runtime/NullClassLoader.java
```diff
@@ -0,0 +1,61 @@
...
+package org.apache.flink.state.api.runtime;
+
+import java.io.InputStream;
+import java.net.URL;
+import java.util.Enumeration;
+
+/**
+ * A classloader that does not work.
+ */
+public class NullClassLoader extends ClassLoader {
```
tzulitai: Add `@Internal`.

flink-libraries/flink-state-processing-api/src/main/java/org/apache/flink/state/api/runtime/SavepointLoader.java
sjwiesman: There was one place where I couldn't get access to the user classloader but it wasn't being used so I decided to go this route. However, if we do the reading on the client the current thread class loader should be the user class loader so that makes it easier.

sjwiesman:
Thanks for the review @tzulitai !

I've addressed all the comments including moving the savepoint loader to run once on the client.

PR #8616, opened 2019-06-04
[FLINK-12718][hive] allow users to specify hive-site.xml location to configure hive metastore client in HiveCatalog

bowenli86:

This PR enables users to specify the path of `hive-site.xml` file as param to build a HiveCatalog. `hive-site.xml` is a standard way of configuring Hive client. Configs in `hive-site` is a superset of the config `hive.metastore.uris` which we currently use to config HiveCatalog. 

## Brief change log

- supported file path and url of `hive-site.xml` as param to construct HiveCatalog

## Verifying this change

  - The serializers: (no)

## Documentation

Documentation will be added later

### DIFF
--- a/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalog.java
+++ b/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalog.java
@@ -77,6 +77,8 @@ import org.apache.thrift.TException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import javax.annotation.Nullable;
+
 import java.io.File;
 import java.net.MalformedURLException;
 import java.net.URL;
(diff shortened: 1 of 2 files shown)


bowenli86:

bowenli86:

xuefuz:
Thanks for the contribution. I have a couple of high-level comments:

1. The PR seems suggesting that hive-site.xml path is always required, which shouldn't be the case. We should just instantiate a HiveConf instance and get the config dir from the environment.

2. There seems to be a good number of constructors. We can probably reduce it by only providing Catalog(String catName, String dbName, String/URL hiveSiteLocation) and taking care of nulls. We can add more on demand later on.

lirui-apache:
LGTM, thanks.

bowenli86:
> Thanks for the contribution. I have a couple of high-level comments:
> 
> 1. The PR seems suggesting that hive-site.xml path is always required, which shouldn't be the case. We should just instantiate a HiveConf instance and get the config dir from the environment.
> 2. There seems to be a good number of constructors. We can probably reduce it by only providing Catalog(String catName, String dbName, String/URL hiveSiteLocation) and taking care of nulls. We can add more on demand later on.

I was thinking that the two issues above can be solved by introducing a builder. I will keep just a few constructors required by the following PR for catalog entries, and add a builder later on

bowenli86:

xuefuz:
I think a builder might be an overkill. We just need to provide a couple of basic constructors and take care of nulls.

Looking at the PR I see we still force user to provide a url for hive-site location. To me that's optional. If user doesn't provide it (or provide null for it), we should just instantiate HiveConf, which takes whatever from the environment. Did I miss anything?

bowenli86:
@xuefuz I added constructors with nullable args and `@Nullable` annotations to remind users of the behaviors. can you take another look

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalog.java
```diff
@@ -109,27 +114,51 @@
...
+		super(catalogName, defaultDatabase == null ? DEFAULT_DB : defaultDatabase);
+
+		this.hiveConf = hiveConf == null ? getHiveConf("") : hiveConf;
 
 		LOG.info("Created HiveCatalog '{}'", catalogName);
 	}
 
-	private static HiveConf getHiveConf(String hiveMetastoreURI) {
-		checkArgument(!StringUtils.isNullOrWhitespaceOnly(hiveMetastoreURI), "hiveMetastoreURI cannot be null or empty");
+	private static HiveConf getHiveConf(String filePath) {
```
xuefuz: Nit: I think instead of having two overloaded getHiveConf(), we can have one that convert string to URL and another that instantiates HiveConf. This might make the code easier to read.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalog.java
```diff
@@ -109,27 +114,51 @@
...
+			}
+		}
+
+		return getHiveConf(url);
+	}
+
+	private static HiveConf getHiveConf(URL hiveSiteUrl) {
+		LOG.info("Setting hive-site location as {}", hiveSiteUrl);
+
+		HiveConf.setHiveSiteLocation(hiveSiteUrl);
```
xuefuz: Nit: it's probably better if we don't rely Hive to handle null. We can call HiveConf.setHiveSiteLocation() only if hiveSiteUrl is not null.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalog.java
bowenli86: make sense to me. changed.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalog.java
bowenli86: Since hive site url is a static member variable, we need to set it even when it's null, in order to support multi hive catalogs. Say 1st catalog has hive site and 2nd doesn't. Not setting it as null for the 2nd would make the 2nd reuse the 1st's hive site file.

bowenli86:
@xuefuz thanks for your review and I adopted your suggestions.

Merging

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalog.java
lirui-apache: Wondering will this incur any concurrency issue, e.g. when user access tables from different Hive catalogs? I think at least some constructors of HiveConf is no longer safe to use after this change, like `HiveConf()` and `HiveConf(Class<?> cls)` which automatically load the specified hive-site.xml.
I think it's better not to rely on static fields of HiveConf. Instead, we can load the hive-site (perhaps `Properties::loadFromXML`?) ourselves and `HiveConf::hiveSiteURL` should always be set to null.

bowenli86: maybe we should reset the HiveConf's hiveSiteUrl back to `null` after creating the HiveConf instance?

lirui-apache: That's also a viable solution, in which case we need a unified place to create HiveConf instances.

PR #8617, opened 2019-06-04
[FLINK-12619][StateBackend]Support TERMINATE/SUSPEND Job with Checkpoint

klion26:

Inspired by the idea of [FLINK-11458](https://issues.apache.org/jira/browse/FLINK-11458), we propose to support terminate/suspend a job with checkpoint. This improvement cooperates with incremental and external checkpoint features, that if checkpoint is retained and this feature is configured, we will trigger a checkpoint before the job stops. It could accelarate job recovery a lot since:
1. No source rewinding required any more.
2. It's much faster than taking a savepoint since incremental checkpoint is enabled.

Please note that conceptually savepoints is different from checkpoint in a similar way that backups are different from recovery logs in traditional database systems. So we suggest using this feature only for job recovery, while stick with FLINK-11458 for the upgrading/cross-cluster-job-migration/state-backend-switch cases.

The current commit does not include rest API(will do this by follow up issue  [FLINK-12733](https://issues.apache.org/jira/browse/FLINK-12733).
## Verifying this change

  - CliFrontendStopTest.java#{stopWithCheckpoint, testStopWithCheckpointWithMaxWM, testStopWithCheckpointAndSavepoint}
  - CheckpointPropertiesTest.java#testSyncCheckpoint
  - CheckpointTypeTest.java #testOrdinalsAreConstant
  - PendingCheckpointTest.java #testSyncCheckpointCannotBeSubsumed
  - CheckpointSerializationTest.java 
  - EventSerializerTest.java #testIsEvent
  - SynchronousCheckpointITCase.java
  - SynchronousCheckpointTest.java
 
  - The serializers: (no)

## Documentation

@aljoscha @StefanRRichter

### DIFF
--- a/flink-clients/src/main/java/org/apache/flink/client/cli/CliFrontend.java
+++ b/flink-clients/src/main/java/org/apache/flink/client/cli/CliFrontend.java
@@ -511,9 +511,9 @@ public class CliFrontend {
 		}
 
 		if (stopOptions.hasCheckpointFlag()) {
-			LOG.info("Running 'stop-with-savepoint' command.");
+			LOG.info("Running 'stop-with-checkpoint' command.");
 		} else {
-			LOG.info("Running 'stop-with-savepoint' command.");
+			LOG.info("Running 'stop-with-checkpoint' command.");
 		}
 
 		final String[] cleanedArgs = stopOptions.getArgs();
--- a/flink-clients/src/main/java/org/apache/flink/client/cli/CliFrontend.java
+++ b/flink-clients/src/main/java/org/apache/flink/client/cli/CliFrontend.java
@@ -500,8 +500,6 @@ public class CliFrontend {
 	 * @param args Command line arguments for the stop action.
 	 */
 	protected void stop(String[] args) throws Exception {
-		LOG.info("Running 'stop-with-savepoint' command.");
-
 		final Options commandOptions = CliFrontendParser.getStopCommandOptions();
 		final Options commandLineOptions = CliFrontendParser.mergeOptions(commandOptions, customCommandLineOptions);
 		final CommandLine commandLine = CliFrontendParser.parse(commandLineOptions, args, false);
@@ -510,6 +508,12 @@ public class CliFrontend {
 		if (stopOptions.isPrintHelp()) {
 			CliFrontendParser.printHelpForStop(customCommandLines);
 			return;
+		}
+
+		if (stopOptions.hasCheckpointFlag()) {
+			LOG.info("Running 'stop-with-savepoint' command.");
+		} else {
+			LOG.info("Running 'stop-with-savepoint' command.");
 		}
 
 		final String[] cleanedArgs = stopOptions.getArgs();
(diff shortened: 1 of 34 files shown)


flink-clients/src/main/java/org/apache/flink/client/cli/CliFrontend.java
```diff
@@ -515,6 +513,12 @@ protected void stop(String[] args) throws Exception {
 			return;
 		}
 
+		if (stopOptions.hasCheckpointFlag()) {
+			LOG.info("Running 'stop-with-savepoint' command.");
```
carp84: here it should be "stop-with-checkpoint"

flink-runtime/src/main/java/org/apache/flink/runtime/checkpoint/CheckpointCoordinator.java
```diff
@@ -424,6 +424,27 @@ public boolean isShutdown() {
 		}
 	}
 
+	public CompletableFuture<CompletedCheckpoint> triggerSynchronousCheckpoint(
```
carp84: It seems to me that we could reuse the existing codes of `triggerSynchronousSavepoint`, adding a flag to indicate whether it's a checkpoint or savepoint. The same applies to `LegacyScheduler#stopWithCheckpoint`

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/LegacyScheduler.java
```diff
@@ -624,4 +624,58 @@ public void declineCheckpoint(final DeclineCheckpoint decline) {
 		return savepointFuture.thenCompose((path) ->
 			terminationFuture.thenApply((jobStatus -> path)));
 	}
+
+	@Override
+	public CompletableFuture<String> stopWithCheckpoint(boolean advanceToEndOfEventTime) {
```
carp84: Please check my comments on `CheckpointCoordinator` about preventing duplicated codes.

aljoscha:
From a first glance I agree with @carp84's comments that we should prevent code duplication as much as possible. Regarding `CANCEL-WITH-SAVEPOINT`, I don't think we need to worry about this one since `STOP-WITH-SAVEPOINT` is meant to replace this in the long run. We only left this in for now to now break the REST API/CLI.

klion26:
thanks for your reviews @carp84 @aljoscha  will address all your comments.

aljoscha:
We should of course wait until the discussion on https://issues.apache.org/jira/browse/FLINK-12619 has converged.

carp84:
Since we reached a consensus after some nice and constructive discussion with @aljoscha , let's resume the review here. Please do necessary rebase and refactor to prepare a new PR @klion26 , thanks.

klion26:
@aljoscha @carp84 I've rebased the master and addressed the comments, please have another look when you're free, thanks.

klion26:
Seems the failed test is related, I'll fix it.

klion26:
Have fixed the failed test.
I think we may add a configure `autoCheckpoint` (whether to trigger a checkpoint when canceling a job), and when canceling a job, we'll do a checkpoint automatically if the  `autoCheckpoint` is true. what do you think @aljoscha

aljoscha:
@klion26 I would suggest to keep `autoCheckpoint` to a separate Jira issue and PR. And focus this PR on purely the stop-with-checkpoint functionality.

klion26:
@aljoscha thanks for the reply, will create a separate Jira issue to that.
close and reopen the pr for trigger another Travis test.

klion26:
Seems Travis build is broken now(all builds failed), I'll retrigger another build when Travis build is ok.

klion26:
Seems Travis build is ok now, close and reopen to trigger another build.

yumengz5:
What if the checkpoint is timeout, the task will block all the time?

klion26:
Hi, @yumengz5  what do you mean "task will block all the time?", would you mind explain it more?

yumengz5:
In my opinion, If the checkpoint is timeout,  this checkpoint will be discarded. But the checkpoint coordinator would handle such abortion as usual and assume that the Flink job continues running. In fact, it looks like "running", but it doesn't process any records. In sync-checkpoint case, every task will be blocked after triggering the checkpoint. Unlocking only happens either when the task is cancelled or when the corresponding checkpoint is acknowledged. I guess stop-with-savepoint may have the same issue. I'm not sure it will truly happen.

klion26:
I'm not sure I fully follow what you mean, in the following, I'm trying to answer your question. Please correct me if I'm not understanding right.
- If you mean "the sync checkpoint/savepoint" is timeout, then the `pendingCheckpoint` will be abort, and the `onCompletionPromis` will complete with an exception, the job exit.
- Currently, when stopping a job, the task will not exit, if you want the task exit, could use `cancel` instead of `stop`

yumengz5:
The onCompletionPromis does complete with an exception, if the pendingCheckpoint has been abort, but the job will not exit and its tasks will not resume to execute because you **don't** notify the syncLatch. For second part, I don't want the tasks to exit. I just want them to unblock and resume to execute.

klion26:
Currently, Flink doesn't' notify task about the expired checkpoint now, [FLINK-8871](https://issues.apache.org/jira/browse/FLINK-8871) will fix it. After FLINK-8871 has been fixed, we can do better here.
Thanks for your comments.

PR #8621, opened 2019-06-05
[FLINK-12682][connectors] StringWriter support custom row delimiter

wisgood:

StringWriter support custom row delimiter

## Brief change log
Add a  row delimiter parameter to the constructor

### DIFF
--- a/flink-connectors/flink-connector-filesystem/src/main/java/org/apache/flink/streaming/connectors/fs/StringWriter.java
+++ b/flink-connectors/flink-connector-filesystem/src/main/java/org/apache/flink/streaming/connectors/fs/StringWriter.java
@@ -42,6 +42,8 @@ public class StringWriter<T> extends StreamWriterBase<T> {
 
 	private String rowDelimiter;
 
+	private byte[] rowDelimiterBytes;
+
 
 	/**
 	 * Creates a new {@code StringWriter} that uses {@code "UTF-8"} charset to convert
@@ -87,7 +89,10 @@ public class StringWriter<T> extends StreamWriterBase<T> {
 	public void write(T element) throws IOException {
 		FSDataOutputStream outputStream = getStream();
 		outputStream.write(element.toString().getBytes(charset));
-		outputStream.write(rowDelimiter.getBytes(charset));
+		if (rowDelimiterBytes == null) {
+			rowDelimiterBytes = rowDelimiter.getBytes(charset);
+		}
+		outputStream.write(rowDelimiterBytes);
 	}
 
 	@Override
--- a/flink-connectors/flink-connector-filesystem/src/main/java/org/apache/flink/streaming/connectors/fs/StringWriter.java
+++ b/flink-connectors/flink-connector-filesystem/src/main/java/org/apache/flink/streaming/connectors/fs/StringWriter.java
@@ -40,7 +40,7 @@ public class StringWriter<T> extends StreamWriterBase<T> {
 
 	private transient Charset charset;
 
-	private byte[] rowDelimiter ;
+	private String rowDelimiter ;
 
 
 	/**
@@ -48,7 +48,7 @@ public class StringWriter<T> extends StreamWriterBase<T> {
 	 * strings to bytes.
 	 */
 	public StringWriter() {
-		this("UTF-8",new byte[] {'\n'});
+		this("UTF-8", "\n");
 	}
 
 	/**
@@ -57,7 +57,7 @@ public class StringWriter<T> extends StreamWriterBase<T> {
 	 *
 	 * @param charsetName Name of the charset to be used, must be valid input for {@code Charset.forName(charsetName)}
 	 */
-	public StringWriter(String charsetName,byte[] rowDelimiter ) {
+	public StringWriter(String charsetName, String rowDelimiter ) {
 		this.charsetName = charsetName;
 		this.rowDelimiter = rowDelimiter;
 	}
@@ -87,7 +87,7 @@ public class StringWriter<T> extends StreamWriterBase<T> {
 	public void write(T element) throws IOException {
 		FSDataOutputStream outputStream = getStream();
 		outputStream.write(element.toString().getBytes(charset));
-		outputStream.write(rowDelimiter);
+		outputStream.write(rowDelimiter.getBytes(charset));
 	}
 
 	@Override
@@ -99,7 +99,7 @@ public class StringWriter<T> extends StreamWriterBase<T> {
 		return charsetName;
 	}
 
-	public byte[] getRowDelimiter() {
+	public String getRowDelimiter() {
 		return rowDelimiter;
 	}
 }
--- a/flink-connectors/flink-connector-filesystem/src/main/java/org/apache/flink/streaming/connectors/fs/StringWriter.java
+++ b/flink-connectors/flink-connector-filesystem/src/main/java/org/apache/flink/streaming/connectors/fs/StringWriter.java
@@ -40,12 +40,15 @@ public class StringWriter<T> extends StreamWriterBase<T> {
 
 	private transient Charset charset;
 
+	private byte[] rowDelimiter ;
+
+
 	/**
 	 * Creates a new {@code StringWriter} that uses {@code "UTF-8"} charset to convert
 	 * strings to bytes.
 	 */
 	public StringWriter() {
-		this("UTF-8");
+		this("UTF-8",new byte[] {'\n'});
 	}
 
 	/**
@@ -54,13 +57,15 @@ public class StringWriter<T> extends StreamWriterBase<T> {
 	 *
 	 * @param charsetName Name of the charset to be used, must be valid input for {@code Charset.forName(charsetName)}
 	 */
-	public StringWriter(String charsetName) {
+	public StringWriter(String charsetName,byte[] rowDelimiter ) {
 		this.charsetName = charsetName;
+		this.rowDelimiter = rowDelimiter;
 	}
 
 	protected StringWriter(StringWriter<T> other) {
 		super(other);
 		this.charsetName = other.charsetName;
+		this.rowDelimiter = other.rowDelimiter;
 	}
 
 	@Override
@@ -82,7 +87,7 @@ public class StringWriter<T> extends StreamWriterBase<T> {
 	public void write(T element) throws IOException {
 		FSDataOutputStream outputStream = getStream();
 		outputStream.write(element.toString().getBytes(charset));
-		outputStream.write('\n');
+		outputStream.write(rowDelimiter);
 	}
 
 	@Override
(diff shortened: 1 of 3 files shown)


wisgood:
Usually who is responsible for code review? @rmetzger

rmetzger:
Hi! Thanks a lot for your contribution to Apache Flink!

Usually, the code review is done by the committer(s)/contributor(s) who discussed the implementation in the JIRA ticket.

As you can see from our [contribution guide](https://flink.apache.org/how-to-contribute.html#propose-an-improvement-or-a-new-feature):

> We recommend to first reach consensus with the community on whether a new feature is required and how to implement a new feature, before starting with the implementation. Some features might be out of scope of the project, and it’s best to discover this early.

In your case, no discussion has happened in the JIRA ticket beforehand.

There's usually a set of committers and contributors who feel responsible for certain components. In your case it's probably @aljoscha who should take a look here.

wisgood:
thanks for your explanation!

wisgood:
@pnowojski Will you have time to take a look? It is a small patch and hopefully won't take too much time. Thanks very much.

wisgood:
@aljoscha  Hi ,do you have time to check it for me? The changes are relatively small, I hope that it will not spent your time too much.Many thanks！

flink-connectors/flink-connector-filesystem/src/main/java/org/apache/flink/streaming/connectors/fs/StringWriter.java
```diff
@@ -82,7 +89,10 @@ public void open(FileSystem fs, Path path) throws IOException {
 	public void write(T element) throws IOException {
 		FSDataOutputStream outputStream = getStream();
 		outputStream.write(element.toString().getBytes(charset));
-		outputStream.write('\n');
+		if (rowDelimiterBytes == null) {
+			rowDelimiterBytes = rowDelimiter.getBytes(charset);
```
pnowojski: Why are you "caching" `rowDelimiterBytes`? Isn't this a premature optimisation?

flink-connectors/flink-connector-filesystem/src/main/java/org/apache/flink/streaming/connectors/fs/StringWriter.java
```diff
@@ -54,13 +59,15 @@ public StringWriter() {
 	 *
 	 * @param charsetName Name of the charset to be used, must be valid input for {@code Charset.forName(charsetName)}
 	 */
-	public StringWriter(String charsetName) {
+	public StringWriter(String charsetName, String rowDelimiter) {
```
pnowojski: please update the `@param` java doc.

Secondly, I'm not sure, but isn't `StringWriter` a part of public API and shouldn't we preserve the old constructor?

flink-connectors/flink-connector-filesystem/src/test/java/org/apache/flink/streaming/connectors/fs/StringWriterTest.java
```diff
@@ -32,7 +32,7 @@
 
 	@Test
 	public void testDuplicate() {
```
pnowojski: Please also add a unit test (in this file?) to cover that the custom delimiter is actually used somehow.

flink-connectors/flink-connector-filesystem/src/main/java/org/apache/flink/streaming/connectors/fs/StringWriter.java
wisgood: @pnowojski  to avoid  serializing this every time.

flink-connectors/flink-connector-filesystem/src/main/java/org/apache/flink/streaming/connectors/fs/StringWriter.java
wisgood: Yes  @pnowojski  @StephanEwen

pnowojski:
I agree with @StephanEwen. If we want to have this optimisation of avoiding the serialisation multiple times, it would be better to have it done eagerly.

flink-connectors/flink-connector-filesystem/src/main/java/org/apache/flink/streaming/connectors/fs/StringWriter.java
```diff
@@ -40,12 +40,18 @@
 
 	private transient Charset charset;
 
+	private String rowDelimiter;
```
pnowojski: Could you mark `rowDelimeter` and `rowDelimiterBytes` fields final (and set them in the constructor) as @StephanEwen suggested? `charsetName` could also be final.

flink-connectors/flink-connector-filesystem/src/main/java/org/apache/flink/streaming/connectors/fs/StringWriter.java
```diff
@@ -82,7 +103,7 @@ public void open(FileSystem fs, Path path) throws IOException {
 	public void write(T element) throws IOException {
 		FSDataOutputStream outputStream = getStream();
 		outputStream.write(element.toString().getBytes(charset));
-		outputStream.write('\n');
+		outputStream.write(rowDelimiterBytes);
```
pnowojski: You are still not testing anywhere, that `rowDelimiterBytes` field is being correctly used. Please add a test in `StringWriterTest` class, where you write some elements with standard and custom delimiter and that result is as you expected.

flink-connectors/flink-connector-filesystem/src/main/java/org/apache/flink/streaming/connectors/fs/StringWriter.java
```diff
@@ -40,12 +40,18 @@
...
+
+	private byte[] rowDelimiterBytes;
+
 	/**
 	 * Creates a new {@code StringWriter} that uses {@code "UTF-8"} charset to convert
 	 * strings to bytes.
 	 */
 	public StringWriter() {
-		this("UTF-8");
+		this("UTF-8", "\n");
```
pnowojski: nit: You can keep this line as it was before (`this("UTF-8")`) to avoid small code duplication.

wisgood: Thanks for your advice! I will submit a change later!

flink-connectors/flink-connector-filesystem/src/main/java/org/apache/flink/streaming/connectors/fs/StringWriter.java
wisgood: If rowDelimiterBytes was marked in final and set in the constructor,is it also necessary to move the the initialization of charset to the constructor ？Thanks ！

flink-connectors/flink-connector-filesystem/src/main/java/org/apache/flink/streaming/connectors/fs/StringWriter.java
wisgood: I will remove rowDelimiterBytes ，to use rowDelimiter.getBytes(charset) directly ，I  think its performance loss is very small.

flink-connectors/flink-connector-filesystem/src/main/java/org/apache/flink/streaming/connectors/fs/StringWriter.java
wisgood: Let me think about how to do it, thank you very much for your suggestion！

flink-connectors/flink-connector-filesystem/src/main/java/org/apache/flink/streaming/connectors/fs/StringWriter.java
pnowojski: Ok, I missed that. Thanks for the explanation

flink-connectors/flink-connector-filesystem/src/main/java/org/apache/flink/streaming/connectors/fs/StringWriter.java
pnowojski: I would keep the `rowDelimiterBytes` field, just to make sure that we do not introduce a performance regression of converting `\n` to bytes via `charset`. Otherwise I think we would need to benchmark this.

wisgood: Hi ,if  rowDelimiterBytes was preserved and  marked in final ,it should be  set in the constructor,but at that moment  the initialization of charset have not  finished.

pnowojski: Yes, as I wrote in another comment, I missed that `rowDelimiterBytes` depends on the `charset` which is initialised only in `open`. 

Normally I would change `charset` to `private final transient` field, but this might require adding `throws IOException` clause to the constructor and would change the current contract a little bit which is a separate independent issue. In that case, I would give +1 for keeping `rowDelimiterBytes` field non-final as you have initially proposed.

wisgood: Ok, I understand ！Thanks  for the explanation

flink-connectors/flink-connector-filesystem/src/main/java/org/apache/flink/streaming/connectors/fs/StringWriter.java
wisgood: @pnowojski  Hi , according to  your suggestions ,I add a test case!

wisgood: @pnowojski  Hello，do you have time to check it  for me?Thanks a lot !

flink-connectors/flink-connector-filesystem/src/test/java/org/apache/flink/streaming/connectors/fs/StringWriterTest.java
```diff
@@ -41,5 +80,62 @@ public void testDuplicate() {
...
+		FSDataInputStream inStream = dfs.open(outputFile);
+		byte[] buffer = new byte[inputData.getBytes(charset).length];
+		readFully(inStream, buffer);
+		inStream.close();
+		String outputData = new String(buffer, charset);
+		Assert.assertEquals(inputData, outputData);
+
+	}
+
+	private void readFully(InputStream in, byte[] buffer) throws IOException {
```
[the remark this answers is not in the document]
wisgood: LGTM!

> I've fixed couple of things myself. Could you check if you have any comments/issues regarding my fixup commit at the top: https://github.com/pnowojski/flink/commits/f12682 ?

LGTM!

pnowojski:
Incorporated in and superseded by https://github.com/apache/flink/pull/9393

PR #8624, opened 2019-06-05
[FLINK-10932]Initial flink-kubernetes module with empty implementation

tianchen92:

*Initialize the skeleton module to start native k8s integration, related to [FLINK-10932](https://issues.apache.org/jira/browse/FLINK-10932)*

## Brief change log

*(for example:)*
  - *Add flink-kubernetes module*
  - *Add Interface of Kubernetes client*

## Verifying this change

*This change is an initial empty implementation, verified by mvn clean install and Travis CI.*

  - The serializers: (no)

## Documentation

  - If yes, how is the feature documented? (the document will be introduced in later pull requests)


flink-kubernetes/pom.xml
```diff
@@ -0,0 +1,120 @@
...
+				</exclusion>
+				<exclusion>
+					<groupId>com.squareup.okhttp3</groupId>
+					<artifactId>okhttp</artifactId>
+				</exclusion>
+				<exclusion>
+					<groupId>com.fasterxml.jackson.core</groupId>
+					<artifactId>jackson-databind</artifactId>
+				</exclusion>
+			</exclusions>
```
tillrohrmann: why are these exclusions needed?

flink-kubernetes/pom.xml
```diff
@@ -0,0 +1,120 @@
...
+			<version>4.0.0</version>
+			<exclusions>
+				<exclusion>
+					<groupId>com.fasterxml.jackson.core</groupId>
+					<artifactId>jackson-core</artifactId>
+				</exclusion>
+				<exclusion>
+					<groupId>com.fasterxml.jackson.core</groupId>
+					<artifactId>jackson-databind</artifactId>
+				</exclusion>
```
tillrohrmann: Why do we need to exclude these dependencies explicitly?

flink-kubernetes/src/main/java/org/apache/flink/kubernetes/FlinkKubernetesOptions.java
```diff
@@ -0,0 +1,37 @@
...
+package org.apache.flink.kubernetes;
+
+import org.apache.flink.configuration.Configuration;
+
+/**
+ * Parameters that will be used in Flink on k8s cluster.
+ * */
+public class FlinkKubernetesOptions {
+
+	private Configuration configuration;
```
tillrohrmann: should be `final`

flink-kubernetes/src/main/java/org/apache/flink/kubernetes/kubeclient/Endpoint.java
```diff
@@ -0,0 +1,42 @@
...
+package org.apache.flink.kubernetes.kubeclient;
+
+/**
+ * Represent an endpoint.
+ * */
+public class Endpoint {
+
+	private String address;
+
+	private int port;
```
tillrohrmann: `final` missing

flink-kubernetes/src/main/java/org/apache/flink/kubernetes/kubeclient/Endpoint.java
```diff
@@ -0,0 +1,42 @@
...
+ */
+
+package org.apache.flink.kubernetes.kubeclient;
+
+/**
+ * Represent an endpoint.
+ * */
+public class Endpoint {
+
+	private String address;
```
tillrohrmann: `final` missing

flink-kubernetes/src/main/java/org/apache/flink/kubernetes/kubeclient/KubeClient.java
```diff
@@ -0,0 +1,63 @@
...
+ */
+
+package org.apache.flink.kubernetes.kubeclient;
+
+import java.util.List;
+
+/**
+ * The client to talk with kubernetes.
+ * */
+public interface KubeClient extends AutoCloseable {
```
tillrohrmann: I would exclude this interface from the initial module commit since not all methods are clear to me yet.

flink-kubernetes/src/main/java/org/apache/flink/kubernetes/FlinkKubernetesOptions.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+package org.apache.flink.kubernetes;
+
+import org.apache.flink.configuration.Configuration;
+
+/**
+ * Parameters that will be used in Flink on k8s cluster.
+ * */
+public class FlinkKubernetesOptions {
```
tillrohrmann: I would exclude this file from the initial commit.

flink-kubernetes/src/main/java/org/apache/flink/kubernetes/kubeclient/Endpoint.java
```diff
@@ -0,0 +1,42 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.kubernetes.kubeclient;
+
+/**
+ * Represent an endpoint.
+ * */
+public class Endpoint {
```
tillrohrmann: Exclude from this PR.

flink-kubernetes/pom.xml
```diff
@@ -0,0 +1,120 @@
...
+		</dependency>
+		<dependency>
+			<groupId>com.fasterxml.jackson.core</groupId>
+			<artifactId>jackson-databind</artifactId>
+			<version>2.7.5</version>
+		</dependency>
+
+		<dependency>
+			<groupId>io.fabric8</groupId>
+			<artifactId>kubernetes-client</artifactId>
```
tillrohrmann: We might wanna relocate the transitive dependencies of this dependency so that user code won't be affected by them (dependency conflicts) (e.g. relocating the jackson dependencies).

tianchen92: Previously I met the build error in travis like this(which not occur in my machine):
`Dependency convergence error for com.fasterxml.jackson.core:jackson-databind:2.7.5 paths to dependency are:
+-org.apache.flink:flink-kubernetes_2.11:1.9-SNAPSHOT
  +-io.fabric8:kubernetes-client:4.0.0
    +-io.fabric8:kubernetes-model:3.0.0
      +-com.fasterxml.jackson.module:jackson-module-jaxb-annotations:2.7.5
        +-com.fasterxml.jackson.core:jackson-databind:2.7.5
and
+-org.apache.flink:flink-kubernetes_2.11:1.9-SNAPSHOT
  +-io.fabric8:kubernetes-client:4.0.0
    +-com.fasterxml.jackson.core:jackson-databind:2.7.7
and
+-org.apache.flink:flink-kubernetes_2.11:1.9-SNAPSHOT
  +-io.fabric8:kubernetes-client:4.0.0
    +-io.fabric8:zjsonpatch:0.3.0
      +-com.fasterxml.jackson.core:jackson-databind:2.3.2`

Seems kubernetes-client has several jackson version which fail the build. So I exclude jackson dependencies in kubenetes-client and include dependency at the mean while. In this way, the problem is resolved.

I don‘t quite understand “relocate the transitive dependencies”, does it mean we specify versions of transitive dependencies in DependencyManagement？I updated this PR in this way, and the dependency error problem could also be resolved.

tianchen92:
@tillrohrmann Thanks for your comments, I have updated this PR, please take a look, thanks very much!

flink-kubernetes/pom.xml
```diff
@@ -0,0 +1,104 @@
...
+				<artifactId>jackson-databind</artifactId>
+				<version>${jackson.version}</version>
+			</dependency>
+			<dependency>
+				<groupId>com.fasterxml.jackson.core</groupId>
+				<artifactId>jackson-annotations</artifactId>
+				<version>${jackson.version}</version>
+			</dependency>
+		</dependencies>
+	</dependencyManagement>
```
tillrohrmann: I think the `dependencyManagement` section is not needed. This section only defines dependencies but does not really specify to use them.

flink-kubernetes/pom.xml
tillrohrmann: If you wanna fix a certain transitive dependency to a certain version, then it is enough to explicitly list this dependency as a direct dependency. Maven's dependency conflict resolution mechanism will use the dependency which is closest to the project root.

With relocating transitive I mean to give them a new namespace so that other code won't see it. The process is described here https://maven.apache.org/plugins/maven-shade-plugin/examples/class-relocation.html and we use it in various modules in Flink: https://github.com/apache/flink/blob/master/flink-shaded-hadoop/pom.xml#L146.

flink-kubernetes/pom.xml
tianchen92: Thanks for your quick reply, I wonder if we remove this, the error below will still occur:
image

Seems dependencyManagement allows to specify the versions of artifacts to be used when they are encountered in transitive dependencies or in dependencies where no version has been specified and this multi versions problem can be solved.

Or any better solutions for the dependency error mentioned above?

tianchen92: Thanks for your pointer, we need add relocate jackson related dependencied with adding  something like <relocation> in this pom, right?

flink-kubernetes/pom.xml
tianchen92: Thanks for your pointer, we need add relocate jackson related dependencied with adding  something like <relocation> in this pom, right?

tianchen92:
@tillrohrmann I added dependency relocation, please take a look, thanks!

tianchen92:
Add direct dependency could not solve dependency conflict in kubernetes-client (see https://travis-ci.org/apache/flink/builds/542176004?utm_source=github_status&utm_medium=notification), so I would keep dependencyManagement.

tianchen92:
@tillrohrmann According to the discuss above, do you have some other comments? Since this PR pending for couple of days, could we merge this and continue to work? thanks a lot!

wangyang0918:
Hi, @tianchen92 
I think you should remove the `dependencyManagement` and do as the following steps. 
1. Add the `exclusions` of `jackson-core` `jackson-annotations` `jackson-databind` in the `kubernetes-client` dependency.
2. Add the direct dependency with the specific version of the above exclusions.
3. Set shaded plugin.

tianchen92:
@wangyang0918 Thanks for your comments, it looks reasonable to me. I have updated this PR @tillrohrmann

flink-kubernetes/pom.xml
```diff
@@ -0,0 +1,139 @@
...
+				<artifactId>maven-shade-plugin</artifactId>
+				<version>3.2.1</version>
+				<executions>
+					<execution>
+						<phase>package</phase>
+						<goals>
+							<goal>shade</goal>
+						</goals>
+						<configuration>
+							<relocations>
```
wangyang0918: <artifactSet>
  <includes combine.children="append">
  </includes>
</artifactSet>

The tag `includes` should be set. Please refer to other module(such as flink-mesos) and update the `maven-shade-plugin` .

flink-kubernetes/pom.xml
```diff
@@ -0,0 +1,139 @@
...
+		</dependency>
+
+	</dependencies>
+
+	<build>
+		<plugins>
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-shade-plugin</artifactId>
+				<version>3.2.1</version>
```
wangyang0918: The version should be removed.

flink-kubernetes/pom.xml
```diff
@@ -0,0 +1,139 @@
...
+			<groupId>org.apache.flink</groupId>
+			<artifactId>flink-runtime_${scala.binary.version}</artifactId>
+			<version>${project.version}</version>
+			<scope>provided</scope>
+		</dependency>
+
+		<dependency>
+			<groupId>com.fasterxml.jackson.core</groupId>
+			<artifactId>jackson-core</artifactId>
+			<version>${jackson.version}</version>
```
wangyang0918: We should use a certain value that the `kubernetes-client` depends on.

tianchen92:
Thanks a lot for your comments @wangyang0918 , updated this PR

wangyang0918:
Hi, @tillrohrmann 
@tianchen92 has updated the PR. Would you mind taking look again?

wangyang0918:
After offline discussion with @tianchen92 , we come to a conclusion maven-shade-plugin (`includes` `filters` and `relocations`) should be used to build a clean flink-kubernetes jar. All the dependencies should be shaded. Please update the PR.

tianchen92:
> After offline discussion with @tianchen92 , we come to a conclusion maven-shade-plugin (`includes` `filters` and `relocations`) should be used to build a clean flink-kubernetes jar. All the dependencies should be shaded. Please update the PR.

Thanks a lot, fixed now.

PR #8625, opened 2019-06-05
[FLINK-12737][table-api][table-planner] Loosen Table class dependencies

dawidwys:

It reduces the dependencies of `TableImpl` that block us from moving it to the api module.

## Brief change log

See commitss

## Verifying this change

## Documentation


zentol:
please update the PR title

dawidwys:
Thank you @zentol for reminding!

dawidwys:
@flinkbot attention @twalthr @aljoscha

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/functions/TemporalTableFunction.java
```diff
@@ -0,0 +1,91 @@
...
+import org.apache.flink.table.operations.TableOperation;
+import org.apache.flink.types.Row;
+
+import java.sql.Timestamp;
+
+/**
+ * Class representing temporal table function over some history table. A
+ * {@link TemporalTableFunction} is also an instance of {@link TableFunction}.
+ *
+ * <p>Currently {@link TemporalTableFunction}s are only supported in streaming.
```
twalthr: nit: `Currently,`

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/functions/TemporalTableFunction.java
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	public TableOperation getUnderlyingHistoryTable() {
+		if (underlyingHistoryTable == null) {
+			throw new IllegalStateException("Accessing table field after planing/serialization");
+		}
+		return underlyingHistoryTable;
+	}
+
+	public static TemporalTableFunction create(
```
twalthr: can we make this package private or avoid that users use this method? The least we can do is marking it `@Internal` but this is still not perfect. The `Impl` approach was nicer.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/functions/TemporalTableFunction.java
```diff
@@ -0,0 +1,91 @@
...
+import java.sql.Timestamp;
+
+/**
+ * Class representing temporal table function over some history table. A
+ * {@link TemporalTableFunction} is also an instance of {@link TableFunction}.
+ *
+ * <p>Currently {@link TemporalTableFunction}s are only supported in streaming.
+ */
+@PublicEvolving
+public class TemporalTableFunction extends TableFunction<Row> {
```
twalthr: `final`

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/functions/TemporalTableFunction.java
dawidwys: We can still go with the `Impl` approach. Will update it.

flink-table/flink-table-planner/src/test/scala/org/apache/flink/table/plan/TimeIndicatorConversionTest.scala
```diff
@@ -253,9 +253,9 @@ class TimeIndicatorConversionTest extends TableTestBase {
         ),
         term("groupBy", "long"),
         term("window", "TumblingGroupWindow('w2, 'newrowtime, 1000.millis)"),
-        term("select", "long", "SUM(int) AS TMP_3", "end('w2) AS TMP_2")
+        term("select", "long", "SUM(int) AS TMP_1", "end('w2) AS TMP_0")
```
twalthr: side comment: If we would `TMP_` to `EXPR$`, I think we would get closer to Calcite's output.

PR #8630, opened 2019-06-05
[FLINK-12667][runtime] Add JobID to TaskExecutorGateway#releasePartitions

zentol:

Adds `JobID` as an argument to `TaskExecutorGateway#releasePartitions` to simplify bookkeeping on the taskmanager side.

Additionally adds tests for calls to said method originating in `Execution`.

### DIFF
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/Execution.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/Execution.java
@@ -1329,7 +1329,7 @@ public class Execution implements AccessExecution, Archiveable<ArchivedExecution
 
 			if (!partitionIds.isEmpty()) {
 				// TODO For some tests this could be a problem when querying too early if all resources were released
-				taskManagerGateway.releasePartitions(partitionIds);
+				taskManagerGateway.releasePartitions(getVertex().getJobId(), partitionIds);
 			}
 		}
 	}
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/Execution.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/Execution.java
@@ -801,7 +801,7 @@ public class Execution implements AccessExecution, Archiveable<ArchivedExecution
 			}
 
 			else if (current == FINISHED || current == FAILED) {
-				// nothing to do any more. finished failed before it could be cancelled.
+				// nothing to do any more. finished/failed before it could be cancelled.
 				// in any case, the task is removed from the TaskManager already
 				sendReleaseIntermediateResultPartitionsRpcCall();
 
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/taskexecutor/TaskExecutor.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/taskexecutor/TaskExecutor.java
@@ -640,7 +640,7 @@ public class TaskExecutor extends RpcEndpoint implements TaskExecutorGateway {
 	}
 
 	@Override
-	public void releasePartitions(Collection<ResultPartitionID> partitionIds) {
+	public void releasePartitions(JobID jobId, Collection<ResultPartitionID> partitionIds) {
 		try {
 			shuffleEnvironment.releasePartitions(partitionIds);
 		} catch (Throwable t) {
--- a/flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/ExecutionTest.java
+++ b/flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/ExecutionTest.java
@@ -19,12 +19,16 @@
 package org.apache.flink.runtime.executiongraph;
 
 import org.apache.flink.api.common.JobID;
+import org.apache.flink.api.java.tuple.Tuple2;
 import org.apache.flink.runtime.checkpoint.JobManagerTaskRestore;
 import org.apache.flink.runtime.checkpoint.TaskStateSnapshot;
 import org.apache.flink.runtime.execution.ExecutionState;
 import org.apache.flink.runtime.executiongraph.restart.NoRestartStrategy;
 import org.apache.flink.runtime.executiongraph.utils.SimpleAckingTaskManagerGateway;
 import org.apache.flink.runtime.instance.SimpleSlot;
+import org.apache.flink.runtime.io.network.partition.ResultPartitionID;
+import org.apache.flink.runtime.io.network.partition.ResultPartitionType;
+import org.apache.flink.runtime.jobgraph.DistributionPattern;
 import org.apache.flink.runtime.jobgraph.JobVertex;
 import org.apache.flink.runtime.jobgraph.JobVertexID;
 import org.apache.flink.runtime.jobmanager.scheduler.LocationPreferenceConstraint;
@@ -54,6 +58,7 @@ import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutionException;
+import java.util.function.Consumer;
 
 import static org.hamcrest.Matchers.containsInAnyOrder;
 import static org.hamcrest.Matchers.equalTo;
@@ -64,6 +69,7 @@ import static org.hamcrest.Matchers.nullValue;
 import static org.hamcrest.Matchers.sameInstance;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertThat;
 import static org.junit.Assert.assertTrue;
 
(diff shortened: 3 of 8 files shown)


flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/ExecutionTest.java
```diff
@@ -293,6 +299,81 @@ public void testSlotAllocationCancellationWhenExecutionCancelled() throws Except
...
+	/**
+	 * Tests that the partitions are released in case of a execution cancellation after the execution is already finished.
+	 */
+	@Test
+	public void testPartitionReleaseOnCancelAfterFinished() throws Exception {
+		testPartitionReleaseAfterFinished(Execution::cancel);
+	}
+
+	/**
+	 * Tests that the partitions are released in case of a execution suspension after the execution is already finished.
```
azagrebin: an execution

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/Execution.java
```diff
@@ -687,7 +687,7 @@ else if (current == RUNNING || current == DEPLOYING) {
 			}
 
 			else if (current == FINISHED || current == FAILED) {
-				// nothing to do any more. finished failed before it could be cancelled.
+				// nothing to do any more. finished/failed before it could be cancelled.
```
azagrebin: unrelated change

flink-runtime/src/main/java/org/apache/flink/runtime/taskexecutor/TaskExecutor.java
```diff
@@ -641,7 +641,7 @@ private void stopTaskExecutorServices() throws Exception {
 	}
 
 	@Override
-	public void releasePartitions(Collection<ResultPartitionID> partitionIds) {
+	public void releasePartitions(JobID jobId, Collection<ResultPartitionID> partitionIds) {
```
azagrebin: a bit weird that we will leave `jobId` here unused, is the future change that needs it going to be so big?

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/Execution.java
zentol: thats why it is in a separate hotfix commit

flink-runtime/src/main/java/org/apache/flink/runtime/taskexecutor/TaskExecutor.java
zentol: it's not gonna be big, but this part is simply already done.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/ExecutionTest.java
```diff
@@ -293,6 +299,81 @@ public void testSlotAllocationCancellationWhenExecutionCancelled() throws Except
...
+	public void testPartitionReleaseOnSuspendAfterFinished() throws Exception {
+		testPartitionReleaseAfterFinished(Execution::suspend);
+	}
+
+	private void testPartitionReleaseAfterFinished(Consumer<Execution> postFinishedExecutionAction) throws Exception {
+		final Tuple2<JobID, Collection<ResultPartitionID>> releasedPartitions = Tuple2.of(null, null);
+		final SimpleAckingTaskManagerGateway taskManagerGateway = new SimpleAckingTaskManagerGateway();
+		taskManagerGateway.setReleasePartitionsConsumer(releasedPartitions::setFields);
+
+		final SimpleSlot slot = new SimpleSlot(
```
zhijiangW: Maybe we could provide an utility for creating `SimpleSlot` in this class, because it would be reused for many tests.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/ExecutionTest.java
```diff
@@ -293,6 +299,81 @@ public void testSlotAllocationCancellationWhenExecutionCancelled() throws Except
...
+		final Execution execution = executionVertex.getCurrentExecutionAttempt();
+
+		execution.allocateAndAssignSlotForExecution(
+			slotProvider,
+			false,
+			LocationPreferenceConstraint.ALL,
+			Collections.emptySet(),
+			TestingUtils.infiniteTime());
+
+		execution.deploy();
```
zhijiangW: Maybe we could further provide an utility for creating `Execution` like below:
`
private CompletableFuture<Execution> createExecution(
		TaskManagerGateway taskManagerGateway,
		JobVertex... vertices) throws Exception {

		SimpleSlot slot = new SimpleSlot(
			new SingleSlotTestingSlotOwner(),
			new LocalTaskManagerLocation(),
			0,
			taskManagerGateway);

		ProgrammedSlotProvider slotProvider = new ProgrammedSlotProvider(1);
		slotProvider.addSlot(vertices[0].getID(), 0, CompletableFuture.completedFuture(slot));

		ExecutionGraph executionGraph = ExecutionGraphTestUtils.createSimpleTestGraph(
			new JobID(),
			slotProvider,
			new NoRestartStrategy(),
			vertices);

		executionGraph.start(TestingComponentMainThreadExecutorServiceAdapter.forMainThread());

		ExecutionJobVertex executionJobVertex = executionGraph.getJobVertex(vertices[0].getID());
		ExecutionVertex executionVertex = executionJobVertex.getTaskVertices()[0];
		Execution execution = executionVertex.getCurrentExecutionAttempt();

		CompletableFuture<Execution> allocationFuture = execution.allocateAndAssignSlotForExecution(
			slotProvider,
			false,
			LocationPreferenceConstraint.ALL,
			Collections.emptySet(),
			TestingUtils.infiniteTime());

		return allocationFuture;
	}
`
Then we could get `Execution` from future, and further get `ExecutionVertex` and `ExecutionGraph` from `Execution`. Then this helper could be reused for many existing tests.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/ExecutionTest.java
```diff
@@ -293,6 +299,81 @@ public void testSlotAllocationCancellationWhenExecutionCancelled() throws Except
...
+		execution.switchToRunning();
+
+		// simulate a case where a cancel/suspend call is too slow and the task is already finished
+		// in this case we have to explicitly release the finished partition
+		// if the task were canceled properly the TM would release the partition automatically
+		execution.markFinished();
+		postFinishedExecutionAction.accept(execution);
+
+		assertEquals(executionGraph.getJobID(), releasedPartitions.f0);
```
zhijiangW: maybe use `assertThat` instead

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/ExecutionTest.java
```diff
@@ -293,6 +299,81 @@ public void testSlotAllocationCancellationWhenExecutionCancelled() throws Except
...
+		assertEquals(executionVertex.getProducedPartitions().size(), releasedPartitions.f1.size());
+		for (ResultPartitionID partitionId : releasedPartitions.f1) {
+			IntermediateResultPartition intermediateResultPartition = executionVertex
+				.getProducedPartitions()
+				.get(partitionId.getPartitionId());
+			assertNotNull(intermediateResultPartition);
```
zhijiangW: this assert might be not necessary

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/ExecutionTest.java
```diff
@@ -293,6 +299,81 @@ public void testSlotAllocationCancellationWhenExecutionCancelled() throws Except
 		assertThat(canceledSlotRequests, equalTo(slotRequests));
 	}
 
+	/**
+	 * Tests that the partitions are released in case of an execution cancellation after the execution is already finished.
+	 */
+	@Test
+	public void testPartitionReleaseOnCancelAfterFinished() throws Exception {
```
gaoyunhaii: I think ...OnCancelingAfterFinished would be better in considering of the gramma, but I think I won't argue for that. :)

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/ExecutionTest.java
```diff
@@ -293,6 +299,81 @@ public void testSlotAllocationCancellationWhenExecutionCancelled() throws Except
...
+		testPartitionReleaseAfterFinished(Execution::cancel);
+	}
+
+	/**
+	 * Tests that the partitions are released in case of an execution suspension after the execution is already finished.
+	 */
+	@Test
+	public void testPartitionReleaseOnSuspendAfterFinished() throws Exception {
```
gaoyunhaii: Similarly, I think "...OnSuspendingAfterFinished" would be better, but I won't argue for that too.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/ExecutionTest.java
zentol: such a utility already exists in #createProgrammedSlotProvider, will update the test to use that instead

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/ExecutionTest.java
zentol: this check is done to ensure that the ids of all released partitions are actually valid. without it the test would pass even if completely random partitions were passed to the task executor.

I'll add a comment to clarify this; I had to think for a bit myself as to what we're checking here.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/ExecutionTest.java
zentol: I agree that we could de-duplicate some code here. I'm concerned that we're baking in a few assumptions (vertex parallelism should be 1, first vertex in the array is special), and there a few subtle differences in tests that where I don't know whether they are significant or not. (For example, `#testTaskRestoreStateIsNulledAfterDeployment` doesn't allocate a slot beforehand). I don't have the time right now to really look into these things, so I'd move any larger refactoring to this class into a follow-up.

PR #8586, opened 2019-05-31
[FLINK-12671][ml] Summarizer

xuyang1706:
## What is the purpose of the change

*Provide a new algorithm Summarizer(summary statistics for Table). User can easily get the total count and the basic column-wise metrics: max, min, mean, variance, standardDeviation, normL1, normL2, the number of missing values and the number of valid values.*

## Brief change log

  - *Implementation of algorithm Summarizer.*
  - *Base definition for the algorithm params.*
  - *Definition and operations for sparse/dense matrix and vector.*

## Verifying this change

*(Please pick either of the following options)*

This change is a trivial rework / code cleanup without any test coverage.

*(no)*

This change is already covered by existing tests, such as *(please describe tests)*.

*(no)*

This change added tests and can be verified as follows:

  - *Added tests for the algorithm Summarizer*
  - *Added tests for DenseVector and SparseVector*
  - *Added tests for DenseMatrix*
  - *Added tests for summary computation process*
  - *Added tests for SummarizerBatchOp*

## Does this pull request potentially affect one of the following parts:

  - Dependencies (does it add or upgrade a dependency): (no)
  - The public API, i.e., is any changed class annotated with `@Public(Evolving)`: (no)
  - The serializers: (no)
  - The runtime per-record code paths (performance sensitive): (no)
  - Anything that affects deployment or recovery: JobManager (and its components), Checkpointing, Yarn/Mesos, ZooKeeper: (no)
  - The S3 file system connector: (no)

## Documentation

  - Does this pull request introduce a new feature? (yes)
  - If yes, how is the feature documented? (not applicable / docs / JavaDocs / not documented)
      * JavaDoc

PR #8632, opened 2019-06-05
[FLINK-12744][ml] add shared params in ml package

xuyang1706:

*add ML common parameters*

## Brief change log

  - *add ML common parameters*
  - *this is sub pr of #8586*
  - *change behavior when use default constructor of param factory*
  - *add shared params in ml package*
  - *add flink-ml module*

## Verifying this change

There are definitions of parameters, mainly name, type, description and other properties. Need not test cases.

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/params/shared/colname/HasOutputCol.java
+++ b/flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/params/shared/colname/HasOutputCol.java
@@ -1,0 +1,47 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.flink.ml.params.shared.colname;
+
+import org.apache.flink.ml.api.misc.param.ParamInfo;
+import org.apache.flink.ml.api.misc.param.ParamInfoFactory;
+import org.apache.flink.ml.api.misc.param.WithParams;
+
+/**
+ * An interface for classes with a parameter specifying the name of the output column.
+ * @see HasOutputCols
+ * @see HasOutputColDefaultAsNull
+ * @see HasOutputColsDefaultAsNull
+ */
+public interface HasOutputCol<T> extends WithParams <T> {
+
+	ParamInfo <String> OUTPUT_COL = ParamInfoFactory
+		.createParamInfo("outputCol", String.class)
+		.setDescription("Name of the output column")
+		.setRequired()
+		.build();
+
+	default String getOutputCol() {
+		return get(OUTPUT_COL);
+	}
+
+	default T setOutputCol(String value) {
+		return set(OUTPUT_COL, value);
+	}
+}
--- a/flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/params/shared/colname/HasOutputColDefaultAsNull.java
+++ b/flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/params/shared/colname/HasOutputColDefaultAsNull.java
@@ -1,0 +1,47 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.flink.ml.params.shared.colname;
+
+import org.apache.flink.ml.api.misc.param.ParamInfo;
+import org.apache.flink.ml.api.misc.param.ParamInfoFactory;
+import org.apache.flink.ml.api.misc.param.WithParams;
+
+/**
+ * An interface for classes with a parameter specifying name of the output column with a null default value.
+ * @see HasOutputCol
+ * @see HasOutputCols
+ * @see HasOutputColsDefaultAsNull
+ */
+public interface HasOutputColDefaultAsNull<T> extends WithParams <T> {
+
+	ParamInfo <String> OUTPUT_COL = ParamInfoFactory
+		.createParamInfo("outputCol", String.class)
+		.setDescription("Name of the output column")
+		.setHasDefaultValue(null)
+		.build();
+
+	default String getOutputCol() {
+		return get(OUTPUT_COL);
+	}
+
+	default T setOutputCol(String value) {
+		return set(OUTPUT_COL, value);
+	}
+}
--- a/flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/params/shared/colname/HasOutputCols.java
+++ b/flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/params/shared/colname/HasOutputCols.java
@@ -1,0 +1,47 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.flink.ml.params.shared.colname;
+
+import org.apache.flink.ml.api.misc.param.ParamInfo;
+import org.apache.flink.ml.api.misc.param.ParamInfoFactory;
+import org.apache.flink.ml.api.misc.param.WithParams;
+
+/**
+ * An interface for classes with a parameter specifying names of multiple output columns.
+ * @see HasOutputCol
+ * @see HasOutputColDefaultAsNull
+ * @see HasOutputColsDefaultAsNull
+ */
+public interface HasOutputCols<T> extends WithParams <T> {
+
+	ParamInfo <String[]> OUTPUT_COLS = ParamInfoFactory
+		.createParamInfo("outputCols", String[].class)
+		.setDescription("Names of the output columns")
+		.setRequired()
+		.build();
+
+	default String[] getOutputCols() {
+		return get(OUTPUT_COLS);
+	}
+
+	default T setOutputCols(String... value) {
+		return set(OUTPUT_COLS, value);
+	}
+}
--- a/flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/params/shared/colname/HasOutputColsDefaultAsNull.java
+++ b/flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/params/shared/colname/HasOutputColsDefaultAsNull.java
@@ -1,0 +1,47 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.flink.ml.params.shared.colname;
+
+import org.apache.flink.ml.api.misc.param.ParamInfo;
+import org.apache.flink.ml.api.misc.param.ParamInfoFactory;
+import org.apache.flink.ml.api.misc.param.WithParams;
+
+/**
+ * An interface for classes with a parameter specifying names of multiple output columns. The default parameter value is null.
+ * @see HasOutputCol
+ * @see HasOutputColDefaultAsNull
+ * @see HasOutputCols
+ */
+public interface HasOutputColsDefaultAsNull<T> extends WithParams <T> {
+
+	ParamInfo <String[]> OUTPUT_COLS = ParamInfoFactory
+		.createParamInfo("outputCols", String[].class)
+		.setDescription("Names of the output columns")
+		.setHasDefaultValue(null)
+		.build();
+
+	default String[] getOutputCols() {
+		return get(OUTPUT_COLS);
+	}
+
+	default T setOutputCols(String... value) {
+		return set(OUTPUT_COLS, value);
+	}
+}
--- a/flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/params/shared/colname/HasPredictionCol.java
+++ b/flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/params/shared/colname/HasPredictionCol.java
@@ -1,0 +1,44 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.flink.ml.params.shared.colname;
+
+import org.apache.flink.ml.api.misc.param.ParamInfo;
+import org.apache.flink.ml.api.misc.param.ParamInfoFactory;
+import org.apache.flink.ml.api.misc.param.WithParams;
+
+/**
+ * An interface for classes with a parameter specifying the column name of the prediction.
+ */
+public interface HasPredictionCol<T> extends WithParams <T> {
+
+	ParamInfo <String> PREDICTION_COL = ParamInfoFactory
+		.createParamInfo("predictionCol", String.class)
+		.setDescription("Column name of prediction.")
+		.setRequired()
+		.build();
+
+	default String getPredictionCol() {
+		return get(PREDICTION_COL);
+	}
+
+	default T setPredictionCol(String value) {
+		return set(PREDICTION_COL, value);
+	}
+}
--- a/flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/params/shared/colname/HasPredictionDetailCol.java
+++ b/flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/params/shared/colname/HasPredictionDetailCol.java
@@ -1,0 +1,45 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.flink.ml.params.shared.colname;
+
+import org.apache.flink.ml.api.misc.param.ParamInfo;
+import org.apache.flink.ml.api.misc.param.ParamInfoFactory;
+import org.apache.flink.ml.api.misc.param.WithParams;
+
+/**
+ * An interface for classes with a parameter specifying the column name of prediction detail.
+ *
+ * <p>The detail is the information of prediction result, such as the probability of each label in classifier.
+ */
+public interface HasPredictionDetailCol<T> extends WithParams <T> {
+
+	ParamInfo <String> PREDICTION_DETAIL_COL = ParamInfoFactory
+		.createParamInfo("predictionDetailCol", String.class)
+		.setDescription("Column name of prediction result, it will include detailed info.")
+		.build();
+
+	default String getPredictionDetailCol() {
+		return get(PREDICTION_DETAIL_COL);
+	}
+
+	default T setPredictionDetailCol(String value) {
+		return set(PREDICTION_DETAIL_COL, value);
+	}
+}
--- a/flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/params/shared/colname/HasReservedCols.java
+++ b/flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/params/shared/colname/HasReservedCols.java
@@ -1,0 +1,44 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.flink.ml.params.shared.colname;
+
+import org.apache.flink.ml.api.misc.param.ParamInfo;
+import org.apache.flink.ml.api.misc.param.ParamInfoFactory;
+import org.apache.flink.ml.api.misc.param.WithParams;
+
+/**
+ * An interface for classes with a parameter specifying the names of the columns to be retained in the output table.
+ */
+public interface HasReservedCols<T> extends WithParams <T> {
+
+	ParamInfo <String[]> RESERVED_COLS = ParamInfoFactory
+		.createParamInfo("reservedCols", String[].class)
+		.setDescription("Names of the columns to be retained in the output table")
+		.setHasDefaultValue(null)
+		.build();
+
+	default String[] getReservedCols() {
+		return get(RESERVED_COLS);
+	}
+
+	default T setReservedCols(String... value) {
+		return set(RESERVED_COLS, value);
+	}
+}
--- a/flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/params/shared/colname/HasSelectedCol.java
+++ b/flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/params/shared/colname/HasSelectedCol.java
@@ -1,0 +1,47 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.flink.ml.params.shared.colname;
+
+import org.apache.flink.ml.api.misc.param.ParamInfo;
+import org.apache.flink.ml.api.misc.param.ParamInfoFactory;
+import org.apache.flink.ml.api.misc.param.WithParams;
+
+/**
+ * An interface for classes with a parameter specifying the name of the table column.
+ * @see HasSelectedColDefaultAsNull
+ * @see HasSelectedCols
+ * @see HasSelectedColsDefaultAsNull
+ */
+public interface HasSelectedCol<T> extends WithParams <T> {
+
+	ParamInfo <String> SELECTED_COL = ParamInfoFactory
+		.createParamInfo("selectedCol", String.class)
+		.setDescription("Name of the selected column used for processing")
+		.setRequired()
+		.build();
+
+	default String getSelectedCol() {
+		return get(SELECTED_COL);
+	}
+
+	default T setSelectedCol(String value) {
+		return set(SELECTED_COL, value);
+	}
+}
--- a/flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/params/shared/colname/HasSelectedColDefaultAsNull.java
+++ b/flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/params/shared/colname/HasSelectedColDefaultAsNull.java
@@ -1,0 +1,47 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.flink.ml.params.shared.colname;
+
+import org.apache.flink.ml.api.misc.param.ParamInfo;
+import org.apache.flink.ml.api.misc.param.ParamInfoFactory;
+import org.apache.flink.ml.api.misc.param.WithParams;
+
+/**
+ * An interface for classes with a parameter specifying the name of the table column with null default value.
+ * @see HasSelectedCol
+ * @see HasSelectedCols
+ * @see HasSelectedColsDefaultAsNull
+ */
+public interface HasSelectedColDefaultAsNull<T> extends WithParams <T> {
+
+	ParamInfo <String> SELECTED_COL = ParamInfoFactory
+		.createParamInfo("selectedCol", String.class)
+		.setDescription("Name of the selected column used for processing")
+		.setHasDefaultValue(null)
+		.build();
+
+	default String getSelectedCol() {
+		return get(SELECTED_COL);
+	}
+
+	default T setSelectedCol(String value) {
+		return set(SELECTED_COL, value);
+	}
+}
--- a/flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/params/shared/colname/HasSelectedCols.java
+++ b/flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/params/shared/colname/HasSelectedCols.java
@@ -1,0 +1,47 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.flink.ml.params.shared.colname;
+
+import org.apache.flink.ml.api.misc.param.ParamInfo;
+import org.apache.flink.ml.api.misc.param.ParamInfoFactory;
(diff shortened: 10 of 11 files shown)


flink-ml-parent/flink-ml-api/src/main/java/org/apache/flink/ml/api/misc/param/Params.java
```diff
@@ -47,14 +70,33 @@
 	 * @throws RuntimeException if the Params doesn't contains the specific parameter, while the
 	 *                          param is not optional but has no default value in the {@code info}
 	 */
-	@SuppressWarnings("unchecked")
 	public <V> V get(ParamInfo<V> info) {
-		V value = (V) paramMap.getOrDefault(info.getName(), info.getDefaultValue());
-		if (value == null && !info.isOptional() && !info.hasDefaultValue()) {
-			throw new RuntimeException(info.getName() +
-				" not exist which is not optional and don't have a default value");
+		Stream<V> stream = getParamNameAndAlias(info)
```
ex00: The `stream` is not a really good name for variable

flink-ml-parent/flink-ml-api/src/main/java/org/apache/flink/ml/api/misc/param/Params.java
```diff
@@ -47,14 +70,33 @@
...
+			.filter(this.params::containsKey)
+			.map(x -> this.params.get(x))
+			.map(x -> valueFromJson(x, info.getValueClass()))
+			.limit(1);
```
ex00: Do you need the stream object here?
How do you feel about `findFirst()` method here?

flink-ml-parent/flink-ml/src/main/java/org/apache/flink/ml/params/shared/HasVectorSizeDv100.java
```diff
@@ -0,0 +1,47 @@
...
+import org.apache.flink.ml.api.misc.param.ParamInfo;
+import org.apache.flink.ml.api.misc.param.ParamInfoFactory;
+import org.apache.flink.ml.params.BaseWithParam;
+
+/**
+ * Vector size of embedding.
+ */
+public interface HasVectorSizeDv100<T> extends BaseWithParam <T> {
+	/**
+	 * @cn embedding的向量长度
```
ex00: Could you remove it?

flink-ml-parent/flink-ml/src/main/java/org/apache/flink/ml/params/shared/colname/HasCategoricalColNames.java
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * Names of the categorical columns used for training in the input table.
+ */
+public interface HasCategoricalColNames<T> extends BaseWithParam <T> {
+
+	ParamInfo <String[]> CATEGORICAL_COL_NAMES = ParamInfoFactory
+		.createParamInfo("categoricalColNames", String[].class)
+		.setDescription("Names of the categorical columns used for training in the input table")
+		.setHasDefaultValue(new String[] {})
```
ex00: I am not sure about this default value for the possible categorical columns, could you explain why do you set this value as default?

flink-ml-parent/flink-ml-api/src/main/java/org/apache/flink/ml/api/misc/param/Params.java
```diff
@@ -93,59 +126,105 @@
...
 			}
+			return mapper.readValue(json, clazz);
 		} catch (IOException e) {
 			throw new RuntimeException("Failed to deserialize json:" + json, e);
 		}
 	}
+
+	private <V> Stream <String> getParamNameAndAlias(
+		ParamInfo <V> paramInfo) {
+		Stream <String> stream = Stream.of(paramInfo.getName());
```
ex00: the same point about the name `stream`

zentol:
Why is this even being reviewed without the PR even being linked to any JIRA?

ex00:
Hi @zentol, I found this PR in https://issues.apache.org/jira/browse/FLINK-12744

xuyang1706:
> Hi, @xuyang1706 thanks for your work!
> I left a few comments to your changes, could you look into when you will have a chance.
> Thanks!

Hi @ex00, sorry for the late reply, I just came back from my vacation. 
Thanks for your helpful comments! I have updated the code and reply them inlines.

flink-ml-parent/flink-ml-api/src/main/java/org/apache/flink/ml/api/misc/param/Params.java
xuyang1706: I changed the name to "paramValue".

flink-ml-parent/flink-ml/src/main/java/org/apache/flink/ml/params/shared/HasVectorSizeDv100.java
xuyang1706: Thanks, removed.

flink-ml-parent/flink-ml/src/main/java/org/apache/flink/ml/params/shared/colname/HasCategoricalColNames.java
xuyang1706: I removed this default value, it is not necessary and maybe confused for user.

flink-ml-parent/flink-ml-api/src/main/java/org/apache/flink/ml/api/misc/param/Params.java
xuyang1706: This name is changed to "paramNames".

flink-ml-parent/flink-ml-api/src/main/java/org/apache/flink/ml/api/misc/param/Params.java
xuyang1706: If not use stream object, we need more codes. 

For the value of parameter may be null, and the findFirst method is not friendly for the null, thus we chose the limit(.) method.

When the user sets a Null as the Value of Param, using the findFirst method in the Stream class will throw NullPointerException.
Looking deep into the code, findFirst returns an instance of Optional<T>,
but Optional<T> can only use non-null value as a constructor argument.
For example, the following code will throw NullPointerException.
<code>
    Stream.of(null, "2").findFirst();
</code>

xuyang1706:
> Why is this even being reviewed without the PR even being linked to any JIRA?

Thanks @zentol . Sorry for late reply, I just came back from my vacation.
I have updated the PR title to show the related JIRA.

c4emmmm:
@bowenli86 @walterddr @Xpray @wuchong @zentol @shaoxuan-wang @ex00 @sunjincheng121 @bluelinsky
I appreciate for your advices while reviewing flink-ml-api. Thanks for your help the pr has been approved and merged into flink.

Now I'd like to invite you to review these pr which is contributing the first library to flink-ml. As far as I know there're three PRs and the order is: 
https://github.com/apache/flink/pull/8632
https://github.com/apache/flink/pull/8631
https://github.com/apache/flink/pull/8586

Please feel free for anything you come up with. I hope this would finally be a good start for flink-ml-lib with your help.

flink-ml-parent/flink-ml/pom.xml
```diff
@@ -0,0 +1,96 @@
...
+				<configuration>
+					<source>1.8</source>
+					<target>1.8</target>
+				</configuration>
+			</plugin>
+		</plugins>
+	</build>
+
+	<properties>
+		<mvn.gson.version>2.8.2</mvn.gson.version>
```
c4emmmm: It seems the property is not necessary, it's better to remove it

flink-ml-parent/flink-ml/src/main/java/org/apache/flink/ml/params/dataproc/HasFraction.java
```diff
@@ -0,0 +1,46 @@
...
+ */
+public interface HasFraction<T> extends BaseWithParam <T> {
+	ParamInfo <Double> FRACTION = ParamInfoFactory
+		.createParamInfo("fraction", Double.class)
+		.setDescription("Proportion of data allocated to left output after splitting")
+		.setRequired()
+		.setValidator(new RangeValidator <>(0.0, 1.0))
+		.build();
+
+	default Double getFraction() {
```
c4emmmm: It's better to use double rather than boxed Double as return type unless you assume a null value is valid, especially the param has already be defined as required, which means there will be null value only if the value is set by users manually. When do you think the value will be null?

flink-ml-parent/flink-ml/src/main/java/org/apache/flink/ml/params/dataproc/HasFraction.java
```diff
@@ -0,0 +1,46 @@
...
+		return getParams().get(FRACTION);
```
c4emmmm: use get(FRACTION) instead of getParams().get(FRACTION)
I saw there are many other places do the same way, all of them should be modified like this

flink-ml-parent/flink-ml/pom.xml
```diff
@@ -0,0 +1,96 @@
...
+			<version>${project.version}</version>
+		</dependency>
+		<dependency>
+			<groupId>org.apache.flink</groupId>
+			<artifactId>flink-table-api-java-bridge_${scala.binary.version}</artifactId>
+			<version>${project.version}</version>
+		</dependency>
+		<dependency>
+			<groupId>org.apache.flink</groupId>
+			<artifactId>flink-table-planner_${scala.binary.version}</artifactId>
```
c4emmmm: The dependency of flink-table-planner is strictly forbidden, please remove the dependency and modify the code correspondingly
For more details, please refer to the mail list https://mail-archives.apache.org/mod_mbox/flink-dev/201905.mbox/browser

flink-ml-parent/flink-ml/pom.xml
```diff
@@ -0,0 +1,96 @@
...
+		<dependency>
+			<groupId>org.apache.flink</groupId>
+			<artifactId>flink-shaded-jackson</artifactId>
+			<version>${jackson.version}-${flink.shaded.version}</version>
+		</dependency>
+
+		<!-- for flink-table-planner -->
+		<dependency>
+			<groupId>com.google.guava</groupId>
+			<artifactId>guava</artifactId>
```
c4emmmm: This should be replaced with flink-shaded-guava

flink-ml-parent/flink-ml/src/main/java/org/apache/flink/ml/params/statistics/SummarizerParams.java
```diff
@@ -0,0 +1,31 @@
...
+
+package org.apache.flink.ml.params.statistics;
+
+import org.apache.flink.ml.params.shared.colname.HasSelectedColNamesDvNull;
+import org.apache.flink.ml.params.shared.colname.HasWeightColNameDvNull;
+
+/**
+ * Params of Summarizer.
+ */
+public interface SummarizerParams<T> extends
```
c4emmmm: I suppose the SummarizerParams is not a common param interface but specific for Summarizer. I think it's better to move this interface to the PR containing Summarizer.

flink-ml-parent/pom.xml
```diff
@@ -35,5 +35,6 @@ under the License.
 
 	<modules>
 		<module>flink-ml-api</module>
+		<module>flink-ml</module>
```
c4emmmm: I think here should be flink-ml-lib and the directory should also be flink-ml-lib rather than flink-ml. It's the artifact id defined in flink-ml-parent/flink-ml/pom.xml and is also the determined module name for ml libraries in FLIP-39.
Maybe all classes in this and related 2 PRs should also be moved from org.apache.flink.ml to  org.apache.flink.ml.lib package.

flink-ml-parent/flink-ml-api/src/main/java/org/apache/flink/ml/api/misc/param/Params.java
c4emmmm: paramValue here has only one element. I think it's better to replace with the only element and it will make the code below much easier.

flink-ml-parent/pom.xml
shaomengwang: There are two suggestions for the name of ml package
* as the subdirectory of flink-ml-parent，flink-ml is simpler
* the new ml package with pipeline api in spark is named ml

+1 for flink-ml

c4emmmm: > There are two suggestions for the name of ml package
> 
> * as the subdirectory of flink-ml-parent，flink-ml is simpler
> * the ml package with pipeline api in other tools is often named ml
> 
> +1 for flink-ml

Why the parent module is named flink-ml-parent is to avoid the conflict with flink-libraries/flink-ml in case of confusing users. For more details please refer to the previous discussion in https://github.com/apache/flink/pull/8402 .
In this premise, no modules should be named flink-ml again. If not so, flink-ml is better to use as the name of the parent module rather than that of a sub module.

flink-ml-parent/flink-ml/src/main/java/org/apache/flink/ml/params/BaseWithParam.java
```diff
@@ -0,0 +1,42 @@
...
+
+import org.apache.flink.ml.api.misc.param.ParamInfo;
+import org.apache.flink.ml.api.misc.param.Params;
+
+/**
+ * Interface for the object, which need set/get parameters.
+ *
+ * @param <T> the type of the object
+ */
+public interface BaseWithParam<T> {
```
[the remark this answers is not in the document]
xuyang1706: For we have refactored the API's definitions of Params and ParamInfo, the BaseWithParam could be replaced with WithParams. Thanks.

flink-ml-parent/flink-ml/pom.xml
xuyang1706: Yes, we removed it.

flink-ml-parent/flink-ml/src/main/java/org/apache/flink/ml/params/dataproc/HasFraction.java
xuyang1706: Thanks for your comment. The parameter's value might be null, we prefer using boxed Double.

flink-ml-parent/flink-ml/src/main/java/org/apache/flink/ml/params/dataproc/HasFraction.java
xuyang1706: Thanks. Directly using get(FRACTION) is more simple, we have modified the similar usages.

flink-ml-parent/flink-ml/pom.xml
xuyang1706: flink-table-planner is not necessary for this PR, we have removed it.

flink-ml-parent/flink-ml/src/main/java/org/apache/flink/ml/params/statistics/SummarizerParams.java
xuyang1706: Yes, let's discuss this params in the other PR.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/params/validators/RangeValidator.java
```diff
@@ -0,0 +1,89 @@
...
+		this.leftInclusive = tag;
+		return this;
+	}
+
+	public RangeValidator <T> withRightInclusive(boolean tag) {
+		this.rightInclusive = tag;
+		return this;
+	}
+
+	public void setLeftInclusive(boolean leftInclusive) {
```
becketqin: Should these two methods be superseded by `withLeftInclusive` and `withRightInclusive`? Any reason we need both flavors?

flink-ml-parent/flink-ml-api/src/main/java/org/apache/flink/ml/api/misc/param/Params.java
ex00: > If not use stream object, we need more codes.
> 
> For the value of parameter may be null, and the findFirst method is not friendly for the null, thus we chose the limit(.) method.
> 
> When the user sets a Null as the Value of Param, using the findFirst method in the Stream class will throw NullPointerException.
> Looking deep into the code, findFirst returns an instance of Optional,
> but Optional can only use non-null value as a constructor argument.
> For example, the following code will throw NullPointerException.
> ` Stream.of(null, "2").findFirst(); `

I suppose if we will have `null` in the stream is an invalid state, and we need to handle this situation.
anyway, we can filtering `null` before getting results, for e.g. by `.filter(Objects::nonNull)`

I agree with @becketqin  and @c4emmmm that the code without stream will be much easier to reading

xuyang1706:
> @xuyang1706 Thanks for the patch. I made a quick pass and left some comments.
> 
> I have to admit that I do not have enough expertise to say whether the various params interfaces added in `o.a.f.ml.params` are necessary or not. It will be great if you can help the reviewer understand this better. One possible way is to enhance the Java doc for them. It might also be better to check in some of the specific `WithParams` subclasses together with the algorithm that actually uses them, instead of in this PR. Doing that would make it a lot easier to understand what those `WithParams` interfaces are for.
> 
> Please let me know what do you think.

@becketqin Thanks for your comments. I removed the most of parameter info definitions, keep the common and typical ones which are helpful to understand current refactoring of ParamInfo and Params. The left parameter infos could be understandable by their name. The others will be added with according algorithms. Current codes might be easier for the reviewers. Thanks.

flink-ml-parent/flink-ml-api/src/main/java/org/apache/flink/ml/api/misc/param/Params.java
```diff
@@ -25,16 +25,39 @@
...
-public class Params implements Serializable {
-	private final Map<String, Object> paramMap = new HashMap<>();
+public class Params implements Serializable, Cloneable {
+	private static final long serialVersionUID = 1L;
+
+	private final Map<String, String> params;
+
+	private transient ObjectMapper mapper;
+
+	public Params() {
```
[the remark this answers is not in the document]
xuyang1706: Yes, we added.

flink-ml-parent/flink-ml-api/src/main/java/org/apache/flink/ml/api/misc/param/Params.java
```diff
@@ -25,16 +25,39 @@
...
  * The map-like container class for parameter. This class is provided to unify the interaction with
  * parameters.
  */
 @PublicEvolving
-public class Params implements Serializable {
-	private final Map<String, Object> paramMap = new HashMap<>();
+public class Params implements Serializable, Cloneable {
+	private static final long serialVersionUID = 1L;
+
+	private final Map<String, String> params;
```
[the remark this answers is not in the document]
xuyang1706: Thanks, we added.

flink-ml-parent/flink-ml-api/src/main/java/org/apache/flink/ml/api/misc/param/Params.java
```diff
@@ -47,14 +70,33 @@
...
+			.map(x -> valueFromJson(x, info.getValueClass()))
+			.limit(1);
+
+		if (info.isOptional()) {
+			if (info.hasDefaultValue()) {
+				return paramValue.reduce(info.getDefaultValue(), (a, b) -> b);
+			} else {
+				return paramValue.collect(Collectors.collectingAndThen(Collectors.toList(),
+					a -> {
+						if (a.isEmpty()) {
```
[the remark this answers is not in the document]
xuyang1706: Some optional parameter have no default value. For example, in the LR model prediction, user need get the predict result, and most of times, the extra predict detail info is not needed. Thus, PredResultColName is required, and PredDetailColName is optional without default value.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/params/validators/RangeValidator.java
xuyang1706: This functionality is not used by left parameter definitions, thus we removed.

flink-ml-parent/pom.xml
xuyang1706: OK, we changed to flink-ml-lib.

flink-ml-parent/flink-ml/pom.xml
xuyang1706: removed.

flink-ml-parent/flink-ml-api/src/main/java/org/apache/flink/ml/api/misc/param/Params.java
xuyang1706: Thanks for comments and code snippet. The implementation without Stream is better. For the case of specified values for multiple alias, I think it is better to throw exception.

flink-ml-parent/flink-ml-api/src/main/java/org/apache/flink/ml/api/misc/param/Params.java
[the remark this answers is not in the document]
becketqin: I am a little confused. So `PredDetailColName` is an example of optional param without default value. Not setting value for it should be a valid case. But our code will throw exception if user did not set value for such params, right?

xuyang1706:
> @xuyang1706 Thanks for updating the patch. I left a few more minor comments.
> 
> It might be important to define and explain the following things in `ParamInfo` Java doc
> 
> 1. Semantic of `optional`.
> 2. Whether a `null` value is acceptable, including for the default value. The current code does not allow user to set a parameter to null (it will be treated as not set), but the default value could be null.

OK, we have explained them in `ParamInfo` Java doc, supported 'null' value and added more test cases.

flink-ml-parent/flink-ml-api/src/main/java/org/apache/flink/ml/api/misc/param/ParamInfo.java
```diff
@@ -67,7 +67,7 @@ public String getName() {
 	 * @return the aliases of the parameter
 	 */
 	public String[] getAlias() {
-		return alias;
+		return alias == null ? new String[0] : alias;
```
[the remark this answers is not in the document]
xuyang1706: OK, added.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/params/shared/colname/HasKeepColNames.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+package org.apache.flink.ml.params.shared.colname;
+
+import org.apache.flink.ml.api.misc.param.ParamInfo;
+import org.apache.flink.ml.api.misc.param.ParamInfoFactory;
+import org.apache.flink.ml.api.misc.param.WithParams;
+
+/**
+ * Names of the columns to be retained in the output table.
```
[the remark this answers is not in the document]
xuyang1706: Thanks, added to java doc.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/params/shared/colname/HasOutputColNameDvNull.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+package org.apache.flink.ml.params.shared.colname;
+
+import org.apache.flink.ml.api.misc.param.ParamInfo;
+import org.apache.flink.ml.api.misc.param.ParamInfoFactory;
+import org.apache.flink.ml.api.misc.param.WithParams;
+
+/**
+ * Name of the output column.
```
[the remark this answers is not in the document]
xuyang1706: Thanks, added to java doc.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/params/shared/colname/HasOutputColNames.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+package org.apache.flink.ml.params.shared.colname;
+
+import org.apache.flink.ml.api.misc.param.ParamInfo;
+import org.apache.flink.ml.api.misc.param.ParamInfoFactory;
+import org.apache.flink.ml.api.misc.param.WithParams;
+
+/**
+ * Names of the output columns.
```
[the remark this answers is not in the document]
xuyang1706: Thanks, added to java doc.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/params/shared/colname/HasOutputColNamesDvNull.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+package org.apache.flink.ml.params.shared.colname;
+
+import org.apache.flink.ml.api.misc.param.ParamInfo;
+import org.apache.flink.ml.api.misc.param.ParamInfoFactory;
+import org.apache.flink.ml.api.misc.param.WithParams;
+
+/**
+ * Names of the output columns.
```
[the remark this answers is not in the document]
xuyang1706: Thanks, added to java doc.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/params/shared/colname/HasPredDetailColName.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+package org.apache.flink.ml.params.shared.colname;
+
+import org.apache.flink.ml.api.misc.param.ParamInfo;
+import org.apache.flink.ml.api.misc.param.ParamInfoFactory;
+import org.apache.flink.ml.api.misc.param.WithParams;
+
+/**
+ * Column name of predicted result, it will include detailed info.
```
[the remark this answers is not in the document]
xuyang1706: OK, we explained it in java doc.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/params/shared/colname/HasPredResultColName.java
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+package org.apache.flink.ml.params.shared.colname;
+
+import org.apache.flink.ml.api.misc.param.ParamInfo;
+import org.apache.flink.ml.api.misc.param.ParamInfoFactory;
+import org.apache.flink.ml.api.misc.param.WithParams;
+
+/**
+ * Column name of predicted result.
```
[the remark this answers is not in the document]
xuyang1706: Thanks, added to java doc.

flink-ml-parent/flink-ml-api/src/main/java/org/apache/flink/ml/api/misc/param/Params.java
```diff
@@ -44,17 +86,39 @@
...
-		V value = (V) paramMap.getOrDefault(info.getName(), info.getDefaultValue());
-		if (value == null && !info.isOptional() && !info.hasDefaultValue()) {
-			throw new RuntimeException(info.getName() +
-				" not exist which is not optional and don't have a default value");
+		String value = null;
+		String usedParamName = null;
+		for (String nameOrAlias : getParamNameAndAlias(info)) {
+			String v = params.get(nameOrAlias);
+			if (value != null && v != null) {
+				throw new IllegalArgumentException(String.format("Duplicate parameters of %s and %s",
```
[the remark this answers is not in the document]
shaomengwang: `params.put` can use `ParamInfo.getName` to get the key. It will not be able to generate duplicates. There is no information about paramName and alias at `putAll` from json. Probably `params.get` is the proper place to check duplicates

I agree with you it should be check duplicates in `ParamInfoFactory.setAlias`

flink-ml-parent/flink-ml-api/src/main/java/org/apache/flink/ml/api/misc/param/Params.java
```diff
@@ -114,38 +171,104 @@ public Params clone() {
...
+			}
+			return mapper.readValue(json, clazz);
 		} catch (IOException e) {
 			throw new RuntimeException("Failed to deserialize json:" + json, e);
 		}
 	}
+
+	private <V> List<String> getParamNameAndAlias(
+		ParamInfo <V> paramInfo) {
+		List<String> paramNames = new ArrayList <>();
```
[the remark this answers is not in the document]
xuyang1706: Thanks, changed.

flink-ml-parent/flink-ml-api/src/main/java/org/apache/flink/ml/api/misc/param/Params.java
```diff
@@ -114,38 +171,104 @@ public Params clone() {
...
 		try {
-			Map<String, String> m = mapper.readValue(json, Map.class);
-			for (Map.Entry<String, String> e : m.entrySet()) {
-				Class<?> valueClass = classMap.getOrDefault(e.getKey(), String.class);
-				paramMap.put(e.getKey(), mapper.readValue(e.getValue(), valueClass));
+			params = mapper.readValue(json, Map.class);
+		} catch (IOException e) {
+			throw new RuntimeException("Failed to deserialize json:" + json, e);
+		}
+		this.params.clear();
```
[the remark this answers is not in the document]
xuyang1706: Yes, I have changed to merge new params to exist params.

flink-ml-parent/flink-ml-api/src/main/java/org/apache/flink/ml/api/misc/param/Params.java
```diff
@@ -114,38 +171,104 @@ public Params clone() {
...
+		this.params.putAll(params);
+	}
+
+	/**
+	 * Factory method for constructing params.
+	 *
+	 * @param json the json string to load
+	 * @return the {@code Params} loaded from the json string.
+	 */
+	public static Params fromJson(String json) {
```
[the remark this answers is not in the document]
xuyang1706: We have redefined the loadJson(), and static method  fromJson() is easy to generate a new Params.

flink-ml-parent/flink-ml-api/src/main/java/org/apache/flink/ml/api/misc/param/Params.java
```diff
@@ -69,20 +133,11 @@
...
-			remove(info);
-			return this;
-		}
-
 		if (info.getValidator() != null && !info.getValidator().validate(value)) {
 			throw new RuntimeException(
 				"Setting " + info.getName() + " as a invalid value:" + value);
 		}
-		paramMap.put(info.getName(), value);
+		params.put(info.getName(), valueToJson(value));
```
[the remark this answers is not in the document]
xuyang1706: If we changed ParamInfo name, we can add the old name to aliases, thus, when we get params from old json, we could also get the param value.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/params/shared/colname/HasKeepColNames.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+package org.apache.flink.ml.params.shared.colname;
+
+import org.apache.flink.ml.api.misc.param.ParamInfo;
+import org.apache.flink.ml.api.misc.param.ParamInfoFactory;
+import org.apache.flink.ml.api.misc.param.WithParams;
+
+/**
+ * An interface for classes with a parameter specifying the names of the columns to be retained in the output table.
```
walterddr: IMO `column` doesn't need to abbreviate. 
Also I prefer `fieldName`

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/params/shared/colname/HasOutputColName.java
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+package org.apache.flink.ml.params.shared.colname;
+
+import org.apache.flink.ml.api.misc.param.ParamInfo;
+import org.apache.flink.ml.api.misc.param.ParamInfoFactory;
+import org.apache.flink.ml.api.misc.param.WithParams;
+
+/**
+ * An interface for classes with a parameter specifying the name of the output column.
```
walterddr: Reading over this comment I am not sure I understand what these interfaces are for.
Also if we are ok to use scala. we might be able to group all of these traits in one file.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/params/shared/colname/HasKeepColNames.java
```diff
@@ -0,0 +1,44 @@
+/*
```
walterddr: Can we have some example in `flink-ml-lib/src/test` ? this would be super helpful. 
Especially if one use case of "shared params" can be included.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/params/shared/colname/HasOutputColName.java
```diff
@@ -0,0 +1,47 @@
...
+ * @see HasOutputColNames
+ * @see HasOutputColNameDvNull
```
walterddr: Not sure I understand what `Dv` stands for. could we elaborate more?

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/params/shared/colname/HasOutputColName.java
c4emmmm: It's recommended to use Scala in Scala apis and Scala libraries in "Apache Flink Code Style and Quality Guide" proposed recently. I don't exactly know whether it means we should not use Scala outside these scenarios but if so it may be better to follow the suggestion and use only java if possible.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/params/shared/colname/HasOutputColName.java
c4emmmm: I suppose it's short for DefaultValue

c4emmmm:
I suggest to move the modification of Params and ParamInfo which are in the flink-ml-api module to another PR and leaving only param definition interfaces here. It follows the contributing instruction better and is indeed clearer.
Since the param definition interfaces are not affected by the modification, I think it's ok to move and then merge them separately. What do you think about it? @walterddr @ex00 @shaomengwang @xuyang1706 @becketqin

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/params/shared/colname/HasKeepColNames.java
xuyang1706: We will provide test case with algorithms later.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/params/shared/colname/HasKeepColNames.java
xuyang1706: Thanks for your advice. But the algorithms' input data type is Table, thus, we choose colName.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/params/shared/colname/HasOutputColName.java
walterddr: good point @c4emmmm . sorry for missing the proposal.

xuyang1706:
> I suggest to move the modification of Params and ParamInfo which are in the flink-ml-api module to another PR and leaving only param definition interfaces here. It follows the contributing instruction better and is indeed clearer.
> Since the param definition interfaces are not affected by the modification, I think it's ok to move and then merge them separately. What do you think about it? @walterddr @ex00 @shaomengwang @xuyang1706 @becketqin

For most of us agreed with this proposal, the modification of Params and ParamInfo are moved to a new subtask(https://issues.apache.org/jira/browse/FLINK-12881) with PR #8776. And we will update this pr when #8776 is merged.

flink-ml-parent/flink-ml-api/src/main/java/org/apache/flink/ml/api/misc/param/Params.java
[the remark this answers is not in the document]
ex00: Thanks, for explanation

flink-ml-parent/flink-ml-api/src/main/java/org/apache/flink/ml/api/misc/param/Params.java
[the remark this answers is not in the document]
ex00: thanks

xuyang1706:
@becketqin @walterddr @ex00 @c4emmmm @zentol  Thanks for all your comments. It seems all comments have been addressed, please let me know if there are more pending questions.

cc @shaoxuan-wang

ex00:
Thanks @xuyang1706 
LGFM

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/params/shared/colname/HasKeepColNames.java
walterddr: my suggestion was following the convention of [RowTypeInfo](https://github.com/apache/flink/blob/master/flink-core/src/main/java/org/apache/flink/api/java/typeutils/RowTypeInfo.java). which IMO most of the table schema is defined against. but either works fine.

shaoxuan-wang:
Thanks for the contribution, @xuyang1706 
The PR looks good to me in general. But I have a few concerns about the naming of interfaces/classes. You have used many unusual abbreviations, for instance DV=>Default value, PRED=>Predicted, in several interfaces/classes, which makes them hard to understand. My feeling is that most `DV` can be replaced by `as`? If it has to be DV, please make sure to use capital characters `DV`, as oppose to `Dv`. For the case of PRED, I do not think it is necessary to shorten the word, as it only reduces a few characters but introduces the ambiguity. What do you think?

flink-ml-parent/flink-ml-lib/pom.xml
```diff
@@ -0,0 +1,65 @@
...
+			<version>${project.version}</version>
+		</dependency>
+		<dependency>
+			<groupId>org.apache.flink</groupId>
+			<artifactId>flink-shaded-jackson</artifactId>
+			<version>${jackson.version}-${flink.shaded.version}</version>
+		</dependency>
+
+		<dependency>
+			<groupId>com.github.fommil.netlib</groupId>
```
shaoxuan-wang: We need include a license file if it has to add this dependency.

flink-ml-parent/flink-ml-lib/pom.xml
```diff
@@ -0,0 +1,65 @@
...
+	</parent>
+	<modelVersion>4.0.0</modelVersion>
+
+	<artifactId>flink-ml-lib_${scala.binary.version}</artifactId>
+	<name>flink-ml-lib</name>
+
+	<build>
+		<plugins>
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
```
shaoxuan-wang: Do we have to add this maven plugins?

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/params/shared/colname/HasKeepColNames.java
```diff
@@ -0,0 +1,44 @@
...
+package org.apache.flink.ml.params.shared.colname;
+
+import org.apache.flink.ml.api.misc.param.ParamInfo;
+import org.apache.flink.ml.api.misc.param.ParamInfoFactory;
+import org.apache.flink.ml.api.misc.param.WithParams;
+
+/**
+ * An interface for classes with a parameter specifying the names of the columns to be retained in the output table.
+ */
+public interface HasKeepColNames<T> extends WithParams <T> {
```
shaoxuan-wang: HasConstantColNames?

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/params/shared/colname/HasKeepColNames.java
```diff
@@ -0,0 +1,44 @@
...
+
+	ParamInfo <String[]> KEEP_COL_NAMES = ParamInfoFactory
+		.createParamInfo("keepColNames", String[].class)
+		.setDescription("Names of the columns to be retained in the output table")
+		.setHasDefaultValue(null)
+		.build();
+
+	default String[] getKeepColNames() {
```
shaoxuan-wang: getConstantColNames

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/params/shared/colname/HasKeepColNames.java
```diff
@@ -0,0 +1,44 @@
...
+		return get(KEEP_COL_NAMES);
+	}
+
+	default T setKeepColNames(String... value) {
```
shaoxuan-wang: setConstantColNames

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/params/shared/colname/HasPredDetailColName.java
```diff
@@ -0,0 +1,45 @@
...
+import org.apache.flink.ml.api.misc.param.ParamInfo;
+import org.apache.flink.ml.api.misc.param.ParamInfoFactory;
+import org.apache.flink.ml.api.misc.param.WithParams;
+
+/**
+ * An interface for classes with a parameter specifying the column name of predicted detailed info.
+ *
+ * <p>The detailed info is the details of prediction result, such as the probability of each label in classifier.
+ */
+public interface HasPredDetailColName<T> extends WithParams <T> {
```
shaoxuan-wang: HasPredicatedInfoDetails?

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/params/shared/colname/HasPredResultColName.java
```diff
@@ -0,0 +1,45 @@
...
+package org.apache.flink.ml.params.shared.colname;
+
+import org.apache.flink.ml.api.misc.param.ParamInfo;
+import org.apache.flink.ml.api.misc.param.ParamInfoFactory;
+import org.apache.flink.ml.api.misc.param.WithParams;
+
+/**
+ * An interface for classes with a parameter specifying the column name of the predicted result.
+ */
+public interface HasPredResultColName<T> extends WithParams <T> {
```
shaoxuan-wang: HasPredicatedResult?

flink-ml-parent/flink-ml-lib/pom.xml
xuyang1706: move to #8631

flink-ml-parent/flink-ml-lib/pom.xml
xuyang1706: removed

xuyang1706:
> I left a few suggestions about the naming. But in general I feel it is hard to understand these interfaces without the examples.

@shaoxuan-wang , thanks for your comments and suggestions. I have renamed the parameter names.

shaoxuan-wang:
Thanks for the updates, @xuyang1706 , it LGTM. I will merge it.

@flinkbot approve all

PR #8644, opened 2019-06-06
[hotfix] fix flink-sql-parser compile error

danny0405:

Fix the flink-sql-parser compile error, before this patch, the CRON build throws error: 
`Could not find goal 'regex-property' in plugin org.codehaus.mojo:build-helper-maven-plugin:1.5 among available goals reserve-network-port, parse-version, maven-version, add-test-resource, add-test-source, add-resource, attach-artifact, add-source, remove-project-artifact`

with environment:
compile - scala 2.12
JDK: openjdk8 Java
PROFILE="-Dhadoop.version=2.8.3 -Pinclude-kinesis -Dinclude_hadoop_aws -Dscala-2.12"

## Brief change log

  - *remove plugin org.codehaus.mojo:build-helper-maven-plugin*

## Verifying this change

no tests

  - The serializers: no

## Documentation

### DIFF
--- a/flink-table/flink-sql-parser/pom.xml
+++ b/flink-table/flink-sql-parser/pom.xml
@@ -222,7 +222,7 @@ under the License.
 					<dependency>
 						<groupId>org.freemarker</groupId>
 						<artifactId>freemarker</artifactId>
-						<version>2.3.25-incubating</version>
+						<version>2.3.28</version>
 					</dependency>
 				</dependencies>
 				<executions>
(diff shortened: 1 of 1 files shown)


zentol:
@KurtYoung @danny0405 Can you give me an explanation on what the issue was and how this fixes it?

danny0405:
> @KurtYoung @danny0405 Can you give me an explanation on what the issue was and how this fixes it?

@zentol It is because the regex-property goal is not configured in the original `build-helper-maven-plugin`, this goal is only used in the CRON build.

While i recheck the plugins and found that this plugin can be removed now, because we already have `javacc-maven-plugin`.

PR #8646, opened 2019-06-06
[FLINK-12735][network] Make shuffle environment implementation independent with IOManager

zhijiangW:

*The current creation of `NetworkEnvironment` relies on `IOManager` from `TaskManagerServices`. It is better not to rely on external components for constructing specific shuffle environment, so the `NetworkEnvironment` could create internal `FileChannelManager` which is already extracted from `IOManager`.*

## Brief change log

  - *Extract file channel related functions in `IOManager` to define internal `FileChannelManager` class*
  - *Make `NetworkEnvironment` creates internal `FileChannelManager` to be used for `ResultPartition`*

## Verifying this change

*covered by existing tests*

## Documentation


zhijiangW:

zhijiangW:

flink-runtime/src/main/java/org/apache/flink/runtime/io/disk/iomanager/IOManager.java
```diff
@@ -272,17 +231,91 @@ public abstract BulkBlockChannelReader createBulkBlockChannelReader(
...
+			// remove all the temp directories
+			for (File path : paths) {
+				try {
+					if (path != null) {
+						if (path.exists()) {
+							FileUtils.deleteDirectory(path);
+							LOG.info("FileChannelManager removed spill file directory {}", path.getAbsolutePath());
+						}
+					}
+				} catch (Throwable t) {
```
azagrebin: too broad exception catch, only `IOException` is expected, others are unexpected and should result in the fast global failure.

flink-runtime/src/main/java/org/apache/flink/runtime/io/disk/iomanager/IOManager.java
```diff
@@ -272,17 +231,91 @@ public abstract BulkBlockChannelReader createBulkBlockChannelReader(
...
+					throw new RuntimeException(
+						"Could not create storage directory for FileChannelManager: " + storageDir.getAbsolutePath());
+				}
+				paths[i] = storageDir;
+
+				LOG.info("FileChannelManager uses directory {} for spill files.", storageDir.getAbsolutePath());
+			}
+		}
+
+		private FileIOChannel.ID createChannel() {
```
azagrebin: nit: import for each `FileIOChannel.*` class to avoid prefixing

flink-runtime/src/main/java/org/apache/flink/runtime/io/disk/iomanager/IOManager.java
```diff
@@ -242,7 +242,7 @@ public abstract BulkBlockChannelReader createBulkBlockChannelReader(
 	/**
 	 * The manager used for creating/deleting file channels based on config temp dirs.
 	 */
-	private static class FileChannelManager {
+	public static class FileChannelManager {
```
azagrebin: I would move it into a separate class because it does not depend on `IOManager` and now also used completely independently from `IOManager`.

flink-runtime/src/main/java/org/apache/flink/runtime/io/disk/iomanager/IOManager.java
```diff
@@ -272,17 +231,91 @@ public abstract BulkBlockChannelReader createBulkBlockChannelReader(
...
+			int newNext = next + 1;
+			nextPath = newNext >= paths.length ? 0 : newNext;
+			return next;
+		}
+
+		private File[] getPaths() {
+			return paths;
+		}
+
+		private void shutdown() {
```
azagrebin: `Autoclosable` instead of `shutdown`?

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartition.java
```diff
@@ -93,12 +92,8 @@
...
 	 * Common constructor.
 	 */
-	public BoundedBlockingSubpartition(
-			int index,
-			ResultPartition parent,
-			Path filePath) throws IOException {
-
-		this(index, parent, MemoryMappedBuffers.create(filePath));
+	BoundedBlockingSubpartition(int index, ResultPartition parent) throws IOException {
+		this(index, parent, MemoryMappedBuffers.create(parent.createFileChannelPath()));
```
azagrebin: `MemoryMappedBuffers.create(parent.createFileChannelPath())`:
I think all complicated construction logic should happen in `ResultPartitionFactory`. `ResultPartition` should not provide convenience methods for construction.

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/NettyShuffleEnvironmentConfiguration.java
```diff
@@ -479,6 +491,7 @@ public int hashCode() {
 		result = 31 * result + floatingNetworkBuffersPerGate;
 		result = 31 * result + (isCreditBased ? 1 : 0);
 		result = 31 * result + (nettyConfig != null ? nettyConfig.hashCode() : 0);
+		result = 31 * result + tempDirs.hashCode();
```
azagrebin: not sure why we actually need `hashCode/equals` for `NettyShuffleEnvironmentConfiguration`,
but we should use `Arrays.hashCode()` and `Arrays.equals()` for arrays here and in `equals`.

flink-runtime/src/test/java/org/apache/flink/runtime/io/disk/iomanager/NoOpIOManager.java
```diff
@@ -1,73 +0,0 @@
...
-import org.apache.flink.runtime.util.EnvironmentInformation;
-
-import java.io.IOException;
-import java.util.List;
-import java.util.concurrent.LinkedBlockingQueue;
-
-/**
- * An {@link IOManager} that cannot do I/O but serves as a mock for tests.
- */
-public class NoOpIOManager extends IOManager {
```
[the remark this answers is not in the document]
zhijiangW: It was actually not used when I submitted this PR before. The latest merged PR brings this issue and I have not noticed this when rebase to solve the conflicts. I would adjust the commit sequence.

flink-runtime/src/main/java/org/apache/flink/runtime/io/disk/iomanager/IOManager.java
zhijiangW: Yes, it could only happen `IOException` here.

flink-runtime/src/main/java/org/apache/flink/runtime/io/disk/iomanager/IOManager.java
zhijiangW: I also considered it before. Because the `IOManager`  relies on `FileChannelManager` to implement some public methods, so I keep `FileChannelManager` inside `IOManager`. It also makes sense to move it outside.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartition.java
zhijiangW: I could refactor `parent.createFileChannelPath()` into factory while creating subpartitions and pass the path in the constructor directly

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/NettyShuffleEnvironmentConfiguration.java
zhijiangW: I also think `hashCode/equals` are not needed in this class. Or  we do not add the `tempDirs` for these methods which might be removed future in separate ticket. Or  add it as you suggested. Which one do you prefer?

azagrebin: True we can consider removal of these methods in a separate issue (we should make sure they are not needed) but then I would prefer to keep them consistent in this PR as long as they exist and add `tempDirs` using the proper array methods.

zhijiangW:
Thanks for the review @azagrebin !
I have updated the codes for addressing above comments.

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/ResultPartitionFactoryTest.java
```diff
@@ -50,7 +51,7 @@ public void testForceConsumptionOnReleaseDisabled() {
 	private static void testForceConsumptionOnRelease(boolean forceConsumptionOnRelease) {
 		ResultPartitionFactory factory = new ResultPartitionFactory(
 			new ResultPartitionManager(),
-			new NoOpIOManager(),
+			new FileChannelManager(new String[] {EnvironmentInformation.getTemporaryFileDirectory()}),
```
azagrebin: we should actually close it at the end to not leak temp folders or introduce `FileChannelManager` interface and provide testing noop implementation, also for `ResultPartitionBuilder`.

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/io/benchmark/StreamNetworkBenchmarkEnvironment.java
```diff
@@ -168,7 +163,6 @@ public void setUp(
 	public void tearDown() {
 		suppressExceptions(senderEnv::close);
 		suppressExceptions(receiverEnv::close);
-		suppressExceptions(ioManager::close);
```
azagrebin: same here, can leak tmp folders w/o close

flink-runtime/src/main/java/org/apache/flink/runtime/io/disk/iomanager/IOManager.java
```diff
@@ -120,41 +121,40 @@ public boolean isProperlyShutDown() {
...
 	/**
 	 * Deletes the file underlying the given channel. If the channel is still open, this
 	 * call may fail.
-	 * 
+	 *
 	 * @param channel The channel to be deleted.
-	 * @throws IOException Thrown if the deletion fails.
 	 */
-	public void deleteChannel(FileIOChannel.ID channel) throws IOException {
+	public void deleteChannel(ID channel) {
```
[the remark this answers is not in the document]
zhijiangW: It seems we could not call `IOManager#deleteChannel` directly atm if not constructing the `IOManager` instance. So it seems not necessary to add `static` now?

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NettyShuffleEnvironment.java
```diff
@@ -386,6 +391,9 @@ public void close() {
 				LOG.warn("Network buffer pool did not shut down properly.", t);
 			}
 
+			// delete all the temp directories
+			fileChannelManager.close();
```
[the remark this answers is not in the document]
zhijiangW: The current `close` would not throw any exceptions actually. And I could wrap it in catch block for avoiding changes future.

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/NettyShuffleEnvironmentConfiguration.java
```diff
@@ -74,7 +78,8 @@ public NettyShuffleEnvironmentConfiguration(
 			boolean isCreditBased,
 			boolean isNetworkDetailedMetrics,
 			boolean forcePartitionReleaseOnConsumption,
-			@Nullable NettyConfig nettyConfig) {
+			@Nullable NettyConfig nettyConfig,
+			@Nullable String[] tempDirs) {
```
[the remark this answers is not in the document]
zhijiangW: it should not be null in my expectation. Just copied the format from above parameter and forgot it.

flink-runtime/src/main/java/org/apache/flink/runtime/io/disk/FileChannelManager.java
```diff
@@ -0,0 +1,119 @@
...
+import java.util.UUID;
+
+/**
+ * The manager used for creating/deleting file channels based on config temp dirs.
+ */
+public class FileChannelManager implements AutoCloseable {
+	private static final Logger LOG = LoggerFactory.getLogger(FileChannelManager.class);
+
+	/** The temporary directories for files. */
+	private final File[] paths;
```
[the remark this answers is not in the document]
zhijiangW: Are there any concerns for array here?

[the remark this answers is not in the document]
azagrebin: the returned array in getter can be modified outside, because immutability cannot be enforced with arrays, alternative is to return a defensive copy

flink-runtime/src/main/java/org/apache/flink/runtime/io/disk/iomanager/IOManager.java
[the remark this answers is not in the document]
azagrebin: `static` is more restrictive like `final` so it provides more guaranties and simplifies future refactoring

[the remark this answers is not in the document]
zhijiangW: Yes, I think the other public methods in `IOManager` could also be static if so.
But I wonder it might not belong to the scope this PR. How about refactoring it if needed future?

flink-runtime/src/main/java/org/apache/flink/runtime/io/disk/FileChannelManager.java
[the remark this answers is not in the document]
zhijiangW: yes, got it.

flink-runtime/src/main/java/org/apache/flink/runtime/io/disk/iomanager/IOManager.java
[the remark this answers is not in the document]
azagrebin: I just commented because the method has been already touched and exception was removed. Other methods use instance fields or what do you mean they can be static?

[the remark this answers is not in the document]
zhijiangW: The other methods are also touched for the issue of removing the prefix import `FileIOChannel`, and the `IOException` actually happens in this method, so I removed it meanwhile.
All the public methods in `IOManager` are used by instance atm, so if we only add static for this method, it seems a bit inconsistent. This PR is for removing legacy codes, so I think the static role should not belong to this scope. WDYT?

flink-runtime/src/main/java/org/apache/flink/runtime/io/disk/FileChannelManager.java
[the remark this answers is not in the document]
zhijiangW: Considering not affecting the interface method which were already used for many places, then I take the defensive copy way here.

zhijiangW:
@azagrebin thanks for the reviews! I have updated the codes for addressing the comments and rebasing.

azagrebin:
@flinkbot approve all

flink-runtime/src/main/java/org/apache/flink/runtime/io/disk/iomanager/IOManager.java
```diff
@@ -59,46 +53,15 @@
...
-						LOG.info("I/O manager removed spill file directory {}", path.getAbsolutePath());
-					}
-				}
-			} catch (Throwable t) {
-				LOG.error("IOManager failed to properly clean up temp file directory: " + path, t);
-			}
+		try {
+			fileChannelManager.close();
+		} catch (Throwable t) {
+			LOG.warn("Cannot close file channel manager properly.", t);
```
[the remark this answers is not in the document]
zhijiangW: Yes, we could remove this log here and only catch `IOException` and throws Exception for this method.

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/NettyShuffleEnvironmentBuilder.java
```diff
@@ -55,7 +54,9 @@
 
 	private MetricGroup metricGroup = UnregisteredMetricGroups.createUnregisteredTaskManagerMetricGroup();
 
-	private IOManager ioManager = new IOManagerAsync();
+	private final String tempDir = EnvironmentInformation.getTemporaryFileDirectory();
+
+	private String[] tempDirs = new String[] {tempDir};
```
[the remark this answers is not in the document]
zhijiangW: The `tempDirs` could be reset in builder, so it could not be defined final here. Do you think it is necessary for define static for both `tempDir` and `tempDirs`?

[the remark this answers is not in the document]
azagrebin: I meant:
```
private static final String[] DEFAULT_TEMP_DIRS = new String[] {EnvironmentInformation.getTemporaryFileDirectory()};

private String[] tempDirs = DEFAULT_TEMP_DIRS;
```

zhijiangW:
I have rebased the codes and addressed the new comments.

flink-runtime/src/test/java/org/apache/flink/runtime/operators/sort/ExternalSortLargeRecordsITCase.java
```diff
@@ -71,7 +71,7 @@ public void beforeTest() {
 	}
 
 	@After
-	public void afterTest() {
+	public void afterTest() throws Exception {
 		this.ioManager.close();
 		if (!this.ioManager.isProperlyShutDown()) {
```
zentol: looking at this it makes it seem like ioManager.close should never throw exceptions. We can already see downsides of this change since, if close() throws an exception, 
a) isProperlyShutDown is never called
b) the memory manager would not be shutdown

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/ResultPartitionFactory.java
```diff
@@ -149,16 +150,15 @@ private void createSubpartitions(
 		}
 	}
 
-	private static void initializeBoundedBlockingPartitions(
+	private void initializeBoundedBlockingPartitions(
 		ResultSubpartition[] subpartitions,
-		ResultPartition parent,
-		IOManager ioManager) {
```
zentol: couldn't you pass in the channelManager instead and keep it static?

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartitionTest.java
```diff
@@ -99,8 +93,8 @@ ResultSubpartition createFailingWritesSubpartition() throws Exception {
 
 	// ------------------------------------------------------------------------
 
-	static Path tmpPath() throws IOException {
-		return new File(TMP_DIR.newFolder(), "subpartition").toPath();
+	static Path tmpPath() {
+		return fileChannelManager.createChannel().getPathFile().toPath();
```
zentol: This commit is strictly speaking not necessary, correct? If so, I'd like to exclude it to prevent merge conflict with #8880.

flink-runtime/src/test/java/org/apache/flink/runtime/operators/sort/ExternalSortLargeRecordsITCase.java
zhijiangW: we could add try catch and ignore the exception here. WDYT?

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartitionTest.java
zhijiangW: It seems a bit redundant for having both `fileChannelManager` and `TMP_DIR` meanwhile for the same function. I am not sure whether only this change would bring conflicts, because we also change the other parts of this class. If you confirm, I could revert this change.

flink-runtime/src/test/java/org/apache/flink/runtime/operators/sort/ExternalSortLargeRecordsITCase.java
zentol: it's not just this test though, there are plenty of others having the same problem.

I'm wondering what value we're adding by changing this behavior. For production at least it makes no difference since we still just log the exception.

zhijiangW: Yes, it might involve in many tests.

Actually my first version was ignoring any exceptions thrown by `IOManager#close`, and @azagrebin suggested ignoring only `IOException` and throwing any unexpected exceptions outside to make quickly fail. 

Actually I think these two options both work on my side. If we revert to previous behavior, we did not change anything during refactoring, otherwise we changed the behavior as now to do as we think in more proper way.

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartitionTest.java
zentol: It is true that there are likely other merge conflicts too, but I disagree with your implied conclusion that this means we don't have to worry about them. More merge conflicts => more time spent on fixing them.

I'm not saying that we shouldn't make this change overall, but let's do it at a later date, when merging it doesn't interfere with other people's critical work.

flink-runtime/src/test/java/org/apache/flink/runtime/operators/sort/ExternalSortLargeRecordsITCase.java
zentol: most of the test were failing anyway if the ioManager shutdown failed due to subsequent assertions on `isProperlyShutdown`; thus b) is no longer a new problem.

I would suggest to go through all tests, remove the `isProperlyShutDown` call if the test follows the same pattern, and potentially remove the method altogether if it is unused at that point.

zhijiangW: Agree, that is a good finding. 
I could solve this previous existing issue in a separate hotfix commit as precondition.

zhijiangW: After thinking through it further, we might not remove the `isProperlyShutDown` call directly, because it is used for checking previous `IOException` which was swallowed during `close`. And `close` itself could only thrown non `IOException` atm.

I also do not like the way of `isProperlyShutDown`, if we want to remove it completely, another option is let `IOManager#close` also throw `IOException` after all the file delete loop. WDYT?

zhijiangW: After confirming with andrey, we could keep the previous behavior to still ignore `IOException` during close, because we just try best to delete files for releasing resources. If one file was already deleted by other factors before, it is no need to cause unnecessary failover. 

If we want to refactor this behavior future, we could do it separately, not to block this PR. I remove `try catch` from `IOManager#close` and add some javadoc for this method.

zhijiangW:
Thanks for the review @zentol !
I have rebased the codes and addressed the above comments.

StephanEwen:
Sorry for jumping in late.

Initially, we added this dependency to make it possible for spilling partitions to use write-behind and separate pre-fetching threads. Do we want to drop this completely?

Also, when it comes to reducing dependencies in the network shuffle environment, what is the gain by replacing IOManager with FileChannelManager?
If this is mainly about the threads that the IOManagerAsync spawns, another option could be to has an IOManagerSync without threads.

zhijiangW:
@StephanEwen thanks for sharing the background and bringing this concern!

The abstract `IOManager` actually has two roles:
- One is like the utility tool for getting file channels which is currently used for some streaming component, like `StateBackend, `BarrierBuffer`, `ShuffleService`.
- The other one is real abstract methods for handing read/write operations which is mainly used for batch in sort-merge.

If we assume that one component should only focus on one single function, it is reasonable to break down `IOManager` into two sub components as now. The other components could only rely on `FileChannelManager` instead of heavy-weight `IOManager`.

Currently the shuffle service would only need internal `FileChannelManager`, I agree with that it might rely on the functions of write-behind or pre-fetching threads provided by `IOManager` future. If so, the other shuffle service implementations could implement a new `IOManager` instance or using current `IOManagerAsync` directly if needed future.

Another option as you suggested is implementing a new `IOManager` instance without spawning threads, which also makes sense. If the shuffle service would need the read/writer functions future, it has to replace another `IOManager` implementation. But I still prefer the current way in PR, because the individual two sub components are both light-weight and they could be used together or separately.

azagrebin:
@StephanEwen Thanks for bringing this up.

True, the idea was to drop `IOManager` dependency in Shuffle Service API because `NettyShuffleEnviroment` basically uses only narrowed scope of IOManager, namely switching between temporary files. The narrowed concern of switching between temporary files is moved into `FileChannelManager` which is also used by the `IOManager`. `NettyShuffleEnviroment` can create its own light-weight  `FileChannelManager` with its own directory prefix.

The valid concern is that if we decide to use `IOManagerAsync` with its threads then the threads should be shared by all IO components which use them. In that case Shuffle Service API has to accept it as a dependency created outside.

As discussed offline, if we decide later that we need `IOManagerAsync` for shuffling then we can revert the change and add the abstract/interface `IOManager` back to `ShuffleEnvironmentContext`.

zhijiangW:
@azagrebin @zentol I submitted another hotfix commit for removing `IOManager#isProperlyShutDown`.

flink-runtime/src/main/java/org/apache/flink/runtime/io/disk/iomanager/IOManagerAsync.java
```diff
@@ -126,22 +131,20 @@ public void close() {
...
+				.map(IOManagerAsync::getWriterThreadCloser)
+				.collect(Collectors.toList());
+
+			Iterable<AutoCloseable> readerThreadCloseables = Arrays.stream(readers)
+				.filter(Objects::nonNull)
+				.filter(Thread::isAlive)
+				.map(IOManagerAsync::getReaderThreadCloser)
+				.collect(Collectors.toList());
+
+			IOUtils.closeAll(writerThreadCloseables, readerThreadCloseables);
```
azagrebin: Instead of modifying `IOUtils.closeAll` method, I would suggest the following approach:
```
        // close writing and reading threads with best effort and log problems
		// first notify all to close, then wait until all are closed

		List<AutoCloseable> closables = new ArrayList<>(2 * writers.length + 2 * readers.length + 1);

		Arrays.stream(writers)
			.filter(Objects::nonNull)
			.filter(Thread::isAlive)
			.map(IOManagerAsync::getWriterThreadCloser)
			.forEach(closables::add);

		Arrays.stream(readers)
			.filter(Objects::nonNull)
			.filter(Thread::isAlive)
			.map(IOManagerAsync::getReaderThreadCloser)
			.forEach(closables::add);

		for (WriterThread wt : writers) {
			closables.add(wt::join);
		}

		for (ReaderThread rt : readers) {
			closables.add(rt::join);
		}

		// make sure we call the super implementation in any case and at the last point,
		// because this will clean up the I/O directories
		closables.add(super::close);

		IOUtils.closeAll(closables);
```
This way we also try to close everything

flink-runtime/src/main/java/org/apache/flink/runtime/io/disk/iomanager/IOManagerAsync.java
```diff
@@ -160,33 +163,31 @@ public void close() {
...
+				throw new IOException("Error while shutting down IO Manager reader thread.", t);
+			}
+		};
+	}
 
 	@Override
 	public void uncaughtException(Thread t, Throwable e) {
 		LOG.error("IO Thread '" + t.getName() + "' terminated due to an exception. Shutting down I/O Manager.", e);
-		close();
+		IOUtils.closeQuietly(this);
```
azagrebin: I think we can log the closing exception here as well

flink-runtime/src/main/java/org/apache/flink/runtime/io/disk/iomanager/IOManagerAsync.java
```diff
@@ -126,22 +131,20 @@ public void close() {
...
-				try {
-					rt.shutdown();
-				}
-				catch (Throwable t) {
-					LOG.error("Error while shutting down IO Manager reader thread.", t);
-				}
-			}
+			Iterable<AutoCloseable> writerThreadCloseables = Arrays.stream(writers)
+				.filter(Objects::nonNull)
+				.filter(Thread::isAlive)
```
azagrebin: do we need `Objects::nonNull` and `Thread::isAlive` filters here?

flink-runtime/src/main/java/org/apache/flink/runtime/io/disk/iomanager/IOManagerAsync.java
zhijiangW: fixed

flink-runtime/src/main/java/org/apache/flink/runtime/io/disk/iomanager/IOManagerAsync.java
zhijiangW: fixed

flink-runtime/src/main/java/org/apache/flink/runtime/io/disk/iomanager/IOManagerAsync.java
zhijiangW: `nonNull` is not necessary, `isAlive` might make sense to avoid the thread was already dead unexpected. WDYT?

flink-runtime/src/main/java/org/apache/flink/runtime/io/disk/iomanager/IOManagerAsync.java
```diff
@@ -118,75 +120,61 @@ public void shutdown() {
...
-	 * @return True, if the IO manager has properly shut down, false otherwise.
-	 */
-	@Override
-	public boolean isProperlyShutDown() {
-		boolean readersShutDown = true;
+
 		for (ReaderThread rt : readers) {
-			readersShutDown &= rt.getState() == Thread.State.TERMINATED;
+			closeables.add(getReaderThreadCloser(rt));
+			closeables.add(rt::join);
```
zentol: join is already called in (Reader/Writer)Thread.shutdown . We likely don't want to wait indefinitely in any case, particularly since we're looping over these and would swallow InterruptedExceptions in `IOUtils#closeAll`..

zhijiangW: I think you are right. I have not looked through the implementation in `shutdown` before, just keep the same behavior as before, which also call `join` after `shutdown`.
I could improve this previous issue in a hotfix commit later.

PR #8647, opened 2019-06-06
[FLINK-12774][build] Pin version of build-helper-maven-plugin

zentol:

Pin version of build-helper-maven-plugin to 1.7 to reduce complexity in the plugins.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -690,7 +690,6 @@ under the License.
 					<plugin>
 						<groupId>org.codehaus.mojo</groupId>
 						<artifactId>build-helper-maven-plugin</artifactId>
-						<version>1.12</version>
 						<executions>
 							<execution>
 								<id>regex-property</id>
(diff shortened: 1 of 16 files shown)


tzulitai:
@flinkbot approve description
@flinkbot approve consensus

tzulitai:
I think the issue tag is incorrect. It should be FLINK-12739.

pom.xml
```diff
@@ -744,6 +743,11 @@ under the License.
 			<build>
 				<pluginManagement>
 					<plugins>
+						<plugin>
+							<groupId>org.codehaus.mojo</groupId>
+							<artifactId>build-helper-maven-plugin</artifactId>
+							<version>1.7</version>
```
tzulitai: Why are do we pin it to 1.7, and not 1.12 (the version that was previously used when the Scala 2.12 profile is enabled)?

zentol: because quite a few modules, primarily scala reliant ones, were overriding the version to 1.7 already, and it apparently did not cause issues.

PR #8648, opened 2019-06-06
[FLINK-12738][network] Remove abstract getPageSize method from InputGate

zhijiangW:

*Currently `InputGate#getPageSize` is only used for constructing `BarrierBuffer`. In order to make abstract `InputGate` simple and clean, we should remove unnecessary abstract methods from it.
Considering the page size could be parsed directly from configuration which could also visible while constructing `BarrierBuffer`, so it is reasonable to do so.*

## Brief change log

  - *Remove `getPageSize` method from `InputGate`*
  -  *Remove `getMemorySegmentSize` method from `BufferProvider` interface*
  - *Remove unnecessary `getMemorySegmentSize` method from `NetworkBufferPool`*

## Verifying this change

*Covered by existing tests*

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/InputProcessorUtil.java
+++ b/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/InputProcessorUtil.java
@@ -25,6 +25,7 @@ import org.apache.flink.configuration.TaskManagerOptions;
 import org.apache.flink.runtime.io.disk.iomanager.IOManager;
 import org.apache.flink.runtime.io.network.partition.consumer.InputGate;
 import org.apache.flink.runtime.jobgraph.tasks.AbstractInvokable;
+import org.apache.flink.runtime.util.ConfigurationParserUtils;
 import org.apache.flink.streaming.api.CheckpointingMode;
 
 import java.io.IOException;
@@ -46,8 +47,10 @@ public class InputProcessorUtil {
 			Configuration taskManagerConfig,
 			String taskName) throws IOException {
 
+		int pageSize = ConfigurationParserUtils.getPageSize(taskManagerConfig);
+
 		BufferStorage bufferStorage = createBufferStorage(
-			checkpointMode, ioManager, inputGate.getPageSize(), taskManagerConfig, taskName);
+			checkpointMode, ioManager, pageSize, taskManagerConfig, taskName);
 		CheckpointBarrierHandler barrierHandler = createCheckpointBarrierHandler(
 			checkpointMode, inputGate.getNumberOfInputChannels(), taskName, toNotifyOnCheckpoint);
 		return new CheckpointedInputGate(inputGate, bufferStorage, barrierHandler);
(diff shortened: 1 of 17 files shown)


zhijiangW:

zhijiangW:

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/InputProcessorUtil.java
```diff
@@ -53,16 +54,17 @@ public static CheckpointBarrierHandler createCheckpointBarrierHandler(
 					+ " must be positive or -1 (infinite)");
 			}
 
+			int pageSize = ConfigurationParserUtils.getPageSize(taskManagerConfig);
 			if (taskManagerConfig.getBoolean(NettyShuffleEnvironmentOptions.NETWORK_CREDIT_MODEL)) {
 				barrierHandler = new BarrierBuffer(
 					inputGate,
-					new CachedBufferBlocker(inputGate.getPageSize()),
+					new CachedBufferBlocker(pageSize),
```
zentol: Doesn't this leave us with an implicit assumption that the pagesize in the input gate corresponds to the configured page size?

What would happen if this would no longer hold?

azagrebin: True, actually we do the implicit coupling between `taskmanager.memory.segment-size` enforcing all shuffle services to use it with no option to use a separate page size. If we do it with `ShuffleEnvironment#getPageSize` then it will be the same for all gates. Not sure whether we should touch it then at all, actually having it in input gate is probably a right thing. Or maybe page size should somehow come from buffers but I do not see how it is possible.

@pnowojski might have more input here.

zhijiangW: Yes, it might bring this concern here. 
My initial version was removing this page size parameter from constructor of `CachedBufferBlocker`. The cached size could be got directly from `BufferOrEvent#getSize`. But it might bring previous behavior because the real size in buffer might be less than the configured page size. But it seems more reasonable to get cached size from data itself.

zhijiangW: For the case of `BufferSpiller` which uses the page size for requesting buffer to fill in disk data. Actually this page size is not relevant with the network buffer size. The default 32K network buffer is only suitable for network transport. But for reading disk data it should not be limited to the default 32K network buffer size. Of course it could simply take this size as default, but I think it has no implicit assumption concern here.  Just like current `IOManager` and `MemoryManager` are both using segment size from the same config option. 

The page size for shuffle service should not be exposed to outside, because it is only used for internal buffer pool and network transport. For other shuffle implementations like DfsShuffleService, it might not need the network buffer pool and no page size at all.

pnowojski: I think this shouldn't be an issue. Parameter passed to `CachedBufferBlocker` is just an estimation and the one passed to `BufferSpiller` is independent of anything else from the network stack.

zhijiangW:
@pnowojski , the travis is green now.

PR #8654, opened 2019-06-06
[FLINK-12647][network] Add feature flag to disable release of consumed blocking partitions

zentol:

Based on #8608; only the last 2 commits are relevant.

The first commit introduces the concept of externally managed partitions, i.e. partitions that are not automatically released by the ShuffleEnvironment on consumption, but instead when instructed to do so by the TaskManager, or more generally as part of the partition lifecycle management. On it's own this commit does not change any behavior, as no part of the system makes use of this flag.
This is primarily a preparatory step for the subsequent commit and future partition lifecycle management.

The second commit introduces a dev feature flag to enable/disable the automatic release of consumed blocking result partitions by the shuffle environment.

If the flag is enabled, which is the default, then the system does not behave differently to how it does currently.
Once the partition lifecycle is fully introduced we will either switch the default to false, or remove the flag entirely.

### DIFF
--- a/flink-core/src/main/java/org/apache/flink/configuration/NettyShuffleEnvironmentOptions.java
+++ b/flink-core/src/main/java/org/apache/flink/configuration/NettyShuffleEnvironmentOptions.java
@@ -21,6 +21,7 @@ package org.apache.flink.configuration;
 import org.apache.flink.annotation.PublicEvolving;
 import org.apache.flink.annotation.docs.ConfigGroup;
 import org.apache.flink.annotation.docs.ConfigGroups;
+import org.apache.flink.annotation.docs.Documentation;
 
 import static org.apache.flink.configuration.ConfigOptions.key;
 
@@ -208,6 +209,11 @@ public class NettyShuffleEnvironmentOptions {
 			.defaultValue(10000)
 			.withDeprecatedKeys("taskmanager.net.request-backoff.max")
 			.withDescription("Maximum backoff in milliseconds for partition requests of input channels.");
+
+	@Documentation.ExcludeFromDocumentation("dev use only; likely temporary")
+	public static final ConfigOption<Boolean> FORCE_PARTITION_RELEASE_ON_CONSUMPTION =
+			key("taskmanager.network.partition.force-release-on-consumption")
+			.defaultValue(true);
 
 	// ------------------------------------------------------------------------
 
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/deployment/ResultPartitionDeploymentDescriptor.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/deployment/ResultPartitionDeploymentDescriptor.java
@@ -48,6 +48,9 @@ public class ResultPartitionDeploymentDescriptor implements Serializable {
 	/** Flag whether the result partition should send scheduleOrUpdateConsumer messages. */
 	private final boolean sendScheduleOrUpdateConsumersMessage;
 
+	/** Whether the result partition is released on consumption. */
+	private final boolean releasedOnConsumption;
+
 	public ResultPartitionDeploymentDescriptor(
 			PartitionDescriptor partitionDescriptor,
 			ShuffleDescriptor shuffleDescriptor,
@@ -58,6 +61,7 @@ public class ResultPartitionDeploymentDescriptor implements Serializable {
 		KeyGroupRangeAssignment.checkParallelismPreconditions(maxParallelism);
 		this.maxParallelism = maxParallelism;
 		this.sendScheduleOrUpdateConsumersMessage = sendScheduleOrUpdateConsumersMessage;
+		this.releasedOnConsumption = partitionDescriptor.getPartitionType() != ResultPartitionType.BLOCKING;
 	}
 
 	public IntermediateDataSetID getResultId() {
@@ -86,6 +90,10 @@ public class ResultPartitionDeploymentDescriptor implements Serializable {
 
 	public boolean sendScheduleOrUpdateConsumersMessage() {
 		return sendScheduleOrUpdateConsumersMessage;
+	}
+
+	public boolean isReleasedOnConsumption() {
+		return releasedOnConsumption;
 	}
 
 	@Override
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/ReleaseOnConsumptionResultPartition.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/ReleaseOnConsumptionResultPartition.java
@@ -1,0 +1,105 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.runtime.io.network.partition;
+
+import org.apache.flink.runtime.io.network.buffer.BufferPool;
+import org.apache.flink.runtime.io.network.buffer.BufferPoolOwner;
+import org.apache.flink.util.function.FunctionWithException;
+
+import java.io.IOException;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import static org.apache.flink.util.Preconditions.checkState;
+
+/**
+ * ResultPartition that releases itself once all subpartitions have been consumed.
+ */
+public class ReleaseOnConsumptionResultPartition extends ResultPartition {
+
+	/**
+	 * The total number of references to subpartitions of this result. The result partition can be
+	 * safely released, iff the reference count is zero. A reference count of -1 denotes that the
+	 * result partition has been released.
+	 */
+	private final AtomicInteger pendingReferences = new AtomicInteger();
+
+	ReleaseOnConsumptionResultPartition(
+			String owningTaskName,
+			ResultPartitionID partitionId,
+			ResultPartitionType partitionType,
+			ResultSubpartition[] subpartitions,
+			int numTargetKeyGroups,
+			ResultPartitionManager partitionManager,
+			FunctionWithException<BufferPoolOwner, BufferPool, IOException> bufferPoolFactory) {
+		super(owningTaskName, partitionId, partitionType, subpartitions, numTargetKeyGroups, partitionManager, bufferPoolFactory);
+	}
+
+	/**
+	 * The partition can only be released after each subpartition has been consumed once per pin operation.
+	 */
+	@Override
+	void pin() {
+		while (true) {
+			int refCnt = pendingReferences.get();
+
+			if (refCnt >= 0) {
+				if (pendingReferences.compareAndSet(refCnt, refCnt + subpartitions.length)) {
+					break;
+				}
+			}
+			else {
+				throw new IllegalStateException("Released.");
+			}
+		}
+	}
+
+	@Override
+	public ResultSubpartitionView createSubpartitionView(int index, BufferAvailabilityListener availabilityListener) throws IOException {
+		int refCnt = pendingReferences.get();
+
+		checkState(refCnt != -1, "Partition released.");
+		checkState(refCnt > 0, "Partition not pinned.");
+
+		return super.createSubpartitionView(index, availabilityListener);
+	}
+
+	@Override
+	void onConsumedSubpartition(int subpartitionIndex) {
+		if (isReleased()) {
+			return;
+		}
+
+		int refCnt = pendingReferences.decrementAndGet();
+
+		if (refCnt == 0) {
+			partitionManager.onConsumedPartition(this);
+		} else if (refCnt < 0) {
+			throw new IllegalStateException("All references released.");
+		}
+
+		LOG.debug("{}: Received release notification for subpartition {}.",
+			this, subpartitionIndex);
+	}
+
+	@Override
+	public String toString() {
+		return "ReleaseOnConsumptionResultPartition " + partitionId.toString() + " [" + partitionType + ", "
+			+ subpartitions.length + " subpartitions, "
+			+ pendingReferences + " pending references]";
+	}
+}
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/ResultPartition.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/ResultPartition.java
@@ -38,7 +38,6 @@ import javax.annotation.Nullable;
 
 import java.io.IOException;
 import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.concurrent.atomic.AtomicInteger;
 
 import static org.apache.flink.util.Preconditions.checkArgument;
 import static org.apache.flink.util.Preconditions.checkElementIndex;
@@ -71,32 +70,25 @@ import static org.apache.flink.util.Preconditions.checkState;
  */
 public class ResultPartition implements ResultPartitionWriter, BufferPoolOwner {
 
-	private static final Logger LOG = LoggerFactory.getLogger(ResultPartition.class);
+	protected static final Logger LOG = LoggerFactory.getLogger(ResultPartition.class);
 
 	private final String owningTaskName;
 
-	private final ResultPartitionID partitionId;
+	protected final ResultPartitionID partitionId;
 
 	/** Type of this partition. Defines the concrete subpartition implementation to use. */
-	private final ResultPartitionType partitionType;
+	protected final ResultPartitionType partitionType;
 
 	/** The subpartitions of this partition. At least one. */
-	private final ResultSubpartition[] subpartitions;
+	protected final ResultSubpartition[] subpartitions;
 
-	private final ResultPartitionManager partitionManager;
+	protected final ResultPartitionManager partitionManager;
 
 	public final int numTargetKeyGroups;
 
 	// - Runtime state --------------------------------------------------------
 
 	private final AtomicBoolean isReleased = new AtomicBoolean();
-
-	/**
-	 * The total number of references to subpartitions of this result. The result partition can be
-	 * safely released, iff the reference count is zero. A reference count of -1 denotes that the
-	 * result partition has been released.
-	 */
-	private final AtomicInteger pendingReferences = new AtomicInteger();
 
 	private BufferPool bufferPool;
 
@@ -281,11 +273,6 @@ public class ResultPartition implements ResultPartitionWriter, BufferPoolOwner {
 	 * Returns the requested subpartition.
 	 */
 	public ResultSubpartitionView createSubpartitionView(int index, BufferAvailabilityListener availabilityListener) throws IOException {
-		int refCnt = pendingReferences.get();
-
-		checkState(refCnt != -1, "Partition released.");
-		checkState(refCnt > 0, "Partition not pinned.");
-
 		checkElementIndex(index, subpartitions.length, "Subpartition not found.");
 
 		ResultSubpartitionView readView = subpartitions[index].createReadView(availabilityListener);
@@ -337,31 +324,15 @@ public class ResultPartition implements ResultPartitionWriter, BufferPoolOwner {
 	@Override
 	public String toString() {
 		return "ResultPartition " + partitionId.toString() + " [" + partitionType + ", "
-				+ subpartitions.length + " subpartitions, "
-				+ pendingReferences + " pending references]";
+				+ subpartitions.length + " subpartitions]";
 	}
 
 	// ------------------------------------------------------------------------
 
 	/**
 	 * Pins the result partition.
-	 *
-	 * <p>The partition can only be released after each subpartition has been consumed once per pin
-	 * operation.
 	 */
 	void pin() {
-		while (true) {
-			int refCnt = pendingReferences.get();
-
-			if (refCnt >= 0) {
-				if (pendingReferences.compareAndSet(refCnt, refCnt + subpartitions.length)) {
-					break;
-				}
-			}
-			else {
-				throw new IllegalStateException("Released.");
-			}
-		}
 	}
 
 	/**
@@ -373,17 +344,8 @@ public class ResultPartition implements ResultPartitionWriter, BufferPoolOwner {
 			return;
 		}
 
-		int refCnt = pendingReferences.decrementAndGet();
-
-		if (refCnt == 0) {
-			partitionManager.onConsumedPartition(this);
-		}
-		else if (refCnt < 0) {
-			throw new IllegalStateException("All references released.");
-		}
-
-		LOG.debug("{}: Received release notification for subpartition {} (reference count now at: {}).",
-				this, subpartitionIndex, pendingReferences);
+		LOG.debug("{}: Received release notification for subpartition {}.",
+				this, subpartitionIndex);
 	}
 
 	public ResultSubpartition[] getAllPartitions() {
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/ResultPartitionFactory.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/ResultPartitionFactory.java
@@ -57,18 +57,22 @@ public class ResultPartitionFactory {
 
 	private final int floatingNetworkBuffersPerGate;
 
+	private final boolean forcePartitionReleaseOnConsumption;
+
 	public ResultPartitionFactory(
 		@Nonnull ResultPartitionManager partitionManager,
 		@Nonnull IOManager ioManager,
 		@Nonnull BufferPoolFactory bufferPoolFactory,
 		int networkBuffersPerChannel,
-		int floatingNetworkBuffersPerGate) {
+		int floatingNetworkBuffersPerGate,
+		boolean forcePartitionReleaseOnConsumption) {
 
 		this.partitionManager = partitionManager;
 		this.ioManager = ioManager;
 		this.networkBuffersPerChannel = networkBuffersPerChannel;
 		this.floatingNetworkBuffersPerGate = floatingNetworkBuffersPerGate;
 		this.bufferPoolFactory = bufferPoolFactory;
+		this.forcePartitionReleaseOnConsumption = forcePartitionReleaseOnConsumption;
 	}
 
 	public ResultPartition create(
@@ -82,6 +86,7 @@ public class ResultPartitionFactory {
 			desc.getPartitionType(),
 			desc.getNumberOfSubpartitions(),
 			desc.getMaxParallelism(),
+			desc.isReleasedOnConsumption() || forcePartitionReleaseOnConsumption,
 			createBufferPoolFactory(desc.getNumberOfSubpartitions(), desc.getPartitionType()));
 	}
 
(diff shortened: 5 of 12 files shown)


flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/ResultPartition.java
```diff
@@ -80,6 +80,8 @@
 	/** Type of this partition. Defines the concrete subpartition implementation to use. */
 	private final ResultPartitionType partitionType;
 
+	private boolean isManagedExternally;
```
zhijiangW: final

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/ResultPartition.java
```diff
@@ -180,6 +184,10 @@ public ResultPartitionType getPartitionType() {
 		return partitionType;
 	}
 
+	public boolean isManagedExternally() {
```
zhijiangW: could be package private

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/ResultPartitionManager.java
```diff
@@ -38,8 +38,18 @@
 
 	private final Map<ResultPartitionID, ResultPartition> registeredPartitions = new HashMap<>(16);
 
+	private final boolean releaseExternallyManagedPartitionsOnConsumption;
+
 	private boolean isShutdown;
 
+	public ResultPartitionManager() {
```
zhijiangW: @VisibleForTesting

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/NettyShuffleEnvironmentConfiguration.java
```diff
@@ -122,6 +126,10 @@ public boolean isNetworkDetailedMetrics() {
 		return isNetworkDetailedMetrics;
 	}
 
+	public boolean isReleaseExternallyManagedPartitionsOnConsumption() {
```
zhijiangW: May be better to unify the method with variable name. 
`isReleasePartititionsOnConsumption| isReleaseBlockingPartitionsOnConsumption| isReleaseExternallyManagedPartitionsOnConsumption`

IMO `isReleasePartititionsOnConsumption` might be simple and proper, no need to concern on whether it is blocking or external.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/ResultPartitionManager.java
```diff
@@ -38,8 +38,18 @@
 
 	private final Map<ResultPartitionID, ResultPartition> registeredPartitions = new HashMap<>(16);
 
+	private final boolean releaseExternallyManagedPartitionsOnConsumption;
```
zhijiangW: Might be named `isReleasePartitionsOnConsumption`.

flink-runtime/src/main/java/org/apache/flink/runtime/deployment/ResultPartitionDeploymentDescriptor.java
```diff
@@ -58,6 +72,7 @@ public ResultPartitionDeploymentDescriptor(
 		KeyGroupRangeAssignment.checkParallelismPreconditions(maxParallelism);
 		this.maxParallelism = maxParallelism;
 		this.sendScheduleOrUpdateConsumersMessage = sendScheduleOrUpdateConsumersMessage;
+		this.isManagedExternally = partitionDescriptor.getPartitionType() == ResultPartitionType.BLOCKING;
```
zhijiangW: Since this property could be got directly based on `ResultPartitionType`, then another option is not defining this property explicitly in `ResultPartitionDeploymentDescriptor` and `ResultPartition` could also provide this property based on internal `ResultPartitionType`.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/ResultPartition.java
```diff
@@ -120,6 +123,7 @@ public ResultPartition(
 		this.partitionType = checkNotNull(partitionType);
 		this.subpartitions = checkNotNull(subpartitions);
 		this.numTargetKeyGroups = numTargetKeyGroups;
+		this.isManagedExternally = isManagedExternally;
```
zhijiangW: As mentioned above, `this.isManagedExternally = (partitionType == ResultPartitionType.BLOCKING)`

flink-core/src/main/java/org/apache/flink/configuration/NettyShuffleEnvironmentOptions.java
```diff
@@ -209,6 +210,11 @@
 			.withDeprecatedKeys("taskmanager.net.request-backoff.max")
 			.withDescription("Maximum backoff in milliseconds for partition requests of input channels.");
 
+	@Documentation.ExcludeFromDocumentation("dev use only; likely temporary")
+	public static final ConfigOption<Boolean> RELEASE_EXTERNALLY_MANAGED_PARTITIONS_ON_CONSUMPTION =
+			key("taskmanager.network.partition.release-externally-managed-on-consumption")
```
zhijiangW: key("taskmanager.network.release-externally-managed-partitions-on-consumption") to consistent with var naming.

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/ResultPartitionManagerTest.java
```diff
@@ -60,4 +61,26 @@ public void testCreateViewForRegisteredPartition() throws Exception {
...
+		final ResultPartition partition = new ResultPartitionBuilder().setIsExternallyManaged(true).build();
+
+		partitionManager.registerResultPartition(partition);
+		partitionManager.onConsumedPartition(partition);
+
+		assertThat(partition.isReleased(), is(false));
+	}
+
+	@Test
+	public void testExternallyManagedPartitionReleaseIfFlagEnabled() {
```
zhijiangW: These two tests could be deduplicated by extracting a boolean value if you like.

flink-core/src/main/java/org/apache/flink/configuration/NettyShuffleEnvironmentOptions.java
zentol: We don't have to be consistent with the variable name. The `partition` is already contained in the key prefix.

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/NettyShuffleEnvironmentConfiguration.java
zentol: `isReleasePartititionsOnConsumption` is ambiguous though; as it stands we always (now, and in the future) want the environment to automatically release consumed partitions if they are not externally managed.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/ResultPartition.java
zentol: By uncoupling this field from the field in the descriptor we are running into the real risk of the logic being out-of-sync, particularly so since the descriptor is a shared API, while the ResultPartition is an internal class.

flink-runtime/src/main/java/org/apache/flink/runtime/deployment/ResultPartitionDeploymentDescriptor.java
zentol: So, my thinking is that the ResultPartitionType has no bearing on whether a partition can be consumed multiple times or is externally managed. These are all implementation details of the ShuffleEnvironment.
The only thing that the type itself specifies is whether or not it is consumable while it is being produced.

Now, so far we _did_ couple these things very explicitly to the type being BLOCKING (hence why this field is also based on the type), but long-term this might be problematic.
For example, we may at some point want to make pipelined partitions re-consumable as well, by also spilling them to disk while streaming them to producers at the same time.
If we bake this property into the type we'd have to add additional ResultPartitionTypes in the future to allow this to be done, which typically is a tricky think to do.

Additionally, the scheduler (at a later point) may decide that a certain blocking partition should not be managed by the JM but by the environment, which a separate flag would allow.
(At which point we'D have to modify the constructor obviously)

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/ResultPartition.java
zhijiangW: I mean we do not add the `isManagedExternally ` property in `ResultPartitionDeploymentDescriptor` because it is only `ResultPartitionType` relevant and we could get this property directly in `ResultPartition`.  Only if `isManagedExternally` might be generated from other properties except `ResultPartitionType` in future?

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/NettyShuffleEnvironmentConfiguration.java
zhijiangW: Agree, so might define the var as `isReleaseExternallyManagedPartitionsOnConsumption` instead of current `isReleaseBlockingPartitionsOnConsumption` because it might bring confusion whether `Blocking` is strictly equivalent with `ExternallyManaged` as now and future.

flink-runtime/src/main/java/org/apache/flink/runtime/deployment/ResultPartitionDeploymentDescriptor.java
zhijiangW: Agree, thanks for the explanation!

zhijiangW: Agree, thanks for the explanation!

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/NettyShuffleEnvironmentConfiguration.java
zentol: yes, thats a good point.

zentol:
@zhijiangW could have another look at the latest commit? Addresses an issue raised in #8687 . https://github.com/apache/flink/pull/8687#discussion_r293196661

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/ResultPartitionTest.java
```diff
@@ -100,6 +101,22 @@ public void testAddOnFinishedBlockingPartition() throws Exception {
 		testAddOnFinishedPartition(ResultPartitionType.BLOCKING);
 	}
 
+	@Test
+	public void testExternallyManagedBlockingPartitionConsumableMultipleTimes() throws IOException {
+		final ResultPartition partition = new ResultPartitionBuilder()
+			.setIsExternallyManaged(true)
+			.setResultPartitionType(ResultPartitionType.BLOCKING)
```
zhijiangW: I guess we do not need to set partition type and number of subpartitions, and the default would make sense.

zentol: The partition being blocking is important for the test so I'd like to not omit it.

zhijiangW:
Yes, I have reviewed the last commit and thought you might create separate ticket for ref-count before. :)

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/ResultPartitionTest.java
zhijiangW: Actually I thought the property of `isExternallyManaged` should control whether it could be consumed multiple times not related to `ResultPartitionType`. But in reality the pipelined type could not support this atm because of the `checkState(readView == null` while `PipelinedSubpartition#createReadView`. So we can keep the current way for setting the blocking type.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/ResultPartitionManager.java
```diff
@@ -113,10 +126,12 @@ void onConsumedPartition(ResultPartition partition) {
 			final ResultPartition previous = registeredPartitions.remove(partition.getPartitionId());
 			// Release the partition if it was successfully removed
 			if (partition == previous) {
-				partition.release();
-				ResultPartitionID partitionId = partition.getPartitionId();
-				LOG.debug("Released partition {} produced by {}.",
-					partitionId.getPartitionId(), partitionId.getProducerId());
+				if (!partition.isManagedExternally() || isReleaseExternallyManagedPartitionsOnConsumption) {
```
zentol: this isn't sufficient; the partition is already removed from `registeredPartitions` above, leading to wrong behavior of `getUnreleasedPartitions`

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/ResultPartition.java
```diff
@@ -381,17 +383,20 @@ void onConsumedSubpartition(int subpartitionIndex) {
 			return;
 		}
 
-		int refCnt = pendingReferences.decrementAndGet();
-
-		if (refCnt == 0) {
+		if (isManagedExternally) {
 			partitionManager.onConsumedPartition(this);
```
zhijiangW: if `isManagedExternally`, we should not call `partitionManager.onConsumedPartition(this)` after one sub partition is consumed.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/ResultPartitionManager.java
zhijiangW: Yes, I also noticed this just now, the state is inconsistent here if removed but not released.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/ResultPartition.java
zentol: It should still be safe to do so (because the RPM basically ignores the call), but yeah this shows that Ih aven't thought this through properly.

zentol: Could we call `pin()` in `createSubpartitionView` and decrement the refcount as usual in `onConsumedPartition`?

zhijiangW: The `pin` would increase the counter as the number of subpartitions, but `createSubpartitionView` only indicates for one of the subpartitions. I need think through this issue again. 
Maybe we could focus on the pin issue in a separate ticket and this PR only introduces the flag but not changed any current behaviors if you like.

zentol: I think I've found a way to do it; pin() will continue to increment the refCount by the number of subpartitions, but createSubpartitionsView will increment it by 1 (if the partition is externally managed).

I'm not sure whether this is a good approach, and what plans people had on how this should be implemented, so I wouldn't oppose deferring this to a follow-up.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/ResultPartitionManager.java
```diff
@@ -110,14 +124,20 @@ void onConsumedPartition(ResultPartition partition) {
...
-				partition.release();
-				ResultPartitionID partitionId = partition.getPartitionId();
-				LOG.debug("Released partition {} produced by {}.",
-					partitionId.getPartitionId(), partitionId.getProducerId());
-			}
+			registeredPartitions.computeIfPresent(
+				partition.getPartitionId(),
+				(resultPartitionID, resultPartition) -> {
+					if (partition == resultPartition) {
+						if (partition.isManagedExternally() && !isReleaseExternallyManagedPartitionsOnConsumption) {
```
zhijiangW: I have another option for this. ATM both `ResultPartition` and `ResultPartitionManager` need check the conditions of releasing partitions consumed, then the logic seems difficult to trace. 
We could make `ResultPartitionManager` not aware of this. The `isReleaseExternallyManagedPartitionsOnConsumption` could be merged with `partition.isManagedExternally()` to generate one final boolean tag in `ResultPartition`. Then the `ResultPartition` checks this tag to decide whether to call `partitionManager.onConsumedPartition` while one sub partition is consumed. The logic in `ResultPartitionManager` would be simple:

- Based on consumption release: The behavior is the same as before after calling `ResultPartitionManger#onConsumedPartition`, remove from map and release directly.

- Based on external release: via `ResultPartitionManager#releasePartition`.

And we only have one boolean tag in `ResultPartition` to handle the external check. WDYT? 
This suggestion is not relevant with pin issue.

zentol: ... I never thought about combining isManagedExternally and the configuration flag. That's a great idea.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/ResultPartition.java
zhijiangW: TBH I do not like the current way because it seems a bit hacky. 
Because during `ResultPartition#onConsumedSubpartition`, the counter would never be 0 for external case, so the `ResultPartitionManager#onConsumedPartition` would never be called, then the internal checking logic in `ResultPartitionManager#onConsumedPartition` seems reductant.

zhijiangW: Maybe we could always `pin` as before, but migrate the checking logic from `ResultPartitionManger` to `ResultPartition` as I mentioned before. I mean `ResultPartitionManager` is not aware of the external issue, the tag in `ResultPartitionManager#isReleaseExternallyManagedPartitionsOnConsumption` could be merged with `ResultPartition#isManagedExternally` to generate a final tag in `ResultPartition`.

Then while `ResultPartition#onConsumedSubpartition`, it could check the final tag to decide whether to call `ResultPartitionManager#onConsumedPartition`. And the reference counter only works when the final tag is false.

zhijiangW: In conclusion:

- do not modify previous `ResultPartitionManger`, only add final tag in `ResultPartition`

- pin logic could keep same as before.

- the ref counter check in `ResultPartition#createSubpartitionView` only works when tag false

- the ref counter check in `ResultPartition#onConsumedSubpartition` only works when tag false

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/ResultPartitionManager.java
zentol: Implemented.

flink-runtime/src/main/java/org/apache/flink/runtime/deployment/ResultPartitionDeploymentDescriptor.java
```diff
@@ -48,6 +50,18 @@
 	/** Flag whether the result partition should send scheduleOrUpdateConsumer messages. */
 	private final boolean sendScheduleOrUpdateConsumersMessage;
 
+	/** Whether the result partition is managed externally.
+	 *
+	 * <p>An externally managed partition is only released if either
+	 * a) the {@link ShuffleEnvironment} is instructed to do so via {@link ShuffleEnvironment#releasePartitions(Collection)},
+	 * b) the production of the partition fails,
+	 * c) the {@link ShuffleEnvironment} shuts down.
```
tillrohrmann: I think the formatting would be better if we used `<ul><li></li>...</ul>`

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/ResultPartitionBuilder.java
```diff
@@ -112,6 +114,11 @@ public ResultPartitionBuilder setBufferPoolFactory(
 		return this;
 	}
 
+	public ResultPartitionBuilder setIsExternallyManaged(boolean isExternallyManaged) {
```
tillrohrmann: could be `isExternallyManaged()` which sets `this.isExternallyManaged = true`. This makes the API a bit conciser.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/ResultPartitionFactory.java
```diff
@@ -62,13 +63,15 @@ public ResultPartitionFactory(
...
-		int floatingNetworkBuffersPerGate) {
+		int floatingNetworkBuffersPerGate,
+		boolean releaseExternallyManagedPartitionsOnConsumption) {
 
 		this.partitionManager = partitionManager;
 		this.ioManager = ioManager;
 		this.networkBuffersPerChannel = networkBuffersPerChannel;
 		this.floatingNetworkBuffersPerGate = floatingNetworkBuffersPerGate;
 		this.bufferPoolFactory = bufferPoolFactory;
+		this.releaseExternallyManagedPartitionsOnConsumption = releaseExternallyManagedPartitionsOnConsumption;
```
tillrohrmann: in a previous commit we named this field `isReleaseExternallyManagedPartitionsOnConsumption`. Might be good to be consistent how we name boolean fields.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/ResultPartition.java
```diff
@@ -381,13 +386,15 @@ void onConsumedSubpartition(int subpartitionIndex) {
...
-		if (refCnt == 0) {
-			partitionManager.onConsumedPartition(this);
-		}
-		else if (refCnt < 0) {
-			throw new IllegalStateException("All references released.");
+			if (refCnt == 0) {
+				partitionManager.onConsumedPartition(this);
+			} else if (refCnt < 0) {
+				throw new IllegalStateException("All references released.");
+			}
```
tillrohrmann: Looking at the changes to the `ResultPartition` I'm was wondering whether it wouldn't make sense to have a `ReleaseOnConsumptionResultPartition` implementation which extends `ResultPartition`. Now we are mixing the logic for the implicit and explicit release in one class.

flink-runtime/src/main/java/org/apache/flink/runtime/deployment/ResultPartitionDeploymentDescriptor.java
```diff
@@ -48,6 +50,18 @@
...
+	 *
+	 * <p>An internally managed partition may be released by the {@link ShuffleEnvironment} at any time it deems
+	 * appropriate, typically after it has been fully consumed once.
+	 * */
+	private final boolean isManagedExternally;
```
tillrohrmann: I think it would be better to name this field `releaseOn/AfterConsumption` because it better describes what this field influences.

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/NettyShuffleEnvironmentConfiguration.java
```diff
@@ -60,6 +60,8 @@
 
 	private final boolean isNetworkDetailedMetrics;
 
+	private final boolean isReleaseExternallyManagedPartitionsOnConsumption;
```
tillrohrmann: If we renamed the field `isManagedExternally` in the `ResultPartitionDeploymentDescriptor` into `releaseOnConsumption`, we could rename this field into `forcePartitionReleaseOnConsumption`.

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/ResultPartitionTest.java
```diff
@@ -100,6 +103,37 @@ public void testAddOnFinishedBlockingPartition() throws Exception {
...
+		assertThat(manager.getUnreleasedPartitions(), contains(partition.getPartitionId()));
+		assertFalse(partition.isReleased());
+
+		// an externally managed blocking partition should be consumable multiple times
+		ResultSubpartitionView subpartitionView2 = partition.createSubpartitionView(0, () -> { });
+		subpartitionView2.notifySubpartitionConsumed();
+
+		// partition should not be released on consumption
+		assertThat(manager.getUnreleasedPartitions(), contains(partition.getPartitionId()));
+		assertFalse(partition.isReleased());
```
tillrohrmann: Here is some code duplication. We could simply execute the calls in a loop.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/ResultPartition.java
zentol: I wouldn't be against introducing an additional ResultPartition class. It is fairly straight-forward to do, and outright prevents anyone from interfering with the refCounts if it is externally managed, which is quite useful.

zentol:
@tillrohrmann I implemented both of your suggestions.

I like the renaming of externallyManaged to releasedOnConsumption; it makes things a lot easier to understand.

As for the ResultPartition sub-class, the empty `pin()` method in the base-class is a bit odd, but otherwise I like it.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/ResultPartition.java
```diff
@@ -337,8 +324,7 @@ public boolean isReleased() {
 	@Override
 	public String toString() {
 		return "ResultPartition " + partitionId.toString() + " [" + partitionType + ", "
-				+ subpartitions.length + " subpartitions, "
-				+ pendingReferences + " pending references]";
+				+ subpartitions.length + " subpartitions";
```
zhijiangW: nit: left `]` in the end.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/ResultPartition.java
```diff
@@ -350,18 +336,6 @@ public String toString() {
 	 * operation.
 	 */
 	void pin() {
```
zhijiangW: The above javadoc `The partition can only be released after each subpartition has been consumed once per pin
operation` might also be adjusted.

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/ResultPartitionBuilder.java
```diff
@@ -112,13 +114,19 @@ public ResultPartitionBuilder setBufferPoolFactory(
...
+
 	public ResultPartition build() {
 		ResultPartitionFactory resultPartitionFactory = new ResultPartitionFactory(
 			partitionManager,
 			ioManager,
 			networkBufferPool,
 			networkBuffersPerChannel,
-			floatingNetworkBuffersPerGate);
+			floatingNetworkBuffersPerGate,
+			true);
```
zhijiangW: After this value is set true, it seems no way to generate `ResultPartition` atm via builder no matter with `releasedOnConsumption` is set true or false.
Another option is make this value false, and let `releasedOnConsumption` default value is true, then the final tag could  also be changed via `ResultPartitionBuilder#isReleasedOnConsumption(false)`.
I am not sure whether to adjust this issue atm or fix when required future.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/ReleaseOnConsumptionResultPartition.java
```diff
@@ -0,0 +1,102 @@
...
+			ResultSubpartition[] subpartitions,
+			int numTargetKeyGroups,
+			ResultPartitionManager partitionManager,
+			FunctionWithException<BufferPoolOwner, BufferPool, IOException> bufferPoolFactory) {
+		super(owningTaskName, partitionId, partitionType, subpartitions, numTargetKeyGroups, partitionManager, bufferPoolFactory);
+	}
+
+	@Override
+	void pin() {
+		while (true) {
```
zhijiangW: Actually I am not sure why it needs `while` loop here before.

zhijiangW:
Thanks for the updates @zentol and the reviews @tillrohrmann .
I like the current way and it looks pretty good now. I only left several nit comments.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/ReleaseOnConsumptionResultPartition.java
gaoyunhaii: Follow Zhijiang's thought, I think we can now remove the pin method from the ResultPartition, since it is only the implementation detail of the ReleaseOnConsumptionResultPartiton. Then we can move the initialization of the pending reference to the Constructor to ensure no concurrent access to the variable, then we can use pendingReferences.set(subpartitions.length) to initialize the variable.

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/ResultPartitionFactoryTest.java
```diff
@@ -0,0 +1,86 @@
...
+ */
+
+package org.apache.flink.runtime.io.network.partition;
+
+import org.apache.flink.runtime.deployment.ResultPartitionDeploymentDescriptor;
+import org.apache.flink.runtime.executiongraph.ExecutionAttemptID;
+import org.apache.flink.runtime.io.disk.iomanager.NoOpIOManager;
+import org.apache.flink.runtime.io.network.buffer.BufferPool;
+import org.apache.flink.runtime.io.network.buffer.BufferPoolFactory;
+import org.apache.flink.runtime.io.network.buffer.BufferPoolOwner;
```
gaoyunhaii: It seems there are unused imports here.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/ReleaseOnConsumptionResultPartition.java
zhijiangW: I created [FLINK-12842](https://issues.apache.org/jira/browse/FLINK-12842) and [FLINK-12843](https://issues.apache.org/jira/browse/FLINK-12843) for the ref-counter issues which could be solved separately after this PR merged, because these issues already exist before and are not in the scope of this PR.

zentol: I agree that we can resolve these in follow-ups.

PR #8661, opened 2019-06-07
 [FLINK-12711][table] Separate function implementation and definition

twalthr:

This PR is a first step toward unified expression stacks. It uncouples runtime implementation from information that is needed during validation and planning.

With this change it is possible to declare properties such as `isDeterminstic` or `requiresOver` also for built-in functions. A next step is to add a `getTypeInference` method to `FunctionDefinition` which enables the removal of all `PlannerExpression` by using `BuiltInFunctionDefinition` instead.

## Brief change log

See commit messages.

## Verifying this change

  - The serializers: no

## Documentation


dawidwys:
+1 for 6c66c85

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/FunctionLookup.java
```diff
@@ -16,19 +16,44 @@
...
 
+import java.util.Optional;
+
 /**
- * Catalog of functions that can resolve the name of a function to a {@link FunctionDefinition}.
+ * Catalog of functions that can resolve the name of a function to a {@link FunctionLookup.Result}.
  */
 @Internal
-public interface FunctionDefinitionCatalog {
+public interface FunctionLookup {
```
dawidwys: I know it will be long but maybe `FunctionDefintionLookup`. It would be cleaner that we lookup only the defintion and we e.g. do not createthe implementation.

twalthr: Yes, but we also return the object identifier. So it is more than just the definition. The catalog is also called `FunctionCatalog` and not `FunctionDefinitionCatalog`.

dawidwys: True, but we do return the object identifier, but it is still closer to `FunctionDefinition` than to e.g. Scalar**Function**, Aggregate**Function** etc. But that was a minor suggestion. I am absolutely fine with that name.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/functions/UserDefinedFunction.java
```diff
@@ -24,11 +24,18 @@
...
+ * (see {@link UserDefinedFunctionDefinition}) and contain a corresponding runtime implementation.
+ *
+ * @see ScalarFunction
+ * @see TableFunction
+ * @see AggregateFunction
+ * @see TableAggregateFunction
  */
 @PublicEvolving
-public abstract class UserDefinedFunction implements Serializable {
+public abstract class UserDefinedFunction implements UserDefinedFunctionDefinition, Serializable {
```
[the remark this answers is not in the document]
twalthr: I removed `UserDefinedFunctionDefinition` for now. It is difficult to clearly separate definition and implementation for UDFs. Let's consider `UserDefinedFunction` as a special kind of `FunctionDefinition`. In the future we could think about a design like the following:

```
InvokableFunction
- open()
- close()

UserDefinedFunction extends InvokableFunction

FunctionDefinition
- getKind()
- getRequirements()
- isDeterministic()
- getTypeInference()
- createInvokableFunction(): Optional<InvokableFunction>

InvokableScalarFunction extends InvokableFunction

InvokableTableFunction extends InvokableFunction
- setCollector()
- collect()

InvokableAsyncTableFunction extends InvokableFunction

InvokableAggregateFunction extends InvokableFunction
- getValue()
- createAccumulator()

InvokableTableAggregateFunction extends InvokableFunction
- createAccumulator()
- createAccumulator()

ScalarFunction extends UserDefinedFunction implements InvokableScalarFunction, FunctionDefinition

TableFunction extends UserDefinedFunction implements InvokableTableFunction, FunctionDefinition

AggregateFunction extends UserDefinedFunction implements InvokableAggregateFunction, FunctionDefinition

TableAggregateFunction extends UserDefinedFunction implements InvokableTableAggregateFunction, FunctionDefinition
```

But this requires major changes in how UDFs are shipped to the cluster and instantiated.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/FunctionCatalog.java
```diff
@@ -104,7 +128,8 @@ public void registerScalarFunction(String name, ScalarFunction function) {
 			foundDefinition = BuiltInFunctionDefinitions.getDefinitions()
 				.stream()
 				.filter(f -> normalizeName(name).equals(normalizeName(f.getName())))
-				.findFirst();
+				.findFirst()
+				.map(Function.identity());
```
dawidwys: neat didn't know about this trick

twalthr:
Thank you @dawidwys. Merging...

PR #8664, opened 2019-06-08
[FLINK-12779][table] Avoid field conflicts when generate field names for non-composite Typeinformation

hequn8128:

This pull request makes the result name of table function more robust. 

We use `FieldInfoUtils.getFieldNames(resultType)` to get the relative field names of the resultType. There are no problem for composite types. For non-composite types, we always set the field name to `f0`. But the `f0` may conflict with the predefined field names. To make it more robust, we should generate a field name with no conflicts. For example, we can use `f0_0` as the field name if `f0` has been used. This is also consistent with the behavior of SQL.

## Brief change log

  - Add an another `getFieldNames` method which takes a list of invalid field names into consideration when generate result field names.
  - Use the new `getFieldNames` to generate TableFunction result field names.
  - Add tests.

## Verifying this change

  - Added `CorrelateTest.testNonCompositeResultType()` to test whether field names are generated correctly.

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/typeutils/FieldInfoUtils.java
+++ b/flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/typeutils/FieldInfoUtils.java
@@ -310,6 +310,37 @@ public class FieldInfoUtils {
 	}
 
 	/**
+	 * Returns field names for a given {@link TypeInformation}. If the input {@link TypeInformation}
+	 * is not a composite type, the result field name should not exist in the invalidNamesForNonCompositeType.
+	 *
+	 * @param inputType The TypeInformation extract the field names.
+	 * @param invalidNamesForNonCompositeType The invalid field names for non-composite types.
+	 * @param <A> The type of the TypeInformation.
+	 * @return An array holding the field names
+	 */
+	public static <A> String[] getFieldNames(TypeInformation<A> inputType, List<String> invalidNamesForNonCompositeType) {
+		validateInputTypeInfo(inputType);
+
+		final String[] fieldNames;
+		if (inputType instanceof CompositeType) {
+			fieldNames = ((CompositeType<A>) inputType).getFieldNames();
+		} else {
+			int i = 0;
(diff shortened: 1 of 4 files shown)


sunjincheng121:
@flinkbot approve description

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/typeutils/FieldInfoUtils.java
```diff
@@ -186,6 +186,37 @@ public static boolean isReferenceByPosition(CompositeType<?> ct, Expression[] fi
...
+	/**
+	 * Returns field names for a given {@link TypeInformation}. If the input {@link TypeInformation}
+	 * is not a composite type, the result field name should not exist in the invalidNamesForNonCompositeType.
+	 *
+	 * @param inputType The TypeInformation extract the field names.
+	 * @param invalidNamesForNonCompositeType The invalid field names for non-composite types.
+	 * @param <A> The type of the TypeInformation.
+	 * @return An array holding the field names
+	 */
+	public static <A> String[] getFieldNames(TypeInformation<A> inputType, List<String> invalidNamesForNonCompositeType) {
```
sunjincheng121: I think adding a new method with the difference prams is pretty good. but the impl logic can be merged into one method! i.e.:
public static <A> String[] getFieldNames(TypeInformation<A> inputType) {
getFieldNames(inputType, Collections.emptyList());
}
public static <A> String[] getFieldNames(TypeInformation<A> inputType, List<String> existingNames) {
...
}
Whtat do you think?

hequn8128: Good idea!

hequn8128:
@sunjincheng121 Thanks a lot for your review and suggestions. I have updated the PR according to your suggestions.

Best, Hequn

sunjincheng121:
@flinkbot approve all
+1 to merged

PR #8665, opened 2019-06-08
[FLINK-12781] [Runtime/REST] include the whole stack trace in response payload

stevenzwu:


stevenzwu:
test failed due to this change

```Response entity too large: DefaultHttpResponse(decodeResult: success, version: HTTP/1.1)
HTTP/1.1 200 OK
Content-Type: application/json; charset=UTF-8
content-length: 6469 Try to raise [rest.client.max-content-length]```

stevenzwu:
latest travis build seems to fail with build timeout

StephanEwen:
Having the stack trace is indeed helpful.

The response can however still fail due to messages with a stack trace longer than the content length.
I think we need to cut the stack trace to be below the max contents length. That may loose some parts of the stack trace in the end, but otherwise the whole response can fail.

stevenzwu:
@StephanEwen that is a great point. I should have looked into what the test failure really means, instead of just increase the max content length in the test class. there are already a few test methods verify the max content length

StephanEwen:
This looks good. Will rebase and test this...

PR #8670, opened 2019-06-10
[hotfix][utils] Replace implemented FutureUtils#toJava with lib method

TisonKun:

Replace implemented `FutureUtils#toJava` with scala's lib method.

## Brief change log

Replace `FutureUtils#toJava` with `scala.compat.java8.FutureConverters#toJava`.

For the unchecked upcast, ref https://github.com/scala/scala-java8-compat/pull/152 but it should be ok in this case.

## Verifying this change

  - The serializers: (no)

## Documentation

cc @zentol @GJL

### DIFF
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/concurrent/FutureUtils.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/concurrent/FutureUtils.java
@@ -26,8 +26,6 @@ import org.apache.flink.util.ExceptionUtils;
 import org.apache.flink.util.function.RunnableWithException;
 import org.apache.flink.util.function.SupplierWithException;
 
-import akka.dispatch.OnComplete;
-
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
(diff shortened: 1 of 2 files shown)


flink-runtime/src/main/java/org/apache/flink/runtime/concurrent/FutureUtils.java
```diff
@@ -807,20 +805,8 @@ public static Time toTime(FiniteDuration finiteDuration) {
...
-					result.completeExceptionally(failure);
-				} else {
-					result.complete(success);
-				}
-			}
-		}, Executors.directExecutionContext());
-
-		return result;
+		//noinspection unchecked
+		return scala.compat.java8.FutureConverters.toJava((Future<T>) scalaFuture).toCompletableFuture();
```
tillrohrmann: Where does this dependency come from? I fear that this is only a transitive dependency on which we should not directly depend.

TisonKun: Thank Till for your review!

It is a dependency comes from org.scala-lang.modules as `org.scala-lang.modules:scala-java8-compat_2.11:0.7.0`, shipped within scala-lang and also mentioned in our `NOTICE` file.

Besides, this module is marked as "currently still experimental" and thus I might give up this hotfix. I had expected to reduce a custom implementation DirectExecutionContext and maintain it.

PR #8677, opened 2019-06-10
[FLINK-12795] [table-planner-blink] Extracted creation & configuration of FrameworkConfig & RelBuilder to separate class in blink planner

godfreyhe:

*Extracted creation & configuration of FrameworkConfig & RelBuilder to separate class in blink planner. 
Both those classes should be configured for a single planning session.They are constructed of static properties, that do not change in a lifecycle of TableEnvironment(e.g. TypeSystem) & dynamic (e.g. default path). The newly introduced PlannerContext class helps to split
those two sets of properties.*

## Brief change log

  - *Extracted creation & configuration of FrameworkConfig & RelBuilder logic into PlannerContext, and make sure that FrameworkConfig & RelBuilder meet static and dynamic properties as mentioned above*

## Verifying this change

This change is already covered by existing tests*.

## Documentation

### DIFF
--- a/flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/PlannerContext.java
+++ b/flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/PlannerContext.java
@@ -107,6 +107,10 @@ public class PlannerContext {
 		return typeFactory;
 	}
 
+	public SchemaPlus getRootSchema() {
+		return frameworkConfig.getDefaultSchema();
+	}
+
 	/**
 	 * Creates a configured {@link FlinkRelBuilder} for a planning session.
 	 *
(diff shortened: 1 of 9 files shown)


flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/PlanningConfigurationBuilder.java
```diff
@@ -0,0 +1,219 @@
...
+import java.util.List;
+
+/**
+ * Utility class to create {@link org.apache.calcite.tools.RelBuilder} or {@link FrameworkConfig} used to create
+ * a corresponding {@link org.apache.calcite.tools.Planner}. It tries to separate static elements in a
+ * {@link org.apache.flink.table.api.TableEnvironment} like: root schema, cost factory, type system etc.
+ * from a dynamic properties like e.g. default path to look for objects in the schema.
+ */
+@Internal
+public class PlanningConfigurationBuilder {
```
KurtYoung: It named as `Builder` but not act like a normal builder. Looks like `PlannerContext` would be a better name?

godfreyhe: make sense

PR #8678, opened 2019-06-10
[FLINK-12708][table] Introduce new source and sink interfaces to make Blink runner work

wuchong:

In order to support Blink runner (especially Blink batch source/sink and lookup source), we need some new source&sink interfaces and changes about TableSchema.

- `AsyncTableFunction`: used as async lookup function, to support async temporal table join (i.e. dimension table join.).
- `LookupableTableSource`: The LookupableTableSource interface adds support for the table to be accessed via key column(s) in a lookup fashion in order to support temporal table join.
BoundedTableSource: used for batch table source, returns a bounded DataStream, not a InputFormat, because we also want to reuse existing streaming source implementation to support bounded source, for example: bounded Kafka source.
- `BoundedTableSink`: used for batch table sink, emit a bounded DataStream, not a OutputFormat. The reason is the same as above.

## Brief change log

This contribution contains 4 commits:

1. Introduce `BoundedTableSource` 
   - add `isBounded` interface to `StreamTableSource`
   - `BoundedTableSource` extends `StreamTableSource` and expose `getInputFormat`
   - removes `BatchTableSource` and `StreamTableSource` in blink planner
   - support it in blink and flink planner
2. Introduce `BoundedTableSink`
   - `BoundedTableSink` extends `StreamTableSink` expose `getInputFormat`
   - removes `BatchTableSink` in blink planner
   - support it in blink and flink planner
3. Introduce `LookupableTableSource`
   - removes `LookupableTableSource` and `LookupConfig` in blink planner
   - support it only in blink planner
4. Expose `getTableStats` in `TableSource`
   - support it in blink and flink planner

## Verifying this change

Add some unit tests and integration tests

  - The serializers: (no)

## Documentation


wuchong:
Hi @twalthr  @aljoscha ,  it would be great if you have time to review it.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/functions/AsyncTableFunction.java
```diff
@@ -0,0 +1,139 @@
...
+	 * Returns the result type of the evaluation method with a given signature.
+	 *
+	 * <p>This method needs to be overridden in case Flink's type extraction facilities are not
+	 * sufficient to extract the {@link TypeInformation} based on the return type of the evaluation
+	 * method. Flink's type extraction facilities can handle basic types or
+	 * simple POJOs but might be wrong for more complex, custom, or composite types.
+	 *
+	 * @return {@link TypeInformation} of result type or <code>null</code> if Flink should determine the type
+	 */
+	public TypeInformation<T> getResultType() {
```
KurtYoung: Use `DataType` instead?

flink-table/flink-table-common/src/main/java/org/apache/flink/table/sources/TableSource.java
```diff
@@ -84,4 +85,9 @@ default DataType getProducedDataType() {
 	default String explainSource() {
 		return TableConnectorUtils.generateRuntimeName(getClass(), getTableSchema().getFieldNames());
 	}
+
+	/** Returns the statistics of the table, returns null if don't know the statistics. */
+	default TableStats getTableStats() {
```
KurtYoung: Return `Optional<TableStats>` instead?

wuchong: Sure

flink-table/flink-table-common/src/main/java/org/apache/flink/table/functions/AsyncTableFunction.java
wuchong: Currently, we do not rely on this method `getResultType` and  `getParameterTypes`, we can get the result type from `TableSource#getResultType` or `TableSource#getProducedDataType`. 

So I removed these methods, and will introduce new methods using `DataType` after `TableFunction`  is refactored.

wuchong:
Comments addressed and conflicts solved.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/functions/AsyncTableFunction.java
KurtYoung: I don't think so. `AsyncTableFunction` should first be a table function, it can be treated as normal table function but executed in an aync way. You current solution will make this interface can only apply to lookup join which i think is somehow coupled. 
We still need to treat these two things separately.

wuchong: Do you mean users have to implement  `AsyncTableFunction#getResultType` even if they have provide the information in `TableSource#getResultType`?

The only concern I have is that the `DataType getResultType()` is not a final solution. According to ["Type System Part II"](https://docs.google.com/document/d/1RM_-XvD25AldUOl7xRSvA3sascWBa1-Jze3i5mvb7WU/edit#), we need to introduce `getOutputTypeInference(): OutputTypeInference` to satisfy all the requirements. Which means even though we expose `DataType getResultType()`, it will also be deprecated and refactored in the near future.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/plan/nodes/physical/batch/BatchExecSink.scala
```diff
@@ -77,11 +78,12 @@ class BatchExecSink[T](
...
       tableEnv: BatchTableEnvironment): StreamTransformation[Any] = {
     val resultTransformation = sink match {
-      case batchTableSink: BatchTableSink[T] =>
+      case boundedTableSink: StreamTableSink[T] =>
+        // we can insert the bounded DataStream into a StreamTableSink
         val transformation = translateToStreamTransformation(withChangeFlag = false, tableEnv)
         val boundedStream = new DataStream(tableEnv.streamEnv, transformation)
-        batchTableSink.emitBoundedStream(
-          boundedStream, tableEnv.getConfig, tableEnv.streamEnv.getConfig).getTransformation
+        boundedTableSink.emitDataStream(boundedStream)
```
JingsongLi: Why not just invoke `getOutputFormat` of `BoundedTableSink`?

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/plan/util/SinkUtil.scala
```diff
@@ -0,0 +1,60 @@
...
+/**
+  * Util for [[org.apache.flink.table.plan.nodes.calcite.Sink]]s.
+  */
+object SinkUtil {
+
+  /**
+    * Gets the newly added sink transformation. The returned sink transformation will be used
+    * to set resources.
+    */
+  def getNewlyAddedSinkTransformation(env: StreamExecutionEnvironment): StreamTransformation[_] = {
```
JingsongLi: It looks very hack.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/plan/nodes/physical/batch/BatchExecSink.scala
wuchong: Because it is a `StreamTableSink`.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/plan/util/SinkUtil.scala
wuchong: Yes. I also left a comment on the code. Under the currently design, we have to return a result transformation for every node (including Sink, mainly used to set resources). I have discussed with @godfreyhe offline and came up with this idea.

If we refactor the node design (fallback to previous design: exclude sink to return a transformation), it will introduce a lot of changes and may delay the progress.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/plan/nodes/physical/batch/BatchExecSink.scala
JingsongLi: According to the current design:
withChangeFlag should be false when source is BoundedTableSink
withChangeFlag should be true when source is BoundedTableSink
I think it's better to separate them and separate BoundedTableSink from DataStream.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/functions/AsyncTableFunction.java
KurtYoung: We can at least keep align with current interfaces until the type system part 2 begins.

wuchong: So let's keep `TypeInformation<T> getResultType()` ?

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/plan/util/SinkUtil.scala
KurtYoung: Looks we need a better solution

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/plan/nodes/physical/batch/BatchExecSink.scala
wuchong: In batch, the `withChangeFlag` should always be false. It doesn't recognize RetractStreamTableSink or UpsertStreamTableSink. We support `StreamTableSink` in batch in order to support various streaming sinks (e.g. kafka).

JingsongLi: No,
If you implement a `RetractSink`(`Tuple<Boolean, Row>`), you want to reuse it in Batch, There must a true `withChangeFlag` to let conversion to convert `BaseRow` to `Tuple`, That's what `withChangeFlag` does. (Internal Blink batch support RetractSink and UpsertSink)

If you don't want support RetractSink and UpsertSink now, please throw exception when sink is RetractSink or UpsertSink.

wuchong: Sure, I will add some checks for this.

flink-table/flink-table-planner/src/main/scala/org/apache/flink/table/api/StreamTableEnvImpl.scala
```diff
@@ -93,6 +93,10 @@ abstract class StreamTableEnvImpl(
                 s"environment. But is: ${execEnv.getStreamTimeCharacteristic}")
         }
 
+      case streamTableSource: StreamTableSource[_] if streamTableSource.isBounded =>
```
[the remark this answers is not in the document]
wuchong: I think it's a protection. If users declares a TableSource is bounded, he may want a batch source which shouldn't be registered into StreamTableEnv. In other word, this check is in order to prohibit to register a `BoundedTableSource`.

wuchong:
Hi @aljoscha , Thanks very much for the suggestions.

Regarding to the `TableStats`. The `TableStats` is the statistics of a table (e.g. row count, ndv, max/min...). It is important information during optimization and is important for the performance improvement of TP-CDS. The `TableStats` was already there but didn't work with `TableSource`. So we expose the `TableStats` in `TableSource`, to support reporting statistics for a table source in this PR.

Regarding to the name of `LookupableTableSource`, do you think the `lookupable` is not appropriate? We use the lookup terminology in Blink runner including lookup join, lookup function, lookup tablesource.
If we use `QueryableTableSource` it might be not align with other terminologies. And we add a `able` suffix to `lookup` to indicate it is a table source which has the ability to lookup, similar to `FilterableTableSource`.

Regarding to the only one `getLookupFunction()`, I'm afraid we can't, because `AsyncTableFunction` doesn't extend `TableFunction`. The same super interface is `UserDefinedFunction`. But IMO, returning a `UserDefinedFunction` is not appropriate here.
I have an idea to make the interface only have one method. That is introuding a `LookupFunctionProvider`. I have created a gist for this idea: https://gist.github.com/wuchong/78fa3fa40323543aac8297fbecae09d8. What do you think about this?

Regarding to how to implement a lookup function, yes, it is not clear in the Javadoc. Because lookup join uses `TableFunction` as a lookup function, the `TableFunction` is not just used as lookup. So we may can't add the lookup function implementation details to the Javadoc of `TableFunction`.
I will add some more instruction to the Javadoc of `getLookupFunction()` and also the documentation with example in the future.

aljoscha:
Cool, thanks! I think this is good to go now. 👍 

We are marking the new lookup and async interfaces as `@Experimental` so we have some leeway in the future when we need to do changes.

However merges this: please make sure to merge the fixups into the respective original commits.

aljoscha:
Of course, also only merge when Travis is green. 😅

wuchong:
Cool! Thanks @aljoscha . I will fix the CI.

PR #8682, opened 2019-06-11
[FLINK-12796][table-planner-blink] Introduce BaseArray and BaseMap to reduce conversion overhead to blink

JingsongLi:

Currently, in internal data format of flink, the array is only BinaryArray, and the map is only BinaryMap. If the user writes a UDAF with arrays as parameters and return values, it will lead to frequent conversion between Java arrays and BinaryArrays (each conversion is equivalent to the entire array of copys), which is very time-consuming.
In order to avoid copy in conversion, BaseArray and BaseMap are introduced as internal formats.
BaseArray is the parent of GenericArray and BinaryArray, providing various read and write operations on an array.
GenericArray is a wrapper class for Java arrays, which internally wraps a Java array. This array stores some elements of internal data format.
Conversion can be avoided when the element type is a primitive type or a type that is consistent internally format and externally format. (Detail see: DataFormatConverters)
After our benchmark, the performance of UDAF using primitive Array has been improved by 10 times.

## Brief change log

Introduce BaseArray, GenericArray.
Introduce BaseMap, GenericMap.
Modify serializers.
Modify ScalarOperatorGens.
Modify DataFormatConverters.

## Verifying this change

ut

  - The serializers: (no)

## Documentation


flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/codegen/calls/ScalarOperatorGens.scala
```diff
@@ -1524,20 +1527,21 @@ object ScalarOperatorGens {
...
          |$valueUpdate
+         |$baseMap = $binaryMap;
        """.stripMargin
     } else {
       // the key or value is not fixed length, re-create the map on every update
       s"""
          |${keyExpr.code}
          |${valueExpr.code}
-         |$mapTerm = $BINARY_MAP.valueOf(${keyExpr.resultTerm}, ${valueExpr.resultTerm});
+         |$baseMap = $BINARY_MAP.valueOf(${keyExpr.resultTerm}, ${valueExpr.resultTerm});
```
liyafan82: This should be binaryMap?

JingsongLi: It is a BinaryMap instance, but it's type term should be BaseMap as return value.

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/dataformat/BinaryMap.java
```diff
@@ -69,6 +74,18 @@ public BinaryArray valueArray() {
 		return values;
 	}
 
+	@Override
+	public Map<Object, Object> toJavaMap(LogicalType keyType, LogicalType valueType) {
+		Object[] keyArray = keys.toClassArray(keyType, getInternalClassForType(keyType));
+		Object[] valueArray = values.toClassArray(keyType, getInternalClassForType(valueType));
```
liyafan82: keyType -> valueType

JingsongLi: nice catch

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/typeutils/BaseArraySerializer.java
```diff
@@ -0,0 +1,280 @@
...
+		Object arr;
+		if (array.isPrimitiveArray()) {
+			switch (eleType.getTypeRoot()) {
+				case BOOLEAN:
+					arr = Arrays.copyOf((boolean[]) array.getArray(), array.numElements());
+					break;
+				case TINYINT:
+					arr = Arrays.copyOf((byte[]) array.getArray(), array.numElements());
+					break;
+				case SMALLINT:
```
liyafan82: do we need to consider char type?

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/types/InternalSerializers.java
```diff
@@ -71,10 +75,12 @@ public static TypeSerializer create(LogicalType type, ExecutionConfig config) {
 				DecimalType decimalType = (DecimalType) type;
 				return new DecimalSerializer(decimalType.getPrecision(), decimalType.getScale());
 			case ARRAY:
-				return BinaryArraySerializer.INSTANCE;
+				return new BaseArraySerializer(((ArrayType) type).getElementType(), config);
```
liyafan82: Each time requires creating a new serializer. Does it cause performance degrade?

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/typeutils/BaseMapSerializer.java
```diff
@@ -0,0 +1,273 @@
...
+	@Override
+	public BaseMap createInstance() {
+		return new BinaryMap();
+	}
+
+	@Override
+	public BaseMap copy(BaseMap from) {
+		if (from instanceof GenericMap) {
+			Map<Object, Object> fromMap = ((GenericMap) from).getMap();
+			HashMap<Object, Object> toMap = new HashMap<>();
```
liyafan82: Here we are using HashMap, so there are some constraints on the key type (the hashCode & equals methods must be implemented properly). 

Are such constraints acceptable, or we should use a list of key-value pairs internally?

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/typeutils/BaseRowSerializer.java
```diff
@@ -165,49 +168,33 @@ public int getArity() {
 
 	/**
 	 * Convert base row to binary row.
-	 * TODO modify it to code gen, and reuse BinaryRow&BinaryRowWriter.
+	 * TODO modify it to code gen.
```
liyafan82: Remove this?

liyafan82:
Can you please provide the benchmark that produces the 10x performance improvement?

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/typeutils/BaseArraySerializer.java
JingsongLi: we don't support logicalType char now.

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/types/InternalSerializers.java
JingsongLi: InternalSerializers.create is used to create serializers. runtime per record process should not invoke it.

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/typeutils/BaseRowSerializer.java
JingsongLi: We still need take this logical to code gen.

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/typeutils/BaseMapSerializer.java
JingsongLi: No,
First: only `IdentityConverter` will use `GenericMap`, others will use `BinaryMap`.
Second: Just wrap user `Map`, users need to guarantee their logic.

liyafan82: "Just wrap user Map, users need to guarantee their logic."
-> the users just guarantee that their Map works correctly according to the Map interface, but it does not mean their Map would work correctly when wrapped in a HashMap. Right?

When that is the case, it may cause some weird problem which is hard to locate and debug. So this is not a reliable solution. 

At the very least, we should write some comment in the JavaDoc about this explicitly.

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/typeutils/BaseRowSerializer.java
liyafan82: I see the code gen related logic is already modified. You mean more changes are needed?

JingsongLi: This is a method comments, it mean move this method implement to code gen instead of using `BinaryWriter.write` and `TypeGetterSetters.get`.

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/typeutils/BaseMapSerializer.java
JingsongLi: This has nothing to do with `GenericMap`. If the user gives a wrong `Map`, it's wrong to turn it into a `BinaryMap` too. It is totally wrong...

liyafan82: The user can provide a "correct" Map, but when the key value is wrapped in a HashMap, it can be a "wrong" Map.

Let me illustrate with an example. Suppose the key type does not implement "hashCode" and "equals" functions correctly (suppose for simplicity, hashCode always returns 0, and equals always returns true).

The input GenericMap can be based on a TreeMap, with a proper comparator to test key equality. 
The input map is a "correct" map, because it implements get/put methods correctly. 

However, when we insert the key/value pairs of the TreeMap into a HashMap, problems will occur. For this example, only one key/value pair will finally exist in the HashMap, which is an unexpected behavior.

JingsongLi: You are right,
`DataFormatConverter` is not wrong, but `copy` maybe have some trouble.
In fact, Runtime `MapSerializer` also has this problem.
I will add some comment.

JingsongLi:
> Can you please provide the benchmark that produces the 10x performance improvement?

You can find the case in `SortAggITCase.testBigDataSimpleArrayUDAF`.
If you want to test previous code, you can modify `PrimitiveLongArrayConverter.toInternalImpl` to use `return BinaryArray.fromPrimitiveArray(value);`.
The difference is so obvious, both in generated code and performance.

liyafan82:
> > Can you please provide the benchmark that produces the 10x performance improvement?
> 
> You can find the case in `SortAggITCase.testBigDataSimpleArrayUDAF`.
> If you want to test previous code, you can modify `PrimitiveLongArrayConverter.toInternalImpl` to use `return BinaryArray.fromPrimitiveArray(value);`.
> The difference is so obvious, both in generated code and performance.

Good job. Thanks.

flink-table/flink-table-runtime-blink/src/test/java/org/apache/flink/table/typeutils/DecimalSerializerTest.java
```diff
@@ -22,7 +22,7 @@
 import org.apache.flink.table.dataformat.Decimal;
 
 /**
- * A test for the {@link BinaryArraySerializer}.
+ * A test for the {@link BaseArraySerializer}.
```
TsReaper: Shoud be "DecimalSerializer"?

flink-table/flink-table-runtime-blink/src/test/java/org/apache/flink/table/typeutils/BinaryRowSerializerTest.java
```diff
@@ -24,7 +24,7 @@
 import org.apache.flink.table.dataformat.BinaryString;
 
 /**
- * A test for the {@link BinaryArraySerializer}.
+ * A test for the {@link BaseArraySerializer}.
```
TsReaper: Should be "BinaryRowSerializer"?

flink-table/flink-table-runtime-blink/src/test/java/org/apache/flink/table/typeutils/BinaryGenericSerializerTest.java
```diff
@@ -23,7 +23,7 @@
 import org.apache.flink.table.dataformat.BinaryGeneric;
 
 /**
- * A test for the {@link BinaryArraySerializer}.
+ * A test for the {@link BaseArraySerializer}.
```
TsReaper: Should be "BinaryGenericSerializer"?

flink-table/flink-table-runtime-blink/src/test/java/org/apache/flink/table/typeutils/BaseMapSerializerTest.java
```diff
@@ -0,0 +1,87 @@
...
+				BinaryMap.valueOf(createArray(1, 2), BaseArraySerializerTest.createArray("11", "haa")),
+				BinaryMap.valueOf(createArray(1, 3, 4), BaseArraySerializerTest.createArray("11", "haa", "ke")),
+				BinaryMap.valueOf(createArray(1, 4, 2), BaseArraySerializerTest.createArray("11", "haa", "ke")),
+				BinaryMap.valueOf(createArray(1, 5, 6, 7), BaseArraySerializerTest.createArray("11", "lele", "haa", "ke"))
+		};
+	}
+
+	private static BinaryArray createArray(int... vs) {
+		BinaryArray array = new BinaryArray();
+		BinaryArrayWriter writer = new BinaryArrayWriter(array, vs.length, 8);
```
TsReaper: 8 should be 4?

flink-table/flink-table-runtime-blink/src/test/java/org/apache/flink/table/typeutils/BaseMapSerializerTest.java
```diff
@@ -0,0 +1,87 @@
...
+
+/**
+ * A test for the {@link BaseMapSerializer}.
+ */
+public class BaseMapSerializerTest extends SerializerTestBase<BaseMap> {
+
+	public BaseMapSerializerTest() {
+		super(new DeeplyEqualsChecker().withCustomCheck(
+				(o1, o2) -> o1 instanceof BaseMap && o2 instanceof BaseMap,
+				(o1, o2, checker) -> newSer().toBinaryMap((BaseMap) o1).equals(newSer().toBinaryMap((BaseMap) o2))
```
TsReaper: I think it's more proper to compare the maps after changing them to Java maps instead of binary maps. For example, consider the following two maps:
{1: 'a', 2: 'b', 3: 'c'} and {3: 'c', 2: 'b', 1: 'a'}
These are actually the same maps, but their key / value order will be different when stored as binary maps, and the equalsTo method of binary map will return false.

JingsongLi: nice catch.

flink-table/flink-table-runtime-blink/src/test/java/org/apache/flink/table/typeutils/BaseMapSerializerTest.java
JingsongLi: ok

flink-table/flink-table-runtime-blink/src/test/java/org/apache/flink/table/typeutils/BinaryGenericSerializerTest.java
JingsongLi: ok

flink-table/flink-table-runtime-blink/src/test/java/org/apache/flink/table/typeutils/BinaryRowSerializerTest.java
JingsongLi: ok

flink-table/flink-table-runtime-blink/src/test/java/org/apache/flink/table/typeutils/DecimalSerializerTest.java
JingsongLi: ok

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/codegen/calls/ScalarOperatorGens.scala
```diff
@@ -1900,8 +1932,72 @@ object ScalarOperatorGens {
...
+          s"""
+             |boolean $resultTerm;
+             |if ($leftTerm instanceof $binaryMapCls && $rightTerm instanceof $binaryMapCls) {
+             |  $resultTerm = $leftTerm.equals($rightTerm);
+             |} else {
+             |  if ($leftTerm.numElements() == $rightTerm.numElements()) {
+             |    $resultTerm = true;
+             |    $mapCls $leftMapTerm = $leftTerm.toJavaMap($keyTypeTerm, $valueTypeTerm);
+             |    $mapCls $rightMapTerm = $rightTerm.toJavaMap($keyTypeTerm, $valueTypeTerm);
+             |
```
TsReaper: Is it possible to first compare the classes of `leftMapTerm` and `rightMapTerm`? If they are equal, just use the `equals` method of Java maps, otherwise compare the keys and values one by one.

flink-table/flink-table-runtime-blink/src/test/java/org/apache/flink/table/typeutils/BaseMapSerializerTest.java
```diff
@@ -0,0 +1,99 @@
...
+		super(new DeeplyEqualsChecker().withCustomCheck(
+				(o1, o2) -> o1 instanceof BaseMap && o2 instanceof BaseMap,
+				(o1, o2, checker) ->
+						// Better is more proper to compare the maps after changing them to Java maps
+						// instead of binary maps. For example, consider the following two maps:
+						// {1: 'a', 2: 'b', 3: 'c'} and {3: 'c', 2: 'b', 1: 'a'}
+						// These are actually the same maps, but their key / value order will be
+						// different when stored as binary maps, and the equalsTo method of binary
+						// map will return false.
+						newSer().toBinaryMap((BaseMap) o1).toJavaMap(INT, STRING)
```
TsReaper: `BaseMap` interface already has the method `toJavaMap`, it'll be more proper to directly write `((Basemap) o1).toJavaMap`

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/codegen/calls/ScalarOperatorGens.scala
JingsongLi: Keys must have right `equals` and `hashcode` for `HashMap`, but values maybe not equivalent, so we must use equals code gen.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/codegen/calls/ScalarOperatorGens.scala
```diff
@@ -1900,8 +1932,72 @@ object ScalarOperatorGens {
...
+        val keyTypeTerm =
+          ctx.addReusableObject(mapType.getKeyType, "keyType", internalTypeCls)
+        val valueTypeTerm =
+          ctx.addReusableObject(mapType.getValueType, "valueType", internalTypeCls)
+
+        val stmt =
+          s"""
+             |boolean $resultTerm;
+             |if ($leftTerm instanceof $binaryMapCls && $rightTerm instanceof $binaryMapCls) {
+             |  $resultTerm = $leftTerm.equals($rightTerm);
```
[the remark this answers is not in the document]
JingsongLi: OK

flink-table/flink-table-runtime-blink/src/test/java/org/apache/flink/table/typeutils/BaseMapSerializerTest.java
JingsongLi: ok

JingsongLi:
Travis passed in https://travis-ci.org/JingsongLi/flink/builds/554513232

PR #8685, opened 2019-06-11
[FLINK-12799][table] Improve expression based TableSchema extraction from DataStream/DataSet

dawidwys:

We should improve the extraction of `TableSchema` from `DataStream/DataSet`. Currently it is split into a few stages:

1. Extract types ignoring time attributes via `FieldInfoUtils#getFieldInfo`
2. Extract the rowtime and proctime positions via `StreamTableEnvImpl#validateAndExtractTimeAttributes`
3. Adjust the indices from 1. using information from 2.

All that could happen in a single pass. This will also deal with the porting/removing of a few methods from `StreamTableEnvImpl`.

## Brief change log

  - Time attributes are also handled during the first pass of `FieldInfoUtils#getFieldInfo`
  - Changed few exceptions types from `TableException` to `ValidationException` as they are thrown when the user provided information is invalid rather than some internal error happens.
  - Fixed checking that no time attributes where provided in `org.apache.flink.table.api.BatchTableEnvImpl#asQueryOperation`

## Verifying this change

## Documentation

### DIFF
--- a/flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/typeutils/FieldInfoUtils.java
+++ b/flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/typeutils/FieldInfoUtils.java
@@ -18,27 +18,35 @@
 
 package org.apache.flink.table.typeutils;
 
+import org.apache.flink.api.common.typeinfo.SqlTimeTypeInfo;
 import org.apache.flink.api.common.typeinfo.TypeInformation;
-import org.apache.flink.api.common.typeinfo.Types;
 import org.apache.flink.api.common.typeutils.CompositeType;
 import org.apache.flink.api.java.typeutils.GenericTypeInfo;
 import org.apache.flink.api.java.typeutils.PojoTypeInfo;
 import org.apache.flink.api.java.typeutils.TupleTypeInfoBase;
 import org.apache.flink.table.api.TableException;
 import org.apache.flink.table.api.TableSchema;
+import org.apache.flink.table.api.Types;
+import org.apache.flink.table.api.ValidationException;
 import org.apache.flink.table.expressions.ApiExpressionDefaultVisitor;
 import org.apache.flink.table.expressions.BuiltInFunctionDefinitions;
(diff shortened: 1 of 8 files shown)


dawidwys:
@aljoscha Could you have a look? Thank you!

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/typeutils/FieldInfoUtils.java
```diff
@@ -381,105 +397,225 @@ public String getFieldName() {
...
-	private static class ExprToFieldInfo extends ApiExpressionDefaultVisitor<Optional<FieldInfo>> {
+	private static class ExprToFieldInfo extends ApiExpressionDefaultVisitor<FieldInfo> {
 
 		private final CompositeType ct;
 
 		private ExprToFieldInfo(CompositeType ct) {
 			this.ct = ct;
 		}
 
+		private ValidationException noFieldFound(String name) {
```
aljoscha: nitpick: my sense would say `fieldNotFound` but this is perfectly fine. 👌

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/typeutils/FieldInfoUtils.java
```diff
@@ -381,105 +397,225 @@ public String getFieldName() {
...
+				validateProcTimeAttributeAppended(call);
+				return createTimeAttributeField(getChildAsReference(call), TimestampKind.PROCTIME, null);
 			}
 
 			return defaultMethod(call);
 		}
 
+		private FieldInfo visitAlias(CallExpression call) {
+			List<Expression> children = call.getChildren();
+			String newName = ExpressionUtils.extractValue(children.get(1), String.class).orElseThrow(() ->
```
aljoscha: You use this exact same code in two places, the two `visitAlias` methods. Could probably be refactored to a helper method.

aljoscha:
I think this looks good to merge now. 👌

PR #8692, opened 2019-06-11
[FLINK-12804] Introduce mailbox-based ExecutorService

StefanRRichter:
This PR introduced an ``ExecutorService`` frontend to the mailbox that can be used by client code. For this purpose, the mailbox was also extended with open->quiesce -> close lifecycle methods.

## Brief change log

- Intoduce ``TaskMailboxExecutorService``, an ``ExecutorService`` with additional  non-blocking alternative to ``execute`` and a ``yield()`` functionality.

- Enhanced ``Mailbox`` with lifecycle methods.

- Improvements to the mailbox usage in ``StreamTask``

## Verifying this change

Introduced ``TaskMailboxExecutorServiceImplTest`` and extended ``MailboxImplTest``.

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/concurrent/FutureUtils.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/concurrent/FutureUtils.java
@@ -28,9 +28,12 @@ import org.apache.flink.util.function.SupplierWithException;
 
 import akka.dispatch.OnComplete;
 
+import javax.annotation.Nonnull;
+
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.List;
 import java.util.concurrent.CancellationException;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionException;
@@ -994,5 +997,28 @@ public class FutureUtils {
 				uncaughtExceptionHandler.uncaughtException(Thread.currentThread(), throwable);
 			}
 		});
+	}
+
+	/**
+	 * Cancels all instances of {@link java.util.concurrent.Future} in the given list of runnables without interrupting.
+	 * This method will suppress unexpected exceptions until the whole list is processed and then rethrow.
+	 *
+	 * @param runnables list of {@link Runnable} candidates to cancel.
+	 */
+	public static void cancelRunnableFutures(@Nonnull List<Runnable> runnables) {
+		RuntimeException suppressedExceptions = null;
+		for (Runnable runnable : runnables) {
+			if (runnable instanceof java.util.concurrent.Future) {
+				try {
+					((java.util.concurrent.Future<?>) runnable).cancel(false);
+				} catch (RuntimeException ex) {
+					// safety net to ensure all candidates get cancelled before we let the exception bubble up.
+					suppressedExceptions = ExceptionUtils.firstOrSuppressed(ex, suppressedExceptions);
+				}
+			}
+		}
+		if (suppressedExceptions != null) {
+			throw suppressedExceptions;
+		}
 	}
 }
--- a/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SourceStreamTask.java
+++ b/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SourceStreamTask.java
@@ -25,8 +25,6 @@ import org.apache.flink.runtime.execution.Environment;
 import org.apache.flink.streaming.api.checkpoint.ExternallyInducedSource;
 import org.apache.flink.streaming.api.functions.source.SourceFunction;
 import org.apache.flink.streaming.api.operators.StreamSource;
-import org.apache.flink.streaming.runtime.tasks.mailbox.Mailbox;
-import org.apache.flink.streaming.runtime.tasks.mailbox.MailboxStateException;
 import org.apache.flink.util.FlinkException;
 
 /**
@@ -103,12 +101,12 @@ public class SourceStreamTask<OUT, SRC extends SourceFunction<OUT>, OP extends S
 	protected void performDefaultAction(ActionContext context) throws Exception {
 		// Against the usual contract of this method, this implementation is not step-wise but blocking instead for
 		// compatibility reasons with the current source interface (source functions run as a loop, not in steps).
-		final LegacySourceFunctionThread sourceThread = new LegacySourceFunctionThread();
+		final LegacySourceFunctionThread sourceThread = new LegacySourceFunctionThread(getName());
 		sourceThread.start();
 
-		// We run an alternative mailbox loop that does not involve default actions and synchronizes around actions.
+		// We run an legacy source mailbox loop that does not involve default actions and synchronizes around actions.
 		try {
-			runAlternativeMailboxLoop();
+			mailboxProcessor.switchToLegacySourceCompatibilityMailboxLoop(getCheckpointLock());
 		} catch (Exception mailboxEx) {
 			// We cancel the source function if some runtime exception escaped the mailbox.
 			if (!isCanceled()) {
@@ -122,21 +120,6 @@ public class SourceStreamTask<OUT, SRC extends SourceFunction<OUT>, OP extends S
 		sourceThread.checkThrowSourceExecutionException();
 
 		context.allActionsCompleted();
-	}
-
-	private void runAlternativeMailboxLoop() throws InterruptedException, MailboxStateException {
-
-		assert taskMailboxExecutor.isMailboxThread() : "Alternative mailbox loop must run in declared mailbox thread!";
-
-		final Mailbox mailbox = taskMailboxExecutor.getMailbox();
-		while (isMailboxLoopRunning()) {
-
-			Runnable letter = mailbox.takeMail();
-
-			synchronized (getCheckpointLock()) {
-				letter.run();
-			}
-		}
 	}
 
 	@Override
@@ -175,7 +158,8 @@ public class SourceStreamTask<OUT, SRC extends SourceFunction<OUT>, OP extends S
 
 		private Throwable sourceExecutionThrowable;
 
-		LegacySourceFunctionThread() {
+		LegacySourceFunctionThread(String taskDescription) {
+			super("Legacy Source Thread - " + taskDescription);
 			this.sourceExecutionThrowable = null;
 		}
 
@@ -185,14 +169,8 @@ public class SourceStreamTask<OUT, SRC extends SourceFunction<OUT>, OP extends S
 				headOperator.run(getCheckpointLock(), getStreamStatusMaintainer(), operatorChain);
 			} catch (Throwable t) {
 				sourceExecutionThrowable = t;
-			}
-
-			try {
-				taskMailboxExecutor.getMailbox().putFirst(mailboxPoisonLetter);
-			} catch (InterruptedException ie) {
-				Thread.currentThread().interrupt();
-			} catch (MailboxStateException me) {
-				LOG.debug("Could not submit poison letter to mailbox.", me);
+			} finally {
+				mailboxProcessor.allActionsCompleted();
 			}
 		}
 
--- a/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SourceStreamTask.java
+++ b/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SourceStreamTask.java
@@ -25,6 +25,7 @@ import org.apache.flink.runtime.execution.Environment;
 import org.apache.flink.streaming.api.checkpoint.ExternallyInducedSource;
 import org.apache.flink.streaming.api.functions.source.SourceFunction;
 import org.apache.flink.streaming.api.operators.StreamSource;
+import org.apache.flink.streaming.runtime.tasks.mailbox.execution.DefaultActionContext;
 import org.apache.flink.util.FlinkException;
 
 /**
@@ -98,7 +99,7 @@ public class SourceStreamTask<OUT, SRC extends SourceFunction<OUT>, OP extends S
 	}
 
 	@Override
-	protected void performDefaultAction(ActionContext context) throws Exception {
+	protected void performDefaultAction(DefaultActionContext context) throws Exception {
 		// Against the usual contract of this method, this implementation is not step-wise but blocking instead for
 		// compatibility reasons with the current source interface (source functions run as a loop, not in steps).
 		final LegacySourceFunctionThread sourceThread = new LegacySourceFunctionThread(getName());
--- a/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SourceStreamTask.java
+++ b/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SourceStreamTask.java
@@ -25,6 +25,8 @@ import org.apache.flink.runtime.execution.Environment;
 import org.apache.flink.streaming.api.checkpoint.ExternallyInducedSource;
 import org.apache.flink.streaming.api.functions.source.SourceFunction;
 import org.apache.flink.streaming.api.operators.StreamSource;
+import org.apache.flink.streaming.runtime.tasks.mailbox.Mailbox;
+import org.apache.flink.streaming.runtime.tasks.mailbox.MailboxStateException;
 import org.apache.flink.util.FlinkException;
 
 /**
@@ -44,8 +46,6 @@ import org.apache.flink.util.FlinkException;
 @Internal
 public class SourceStreamTask<OUT, SRC extends SourceFunction<OUT>, OP extends StreamSource<OUT, SRC>>
 	extends StreamTask<OUT, OP> {
-
-	private static final Runnable SOURCE_POISON_LETTER = () -> {};
 
 	private volatile boolean externallyInducedCheckpoints;
 
@@ -114,6 +114,7 @@ public class SourceStreamTask<OUT, SRC extends SourceFunction<OUT>, OP extends S
 			if (!isCanceled()) {
 				cancelTask();
 			}
+			sourceThread.interrupt();
 			throw mailboxEx;
 		}
 
@@ -123,14 +124,14 @@ public class SourceStreamTask<OUT, SRC extends SourceFunction<OUT>, OP extends S
 		context.allActionsCompleted();
 	}
 
-	private void runAlternativeMailboxLoop() throws InterruptedException {
+	private void runAlternativeMailboxLoop() throws InterruptedException, MailboxStateException {
 
-		while (true) {
+		assert taskMailboxExecutor.isMailboxThread() : "Alternative mailbox loop must run in declared mailbox thread!";
+
+		final Mailbox mailbox = taskMailboxExecutor.getMailbox();
+		while (isMailboxLoopRunning()) {
 
 			Runnable letter = mailbox.takeMail();
-			if (letter == SOURCE_POISON_LETTER) {
-				break;
-			}
 
 			synchronized (getCheckpointLock()) {
 				letter.run();
@@ -184,8 +185,14 @@ public class SourceStreamTask<OUT, SRC extends SourceFunction<OUT>, OP extends S
 				headOperator.run(getCheckpointLock(), getStreamStatusMaintainer(), operatorChain);
 			} catch (Throwable t) {
 				sourceExecutionThrowable = t;
-			} finally {
-				mailbox.clearAndPut(SOURCE_POISON_LETTER);
+			}
+
+			try {
+				taskMailboxExecutor.getMailbox().putFirst(mailboxPoisonLetter);
+			} catch (InterruptedException ie) {
+				Thread.currentThread().interrupt();
+			} catch (MailboxStateException me) {
+				LOG.debug("Could not submit poison letter to mailbox.", me);
 			}
 		}
 
--- a/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SourceStreamTask.java
+++ b/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SourceStreamTask.java
@@ -28,6 +28,8 @@ import org.apache.flink.streaming.api.operators.StreamSource;
 import org.apache.flink.streaming.runtime.tasks.mailbox.execution.DefaultActionContext;
 import org.apache.flink.util.FlinkException;
 
+import java.util.concurrent.CompletableFuture;
+
 /**
  * {@link StreamTask} for executing a {@link StreamSource}.
  *
@@ -100,27 +102,20 @@ public class SourceStreamTask<OUT, SRC extends SourceFunction<OUT>, OP extends S
 
 	@Override
 	protected void performDefaultAction(DefaultActionContext context) throws Exception {
+
+		context.suspendDefaultAction();
+
 		// Against the usual contract of this method, this implementation is not step-wise but blocking instead for
 		// compatibility reasons with the current source interface (source functions run as a loop, not in steps).
 		final LegacySourceFunctionThread sourceThread = new LegacySourceFunctionThread(getName());
 		sourceThread.start();
-
-		// We run an legacy source mailbox loop that does not involve default actions and synchronizes around actions.
-		try {
-			mailboxProcessor.switchToLegacySourceCompatibilityMailboxLoop(getCheckpointLock());
-		} catch (Exception mailboxEx) {
-			// We cancel the source function if some runtime exception escaped the mailbox.
-			if (!isCanceled()) {
-				cancelTask();
+		sourceThread.getCompletionFuture().whenComplete((Void ignore, Throwable sourceThreadThrowable) -> {
+			if (sourceThreadThrowable == null) {
+				mailboxProcessor.allActionsCompleted();
+			} else {
+				mailboxProcessor.reportThrowable(sourceThreadThrowable);
 			}
-			sourceThread.interrupt();
-			throw mailboxEx;
-		}
-
-		sourceThread.join();
-		sourceThread.checkThrowSourceExecutionException();
-
-		context.allActionsCompleted();
+		});
 	}
 
 	@Override
@@ -157,28 +152,25 @@ public class SourceStreamTask<OUT, SRC extends SourceFunction<OUT>, OP extends S
 	 */
 	private class LegacySourceFunctionThread extends Thread {
 
-		private Throwable sourceExecutionThrowable;
+		private final CompletableFuture<Void> completionFuture;
 
 		LegacySourceFunctionThread(String taskDescription) {
 			super("Legacy Source Thread - " + taskDescription);
-			this.sourceExecutionThrowable = null;
+			this.completionFuture = new CompletableFuture<>();
 		}
 
 		@Override
 		public void run() {
 			try {
 				headOperator.run(getCheckpointLock(), getStreamStatusMaintainer(), operatorChain);
+				completionFuture.complete(null);
 			} catch (Throwable t) {
-				sourceExecutionThrowable = t;
-			} finally {
-				mailboxProcessor.allActionsCompleted();
+				completionFuture.completeExceptionally(t);
 			}
 		}
 
-		void checkThrowSourceExecutionException() throws Exception {
-			if (sourceExecutionThrowable != null) {
-				throw new Exception(sourceExecutionThrowable);
-			}
+		CompletableFuture<Void> getCompletionFuture() {
+			return completionFuture;
 		}
 	}
 }
--- a/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
+++ b/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
@@ -474,7 +474,6 @@ public abstract class StreamTask<OUT, OP extends StreamOperator<OUT>>
 
 	@Override
 	public final void cancel() throws Exception {
-		mailboxProcessor.cancelMailboxExecution();
 		isRunning = false;
 		canceled = true;
 
@@ -485,6 +484,7 @@ public abstract class StreamTask<OUT, OP extends StreamOperator<OUT>>
 			cancelTask();
 		}
 		finally {
+			mailboxProcessor.cancelMailboxExecution();
 			cancelables.close();
 		}
 	}
--- a/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
+++ b/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
@@ -59,8 +59,8 @@ import org.apache.flink.streaming.runtime.partitioner.StreamPartitioner;
 import org.apache.flink.streaming.runtime.streamrecord.StreamRecord;
 import org.apache.flink.streaming.runtime.streamstatus.StreamStatusMaintainer;
 import org.apache.flink.streaming.runtime.tasks.mailbox.execution.DefaultActionContext;
+import org.apache.flink.streaming.runtime.tasks.mailbox.execution.MailboxExecutor;
 import org.apache.flink.streaming.runtime.tasks.mailbox.execution.MailboxProcessor;
-import org.apache.flink.streaming.runtime.tasks.mailbox.execution.TaskMailboxExecutor;
 import org.apache.flink.util.ExceptionUtils;
 import org.apache.flink.util.Preconditions;
 
@@ -489,7 +489,7 @@ public abstract class StreamTask<OUT, OP extends StreamOperator<OUT>>
 		}
 	}
 
-	public TaskMailboxExecutor getTaskMailboxExecutor() {
+	public MailboxExecutor getTaskMailboxExecutor() {
 		return mailboxProcessor.getTaskMailboxExecutor();
 	}
 
--- a/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
+++ b/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
@@ -1339,7 +1339,7 @@ public abstract class StreamTask<OUT, OP extends StreamOperator<OUT>>
 	 */
 	public final class ActionContext {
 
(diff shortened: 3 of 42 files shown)


StefanRRichter:
CC @1u0 @pnowojski

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
```diff
@@ -252,23 +262,35 @@ protected StreamTask(
...
+		mailboxLoopRunning = true;
+
 		while (true) {
+
 			if (mailbox.hasMail()) {
 				Optional<Runnable> maybeLetter;
 				while ((maybeLetter = mailbox.tryTakeMail()).isPresent()) {
-					Runnable letter = maybeLetter.get();
-					if (letter == POISON_LETTER) {
+					if (!mailboxLoopRunning) {
```
1u0: **NB:** you are silently ignoring the letter here.

Should there be an additional check for `mailboxLoopRunning` before calling `performDefaultAction` too?

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
```diff
@@ -395,6 +417,9 @@ public final void invoke() throws Exception {
 				// make sure no new timers can come
 				timerService.quiesce();
 
+				// let mailbox execution reject all new letters from this point
+				taskMailboxExecutor.shutdown();
```
1u0: Is it possible that there are still some letters in the mailbox queue that would require to be executed (including future cases to host timer service and checkpoints via mailbox)?
Can `StreamTask::closeAllOperators()` still need the mailbox loop (e.g. for timers)?

My impression, that the old `StreamTask::run()` method (prior #8442 PR) finished running, when the source has ended. Other threads could still do valid operations
So the mailbox loop run is not the same as `StreamTask::run()`, but should span until `StreamTask.isRunning` is `true`.

Otherwise, we would need to change this code again (the `StreamTask` and the loop in `SourceStreamTask`) to support timers in compatible way.

flink-runtime/src/main/java/org/apache/flink/runtime/concurrent/FutureUtils.java
```diff
@@ -995,4 +998,27 @@ public static void handleUncaughtException(CompletableFuture<?> completableFutur
...
+
+	/**
+	 * Cancels all instances of {@link java.util.concurrent.Future} in the given list of runnables without interrupting.
+	 * This method will suppress unexpected exceptions until the whole list is processed and then rethrow.
+	 *
+	 * @param runnables list of {@link Runnable} candidates to cancel.
+	 */
+	public static void cancelRunnableFutures(@Nonnull List<Runnable> runnables) {
+		RuntimeException suppressedExceptions = null;
+		for (Runnable droppedLetter : runnables) {
```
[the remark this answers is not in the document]
StefanRRichter: Good idea, I overlook that when I moved the code outside of the mailbox.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
StefanRRichter: I think that can be fixed by moving the before taking letters. The purpose of this bit more complex logic is that we assume change of `mailboxLoopRunning` is always connected to a letter and with this we do not have to check this field on each loop iteration, but only when we process mail. Therefore no check is required again for the default action.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
StefanRRichter: Currently I don't see any concrete case that would need this. Timers, checkpoints, operators, etc should all be completed at this point. If an operator needs to ensure that something completes first, it can do so in its `close` method - like the Async operator will do for example. For as long as there is no real use case I would strictly avoid running anything here so that the shutdown can complete asap.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
```diff
@@ -467,7 +496,12 @@ public final void invoke() throws Exception {
 
 	@Override
 	public final void cancel() throws Exception {
-		mailbox.clearAndPut(POISON_LETTER);
+		try {
+			List<Runnable> droppedRunnables = taskMailboxExecutor.getMailbox().clearAndPut(mailboxPoisonLetter);
```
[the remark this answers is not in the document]
StefanRRichter: Need to think about this suggestion. Sounds good if it works.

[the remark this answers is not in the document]
StefanRRichter: I had a closer look and the suggestion would change semantics and break some tests, so I prefer to keep my approach.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/Mailbox.java
```diff
@@ -20,17 +20,85 @@
...
+
+	/**
+	 * Adds the given action to the head of the mailbox. This method will block if the mailbox is full and
+	 * should therefore only be called from outside the mailbox main-thread to avoid deadlocks.
+	 *
+	 * @param priorityLetter action to enqueue to the head of the mailbox.
+	 * @throws InterruptedException on interruption.
+	 * @throws MailboxStateException if the mailbox is quiesced or closed.
+	 */
+	void putFirst(@Nonnull Runnable priorityLetter) throws InterruptedException, MailboxStateException;
```
pnowojski: If every method enqueuing the `Runnable` returned a `CompletableFuture<?>`, we would have a better mechanism of canceling them, instead of instance of checking the `Runnable`, right? Plus it would give us a generic mechanism for checking whether an action has completed or not (especially useful in tests). 

Are there some drawbacks of using `CompletableFuture`?

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/MailboxImpl.java
```diff
@@ -158,44 +174,108 @@ public boolean tryPutMail(@Nonnull Runnable letter) {
 	}
 
 	@Override
-	public void putMail(@Nonnull Runnable letter) throws InterruptedException {
+	public void putMail(@Nonnull Runnable letter) throws InterruptedException, MailboxStateException {
```
pnowojski: Shouldn't we add a checkstate that we are not putting something from the mailbox thread?

StefanRRichter: From how the mailbox is threated in the PR, it is more of a datastructure to track runnable objects and acts somewhat like a queue. Not sure if I would implement knowledge about threads and what threads are allowed to do what here. This is now implemented on the level of the executor, i.e. the user-facing facade.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/Mailbox.java
StefanRRichter: I was thinking about that and decided against it for a couple of reasons. First, overhead is introduced even for caller that do not care about completion, cancelation etc., which seems to be most cases. Second, you can always build this on top and also the user-facing fassade of the executor service has methods that return `Future`. If you wonder why not `CompletableFuture` instead, it is because the user can always create such futures themselves via `CompletableFuture.runAsync(Runnable, Executor)`. However, there is a downside to this.
Such calls enqueue a special instance of `RunnableFuture` under the hood, but calling cancel on them is never propagated back to the `CompletableFuture`. That is just a general shortcoming of this implementation in the JDK in my opinion. So, if you want to also propagate cancellation to completable futures, that would mean you have to remember all of them, in parallel to the `Runnable`s, because they are not `Runnable`. So there are many options already possible with this interface that does not impose return values on all clients:
- no feedback
- Future
- compleable future via `runAsync`
- your own feedback mechanism if you submit instances of `Runnable(Future)`

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/TaskMailboxExecutorService.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+public interface TaskMailboxExecutorService extends TaskMailboxExecutor, ExecutorService {
+
+	/**
+	 * Returns the mailbox that manages the execution order.
+	 *
+	 * @return the mailbox.
+	 */
+	@Nonnull
+	Mailbox getMailbox();
```
Aitozi: Why here needs to expose the mailbox ?  Can we delegate all the interactions with mailbox to the MailboxExecutorService?

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
```diff
@@ -462,12 +486,20 @@ public final void invoke() throws Exception {
...
+			FutureUtils.cancelRunnableFutures(taskMailboxExecutor.shutdownNow());
 		}
 	}
 
 	@Override
 	public final void cancel() throws Exception {
-		mailbox.clearAndPut(POISON_LETTER);
+		try {
+			List<Runnable> droppedRunnables = taskMailboxExecutor.getMailbox().clearAndPut(mailboxPoisonLetter);
+			FutureUtils.cancelRunnableFutures(droppedRunnables);
```
Aitozi: At the moment the pending letters are not under real execution although they may be a future instance, why needs to do the cancel operation on them ?

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/Mailbox.java
```diff
@@ -20,17 +20,85 @@
...
+	 * The effect of this is that all pending letters in the mailbox are dropped and the given priorityLetter
+	 * is enqueued to the head of the mailbox. Dropped letters are returned. This method should only be invoked
+	 * by code that has ownership of the mailbox object and only rarely used, e.g. to submit special events like
+	 * shutting down the mailbox loop.
+	 *
+	 * @param priorityLetter action to enqueue atomically after the mailbox was cleared.
+	 * @throws MailboxStateException if the mailbox is quiesced or closed.
+	 */
+	@Nonnull
+	List<Runnable> clearAndPut(@Nonnull Runnable priorityLetter) throws MailboxStateException;
```
Aitozi: with `putFirst` to put poisionLetter at the head, do we still need this `clearAndPut`?

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
StefanRRichter: Currently this is done to signal interested clients that their letter will never be executed. I think this can be useful for shutdown in some cases.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/Mailbox.java
StefanRRichter: Main reason is that `putFirst` can block if the queue is currently full and therefore should not be invoked from the task thread, `clearAndPut` never blocks. So there is a distinct use-case for each.

StefanRRichter:
Updated the PR in a couple of ways, for example enforcing that some methods of the executor must be called from inside/outside of the mailbox thread, a check if we are alreay inside the mailbox thread and I removed potentially dangerous and not really required waiting (for mail or capacity) methods.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SourceStreamTask.java
```diff
@@ -123,14 +124,12 @@ protected void performDefaultAction(ActionContext context) throws Exception {
 		context.allActionsCompleted();
 	}
 
-	private void runAlternativeMailboxLoop() throws InterruptedException {
+	private void runAlternativeMailboxLoop() throws InterruptedException, MailboxStateException {
 
-		while (true) {
+		final Mailbox mailbox = taskMailboxExecutor.getMailbox();
```
1u0: Because you are using the mailbox here directly, you can also add a precondition similar to `taskMailboxExecutor.checkIsMailboxThread()`.

Otherwise, it's implicitly not compatible with some tests now. I've worked on a fix in the checkpoints-to-mailbox ticket, 
if you want, I can try to cherry pick them before your commit, to properly check that tests are good with your PR.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
```diff
@@ -252,23 +261,31 @@ protected StreamTask(
 	 * Runs the stream-tasks main processing loop.
 	 */
 	private void run() throws Exception {
-		final ActionContext actionContext = new ActionContext();
+		final Mailbox mailbox = taskMailboxExecutor.getMailbox();
```
1u0: Same as for `SourceStreamTask.runAlternativeMailboxLoop()`, you can add a precondition check that this method is called in the proper thread.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
```diff
@@ -1329,33 +1364,71 @@ public void tryHandleCheckpointException(CheckpointMetaData checkpointMetaData,
...
 		 * This method must be called to end the stream task when all actions for the tasks have been performed.
 		 */
 		public void allActionsCompleted() {
-			mailbox.clearAndPut(POISON_LETTER);
+			try {
+				if (Thread.currentThread() == mailboxThread) {
+					if (!mailbox.tryPutFirst(mailboxPoisonLetter)) {
+						// mailbox is full - in this particular case we know for sure that we will still run through the
+						// break condition check inside the mailbox loop and so we can just run directly.
+						mailboxPoisonLetter.run();
```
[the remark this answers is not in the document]
StefanRRichter: I think your comment was on an outdated version, it does this already.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
```diff
@@ -1329,33 +1364,71 @@ public void tryHandleCheckpointException(CheckpointMetaData checkpointMetaData,
...
-			mailbox.putMail(actionUnavailableLetter);
+		public void actionsUnavailable() {
+			putOrExecuteDirectly(actionUnavailableLetter);
+		}
+
+		private void putOrExecuteDirectly(Runnable letter) {
+			try {
+				if (Thread.currentThread() == mailboxThread) {
+					if (!mailbox.tryPutMail(letter)) {
+						letter.run();
```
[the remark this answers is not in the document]
StefanRRichter: This is a private method, that is only there to avoid some code duplication. No caller uses it directly, and for the two indirect use-cases that are exposed, they are about two special "signal" letters for which the sematics should be perfectly fine. But maybe you can go into more detail if you think I overlook a problem?

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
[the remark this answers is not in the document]
1u0: I'm reading commit version `d235a0d8b3ca2d60c91165fe3c3625437063392e`.

What I mean in my comment, that why wouldn't you just use:

```java
if (mailboxExecutorService.isMailboxThread()) {
    mailboxPoisonLetter.run();
} else {
    taskMailboxExecutor.getMailbox().putFirst(mailboxPoisonLetter);
}
```
?

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
```diff
@@ -1329,33 +1364,66 @@ public void tryHandleCheckpointException(CheckpointMetaData checkpointMetaData,
...
 	public final class ActionContext {
 
-		private final Runnable actionUnavailableLetter = ThrowingRunnable.unchecked(mailbox::waitUntilHasMail);
+		private final Runnable actionUnavailableLetter;
+		private final TaskMailboxExecutorServiceImpl mailboxExecutorService;
+
+		public ActionContext(TaskMailboxExecutorServiceImpl mailboxExecutorService) {
+			final Mailbox mailbox = taskMailboxExecutor.getMailbox();
+			this.actionUnavailableLetter = ThrowingRunnable.unchecked(() -> mailbox.takeMail().run());
+			this.mailboxExecutorService = mailboxExecutorService;
```
1u0: Nitpick: you are mixing two fields in the `ActionContext` implementation `TaskMailboxExecutorServiceImpl 
 mailboxExecutorService // this` and `TaskMailboxExecutorServiceImpl taskMailboxExecutor // enclosing StreamTask`.

Although, I'd expect that they both reference the same object.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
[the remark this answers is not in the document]
1u0: I agree for this being used only for signal letters and hopefully that `ActionContet` is used only internally.

By "weird" I mean that there can be situations at runtime, where the caller, for example that uses `ActionContext.actionsUnavailable()` may be blocked (from time-to-time) + it may execute a letter.

Imaginable example:
```java
// In some sub class of StreamTask
protected void performDefaultAction(ActionConext actionContext) {
   acquireResourceA();
   actionContext.actionsUnavailable();
   releaseA();
}
```

In this case, an execution can happen in different orders:
```
acquireResourceA
releaseA
// continue in the mailbox loop
```
OR
```
acquireResourceA
// execute some letter from the mailbox
releaseA
```

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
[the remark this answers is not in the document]
StefanRRichter: I think you are right, we can simplify it as you suggested.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SourceStreamTask.java
1u0: @StefanRRichter, the patch for tests: #8858

StefanRRichter:
Hi everybody! I updated and extended the PR quiet a bit, including a refactoring that makes testing mailbox execution possible and an improved signaling for (un)availability of default action. PR now also sits on top of #8858.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/MailboxProcessor.java
```diff
@@ -0,0 +1,319 @@
...
+	 * continue running, false if it should stop. In the current design, this method also evaluates all control flag
+	 * changes. This keeps the hot path in {@link #runMailboxLoop()} free from any other flag checking, at the cost
+	 * that all flag changes must make sure that the mailbox signals mailbox#hasMail.
+	 */
+	private boolean processMail(Mailbox mailbox) throws MailboxStateException, InterruptedException {
+
+		if (!mailbox.hasMail()) {
+			// We can directly return true because all changes to #isMailboxLoopRunning must be connected to
+			// mailbox.hasMail() == true.
+			return true;
```
1u0: Why not remove this `if` at all here? Currently I see some workarounds that check that the mailbox is not empty, but making it all fragile and not easy to reason about, imo.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/MailboxProcessor.java
```diff
@@ -0,0 +1,319 @@
...
+
+		SuspendDefaultActionRunnable() {
+			this.valid = true;
+		}
+
+		@Override
+		public void resume() {
+			try {
+				if (taskMailboxExecutor.isMailboxThread()) {
+					resumeInternal();
```
1u0: I'd move the mailbox thread clause outside of `try`:
```java
if (taskMailboxExecutor.isMailboxThread()) {
    resumeInternal();
    return;
}
try {
    ...
} catch (...) {
    ...
}

File: flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/TaskMailboxExecutorServiceImpl.java [hunk 17]
```diff
@@ -0,0 +1,154 @@
...
+	}
+
+	@Override
+	public boolean isTerminated() {
+		return mailbox.getState() == Mailbox.State.CLOSED;
+	}
+
+	@Override
+	public boolean awaitTermination(long timeout, @Nonnull TimeUnit unit) {
+		return isTerminated();
```
1u0: Nitpick: the implementation of this method is not strictly confirms `ExecutorService.awaitTermination()` contracts.
I think we are not using it directly or implicitly, so maybe just make it throw unimplemented exception?

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/MailboxProcessor.java
```diff
@@ -0,0 +1,319 @@
...
+			} catch (MailboxStateException me) {
+				LOG.debug("Action context could not submit letter to mailbox.", me);
+			}
+		}
+
+		private void resumeInternal() {
+			if (valid) {
+				valid = false;
+				suspendedDefaultAction = null;
+			}
```
1u0:
```java
if (suspendedDefaultAction == this) {
    suspendedDefaultAction = null;
}
```
?

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
```diff
@@ -226,13 +219,18 @@ protected StreamTask(
 		this.accumulatorMap = getEnvironment().getAccumulatorRegistry().getUserMap();
 		this.recordWriters = createRecordWriters(configuration, environment);
 		this.syncSavepointLatch = new SynchronousSavepointLatch();
-		this.mailbox = new MailboxImpl();
+		this.mailboxProcessor = new MailboxProcessor(this);
```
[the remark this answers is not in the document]
StefanRRichter: I did this, so as to avoid any changes to all the subclasses - but I guess then it's ok and I will simply have a couple more touched files.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/MailboxProcessor.java
```diff
@@ -0,0 +1,300 @@
...
+
+	/**
+	 * Runs the mailbox processing loop. This is where the main work is done.
+	 */
+	public void runMailboxLoop() throws Exception {
+
+		assert taskMailboxExecutor.isMailboxThread() :
+			"StreamTask::run must be executed by declared mailbox thread!";
+
+		final Mailbox localMailbox = mailbox;
```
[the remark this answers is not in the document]
StefanRRichter: I have seen this in many JDK classes that they copy fields, even `final` fields into local variables for hot paths. Maybe it is not required (anymore?) but I asume they have (had?) a reason to do so.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/MailboxProcessor.java
```diff
@@ -0,0 +1,319 @@
...
+	/**
+	 * This method exists to run the current sources with the mailbox model in a compatibility mode. Once we drop the
+	 * old source interface for the new one (FLIP-27) this method can eventually go away.
+	 *
+	 * @param checkpointLock the task's checkpointing lock.
+	 * @throws MailboxStateException if mailbox is closed.
+	 * @throws InterruptedException on interruption.
+	 */
+	public void switchToLegacySourceCompatibilityMailboxLoop(
+		final Object checkpointLock) throws MailboxStateException, InterruptedException {
```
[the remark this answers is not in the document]
StefanRRichter: It's just two lines and so if we really want to remove that later it is no big deal.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/MailboxProcessor.java
StefanRRichter: Because it saves you to read a field in every loop iteration.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/MailboxProcessor.java
StefanRRichter: Makes sense, the method worked a bit different before but now we can limit the scope of the block 👍

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/TaskMailboxExecutorServiceImpl.java
StefanRRichter: Ok, unsupported seems good enough.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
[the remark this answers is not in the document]
1u0: I think you wouldn't need to touch subclasses with the change I've proposed.

My main motivation is to not have `MailboxDefaultAction` implementation exposed as `public` api of `StreamTask`, as it's more internal implementation detail.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/MailboxProcessor.java
StefanRRichter: I think it is better like this, or else client code could accidentally keep a reference on one outdated suspended action forever and it would always be able to manipulate the flag.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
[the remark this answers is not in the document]
StefanRRichter: I would have to prefix the class name. But anyways I changed and refactored it a bit more, I was already tempted to rename `ActionContext` to `DefaultActionContext` before.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/MailboxProcessor.java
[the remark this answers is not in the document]
1u0: Maybe it makes sense.
Would it be reasonable then to apply same optimization trick to `switchToLegacySourceCompatibilityMailboxLoop` loop?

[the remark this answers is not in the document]
StefanRRichter: It probably would not hurt.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/MailboxProcessor.java
1u0: If it's about optimization, what do you think about the following variant:
```java
if (mailbox.hasMail()) {
	// TODO consider batched draining into list and/or limit number of executed letters
	// Take letters in a non-blockingly and execute them.
	Optional<Runnable> maybeLetter;
	while (isMailboxLoopRunning() && (maybeLetter = mailbox.tryTakeMail()).isPresent()) {
		maybeLetter.get().run();
	}
}

// If the default action is currently not available, we can run a blocking mailbox execution until the default
// action becomes available again.
while (isDefaultActionUnavailable() && isMailboxLoopRunning()) {
	mailbox.takeMail().run();
}

return isMailboxLoopRunning();
```
?

StefanRRichter: I was thinking about that, still this one reads two more fields per iteration than the other variant, at the cost of having to do a bit more in the uncommon path (adding a letter).

StefanRRichter: The suggestion would also be only a partial fix, not covering the case where we want to terminate the loop.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/MailboxProcessor.java
1u0: The suggested snipped should behave exactly the same as it's now in the PR.

>  client code could accidentally keep a reference on one outdated suspended action forever and it would always be able to manipulate the flag.

If by "outdated" you mean a `SuspendDefaultActionRunnable` instance that was already called `resumeInternal()`, then it would set `valid = false` and next invocations of this method have no effect.

On the other hand, you have invariants in the code, that:
 * if `SuspendDefaultActionRunnable.valid`, then it's the same instance that referenced in `MailboxProcessor.suspendedDefaultAction`.
 * if there are multiple instances of `SuspendDefaultActionRunnable` returned by `MailboxProcessor.suspendDefaultAction()`, only the latest one can have `valid = true`.

StefanRRichter: True, so I will remove the flag.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/MailboxProcessor.java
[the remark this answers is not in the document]
StefanRRichter: FYI, I removed the legacy loop method in favor of running it by signaling a suspended default action until the legacy source thread is finished.

StefanRRichter:
Thanks a lot for the reviews, will squash and merge when Travis is green.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/MailboxImpl.java
```diff
@@ -38,39 +40,44 @@
 public class MailboxImpl implements Mailbox {
 
 	/**
-	 * The enqueued letters.
+	 * Node for the embedded linked list that represents the mailbox queue.
```
pnowojski: nit: "embedded single linked list"?

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/MailboxImpl.java
```diff
@@ -38,39 +40,44 @@
...
+	private Node head;
 
 	/**
-	 * Condition that is triggered when the buffer is no longer full.
+	 * The tail of the linked list of mailbox actions.
 	 */
-	@GuardedBy("lock")
-	private final Condition notFull;
+	@Nullable
+	private Node tail;
```
pnowojski: I think the singly linked list implementation should be extracted to a separate class. Especially that we might want to evolve this structure in the future, with for example an iterator that can remove things efficiently.

Also, is there no singly linked list implementation anywhere in our class path (dependencies?).

Another idea (maybe even better), to not pre mature optimise things: just use doubly `LinkedList`? In fact I think I would be in favour of just doing that as it will save us both implementation, maintaining & testing efforts.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/MailboxImpl.java
```diff
@@ -38,39 +40,44 @@
...
-	 * Index of the ring buffer head.
+	 * Lock for all concurrent ops.
 	 */
-	@GuardedBy("lock")
-	private int headIndex;
+	private final ReentrantLock lock;
 
 	/**
-	 * Index of the ring buffer tail.
+	 * Condition that is triggered when the buffer is no longer empty.
```
pnowojski: `buffer` is no longer relevant. Java doc's link (`{@link #ringBuffer}`) would have guarded against that.

Extracting linked list to a separate class/using `LinkedList` would also allow to keep the java doc link to a new field.

StefanRRichter:
Have added the requested changes and rebased. Failing test `throwingExceptionOnCallbackWithNoRestartsShouldFailTheSuspend` also fails on master for me so I guess it is unrelated.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/MailboxImpl.java
```diff
@@ -80,7 +58,7 @@
 	private final Condition notEmpty;
 
 	/**
-	 * Number of letters in the mailbox.
+	 * Number of letters in the mailbox. We track it separately from the queue#size to avoid locking on {@link #hasMail()}.
```
pnowojski: nitty nit: Does the `{@link qeuue#size}` work? or `{@link queue}#size`?

StefanRRichter: I tried that, it did not work.

StefanRRichter:
Thanks for the reivew @pnowojski . Will merge.

PR #8693, opened 2019-06-11
[FLINK-8871] Support to cancel checkpoing via notification

Myasuka:

[**update**]
Use notification to abort checkpoint and take effect in the phase of trigger/before executing sync and executing async phase of `StreamTask`.

~~Use notification to abort checkpoint, instead of previous inflexible `CancelCheckpointMarker` mechanism.~~

## Brief change log

  -  If some specific checkpoint is notified as aborted. That checkpoint-to-abort would be stored in  `StreamTask`. When `StreamTask` starts to trigger or in sync or async phase of checkpoint, the target checkpoint would be aborted.

## Verifying this change

  - Add tests in `OneInputStreamTaskTest` to verify the aborted checkpoint could take effect in sync phase and async phase.

  - The serializers: **no**

## Documentation


Myasuka:
Previous check failed due to downloading kafka tar package unsuccessfully (please refer to build [logs](https://api.travis-ci.org/v3/job/544220294/log.txt)). Rebase latest code to re-trigger again.

StephanEwen:
@Myasuka Thank you for working on this feature. This will be a good addition to Flink. The general mechanism makes sense to me.

I noticed a few issues in the code that would be great to address before merging.

**(1)** The trigger of checkpoints on barrier needs also to check if the checkpoint is already aborted. Similar to the triggering on RPC.
  
**(2)** I think it would be worth considering to refactor all checkpointing parts into a single place. Something like a `Checkpoints` class that captures all currently running and pending / aborted checkpoints. 

That would simplify a lot, and also make tests much easier, because tests would not need to deal with mocking the `StreamTask`.

**(3)** The PR introduces more lock synchronization, while the remaining parts are going for the mailbox instead. At least the checkpoint lock should be avoided in favor of the mailbox executor.

**(4)** We should probably rename `notifyCheckpointAbort(...)` to `notifyCheckpointAborted(...)`

**(5)** The `notifyCheckpointAbort` method in the `CheckpointListener` interface should probably not be a default method. Implementers should generally be forced to think about what to do when checkpoint is aborted.

I think given the abstractions right now, it would make sense for the `AbstractStreamOperator` to have a default implementation of that method that is empty.

Myasuka:
@StephanEwen thanks for your review and sorry for the reply. I have actually almost completed the refactoring checkpointing part of StreamTask locally and just crated a new issue [FLINK-14652](https://issues.apache.org/jira/browse/FLINK-14652) to track this.

BTW, what do you think of removing `lock` within notify checkpoint complete/aborted interfaces? From my side, as we still need to support legacy source, we cannot easily remove these `synchronized (lock)` in the notifications now.

StephanEwen:
About the lock removal: If we make actions go through the mailbox, then the mailbox implementation takes care of abstracting the backwards compatibility for legacy sources.

The fact that we have legacy sources should ideally not be visible in this code at all, otherwise we have to touch the code again when we remove the legacy sources or change the compatibility approach or locking model.

rkhachatryan:
Thanks for the contribution, @Myasuka .

Can you please rebase your changes to the latest master?
As we moved most checkpointing-related functionality recently from `StreamTask` to `SubtaskCheckpointCoordinator`, I think task changes should go there.

I have some high-level remarks:
1. does it make sense to split the commit (or even PR) into several pieces addressing JM and TM separately?
2. can you add a component name to the commit message?
3. `AsyncCheckpointRunnable.cancelled` seems not to be used. And why not to add a new AsyncCheckpointState instead (or reuse `DISCARDED`)?
4. I don't see how `AsyncCheckpointRunnable` is canceled or its thread is stopped; which is the original goal of this PR I believe

Myasuka:
@rkhachatryan Thanks for your review!
I'll rebase master code on top of https://github.com/apache/flink/pull/11721 recently.

1. > does it make sense to split the commit (or even PR) into several pieces addressing JM and TM separately?
    
I also previous single commit is a bit large, I'll split into several smaller ones.

2. > can you add a component name to the commit message?

Sure, no problem. It was my fault to missing component information.

3. > AsyncCheckpointRunnable.cancelled seems not to be used. And why not to add a new AsyncCheckpointState instead (or reuse DISCARDED)?

Previous `AsyncCheckpointRunnable.cancelled` is used for testing to verify whether this async runnable is cancelled. The reason why I not add a new `AsyncCheckpointState` is due to I reuse `AsyncCheckpointRunnable#close` to cancel the running thread which would set the state as `DISCARDED`. I think this is an open question to discuss to search for a better solution.

4. > I don't see how AsyncCheckpointRunnable is canceled or its thread is stopped; which is the original goal of this PR I believe

As I said above, I reuse `AsyncCheckpointRunnable#close` to cancel the running async checkpoint task.

Myasuka:
@rkhachatryan Would you please review this PR again?
I have split into two commits now and rebase with latest code.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
```diff
@@ -182,7 +181,7 @@
 	protected final TimerService timerService;
 
 	/** The currently active background materialization threads. */
-	private final CloseableRegistry cancelables = new CloseableRegistry();
+	private final AsyncCheckpointRunnableRegistry cancelables = new AsyncCheckpointRunnableRegistry();
```
rkhachatryan, line 184: This registry stores not only `AsyncCheckpointRunnableRegistry` but also `SubtaskCheckpointCoordinator` (and other objects).

Setting it to `AsyncCheckpointRunnableRegistry` makes me think it only stores Runnables and should go into `SubtaskCheckpointCoordinatorImpl`.

I'd suggest to explicitly keep track of async checkpointing tasks in `SubtaskCheckpointCoordinatorImpl`.

flink-runtime/src/main/java/org/apache/flink/runtime/checkpoint/CheckpointCoordinator.java
```diff
@@ -1072,6 +1073,7 @@ public void run() {
 					}
 				});
 
+				sendAbortedMessages(checkpointId, pendingCheckpoint.getCheckpointTimestamp());
```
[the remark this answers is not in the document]
Myasuka: I think this is a good question. We need notify checkpoint complete to tasks ASAP if we use two-phase commit mechanism, thus `sendAcknowledgeMessages` is executed synchronously. Notify checkpoint aborted to tasks should not be so urgent and could be executed asynchronously. I'll change this behavior in this PR.

flink-runtime/src/main/java/org/apache/flink/runtime/state/CheckpointListener.java
```diff
@@ -38,4 +38,12 @@
...
 	 */
 	void notifyCheckpointComplete(long checkpointId) throws Exception;
+
+	/**
+	 * This method is called as a notification once a distributed checkpoint has been aborted.
+	 *
+	 * @param checkpointId The ID of the checkpoint that has been aborted.
+	 * @throws Exception
+	 */
+	default void notifyCheckpointAborted(long checkpointId) throws Exception {}
```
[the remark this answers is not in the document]
Myasuka: I think this could still stay at `CheckpointListener.java` as this method could take effect in many places. Some state backend could leverage this method to do some cleanup things, especially those disAgg (separating storage and computation) state-backends. Moreover, we also want to introduce `notifyCheckpointSubsume` if possible in future, and too many interfaces would be a bit over design.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/AbstractStreamOperator.java
```diff
@@ -340,7 +340,16 @@ public void initializeState(StateInitializationContext context) throws Exception
 
 	@Override
 	public void notifyCheckpointComplete(long checkpointId) throws Exception {
-		stateHandler.notifyCheckpointComplete(checkpointId);
+		if (stateHandler != null) {
```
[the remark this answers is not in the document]
Myasuka: Emmm, it's just for unit test with operator not calling `initializeState`, I could try to remove this if possble.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/StreamOperatorStateHandler.java
```diff
@@ -222,6 +222,12 @@ public void notifyCheckpointComplete(long checkpointId) throws Exception {
 		}
 	}
 
+	public void notifyCheckpointAborted(long checkpointId) throws Exception {
+		if (keyedStateBackend != null) {
+			keyedStateBackend.notifyCheckpointAborted(checkpointId);
+		}
+	}
```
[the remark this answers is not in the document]
Myasuka: If user develop their own state backend and rely on this method to do some necessary cleanup, this must be implemented here. Actually, that's what we rely for our internal enterprise state-backend.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
Myasuka: In the beginning, I plan to wrap `CloseableRegistry` and found I have to use another lock to synchronized. That's why I currently have to extend `CloseableRegistry` to leverage the same `getSynchronizationLock()`.
And I agree the name of `AsyncCheckpointRunnableRegistry` is misleading, I'll prepare to wrap `CloseableRegistry` again as I think another internal lock should not harm much.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SubtaskCheckpointCoordinatorImpl.java
```diff
@@ -93,6 +101,8 @@
 		this.unalignedCheckpointEnabled = unalignedCheckpointEnabled;
 		this.prepareInputSnapshot = prepareInputSnapshot;
 		this.closeableRegistry.registerCloseable(this);
+		this.abortedCheckpointIds = new ConcurrentSkipListSet<>();
```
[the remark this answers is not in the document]
Myasuka: Good catch, this feature is developed when `StreamTask` has no mailbox, that's why we use a `ConcurrentSkipListSet` here, will refactor to `TreeSet`.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SubtaskCheckpointCoordinatorImpl.java
```diff
@@ -183,6 +209,41 @@ public void notifyCheckpointComplete(long checkpointId, OperatorChain<?, ?> oper
 		env.getTaskStateManager().notifyCheckpointComplete(checkpointId);
 	}
 
+	@Override
+	public void notifyCheckpointAborted(long checkpointId, OperatorChain<?, ?> operatorChain, Supplier<Boolean> isRunning) throws Exception {
```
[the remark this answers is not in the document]
Myasuka: `notifyCheckpointAborted` is executed asynchronously, and I prefer a `Suppluer` here to provide more accurate result.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SubtaskCheckpointCoordinatorImpl.java
```diff
@@ -183,6 +209,41 @@ public void notifyCheckpointComplete(long checkpointId, OperatorChain<?, ?> oper
...
+
+		if (isRunning.get()) {
+			LOG.debug("Notification of aborted checkpoint for task {}", taskName);
+			// only happens when the task always received checkpoints to abort but never trigger or executing.
+			if (abortedCheckpointIds.size() >= DEFAULT_MAX_RECORDED_ABOTRED_CHECKPOINTS) {
+				abortedCheckpointIds.pollFirst();
+			}
```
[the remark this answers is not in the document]
Myasuka: This feature develops as once two very large jobs queued a lot of checkpoint requests (cannot get the checkpoint lock) on task side. The checkpoint on JM side just expire and trigger new one, while they cannot execute on task side. After nearly one day, many queue checkpoints flood as the lock is released and lead to many operators to execute checkpoint at the same time continuously, which lead to the DFS crashed and unavailable.
That's why we store those notified checkpoints, not sure whether have been explained clearly.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SubtaskCheckpointCoordinatorImpl.java
```diff
@@ -183,6 +209,41 @@ public void notifyCheckpointComplete(long checkpointId, OperatorChain<?, ?> oper
...
+
+			boolean canceled = closeableRegistry.cancelAsyncCheckpointRunnable(checkpointId);
```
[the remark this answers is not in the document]
Myasuka: I think the `channelWrittenFuture` is not canceled is during normal cancel phase existed even this feature is not introduced. And storing the `Future<?>` cannot ensure canceling that future could discard those written state handles, thus we should store the `AsyncCheckpointRunnable` itself.

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/SubtaskCheckpointCoordinatorTest.java
```diff
@@ -58,4 +83,212 @@ public void testNotifyCheckpointComplete() throws Exception {
...
+
+		subtaskCheckpointCoordinator.checkpointState(
+			new CheckpointMetaData(checkpointId, System.currentTimeMillis()),
+			CheckpointOptions.forCheckpointWithDefaultLocation(),
+			new CheckpointMetrics(),
+			operatorChain,
+			() -> true);
+		assertFalse(checkpointOperator.isCheckpointed());
+		assertEquals(0, subtaskCheckpointCoordinator.getAbortedCheckpointSize());
+		assertEquals(0, subtaskCheckpointCoordinator.getCloseableRegistry().getAsyncCheckpointRunnables().size());
```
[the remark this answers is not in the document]
Myasuka: Actually, `checkpointOperator.isCheckpointed()` has already verified that "operator snapshots are not taken".
I could add `assertEquals(-1, stateManager.getReportedCheckpointId());` to verify `TaskStateManager` has not been reported.

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/SubtaskCheckpointCoordinatorTest.java
```diff
@@ -58,4 +83,212 @@ public void testNotifyCheckpointComplete() throws Exception {
...
+		}
+	}
+
+	private <T> OneInputStreamOperator<T, T> streamOperatorWithSnapshot(OperatorSnapshotFutures operatorSnapshotResult) throws Exception {
+		@SuppressWarnings("unchecked")
+		OneInputStreamOperator<T, T> operator = mock(OneInputStreamOperator.class);
+		when(operator.getOperatorID()).thenReturn(new OperatorID());
+
+		when(operator.snapshotState(anyLong(), anyLong(), any(CheckpointOptions.class), any(CheckpointStreamFactory.class)))
+			.thenReturn(operatorSnapshotResult);
```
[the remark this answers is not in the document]
Myasuka: Thanks for you suggestion, I'll enrich `CheckpointOperator`.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/AbstractStreamOperator.java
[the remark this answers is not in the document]
rkhachatryan: That would be nice because it makes the developer think it's also nullable in production code.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SubtaskCheckpointCoordinatorImpl.java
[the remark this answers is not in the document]
rkhachatryan: I'm not sure I fully understood your explanation. I think you are talking about why the limit exists at all, while I meant that it should be enforced later - after the line 252:
```
abortedCheckpointIds.add(checkpointId);
```

flink-runtime/src/main/java/org/apache/flink/runtime/state/CheckpointListener.java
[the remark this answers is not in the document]
rkhachatryan: > Some state backend could leverage this method to do some cleanup things, especially those disAgg (separating storage and computation) state-backends. 

If any of state backends will implement it, it won't be too difficult to add a call.

> too many interfaces would be a bit over design.

I think it's better to have many small interfaces rather than a bigger one (I believe it's a general recommendation).
It's also not over design here because this is an important part of the system.

As it is now, there is a lot of "dead" code with empty methods which complicates the design.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
rkhachatryan: But why not to store `AsyncCheckpointRunnable`s in `SubtaskCheckpointCoordinatorImpl`?
That is, move `Map<Long, AsyncCheckpointRunnable> checkpoints` from `AsyncCheckpointRunnableRegistry` to `SubtaskCheckpointCoordinatorImpl`.

That would make it more clear and flexible (we could change the type in map, we could avoid passing registry to `SubtaskCheckpointCoordinatorImpl` etc.).

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/SubtaskCheckpointCoordinatorTest.java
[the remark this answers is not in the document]
rkhachatryan: Makes sense. 
Another thing I meant is that we shouldn't expose implementation details unnecessarily. 
So ideally we could remove `SubtaskCheckpointCoordinatorImpl.getAsyncCheckpointRunnableRegistry` and 
`AsyncCheckpointRunnableRegistry.getAsyncCheckpointRunnables`.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/StreamOperatorStateHandler.java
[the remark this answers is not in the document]
rkhachatryan: I highlighted the wrong lines, sorry.
I meant that there is already a default implementation in interface and there is no need in empty methods in `RocksDBKeyedStateBackend` and `HeapKeyedStateBackend`:
```
@Override
public void notifyCheckpointAborted(long checkpointId) {
	// nothing to do
}
```

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SubtaskCheckpointCoordinatorImpl.java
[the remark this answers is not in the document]
rkhachatryan: `channelWrittenFuture` is canceled by the `ChannelStateWriter` during "normal" abort (its state handle doesn't need to be discarded). 
This PR only "speeds up" the abort, cancellation marker will follow, right?  So eventually it will be canceled too.

Any other interruptible operation (e.g. reporting to JM) should be interrupted by canceling the future returned by the `executorService`.  But this is probably out of scope of this PR.

flink-runtime/src/test/java/org/apache/flink/runtime/state/TaskLocalStateStoreImplTest.java
```diff
@@ -153,6 +153,21 @@ public void confirmCheckpoint() throws Exception {
...
+	@Test
+	public void abortCheckpoint() throws Exception {
+
+		final int chkCount = 4;
+		final int aborted = chkCount - 2;
+		List<TaskStateSnapshot> taskStateSnapshots = storeStates(chkCount);
+		taskLocalStateStore.abortCheckpoint(aborted);
+		checkPrunedAndDiscarded(taskStateSnapshots, aborted, aborted + 1);
+		checkStoredAsExpected(taskStateSnapshots, 0, aborted);
+		checkStoredAsExpected(taskStateSnapshots, aborted + 1, chkCount);
```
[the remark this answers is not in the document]
Myasuka: `taskLocalStateStore.retrieveLocalState(checkpointId)` would actually already check snapshot was removed from `TaskLocalStateStoreImpl.storedTaskStateByCheckpointID`.

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/SubtaskCheckpointCoordinatorTest.java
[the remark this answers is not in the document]
Myasuka: Got your point, already removed this two methods and verify the logic using other ways.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
Myasuka: Because `asyncCheckpointRunnable` is registered to `closeableRegistry` when it begins to run asynchronously, we have to introduce something to wrap the `checkpointId`, `AsyncCheckpointRunnable` and `closeableRegistry` to let cancellation available, which is the role `AsyncCheckpointRunnableRegistry` played. Even we move `Map<Long, AsyncCheckpointRunnable> checkpoints` from `AsyncCheckpointRunnableRegistry` to `SubtaskCheckpointCoordinatorImpl`, we still need something to warp them all.

Moreover, since we register the `closeableRegistry` within `AsyncCheckpointRunnable`, and we create the runnable in `SubtaskCheckpointCoordinatorImpl`, I'm afraid it's not easy to avoid passing registry to `SubtaskCheckpointCoordinatorImpl`.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SubtaskCheckpointCoordinatorImpl.java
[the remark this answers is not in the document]
Myasuka: Thanks for your clarification, I would also try to call `ChannelStateWriter#abort` during `SubtaskCheckpointCoordinator#notifyCheckpointAborted` and add logic to verify this in unit test.

Myasuka:
@flinkbot run azure

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SubtaskCheckpointCoordinatorImpl.java
```diff
@@ -60,17 +63,22 @@
...
+	private final AsyncCheckpointRunnableRegistry asyncCheckpointRunnableRegistry;
 	private final ExecutorService executorService;
 	private final Environment env;
 	private final AsyncExceptionHandler asyncExceptionHandler;
 	private final ChannelStateWriter channelStateWriter;
 	private final StreamTaskActionExecutor actionExecutor;
 	private final boolean unalignedCheckpointEnabled;
 	private final BiFunctionWithException<ChannelStateWriter, Long, CompletableFuture<Void>, IOException> prepareInputSnapshot;
+	/** The IDs of the checkpoint for which we are notified aborted. */
+	private final NavigableSet<Long> abortedCheckpointIds;
```
pnowojski, line 79: Why do we need to track the aborted checkpoints?

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
rkhachatryan: > we still need something to warp them all.

Do you mean a scenario when all runnables need to be closed? 
For this we can explicitly iterate over a map and close them in `SubtaskCheckpointCoordinatorImpl.close()` (it's already registered with `closeableRegistry`).


> it's not easy to avoid passing registry

I didn't mean we should do it now but it's one thing we can consider in future. Having it tied to runnables makes it a bit harder.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SubtaskCheckpointCoordinatorImpl.java
```diff
@@ -183,6 +235,42 @@ public void notifyCheckpointComplete(long checkpointId, OperatorChain<?, ?> oper
...
+	public void notifyCheckpointAborted(long checkpointId, OperatorChain<?, ?> operatorChain, Supplier<Boolean> isRunning) throws Exception {
+
+		if (isRunning.get()) {
+			LOG.debug("Notification of aborted checkpoint for task {}", taskName);
+			// only happens when the task always received checkpoints to abort but never trigger or executing.
+			if (abortedCheckpointIds.size() >= maxRecordAbortedCheckpoints) {
+				abortedCheckpointIds.pollFirst();
+			}
+
+			channelStateWriter.abort(checkpointId, new NotifiedCheckpointAbortedException(checkpointId));
```
rkhachatryan, line 248: `try/catch`?
same for the operators

flink-runtime/src/main/java/org/apache/flink/runtime/state/heap/HeapKeyedStateBackend.java
```diff
@@ -311,6 +311,11 @@ public void notifyCheckpointComplete(long checkpointId) {
 		//Nothing to do
 	}
 
+	@Override
+	public void notifyCheckpointAborted(long checkpointId) {
+		// nothing to do
+	}
+
```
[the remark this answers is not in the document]
rkhachatryan: (or the other way around, as @StephanEwen mentioned: no default implementation)

flink-runtime/src/main/java/org/apache/flink/runtime/state/CheckpointListener.java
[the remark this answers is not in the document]
rkhachatryan: As there are already such state backends, as @Myasuka mentioned below, it makes sense
(regarding multiple interfaces, they can simply be combined (with `extends`) and given to the user).

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SubtaskCheckpointCoordinatorImpl.java
Myasuka: The notification of aborted checkpoints could come before task start to execute the checkpoint.
We store aborted checkpoints so that we could avoid to execute any sync or async phase of checkpoint on task side when we know this checkpoint has been aborted.

rkhachatryan:
Thanks for the updates @Myasuka 

There are still some questions open, please see the comments above.

Besides that, can you please list some IT cases covering checkpoint cancelation in the description?
 
I'm a bit concerned about the following scenarios:
1. a barrier arrives for an already canceled checkpoint; alignment still happens as usual but the result is discarded, right?
2. recently, upstream tasks started to block in aligned mode (#11351). With this change, they still will be unblocked by the downstream despite of cancelation via RPC, right?

flink-runtime/src/main/java/org/apache/flink/runtime/state/heap/HeapKeyedStateBackend.java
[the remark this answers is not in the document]
Myasuka: I forget Stephan's suggestion to remove the default implementation, I'll update to remove the default implementation of `notifyCheckpointAborted`

Myasuka:
@rkhachatryan thanks for your review. It's quite late for me at my time now and plan to update at my day time.
Before go to sleep, I could answer the 1st question: barrier alignment would still happen as usual, current PR would not affect barrier align part as discussed in the beginning, and we could improve this in the future.

flink-state-backends/flink-statebackend-rocksdb/src/main/java/org/apache/flink/contrib/streaming/state/RocksDBKeyedStateBackend.java
```diff
@@ -464,6 +464,11 @@ public void notifyCheckpointComplete(long completedCheckpointId) throws Exceptio
 		}
 	}
 
+	@Override
+	public void notifyCheckpointAborted(long checkpointId) {
+		// nothing to do
+	}
+
```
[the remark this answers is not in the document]
Myasuka: I prefer the idea to not leave default implementation of `notifyCheckpointAborted`, thus leave it here could be acceptable.

Myasuka:
For the 2nd question @romanovacca , after discussed with @zhijiangW offline, we believe this PR with cancellation via RPC would not take part in the data transmission in channel, thus the upstream task would be unblocked once the channel is unblocked which is the same logic as before.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SubtaskCheckpointCoordinatorImpl.java
Myasuka: Would you please give more clear explanations for this? Did you mean we cannot call `channelStateWriter.abort` twice. Moreover, what do you mean for `same for the operators`?

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SubtaskCheckpointCoordinatorImpl.java
```diff
@@ -183,6 +235,42 @@ public void notifyCheckpointComplete(long checkpointId, OperatorChain<?, ?> oper
...
+
+			if (!canceled) {
+				if (checkpointId > lastCheckpointId) {
+					// only record checkpoints that have not triggered on task side.
+					abortedCheckpointIds.add(checkpointId);
+				}
+			}
+
+			for (StreamOperatorWrapper<?, ?> operatorWrapper : operatorChain.getAllOperators(true)) {
+				operatorWrapper.getStreamOperator().notifyCheckpointAborted(checkpointId);
```
rkhachatryan, line 259: What if `notifyCheckpointAborted` fails for one operator, should we continue to the next?

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SubtaskCheckpointCoordinatorImpl.java
rkhachatryan: Sorry for not being clear. We shouldn't call `channelStateWriter.abort` before receiving a cancelation marker.

> Moreover, what do you mean for same for the operators?

Added a separate comment.

Myasuka: If the async checkpoint on task side is aborted via RPC message, and we did not call `channelStateWriter.abort`. If above stream did not broadcast a cancellation marker to downstream, what happened to `channelStateWriter` as that is not called as aborted.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SubtaskCheckpointCoordinatorImpl.java
Myasuka: I think this is the same for `notifyCheckpointComplete`, catch the exception and continue to the next operator should be a solution.

Myasuka:
I think current IT cases do not target to verify the cancel notification, I'll create one to target to veriy this.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
Myasuka: Actually, current `AsyncCheckpointRunnableRegistry` is a wrapper of `closeableRegistry` and ` Map<Long, AsyncCheckpointRunnable> checkpoints`.
And if we do not use a wrapper but with two separate `closeableRegistry` and `checkpoints`,  the `AsyncCheckpointRunnable` would register itself to `closeableRegistry` and put itself with checkpoint id to `checkpoints` when begin async phase. 
Once it reach the finally phase of the runnable, it will remove from the `checkpoints` and then unregister from `closeableRegistry`. Actually, I cannot see any real difference compared with current `AsyncCheckpointRunnableRegistry`.
Perhaps the name of `AsyncCheckpointRunnableRegistry` could easily mislead developers? And a name like wrapper is better?

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SubtaskCheckpointCoordinatorImpl.java
rkhachatryan: In that case, `CheckpointBarrierHandler` will call abort on the next barrier.
Though I don't think it's possible that task skips the cancellation marker.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
rkhachatryan: I think that `SubtaskCheckpointCoordinator` shouldn't delegate *its* responsibility of maintaining _its_ runnables to `StreamTask.closeableRegistry`. Changing the names only masquerades it.

Using the same registry creates unnecessary connection, making it more difficult to change `StreamTask` (for example, to remove registry type).

At runtime, I guess behavior is more or less the same.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SubtaskCheckpointCoordinatorImpl.java
```diff
@@ -183,6 +235,42 @@ public void notifyCheckpointComplete(long checkpointId, OperatorChain<?, ?> oper
...
 	}
 
+	@Override
+	public void notifyCheckpointAborted(long checkpointId, OperatorChain<?, ?> operatorChain, Supplier<Boolean> isRunning) throws Exception {
+
+		if (isRunning.get()) {
+			LOG.debug("Notification of aborted checkpoint for task {}", taskName);
+			// only happens when the task always received checkpoints to abort but never trigger or executing.
+			if (abortedCheckpointIds.size() >= maxRecordAbortedCheckpoints) {
+				abortedCheckpointIds.pollFirst();
```
rkhachatryan, line 245: Why do we remove the oldest checkpoint? Why not the one that was _canceled first_?
I guess both are valid options, but for the 2nd we could use `LinkedHashMap` with O(1) time complexity.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SubtaskCheckpointCoordinatorImpl.java
rkhachatryan: For `notifyCheckpointComplete` it's less important because operators are expected to work correctly even without the notification (and `channelStateWriter` has only in-memory resources).

Another question: this is a new method and we call it only on RPC notification, right? Should we also call it in `abortCheckpointOnBarrier`?

flink-tests/src/test/java/org/apache/flink/test/checkpointing/NotifyCheckpointAbortedITCase.java
```diff
@@ -0,0 +1,576 @@
...
+			// found the local state manager has stored states for checkpoint-1.
+			localStoredStates.addAll(collectLocalStoredStates());
+			if (!localStoredStates.isEmpty()) {
+				break;
+			}
+		}
+
+		// let the checkpoint-1 failed finally.
+		TestingCompletedCheckpointStore.abortCheckpointLatch.trigger();
+		while (deadline.hasTimeLeft()) {
```
rkhachatryan, line 229: I think this check will pass if the deadline is missed. Should it fail instead?

flink-tests/src/test/java/org/apache/flink/test/checkpointing/NotifyCheckpointAbortedITCase.java
```diff
@@ -0,0 +1,576 @@
...
+			}
+		}
+
+		// wait for StuckAsyncCheckpointMap notified as checkpoint aborted.
+		StuckAsyncCheckpointMap.checkpointAbortedLatch.await(deadline.timeLeft().toMillis(), TimeUnit.MILLISECONDS);
+		assertTrue(stuckAsyncSnapshotStrategy.blockingRunnableFuture.isCancelled());
+
+		// verify BeforeExecuteCheckpointSink never execute checkpoint-1.
+		while (deadline.hasTimeLeft()) {
+			if (!BeforeExecuteCheckpointSink.snapshotIds.isEmpty()) {
```
rkhachatryan, line 243: Can we use accumulators instead of a static field?
It could save from sync/wait and improve readability (no static field).

flink-tests/src/test/java/org/apache/flink/test/checkpointing/NotifyCheckpointAbortedITCase.java
```diff
@@ -0,0 +1,576 @@
...
+	 * by 'DeclineSink'. Then we verify the async runnable future of 'StuckAsyncCheckpointMap' was canceled as expected,
+	 * and 'BeforeExecuteCheckpointSink' did not execute the sync phase of checkpoint.
+	 *
+	 * <p>The job graph looks like:
+	 * NormalSource --> keyBy --> NormalMap --> StuckAsyncCheckpointMap --> DeclineSink
+	 *                                  |
+	 *                                  |--> BarrierDelayMap -> BeforeExecuteCheckpointSink
+	 */
+	@Test(timeout = TEST_TIMEOUT)
+	public void testNotifyCheckpointAborted() throws Exception {
```
rkhachatryan, line 183: This test does a great job of testing the functionality :+1: 

However, considering its complexity and unit-test coverage, I think a valid trade-off would be to simply test that operator is notified that the checkpoint is aborted.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SubtaskCheckpointCoordinatorImpl.java
Myasuka: I think this suggestion sounds reasonable, changed to `ArrayDeque`.

flink-runtime/src/main/java/org/apache/flink/runtime/state/CheckpointListener.java
```diff
@@ -38,4 +38,12 @@
...
 	 */
 	void notifyCheckpointComplete(long checkpointId) throws Exception;
+
+	/**
+	 * This method is called as a notification once a distributed checkpoint has been aborted.
+	 *
+	 * @param checkpointId The ID of the checkpoint that has been aborted.
+	 * @throws Exception
+	 */
+	void notifyCheckpointAborted(long checkpointId) throws Exception;
```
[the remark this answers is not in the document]
Myasuka: Once this PR merged, I planned to add release notes in JIRA. From my previous experience, this interface change might not publish in the final release notes of flink official website.

pnowojski:
There is a test failure:
```
SubtaskCheckpointCoordinatorTest.testNotifyCheckpointAbortedDuringAsyncPhase:195 » IllegalArgument
```

Myasuka:
@pnowojski thanks for your remind.
I think I have addressed all comments to update the commits.

Since current CI is still pending, my local pipeline based on the same commit [d023c50](https://github.com/apache/flink/commit/d023c507a423b31f38203581c8eed27d28cb3a3c) get a [green result](https://myasuka.visualstudio.com/flink/_build/results?buildId=60&view=results).

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/AsyncCheckpointRunnables.java
```diff
@@ -0,0 +1,85 @@
...
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Track the {@link AsyncCheckpointRunnable}s in {@link SubtaskCheckpointCoordinator}.
+ */
+@Internal
+public class AsyncCheckpointRunnables implements Closeable {
```
rkhachatryan, line 36: I doubt whether we need this class, given that:
1. the only user is `SubtaskCheckpointCoordinatorImpl`
2. it's lifecycle is bound to one of `SubtaskCheckpointCoordinatorImpl`
3. the only field is a map (after removing the lock)

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SubtaskCheckpointCoordinatorImpl.java
```diff
@@ -126,6 +162,15 @@ public ChannelStateWriter getChannelStateWriter() {
 		return channelStateWriter;
 	}
 
+	@VisibleForTesting
+	int getAbortedCheckpointSize() {
+		return abortedCheckpointIds.size();
+	}
+
+	private boolean checkpointAlreadyAborted(long checkpointId) {
```
rkhachatryan, line 170: ~~nit~~ ambiguous name (not `nit` because I indeed suggested a wrong name ))

`getAndClearAbortedStatus`?

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SubtaskCheckpointCoordinatorImpl.java
```diff
@@ -142,6 +187,13 @@ public void checkpointState(
 		// We generally try to emit the checkpoint barrier as soon as possible to not affect downstream
 		// checkpoint alignments
 
+		// Step (0): Record the last triggered checkpointId.
+		lastCheckpointId = metadata.getCheckpointId();
```
rkhachatryan, line 191: `lastCheckpointId = Math.max(lastCheckpointId, metadata.getCheckpointId());`
?

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SubtaskCheckpointCoordinatorImpl.java
```diff
@@ -328,6 +421,7 @@ private OperatorSnapshotFutures buildOperatorSnapshotFutures(
 
 	@Override
 	public void close() throws IOException {
+		asyncCheckpointRunnables.close();
```
rkhachatryan, line 424: `try/catch` (or guava Closer or some other way to close both no matter what?)

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SubtaskCheckpointCoordinatorImpl.java
```diff
@@ -384,4 +478,12 @@ private static OperatorSnapshotFutures checkpointStreamOperator(
 			throw ex;
 		}
 	}
+
+	private static class NotifiedCheckpointAbortedException extends Exception {
```
rkhachatryan, line 482: Couldn't find any usages (remove?).

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SubtaskCheckpointCoordinatorImpl.java
```diff
@@ -126,6 +162,15 @@ public ChannelStateWriter getChannelStateWriter() {
 		return channelStateWriter;
 	}
 
+	@VisibleForTesting
+	int getAbortedCheckpointSize() {
+		return abortedCheckpointIds.size();
+	}
+
+	private boolean checkpointAlreadyAborted(long checkpointId) {
+		return abortedCheckpointIds.remove(checkpointId);
```
rkhachatryan, line 171: This is `O(n)`. 
I [suggested](https://github.com/apache/flink/pull/8693#discussion_r425881776) to use `LinkedHashMap`.

flink-state-backends/flink-statebackend-rocksdb/src/main/java/org/apache/flink/contrib/streaming/state/snapshot/RocksIncrementalSnapshotStrategy.java
```diff
@@ -179,6 +179,13 @@ public void notifyCheckpointComplete(long completedCheckpointId) {
 		}
 	}
 
+	@Override
+	public void notifyCheckpointAborted(long abortedCheckpointId) throws Exception {
+		synchronized (materializedSstFiles) {
+			materializedSstFiles.keySet().removeIf(checkpointId -> checkpointId == abortedCheckpointId);
```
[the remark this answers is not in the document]
Myasuka: Good catch

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/AsyncCheckpointRunnables.java
```diff
@@ -0,0 +1,85 @@
...
+	private final Map<Long, AsyncCheckpointRunnable> checkpoints;
+	/** Lock that guards state of this registry. **/
+	private final Object lock;
+
+	public AsyncCheckpointRunnables() {
```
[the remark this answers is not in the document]
Myasuka: I think this is something related to code style and no such guide in Flink official code-style guide. I prefer explicit constructor.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SubtaskCheckpointCoordinatorImpl.java
Myasuka: I think `getAndClearAbortedStatus` did not give explicit description of the result when checking the notified checkpoint-id. I think `isCheckpointAlreadyAborted` might be better than original name. 

Moreover, I think we don't need to explain that we will clear the aborted checkpoint-id if found which is too technically.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SubtaskCheckpointCoordinatorImpl.java
Myasuka: `checkpointState` is executed in sync phase of checkpoint on task which should behave as a strictly increasing order due to the barrier align mechanism.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SubtaskCheckpointCoordinatorImpl.java
Myasuka: good catch, this is something implemented in previous commits but already useless.

flink-state-backends/flink-statebackend-rocksdb/src/main/java/org/apache/flink/contrib/streaming/state/snapshot/RocksIncrementalSnapshotStrategy.java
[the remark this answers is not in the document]
Myasuka: Moreover, I'll implement `RocksDBKeyedStateBackend#notifyCheckpointAborted`

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SubtaskCheckpointCoordinatorImpl.java
Myasuka: Thanks for your remind.
I'll use a set based on `LinkedHashMap` to implement a set which would remove oldest checkpoint id automatically.
~~~java
private Set<Long> createAbortedCheckpointSetWithLimitSize(int maxRecordAbortedCheckpoints) {
	return Collections.newSetFromMap(new LinkedHashMap<Long, Boolean>() {

		@Override
		protected boolean removeEldestEntry(Map.Entry<Long, Boolean> eldest) {
			return size() > maxRecordAbortedCheckpoints;
		}
	});
}
~~~

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/AsyncCheckpointRunnables.java
```diff
@@ -0,0 +1,85 @@
...
+		this.checkpoints = new HashMap<>();
+		this.lock = new Object();
+	}
+
+	public void registerAsyncCheckpointRunnable(long checkpointId, AsyncCheckpointRunnable asyncCheckpointRunnable) {
+		synchronized (lock) {
+			checkpoints.put(checkpointId, asyncCheckpointRunnable);
```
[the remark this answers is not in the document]
Myasuka: > Can we add a check that we don't already have this checkpointId?
The checkpoint ids should be unique on task side, and I think adding a check here could be safer, if the `checkpointId` already registered, we should throw exception out.

> Ideally, we should also check that this registry was not closed yet
I think this is reasonable, which is similar to `AbstractCloseableRegistry`.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/AsyncCheckpointRunnables.java
```diff
@@ -0,0 +1,85 @@
...
+import java.util.Map;
+
+/**
+ * Track the {@link AsyncCheckpointRunnable}s in {@link SubtaskCheckpointCoordinator}.
+ */
+@Internal
+public class AsyncCheckpointRunnables implements Closeable {
+	private final Map<Long, AsyncCheckpointRunnable> checkpoints;
+	/** Lock that guards state of this registry. **/
+	private final Object lock;
```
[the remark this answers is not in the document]
Myasuka: If we add check to `closed`, we could still need the `lock` instead of introducing `ConcurrentMap`.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SubtaskCheckpointCoordinatorImpl.java
Myasuka: OK, I'll ensure both the closeables could be closed.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/AsyncCheckpointRunnables.java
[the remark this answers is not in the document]
rkhachatryan: But then this check can (should) be moved to `SubtaskCheckpointCoordinatorImpl`.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SubtaskCheckpointCoordinatorImpl.java
rkhachatryan: By convention, `isXxx` should be used for boolean getters.
I think it's always good to give self-explaining names to methods to prevent surprises and looking into their code.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SubtaskCheckpointCoordinatorImpl.java
rkhachatryan: It can also be triggered through RPC (for sources).
I agree that it shouldn't be possible now, but I think it's good to have such a guard or check to protect against bugs or changes.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/AsyncCheckpointRunnables.java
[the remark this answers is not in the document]
rkhachatryan: Yes, you're right.
And we have to add this check to maintain current behavior.
(both these fields however can be moved to `SubtaskCheckpointCoordinatorImpl` as I suggested [here](https://github.com/apache/flink/pull/8693#discussion_r427145436)).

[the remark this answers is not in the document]
Myasuka: I think your suggestion make sense, and the only other place using it is `LocalStateForwardingTest#testReportingFromSnapshotToTaskStateManager` ans I will pass a `SubtaskCheckpointCoordinatorImpl` to the created `AsyncCheckpointRunnable`.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SubtaskCheckpointCoordinatorImpl.java
Myasuka: How about `checkAndClearAbortedStatus` as I need to return boolean result.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SubtaskCheckpointCoordinatorImpl.java
Myasuka: If so, the condition below have to change which lead to I have to record a lot of unnecessary checkpoint ids.

~~~java
if (checkpointId > lastCheckpointId) {
	// only record checkpoints that have not triggered on task side.
       abortedCheckpointIds.add(checkpointId);
}
~~~

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SubtaskCheckpointCoordinatorImpl.java
rkhachatryan: :+1:

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SubtaskCheckpointCoordinatorImpl.java
rkhachatryan: We can just force the assumption with
```Preconditions.checkArgument(lastCheckpointId < metadata.getCheckpointId());```

carp84:
Already merged into master via:
* f0ed29c06d331892a06ee9bddea4173d6300835d
* fae6a6cad3f4fe30d80cc9dd664b2efc72c0be36
* 1f6b1e6b8ffb38f196c2a1c6348cb96935e9a9cc

release-1.11 via:
* fa731288518c8ebf66f40b4e0e9b1929546b6257
* fcacc42e17f00cb47c5c16fe75af035f784ae1fa
* 6591263c8871898f602d122aac336f2ad63bbb1c

PR #8695, opened 2019-06-11
[FLINK-12805][FLINK-12808][FLINK-12809][table-api] Introduce PartitionableTableSource and PartitionableTableSink and OverwritableTableSink

wuchong:

Introduce `PartitionableTableSource` for partition pruning, and introduce `OverwritableTableSink` for supporting insert overwrite, and introduce `PartitionableTableSink` for support writing data into partitions.

All of them are only supported in Blink planner in the first version.

## Brief change log

This pull request only contains the three interface. We didn't include any implementations because: 
 - `OverwritableTableSink` depends on the `INSERT OVERWRITE` syntax. 
 - `PartitionableTableSink` depends on the `INSERT PARTITION` syntax.
 - `PartitionableTableSource` depends on the filter pushdown optimization implementation in Blink planner.

## Verifying this change

N/A

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-table/flink-table-common/src/main/java/org/apache/flink/table/sinks/PartitionableTableSink.java
+++ b/flink-table/flink-table-common/src/main/java/org/apache/flink/table/sinks/PartitionableTableSink.java
@@ -1,0 +1,112 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.table.sinks;
+
+import org.apache.flink.annotation.Experimental;
+
+import java.util.List;
+import java.util.Map;
+
+/**
+ * An interface for partitionable {@link TableSink}. A partitionable sink can writes
+ * query results to partitions.
+ *
+ * <p>Partition columns are defined via {@link #getPartitionFieldNames()} and the field names
+ * should be sorted in a strict order. And all the partition fields should exist in the
+ * {@link TableSink#getTableSchema()}.
+ *
+ * <p>For example, a partitioned table named {@code my_table} with a table schema
+ * {@code [a INT, b VARCHAR, c DOUBLE, dt VARCHAR, country VARCHAR]} is partitioned on columns
+ * {@code dt, country}. Then {@code dt} is the first partition column, and
+ * {@code country} is the secondary partition column.
+ *
+ * <p>We can insert data into table partitions using INSERT INTO PARTITION syntax, for example:
+ * <pre>
+ * <code>
+ *     INSERT INTO my_table PARTITION (dt='2019-06-20', country='bar') select a, b, c from my_view
+ * </code>
+ * </pre>
+ * When all the partition columns are set a value in PARTITION clause, it is inserting into a
+ * static partition. It will writes the query result into a static partition,
+ * i.e. {@code dt='2019-06-20', country='bar'}. The user specified static partitions will be told
+ * to the sink via {@link #setStaticPartition(Map)}.
+ *
+ * <p>The INSERT INTO PARTITION syntax also supports dynamic partition inserts.
+ * <pre>
+ * <code>
+ *     INSERT INTO my_table PARTITION (dt='2019-06-20') select a, b, c, country from another_view
+ * </code>
+ * </pre>
+ * When partial partition columns (prefix part of all partition columns) are set a value in
+ * PARTITION clause, it is writing the query result into a dynamic partition. In the above example,
+ * the static partition part is {@code dt='2019-06-20'} which will be told to the sink via
+ * {@link #setStaticPartition(Map)}. And the {@code country} is the dynamic partition which will be
+ * get from each record.
+ */
+@Experimental
+public interface PartitionableTableSink {
+
+	/**
+	 * Gets the partition field names of the table. The partition field names should be sorted in
+	 * a strict order, i.e. they have the order as specified in the PARTITION statement in DDL.
+	 * This should be an empty set if the table is not partitioned.
+	 *
+	 * <p>All the partition fields should exist in the {@link TableSink#getTableSchema()}.
+	 *
+	 * @return partition field names of the table, empty if the table is not partitioned.
+	 */
+	List<String> getPartitionFieldNames();
+
+	/**
+	 * Sets the static partition into the {@link TableSink}. The static partition may be partial
+	 * of all partition columns. See the class Javadoc for more details.
+	 *
+	 * <p>The static partition is represented as a {@code Map<String, String>} which maps from
+	 * partition field name to partition value. The partition values are all encoded as strings,
+	 * i.e. encoded using String.valueOf(...). For example, if we have a static partition
+	 * {@code f0=1024, f1="foo", f2="bar"}. f0 is an integer type, f1 and f2 are string types.
+	 * They will all be encoded as strings: "1024", "foo", "bar". And can be decoded to original
+	 * literals based on the field types.
+	 *
+	 * @param partitions user specified static partition
+	 */
+	void setStaticPartition(Map<String, String> partitions);
+
+	/**
+	 * If returns true, sink can trust all records will definitely be grouped by partition fields
+	 * before consumed by the {@link TableSink}, i.e. the sink will receive all elements of one
+	 * partition and then all elements of another partition, elements of different partitions
+	 * will not be mixed. For some sinks, this can be used to reduce number of the partition
+	 * writers to improve writing performance.
+	 *
+	 * <p>This method is used to configure the behavior of input whether to be grouped by partition,
+	 * if true, at the same time the sink should also configure itself, i.e. set an internal field
+	 * that changes the writing behavior (writing one partition at a time).
+	 *
+	 * @param supportsGrouping whether the execution mode supports grouping,
+	 *                            e.g. grouping (usually use sort to implement) is only supported
+	 *                            in batch mode, not supported in streaming mode.
+	 *
+	 * @return whether data need to be grouped by partition before consumed by the sink. Default is false.
+	 * If {@code supportsGrouping} is false, it should never return true (requires grouping), otherwise it will fail.
+	 */
+	default boolean configurePartitionGrouping(boolean supportsGrouping) {
+		return false;
+	}
+}
--- a/flink-table/flink-table-common/src/main/java/org/apache/flink/table/sources/PartitionableTableSource.java
+++ b/flink-table/flink-table-common/src/main/java/org/apache/flink/table/sources/PartitionableTableSource.java
@@ -1,0 +1,68 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.table.sources;
+
+import org.apache.flink.annotation.Experimental;
+import org.apache.flink.table.sinks.TableSink;
+
+import java.util.List;
+import java.util.Map;
+
+/**
+ * An interface for partitionable {@link TableSource}.
+ *
+ * <p>A {@link PartitionableTableSource} can exclude partitions from reading, which
+ * includes skipping the metadata. This is especially useful when there are thousands
+ * of partitions in a table.
+ *
+ * <p>A partition is represented as a {@code Map<String, String>} which maps from partition
+ * field name to partition value. Since the map is NOT ordered, the correct order of partition
+ * fields should be obtained via {@link #getPartitionFieldNames()}.
+ */
+@Experimental
+public interface PartitionableTableSource {
+
+	/**
+	 * Returns all the partitions of this {@link PartitionableTableSource}.
+	 */
+	List<Map<String, String>> getPartitions();
+
+	/**
+	 * Gets the partition field names of the table. The partition field names should be sorted in
+	 * a strict order, i.e. they have the order as specified in the PARTITION statement in DDL.
+	 * This should be an empty set if the table is not partitioned.
+	 *
+	 * <p>All the partition fields should exist in the {@link TableSink#getTableSchema()}.
+	 *
+	 * @return partition field names of the table, empty if the table is not partitioned.
+	 */
+	List<String> getPartitionFieldNames();
+
+	/**
+	 * Applies the remaining partitions to the table source. The {@code remainingPartitions} is
+	 * the remaining partitions of {@link #getPartitions()} after partition pruning applied.
+	 *
+	 * <p>After trying to apply partition pruning, we should return a new {@link TableSource}
+	 * instance which holds all pruned-partitions.
+	 *
+	 * @param remainingPartitions Remaining partitions after partition pruning applied.
+	 * @return A new cloned instance of {@link TableSource} holds all pruned-partitions.
+	 */
+	TableSource applyPartitionPruning(List<Map<String, String>> remainingPartitions);
+}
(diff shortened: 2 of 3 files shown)


KurtYoung:
split to 3 commits?

wuchong:
Sure @KurtYoung , I have split into 3 commits.

wuchong:
Hi @twalthr @aljoscha , do you have time to review the source sink interfaces for Hive functionality?

flink-table/flink-table-common/src/main/java/org/apache/flink/table/sinks/PartitionableTableSink.java
```diff
@@ -0,0 +1,71 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.flink.table.sinks;
+
+import java.util.List;
+import java.util.Map;
+
+/**
+ * An abstract class with trait about partitionable table sink. This is mainly used for
```
lirui-apache: Why is it mainly used for static partition?

wuchong: I'm not sure, this is the current design in Blink.  Do we support dynamic partition for sink in Blink?

lirui-apache: Yes we do. Actually the example in this very doc suggests it supports dynamic partitioning.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/sources/PartitionableTableSource.java
```diff
@@ -0,0 +1,76 @@
...
+
+package org.apache.flink.table.sources;
+
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Adds support for partition pruning to a [[TableSource]].
+ * A {@link TableSource} extending this interface is a partition table and able to
+ * prune partitions before reading data.
```
godfreyhe: another usage is to reduce reading metadata (especially there are thousands of files)

flink-table/flink-table-common/src/main/java/org/apache/flink/table/sources/PartitionableTableSource.java
```diff
@@ -0,0 +1,76 @@
...
+	 * it is recommended that we still return a new {@link TableSource} instance since we will
+	 * mark the returned instance as partition pruned and will never try again.
+	 *
+	 * @param remainingPartitions Remaining partitions after partition pruning applied.
+	 * @return A new cloned instance of {@link TableSource}.
+	 */
+	TableSource applyPartitionPruning(List<Map<String, String>> remainingPartitions);
+
+	/**
+	 * Whether drop partition predicates after apply partition pruning.
```
godfreyhe: could you explain it more clearly why we need this method ?

godfreyhe: if the values of partition fields in a partition are not unique, the partition predicates could not be dropped after partition pruning, otherwise the result is incorrect. 
e.g. a table source is partition by hour,  and its partition field is `timestamp` which precision is milliseconds. each partition contains multiple `timestamp` values. we also could use partition predicates to filter partitions, but the remaining partitions may contains unexpected result which could not meet the partition predicates. the partition predicates could be reserved.
however, how to filter partitions is very special too. 

I think we should delete this method and require that the values of partition fields in every partition should be unique. otherwise use should implement another partition prune rule to meet special needs.

wuchong: I agree. `supportDropPartitionPredicate()` has been removed.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/sinks/PartitionableTableSink.java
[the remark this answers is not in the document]
lirui-apache: I guess we can remove that statement, since we'll support dynamic partitioning in Flink.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/sinks/PartitionableTableSink.java
```diff
@@ -0,0 +1,71 @@
...
+ * static partitions. For sql statement:
+ * <pre>
+ * <code>
+ * INSERT INTO A PARTITION(a='ab', b='cd') select c, d from B
+ * </code>
+ * </pre>
+ * We Assume the A has partition columns as &lt;a&gt;, &lt;b&gt;, &lt;c&gt;.
+ * The columns &lt;a&gt; and &lt;b&gt; are called static partition columns, while c is called
```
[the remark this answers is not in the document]
lirui-apache: I think we should give the definition of table `A` if we intend to offer a valid example. It's true the dynamic column should appear last, but the column names in `SELECT` don't have to be the same as the column names in the destination table -- so it's hard to tell w/o a DDL :)

flink-table/flink-table-common/src/main/java/org/apache/flink/table/sinks/PartitionableTableSink.java
```diff
@@ -0,0 +1,71 @@
...
+	 * Get the partition keys of the table. This should be an empty set if the table is not partitioned.
+	 *
+	 * @return partition keys of the table
+	 */
+	List<String> getPartitionKeys();
+
+	/**
+	 * Sets the static partitions into the {@link TableSink}.
+	 * @param partitions mapping from static partition column names to string literal values.
+	 *                      String literals will be quoted using {@code '}, for example,
```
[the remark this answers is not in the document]
wuchong: This is a mistake. I think the key-value meaning should align with `CatalogPartitionSpec`. If I understand correctly, all the values are encoded as string (i.e. encoded using `String.valueOf(...)`). For example, there is a static partition `f0=1024, f1="foo", f2="bar"`.  `f0` is an integer type, `f1` and `f2` are string types. They will all be encoded as strings: "1024", "foo", "bar". And will be decoded to original type literals based on the field types. Do I understand correctly? @bowenli86 

I will update this to the javadoc.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/sources/PartitionableTableSource.java
```diff
@@ -0,0 +1,68 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.flink.table.sources;
+
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Adds support for partition pruning to a [[TableSource]].
```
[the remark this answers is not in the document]
wuchong: I do a little change: `A partition is represented as a Map<String, String>` instead of `Partitions are ...`. Because a `Map<String, String>` represents only one partition which is consisted of several field names -> static field values.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/sources/PartitionableTableSource.java
```diff
@@ -0,0 +1,68 @@
...
+	 *
+	 * <p>After trying to apply partition pruning, we should return a new {@link TableSource}
+	 * instance which holds all pruned-partitions. Even if we actually pruned nothing,
+	 * it is recommended that we still return a new {@link TableSource} instance since we will
+	 * mark the returned instance as partition pruned and will never try again.
+	 *
+	 * @param remainingPartitions Remaining partitions after partition pruning applied.
+	 * @return A new cloned instance of {@link TableSource}.
+	 */
+	TableSource applyPartitionPruning(List<Map<String, String>> remainingPartitions);
```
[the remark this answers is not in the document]
godfreyhe: yes, `isPartitionPruned` could be deleted. the partition predicates should be drop from `Filter` while appling  push partition into table source rule. the `Filter` will be only matched once, and there is no infinite loop.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/sinks/PartitionableTableSink.java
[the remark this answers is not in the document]
wuchong: As partition fields are part of the schema, the table sink doesn't need to know whether it is a static partition or dynamic partition, and even doesn't need to know the specific static partition, it can be get from records. So I removed the comments about dynamic partition and `setStaticPartitions(...)` method.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/sinks/PartitionableTableSink.java
```diff
@@ -0,0 +1,71 @@
...
+	 * one partition at a time.
+	 *
+	 * <p>A sink should consider whether to override this especially when it needs buffer
+	 * data before writing.
+	 *
+	 * <p>Notes:
+	 * 1. If returns true, the output data will be sorted <strong>locally</strong> after partitioning.
+	 * 2. Default returns true, if the table is partitioned.
+	 */
+	default boolean sortLocalPartition() {
```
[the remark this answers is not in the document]
wuchong: I think it's a requirement, i.e. data must to be sorted by partition if returns true. However, when it is in streaming mode, sort is not supported. So I extended this method a bit to pass a `supportSort` parameter in.

I changed the method name to `requiresSortByPartition` instead of `requiresPartitionGrouping` , because it is indeed sorting by partitions. However, grouping by partitions may confuse users it is keyBy partitions.

wuchong:
Hi @aljoscha @godfreyhe , I have updated this PR with following changes:

Changes to `PartitionableTableSink`:
1. Rename `getPartitionKeys` to `getPartitionFieldNames`
2. Remove `setStaticPartitions(Map<String, String> partitions)` method
 - Because partition fields are part of the schema, table sink doesn't need to know the specific static partition, it can be get from records.
3. Change `boolean sortLocalPartition()` method to `boolean requiresSortByPartition(boolean supportSort)`

Changes to `PartitionableTableSource`:
1. Rename `getAllPartitions` to `getPartitions`
2. Rename `getPartitionKeys` to `getPartitionFieldNames`
3. Remove `boolean isPartitionPruned()` method

lirui-apache:
> Because partition fields are part of the schema, table sink doesn't need to know the specific static partition, it can be get from records.

Table sink can tell which columns are partition columns in a record. But without `setStaticPartitions `, how can it know which partition columns are static and which are dynamic?

flink-table/flink-table-common/src/main/java/org/apache/flink/table/sinks/PartitionableTableSink.java
```diff
@@ -0,0 +1,67 @@
...
+ * An interface for partitionable {@link TableSink TableSinks}. A partitionable sink can writes
+ * query results to partitions.
+ *
+ * <p>Partition columns are defined via {@link #getPartitionFieldNames()} and the field names
+ * should be sorted in a strict order. And all the partition fields should exist in the
+ * {@link TableSink#getTableSchema()}.
+ *
+ * <p>For example, a partitioned table named {@code my_table} with a table schema
+ * {@code [a INT, b VARCHAR, c DOUBLE, dt VARCHAR, country VARCHAR]} is partitioned on columns
+ * {@code ds, country}. Then {@code ds} is the first partition column, and
```
bowenli86: there's a mismatch between the column names of `ds` and the`dt` above

flink-table/flink-table-common/src/main/java/org/apache/flink/table/sinks/PartitionableTableSink.java
```diff
@@ -0,0 +1,67 @@
...
+	 * <p>All the partition fields should exist in the {@link TableSink#getTableSchema()}.
+	 *
+	 * @return partition field names of the table, empty if the table is not partitioned.
+	 */
+	List<String> getPartitionFieldNames();
+
+	/**
+	 * If returns true, sink can trust all records will definitely be sorted by partition fields
+	 * before consumed by the {@link TableSink}, i.e. the sink will receive data one partition
+	 * at a time. For some sinks, this can be used to reduce the partition writers to improve
```
bowenli86: reduce "what" of the partition writers? number, workload?

flink-table/flink-table-common/src/main/java/org/apache/flink/table/sinks/PartitionableTableSink.java
```diff
@@ -0,0 +1,67 @@
...
+	 * writing performance.
+	 *
+	 * @param supportSort whether the execution mode supports sort, e.g. sort is only supported in
+	 *                    batch mode, not supported in stream mode.
```
bowenli86:
```suggestion
	 *                    batch mode, not supported in streaming mode.
```

flink-table/flink-table-common/src/main/java/org/apache/flink/table/sources/PartitionableTableSource.java
```diff
@@ -0,0 +1,68 @@
...
+package org.apache.flink.table.sources;
+
+import org.apache.flink.annotation.Experimental;
+import org.apache.flink.table.sinks.TableSink;
+
+import java.util.List;
+import java.util.Map;
+
+/**
+ * An interface for partitionable {@link TableSource TableSources}.
```
bowenli86:
```suggestion
 * An interface for partitionable {@link TableSource}.
```

flink-table/flink-table-common/src/main/java/org/apache/flink/table/sinks/PartitionableTableSink.java
wuchong: reduce number of writers

wuchong:
After discussed with @lirui-apache offline, I think it make sense to let sink know what's the static partition part. Then the sink will have some optimization when it's not a dynamic partition inserting. So I add back the `setStaticPartition`. Could you have a look again @lirui-apache ?

lirui-apache:
Thanks @wuchong for updating. LGTM now.

godfreyhe:
LGTM

flink-table/flink-table-common/src/main/java/org/apache/flink/table/sinks/PartitionableTableSink.java
bowenli86: can you update the javadoc?

wuchong: Yeah, I have updated the javadoc.

wuchong:
Hi @aljoscha , do you have time to have a look again?

aljoscha:
I'll have another look 👌

flink-table/flink-table-common/src/main/java/org/apache/flink/table/sinks/PartitionableTableSink.java
```diff
@@ -0,0 +1,106 @@
...
+	 */
+	List<String> getPartitionFieldNames();
+
+	/**
+	 * Sets the static partition into the {@link TableSink}. The static partition may be partial
+	 * of all partition columns. See the class Javadoc for more details.
+	 *
+	 * <p>The static partition is represented as a {@code Map<String, String>} which maps from
+	 * partition field name to partition value. The partition values are all encoded as strings,
+	 * i.e. encoded using String.valueOf(...). For example, there is a static partition
```
aljoscha:
```suggestion
	 * i.e. encoded using String.valueOf(...). For example, if we have static partitions
```

flink-table/flink-table-common/src/main/java/org/apache/flink/table/sinks/PartitionableTableSink.java
```diff
@@ -0,0 +1,106 @@
...
+	 * {@code f0=1024, f1="foo", f2="bar"}. f0 is an integer type, f1 and f2 are string types.
+	 * They will all be encoded as strings: "1024", "foo", "bar". And can be decoded to original
+	 * literals based on the field types.
+	 *
+	 * @param partitions user specified static partition
+	 */
+	void setStaticPartition(Map<String, String> partitions);
```
aljoscha:
```suggestion
	void setStaticPartitions(Map<String, String> partitions);
```

flink-table/flink-table-common/src/main/java/org/apache/flink/table/sinks/PartitionableTableSink.java
```diff
@@ -0,0 +1,106 @@
...
+	 * of all partition columns. See the class Javadoc for more details.
+	 *
+	 * <p>The static partition is represented as a {@code Map<String, String>} which maps from
+	 * partition field name to partition value. The partition values are all encoded as strings,
+	 * i.e. encoded using String.valueOf(...). For example, there is a static partition
+	 * {@code f0=1024, f1="foo", f2="bar"}. f0 is an integer type, f1 and f2 are string types.
+	 * They will all be encoded as strings: "1024", "foo", "bar". And can be decoded to original
+	 * literals based on the field types.
+	 *
+	 * @param partitions user specified static partition
```
aljoscha:
```suggestion
	 * @param partitions user specified static partitions
```

flink-table/flink-table-common/src/main/java/org/apache/flink/table/sinks/PartitionableTableSink.java
```diff
@@ -0,0 +1,106 @@
...
+	 * at a time. For some sinks, this can be used to reduce number of the partition writers
+	 * to improve writing performance.
+	 *
+	 * @param supportSort whether the execution mode supports sort, e.g. sort is only supported in
+	 *                    batch mode, not supported in streaming mode.
+	 *
+	 * @return whether data need to be sorted by partition before consumed by the sink. Default is false.
+	 * If {@code supportSort} is false, it should never return true (requires sort), otherwise it will fail.
+	 */
+	default boolean requiresSortByPartition(boolean supportSort) {
```
aljoscha: I think this method is somewhat tricky because it both asks the source about what it wants and at the same time the source will configure itself. I.e, a source that returns `true` will set an internal field that changes its behaviour. I didn't consider that second aspect before, that's why I'm only noticing now. So maybe we should call it `configurePartitionGrouping(boolean)` and explicitly mention this also in the Javadoc?

wuchong:
Merged with 

- 6e4805673739fa1f8e85389d38ef6f3de20fc005
- fb111ef22b8f87dd25e61160678fc8857dffcfe0
- c3f2ad8c8c10097dc7f680ccdf2e9b99babc349c

PR #8697, opened 2019-06-12
[FLINK-12810][table-planner-blink] Support to run a TableAPI query like 'table.select('a, 'b, 'c)'

beyond1920:
The pr aims to support to run a TableAPI query like 'table.select('a, 'b, 'c)'

## Brief change log
1. add RexNodeConverter to convert Expression to RexNode
2. Introduce RexPlannerExpression to wraps a RexNode, which is similar to RexPlannerExpression in flink-planner module.
3. add QueryOperationConverter to convert QueryOperation tree to RelNode
4. Introduce PlannerQueryOperation to wraps a RelNode in QueryOperation, which is similar to PlannerQueryOperation in flink-planner module.
5. add OperationTreeBuilder and relative classes(ExpressionResolver, TableReferenceLookup, FieldReferenceLoopup, ExpandColumnFunctionsRule, LookupCallByNameRule, ResolverRule, StarReferenceFlattingRule, VerifyNoUnresolvedExpressionRule), which is almost copy from flink-planner module to blink-planner module, except that we remove dependency on PlannerExpression. *Note: This is temporal solution, since OperationTreeBuilder will be moved to api module finally.*
6. add implicit conversion in expressionDsl

## Verifying this change
ITCase

  - The serializers: (no)

## Documentation


tools/maven/suppressions.xml
```diff
@@ -19,43 +19,49 @@ under the License.
 -->
 
 <!DOCTYPE suppressions PUBLIC
-		"-//Puppy Crawl//DTD Suppressions 1.1//EN"
```
JingsongLi: Why change this?

aljoscha:
At the very least, this would have to be split up into several commits. And, as I said in the Jira issue, I'm skeptical about copying all those classes over to the Blink runner.

beyond1920:
@aljoscha Thanks for your review, I will split the issue later. About copy OperationTreeBuilder to Blink planner, as I response in jira, it is a temporal solution to run all API testcases as soon as possible.

PR #8700, opened 2019-06-12
[FLINK-12657][hive] Integrate Flink with Hive UDF

bowenli86:

This PR integrates Flink with Hive UDF.

Background: given that integrating Flink with Hive functions has some hard dependencies on function definitions and type system rework part II and is currently blocked by these dependencies, we decide to take an eager approach by merging Hive functions related code into Flink first, and then adapt to any dependencies when they are done. This way, we are more confident with the timing, and minimize our risk of not being able to make it to 1.9 release.

## Brief change log

- introduced `HiveFunction` as a temp workaround to get input/output arg types
- added `HiveFunctionWrapper` as wrapper of Hive functions
- developed `HiveSimpleUDF` to hook Hive UDF with Flink
- added `HiveInspector` for data inspection and conversions
- added unit tests for `HiveSimpleUDF` using a few real Hive UDFs

Note that the data conversion currently only covers a few basic types. Conversions of more typed data will be added later on.

## Verifying this change

- added `HiveSimpleUDFTest`

  - The serializers:(yes)
  - The S3 file system connector: (yes)

## Documentation

Doc will be added later

### DIFF
--- a/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/functions/hive/HiveFunction.java
+++ b/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/functions/hive/HiveFunction.java
@@ -1,0 +1,62 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.table.functions.hive;
+
+import org.apache.flink.annotation.Internal;
+import org.apache.flink.table.types.DataType;
+
+/**
+ * Interface for Hive simple udf, generic udf, and generic udtf.
+ * TODO: Note: this is only a temporary interface for workaround when Flink type system and udf system
+ * 	rework is not finished. Should adapt to Flink type system and Flink UDF framework later on.
+ */
+@Internal
+public interface HiveFunction {
+
+	/**
+	 * Set arguments and argTypes for Function instance.
+	 * In this way, the correct method can be really deduced by the function instance.
+	 *
+	 * <p>Framework should guarantee that this is always called
+	 * before {@link #getHiveResultType(Object[] constantArguments, DataType[] argTypes)}.
+	 *
+	 * @param constantArguments arguments of a function call (only literal arguments
+	 *                  are passed, nulls for non-literal ones)
+	 * @param argTypes types of arguments
+	 */
+	void setArgumentTypesAndConstants(Object[] constantArguments, DataType[] argTypes);
+
+	/**
+	 * Get result type by arguments and argTypes.
+	 *
+	 * <p>We can't use getResultType(Object[], Class[]). The Class[] is the classes of what
+	 * is defined in eval(), for example, if eval() is "public Integer eval(Double)", the
+	 * argTypes would be Class[Double]. However, in our wrapper, the signature of eval() is
+	 * "public Object eval(Object... args)", which means we cannot get any info from the interface.
+	 *
+	 * <p>Framework should guarantee that this is always called
+	 * 	 after {@link #setArgumentTypesAndConstants(Object[] constantArguments, DataType[] argTypes)}.
+	 *
+	 * @param constantArguments arguments of a function call (only literal arguments
+	 *                  are passed, nulls for non-literal ones)
+	 * @param argTypes types of arguments
+	 * @return result type.
+	 */
+	DataType getHiveResultType(Object[] constantArguments, DataType[] argTypes);
+}
--- a/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/functions/hive/HiveFunction.java
+++ b/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/functions/hive/HiveFunction.java
@@ -33,9 +33,6 @@ public interface HiveFunction {
 	 * Set arguments and argTypes for Function instance.
 	 * In this way, the correct method can be really deduced by the function instance.
 	 *
-	 * <p>Framework should guarantee that this is always called
-	 * before {@link #getHiveResultType(Object[] constantArguments, DataType[] argTypes)}.
-	 *
 	 * @param constantArguments arguments of a function call (only literal arguments
 	 *                  are passed, nulls for non-literal ones)
 	 * @param argTypes types of arguments
@@ -49,9 +46,6 @@ public interface HiveFunction {
 	 * is defined in eval(), for example, if eval() is "public Integer eval(Double)", the
 	 * argTypes would be Class[Double]. However, in our wrapper, the signature of eval() is
 	 * "public Object eval(Object... args)", which means we cannot get any info from the interface.
-	 *
-	 * <p>Framework should guarantee that this is always called
-	 * 	 after {@link #setArgumentTypesAndConstants(Object[] constantArguments, DataType[] argTypes)}.
 	 *
 	 * @param constantArguments arguments of a function call (only literal arguments
 	 *                  are passed, nulls for non-literal ones)
--- a/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/functions/hive/HiveScalarFunction.java
+++ b/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/functions/hive/HiveScalarFunction.java
@@ -24,6 +24,7 @@ import org.apache.flink.table.functions.FunctionContext;
 import org.apache.flink.table.functions.ScalarFunction;
 import org.apache.flink.table.types.CollectionDataType;
 import org.apache.flink.table.types.DataType;
+import org.apache.flink.table.types.logical.LogicalTypeRoot;
 import org.apache.flink.table.types.utils.LegacyTypeInfoDataTypeConverter;
 
 import org.apache.hadoop.hive.ql.exec.UDF;
@@ -79,7 +80,8 @@ public abstract class HiveScalarFunction<UDFType> extends ScalarFunction impleme
 	@Override
 	public void open(FunctionContext context) {
 		openInternal();
-		isArgsSingleArray = argTypes.length == 1 && (argTypes[0] instanceof CollectionDataType);
+		isArgsSingleArray = argTypes.length == 1 &&
+			(argTypes[0] instanceof CollectionDataType && argTypes[0].getLogicalType().getTypeRoot().equals(LogicalTypeRoot.ARRAY));
 	}
 
 	/**
--- a/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/functions/hive/HiveScalarFunction.java
+++ b/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/functions/hive/HiveScalarFunction.java
@@ -22,15 +22,15 @@ import org.apache.flink.annotation.Internal;
 import org.apache.flink.api.common.typeinfo.TypeInformation;
 import org.apache.flink.table.functions.FunctionContext;
 import org.apache.flink.table.functions.ScalarFunction;
-import org.apache.flink.table.types.CollectionDataType;
 import org.apache.flink.table.types.DataType;
+import org.apache.flink.table.types.logical.ArrayType;
+import org.apache.flink.table.types.logical.LogicalType;
 import org.apache.flink.table.types.logical.LogicalTypeRoot;
 import org.apache.flink.table.types.utils.LegacyTypeInfoDataTypeConverter;
 
 import org.apache.hadoop.hive.ql.exec.UDF;
 import org.apache.hadoop.hive.ql.udf.generic.GenericUDF;
 import org.apache.hadoop.hive.serde2.objectinspector.ObjectInspector;
-
 
 /**
  * Abstract class to provide more information for Hive {@link UDF} and {@link GenericUDF} functions.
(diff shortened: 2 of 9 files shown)


KurtYoung:
@flinkbot attention @JingsongLi

bowenli86:
cc @xuefuz @JingsongLi @lirui-apache @zjuwangg

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/functions/hive/HiveFunction.java
```diff
@@ -0,0 +1,62 @@
...
+	 *
+	 * <p>Framework should guarantee that this is always called
+	 * 	 after {@link #setArgumentTypesAndConstants(Object[] constantArguments, DataType[] argTypes)}.
+	 *
+	 * @param constantArguments arguments of a function call (only literal arguments
+	 *                  are passed, nulls for non-literal ones)
+	 * @param argTypes types of arguments
+	 * @return result type.
+	 */
+	DataType getHiveResultType(Object[] constantArguments, DataType[] argTypes);
```
xuefuz: Just curious, if setArgumentTypesAndConstants() is called before this one is called with the same arguments, why do we need those arguments again?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/functions/hive/HiveScalarFunction.java
```diff
@@ -0,0 +1,107 @@
...
+ */
+@Internal
+public abstract class HiveScalarFunction<UDFType> extends ScalarFunction implements HiveFunction {
+
+	protected final HiveFunctionWrapper<UDFType> hiveFunctionWrapper;
+
+	protected Object[] constantArguments;
+	protected DataType[] argTypes;
+
+	protected transient UDFType function;
```
xuefuz: hiveFunctionWrapper contains such an instance, right? Why we need this here?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/functions/hive/HiveScalarFunction.java
```diff
@@ -0,0 +1,107 @@
...
+					.getAnnotation(org.apache.hadoop.hive.ql.udf.UDFType.class);
+
+			return udfType != null && udfType.deterministic() && !udfType.stateful();
+		} catch (ClassNotFoundException e) {
+			throw new FlinkHiveUDFException(e);
+		}
+	}
+
+	@Override
+	public TypeInformation getResultType(Class[] signature) {
```
xuefuz: How come we don't implemment HiveFunction.getResultType()?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/functions/hive/HiveSimpleUDF.java
```diff
@@ -0,0 +1,135 @@
...
+			throw new FlinkHiveUDFException(
+				String.format("Failed to open HiveSimpleUDF from %s", hiveFunctionWrapper.getClassName()), e);
+		}
+	}
+
+	@Override
+	public Object evalInternal(Object[] args) {
+		checkArgument(args.length == conversions.length);
+
+		if (!allIdentityConverter) {
```
xuefuz: How do we gain from this optimization?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/functions/hive/HiveFunction.java
bowenli86: Good catch, the javadoc is a bit misleading. The fact is that `getHiveResultType` can be called independently thru its parent method `getResultType()` without calling `setArgumentTypesAndConstants()` in Blink. Not sure about if that will be the case in Flink. 

I will remove the wrong part of javadoc

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/functions/hive/HiveSimpleUDF.java
bowenli86: the flag indicates if all input args can stay the same and don't need to be converted to a Hive object. So if none of them needs to be converted, we can skip the conversion and directly invoke Hive UDF with them.

We can optimize more, if necessary later on, by remembering index of input args that need to be converted and only convert a subset of input args. Currently I'll just make it the same as that in Blink

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/functions/hive/conversion/HiveInspectors.java
```diff
@@ -0,0 +1,248 @@
...
+			}
+		}
+
+		// TODO: handle complex types like struct, list, and map
+
+		throw new FlinkHiveUDFException(
+			String.format("Unwrap does not support ObjectInspector '%s' yet", inspector));
+	}
+
+	public static ObjectInspector getObjectInspector(Class clazz) {
```
xuefuz: Is this method used anywhere?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/functions/hive/HiveScalarFunction.java
bowenli86: @xuefuz sorry that I accidentally deleted your comment here. So your question is what's the difference between this function instance and the one in `HiveFunctionWrapper`.

The one in `HiveFunctionWrapper` acts as a cache for GenericUDF, GenericUDTA, and UDAFResolver, but not `UDF` (I'm still not sure why not `UDF`, that logic is borrowed from Spark). So this function instance in `HiveScalarFunction` is actually the one gonna be invoked to process data.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/functions/hive/conversion/HiveInspectors.java
bowenli86: it's used in `HiveSimpleUDF#getHiveResultType()`

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/functions/hive/HiveFunction.java
xuefuz: OK. Let's see.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/functions/hive/HiveScalarFunction.java
xuefuz: okay.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/functions/hive/HiveSimpleUDF.java
xuefuz: Yeah, but I was wondering how or how much it saves and whether it's worth the effort. It requires all be identityconverter, which is quite limiting in my opinion. Looking at IdentityConverter, the conversion just returns the original input, which is very light.

I'm trying to see if I have missed anything.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/functions/hive/conversion/HiveInspectors.java
xuefuz: ok

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/functions/hive/HiveSimpleUDF.java
bowenli86: just by the logic itself, I think it can be faster for complex types and nested types, I don't have benchmarks and users' usage pattern.

I don't think it's critically important, either removing or keeping it. This is more of referencing how we did it in Blink rather than intentionally trying to optimize it hard. If you feel really strong on removing it, we can surely do that

xuefuz: It's okay. I was just to see if I missed anything obvious.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/functions/hive/HiveScalarFunction.java
JingsongLi: I think the reason is in Hive `UDF` comments:
```
One instance of this class will be instantiated per JVM and it will not be called concurrently.
```

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/functions/hive/HiveScalarFunction.java
```diff
@@ -0,0 +1,107 @@
...
+		return LegacyTypeInfoDataTypeConverter.toLegacyTypeInfo(
+			getHiveResultType(this.constantArguments, this.argTypes));
+	}
+
+	@Override
+	public void open(FunctionContext context) {
+		openInternal();
+		isArgsSingleArray = argTypes.length == 1 && (argTypes[0] instanceof CollectionDataType);
```
JingsongLi: Only Object[] is called `isArgsSingleArray`.
Not include primitiveArray, multiset.
I think we should make more judgments and add tests.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/functions/hive/HiveSimpleUDF.java
JingsongLi: If there are more than one parameter, this will lead to multiple virtual function calls, and the cost of virtual function calls is not small. This optimization is so simple that I think it's worth doing.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/functions/hive/conversion/HiveInspectors.java
```diff
@@ -0,0 +1,248 @@
...
+		}
+
+		throw new FlinkHiveUDFException(
+			String.format("Flink doesn't support convert object conversion for %s yet", inspector));
+	}
+
+	/**
+	 * Unwraps an object from Hive object to Flink object with an ObjectInspector.
+	 */
+	public static Object unwrap(ObjectInspector inspector, Object data) {
```
JingsongLi: Why not merge `unwrap` and `HiveObjectConversion`?
I think `HiveObjectConversion` should provide two method: `toHiveObject` and `toFlinkObject`.
Just like blink `DataFormatConverters`.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/functions/hive/conversion/HiveInspectors.java
```diff
@@ -0,0 +1,248 @@
...
+			} else if (inspector instanceof DoubleObjectInspector) {
+				DoubleObjectInspector oi = (DoubleObjectInspector) inspector;
+
+				return oi.preferWritable() ?
+					oi.get(data) :
+					oi.getPrimitiveWritableObject(data);
+			}
+		}
+
+		// TODO: handle complex types like struct, list, and map
```
JingsongLi: May not only complex types, for example, spark handle:
```
    case _: WritableStringObjectInspector => StringType
    case _: JavaStringObjectInspector => StringType
    case _: WritableHiveVarcharObjectInspector => StringType
    case _: JavaHiveVarcharObjectInspector => StringType
    case _: WritableHiveCharObjectInspector => StringType
    case _: JavaHiveCharObjectInspector => StringType
```

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/functions/hive/HiveScalarFunction.java
bowenli86: Good catch on the multiset thing!  What do you mean by the primitive array case? 

BTW, the point here is that `HiveFunction` and its APIs are *temporary*. We don't know how and in which form the types are gonna be passed in yet, and this part should be refactored later to adapt to any outcome of the type system rework. So I think it's ok to not over-engineer the testing part for now.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/functions/hive/conversion/HiveInspectors.java
bowenli86: IIUC, the example you give is in `inspectorToDataType()` and it's used to get a Spark data type from a Hive inspector, and shouldn't that match our `toFlinkType()` rather than our `unwrap()`?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/functions/hive/conversion/HiveInspectors.java
bowenli86: that's might be doable. I need to spend some time to take a look at it because, unlike DataFormatConverters,  there's a subtle difference between Flink -> Hive and Hive -> Flink: 

From Flink-> Hive, it will always be thru `JavaXxxObjectInspector`. But from Hive -> Flink, it can be thru either `WritableXxxObjectInspector` or `JavaXxxObjectInspector`

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/functions/hive/HiveScalarFunction.java
JingsongLi: For example, int[] pass to eval(Object...) is ok, but Integer[] pass to eval(Object...) is wrong.
So we need set `isArgsSingleArray` when Integer[] instead of int[]. (int[] is primitive array)

IMO, this is a wrong code, I don't think we should merge it.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/functions/hive/conversion/HiveInspectors.java
JingsongLi:
```
      case _: JavaHiveVarcharObjectInspector =>
        withNullSafe { o =>
            val s = o.asInstanceOf[UTF8String].toString
            new HiveVarchar(s, s.length)
        }
      case _: JavaHiveCharObjectInspector =>
        withNullSafe { o =>
            val s = o.asInstanceOf[UTF8String].toString
            new HiveChar(s, s.length)
          }
```
wrap also have these.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/functions/hive/conversion/HiveInspectors.java
JingsongLi: Yeah, I get it.
But I recommend aligning `getConversion` and `unwrap`, calling them similar names, and returning classes (rather than direct conversions).
It doesn't seem to take much effort.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/functions/hive/conversion/HiveInspectors.java
bowenli86: I see what you mean now. Primitive types like char, varchar, timestamp, date, decimal, etc will added immediately afterwards

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/functions/hive/HiveScalarFunction.java
bowenli86: >  int[] pass to eval(Object...) is ok, but Integer[] pass to eval(Object...) is wrong. So we need set isArgsSingleArray when Integer[] instead of int[]

sorry, I don't really understand what you are trying to say....  Let's sync offline

bowenli86: Synced with @JingsongLi offline, and he has updated his comment.

The root cause is that, different from Blink, Flink supports nullable field and therefore both primitive array and boxed array now. If an array of int column is not nullable, the type will be int[]; if it's nullable, the type will be Integer[].

I need to take another look at this issue and what it implies.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/functions/hive/conversion/HiveInspectors.java
bowenli86: Giving it a re-evaluation, I think it's better not to do it now as I can't see much value out of it. The main reason you choose the `HiveObjectConversion` way is to optimize conversions of mult input args, but it doesn't apply to output arg as it's always only one. The current structure of "unwrap" is the same as that in Spark, making it easy for us to compare and question on. 

I will rename it to "toFlinkObject(...)" to make the names more aligned

bowenli86:
I created FLINK-12854 to evaluate if we should support primitive types and convert them for Hive functions. This is not critical and thus may or may not happen in 1.9.

@JingsongLi Can you take another look?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/functions/hive/conversion/HiveInspectors.java
JingsongLi: > The current structure of "unwrap" is the same as that in Spark, making it easy for us to compare and question on.

In fact, Spark unwrap return a Function instead of runtime `instanceof`.

JingsongLi:
+1 to merge

bowenli86:
@JingsongLi @xuefuz thanks for your review! 

Merging

PR #8704, opened 2019-06-12
[FLINK-12812][runtime] Set resource profiles for task slots

xintongsong:

This pull request sets resource profiles for task slots according to resources the TaskExecutor is started with. This is one of the steps towards enabling fine grained resource management.

## Brief change log

- Add managed memory in ResourceSpec and ResourceProfile.
- Set resource profiles for task slots when starting TaskManagers.
- Set resource profiles for pending task manager slots.

## Verifying this change

 - Added TaskExecutorTest#testSlotReportResourceProfile that validates that slot report created by the TaskExecutor contains correct task slot resource profiles.
- Added YarnResourceManagerTest#testCreateSlotsPerWorker that validates that slot resource profile calculated by TaskManagers (for slot tables) and ResourceManagers (for pending task manager slots) are the same.

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-core/src/main/java/org/apache/flink/api/common/operators/ResourceSpec.java
+++ b/flink-core/src/main/java/org/apache/flink/api/common/operators/ResourceSpec.java
@@ -43,6 +43,7 @@ import java.util.Objects;
  *     <li>Direct Memory Size</li>
  *     <li>Native Memory Size</li>
  *     <li>State Size</li>
+ *     <li>Managed Memory Size</li>
  *     <li>Extended resources</li>
  * </ol>
  */
@@ -51,7 +52,7 @@ public class ResourceSpec implements Serializable {
 
 	private static final long serialVersionUID = 1L;
 
-	public static final ResourceSpec DEFAULT = new ResourceSpec(0, 0, 0, 0, 0);
+	public static final ResourceSpec DEFAULT = new ResourceSpec(0, 0, 0, 0, 0, 0);
 
 	public static final String GPU_NAME = "GPU";
 
@@ -70,6 +71,9 @@ public class ResourceSpec implements Serializable {
 	/** How many state size in mb are used. */
 	private final int stateSizeInMB;
 
+	/** The required amount of managed memory (in MB). */
+	private final int managedMemoryInMB;
+
 	private final Map<String, Resource> extendedResources = new HashMap<>(1);
 
 	/**
@@ -80,6 +84,7 @@ public class ResourceSpec implements Serializable {
 	 * @param directMemoryInMB The size of the java nio direct memory, in megabytes.
 	 * @param nativeMemoryInMB The size of the native memory, in megabytes.
 	 * @param stateSizeInMB The state size for storing in checkpoint.
+	 * @param managedMemoryInMB The size of managed memory, in megabytes.
 	 * @param extendedResources The extended resources, associated with the resource manager used
 	 */
 	protected ResourceSpec(
@@ -88,12 +93,14 @@ public class ResourceSpec implements Serializable {
 			int directMemoryInMB,
 			int nativeMemoryInMB,
 			int stateSizeInMB,
+			int managedMemoryInMB,
 			Resource... extendedResources) {
 		this.cpuCores = cpuCores;
 		this.heapMemoryInMB = heapMemoryInMB;
 		this.directMemoryInMB = directMemoryInMB;
 		this.nativeMemoryInMB = nativeMemoryInMB;
 		this.stateSizeInMB = stateSizeInMB;
+		this.managedMemoryInMB = managedMemoryInMB;
 		for (Resource resource : extendedResources) {
 			if (resource != null) {
 				this.extendedResources.put(resource.getName(), resource);
@@ -114,7 +121,8 @@ public class ResourceSpec implements Serializable {
 				this.heapMemoryInMB + other.heapMemoryInMB,
 				this.directMemoryInMB + other.directMemoryInMB,
 				this.nativeMemoryInMB + other.nativeMemoryInMB,
-				this.stateSizeInMB + other.stateSizeInMB);
+				this.stateSizeInMB + other.stateSizeInMB,
+				this.managedMemoryInMB + other.managedMemoryInMB);
 		target.extendedResources.putAll(extendedResources);
 		for (Resource resource : other.extendedResources.values()) {
 			target.extendedResources.merge(resource.getName(), resource, (v1, v2) -> v1.merge(v2));
@@ -142,6 +150,10 @@ public class ResourceSpec implements Serializable {
 		return this.stateSizeInMB;
 	}
 
+	public int getManagedMemory() {
+		return this.managedMemoryInMB;
+	}
+
 	public double getGPUResource() {
 		Resource gpuResource = extendedResources.get(GPU_NAME);
 		if (gpuResource != null) {
@@ -162,7 +174,7 @@ public class ResourceSpec implements Serializable {
 	 */
 	public boolean isValid() {
 		if (this.cpuCores >= 0 && this.heapMemoryInMB >= 0 && this.directMemoryInMB >= 0 &&
-				this.nativeMemoryInMB >= 0 && this.stateSizeInMB >= 0) {
+				this.nativeMemoryInMB >= 0 && this.stateSizeInMB >= 0 && managedMemoryInMB >= 0) {
 			for (Resource resource : extendedResources.values()) {
 				if (resource.getValue() < 0) {
 					return false;
@@ -187,7 +199,8 @@ public class ResourceSpec implements Serializable {
 		int cmp3 = Integer.compare(this.directMemoryInMB, other.directMemoryInMB);
 		int cmp4 = Integer.compare(this.nativeMemoryInMB, other.nativeMemoryInMB);
 		int cmp5 = Integer.compare(this.stateSizeInMB, other.stateSizeInMB);
-		if (cmp1 <= 0 && cmp2 <= 0 && cmp3 <= 0 && cmp4 <= 0 && cmp5 <= 0) {
+		int cmp6 = Integer.compare(this.managedMemoryInMB, other.managedMemoryInMB);
+		if (cmp1 <= 0 && cmp2 <= 0 && cmp3 <= 0 && cmp4 <= 0 && cmp5 <= 0 && cmp6 <= 0) {
 			for (Resource resource : extendedResources.values()) {
 				if (!other.extendedResources.containsKey(resource.getName()) ||
 					other.extendedResources.get(resource.getName()).getResourceAggregateType() != resource.getResourceAggregateType() ||
@@ -211,6 +224,7 @@ public class ResourceSpec implements Serializable {
 					this.directMemoryInMB == that.directMemoryInMB &&
 					this.nativeMemoryInMB == that.nativeMemoryInMB &&
 					this.stateSizeInMB == that.stateSizeInMB &&
+					this.managedMemoryInMB == that.managedMemoryInMB &&
 					Objects.equals(this.extendedResources, that.extendedResources);
 		} else {
 			return false;
@@ -225,6 +239,7 @@ public class ResourceSpec implements Serializable {
 		result = 31 * result + directMemoryInMB;
 		result = 31 * result + nativeMemoryInMB;
 		result = 31 * result + stateSizeInMB;
+		result = 31 * result + managedMemoryInMB;
 		result = 31 * result + extendedResources.hashCode();
 		return result;
 	}
@@ -240,7 +255,8 @@ public class ResourceSpec implements Serializable {
 				", heapMemoryInMB=" + heapMemoryInMB +
 				", directMemoryInMB=" + directMemoryInMB +
 				", nativeMemoryInMB=" + nativeMemoryInMB +
-				", stateSizeInMB=" + stateSizeInMB + extend +
+				", stateSizeInMB=" + stateSizeInMB +
+				", managedMemoryInMB=" + managedMemoryInMB + extend +
 				'}';
 	}
 
@@ -258,6 +274,7 @@ public class ResourceSpec implements Serializable {
 		private int directMemoryInMB;
 		private int nativeMemoryInMB;
 		private int stateSizeInMB;
+		private int managedMemoryInMB;
 		private GPUResource gpuResource;
 
 		public Builder setCpuCores(double cpuCores) {
@@ -285,6 +302,11 @@ public class ResourceSpec implements Serializable {
 			return this;
 		}
 
+		public Builder setManagedMemoryInMB(int managedMemory) {
+			this.managedMemoryInMB = managedMemory;
+			return this;
+		}
+
 		public Builder setGPUResource(double gpus) {
 			this.gpuResource = new GPUResource(gpus);
 			return this;
@@ -297,6 +319,7 @@ public class ResourceSpec implements Serializable {
 				directMemoryInMB,
 				nativeMemoryInMB,
 				stateSizeInMB,
+				managedMemoryInMB,
 				gpuResource);
 		}
 	}
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/clusterframework/types/ResourceProfile.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/clusterframework/types/ResourceProfile.java
@@ -284,6 +284,7 @@ public class ResourceProfile implements Serializable, Comparable<ResourceProfile
 			return this.cpuCores == that.cpuCores &&
 					this.heapMemoryInMB == that.heapMemoryInMB &&
 					this.directMemoryInMB == that.directMemoryInMB &&
+					this.nativeMemoryInMB == that.nativeMemoryInMB &&
 					this.networkMemoryInMB == that.networkMemoryInMB &&
 					Objects.equals(extendedResources, that.extendedResources);
 		}
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/clusterframework/types/ResourceProfile.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/clusterframework/types/ResourceProfile.java
@@ -51,7 +51,7 @@ public class ResourceProfile implements Serializable, Comparable<ResourceProfile
 	public static final ResourceProfile UNKNOWN = new ResourceProfile(-1.0, -1);
 
 	/** ResourceProfile which matches any other ResourceProfile. */
-	public static final ResourceProfile ANY = new ResourceProfile(Double.MAX_VALUE, Integer.MAX_VALUE, Integer.MAX_VALUE, Integer.MAX_VALUE, Integer.MAX_VALUE, Collections.emptyMap());
+	public static final ResourceProfile ANY = new ResourceProfile(Double.MAX_VALUE, Integer.MAX_VALUE, Integer.MAX_VALUE, Integer.MAX_VALUE, Integer.MAX_VALUE, Integer.MAX_VALUE, Collections.emptyMap());
 
 	// ------------------------------------------------------------------------
 
@@ -70,6 +70,9 @@ public class ResourceProfile implements Serializable, Comparable<ResourceProfile
 	/** Memory used for the task in the slot to communicate with its upstreams. Set by job master. */
 	private final int networkMemoryInMB;
 
+	/** The required amount of managed memory (in MB). */
+	private final int managedMemoryInMB;
+
 	/** A extensible field for user specified resources from {@link ResourceSpec}. */
 	private final Map<String, Resource> extendedResources = new HashMap<>(1);
 
@@ -83,6 +86,7 @@ public class ResourceProfile implements Serializable, Comparable<ResourceProfile
 	 * @param directMemoryInMB The size of the direct memory, in megabytes.
 	 * @param nativeMemoryInMB The size of the native memory, in megabytes.
 	 * @param networkMemoryInMB The size of the memory for input and output, in megabytes.
+	 * @param managedMemoryInMB The size of managed memory, in megabytes.
 	 * @param extendedResources The extended resources such as GPU and FPGA
 	 */
 	public ResourceProfile(
@@ -91,12 +95,14 @@ public class ResourceProfile implements Serializable, Comparable<ResourceProfile
 			int directMemoryInMB,
 			int nativeMemoryInMB,
 			int networkMemoryInMB,
+			int managedMemoryInMB,
 			Map<String, Resource> extendedResources) {
 		this.cpuCores = cpuCores;
 		this.heapMemoryInMB = heapMemoryInMB;
 		this.directMemoryInMB = directMemoryInMB;
 		this.nativeMemoryInMB = nativeMemoryInMB;
 		this.networkMemoryInMB = networkMemoryInMB;
+		this.managedMemoryInMB = managedMemoryInMB;
 		if (extendedResources != null) {
 			this.extendedResources.putAll(extendedResources);
 		}
@@ -109,7 +115,7 @@ public class ResourceProfile implements Serializable, Comparable<ResourceProfile
 	 * @param heapMemoryInMB The size of the heap memory, in megabytes.
 	 */
 	public ResourceProfile(double cpuCores, int heapMemoryInMB) {
-		this(cpuCores, heapMemoryInMB, 0, 0, 0, Collections.emptyMap());
+		this(cpuCores, heapMemoryInMB, 0, 0, 0, 0, Collections.emptyMap());
 	}
 
 	/**
@@ -123,6 +129,7 @@ public class ResourceProfile implements Serializable, Comparable<ResourceProfile
 				other.directMemoryInMB,
 				other.nativeMemoryInMB,
 				other.networkMemoryInMB,
+				other.managedMemoryInMB,
 				other.extendedResources);
 	}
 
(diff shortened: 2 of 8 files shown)


flink-yarn/src/test/java/org/apache/flink/yarn/YarnResourceManagerTest.java
```diff
@@ -522,4 +534,68 @@ public void testOnContainerCompleted() throws Exception {
 			});
 		}};
 	}
+
+	/**
+	 * Test that RM and TM calculate same slot resource profile.
```
gaoyunhaii: Tests

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleEnvironment.java
```diff
@@ -163,4 +163,11 @@
 	boolean updatePartitionInfo(
 		ExecutionAttemptID consumerID,
 		PartitionInfo partitionInfo) throws IOException, InterruptedException;
+
+	/**
+	 * Get total memory size for network buffers in bytes.
```
gaoyunhaii: Gets

flink-runtime/src/test/java/org/apache/flink/runtime/taskexecutor/TaskExecutorTest.java
```diff
@@ -1673,9 +1681,11 @@ public void testTerminationOnFatalError() throws Throwable {
 	@Test(timeout = 10000L)
 	public void testSyncSlotsWithJobMasterByHeartbeat() throws Exception {
 		final TaskSlotTable taskSlotTable = new TaskSlotTable(
-				Arrays.asList(ResourceProfile.UNKNOWN, ResourceProfile.UNKNOWN),
-				timerService);
-		final TaskManagerServices taskManagerServices = new TaskManagerServicesBuilder().setTaskSlotTable(taskSlotTable).build();
+			Arrays.asList(ResourceProfile.UNKNOWN, ResourceProfile.UNKNOWN),
+			timerService);
+		final TaskManagerServices taskManagerServices = new TaskManagerServicesBuilder().setTaskSlotTable(taskSlotTable)
+
```
gaoyunhaii: Remove this empty line.

flink-runtime/src/test/java/org/apache/flink/runtime/clusterframework/types/ResourceProfileTest.java
```diff
@@ -31,10 +31,10 @@
...
 	@Test
 	public void testMatchRequirement() throws Exception {
-		ResourceProfile rp1 = new ResourceProfile(1.0, 100, 100, 100, 0, Collections.emptyMap());
-		ResourceProfile rp2 = new ResourceProfile(1.0, 200, 200, 200, 0, Collections.emptyMap());
-		ResourceProfile rp3 = new ResourceProfile(2.0, 100, 100, 100, 0, Collections.emptyMap());
-		ResourceProfile rp4 = new ResourceProfile(2.0, 200, 200, 200, 0, Collections.emptyMap());
+		ResourceProfile rp1 = new ResourceProfile(1.0, 100, 100, 100, 0, 0, Collections.emptyMap());
+		ResourceProfile rp2 = new ResourceProfile(1.0, 200, 200, 200, 0, 0, Collections.emptyMap());
+		ResourceProfile rp3 = new ResourceProfile(2.0, 100, 100, 100, 0, 0, Collections.emptyMap());
+		ResourceProfile rp4 = new ResourceProfile(2.0, 200, 200, 200, 0, 0, Collections.emptyMap());
```
gaoyunhaii: Do we need to add a test for the newly added `approximate` method ？

flink-runtime/src/test/java/org/apache/flink/runtime/taskexecutor/TaskExecutorTest.java
```diff
@@ -1744,6 +1757,39 @@ public void testSyncSlotsWithJobMasterByHeartbeat() throws Exception {
 		}
 	}
 
+	@Test
+	public void testSlotReportResourceProfile() throws Throwable {
+		Configuration config = new Configuration();
+		config.setInteger(TaskManagerOptions.NUM_TASK_SLOTS, 2);
+		config.setString(NettyShuffleEnvironmentOptions.NETWORK_BUFFERS_MEMORY_MIN, "100m");
```
gaoyunhaii: I think it may be better to define the test data in variables, and get the target value in the following `assertEquals` with the relationship between the variables, like
<pre>
    int networkMemoryInMB = 100;
    ....
    assertEquals(networkMemoryInMB / numSlots, resourceProfile.getNetworkMemoryInMB());
</pre>

flink-yarn/src/main/java/org/apache/flink/yarn/YarnResourceManager.java
```diff
@@ -394,8 +394,10 @@ public void onContainersAllocated(List<Container> containers) {
 
 					try {
 						// Context information used to start a TaskExecutor Java process
+						// Use requested resource instead of allocated container resource, to make sure resource
```
gaoyunhaii: 1. Uses
2. A comma may be required for the last sentence.

flink-runtime/src/main/java/org/apache/flink/runtime/resourcemanager/slotmanager/SlotManager.java
```diff
@@ -585,7 +585,8 @@ private TaskManagerSlot createAndRegisterTaskManagerSlot(SlotID slotId, Resource
 	@Nullable
 	private PendingTaskManagerSlot findExactlyMatchingPendingTaskManagerSlot(ResourceProfile resourceProfile) {
 		for (PendingTaskManagerSlot pendingTaskManagerSlot : pendingSlots.values()) {
-			if (pendingTaskManagerSlot.getResourceProfile().equals(resourceProfile)) {
+			// there might be slight error between slot resource profiles calculated by RM / TM
+			if (pendingTaskManagerSlot.getResourceProfile().approximate(resourceProfile)) {
```
gaoyunhaii: I'm a little worry that the approximation may not be tolerant in TM side. For example, if the network memory of the available slot is just 1MB less than the request slot, it is very likely to cause the InsufficientBufferException in the TM side. 

I'm not sure why we can't make the request exactly the same with the available slot ? Or at least we must ensure pendingTaskManagerSlot.getResourceProfile().isMatching(resourceProfile) is also true.

flink-yarn/src/test/java/org/apache/flink/yarn/YarnResourceManagerTest.java
```diff
@@ -522,4 +534,68 @@ public void testOnContainerCompleted() throws Exception {
...
+		config2.setString(TaskManagerOptions.MANAGED_MEMORY_SIZE, "789m");
+		testCreateSlotsPerWorker(config2,  Resource.newInstance(800, 50));
+
+		Configuration config3 = new Configuration();
+		config3.setString(TaskManagerOptions.MANAGED_MEMORY_SIZE, "300m");
+		config3.setBoolean(TaskManagerOptions.MEMORY_OFF_HEAP, true);
+		testCreateSlotsPerWorker(config3,  Resource.newInstance(2000, 60));
+
+		Configuration config4 = new Configuration();
+		config4.setInteger(NettyShuffleEnvironmentOptions.NETWORK_NUM_BUFFERS, 50);
```
gaoyunhaii: I think it may be better to use the new NETWORK_BUFFERS_MEMORY_MIN/NETWORK_BUFFERS_MEMORY_MAX configuration instead.

gaoyunhaii:
Thank @xintongsong for opening the PR! The PR looks good to me as a whole and I have left some comments.

flink-runtime/src/main/java/org/apache/flink/runtime/resourcemanager/slotmanager/SlotManager.java
xintongsong: The reason we need approximation is due to precision issue. Currently slot profile are calculated on both RM and TM, with the same rules but different implementations. As a result, the calculation results may have slight difference. We are planing to unify the memory calculation so that RM and TM use the same implementation in the next version. Once we have that, the approximation could be removed.

xintongsong:
Thanks a lot for the review, @gaoyunhaii. I just addressed your comments.

xintongsong:

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleEnvironment.java
```diff
@@ -171,4 +171,11 @@ ShuffleIOOwnerContext createShuffleIOOwnerContext(
 	boolean updatePartitionInfo(
 		ExecutionAttemptID consumerID,
 		PartitionInfo partitionInfo) throws IOException, InterruptedException;
+
+	/**
+	 * Gets total memory size for network buffers in bytes.
+	 *
+	 * @return network memory size in bytes.
+	 */
+	long getMemorySize();
```
zhijiangW: Actually we considered this issue in [FLINK-12110](https://issues.apache.org/jira/browse/FLINK-12110). 
I guess we might also need to know the memory type besides size, because we could not make an assumption of direct memory as now based on pluggable shuffle service.

zhijiangW: My suggestion is not covering the network resources in `ResourceProfile` atm. As I know, the `ResourceProfile` from execution only covers managed memory from table api. If so we could also cover the managed memory to match as the first step, then focus on network resources future because it might involved in how to expose/calculate network resource on RM/JM side.

flink-runtime/src/main/java/org/apache/flink/runtime/taskexecutor/TaskManagerServices.java
```diff
@@ -325,6 +321,40 @@ public static TaskManagerServices fromConfiguration(
...
+	private static TaskSlotTable createTaskSlotTable(
+			TaskManagerServicesConfiguration taskManagerServicesConfiguration,
+			long networkMemorySize,
+			long managedMemorySize) {
+
+		final int numOfSlots = taskManagerServicesConfiguration.getNumberOfSlots();
+		final long heapMemoryBytesPerSlot = (taskManagerServicesConfiguration.getMemoryType() == MemoryType.HEAP ?
+			taskManagerServicesConfiguration.getMaxJvmHeapMemory() - managedMemorySize :
+			taskManagerServicesConfiguration.getMaxJvmHeapMemory()) / numOfSlots;
+		final long networkMemoryBytesPerSlot = networkMemorySize / numOfSlots;
```
zhijiangW: It might not cover network resource here as I suggested above.

flink-runtime/src/main/java/org/apache/flink/runtime/resourcemanager/ResourceManager.java
```diff
@@ -1191,8 +1197,32 @@ protected int getNumberRequiredTaskManagerSlots() {
 	//  Helper methods
 	// ------------------------------------------------------------------------
 
-	protected static Collection<ResourceProfile> createSlotsPerWorker(int numSlots) {
-		return Collections.nCopies(numSlots, ResourceProfile.ANY);
+	@VisibleForTesting
+	public static Collection<ResourceProfile> createSlotsPerWorker(Configuration config, long totalMemoryMB, int numSlots) {
+		final long cutoffMB = ContaineredTaskManagerParameters.calculateCutoffMB(config, totalMemoryMB);
+
+		final long networkMB = NettyShuffleEnvironmentConfiguration.calculateNetworkBufferMemory(
```
zhijiangW: In reality `NettyShuffleEnvironmentConfiguration.calculateNetworkBufferMemory` should not be exposed outside based on pluggable shuffle service proposed in [FLINK-12110](https://issues.apache.org/jira/browse/FLINK-12110), so I suggest not covering network factor in `ResourceProfile` atm.

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleEnvironment.java
xintongsong: Thanks for the input, @zhijiangW. I agree with you that we do not necessarily need matching network memory at the moment. I'll update the PR accordingly.

flink-runtime/src/main/java/org/apache/flink/runtime/resourcemanager/ResourceManager.java
xintongsong: I agree that we do not need to cover network memory in `ResourceProfile` in this version. However, I'm afraid we need to access `NettyShuffleEnvironmentConfiguration.calculateNetworkBufferMemory` on RM side anyway. Otherwise, we can not calculate managed memory size from the total container memory size. I think in the future we need a dedicated data structure to unify all the task executor resource calculations, exposing accesses of individual factors while hiding the detailed calculation steps.

xintongsong:
Thank you for the review, @zhijiangW. I'll address your comments.

flink-runtime/src/main/java/org/apache/flink/runtime/taskexecutor/TaskExecutor.java
```diff
@@ -1621,6 +1621,11 @@ TaskExecutorToResourceManagerConnection getResourceManagerConnection() {
 		return resourceManagerHeartbeatManager;
 	}
 
+	@VisibleForTesting
+	SlotReport getSlotReport() {
```
StephanEwen: You can probably drop this method if you move the code that computes the slot resources into a separate method:
`computeSlotResourceProfile(taskManagerServicesConfiguration)` as suggested. Then you can test that logic in isolation and do not need access to the slot requests.

flink-runtime/src/main/java/org/apache/flink/runtime/taskexecutor/TaskManagerServices.java
```diff
@@ -258,20 +259,13 @@ public static TaskManagerServices fromConfiguration(
 
 		// this call has to happen strictly after the network stack has been initialized
 		final MemoryManager memoryManager = createMemoryManager(taskManagerServicesConfiguration);
+		final long managedMemorySize = memoryManager.getMemorySize();
 
 		final BroadcastVariableManager broadcastVariableManager = new BroadcastVariableManager();
 
-		final List<ResourceProfile> resourceProfiles = new ArrayList<>(taskManagerServicesConfiguration.getNumberOfSlots());
```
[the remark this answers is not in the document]
StephanEwen: This also allows for targeted testing of the resource computation code, which avoids many of the complexity of the new tests.

flink-runtime/src/test/java/org/apache/flink/runtime/taskexecutor/TaskExecutorTest.java
```diff
@@ -1729,14 +1738,16 @@ public void testSyncSlotsWithJobMasterByHeartbeat() throws Exception {
 			final AllocationID allocationIdOnlyInJM = new AllocationID();
 			final AllocationID allocationIdOnlyInTM = new AllocationID();
 
-			taskExecutorGateway.requestSlot(slotId1, jobId, allocationIdInBoth, "foobar", testingResourceManagerGateway.getFencingToken(), timeout);
-			taskExecutorGateway.requestSlot(slotId2, jobId, allocationIdOnlyInTM, "foobar", testingResourceManagerGateway.getFencingToken(), timeout);
+			taskExecutorGateway.requestSlot(slotId1, jobId, allocationIdInBoth, "foobar",
```
StephanEwen: Lots of unrelated reformatting. Per Flink coding guides, these should be done in a change.

flink-runtime/src/test/java/org/apache/flink/runtime/taskexecutor/TaskExecutorTest.java
```diff
@@ -1845,6 +1860,42 @@ public void testSlotReportDoesNotContainStaleInformation() throws Exception {
 		}
 	}
 
+	@Test
+	public void testSlotReportResourceProfile() throws Throwable {
```
StephanEwen: I think this test shows that the abstractions are too intermingled.

The code that was added in TaskExecutor was a change to replace the UNKNOWN profile attached to a slot by a profile with proper values. The code only takes the task manager's config (or managed memory size), and determines the size of the slots.

Such a change should not need the creation of a full TaskExecutor to test it. That means something is not right in the abstraction.

This test setup means that every time the TaskExecutor setup changes, the test for the simple code "config -> profile" needs to be adjusted, which contributes to making maintenance complex.

flink-runtime/src/main/java/org/apache/flink/runtime/resourcemanager/slotmanager/SlotManager.java
StephanEwen: Why does the RM re-calculate the resource profile? This assumes that the RM and TM not only follow the same rules, but have the same config. That is actually not always the case, I have seen standalone setups where different TMs had different configs.

xintongsong:
Thanks for the review, @StephanEwen. 

I would like to explain regarding your concern about the assumption of RM/TM having same configuration:

The reason we need to calculate TM's slot resource profiles on RM side is that, we need to set resource profile for `PendingTaskManagerSlot` before the corresponding TM is started. 

Currently, Flink can assign a pending slot to a slot request before the TM is started and registered. In this way, the subsequent slot requests will first consume slots on the pending TM (for multi-slot TMs) before requesting and launching a new one. When the TM is registered, the SlotManager matches the registered new slot to a `PendingTaskManagerSlot` with the same resource profile, and assigns the registered slot to the same slot request that the pending slot is assigned to (if any).

Before this PR, both the pending slot on RM side and the actual slot on TM side have the same resource profile `ANY`, which can be matched with the method `equals`. Since this PR sets the slot resource profile on TM side to the actual resource of the slot, we need to set the resource profile for the pending slots on RM side in the same way. This is way I introduced calculating TM's slot resource profiles on RM side, and the approximate matching.

The assigning over pending slots and the RM side slot resource calculating only happens on Yarn/Mesos. In these scenarios, TMs do have the same configuration as RM does, which is transmitted from RM side. For a standalone cluster, there should be no pending slots because RM can not actively start any TM.

Except for the `PendingTaskManagerSlot`, RM does use the slot resource profile reported from TM for matching slot request against registered slots, and converting requested `UNKNOWN` resource profile to a default value (as shown in the following PR #8846 for dynamic managed memory). Therefore, it should not cause problems on a standalone cluster with TMs having different configs.

It's my bad not making these clear in codes and comments. For the rest of your comments, I'll address them ASAP. I especially admire your suggestions on encapsulation and simplifying tests. It's a good lesson for me. Thank you again.

flink-core/src/main/java/org/apache/flink/api/common/operators/ResourceSpec.java
```diff
@@ -240,7 +254,8 @@ public String toString() {
 				", heapMemoryInMB=" + heapMemoryInMB +
 				", directMemoryInMB=" + directMemoryInMB +
 				", nativeMemoryInMB=" + nativeMemoryInMB +
-				", stateSizeInMB=" + stateSizeInMB + extend +
+				", stateSizeInMB=" + stateSizeInMB +
+				", managedMemoryInMB=" + managedMemoryInMB + extend +
```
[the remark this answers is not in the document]
xintongsong: I don't think there is any missing space. The string `extend` is either start with ", " or empty.

flink-runtime/src/main/java/org/apache/flink/runtime/clusterframework/types/ResourceProfile.java
```diff
@@ -284,12 +301,39 @@ else if (obj != null && obj.getClass() == ResourceProfile.class) {
...
 					Objects.equals(extendedResources, that.extendedResources);
 		}
 		return false;
 	}
 
+	public boolean approximate(ResourceProfile that) {
+		if (that == null) {
+			return false;
+		}
+		if (Math.abs(this.cpuCores - that.cpuCores) > 1e-6f) {
```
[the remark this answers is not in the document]
xintongsong: Yes, this is about the rounding errors during the profile calculation.

The calculation of profiles involves multiplying integer values with floating values. Rounding the floating value product to integer value will cause error. E.g., rounded result of `Total * Fraction` and `Total - Total * (1 - Fraction)` may be different, where `Total` is an integer value and `Fraction` is a floating value.

Each time we do such rounding, we may get an error with max value 1. Since there are two of such fraction based calculation (for managed memory and network memory), I set the max error allowed here to 2.

This approximate matching is only used for matching `PendingTaskManagerSlot` with slot registered from TM. It can be replaced with exact matching once we unify the TM resource configuration. After we unify the TM resource configurations, there will be no more resource calculations on TM side. For Yarn, RM will calculate the resource profile and pass the calculation result to the TM to be started, so the pending slots and the actual slots should have exact same profiles. For standalone, there shouldn't be any pending slots.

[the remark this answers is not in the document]
xintongsong: I'll add comments for these explanations.

flink-runtime/src/main/java/org/apache/flink/runtime/clusterframework/types/ResourceProfile.java
```diff
@@ -301,7 +345,8 @@ public String toString() {
 			", heapMemoryInMB=" + heapMemoryInMB +
 			", directMemoryInMB=" + directMemoryInMB +
 			", nativeMemoryInMB=" + nativeMemoryInMB +
-			", networkMemoryInMB=" + networkMemoryInMB + resources +
+			", networkMemoryInMB=" + networkMemoryInMB +
+			", managedMemoryInMB=" + managedMemoryInMB + resources +
```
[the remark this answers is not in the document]
xintongsong: Same as above. The string `resources` is either start with ", " or empty.

flink-runtime/src/main/java/org/apache/flink/runtime/clusterframework/types/ResourceProfile.java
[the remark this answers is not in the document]
xintongsong: As we discussed offline, I removed this method.

flink-runtime/src/main/java/org/apache/flink/runtime/resourcemanager/slotmanager/SlotManager.java
xintongsong: As we discussed offline, I reverted changes here.

flink-runtime/src/main/java/org/apache/flink/runtime/taskexecutor/TaskManagerServices.java
```diff
@@ -325,6 +319,38 @@ public static TaskManagerServices fromConfiguration(
 			.createShuffleEnvironment(shuffleEnvironmentContext);
 	}
 
+	private static TaskSlotTable createTaskSlotTable(
+			TaskManagerServicesConfiguration taskManagerServicesConfiguration,
+			long managedMemorySize) {
+
+		final int numOfSlots = taskManagerServicesConfiguration.getNumberOfSlots();
+		final long heapMemoryBytesPerSlot = (taskManagerServicesConfiguration.getMemoryType() == MemoryType.HEAP ?
```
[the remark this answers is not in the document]
xintongsong: As we discussed offline, I reverted this change.

flink-runtime/src/main/java/org/apache/flink/runtime/taskexecutor/TaskManagerServices.java
```diff
@@ -358,14 +384,14 @@ private static MemoryManager createMemoryManager(
 			float memoryFraction = taskManagerServicesConfiguration.getMemoryFraction();
 
 			if (memType == MemoryType.HEAP) {
-				long freeHeapMemoryWithDefrag = taskManagerServicesConfiguration.getFreeHeapMemoryWithDefrag();
-				// network buffers allocated off-heap -> use memoryFraction of the available heap:
-				long relativeMemSize = (long) (freeHeapMemoryWithDefrag * memoryFraction);
+				long maxJvmHeapMemory = taskManagerServicesConfiguration.getMaxJvmHeapMemory();
```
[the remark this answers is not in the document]
xintongsong: As we discussed offline, I reverted this change.

flink-core/src/main/java/org/apache/flink/api/common/operators/ResourceSpec.java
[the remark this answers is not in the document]
StephanEwen: You are right, the delimiter is in the first loop so all is fine.

flink-runtime/src/main/java/org/apache/flink/runtime/clusterframework/types/ResourceProfile.java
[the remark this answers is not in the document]
StephanEwen: You are right!

StephanEwen:
Thanks @xintongsong this looks really good now.

Will take this over and merge the PR.

There are some minor adjustments I would like to do (like de-duplicate configuration logic between Yarn/Mesos and and de-duplicate memory size logic between TaskManagerServices and ResourceManager) but I would apply these on the fly during the merge, due to the feature freeze deadline...

StephanEwen:
I have applied the fixes here: https://github.com/StephanEwen/incubator-flink/tree/xintong

Trying to merge this, unfortunately battling with some tests that are very sensitive to rounding errors in the whole bytes/megabytes computation.

StephanEwen:
Manually merged in c8b93b354d6d72ef3cacc9b0b9adcdb1df26e625

PR #8706, opened 2019-06-12
[FLINK-12814][sql-client] Support a traditional and scrolling view of…

docete:
… result

Introduce a traditional and scrolling (non-interactive) view of result in table mode. 

## Brief change log

- Introduce a new interface **FinalizedResult** which represented as the finalized result consisting all latest insert records.
- Introduce an execution option: execution.interactive-view.enabled. CliClient works in original way when it's set to 'true'. By default it is 'true'.
- Renamed **MaterializedCollectBatchResult** to **CollectBatchResult** and implements **FinalizedResult**. When execution.interactive-view.enabled is set to 'false', **CollectBatchResult** would return all fetched rows. CliClient output the rows in a non-interactive way.
- Introduce **FinalizedCollectStreamResult** which implements **FinalizedResult**. When execution.interactive-view.enabled is set to 'false', **FinalizedCollectStreamResult** would submit user's query and wait until the job finished, failed, interrupted by user(ctrl+c) or  time out(in a user defined timeout). And return all fetched rows. CliClient output the rows in a non-interactive way. 

## Verifying this change
 - Manually verified the change by running a standalone cluster with 1 JobManager and 1 TaskManager, a sql client(with execution.interactive-view.enabled=false), and submitting queries to sql client, verifying that result rows are printed in a traditional and scrolling way.

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/cli/CliClient.java
+++ b/flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/cli/CliClient.java
@@ -28,6 +28,7 @@ import org.apache.flink.table.client.gateway.ProgramTargetDescriptor;
 import org.apache.flink.table.client.gateway.ResultDescriptor;
 import org.apache.flink.table.client.gateway.SessionContext;
 import org.apache.flink.table.client.gateway.SqlExecutionException;
+import org.apache.flink.types.Row;
 
 import org.jline.reader.EndOfFileException;
 import org.jline.reader.LineReader;
@@ -464,6 +465,7 @@ public class CliClient {
 	}
 
 	private void callSelect(SqlCommandCall cmdCall) {
+		final long start = System.currentTimeMillis();
 		final ResultDescriptor resultDesc;
 		try {
 			resultDesc = executor.executeQuery(context, cmdCall.operands[0]);
@@ -471,21 +473,30 @@ public class CliClient {
 			printExecutionException(e);
 			return;
 		}
-		final CliResultView view;
-		if (resultDesc.isMaterialized()) {
-			view = new CliTableResultView(this, resultDesc);
+		final long end = System.currentTimeMillis();
+
+		if (resultDesc.isFinalized()) {
+			printRows(resultDesc, start, end);
+			executor.cancelQuery(context, resultDesc.getResultId());
+		} else if (resultDesc.isMaterialized()) {
+			final CliResultView view;
+			if (resultDesc.isMaterialized()) {
+				view = new CliTableResultView(this, resultDesc);
+			} else {
+				view = new CliChangelogResultView(this, resultDesc);
+			}
+
+			// enter view
+			try {
+				view.open();
+
+				// view left
+				printInfo(CliStrings.MESSAGE_RESULT_QUIT);
+			} catch (SqlExecutionException e) {
+				printExecutionException(e);
+			}
 		} else {
-			view = new CliChangelogResultView(this, resultDesc);
-		}
-
-		// enter view
-		try {
-			view.open();
-
-			// view left
-			printInfo(CliStrings.MESSAGE_RESULT_QUIT);
-		} catch (SqlExecutionException e) {
-			printExecutionException(e);
+			throw new SqlExecutionException("Unsupported result type");
 		}
 	}
 
@@ -611,6 +622,20 @@ public class CliClient {
 	private void printInfo(String message) {
 		terminal.writer().println(CliStrings.messageInfo(message).toAnsi());
 		terminal.flush();
+	}
+
+	/**
+	 * Output Results in non-interactive way.
+	 *
+	 * @return
+	 */
+	void printRows(final ResultDescriptor desc, long start, long end) {
+		final String msg = "Time taken: %.03f seconds, Fetched %d row(s).";
+		List<Row> result = this.executor.retrieveResult(desc.getResultId());
+		terminal.writer().println("OK");
+		result.forEach((row) -> terminal.writer().println(row.toString()));
+		terminal.writer().println(String.format(msg, (end - start) / 1000.0, result.size()));
+		terminal.writer().flush();
 	}
 
 	// --------------------------------------------------------------------------------------------
--- a/flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/config/entries/ExecutionEntry.java
+++ b/flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/config/entries/ExecutionEntry.java
@@ -108,6 +108,10 @@ public class ExecutionEntry extends ConfigEntry {
 
 	public static final String EXECUTION_CURRENT_DATABASE = "current-database";
 
+	private static final String EXECUTION_INTERACTIVE_VIEW_ENABLED = "interactive-view.enabled";
+
+	private static final String EXECUTION_MAX_QUERY_TIMEOUT_MS = "max-query-timeout-ms";
+
 	private ExecutionEntry(DescriptorProperties properties) {
 		super(properties);
 	}
@@ -152,6 +156,8 @@ public class ExecutionEntry extends ConfigEntry {
 		properties.validateInt(EXECUTION_RESTART_STRATEGY_MAX_FAILURES_PER_INTERVAL, true, 1);
 		properties.validateString(EXECUTION_CURRENT_CATALOG, true, 1);
 		properties.validateString(EXECUTION_CURRENT_DATABASE, true, 1);
+		properties.validateBoolean(EXECUTION_INTERACTIVE_VIEW_ENABLED, true);
+		properties.validateInt(EXECUTION_MAX_QUERY_TIMEOUT_MS, true, -1);
 	}
 
 	public EnvironmentSettings getEnvironmentSettings() {
@@ -310,6 +316,12 @@ public class ExecutionEntry extends ConfigEntry {
 		return properties.getOptionalString(EXECUTION_CURRENT_DATABASE);
 	}
 
+	public int getMaxQueryTimeoutMs() {
+		// defaultValue -1 means do not timeout
+		return properties.getOptionalInt(EXECUTION_MAX_QUERY_TIMEOUT_MS)
+			.orElseGet(() -> useDefaultValue(EXECUTION_MAX_QUERY_TIMEOUT_MS, -1));
+	}
+
 	public boolean isChangelogMode() {
 		return properties.getOptionalString(EXECUTION_RESULT_MODE)
 			.map((v) -> v.equals(EXECUTION_RESULT_MODE_VALUE_CHANGELOG))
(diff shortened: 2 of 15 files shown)


docete:
@twalthr Please take a look at this if you have time.

bowenli86:
@docete can you please rebase this PR?

flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/config/entries/ExecutionEntry.java
```diff
@@ -152,6 +156,8 @@ protected void validate(DescriptorProperties properties) {
 		properties.validateInt(EXECUTION_RESTART_STRATEGY_MAX_FAILURES_PER_INTERVAL, true, 1);
 		properties.validateString(EXECUTION_CURRENT_CATALOG, true, 1);
 		properties.validateString(EXECUTION_CURRENT_DATABASE, true, 1);
+		properties.validateBoolean(EXECUTION_INTERACTIVE_VIEW_ENABLED, true);
+		properties.validateInt(EXECUTION_MAX_QUERY_TIMEOUT_MS, true, -1);
```
bowenli86: nit: feels like we can set a default timeout when EXECUTION_INTERACTIVE_VIEW_ENABLED is false?

flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/gateway/local/result/CollectBatchResult.java
```diff
@@ -126,9 +149,24 @@ else if (!snapshotted) {
 		}
 	}
 
+	@Override
+	public List<Row> retrieveResult() {
+		try {
+			if (executionException != null) {
+				throw executionException;
+			}
+			// wait until retrievalT
```
bowenli86:
```suggestion
			// wait until retrieval
```

flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/gateway/local/result/CollectBatchResult.java
```diff
@@ -126,9 +149,24 @@ else if (!snapshotted) {
...
+			//hread finished.
```
bowenli86:
```suggestion
			// read finished.
```

flink-table/flink-sql-client/src/test/java/org/apache/flink/table/client/gateway/utils/UserDefinedFunctions.java
```diff
@@ -47,6 +47,27 @@ public String eval(Integer i) {
 		}
 	}
 
+	/**
+	 * The scalar function for SQL Client test.
+	 */
+	public static class TimeConsumingUDF extends ScalarFunction {
```
bowenli86: in which test is this class used?

docete:
> wondering should value of EXECUTION_INTERACTIVE_VIEW_ENABLED be deduced automatically based on execution mode? e.g. streaming -> true, batch -> false?

result mode(non-interactive/original interactive) and execution mode(streaming/batch) should be orthogonal.

flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/cli/CliClient.java
```diff
@@ -613,6 +624,20 @@ private void printInfo(String message) {
 		terminal.flush();
 	}
 
+	/**
+	 * Output Results in non-interactive way.
+	 *
+	 * @return
+	 */
+	void printRows(final ResultDescriptor desc, long start, long end) {
+		final String msg = "Time taken: %.03f seconds, Fetched %d row(s).";
```
twalthr: Use `printInfo`. For distinguishing between rows and info rows.

flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/cli/CliClient.java
```diff
@@ -613,6 +624,20 @@ private void printInfo(String message) {
 		terminal.flush();
 	}
 
+	/**
+	 * Output Results in non-interactive way.
+	 *
+	 * @return
+	 */
+	void printRows(final ResultDescriptor desc, long start, long end) {
```
twalthr: Give method a more meaningful name such as `printNonInteractiveResult`. Then the Javadoc is not required anymore.

flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/gateway/local/result/BasicResult.java
```diff
@@ -50,4 +50,12 @@ public String getWebInterfaceUrl() {
 		}
 		return webInterfaceUrl;
 	}
+
+	/**
+	 * Result types.
```
twalthr: Make this a useful Javadoc and explain the difference between those two types.

flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/gateway/local/result/FinalizedCollectStreamResult.java
```diff
@@ -0,0 +1,254 @@
...
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+/**
+ * Collects results and returns them as finalized table.
+ *
+ * @param <C> cluster id to which this result belongs to
+ */
+public class FinalizedCollectStreamResult<C> extends BasicResult<C> implements DynamicResult<C>, FinalizedResult<C> {
```
twalthr: Add a test for this logic. One IT case in LocalExecutorITCase and unit tests for the `Result` class.

flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/gateway/local/result/FinalizedCollectStreamResult.java
```diff
@@ -0,0 +1,254 @@
...
+
+	private static final Logger LOG = LoggerFactory.getLogger(FinalizedCollectStreamResult.class);
+	private final TypeInformation<Row> outputType;
+	private final SocketStreamIterator<Tuple2<Boolean, Row>> iterator;
+	private final CollectStreamTableSink collectTableSink;
+	private final ResultRetrievalThread retrievalThread;
+	private final CountDownLatch resultLatch;
+	private final List<Row> finalizedTable;
+	/**
+	 * Caches the last row position for faster access. The position might not be exact (if rows
```
twalthr: You copied this from an other class, right? Please abstract correctly to remove code duplication.

flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/config/entries/ExecutionEntry.java
```diff
@@ -108,6 +108,10 @@
 
 	public static final String EXECUTION_CURRENT_DATABASE = "current-database";
 
+	private static final String EXECUTION_INTERACTIVE_VIEW_ENABLED = "interactive-view.enabled";
```
twalthr: I think I asked that in the past but why do we need an additional property? We have a property `result-mode` which could be extended by another enum value. Every additional config param should be well thought through because it increases the complexity for the user.

flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/config/entries/ExecutionEntry.java
```diff
@@ -108,6 +108,10 @@
...
+
+	private static final String EXECUTION_MAX_QUERY_TIMEOUT_MS = "max-query-timeout-ms";
```
twalthr: When do you think a user wants to set this? If the connection to the cluster is bad?

flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/cli/CliClient.java
```diff
@@ -464,28 +465,38 @@ private void callExplain(SqlCommandCall cmdCall) {
...
 			printExecutionException(e);
 			return;
 		}
-		final CliResultView view;
-		if (resultDesc.isMaterialized()) {
-			view = new CliTableResultView(this, resultDesc);
-		} else {
-			view = new CliChangelogResultView(this, resultDesc);
-		}
+		final long end = System.currentTimeMillis();
```
twalthr: How can a user cancel the submitted query?

twalthr:
Actually, this PR requires a FLIP because it introduces new user-facing features.

flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/cli/CliClient.java
docete: User can wait a timeout(setting by max-query-timeout-ms) or just type CTRL + C to terminate the submitted query.

flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/cli/CliClient.java
docete: OK

flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/cli/CliClient.java
docete: OK

flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/config/entries/ExecutionEntry.java
docete: result-mode describes how we organize the result (change-log or table). and this switch describes how we show the result to user(interactive or non-interactive). IMO they are two different dimensions.

flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/config/entries/ExecutionEntry.java
docete: It offers an overall protection for the user or downstream projects submitting a long-running query(e.g., select * from kafakTable). IMO for DQL cases in a streaming scenario, user always expects the result is finalizable(He/She can not stare at the screen all of the day). For user, he can use CTRL+C signal. For downstream project, a timeout should be better.

flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/gateway/local/result/BasicResult.java
docete: OK

flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/gateway/local/result/FinalizedCollectStreamResult.java
docete: OK, will add test case soon.

flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/gateway/local/result/FinalizedCollectStreamResult.java
docete: Yes, copied some code from CollectStreamResult. Will do a retractor to remove duplication code.

docete:
> Actually, this PR requires a FLIP because it introduces new user-facing features.

Shall I complement a FLIP, or just address feedbacks in this PR?

PR #8707, opened 2019-06-12
[FLINK-12815] [table-planner-blink] Supports CatalogManager in blink planner

godfreyhe:

*Supports CatalogManager in blink planner (which is what FLINK-11476 has done)*

## Brief change log

  - *all table registrations and lookups in TableEnvironment go through CatalogManager*
  - *added Calcite's Schema structure corresponding to the CatalogManager. This enables Calcite to lookup for Tables through CatalogManager*
  - *introduce CrudExternalCatalog and InMemoryExternalCatalog and catalog exceptions*
  - *add FileSystem and OldCsv descriptors for test*

## Verifying this change

  - *All existent tests should pass, except catalog name and database name are added to table scan digest*
  - *Added test that validates CatalogManager logic in blink planner*

## Documentation


dawidwys:
We should definitely split this commit. We must not merge commits of this size.
It is impossible to:
1. review 
2. understand reasoning behind the change in the future,
3. track back to original changes.

godfreyhe:
thanks for your suggestion @dawidwys , i will try to split this PR into 4 PRs as following:
1. introduce CrudExternalCatalog and related catalog exceptions
2. the intermediate optimization result does not need register as table in batch and stream CommonSubGraphBasedOptimizer
3. unify BatchTableSourceTable and StreamTableSourceTable to TableSourceTable
4. Supports CatalogManager in blink planner

however the last PR is also very enormous due to catalog name and database name will be added for each test case result and there are about 100 test xml files.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/plan/util/RelShuttles.scala
```diff
@@ -97,14 +97,15 @@ class ExpandTableScanShuttle extends RelShuttleImpl {
...
+    * Converts [[LogicalTableScan]] the result [[RelNode]] tree
+    * by calling [[QueryOperationCatalogViewTable]]#toRel
     */
   override def visit(scan: TableScan): RelNode = {
     scan match {
       case tableScan: LogicalTableScan =>
-        val relTable = tableScan.getTable.unwrap(classOf[RelTable])
-        if (relTable != null) {
-          val rel = relTable.toRel(RelOptUtil.getContext(tableScan.getCluster), tableScan.getTable)
+        val viewTable = tableScan.getTable.unwrap(classOf[QueryOperationCatalogViewTable])
```
docete: AbstractCatalogView from Catalog has no QueryOperation. We must get the underlie QueryOperation and convert to QueryOperationCatalogViewTable somewhere.

docete:
+1 LGTM, leave a comment about CatalogView expanding.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/plan/util/RelShuttles.scala
godfreyhe: thanks for your suggestion @docete , I have add a `TODO` in DatabaseCalciteSchema, and will support expanding `GenericCatalogView` later

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/catalog/CatalogCalciteSchema.java
```diff
@@ -0,0 +1,117 @@
...
+
+	/**
+	 * Look up a sub-schema (database) by the given sub-schema name.
+	 *
+	 * @param schemaName name of sub-schema to look up
+	 * @return the sub-schema with a given database name, or null
+	 */
+	@Override
+	public Schema getSubSchema(String schemaName) {
+
```
danny0405: This name `getSubSchema` seems a little misleading, since we always returns a `DatabaseCalciteSchema`, i know that `getSubSchema` is a interface defined by Calcite, but maybe we can have an abstraction of our own.

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/catalog/CatalogManagerCalciteSchema.java
```diff
@@ -0,0 +1,123 @@
...
+ * the full path of objects is of format [catalog_name].[db_name].[meta-object_name].
+ *
+ * <p>It also supports {@link ExternalCatalog}s. An external catalog maps 1:1 to the Calcite's schema.
+ */
+@Internal
+public class CatalogManagerCalciteSchema implements Schema {
+
+	private final CatalogManager catalogManager;
+	private boolean isBatch;
+
```
danny0405: Does this mean we can only have one kind of table for the whole `CatalogManager` ? Just curious why the `CatalogManager` would take care of this attribute, isn't this an attribute of just a single `table` ?

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/catalog/DatabaseCalciteSchema.java
```diff
@@ -0,0 +1,159 @@
...
+		return new HashSet<>();
+	}
+
+	@Override
+	public Collection<Function> getFunctions(String s) {
+		return new HashSet<>();
+	}
+
+	@Override
+	public Set<String> getFunctionNames() {
```
danny0405: Some of the interfaces inherited from Calcite `Schema` seems useless now, can we have a uniform impl for all these Flink schema implementations ?

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/catalog/FlinkTempCatalogTable.java
```diff
@@ -0,0 +1,79 @@
...
+
+import org.apache.calcite.schema.impl.AbstractTable;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+
+/**
+ * Represents AbstractTable in Flink catalog.
+ */
```
danny0405: Please give more comments about why this special table is needed and how it should be used.

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/catalog/CatalogCalciteSchema.java
godfreyhe: do you mean we need an interface which defines methods like: `getDatabase(String database)` or `getTable(String table)` ?

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/catalog/CatalogManagerCalciteSchema.java
godfreyhe: we unify `BatchTableSourceTable` and `StreamTableSourceTable` into `TableSourceTable`,  and batch `TableSourceTable` and stream `TableSourceTable` have different behaviors in infering table row type.

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/catalog/DatabaseCalciteSchema.java
godfreyhe: make sense

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/catalog/FlinkTempCatalogTable.java
godfreyhe: This class should be delete due to there is no need create a table to register, `DataStream` or `TableSource` should be wrapped as `QueryOperation` and registered as table

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/plan/util/RelShuttles.scala
dawidwys: This is something that has to be done also for the legacy planner. The problem is that we have to pass somehow the sql parser, as the CatalogView translates to a string query.

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/catalog/CatalogCalciteSchema.java
dawidwys: I don't necessarily understand @danny0405's concerns. We do have abstractions of our own. This class is just an adapter between our abstractions (Catalog, CatalogManager, Database etc.) and Calcite's (Schema).

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/plan/util/RelShuttles.scala
godfreyhe: yes, we meet this problem in blink code. and we put the `FlinkPlannerImpl` instance into the calcite planner context, and get it from context to parser a string query in `CatalogTableToTableSourceRule`. it's a little hack! We will try to find a clean way to resolve it  in next pr.

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/catalog/CatalogCalciteSchema.java
danny0405: @dawidwys If we see Calcite's `Schema` interfaces definition, `Schema#getTable`, `Schema#getTableNames`, `Schema#getFunction` ... It means that for every level schema, we could fetch table/functions, but for this abstraction, `CatalogCalciteSchema` can only see database, and `DataBaseCalciteSchema` can only see table.

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/catalog/CatalogManagerCalciteSchema.java
dawidwys: We don't need that flag in blink. In the legacy planner we need that because we use either `BatchTableSource` or `StreamTableSource` So the answer to original @danny0405  question is yes, in the legacy planner we can have only a single kind of `TableSource`. We should get rid of it for blink planner.

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/catalog/CatalogCalciteSchema.java
dawidwys: Still don't necessarily get that. What do you suggest? To have additional level of abstractions:

    `Schema` <- `SubSchemaOnlySchema` <- `CatalogCalciteSchema` 

where `DatabaseOnlySchema` would have methods like that:

	@Override
	public Schema getSubSchema(String schemaName) {
		getDatabaseSchema(schemaName)
	}

        protected DatabaseCalciteSchema getDatabaseSchema(String schemaName)

or sth similar? I think this is would be over-engineering, but I am not strongly against it

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/catalog/CatalogManagerCalciteSchema.java
godfreyhe: currently, the table need to distinguish between batch table and stream table. that is the cornerstone. `Table` is from `Schema#getTable`, so i think `Schema` also need know the mode of a job.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/descriptors/OldCsv.scala
```diff
@@ -0,0 +1,200 @@
...
+  * Note: This descriptor describes Flink's non-standard CSV table source/sink. In the future, the
+  * descriptor will be replaced by a proper RFC-compliant version. Use the RFC-compliant `Csv`
+  * format in the dedicated `flink-formats/flink-csv` module instead when writing to Kafka. Use the
+  * old one for stream/batch filesystem operations for now.
+  *
+  * @deprecated Use the RFC-compliant `Csv` format in the dedicated
+  *            `flink-formats/flink-csv` module instead when writing to Kafka.
+  */
+@Deprecated
+@deprecated
```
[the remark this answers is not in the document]
godfreyhe: `Deprecated` is for java, and `deprecated` is for scala.

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/catalog/CatalogCalciteSchema.java
godfreyhe: That makes the schema more complex, i think.

godfreyhe:
@dawidwys do you have any other suggestions? this PR has blocked other blink merging work, I hope to speed it up. thanks~

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/catalog/CrudExternalCatalog.scala
```diff
@@ -0,0 +1,109 @@
...
+
+import org.apache.flink.table.api._
+
+/**
+  * The CrudExternalCatalog provides methods to create, drop, and alter (sub-)catalogs or tables.
+  *
+  * @deprecated use [[Catalog]] instead.
+  */
+@deprecated
+trait CrudExternalCatalog extends ExternalCatalog {
```
dawidwys: Let's not copy those classes. I think we should either move them to common module or leave them in the flink planner and throw exception in blink-planner if they are used. For this PR I would do the latter.

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/operations/RichTableSourceQueryOperation.java
```diff
@@ -0,0 +1,67 @@
...
+import java.util.List;
+import java.util.Map;
+
+/**
+ * A {@link TableSourceQueryOperation} with {@link FlinkStatistic} and qualifiedName.
+ */
+@Internal
+public class RichTableSourceQueryOperation<T> extends TableSourceQueryOperation<T> {
+	private final FlinkStatistic statistic;
+	private List<String> qualifiedName;
```
dawidwys: Why do we need the qualifiedName? Where does the qualified name comes from? The purpose of having TableSourceQueryOperation/DataStreamQueryOperation to not have a connection to the catalog, as they are created when e.g. `fromTableSource(TableSource)` or `fromDataStream(DataStream)` which is an inline operation.

If I understand it correctly this is needed for your tests to have a deterministic naming. To be honest I think this shows a problem in the tests infrastructure. As a workaround though I would change the `List<String> qualifiedName` to `String uniqueId`.

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/plan/QueryOperationConverter.java
```diff
@@ -130,8 +138,39 @@ public RelNode visit(QueryOperation other) {
 		}
 
 		@Override
-		public <U> RelNode visit(TableSourceQueryOperation<U> tableSourceTable) {
-			throw new UnsupportedOperationException("Unsupported now");
+		public <U> RelNode visit(TableSourceQueryOperation<U> tableSourceOperation) {
+			TableSource<?> tableSource = tableSourceOperation.getTableSource();
+			boolean isBatch = tableSourceOperation.isBatch();
```
dawidwys: This will never be true in Blink. You should use the flag on `TableSource`. The `isBatch` tells if the `TableSource` is `StreamTableSource` or `BatchTableSource`.

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/plan/QueryOperationConverter.java
```diff
@@ -130,8 +138,39 @@ public RelNode visit(QueryOperation other) {
...
+				relBuilder.getRelOptSchema(),
+				tableSourceTable.getRowType(relBuilder.getTypeFactory()),
+				names,
+				tableSourceTable);
+
+			// when LogicalTableScan converts a table scan with QueryOperationCatalogViewTable to a new table scan
+			// with TableSourceTable, both two table scans have same digest due to both two tables have
+			// exactly the same table name. so use EnumerableTableScan instead of LogicalTableScan to
+			// make sure two table scans have different digest.
+			if (createEnumerableTableScan) {
```
dawidwys: If I understand this correctly this is the problem of introducing the `qualifiedName` in `RichTableSourceQueryOperation`, right? Can we not introduce hacks in production code to make the tests work?

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/operations/RichTableSourceQueryOperation.java
godfreyhe: the `qualifiedName` is used for creating `FlinkRelOptTable` in `QueryOperationConverter` class, and the plan could show given table name while registering table source.
the qualified name is from `DatabaseCalciteSchema`:
`
			if (table instanceof QueryOperationCatalogView) {
				QueryOperationCatalogView view = (QueryOperationCatalogView) table;
				QueryOperation operation = view.getQueryOperation();
				if (operation instanceof DataStreamQueryOperation) {
					List<String> qualifiedName = Arrays.asList(catalogName, databaseName, tableName);
					((DataStreamQueryOperation) operation).setQualifiedName(qualifiedName);
				} else if (operation instanceof RichTableSourceQueryOperation) {
					List<String> qualifiedName = Arrays.asList(catalogName, databaseName, tableName);
					((RichTableSourceQueryOperation) operation).setQualifiedName(qualifiedName);
				}
				return QueryOperationCatalogViewTable.createCalciteTable(view);
			} else if (table instanceof ConnectorCatalogTable) {
...
`
for unregister table source (`fromTableSource(TableSource)`) or data stream (`fromDataStream(DataStream)`), their name is "unregistered_XXX" which is same with flink planner.

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/plan/QueryOperationConverter.java
godfreyhe: in blink, we always use `StreamTableSource` and use the `isBounded` flag ? do i understand correctly ?

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/catalog/CrudExternalCatalog.scala
godfreyhe: sure

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/plan/QueryOperationConverter.java
godfreyhe: As above.

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/operations/RichTableSourceQueryOperation.java
dawidwys: What do you mean it is lost? The name there is a name of a view (`QueryOperationCatalogView` is a view). Views are expanded, so I think it totally makes sense that it is not present in the explain.

I would say the blink behavior is misleading as you are overwriting original name with a name of a view.

dawidwys: Ok, but I think I get your point as it changes the old behavior slightly when registered via: `registerDataStream(String name, DataStream)`. Let me think about it for a second.

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/plan/QueryOperationConverter.java
dawidwys: yes

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/operations/RichTableSourceQueryOperation.java
dawidwys: To be honest. I see pros & cons of that solution. 

My personal opinion is that `DataStreamQueryOperation` and `TableSourceQueryOperation` are inline operations and are always registered as views, thus expanded during optimization if accessed via catalog.

How much of an effort would it be to have a consistent behavior with flink planner for this PR? I would prefer to move the discussion what name/digest to use to a separate issue, if possible.

godfreyhe: about 7000+ test cases will be changed if blink-planner has a consistent behavior with flink planner.  My suggestion is: I add a `FIXME` about `qualifiedName`, and fix them in another issue.

dawidwys: I thought so :( I think this is a reasonable trade off for now.

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/plan/QueryOperationConverter.java
godfreyhe: There is a special case: `LookupableTableSource`, which extends `TableSource`. so the `isBatch` will also be used here if a `TableSourceQueryOperation` wraps a `LookupableTableSource`.

KurtYoung:
I will merge it after travis go green

KurtYoung:
The cache problem is happening again, I will verify this locally.

PR #8708, opened 2019-06-12
[FLINK-12816][streaming] set maxParallelism to 1 when DataStreamSource is not isParallel

XuPingyong:

Now when DataStreamSource is not parallel, only parallelism is set to 1, maxParallelism should be also set to 1.
 
As when SingleOutputStreamOperator is not  parallel, it set not only parallelism but also maxParallelism to 1

## Brief change log

  - set maxParallelism to 1 when DataStreamSource is not isParallel 

## Verifying this change

DataStreamTest

  - The serializers: ( no)
  - The S3 file system connector: (no )

## Documentation

### DIFF
--- a/flink-streaming-java/src/main/java/org/apache/flink/streaming/api/datastream/DataStreamSource.java
+++ b/flink-streaming-java/src/main/java/org/apache/flink/streaming/api/datastream/DataStreamSource.java
@@ -41,6 +41,7 @@ public class DataStreamSource<T> extends SingleOutputStreamOperator<T> {
 		this.isParallel = isParallel;
 		if (!isParallel) {
 			setParallelism(1);
+			setMaxParallelism(1);
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


XuPingyong:
@pnowojski  What do you think?

wuchong:
Thanks for the contribution @XuPingyong .  The change is simple and straightforward.

+1 from my side.

Btw, please keep the headers and lists markdown syntax to make the description easy to read.

pnowojski:
As you can see there are related test failures caused by this change. 

One coming from `libraries` is benign.

However this change brakes a savepoint restoration compatibility. As far as I know, there is some check somewhere, that ensures (for good reasons) that `maxParallelism` of the current job and of the savepoint from which we are restoring is the same. 

I'm pretty sure that we can not fix this that simply as this PR wants to do it. It also brings me a question, whether this change is worth the effort. Could you explain what exactly are you trying to improve/solve here?

XuPingyong:
@pnowojski  Thanks for your review. I have not considered the compatibility issue. I close it and think again.

PR #8714, opened 2019-06-12
[FLINK-12820][Connectors / Cassandra] Support ignoring writing nulls 

ozancicek:
Added support to ignore null fields when writing to c* for tuple types. 

## Brief change log
 - Added 'ignoreNullFields' config parameter which can be enabled
 - Unset nulls from bound statement.

## Verifying this change

 Added a test case to CassandraConnectorITCase testing partial column updates

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-connectors/flink-connector-cassandra/src/main/java/org/apache/flink/streaming/connectors/cassandra/CassandraSinkBaseConfig.java
+++ b/flink-connectors/flink-connector-cassandra/src/main/java/org/apache/flink/streaming/connectors/cassandra/CassandraSinkBaseConfig.java
@@ -39,7 +39,7 @@ public final class CassandraSinkBaseConfig implements Serializable  {
 	public static final Duration DEFAULT_MAX_CONCURRENT_REQUESTS_TIMEOUT = Duration.ofMillis(Long.MAX_VALUE);
 
 	/**
-	 * The default option to ignore null fields on insert. By default, {@code Boolean.FALSE}.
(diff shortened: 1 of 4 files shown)


flink-connectors/flink-connector-cassandra/src/main/java/org/apache/flink/streaming/connectors/cassandra/CassandraSinkBaseConfig.java
```diff
@@ -38,6 +38,12 @@
 	 */
 	public static final Duration DEFAULT_MAX_CONCURRENT_REQUESTS_TIMEOUT = Duration.ofMillis(Long.MAX_VALUE);
 
+	/**
+	 * The default option to ignore null fields on insert. By default, {@code Boolean.FALSE}.
```
Myasuka: should be `on insertion`

ozancicek: Thanks, fixed!

StephanEwen:

StephanEwen:
Overall approach looks good.

I would ask you to switch to use primitive booleans rather than boxed booleans, to be consistent with the approach in the remainder of Flink. This also avoid accidental nullability and late null pointer exceptions.

Otherwise looks good to me.

StephanEwen:
@flinkbot approve all

StephanEwen:
Looks good to me, merging this...

PR #8715, opened 2019-06-12
[FLINK-12821][table][cep] Fix the bug that fix time quantifier can not be the last element of a pattern

dianfu:

*This pull request fix the bug that fix time quantifier can not be the last element of a pattern. Exception will be thrown for pattern "a{2}". Actually this pattern is not greedy and we should support it.*

## Brief change log

  - *If the quantifier is fixed, we should not mark it as greedy.*

## Verifying this change

  - *Added tests in PatternTranslatorTest#testQuantifiers*

  - The serializers: (no)

## Documentation


dianfu:
@flinkbot attention @dawidwys

dawidwys:
Thank you for the fix @dianfu, merged.

dianfu:
@dawidwys thanks a lot for the review and commit. :)

PR #8720, opened 2019-06-13
[FLINK-12771][hive] Support ConnectorCatalogTable in HiveCatalog

bowenli86:

This PR adds support for operations of `ConnectorCatalogTable` in HiveCatalog.

## Brief change log

- added support for operations of `ConnectorCatalogTable` in catalog APIs in HiveCatalog
- developed `ConnectorCatalogTableTestBase` and two test impl for each catalog to test catalog on `ConnectorCatalogTable`

## Verifying this change

- base test class `ConnectorCatalogTableTestBase`
- `GenericInMemoryConnectorCatalogTableTest`  and `HiveConnectorCatalogTableTest` for each catalog impl

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalog.java
+++ b/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalog.java
@@ -328,9 +328,8 @@ public class HiveCatalog extends AbstractCatalog {
 		}
 
 		if (table instanceof ConnectorCatalogTable) {
-			ConnectorCatalogTable connectorTable = connectorTables.get(tablePath);
 
-			if (connectorTable != null && !tableExistsInHiveMetastore(tablePath)) {
+			if (tableExists(tablePath)) {
 				if (!ignoreIfExists) {
 					throw new TableAlreadyExistException(getName(), tablePath);
 				}
@@ -362,10 +361,15 @@ public class HiveCatalog extends AbstractCatalog {
 
 		ObjectPath newPath = new ObjectPath(tablePath.getDatabaseName(), newTableName);
 
-		ConnectorCatalogTable connectorTable = connectorTables.get(tablePath);
+		ConnectorCatalogTable connectorTable = connectorTables.remove(tablePath);
 
 		if (connectorTable != null) {
-			connectorTables.put(newPath, connectorTable);
+			if (connectorTables.containsKey(newPath)) {
+				throw new TableAlreadyExistException(getName(), newPath);
+			} else {
+				connectorTables.put(newPath, connectorTable);
+			}
+
 			return;
 		}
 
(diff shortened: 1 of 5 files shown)


dawidwys:
I don't think this feature makes any sense. This makes the `HiveCatalog` also an instance of `InMemoryCatalog`.

I think if a table cannot be converted to properties (as is the case for `ConnectorCatalogTable` an exception should be thrown. We should allow storing those kind of tables only in an in memory catalog.

bowenli86:
Hi @dawidwys , shall we move the discussion to the JIRA? I've added context there

bowenli86:

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalog.java
```diff
@@ -313,6 +327,20 @@ public void createTable(ObjectPath tablePath, CatalogBaseTable table, boolean ig
 			throw new DatabaseNotExistException(getName(), tablePath.getDatabaseName());
 		}
 
+		if (table instanceof ConnectorCatalogTable) {
+			ConnectorCatalogTable connectorTable = connectorTables.get(tablePath);
+
+			if (connectorTable != null && !tableExistsInHiveMetastore(tablePath)) {
```
xuefuz: Why cannot we just call tableExists()?

bowenli86: Good catch! Not sure why I did it that way.... Changed

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalog.java
```diff
@@ -332,11 +359,24 @@ public void renameTable(ObjectPath tablePath, String newTableName, boolean ignor
 		checkNotNull(tablePath, "tablePath cannot be null");
 		checkArgument(!StringUtils.isNullOrWhitespaceOnly(newTableName), "newTableName cannot be null or empty");
 
+		ObjectPath newPath = new ObjectPath(tablePath.getDatabaseName(), newTableName);
+
+		ConnectorCatalogTable connectorTable = connectorTables.remove(tablePath);
+
+		if (connectorTable != null) {
+			if (connectorTables.containsKey(newPath)) {
+				throw new TableAlreadyExistException(getName(), newPath);
```
xuefuz: If we get here because of this exception, the table is already dropped, which doesn't seem right.

xuefuz:
Also, since we are using a map to store the temp tables with combo of db name and table name as the key, I'm wondering what impact those database operations might have, such as drop a database.

bowenli86:
@xuefuz good idea to add tests for mixed operations, I'll add them. As I can think of for now, there should be no impact on db operations except dropping database needs to ensure it also doesn't have temp tables. Will add that too.

I'll hold on this PR a little bit util we reach consensus with Dawid

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalog.java
```diff
@@ -313,6 +327,19 @@ public void createTable(ObjectPath tablePath, CatalogBaseTable table, boolean ig
 			throw new DatabaseNotExistException(getName(), tablePath.getDatabaseName());
 		}
 
+		if (table instanceof ConnectorCatalogTable) {
+
+			if (tableExists(tablePath)) {
```
lirui-apache: I think we'll need some synchronization here if we want to achieve thread safety.

bowenli86:
I will close this PR for now as we cannot reach consensus of having this feature with the community

PR #8721, opened 2019-06-13
[FLINK-12823][datastream] PartitionTransformation supports DataExchan…

ifndef-SleePy:
…geMode property

* Since `StreamTransformation` would also support batch runner in 1.9, the result partition type of `StreamTransformation` should not be hard-coded with `PIPELINED_BOUNDED`
* We need to provide a way for upper level API upon `StreamTransformation` to configure the result partition type of edge

## Brief change log

* Expose a property `DataExchangeMode` of `PartitionTransformation` as an internal API of `StreamTransformation`
* Pass the `DataExchangeMode` to `StreamEdge`
* `StreamingJobGraphGenerator` chooses the appropriate result partition type based on `DataExchangeMode` of `StreamEdge`

## Verifying this change

* Add an unit test of `StreamingJobGraphGeneratorTest`

  - The serializers: no

## Documentation

### DIFF
--- a/flink-streaming-java/src/main/java/org/apache/flink/streaming/api/graph/StreamGraph.java
+++ b/flink-streaming-java/src/main/java/org/apache/flink/streaming/api/graph/StreamGraph.java
@@ -26,6 +26,7 @@ import org.apache.flink.api.common.typeinfo.TypeInformation;
 import org.apache.flink.api.common.typeutils.TypeSerializer;
 import org.apache.flink.api.java.functions.KeySelector;
 import org.apache.flink.api.java.tuple.Tuple2;
+import org.apache.flink.api.java.tuple.Tuple3;
 import org.apache.flink.api.java.typeutils.MissingTypeInfo;
 import org.apache.flink.optimizer.plan.StreamingPlan;
 import org.apache.flink.runtime.jobgraph.JobGraph;
@@ -35,6 +36,7 @@ import org.apache.flink.streaming.api.collector.selector.OutputSelector;
 import org.apache.flink.streaming.api.environment.CheckpointConfig;
 import org.apache.flink.streaming.api.environment.StreamExecutionEnvironment;
 import org.apache.flink.streaming.api.operators.StreamOperatorFactory;
+import org.apache.flink.streaming.api.transformations.ShuffleMode;
 import org.apache.flink.streaming.runtime.partitioner.ForwardPartitioner;
 import org.apache.flink.streaming.runtime.partitioner.RebalancePartitioner;
 import org.apache.flink.streaming.runtime.partitioner.StreamPartitioner;
@@ -86,7 +88,7 @@ public class StreamGraph extends StreamingPlan {
 	private Set<Integer> sinks;
 	private Map<Integer, Tuple2<Integer, List<String>>> virtualSelectNodes;
 	private Map<Integer, Tuple2<Integer, OutputTag>> virtualSideOutputNodes;
-	private Map<Integer, Tuple2<Integer, StreamPartitioner<?>>> virtualPartitionNodes;
+	private Map<Integer, Tuple3<Integer, StreamPartitioner<?>, ShuffleMode>> virtualPartitionNodes;
 
 	protected Map<Integer, String> vertexIDtoBrokerID;
 	protected Map<Integer, Long> vertexIDtoLoopTimeout;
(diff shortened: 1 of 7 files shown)


aljoscha:
I think we should not reuse the existing `DataExchangeMode` because it is a runtime class in the `flink-runtime` package. At some point we want to decouple the `StreamTransformations` and the `DataStream` API from runtime so we should not introduce more dependencies. I would suggest to add a new enum next to `PartitionTransformation` (in the same package) for this purpose, it should probably also be declared `@PublicEvolving`.

ifndef-SleePy:
Hi @aljoscha

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/graph/StreamGraph.java
```diff
@@ -428,7 +436,11 @@ private void addEdgeInternal(Integer upStreamVertexID,
 				}
 			}
 
-			StreamEdge edge = new StreamEdge(upstreamNode, downstreamNode, typeNumber, outputNames, partitioner, outputTag);
+			if (shuffleMode == null) {
```
aljoscha: Can it happen that `shuffleMode` is `null`? We should add a null check in `PartitionTransform` and mark the field `@Nonnull`.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/transformations/PartitionTransformation.java
```diff
@@ -69,6 +88,15 @@ public PartitionTransformation(StreamTransformation<T> input, StreamPartitioner<
 		return partitioner;
 	}
 
+	/**
```
aljoscha: I don't like these redundant descriptions to much, this can be simplified to:
```
/**
 * Returns the {@link ShuffleMode} of this {@link PartitionTransformation}.
 */
```

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/transformations/PartitionTransformation.java
```diff
@@ -79,6 +107,6 @@ public PartitionTransformation(StreamTransformation<T> input, StreamPartitioner<
 
 	@Override
 	public final void setChainingStrategy(ChainingStrategy strategy) {
-		throw new UnsupportedOperationException("Cannot set chaining strategy on Union Transformation.");
+		throw new UnsupportedOperationException("Cannot set chaining strategy on Partition Transformation.");
```
aljoscha: Nice catch! 😅

flink-streaming-java/src/test/java/org/apache/flink/streaming/api/graph/StreamingJobGraphGeneratorTest.java
```diff
@@ -133,8 +136,8 @@ public void testChainStartEndSetting() throws Exception {
 				public Integer map(Integer value) throws Exception {
 					return value;
 				}
-			})
-			.print();
+			}).setParallelism(2)
```
aljoscha: Is this change needed? It seems to be an unrelated test.

ifndef-SleePy: It's an unrelated change in a sense.
I found it when adding unit test. It's a potential unstable case since it depends on the `StreamExecutionEnvironment.defaultLocalParallelism` which must be not 1 in this case.
However I just found someone has fix it yesterday. What a coincident! I can remove it from this PR.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/graph/StreamGraph.java
ifndef-SleePy: Yes, it can be null here. Because there might be no `PartitionTransformation` if user do not specify explicitly. And then `partitioner` and `shuffleMode` would be null here.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/transformations/PartitionTransformation.java
ifndef-SleePy: Done

ifndef-SleePy:
Hi @aljoscha 
I have rebased master and addressed comments.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/graph/StreamGraph.java
aljoscha: Ok, makes sense

aljoscha:
Thanks! I think I can merge once Travis is green. 👌

ifndef-SleePy:
Travis failed, there is am importing error in python module. I don't think it is caused by this PR. I will try to rebase master and trigger travis checking again.

ifndef-SleePy:
Python module is fine but connectors module failed this time on Travis checking. It seems a travis relevant issue. I have filed a JIRA issue for this, https://issues.apache.org/jira/browse/FLINK-12866
Will trigger checking again through close and reopening, hope it works.

aljoscha:
I merged this. Thanks for working on this with me! 👍

PR #8723, opened 2019-06-13
[FLINK-12827][table-planner] Extract Optimizer from TableEnvImpl

dawidwys:

* Extract `Optimizer` from `Batch/StreamTableEnvImpl` to better separate concerns.

## Verifying this change

## Documentation

### DIFF
--- a/flink-table/flink-table-planner/src/main/scala/org/apache/flink/table/plan/BatchOptimizer.scala
+++ b/flink-table/flink-table-planner/src/main/scala/org/apache/flink/table/plan/BatchOptimizer.scala
@@ -27,7 +27,8 @@ import org.apache.flink.table.plan.rules.FlinkRuleSets
 import org.apache.flink.table.planner.PlanningConfigurationBuilder
 
 /**
-  * An [[Optimizer]] that can be used for optimizing a batch plan.
+  * An [[Optimizer]] that can be used for optimizing a batch plan. Should be used to create an
+  * optimized tree from a logical input tree.
   *
   * @param calciteConfig provider for [[CalciteConfig]]. It is a provider because the
   *                      [[org.apache.flink.table.api.TableConfig]] in a
@@ -56,12 +57,12 @@ class BatchOptimizer(
   }
 
   /**
-    * Returns the built-in normalization rules that are defined by the environment.
+    * Returns the built-in normalization rules that are defined by the optimizer.
     */
   protected def getBuiltInNormRuleSet: RuleSet = FlinkRuleSets.DATASET_NORM_RULES
 
   /**
-    * Returns the built-in optimization rules that are defined by the environment.
+    * Returns the built-in optimization rules that are defined by the optimizer.
     */
   protected def getBuiltInPhysicalOptRuleSet: RuleSet = FlinkRuleSets.DATASET_OPT_RULES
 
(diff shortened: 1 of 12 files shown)


flink-table/flink-table-planner/src/main/scala/org/apache/flink/table/plan/Optimizer.scala
```diff
@@ -0,0 +1,264 @@
...
+import org.apache.flink.table.plan.nodes.FlinkConventions
+import org.apache.flink.table.plan.rules.FlinkRuleSets
+import org.apache.flink.table.planner.PlanningConfigurationBuilder
+
+import scala.collection.JavaConverters._
+
+/**
+  * Common functionalities for both [[StreamOptimizer]] and [[BatchOptimizer]]. An [[Optimizer]]
+  * should be used to create an optimized tree from a logical input tree.
+  * See [[StreamOptimizer.optimize]]and [[BatchOptimizer.optimize]]
```
aljoscha: nit: missing space

flink-table/flink-table-planner/src/main/scala/org/apache/flink/table/plan/StreamOptimizer.scala
```diff
@@ -0,0 +1,125 @@
...
+    */
+  def optimize(
+    relNode: RelNode,
+    updatesAsRetraction: Boolean,
+    relBuilder: RelBuilder): RelNode = {
+    val convSubQueryPlan = optimizeConvertSubQueries(relNode)
+    val expandedPlan = optimizeExpandPlan(convSubQueryPlan)
+    val decorPlan = RelDecorrelator.decorrelateQuery(expandedPlan, relBuilder)
+    val planWithMaterializedTimeAttributes =
+      RelTimeIndicatorConverter.convert(decorPlan, relBuilder.getRexBuilder)
```
aljoscha: Just something I noticed: documentation is a bit sparse and it's hard for me to find out what a `RelTimeIndicatorConverter` actually does, because it has no class-level (or Object-level) documentation. 😅 

Not for this PR to fix, though...

flink-table/flink-table-planner/src/main/scala/org/apache/flink/table/plan/BatchOptimizer.scala
```diff
@@ -0,0 +1,68 @@
...
+    val convSubQueryPlan = optimizeConvertSubQueries(relNode)
+    val expandedPlan = optimizeExpandPlan(convSubQueryPlan)
+    val decorPlan = RelDecorrelator.decorrelateQuery(expandedPlan)
+    val normalizedPlan = optimizeNormalizeLogicalPlan(decorPlan)
+    val logicalPlan = optimizeLogicalPlan(normalizedPlan)
+    optimizePhysicalPlan(logicalPlan, FlinkConventions.DATASET)
+  }
+
+  /**
+    * Returns the built-in normalization rules that are defined by the environment.
```
aljoscha: I guess these are not defined by the environment anymore but by the optimiser. Probably the same in the `StreamOptimizer`.

dawidwys:
As for the `Law of Demeter` I totally agree. The problem is there are hundreds of tests that depend on the monolithic design of the `TableEnvironment`. We would have to rewrite most of them, which would be rather tricky and definitely not quick.

flink-table/flink-table-planner/src/main/scala/org/apache/flink/table/plan/StreamOptimizer.scala
dawidwys: Totally agree. We try to add proper javadocs for the newly introduced interfaces/classes.

PR #8729, opened 2019-06-13
[FLINK-12831][table-planner][table-api-java] Split FunctionCatalog into Flink & Calcite specific parts

dawidwys:

This PR splits FunctionCatalog into Flink and Calcite specific parts. This is a necessary step to move the `TableEnvironment` to api module. 

## Brief change log
- ported functions to validate & extract types from user function (ported from UserDefinedFunctionUtils)
- ported & moved `FunctionCatalog` to api-java module
- Introduced `org.apache.flink.table.catalog.CatalogOperatorTable` as a thin adapter between `FunctionCatalog` & `SqlOperatorTable`

## Verifying this change

## Documentation

### DIFF
--- a/flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/functions/UserFunctionsTypeHelper.java
+++ b/flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/functions/UserFunctionsTypeHelper.java
@@ -1,0 +1,176 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.table.functions;
+
+import org.apache.flink.annotation.Internal;
+import org.apache.flink.api.common.typeinfo.TypeInformation;
+import org.apache.flink.api.java.typeutils.TypeExtractor;
+import org.apache.flink.table.api.ValidationException;
+import org.apache.flink.util.InstantiationUtil;
+
+import java.util.Arrays;
+
+/**
+ * Utility methods for validation and extraction of types during function registration in a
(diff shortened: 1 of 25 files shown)


flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/functions/UserFunctionsTypeHelper.java
```diff
@@ -0,0 +1,176 @@
...
+	/**
+	 * Check whether this is a Scala object. It is forbidden to use {@link TableFunction} implemented
+	 * by a Scala object, since concurrent risks.
+	 */
+	public static void validateNotSingleton(Class<?> clazz) {
+		// TODO it is not a good way to check singleton. Maybe improve it further.
+		if (Arrays.stream(clazz.getFields()).anyMatch(f -> f.getName().equals("MODULE$"))) {
+			throw new ValidationException(String.format(
+				"TableFunction implemented by class %s " +
+					"is a Scala object, it is forbidden since concurrent risks.", clazz.getCanonicalName()));
```
aljoscha: I know it's not in the scope of the PR but this should be `this is forbidden because of concurrency problems when using them`. Or something like it.

flink-table/flink-table-planner/src/main/java/org/apache/flink/table/catalog/CatalogOperatorTable.java
```diff
@@ -0,0 +1,142 @@
...
+import org.apache.calcite.sql.SqlSyntax;
+
+import java.util.List;
+import java.util.Optional;
+
+/**
+ * Thin adapter between {@link SqlOperatorTable} and {@link FunctionCatalog}.
+ */
+@Internal
+public class CatalogOperatorTable implements SqlOperatorTable {
```
aljoscha: I think I don't really have a better name but maybe making it obvious that this is a bridge or adapter would be useful.

dawidwys: This sort of inline with naming of similar pattern for `CatalogManagerCalciteSchema` which bridges/adapts between `CatalogManager` & Calcite's `Schema`.

flink-table/flink-table-planner/src/main/java/org/apache/flink/table/catalog/FunctionCatalogOperatorTable.java
```diff
@@ -0,0 +1,139 @@
...
+
+		String name = opName.getSimple();
+		Optional<FunctionDefinition> candidateFunction = functionCatalog.lookupFunction(name);
+
+		candidateFunction.flatMap(functionDefinition ->
+			convertToSqlFunction(category, name, functionDefinition)
+		).ifPresent(operatorList::add);
+	}
+
+	private boolean isUserFunction(SqlFunctionCategory category) {
```
docete: Use 'isNotUserFunction' for more clear and less misleading

dawidwys: Good catch! Thank you!

PR #8731, opened 2019-06-13
[FLINK-11878][runtime] Implement the runtime handling of BoundedOneInput and BoundedMultiInput

sunhaibotb:

This pull request implements the runtime handling of the BoundedOneInput and BoundedMultiInput 
 interfaces. This will give the bounded stream operators an opportunity to process certain logic when the EndOfInput  notification arrives.  For example,  the hash join operator knows when the read to the build side has finished, then it can finish the build phase and prepare for the probe phase.

## Brief change log

  - Change `xInputProcessor` (`StreamInputProcessor`, `StreamTwoInputProcessor` and `StreamTwoInputSelecdtableProcessor`) to add the `endInput()` method call of the header operator on the operator chain.
  - Change `xStreamTask` (`SourceStreamTask`, `OneInputStreamTask`, `TwoInputStreamTask` and `TwoInputSelectableStreamTask`) to add the `endInput()` method calls for other non-header operators on the operator chain.

## Verifying this change

  - Added test that validates that `SourceStreamTask` correctly handles `EndOfInput` notifications.
  - Added test that validates that `OneInputStreamTask` correctly handles `EndOfInput` notifications.
  - Added test that validates that `TwoInputStreamTask` correctly handles `EndOfInput` notifications.
  - Added test that validates that `TwoInputSelectableStreamTask` correctly handles `EndOfInput` notifications.

## Documentation

### DIFF
--- a/flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/StreamSource.java
+++ b/flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/StreamSource.java
@@ -27,6 +27,7 @@ import org.apache.flink.streaming.api.watermark.Watermark;
 import org.apache.flink.streaming.runtime.streamrecord.LatencyMarker;
 import org.apache.flink.streaming.runtime.streamrecord.StreamRecord;
 import org.apache.flink.streaming.runtime.streamstatus.StreamStatusMaintainer;
+import org.apache.flink.streaming.runtime.tasks.OperatorChain;
 import org.apache.flink.streaming.runtime.tasks.ProcessingTimeCallback;
 import org.apache.flink.streaming.runtime.tasks.ProcessingTimeService;
 
@@ -55,13 +56,17 @@ public class StreamSource<OUT, SRC extends SourceFunction<OUT>> extends Abstract
 		this.chainingStrategy = ChainingStrategy.HEAD;
 	}
 
-	public void run(final Object lockingObject, final StreamStatusMaintainer streamStatusMaintainer) throws Exception {
-		run(lockingObject, streamStatusMaintainer, output);
+	public void run(final Object lockingObject,
+			final StreamStatusMaintainer streamStatusMaintainer,
+			final OperatorChain<?, ?> operatorChain) throws Exception {
+
+		run(lockingObject, streamStatusMaintainer, output, operatorChain);
 	}
 
 	public void run(final Object lockingObject,
 			final StreamStatusMaintainer streamStatusMaintainer,
-			final Output<StreamRecord<OUT>> collector) throws Exception {
+			final Output<StreamRecord<OUT>> collector,
+			final OperatorChain<?, ?> operatorChain) throws Exception {
 
 		final TimeCharacteristic timeCharacteristic = getOperatorConfig().getTimeCharacteristic();
 
@@ -96,9 +101,14 @@ public class StreamSource<OUT, SRC extends SourceFunction<OUT>> extends Abstract
 
 			// if we get here, then the user function either exited after being done (finite source)
 			// or the function was canceled or stopped. For the finite source case, we should emit
-			// a final watermark that indicates that we reached the end of event-time
+			// a final watermark that indicates that we reached the end of event-time, and end inputs
+			// of the operator chain
 			if (!isCanceledOrStopped()) {
 				advanceToEndOfEventTime();
+
+				synchronized (lockingObject) {
+					operatorChain.endInput(1);
+				}
 			}
 		} finally {
 			// make sure that the context is closed in any case
--- a/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamInputProcessor.java
+++ b/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamInputProcessor.java
@@ -36,6 +36,7 @@ import org.apache.flink.streaming.runtime.streamrecord.StreamRecord;
 import org.apache.flink.streaming.runtime.streamstatus.StatusWatermarkValve;
 import org.apache.flink.streaming.runtime.streamstatus.StreamStatus;
 import org.apache.flink.streaming.runtime.streamstatus.StreamStatusMaintainer;
+import org.apache.flink.streaming.runtime.tasks.OperatorChain;
 import org.apache.flink.streaming.runtime.tasks.StreamTask;
 
 import org.slf4j.Logger;
@@ -69,6 +70,8 @@ public class StreamInputProcessor<IN> {
 
 	private final Object lock;
 
+	private final OperatorChain<?, ?> operatorChain;
+
 	// ---------------- Status and Watermark Valve ------------------
 
 	/** Valve that controls how watermarks and stream statuses are forwarded. */
@@ -96,7 +99,8 @@ public class StreamInputProcessor<IN> {
 			OneInputStreamOperator<IN, ?> streamOperator,
 			TaskIOMetricGroup metrics,
 			WatermarkGauge watermarkGauge,
-			String taskName) throws IOException {
+			String taskName,
+			OperatorChain<?, ?> operatorChain) throws IOException {
 
 		InputGate inputGate = InputGateUtil.createInputGate(inputGates);
 
@@ -120,6 +124,8 @@ public class StreamInputProcessor<IN> {
 
 		this.watermarkGauge = watermarkGauge;
 		metrics.gauge("checkpointAlignmentTime", barrierHandler::getAlignmentDurationNanos);
+
+		this.operatorChain = checkNotNull(operatorChain);
 	}
 
 	public boolean processInput() throws Exception {
@@ -128,7 +134,7 @@ public class StreamInputProcessor<IN> {
 		StreamElement recordOrMark = input.pollNextNullable();
 		if (recordOrMark == null) {
 			input.isAvailable().get();
-			return !input.isFinished();
+			return !checkFinished();
 		}
 		int channel = input.getLastChannel();
 		checkState(channel != StreamTaskInput.UNSPECIFIED);
@@ -172,6 +178,16 @@ public class StreamInputProcessor<IN> {
 				numRecordsIn = new SimpleCounter();
 			}
 		}
+	}
+
+	private boolean checkFinished() throws Exception {
+		boolean isFinished = input.isFinished();
+		if (isFinished) {
+			synchronized (lock) {
+				operatorChain.endInput(1);
+			}
+		}
+		return isFinished;
 	}
 
 	public void cleanup() throws Exception {
--- a/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/OperatorChain.java
+++ b/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/OperatorChain.java
@@ -38,10 +38,14 @@ import org.apache.flink.streaming.api.collector.selector.DirectedOutput;
 import org.apache.flink.streaming.api.collector.selector.OutputSelector;
 import org.apache.flink.streaming.api.graph.StreamConfig;
 import org.apache.flink.streaming.api.graph.StreamEdge;
+import org.apache.flink.streaming.api.operators.BoundedMultiInput;
+import org.apache.flink.streaming.api.operators.BoundedOneInput;
+import org.apache.flink.streaming.api.operators.InputSelection;
 import org.apache.flink.streaming.api.operators.OneInputStreamOperator;
 import org.apache.flink.streaming.api.operators.Output;
 import org.apache.flink.streaming.api.operators.StreamOperator;
 import org.apache.flink.streaming.api.operators.StreamOperatorFactory;
+import org.apache.flink.streaming.api.operators.TwoInputStreamOperator;
 import org.apache.flink.streaming.api.watermark.Watermark;
 import org.apache.flink.streaming.runtime.io.RecordWriterOutput;
 import org.apache.flink.streaming.runtime.metrics.WatermarkGauge;
@@ -79,6 +83,9 @@ public class OperatorChain<OUT, OP extends StreamOperator<OUT>> implements Strea
 
 	private static final Logger LOG = LoggerFactory.getLogger(OperatorChain.class);
 
+	/**
+	 * Stores all operators on this chain in reverse order.
+	 */
 	private final StreamOperator<?>[] allOperators;
 
 	private final RecordWriterOutput<?>[] streamOutputs;
@@ -94,6 +101,9 @@ public class OperatorChain<OUT, OP extends StreamOperator<OUT>> implements Strea
 	 * this value is {@link StreamStatus#IDLE}.
 	 */
 	private StreamStatus streamStatus = StreamStatus.ACTIVE;
+
+	/** The flag that tracks finished inputs. */
+	private InputSelection finishedInputs = new InputSelection.Builder().build();
 
 	public OperatorChain(
 			StreamTask<OUT, OP> containingTask,
@@ -221,6 +231,50 @@ public class OperatorChain<OUT, OP extends StreamOperator<OUT>> implements Strea
 			final StreamOperator<?> op = operators[i];
 			if (op != null) {
 				op.prepareSnapshotPreBarrier(checkpointId);
+			}
+		}
+	}
+
+	/**
+	 * Ends an input (specified by {@code inputId}) of the {@link StreamTask}. The {@code inputId}
+	 * is numbered starting from 1, and `1` indicates the first input.
+	 *
+	 * @param inputId The ID of the input.
+	 * @throws Exception if some exception happens in the endInput function of an operator.
+	 */
+	public void endInput(int inputId) throws Exception {
+		if (finishedInputs.areAllInputsSelected()) {
+			return;
+		}
+
+		if (headOperator instanceof TwoInputStreamOperator) {
+			if (finishedInputs.isInputSelected(inputId)) {
+				return;
+			}
+
+			if (headOperator instanceof BoundedMultiInput) {
+				((BoundedMultiInput) headOperator).endInput(inputId);
+			}
+
+			finishedInputs = InputSelection.Builder
+				.from(finishedInputs)
+				.select(finishedInputs.getInputMask() == 0 ? inputId : -1)
+				.build();
+		} else {
+			// here, the head operator is a stream source or an one-input stream operator,
+			// so all inputs are finished
+			finishedInputs = new InputSelection.Builder()
+				.select(-1)
+				.build();
+		}
+
+		if (finishedInputs.areAllInputsSelected()) {
+			// executing #endInput() happens from head to tail operator in the chain
+			for (int i = allOperators.length - 1; i >= 0; i--) {
+				StreamOperator<?> operator = allOperators[i];
+				if (operator instanceof BoundedOneInput) {
+					((BoundedOneInput) operator).endInput();
+				}
 			}
 		}
 	}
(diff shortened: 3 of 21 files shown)


flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/OneInputStreamTaskTest.java
```diff
@@ -597,6 +603,54 @@ public void testQuiesceTimerServiceAfterOpClose() throws Exception {
...
+			.finish();
+
+		ConcurrentLinkedQueue<Object> expectedOutput = new ConcurrentLinkedQueue<>();
+
+		testHarness.invoke();
+		testHarness.waitForTaskRunning();
+
+		TestBoundedOneInputOperator headOperator = (TestBoundedOneInputOperator) testHarness.getTask().headOperator;
+
+		testHarness.processElement(new StreamRecord<>("Hello-0-0"), 0, 0);
```
pnowojski: Why are you passing `inputGate` and `channel` parameters? does it matter from where did the element come from?

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/OneInputStreamTaskTest.java
```diff
@@ -597,6 +603,54 @@ public void testQuiesceTimerServiceAfterOpClose() throws Exception {
...
+		testHarness.endInput(0,  0);
+		testHarness.processElement(new StreamRecord<>("Hello-0-1"), 0, 1);
+		testHarness.endInput(0,  1);
+
+		headOperator.waitOutputRecords(2);
+
+		testHarness.processElement(new StreamRecord<>("Hello-1-0"), 1, 0);
+		testHarness.processElement(new StreamRecord<>("Hello-1-1"), 1, 1);
```
pnowojski: Wouldn't it be better to reorder this to:
```
		testHarness.processElement(new StreamRecord<>("Hello-1-0"), 1, 0);
		testHarness.endInput(1,  0);
		testHarness.processElement(new StreamRecord<>("Hello-1-1"), 1, 1);
		testHarness.endInput(1,  1);
```
? As it is now, we are not sure after which `endInput()` call operators received `endInput()` notification.

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/OneInputStreamTaskTest.java
```diff
@@ -1011,5 +1065,70 @@ protected void handleWatermark(Watermark mark) {
...
+		public void endInput() {
+			output.collect(new StreamRecord<>("[" + name + "]: Bye"));
+			outputRecords++;
+
+			synchronized (this) {
+				this.notifyAll();
+			}
+		}
+
+		public void waitOutputRecords(int expectedRecords) throws Exception {
```
pnowojski: Why do you need this method?

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/OneInputStreamTaskTest.java
```diff
@@ -1011,5 +1065,70 @@ protected void handleWatermark(Watermark mark) {
 			output.emitWatermark(mark);
 		}
 	}
+
+	/**
+	 * Uses to test handling the EndOfInput notification.
+	 */
+	static class TestBoundedOneInputOperator extends AbstractStreamOperator<String>
```
pnowojski: As this is shared across three different tests, I would extract it to a separate file (move to upper level).

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/SourceStreamTaskTest.java
```diff
@@ -291,5 +325,34 @@ public void run(SourceContext<String> ctx) throws Exception {
 		@Override
 		public void cancel() {}
 	}
+
+	private static class TestBoundedSource implements SourceFunction<String> {
```
pnowojski: I'm pretty sure there is already something like that defined somewhere. For example:
`org.apache.flink.test.streaming.runtime.StreamTaskSelectiveReadingITCase.TestStringSource`
`org.apache.flink.streaming.runtime.tasks.SourceStreamTaskTest.MockSource`
or even production class
`org.apache.flink.streaming.api.functions.source.FromElementsFunction`

Please re-use one of them, preferably `FromElementsFunction`.

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/TwoInputStreamTaskTest.java
```diff
@@ -596,6 +602,66 @@ public TaskMetricGroup getMetricGroup() {
 		testHarness.waitForTaskCompletion();
 	}
 
+	@Test
+	public void testHandlingEndOfInput() throws Exception {
+		final TwoInputStreamTaskTestHarness<String, String, String> testHarness = new TwoInputStreamTaskTestHarness<>(
+			isInputSelectable ? TwoInputSelectableStreamTask::new : TwoInputStreamTask::new,
+			3, 2, new int[] {1, 2, 2},
```
pnowojski: nit: one parameter per line

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/TwoInputStreamTaskTest.java
```diff
@@ -596,6 +602,66 @@ public TaskMetricGroup getMetricGroup() {
...
+			.finish();
+
+		ConcurrentLinkedQueue<Object> expectedOutput = new ConcurrentLinkedQueue<Object>();
+
+		testHarness.invoke();
+		testHarness.waitForTaskRunning();
+
+		TestBoundedTwoInputOperator headOperator = (TestBoundedTwoInputOperator) testHarness.getTask().headOperator;
+
+		testHarness.processElement(new StreamRecord<>("Hello-1"), 0, 0);
```
pnowojski: ditto: Do you need to pass those extra arguments like `0, 0` or `0, 1`?

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/TwoInputStreamTaskTest.java
```diff
@@ -671,5 +737,100 @@ public InputSelection nextSelection() {
...
+			} else {
+				outputRecords2++;
+			}
+
+			synchronized (this) {
+				this.notifyAll();
+			}
+		}
+
+		public void waitOutputRecords(int inputId, int expectedRecords) throws Exception {
```
pnowojski: ditto: why do you need this method?

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamInputProcessor.java
```diff
@@ -125,10 +126,14 @@ public StreamInputProcessor(
 	public boolean processInput() throws Exception {
 		initializeNumRecordsIn();
 
+		if (input.isFinished()) {
```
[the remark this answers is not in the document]
sunhaibotb: > To call endInput only once?

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/OperatorChain.java
```diff
@@ -225,6 +226,21 @@ public void prepareSnapshotPreBarrier(long checkpointId) throws Exception {
...
 	}
 
+	/**
+	 * Starting from the second operator, go forward through the operator chain to notify
+	 * each operator that its input has ended.
+	 *
+	 * @throws Exception if some exception happens in the endInput function of an operator.
+	 */
+	public void endOperatorInputs() throws Exception {
+		for (int i = allOperators.length - 2; i >= 0; i--) {
```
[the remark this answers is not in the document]
sunhaibotb: The `endInput()` of the head operator has been called in `xInputProcessor`. It's called in `xInputProcessor` because the `endInput()` of the two input operator must be called immediately when one of both inputs arrives the end.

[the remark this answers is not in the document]
sunhaibotb: I'll change the method name `endOperatorInputs` to the appropriate one (for example, `endNonHeadOperatorInputs`).

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/OperatorChain.java
```diff
@@ -225,6 +226,21 @@ public void prepareSnapshotPreBarrier(long checkpointId) throws Exception {
 		}
 	}
```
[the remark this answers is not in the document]
sunhaibotb: I will add a comment for it.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamInputProcessor.java
```diff
@@ -174,6 +179,18 @@ private void initializeNumRecordsIn() {
 		}
 	}
 
+	private boolean checkFinished() throws Exception {
+		boolean isFinished = input.isFinished();
+		if (isFinished) {
+			if (streamOperator instanceof BoundedOneInput) {
```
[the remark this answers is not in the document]
sunhaibotb: > Why here are you endInput for the head operator and inside the task for the rest of the operators?

The reason is that we don't want `xInputProcessor` to see `OperatorChain`. If we change this point and let `xInputProcessor` see `OperatorChain`, we can do that like you said.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/OperatorChain.java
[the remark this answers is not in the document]
sunhaibotb: > I think you could simplify the code if you iterated through all operators in OperatorChain (not starting from the second one), check each one of them if there are instanceof BoundedOneInput and end them if they are.
>
>This would work for:
>

> - OneInputStreamTask - because all of them are/can be BoundedOneInput
> - SourceStreamTask - because head will not be BoundedOneInput
> - TwoInputStreamTask - because head will not be BoundedOneInput

You are right. However, if chaining non-header two input operator is allowed in the future, it will have problems.

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/OneInputStreamTaskTest.java
sunhaibotb: This test was written a long time ago, when they was needed. Now `inputGate` and `channel` parameters are not needed, and I will delete them.

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/TwoInputStreamTaskTest.java
sunhaibotb: `StreamTwoInputProcessor` has not been changed to use `StreamTaskNetworkInput`, whether the inputs are finished is judged by two added  BitSet members, so `inputGate` and `channel` parameters need to be passed to verify the correctness of this logic.

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/OneInputStreamTaskTest.java
sunhaibotb: This test was written a long time ago, when it was needed. Now not needed, and I will remove it.

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/TwoInputStreamTaskTest.java
sunhaibotb: This ensures that the output is strictly ordered, to verify whether `endInput()` is called immediately when `input1` arrives the end.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/OperatorChain.java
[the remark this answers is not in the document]
sunhaibotb: Maybe we can use the following structure?

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamInputProcessor.java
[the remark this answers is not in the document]
sunhaibotb: If let `xInputProcessor` see non-head operators, we can use the following structure. (`OperatorChain` is still invisible to `XInputProcessor`)

```
============== xInputProcessor.java ==============

private boolean checkFinished() throws Exception {
                        ......
    // invoke endInput(...) of the head operator
    ......
   // invoke endInput(...) of other operators
   streamTask.endNonHeadOperatorInputs()
}

============== xStreamTask.java ==============

public void endNonHeadOperatorInputs throws Exception {
    operatorChain.endNonHeadOperatorInputs();
}

============== OperatorChain.java ==============

public void endNonHeadOperatorInputs throws Exception {
                       ......
}

```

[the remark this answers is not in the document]
pnowojski: It would be better to have this logic in one place and not distributed across two different classes (head is being marked as ended in the processor, while non heads are marked via task).

Also what you proposed adds an (additional) cyclic dependency between Task and Processor, which is also not good.

Maybe the best approach would be to encapsulate this whole logic in `OperatorChain` class and pass it to the processors? For example by adding `OperatorChain#endInput(int inputId)` method, which would:
- if the head is `TwoInput`, it would track which inputs have ended. If all of them, then it would propagate the `endInput` call down the operator chain.
- if the head is `OneInput`, it would just end all of the inputs

For the future, we can not be sure how will be `TwoInputStreamOperator` chaining implemented, but if we keep `OperatorChain` class, the support for propagating `endInput` would need to be adjusted only there.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/StreamSource.java
```diff
@@ -143,7 +143,7 @@ protected void markCanceledOrStopped() {
 	 * Checks whether the source has been canceled or stopped.
 	 * @return True, if the source is canceled or stopped, false is not.
 	 */
-	protected boolean isCanceledOrStopped() {
+	public boolean isCanceledOrStopped() {
```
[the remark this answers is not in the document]
sunhaibotb: I will add a test for the case of cancel.

sunhaibotb:
The PR has been update.  Please review it, thank you very much. @pnowojski

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/operators/StreamSourceOperatorLatencyMetricsTest.java
```diff
@@ -163,7 +164,7 @@ private void testLatencyMarkEmission(int numberLatencyMarkers, OperatorSetupOper
 		operatorSetup.setupSourceOperator(operator, testProcessingTimeService);
 
 		// run and wait to be stopped
-		operator.run(new Object(), mock(StreamStatusMaintainer.class), new CollectorOutput<Long>(output));
+		operator.run(new Object(), mock(StreamStatusMaintainer.class), new CollectorOutput<Long>(output), mock(OperatorChain.class));
```
pnowojski: Please do not introduce new mockito mocked classes. As usually either:
- please instantiate a real `OpeartorChain`
- please instantiate a real `OpeartorChain` with maybe mocked content `RecordWriters `
- implement `OperatorChain` mock, maybe a `NoOpOperatorChain`

Keep in mind that `mocked` I mean a proper mocks, not mockito.

(You can read about the reasoning behind not using mockito in the proposed code style guideline)

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/TwoInputStreamTaskTest.java
pnowojski: Is there a reason why you can not use `testHarness.waitForInputProcessing`?

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/OperatorChain.java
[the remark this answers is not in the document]
pnowojski: I meant above `allOperators` field, since it wasn't obvious to me when I was reading the code.

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/TwoInputStreamTaskTest.java
sunhaibotb: You're right, `testHarness.waitForInputProcessing` works well. I was misled by its name before.

sunhaibotb:
The comments were addressed, and the PR has been updated.  @pnowojski

sunhaibotb:
The Travis failure has nothing to do with this PR. After you review, if there are no questions, I'll squash all commits in one and rebase on the latest master. @pnowojski

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/operators/StreamSourceOperatorLatencyMetricsTest.java
```diff
@@ -164,7 +166,10 @@ private void testLatencyMarkEmission(int numberLatencyMarkers, OperatorSetupOper
 		operatorSetup.setupSourceOperator(operator, testProcessingTimeService);
 
 		// run and wait to be stopped
-		operator.run(new Object(), mock(StreamStatusMaintainer.class), new CollectorOutput<Long>(output), mock(OperatorChain.class));
+		OperatorChain<?, ?> operatorChain = new OperatorChain<>(
+			operator.getContainingTask(),
+			StreamTask.createRecordWriters(operator.getOperatorConfig(), new MockEnvironmentBuilder().build()));
+		operator.run(new Object(), mock(StreamStatusMaintainer.class), new CollectorOutput<Long>(output), operatorChain);
```
pnowojski: I think you should add `try/finally` block with `operatorChain.releaseOutputs()`, otherwise there will be a resource leak.

sunhaibotb: You're right. I overlooked that.

sunhaibotb:
The PR has been updated, and rebased on the latest master.  Thank you very much for the high quality review. @pnowojski

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/operators/StreamSourceOperatorLatencyMetricsTest.java
pnowojski: Thanks for fixing this also in other places.

sunhaibotb:
It seems that unstable case led to the failure of Travis. I have re-triggered the Travis through squashing all commits in one.

sunhaibotb:
The travis has been green. The later pushes were to re-trigger Travis because Travis was not successful for external reasons, and those pushes only rebased on the master and did not make any code changes.

pnowojski:
Thanks! Merging.

PR #8736, opened 2019-06-14
[FLINK-12846][table] Carry primary key and unique key information in TableSchema

wuchong:

The primary key and unique key is a standard meta information in SQL. And they are important information for optimization, for example, AggregateRemove, AggregateReduceGrouping and state layout optimization for TopN and Join.

So in this PR, we extend TableSchema to carry more information about primary key and unique keys. So that the TableSource can declare this meta information.

The primary key and unique key information will only work in Blink planner currently. Flink planner will just ignore the information.

## Brief change log

This pull request contains 2 commits:

1. Add primary key and unique key to `TableSchema` with some Javadocs. 
2. Use the primary key and unique key information in Blink planner. 
   - We did some refactor to `TableSourceTable` to make it only build the `FlinkStatistic` from `TableSource` instead of passing in from outside. Along with some tests changes. 

## Verifying this change

1. Adds a `TableSchemaTest` to check the schema builder methods.
2. Adds `FlinkRelMdUniqueKeysTest#testGetUniqueKeysOnTableSourceScan`  test to check the optimizer can get the unique key information from  `TableSchema` from `TableSource`.

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-table/flink-table-common/src/main/java/org/apache/flink/table/api/TableSchema.java
+++ b/flink-table/flink-table-common/src/main/java/org/apache/flink/table/api/TableSchema.java
@@ -56,14 +56,20 @@ import static org.apache.flink.table.types.utils.TypeConversions.fromLegacyInfoT
  *         A primary key or unique key on source will be simply trusted, we won't validate the
  *         constraint. The primary key and unique key information will then be used for query
  *         optimization. If a bounded or unbounded table source defines any primary key or
- *         unique key, it must contain a unique value for each row of data, you cannot have
- *         two records having the same value of that field(s).
+ *         unique key, it must contain a unique value for each row of data. You cannot have
+ *         two records having the same value of that field(s), otherwise, the query result
+ *         may be wrong.
  *     </li>
  *     <li>
  *         A primary key or unique key on sink is a weak constraint. Currently, we won't validate
  *         the constraint, but we may add some check in the future to validate whether the
  *         primary/unique key of the query matches the primary/unique key of the sink during
  *         compile time.
+ *     </li>
+ *     <li>
+ *         The difference between primary key and unique key is that there is only one primary key
+ *         and there can be more than one unique key. And a primary key doesn't need to be declared
+ *         in unique key list again.
  *     </li>
  * </ul>
  */
@@ -364,9 +370,9 @@ public class TableSchema {
 
 		private List<DataType> fieldDataTypes;
 
-		private List<String> primaryKey;
+		private Set<String> primaryKey;
 
-		private List<List<String>> uniqueKeys;
+		private Set<Set<String>> uniqueKeys;
 
 		public Builder() {
 			fieldNames = new ArrayList<>();
@@ -432,7 +438,14 @@ public class TableSchema {
 						"' is not existed in the schema.");
 				}
 			}
-			primaryKey = Arrays.asList(fields);
+			Set<String> keyFields = new HashSet<>(Arrays.asList(fields));
+			if (uniqueKeys != null && !uniqueKeys.isEmpty()) {
+				if (uniqueKeys.contains(keyFields)) {
+					throw new IllegalArgumentException(
+						"The primary key fields have already been declared as unique key.");
+				}
+			}
+			this.primaryKey = keyFields;
 			return this;
 		}
 
@@ -451,10 +464,23 @@ public class TableSchema {
 						"' is not existed in the schema.");
 				}
 			}
-			if (uniqueKeys == null) {
-				uniqueKeys = new ArrayList<>();
+			Set<String> keyFields = new HashSet<>(Arrays.asList(fields));
+			if (primaryKey != null) {
+				if (primaryKey.equals(keyFields)) {
+					throw new IllegalArgumentException(
+						"The unique key fields have already been declared as primary key.");
+				}
 			}
-			uniqueKeys.add(Arrays.asList(fields));
+			if (uniqueKeys != null && !uniqueKeys.isEmpty()) {
+				if (uniqueKeys.contains(keyFields)) {
+					throw new IllegalArgumentException(
+						"The unique key have already been declared.");
+				}
+			}
+			if (uniqueKeys == null) {
+				uniqueKeys = new HashSet<>();
+			}
+			uniqueKeys.add(keyFields);
 			return this;
 		}
 
--- a/flink-table/flink-table-common/src/main/java/org/apache/flink/table/api/TableSchema.java
+++ b/flink-table/flink-table-common/src/main/java/org/apache/flink/table/api/TableSchema.java
@@ -34,6 +34,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 
 import static org.apache.flink.table.api.DataTypes.FIELD;
@@ -43,7 +44,28 @@ import static org.apache.flink.table.types.utils.TypeConversions.fromDataTypeToL
 import static org.apache.flink.table.types.utils.TypeConversions.fromLegacyInfoToDataType;
 
 /**
- * A table schema that represents a table's structure with field names and data types.
+ * <p>A table schema that represents a table's structure with field names and data types and some
+ * constraint information (e.g. primary key, unique key).</p><br/>
+ *
+ * <p>Concepts about primary key and unique key:</p>
+ * <ul>
+ *     <li>
+ *         Primary key and unique key can consist of single or multiple columns (fields).
+ *     </li>
+ *     <li>
+ *         A primary key or unique key on source will be simply trusted, we won't validate the
+ *         constraint. The primary key and unique key information will then be used for query
+ *         optimization. If a bounded or unbounded table source defines any primary key or
+ *         unique key, it must contain a unique value for each row of data, you cannot have
+ *         two records having the same value of that field(s).
+ *     </li>
+ *     <li>
+ *         A primary key or unique key on sink is a weak constraint. Currently, we won't validate
+ *         the constraint, but we may add some check in the future to validate whether the
+ *         primary/unique key of the query matches the primary/unique key of the sink during
+ *         compile time.
+ *     </li>
+ * </ul>
  */
 @PublicEvolving
 public class TableSchema {
@@ -56,9 +78,15 @@ public class TableSchema {
 
 	private final Map<String, Integer> fieldNameToIndex;
 
-	private TableSchema(String[] fieldNames, DataType[] fieldDataTypes) {
+	private final String[] primaryKey;
+
+	private final String[][] uniqueKeys;
+
+	private TableSchema(String[] fieldNames, DataType[] fieldDataTypes, String[] primaryKey, String[][] uniqueKeys) {
 		this.fieldNames = Preconditions.checkNotNull(fieldNames);
 		this.fieldDataTypes = Preconditions.checkNotNull(fieldDataTypes);
+		this.primaryKey = primaryKey;
+		this.uniqueKeys = uniqueKeys;
 
 		if (fieldNames.length != fieldDataTypes.length) {
 			throw new TableException(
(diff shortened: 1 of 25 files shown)


wuchong:
@godfreyhe , do you have time to have a look?

flink-table/flink-table-common/src/main/java/org/apache/flink/table/api/TableSchema.java
```diff
@@ -244,13 +305,17 @@ public boolean equals(Object o) {
 		}
 		TableSchema schema = (TableSchema) o;
 		return Arrays.equals(fieldNames, schema.fieldNames) &&
-			Arrays.equals(fieldDataTypes, schema.fieldDataTypes);
+			Arrays.equals(fieldDataTypes, schema.fieldDataTypes) &&
+			Arrays.equals(primaryKey, schema.primaryKey) &&
+			Arrays.equals(uniqueKeys, schema.uniqueKeys);
```
docete: Arrays.equals doesn't "work" for two dimensional arrays. Use Arrays.deepEquals instead.

flink-table/flink-table-common/src/test/java/org/apache/flink/table/api/TableSchemaTest.java
```diff
@@ -0,0 +1,79 @@
...
+public class TableSchemaTest {
+
+	@Test
+	public void testToString() {
+		TableSchema.Builder builder = TableSchema.builder()
+			.field("a", DataTypes.INT())
+			.field("b", DataTypes.STRING())
+			.field("c", DataTypes.BIGINT());
+
+		String expected = "root\n |-- a: INT\n |-- b: STRING\n |-- c: BIGINT\n";
```
docete: field type should come with nullable information.
seems we should modify DataType.toString().

flink-table/flink-table-common/src/test/java/org/apache/flink/table/api/TableSchemaTest.java
```diff
@@ -0,0 +1,79 @@
...
+package org.apache.flink.table.api;
+
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+
+/**
+ * Tests for {@link TableSchema}.
+ */
+public class TableSchemaTest {
```
docete: add equals test(especially for uniqueKeys).

flink-table/flink-table-common/src/main/java/org/apache/flink/table/api/TableSchema.java
wuchong: Good catch! I also change the hashcode to `Arrays.deepHashCode`.

flink-table/flink-table-common/src/test/java/org/apache/flink/table/api/TableSchemaTest.java
wuchong: The default is nullable, and nullable will not be print in `toString`. If it is not null, the toString will be `INT NOT NULL`.

flink-table/flink-table-common/src/test/java/org/apache/flink/table/api/TableSchemaTest.java
wuchong: Sure

wuchong:
Thanks @docete , I have addressed the comments.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/api/TableSchema.java
```diff
@@ -43,7 +44,28 @@
 import static org.apache.flink.table.types.utils.TypeConversions.fromLegacyInfoToDataType;
 
 /**
- * A table schema that represents a table's structure with field names and data types.
+ * <p>A table schema that represents a table's structure with field names and data types and some
+ * constraint information (e.g. primary key, unique key).</p><br/>
+ *
+ * <p>Concepts about primary key and unique key:</p>
```
godfreyhe: do we distinguish primary key from unique key? in current javadoc, they have no difference.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/api/TableSchema.java
```diff
@@ -343,13 +412,66 @@ public Builder field(String name, TypeInformation<?> typeInfo) {
...
+				primaryKey == null,
+				"A primary key " + primaryKey +
+					" have been defined, can not define another primary key " +
+					Arrays.toString(fields));
+			for (String field : fields) {
+				if (!fieldNames.contains(field)) {
+					throw new IllegalArgumentException("The field '" + field +
+						"' is not existed in the schema.");
+				}
+			}
```
godfreyhe: that means we must build fieldNames first ?

flink-table/flink-table-common/src/main/java/org/apache/flink/table/api/TableSchema.java
```diff
@@ -343,13 +412,66 @@ public Builder field(String name, TypeInformation<?> typeInfo) {
...
+			if (uniqueKeys == null) {
+				uniqueKeys = new ArrayList<>();
+			}
+			uniqueKeys.add(Arrays.asList(fields));
```
godfreyhe: should distinct uniqueKey ?

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/plan/schema/TableSourceTable.scala
```diff
@@ -19,20 +19,24 @@
...
+import java.util
+
 /**
   * Abstract class which define the interfaces required to convert a [[TableSource]] to
   * a Calcite Table
   */
 class TableSourceTable[T](
     val tableSource: TableSource[T],
-    val isStreaming: Boolean,
-    val statistic: FlinkStatistic)
```
godfreyhe: if the `statistic` is deleted, how to store the statistic from catalog?

flink-table/flink-table-planner-blink/src/test/scala/org/apache/flink/table/util/TableTestBase.scala
```diff
@@ -147,13 +161,15 @@ abstract class TableTestUtil(test: TableTestBase) {
...
     * @param statistic statistic of current table
+    * @param uniqueKeys unique keys of current table
     * @return returns the registered [[Table]].
     */
   def addTableSource(
       name: String,
       types: Array[TypeInformation[_]],
       fields: Array[String],
-      statistic: FlinkStatistic = FlinkStatistic.UNKNOWN): Table
+      statistic: TableStats,
```
godfreyhe: we may add more info for `FlinkStatistic`, use `FlinkStatistic` instead of each fields to make sure this method and related test cases are stable.

flink-table/flink-table-planner-blink/src/test/scala/org/apache/flink/table/plan/metadata/MetadataTestUtil.scala
```diff
@@ -121,6 +123,22 @@ object MetadataTestUtil {
...
+      .field("height", DataTypes.DOUBLE())
+      .field("sex", DataTypes.STRING())
+      .field("class", DataTypes.INT())
+      .uniqueKey("id")
+      .build()
+    val tableSource = new TestTableSource(true, schema)
+    new TableSourceTable[BaseRow](tableSource, false)
+  }
+
+
```
godfreyhe: delete redundant blank line

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/plan/schema/TableSourceTable.scala
godfreyhe: Another scenario: rules (like `PushProjectIntoTableSourceScanRule`) does not change statistics   , so the new TableSource created by the rule could reuse the original TableSource, and avoid to call `TableSource`#getTableStats method which is high cost.
so the ` def copy(statistic: FlinkStatistic): FlinkTable` method defined in `FlinkTable` should not be delete too.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/api/TableSchema.java
wuchong: The difference between primary key and unique key is that there is only one primary key and there can be more than one unique key. And a primary key doesn't need to be declared in unique key list again.

I will add this to the class Javadoc.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/api/TableSchema.java
wuchong: Yes. I think so.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/api/TableSchema.java
wuchong: Sure, I will add that.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/plan/schema/TableSourceTable.scala
wuchong: 1. the statistic from catalog should be restored via `TableSource#getTableStats()`.
2. I think if the TableSource is not changed, then we don't need to re-construct a new `TableSourceTable`. I can reuse the original `TableSourceTable` and avoid calling `TableSource#getTableStats` again.

flink-table/flink-table-planner-blink/src/test/scala/org/apache/flink/table/util/TableTestBase.scala
wuchong: If the more information we need add to `FlinkStatistic`, I think it should also be included in `TableStats`. Regarding to the `relModifiedMonotonicity`, it is only be used internally in intermediate table source (`IntermediateRelTable`) which keeps `FlinkStatistic` as the constructor parameter.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/plan/schema/TableSourceTable.scala
godfreyhe: 1. `TableSource` should be decoupled with `Catalog`, or add a method like `setTableStats` for `TableSource` interface.
2. The TableSource may be changed,  e.g. project push down into table source

wuchong: 1. Yes, maybe we need something like `setTableStats`, but this is out of the scope of this issue.
2. If the TableSource is changed, shouldn't we always to create a new `TableSourceTable` and `getTableStats()` again? How do we know the stats is not changed?

flink-table/flink-table-planner-blink/src/test/scala/org/apache/flink/table/util/TableTestBase.scala
godfreyhe: what's the problem using `FlinkStatistic` ?

wuchong: We removed the `FlinkStatistic` from the constructor of `TableSourceTable` to make the statistic deriving simple. That's why we can't build a `TableSourceTable` from `FlinkStatistic`.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/plan/schema/TableSourceTable.scala
godfreyhe: 1. I do not think it‘s a good idea to add `setTableStats ` for `TableSource`, one big reason is each `TableSource` must implement this method and default implementation does not work due to a TableSource does not know whether the catalog could provide its stats.

3. yes, `TableSource` is immutable in `TableSourceTable`. currently, whether use the original stats or use unknown stats for new TableSourceTable is decided in each rule. you could reference this related code in `PushProjectIntoTableSourceScanRule` and `PushFilterIntoTableSourceScanRule` in blink inner code.

wuchong:
This PR will be splitted into several PRs. Let's continue the first one in #10213.

PR #8738, opened 2019-06-14
[FLINK-12845][sql-client] Execute multiple statements in command line…

docete:
… or sql script file

Allows sql client execute multiple statements from command line or sql file script.

## Brief change log
multiple statements may come from
- copy and paste to command line
- sql script files from SOURCE command (SOURCE myscript.sql)
- pipeline (sql-client.sh embedded < myscript.sql)
- incoming -f option

## Verifying this change

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/cli/CliClient.java
+++ b/flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/cli/CliClient.java
@@ -641,7 +641,7 @@ public class CliClient {
 	/**
 	 * Split a SemiColon-separated String, but ignore SemiColons in quotes.
 	 */
-	private static List<String> splitSemiColon(String line) {
+	static List<String> splitSemiColon(String line) {
 		boolean inSingleQuotes = false;
 		boolean inDoubleQuotes = false;
 		boolean escape = false;
--- a/flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/cli/CliClient.java
+++ b/flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/cli/CliClient.java
@@ -49,6 +49,7 @@ import java.nio.charset.Charset;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -196,8 +197,8 @@ public class CliClient {
 			if (line == null) {
 				continue;
 			}
-			final Optional<SqlCommandCall> cmdCall = parseCommand(line);
-			cmdCall.ifPresent(this::callCommand);
+
+			run(line);
 		}
 	}
 
@@ -572,8 +573,7 @@ public class CliClient {
 		terminal.flush();
 
 		// try to run it
-		final Optional<SqlCommandCall> call = parseCommand(stmt);
-		call.ifPresent(this::callCommand);
+		run(stmt);
 	}
 
 	// --------------------------------------------------------------------------------------------
@@ -613,6 +613,19 @@ public class CliClient {
 		terminal.flush();
 	}
 
+	private void run(String stmt) {
+		final List<String> statements = splitSemiColon(stmt);
+		for (String statement : statements) {
+			final Optional<SqlCommandCall> call = parseCommand(statement);
+
+			if (!call.isPresent()) {
+				// if we found a invalid command, issue error message and break
+				break;
+			}
+			call.ifPresent(this::callCommand);
+		}
+	}
+
 	// --------------------------------------------------------------------------------------------
 
 	private static Terminal createDefaultTerminal() {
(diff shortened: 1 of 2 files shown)


wuchong:
Hi @docete , could you add some test to verify the new feature? We should always add some tests when introduce new features.

docete:
> Hi @docete , could you add some test to verify the new feature? We should always add some tests when introduce new features.

added unit test.

bowenli86:
@docete can you rebase this PR?

bowenli86:
LGTM!

@wuchong @KurtYoung @twalthr do you have any concerns?

wuchong:
After revisiting the SQL CLI. I'm not sure about this feature. 
What will happen when the command is "show functions;show tables;show views;" ? Only show the result of last command?

Will this feature be covered by `-f <sql-script>` ?

docete:
> After revisiting the SQL CLI. I'm not sure about this feature.
> What will happen when the command is "show functions;show tables;show views;" ? Only show the result of last command?

will run & show these 3 statements one by one.

> 
> Will this feature be covered by `-f <sql-script>` ?

No. user may copy SQL statements and paste on a running cli. -f feature can't cover this scenario.

flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/cli/CliClient.java
```diff
@@ -624,4 +637,56 @@ private static Terminal createDefaultTerminal() {
...
+	 * Split a SemiColon-separated String, but ignore SemiColons in quotes.
+	 */
+	static List<String> splitSemiColon(String line) {
+		boolean inSingleQuotes = false;
+		boolean inDoubleQuotes = false;
+		boolean escape = false;
+
+		// normalize
+		line = line.replaceAll("--[^\r\n]*", ""); // remove single-line comments
+		line = line.replaceAll("/\\*[\\w\\W]*?(?=\\*/)\\*/", ""); // remove double-line comments
```
wuchong: This is the another concern from my side. We will lose the original text and character positions after replacing. 
This will make users confused if the SQL has parser error because the position is not correct. 

Another option is support all parsers in `flink-sql-parser`. The sql parser supports multi statement naturally and have proper exceptions. In this way, we will not need this "hack" replacing. But we may need more works. 

cc @danny0405 

Besides of that, we have met some issues in internal blink when removing comments and

JingsongLi:
Agree with using proper sql parser, @twalthr you are right, we have encountered too many parsing-related problems internally because of regex, which make it difficult to use special symbols(Especially backslash) in SQL, leading to various strange problems.

docete:
@wuchong @twalthr @JingsongLi  IMO the client should have the ability to split input lines into commands( ";" as delimiter). And this should not depends on sql parser. There are several special commands expect SQL statement a SQL client should support. Take SET command(e.g. "SET key1=value1") as example, Should we expend the sql parser to support this?

flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/cli/CliClient.java
danny0405: Support multiline sql parse work for sql-parser seems the right way to go, just one question: after parsing, we got a `SqlNodeList` with each node comes from a sql statement.

So we may need to unparse the node to sql text again when passing them to executor. The unparsed text may be different from the original one, so the error message may not be exact accurate sometimes.

wuchong: If it is an invalid multiline sql string, then it should fail in sql parser instead of passing to executor?

danny0405: Yeah, we should  just to keep the unparsed sql correct.

wuchong:
> I agree the current PR is way too hacky. However, I don't see the need of extending SQL parser just to support the multi-command line either as there can be commands that's not SQL at all. As far as I can see, we need to split the command line by ';' (certainly take care of possible escaped ';'), pre-process each segment to determine the command type (which we are already doing), and invoke sql parser only if the command is a sql command.

The problem is how to split command by semi-colon. Using regex is a hack way and we have met many issues because of it in the internal blink.

twalthr:
As far as I know Hive is also not SQL compliant and backslashes should not be used for escaping unless used with the following syntax:

```
String literals must be enclosed in single quotes (e.g., SELECT 'Hello World'). Duplicate a single quote for escaping (e.g., SELECT 'It''s me.'). Unicode characters are supported in string literals. If explicit unicode code points are required, use the following syntax:

Use the backslash (\) as escaping character (default): SELECT U&'\263A'
Use a custom escaping character: SELECT U&'#263A' UESCAPE '#'
```

What we need is a slightly different architecture for the SQL Client. Currently the DDL, DCL commands are parsed on the CLI side. However, this should happen in TableEnvironment instead such that the CLI is just a thin wrapper around the core API.

PR #8742, opened 2019-06-14
[FLINK-11879] Add validators for the uses of InputSelectable, BoundedOneInput and BoundedMultiInput

sunhaibotb:

This pull request add the following validators for the uses of `InputSelectable`, `BoundedOneInput` and `BoundedMultiInput`.
- Rejects the jobs containing operators which implement `InputSelectable` in case that checkpointing is enabled.
- Rejects the jobs containing operators which implement `BoundedInput` or `BoundedMultiInput` in case that checkpointing is enabled.
- Rejects the jobs containing operators which implement `InputSelectable` in case that credit-based flow control is disabled.

## Brief change log

  -  In `StreamingJobGraphGenerator`, add a validator that does not support the operators which implement `InputSelectable` or `BoundedOneInput` or `BoundedMultiInput` when checkpointing is enabled.
  -  In `StreamTask`, add a validator that does not support the operators which implement `InputSelectable` when credit-based mode is disabled.

## Verifying this change

  - Added test that validates the correctness of the new validator in `StreamingJobGraphGenerator`.
  - Added test that validates the correctness of the new validator in `StreamTask`.

## Documentation

### DIFF
--- a/flink-streaming-java/src/main/java/org/apache/flink/streaming/api/graph/StreamingJobGraphGenerator.java
+++ b/flink-streaming-java/src/main/java/org/apache/flink/streaming/api/graph/StreamingJobGraphGenerator.java
@@ -47,6 +47,7 @@ import org.apache.flink.streaming.api.CheckpointingMode;
 import org.apache.flink.streaming.api.checkpoint.WithMasterCheckpointHook;
 import org.apache.flink.streaming.api.environment.CheckpointConfig;
 import org.apache.flink.streaming.api.operators.ChainingStrategy;
+import org.apache.flink.streaming.api.operators.InputSelectable;
 import org.apache.flink.streaming.api.operators.StreamOperatorFactory;
 import org.apache.flink.streaming.api.operators.UdfStreamOperatorFactory;
 import org.apache.flink.streaming.api.transformations.ShuffleMode;
@@ -140,6 +141,7 @@ public class StreamingJobGraphGenerator {
 	}
 
 	private JobGraph createJobGraph() {
+		preValidate();
 
 		// make sure that all vertices start immediately
 		jobGraph.setScheduleMode(streamGraph.getScheduleMode());
@@ -176,6 +178,35 @@ public class StreamingJobGraphGenerator {
 		}
 
 		return jobGraph;
+	}
+
+	@SuppressWarnings("deprecation")
+	private void preValidate() {
+		CheckpointConfig checkpointConfig = streamGraph.getCheckpointConfig();
+
+		if (checkpointConfig.isCheckpointingEnabled()) {
+			// temporarily forbid checkpointing for iterative jobs
+			if (streamGraph.isIterative() && !checkpointConfig.isForceCheckpointing()) {
+				throw new UnsupportedOperationException(
+					"Checkpointing is currently not supported by default for iterative jobs, as we cannot guarantee exactly once semantics. "
+						+ "State checkpoints happen normally, but records in-transit during the snapshot will be lost upon failure. "
+						+ "\nThe user can force enable state checkpoints with the reduced guarantees by calling: env.enableCheckpointing(interval,true)");
+			}
+
+			ClassLoader classLoader = Thread.currentThread().getContextClassLoader();
+			for (StreamNode node : streamGraph.getStreamNodes()) {
+				StreamOperatorFactory operatorFactory = node.getOperatorFactory();
+				if (operatorFactory != null) {
+					Class<?> operatorClass = operatorFactory.getStreamOperatorClass(classLoader);
+					if (InputSelectable.class.isAssignableFrom(operatorClass)) {
+
+						throw new UnsupportedOperationException(
+							"Checkpointing is currently not supported for operators that implement InputSelectable:"
+								+ operatorClass.getName());
+					}
+				}
+			}
+		}
 	}
 
 	private void setPhysicalEdges() {
--- a/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
+++ b/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
@@ -20,6 +20,7 @@ package org.apache.flink.streaming.runtime.tasks;
 import org.apache.flink.annotation.Internal;
 import org.apache.flink.annotation.VisibleForTesting;
 import org.apache.flink.api.common.accumulators.Accumulator;
+import org.apache.flink.configuration.NettyShuffleEnvironmentOptions;
 import org.apache.flink.configuration.TaskManagerOptions;
 import org.apache.flink.core.fs.CloseableRegistry;
 import org.apache.flink.core.fs.FileSystemSafetyNet;
@@ -349,6 +350,9 @@ public abstract class StreamTask<OUT, OP extends StreamOperator<OUT>>
 			operatorChain = new OperatorChain<>(this, recordWriters);
 			headOperator = operatorChain.getHeadOperator();
 
+			// check environment for selective reading
+			checkSelectiveReadingEnv();
+
 			// task specific initialization
 			init();
 
@@ -603,6 +607,19 @@ public abstract class StreamTask<OUT, OP extends StreamOperator<OUT>>
 	boolean isSerializingTimestamps() {
 		TimeCharacteristic tc = configuration.getTimeCharacteristic();
 		return tc == TimeCharacteristic.EventTime | tc == TimeCharacteristic.IngestionTime;
+	}
+
+	private void checkSelectiveReadingEnv() {
+		boolean isCreditBased = getEnvironment()
+			.getTaskManagerInfo()
+			.getConfiguration()
+			.getBoolean(NettyShuffleEnvironmentOptions.NETWORK_CREDIT_MODEL);
+
+		if (!isCreditBased && operatorChain.hasSelectiveReadingOperator()) {
+			throw new UnsupportedOperationException(
+				"The operator that implements the InputSelectable interface is not supported in the non-credited" +
+					" network mode. (please set 'taskmanager.network.credit-model' to 'true')");
+		}
 	}
 
 	// ------------------------------------------------------------------------
--- a/flink-streaming-java/src/test/java/org/apache/flink/streaming/api/graph/StreamingJobGraphGeneratorTest.java
+++ b/flink-streaming-java/src/test/java/org/apache/flink/streaming/api/graph/StreamingJobGraphGeneratorTest.java
@@ -25,6 +25,7 @@ import org.apache.flink.api.common.io.InputFormat;
 import org.apache.flink.api.common.io.OutputFormat;
 import org.apache.flink.api.common.operators.ResourceSpec;
 import org.apache.flink.api.common.operators.util.UserCodeWrapper;
+import org.apache.flink.api.common.typeinfo.BasicTypeInfo;
 import org.apache.flink.api.common.typeinfo.TypeInformation;
 import org.apache.flink.api.dag.Transformation;
 import org.apache.flink.api.java.io.DiscardingOutputFormat;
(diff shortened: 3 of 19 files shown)


flink-streaming-java/src/main/java/org/apache/flink/streaming/api/graph/StreamingJobGraphGenerator.java
```diff
@@ -661,18 +664,31 @@ private void configureCheckpointing() {
...
+			StreamOperatorFactory operatorFactory = node.getOperatorFactory();
+			if (operatorFactory instanceof UdfStreamOperatorFactory) {
+				Function f = ((UdfStreamOperatorFactory) operatorFactory).getUserFunction();
 
 				if (f instanceof WithMasterCheckpointHook) {
 					hooks.add(new FunctionMasterCheckpointHookFactory((WithMasterCheckpointHook<?>) f));
 				}
 			}
+
+			if (cfg.isCheckpointingEnabled() && operatorFactory != null) {
```
1u0: I'd move this check (for all nodes) into a separate function and invoke the new function earlier (in `StreamingJobGraphGenerator.createJobGraph()` call or even before `StreamingJobGraphGenerator` is used).

`StreamGraph.getJobGraph()` already has a validation check before creating a `JobGraph`. Imo, it's a better place to carry such validation for now (excluding the case for `InputSelectable` and network credit-based flow control).

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
```diff
@@ -337,6 +338,9 @@ public final void invoke() throws Exception {
 			operatorChain = new OperatorChain<>(this, recordWriters);
 			headOperator = operatorChain.getHeadOperator();
 
+			// check environment for selective reading
+			checkSelectiveReadingEnv();
```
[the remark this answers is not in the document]
sunhaibotb: If validated on the JM side, one way to determine whether a job contains selective reading operator is to mark it by adding a boolean field in `JobGraph` when generating JobGraph, and another way is to deserialize `StreamOperatorFactory` from `StreamConfig`. I don't think the second way is good, because JM currently does not need to deserialize `StreamConfig`, which will increase the time cost. Considering that the default value of `taskmanager.network.credit-model` is true, this check is a protective code and it does not normally trigger a check exception. In addition, `taskmanager.network.credit-model` is an option that is deprecated, and the validation will not be needed after it is removed. So I think it's easier to add validation here and to remove it in the future.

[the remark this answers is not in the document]
1u0: My main suspect was that you may not have the same environment configuration at JM's side.

But regarding `OperatorChain.hasSelectiveReadingOperator()` implementation, I think delegating it to JM should not add more overhead. You already have
```java
Class<?> operatorClass = operatorFactory.getStreamOperatorClass();
InputSelectable.class.isAssignableFrom(operatorClass);
```
check in [`StreamingJobGraphGenerator`](https://github.com/apache/flink/blob/9ab0b7d3b23c8c548edd0723be34e143e144afaf/flink-streaming-java/src/main/java/org/apache/flink/streaming/api/graph/StreamingJobGraphGenerator.java#L682-L683).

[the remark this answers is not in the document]
sunhaibotb: I'm not sure if what you're talking about is the first way I said. If yes, I don't think it's necessary to add a boolean field to the job graph for the checking logic which will be removed in the future. If not, I understand you're talking about the second way as I said, that `StreamOperatorFactory` needs to be deserialized from `StreamConfig` on the JM side, which will increase the time cost ( Please note that JM currently does not need to deserialize `StreamConfig`). If neither method is not, I don't know what your proposal is like.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/graph/StreamingJobGraphGenerator.java
sunhaibotb: The following validation that you are talking about allows to enable the forced checkpointing for the iterative streaming, which means that no exception is thrown in that case. In this PR, we need to add the check to cover this case, otherwise `IterateITCase#testWithCheckPointing` will fail.
```
if (isIterative() && checkpointConfig.isCheckpointingEnabled() && !checkpointConfig.isForceCheckpointing()) 
```

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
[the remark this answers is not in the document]
1u0: My understanding was that `StreamingJobGraphGenerator.createJobGraph()` happens in JM side only. But it looks like it is also called at sql-client, where you may not have the same environment configuration.

> taskmanager.network.credit-model is an option that is deprecated, and the validation will not be needed after it is removed. So I think it's easier to add validation here and to remove it in the future.

I'm fine with this.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/graph/StreamingJobGraphGenerator.java
pnowojski: I agree with @1u0. This method is definitely already too big if we want to add some new functionality here, it must be decomposed into smaller pieces. But I also agree that this check would be better to perform earlier. Maybe in some `private void  StreamingJobGraphGenerator#validateJobGraph()` method (together with the present validation that happens in `StreamGraph#getJobGraph`). 

With this `validateJobGraph()` either we could call it pre-emptively as a first thing in `createJobGraph()` or expect a user to call it manually before `createJobGraph()`.

sunhaibotb:
Thanks for reviewing @pnowojski  @1u0 . 
The code has been updated.  Because of the conflict with the latest master branch, I rebased on the latest master and forcibly pushed. You can just look at `StreamingJobGraphGenerator` and `StreamGraph`.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/graph/StreamingJobGraphGenerator.java
```diff
@@ -178,6 +182,36 @@ private JobGraph createJobGraph() {
...
+			for (StreamNode node : streamGraph.getStreamNodes()) {
+				StreamOperatorFactory operatorFactory = node.getOperatorFactory();
+				if (operatorFactory != null) {
+					Class<?> operatorClass = operatorFactory.getStreamOperatorClass();
+					if (InputSelectable.class.isAssignableFrom(operatorClass)
+						|| BoundedOneInput.class.isAssignableFrom(operatorClass)
+						|| BoundedMultiInput.class.isAssignableFrom(operatorClass)) {
+
+						throw new UnsupportedOperationException(
+							"Checkpointing currently does not supported the operator that implements" +
```
1u0: I suggest to change it to `"Checkpointing is currently not supported for operators that implement "`.

flink-streaming-java/src/test/java/org/apache/flink/streaming/api/graph/StreamingJobGraphGeneratorTest.java
```diff
@@ -673,4 +682,87 @@ public void testBlockingConnectionsBetweenChainsEnabled() {
...
+
+			DataStreamSource<String> source1 = env.fromElements("1");
+			DataStreamSource<Integer> source2 = env.fromElements(1);
+			source1.connect(source2)
+				.transform("test",
+					BasicTypeInfo.STRING_TYPE_INFO,
+					new TestAnyModeReadingStreamOperator("test operator"))
+				.print();
+
+			StreamingJobGraphGenerator.createJobGraph(env.getStreamGraph());
```
1u0: What about refactoring this test, by extracting building a sample job graph (with operator under the test as a parameter) via a helper method and also, splitting the test into three separate tests?

If they run separately, you can also use simpler structure
```
@Test(expected = UnsupportedOperationException.class)
public void testValidationForOperator...() throws Exception {
    // setup graph:
    ...
    StreamingJobGraphGenerator.createJobGraph(env.getStreamGraph());
    fail("unreachable");
}

File: flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/runtime/CodeGenOperatorFactory.java [hunk 5]
```diff
@@ -59,7 +59,12 @@ public ChainingStrategy getChainingStrategy() {
 
 	@Override
 	public boolean isOperatorSelectiveReading() {
-		return InputSelectable.class.isAssignableFrom(generatedClass.getClass(Thread.currentThread().getContextClassLoader()));
+		return InputSelectable.class.isAssignableFrom(getStreamOperatorClass());
+	}
+
+	@Override
+	public Class<? extends StreamOperator> getStreamOperatorClass() {
+		return generatedClass.getClass(Thread.currentThread().getContextClassLoader());
```
1u0: Is it possible that this code can be triggered in a settings where the context class loader is not a user code class loader?

sunhaibotb: I will add a `Classloader` type parameter to this method.

sunhaibotb:
The code has been updated, and please review it again. Thanks @1u0

1u0:
LGTM, % with still open question:

> `generatedClass.getClass(Thread.currentThread().getContextClassLoader());`
>
> Is it possible that this code can be triggered in a settings where the context class loader is not a user code class loader?

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/runtime/CodeGenOperatorFactory.java
```diff
@@ -59,7 +59,12 @@ public ChainingStrategy getChainingStrategy() {
 
 	@Override
 	public boolean isOperatorSelectiveReading() {
-		return InputSelectable.class.isAssignableFrom(generatedClass.getClass(Thread.currentThread().getContextClassLoader()));
+		return InputSelectable.class.isAssignableFrom(getStreamOperatorClass(Thread.currentThread().getContextClassLoader()));
+	}
+
+	@Override
+	public Class<? extends StreamOperator> getStreamOperatorClass(ClassLoader classLoader) {
```
pnowojski: this method is inconsistent with `isOperatorSelectiveReading()` api that just uses `getContextClassLoader()`. Please make them consistent and use the same mechanism: either passing `classLoader` in both places or using `getContextClassLoader` in both.

sunhaibotb: Because a general method `getStreamOperatorClass()` is added, `isOperatorSelectiveReading()` is not needed and can be dropped. It is planned to drop `isOperatorSelectiveReading()` from the PR of the JIRA (https://issues.apache.org/jira/browse/FLINK-13051), so no changes had been made to make them consistent. But the best thing to do is to keep them consistency first whether `isOperatorSelectiveReading()` is dropped or not. I will solve it.

sunhaibotb:
The code has been updated. Because of the conflict with the latest master branch, I rebased on the latest master and forcibly pushed. 

A new problem is that batch and streaming will share some operators, some of which may need to implement `Bounded[Multi|One]Input` in the batch mode to do something in the `endInput()` method, such as `ContinuousFileReaderOperator` (please look at  [issue#13376](https://issues.apache.org/jira/browse/FLINK-13376)). But if these operators implement `Bounded[Multi|One]Input`, streaming will not be able to use them because of the validation in this PR.

After I re-checked the code related to the checkpoint handler, I think that there was no need to add the validator for the use of `Bounded[Multi|One]Input`. Because as long as an `EndOfPartition` event arrives, the current checkpointing mechanism will not generate a successful checkpoint after that. That is, failure to handle the `endInput` state for checkpointing does not cause correctness problems. What do you think? @pnowojski 

(In the latest code of this PR, I've removed the validation on `Bounded[Multi|One]Input`. If I'm wrong, I'll add it back.)

sunhaibotb:
The conflict has resolved and test passed. Quick link to the CI Report: https://github.com/apache/flink/pull/8742#issuecomment-510731561g

pnowojski:
Thanks @sunhaibotb . Merging.

PR #8322, opened 2019-04-30
[FLINK-12364] Introduce a CheckpointFailureManager to centralized manage checkpoint failure

yanghua:
## What is the purpose of the change

*This pull request introduces a CheckpointFailureManager to centralized manage checkpoint failure*

## Brief change log

  - *Introduce a CheckpointFailureManager to centralized manage checkpoint failure*
  - *Add a test case for CheckpointFailureManager*

## Verifying this change

This change is already covered by existing tests, such as *CheckpointFailureManagerTest*.

## Does this pull request potentially affect one of the following parts:

  - Dependencies (does it add or upgrade a dependency): (yes / **no**)
  - The public API, i.e., is any changed class annotated with `@Public(Evolving)`: (yes / **no**)
  - The serializers: (yes / **no** / don't know)
  - The runtime per-record code paths (performance sensitive): (yes / **no** / don't know)
  - Anything that affects deployment or recovery: JobManager (and its components), Checkpointing, Yarn/Mesos, ZooKeeper: (yes / **no** / don't know)
  - The S3 file system connector: (yes / **no** / don't know)

## Documentation

  - Does this pull request introduce a new feature? (yes / **no**)
  - If yes, how is the feature documented? (not applicable / docs / JavaDocs / **not documented**)

PR #8745, opened 2019-06-14
[FLINK-11662] Disable task to fail on checkpoint errors

Myasuka:

FLINK-11662 has been a known and urgent issue after Flink-1.8 release. There exist two parts to fix this problem, one is from task side to disable task to fail on checkpoint errors. Another part is to let JM  to decide when to fail the whole job which has been part of work in https://github.com/apache/flink/pull/8322. 

This PR mainly focus on the task side and choose to base on https://github.com/apache/flink/pull/8322 before it merged since that is very unlikely to still have significant changes at this point. Once https://github.com/apache/flink/pull/8322 merged, this PR would then rebased on the latest code.

## Brief change log

  - Deprecate `isFailTaskOnCheckpointError` and `setFailTaskOnCheckpointError` in `ExecutionConfig`. 
 - Deprecate `isFailOnCheckpointingErrors` and `setFailOnCheckpointingErrors` in `CheckpointConfig`. 
  - Remove `FailingCheckpointExceptionHandler`

## Verifying this change

  - Refactor `CheckpointExceptionHandlerConfigurationTest` and `CheckpointExceptionHandlerTest` to verify the default logic changed.
 - Refactor `StreamTaskTest` to verify the default logic changed on task side.

  - The serializers: **no**

## Documentation

### DIFF
--- a/flink-core/src/main/java/org/apache/flink/api/common/ExecutionConfig.java
+++ b/flink-core/src/main/java/org/apache/flink/api/common/ExecutionConfig.java
@@ -154,7 +154,10 @@ public class ExecutionConfig implements Serializable, Archiveable<ArchivedExecut
 	/** This flag defines if we use compression for the state snapshot data or not. Default: false */
 	private boolean useSnapshotCompression = false;
 
-	/** Determines if a task fails or not if there is an error in writing its checkpoint data. Default: true */
+	/**
+	 * @deprecated Should no longer be used because we would not support to let task directly fail on checkpoint error.
+	 */
+	@Deprecated
 	private boolean failTaskOnCheckpointError = true;
 
 	/** The default input dependency constraint to schedule tasks. */
@@ -948,20 +951,22 @@ public class ExecutionConfig implements Serializable, Archiveable<ArchivedExecut
 	}
 
 	/**
-	 * This method is visible because of the way the configuration is currently forwarded from the checkpoint config to
-	 * the task. This should not be called by the user, please use CheckpointConfig.isFailTaskOnCheckpointError()
-	 * instead.
+	 * @deprecated This method takes no effect since we would not forward the configuration from the checkpoint config
+	 * to the task, and we have not supported task to fail on checkpoint error.
+	 * Please use CheckpointConfig.getTolerableCheckpointFailureNumber() to know the behavior on checkpoint errors.
 	 */
+	@Deprecated
 	@Internal
 	public boolean isFailTaskOnCheckpointError() {
 		return failTaskOnCheckpointError;
 	}
 
 	/**
-	 * This method is visible because of the way the configuration is currently forwarded from the checkpoint config to
-	 * the task. This should not be called by the user, please use CheckpointConfig.setFailOnCheckpointingErrors(...)
-	 * instead.
+	 * @deprecated This method takes no effect since we would not forward the configuration from the checkpoint config
+	 * to the task, and we have not supported task to fail on checkpoint error.
+	 * Please use CheckpointConfig.setTolerableCheckpointFailureNumber(int) to determine the behavior on checkpoint errors.
 	 */
+	@Deprecated
 	@Internal
 	public void setFailTaskOnCheckpointError(boolean failTaskOnCheckpointError) {
 		this.failTaskOnCheckpointError = failTaskOnCheckpointError;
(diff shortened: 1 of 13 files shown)


Myasuka:
@flinkbot attention @StefanRRichter

flink-core/src/main/java/org/apache/flink/api/common/ExecutionConfig.java
```diff
@@ -154,8 +154,11 @@
 	/** This flag defines if we use compression for the state snapshot data or not. Default: false */
 	private boolean useSnapshotCompression = false;
 
-	/** Determines if a task fails or not if there is an error in writing its checkpoint data. Default: true */
-	private boolean failTaskOnCheckpointError = true;
+	/**
+	 * @deprecated Should no longer be used because we would not support to let task fail on checkpoint error.
```
StefanRRichter: I think we should mention the (kind of) replacement flag for tolerable checkpoint errors here. And also how the two interact.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
```diff
@@ -323,9 +323,8 @@ public final void invoke() throws Exception {
 
 			CheckpointExceptionHandlerFactory cpExceptionHandlerFactory = createCheckpointExceptionHandlerFactory();
 
-			synchronousCheckpointExceptionHandler = cpExceptionHandlerFactory.createCheckpointExceptionHandler(
-				getExecutionConfig().isFailTaskOnCheckpointError(),
-				getEnvironment());
+			synchronousCheckpointExceptionHandler = cpExceptionHandlerFactory
```
StefanRRichter: I wonder if we still need the sync and async exception handler, factoy, etc ... because there is only a single implementation left now and that could be inlined again.

Myasuka: Agree, there is no more benefit to distinguish these two exception handlers. Already removed the `asynchronousCheckpointExceptionHandler` in `StreamTask`.

flink-runtime/src/main/java/org/apache/flink/runtime/checkpoint/CheckpointFailureManager.java
```diff
@@ -31,18 +31,24 @@
...
 	private final Set<Long> countedCheckpointIds;
 
 	public CheckpointFailureManager(int tolerableCpFailureNumber, FailJobCallback failureCallback) {
-		checkArgument(tolerableCpFailureNumber >= 0,
+		checkArgument(tolerableCpFailureNumber >= 0 || tolerableCpFailureNumber == UNDEFINED_TOLERABLE_CHECKPOINT_NUMBER,
 			"The tolerable checkpoint failure number is illegal, " +
 				"it must be greater than or equal to 0 .");
-		this.tolerableCpFailureNumber = tolerableCpFailureNumber;
+		// set tolerableCpFailureNumber as 0 when passing UNDEFINED_TOLERABLE_CHECKPOINT_NUMBER.
+		if (tolerableCpFailureNumber == UNDEFINED_TOLERABLE_CHECKPOINT_NUMBER) {
```
StefanRRichter: I wonder if we should introduce and handle the `UNDEFINED_TOLERABLE_CHECKPOINT_NUMBER` in this class. It seems to me that this is more an implementation detail of how the configuration is determined and propagated. If possible, I would suggest to resolve this already before at the time when the configuration is determined and not in the constuctor of this class. I would suggest to pass only "valid" and resolved configuration here.

Myasuka: I agree to make `tolerableCheckpointFailureNumber` valid before `CheckpointFailureManager` or `CheckpointCoordinatorConfiguration`. But I still want to introduce `UNDEFINED_TOLERABLE_CHECKPOINT_NUMBER` at least in `CheckpointConfig` to judge whether `setFailOnCheckpointingErrors` would take effect. 
The only left problem is when we want to call `CheckpointConfig#getTolerableCheckpointFailureNumber`. If that field was already `-1`, shall we set the `tolerableCheckpointFailureNumber` as `0` or just return `0` without setting that. I prefer to return `0` but leave `tolerableCheckpointFailureNumber` still as `-1` to avoid such side effects. What do you think?

StefanRRichter: Yes, your suggestion sounds good and I am also not against this in `CheckpointConfig`, I just think `CheckpointFailureManager` constructor is a bit late.

PR #8751, opened 2019-06-16
[FLINK-11937][StateBackend]Resolve small file problem in RocksDB incremental checkpoint

klion26:

This pr wants to fix the problem of too many small files in RocksDB incremental checkpoint.
- Resue the same underlying file in one checkpoint of one operator, this means we just generate a single file for one checkpoint.
- Do not support to use `ByteStreamStateHandle` in the new feature FSCSOS, will always flush state to file.

This PR  includes 7 commits:
- 924a60c   Add a parameter checkpointId for CheckpointStreamFactory#createCheckpointStateOutputStream, this is a pre-work for the new feature FSCSOS
- a150108   Expose maxConcurrentCheckpoints from StreamConfig, a pre-work for the new feature FSCSOS, used to close the earlier opened underlying file which used in a failed checkpoint.
- a020eed Introduce all the necessary components for the new feature FSCSOS and some necessary tests to verify it.
- 5609d30  Add a new statehandle type in SvaepointV2Serializer, **DO NOT** change the layout, and some tests to verify it
- 969a58d  Expose getSharedState() in IncrementalKeyedStateHandle, this is a pre-work to enable the FSCSOS feature.
- 68fa3be  Enable the new feature FSCSOS — includes  introduce RocksDBSegmentStateBackendFactory, and modify SharedStateRegisty,  StateBackendLoader, RocksDBStateBackend, RocksIncrementalSnapshotStrategy , and so on
- 82127e2  Add some tests to verify the new introduced feature FSCSOS, includes SharedStateRegistryTest, RocksDBSegmentStateBackendFactoryTest and RocksDBStateBackendTest.

## Brief change log

This pr introduces some necessary classes, such as
- FsSegmentCheckpointStorage
- FsCheckpointStreamFactory
- FsSegmentCheckpointStorageLocation
- FsSegmentCheckpointStreamFactory
- FsSegmentStateBackend
- FsSegmentStateHandle
This pr modifies some exist classes to support this feature, such as:
- SharedStateRegistry
- RocksDBStateBackend
and add some tests to verify the correct.

## Verifying this change
This pr introduces some test and modified some exist test to verify the changes.
- SharedStateRegistryTest
- RocksDBStateBackendTest
- FsSegmentCheckpointStorageTest
- FsSegmentCheckpointStorageTest.java
- FsSegmentStateBackendEntropyTest.java
- FsSegmentStateHandleTest.java
- RocksDBSegmentStateBackendFactoryTest.java
- SavepointV2SerializerTest.java

  - The serializers: (**no**)
  - Anything that affects deployment or recovery: JobManager (and its components), Checkpointing, Yarn/Mesos, ZooKeeper: (**yes**)

## Documentation

  - If yes, how is the feature documented? (**JavaDocs**)

### DIFF
--- a/flink-state-backends/flink-statebackend-rocksdb/src/main/java/org/apache/flink/contrib/streaming/state/snapshot/RocksIncrementalSnapshotStrategy.java
+++ b/flink-state-backends/flink-statebackend-rocksdb/src/main/java/org/apache/flink/contrib/streaming/state/snapshot/RocksIncrementalSnapshotStrategy.java
@@ -69,7 +69,6 @@ import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Set;
 import java.util.SortedMap;
 import java.util.UUID;
 import java.util.concurrent.RunnableFuture;
@@ -98,7 +97,7 @@ public class RocksIncrementalSnapshotStrategy<K> extends RocksDBSnapshotStrategy
 
 	/** Stores the materialized sstable files from all snapshots that build the incremental history. */
 	@Nonnull
-	private final SortedMap<Long, Set<StateHandleID>> materializedSstFiles;
+	private final SortedMap<Long, Map<StateHandleID, StreamStateHandle>> materializedSstFiles;
 
 	/** The identifier of the last completed checkpoint. */
 	private long lastCompletedCheckpointId;
@@ -120,7 +119,7 @@ public class RocksIncrementalSnapshotStrategy<K> extends RocksDBSnapshotStrategy
 		@Nonnull CloseableRegistry cancelStreamRegistry,
 		@Nonnull File instanceBasePath,
 		@Nonnull UUID backendUID,
-		@Nonnull SortedMap<Long, Set<StateHandleID>> materializedSstFiles,
+		@Nonnull SortedMap<Long, Map<StateHandleID, StreamStateHandle>> materializedSstFiles,
 		long lastCompletedCheckpointId,
 		int numberOfTransferingThreads) {
 
@@ -155,7 +154,7 @@ public class RocksIncrementalSnapshotStrategy<K> extends RocksDBSnapshotStrategy
 		LOG.trace("Local RocksDB checkpoint goes to backup path {}.", snapshotDirectory);
 
 		final List<StateMetaInfoSnapshot> stateMetaInfoSnapshots = new ArrayList<>(kvStateInformation.size());
-		final Set<StateHandleID> baseSstFiles = snapshotMetaData(checkpointId, stateMetaInfoSnapshots);
+		final Map<StateHandleID, StreamStateHandle> baseSstFiles = snapshotMetaData(checkpointId, stateMetaInfoSnapshots);
 
 		takeDBNativeCheckpoint(snapshotDirectory);
 
@@ -219,12 +218,12 @@ public class RocksIncrementalSnapshotStrategy<K> extends RocksDBSnapshotStrategy
 		}
 	}
 
-	private Set<StateHandleID> snapshotMetaData(
+	private Map<StateHandleID, StreamStateHandle> snapshotMetaData(
 		long checkpointId,
 		@Nonnull List<StateMetaInfoSnapshot> stateMetaInfoSnapshots) {
 
 		final long lastCompletedCheckpoint;
-		final Set<StateHandleID> baseSstFiles;
+		final Map<StateHandleID, StreamStateHandle> baseSstFiles;
 
 		// use the last completed checkpoint as the comparison base.
 		synchronized (materializedSstFiles) {
@@ -280,13 +279,13 @@ public class RocksIncrementalSnapshotStrategy<K> extends RocksDBSnapshotStrategy
 
 		/** All sst files that were part of the last previously completed checkpoint. */
 		@Nullable
-		private final Set<StateHandleID> baseSstFiles;
+		private final Map<StateHandleID, StreamStateHandle> baseSstFiles;
 
 		private RocksDBIncrementalSnapshotOperation(
 			long checkpointId,
 			@Nonnull CheckpointStreamFactory checkpointStreamFactory,
 			@Nonnull SnapshotDirectory localBackupDirectory,
-			@Nullable Set<StateHandleID> baseSstFiles,
+			@Nullable Map<StateHandleID, StreamStateHandle> baseSstFiles,
 			@Nonnull List<StateMetaInfoSnapshot> stateMetaInfoSnapshots) {
 
 			this.checkpointStreamFactory = checkpointStreamFactory;
@@ -320,7 +319,7 @@ public class RocksIncrementalSnapshotStrategy<K> extends RocksDBSnapshotStrategy
 				uploadSstFiles(sstFiles, miscFiles);
 
 				synchronized (materializedSstFiles) {
-					materializedSstFiles.put(checkpointId, sstFiles.keySet());
+					materializedSstFiles.put(checkpointId, sstFiles);
 				}
 
 				final IncrementalRemoteKeyedStateHandle jmIncrementalKeyedStateHandle =
@@ -343,7 +342,7 @@ public class RocksIncrementalSnapshotStrategy<K> extends RocksDBSnapshotStrategy
 							directoryStateHandle,
 							keyGroupRange,
 							metaStateHandle.getTaskLocalSnapshot(),
-							sstFiles.keySet());
+							sstFiles);
 
 					snapshotResult = SnapshotResult.withLocalState(jmIncrementalKeyedStateHandle, localDirKeyedStateHandle);
 				} else {
@@ -445,7 +444,7 @@ public class RocksIncrementalSnapshotStrategy<K> extends RocksDBSnapshotStrategy
 				final StateHandleID stateHandleID = new StateHandleID(fileName);
 
 				if (fileName.endsWith(SST_FILE_SUFFIX)) {
-					final boolean existsAlready = baseSstFiles != null && baseSstFiles.contains(stateHandleID);
+					final boolean existsAlready = baseSstFiles != null && baseSstFiles.containsKey(stateHandleID);
 
 					if (existsAlready) {
 						// we introduce a placeholder state handle, that is replaced with the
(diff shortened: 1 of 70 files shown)


StephanEwen:
I like the idea of fixing the small file problem a lot - I think it is very relevant in practice.

Before looking at the code, we need a lot more information for a good review:

  - An extensive high-level description of the changes, including the changes to state handles, checkpoint metadata, etc.
  - Is this replacing the existing functionality or a new option that users activate?
  - What is the compatibility for existing users?
  - Can this be abstracted within the state backend without changes to the state handle system? If not, why and could it be done with a bit different approach? Something that is in the state backend is easier/less sensitive than major changes to the shared state registry, etc.

Given this list, I am wondering if this is a FLIP-worthy change in the end.

I think this also needs attention from @tzulitai for checking if anything about snapshot metadata is affected here, as well as @aljoscha @kl0u to check if this may affect savepoint / checkpoint properties etc.

StephanEwen:
@flinkbot attention @tzulitai @aljoscha @kl0u

klion26:
@StephanEwen thanks for the comments, I'm trying to answer the questions below:
- As a high-level description: this change introduces a new state handle(`FsSegmentStateHandle`), modify checkpoint metadata(**Do not modify the layout**, just add a type for `FsSegmentStateHandle` in `SavepointV2Serializer`), add some information in `SharedRegistry` to track the underlying file's reference, and other necessary modifications.
- This is a new option needs users to activate
- I think there are no compatibility problems, **for the checkpoint meta we don't change the layout**, just add a new type for the new state handle, for restoring from the exist `FileStateHandle`, we'll delivery to `RocksDBStateDownloader#downloadDataForStateHandle` and will read both the `FileStateHandle`  and `FsSegmentStateHandle` correctly, for `SharedStateRegistry` all the modifications just affect the newly introduced state handle only.
- In my opinion, we can't just do this change in state bankend. I'm trying to give the reasons below:
   - first, we need to track the position(start position and end position) in the file of currently state handle(because after applying this change, each state handle is mapping to a block of one file)
   - second, we need to track the reference count of the underlying file, so that we can delete the file in the feature in time(not too early and not too late, delete the file too early will encounter problems of `IOException`, delete too late will consume too may disk space).

Please let me know if I need to give more information, sir.

carp84:
@StephanEwen Thanks for the attention sir. Besides above answers from @klion26 , there's also a [design doc](https://docs.google.com/document/d/1ukLfqNt44yqhDFL3uIhd68NevVdawccb6GflGNFzLcg) and some offline review given by @StefanRRichter and @tzulitai , FYI.

klion26:
Travis failed because of some other component's checkstyle's problem, close and reopen to trigger Travis build again.

klion26:
@StephanEwen  @tzulitai @aljoscha @kl0u Could you please take a look at this, thanks.
Besides the before comments. I also updated the description of the PR, mainly added this two description about this pr:
- Resue the same underlying file in one checkpoint of one operator, this means we just generate a single file for one checkpoint.
- Do not support to use ByteStreamStateHandle in the new feature FSCSOS, will always flush state to file.

the failed test is irrelevant, there is an issue [FLINK-9900](https://issues.apache.org/jira/browse/FLINK-9900) tracking it.

flink-state-backends/flink-statebackend-rocksdb/src/main/java/org/apache/flink/contrib/streaming/state/snapshot/RocksIncrementalSnapshotStrategy.java
```diff
@@ -154,8 +160,12 @@ public RocksIncrementalSnapshotStrategy(
 		final SnapshotDirectory snapshotDirectory = prepareLocalSnapshotDirectory(checkpointId);
 		LOG.trace("Local RocksDB checkpoint goes to backup path {}.", snapshotDirectory);
 
+		if (this.checkpointStreamFactory == null) {
+			this.checkpointStreamFactory = checkpointStreamFactory;
```
streaming-olap: The line 175 may use this.checkpointStreamFactory replace  checkpointStreamFactory.

yumengz5:
Do you deal with the storage amplification in this pr?

streaming-olap:
> Do you deal with the storage amplification in this pr?

Yes, I introduce this pr into the production in my company, but there is some bug which causes the connection leak . If you introduce this pr into th production, you should fix the bug.

klion26:
@yumengz5 thanks for the question, this pr did not solve the storage amplification.
@streaming-olap thanks for the information and the feedback. good to know that you introduce this to your company.

jiexray:
https://github.com/klion26/flink/blob/2dda201498d913b3fbf2e7d078dd5440652e6a19/flink-state-backends/flink-statebackend-rocksdb/src/main/java/org/apache/flink/contrib/streaming/state/snapshot/RocksIncrementalSnapshotStrategy.java#L191

Hello, the FSDataOutputStream for a checkpoint is closed, when the checkpoint is notified from JM. Why do you delay the close of the output stream. It can be closed when the RocksStateUploader finishes uploading all sst files for the current checkpoint. Since the output stream has already been unuseful,  keeping it open is a waste.

StephanEwen:
@klion26 and @carp84 The problem that this PR wants to solve should also be solved by [FLIP-158](https://cwiki.apache.org/confluence/display/FLINK/FLIP-158%3A+Generalized+incremental+checkpoints). Using the FLIP-158 approach with a longer checkpoint interval is also possible, it doesn't necessarily have to go with very short intervals.

And FLIP-158 would also have the advantage that it does not change the way that streams need to stay open, artifacts span different checkpoints, etc. which makes the whole shared state management much more complex.

What do you think about pursuing FLIP-158 further instead?

carp84:
@StephanEwen please correct me if I'm wrong, but I think the issue FLIP-158 is trying to resolve is orthogonal with this one (FLINK-11937).

On one hand, IIUC, the snapshot interval (for generating SST files (take RocksDB for example) to truncate change-logs) in FLIP-158 design would be configurable, and if it's set to some value similar to the old checkpoint interval, eg. 10min, then we will have similar small file problem as observed now. Actually I don't think this snapshot interval should be too long since it will decide how much logs to replay during restore thus affecting the recovery speed.

OTOH, I'm not sure about the value of using change-log based checkpoint with long checkpoint interval (like more than 10min). Saving change logs will consume additional network bandwidth and disk space (since the SST uploading process is reserved for log truncation) and increase the latency of routine record processing (for "double-writing"), which is a good trade-off for momentary checkpoint interval but not that effective with long ones, IMHO.

Not sure whether I'm missing anything, and please let me know your thoughts. Thanks.

StephanEwen:
@carp84 In FLIP-158, the checkpoint interval doesn't influence when RocksDB takes its checkpoint. Instead, RocksDB backend decides itself when to checkpoint, and can do this for example always when we estimate that a full write buffer has been accumulated (like when gathered as many bytes 0.95 x write_buffer_size). That way we avoid small SST files. What may remain are small metadata files (like few kilobytes size), but we could generally inline the metadata into the state handles directly.

carp84:
Thanks for the clarification @StephanEwen . I think size-based and time-based snapshot are two options each with their advantage. The size-based snapshot for log truncation/consolidation could indeed prevent the small file problem but will also have possibly longer recovery time (we will replay as much as 0.95 x write_buffer_size logs during restore before the size-based flush triggered). The time-based snapshot could compact/truncate the log at fixed pace but may introduce small file problem, unless we introduce some mechanism similar to the proposal here, which I think is still valuable.

And just to confirm, are we aiming at completely replacing the snapshot-based checkpoint with log-based checkpoint in the future? Or both will be reserved for different user scenario? If the later, I think we still need to resolve the small file problems for snapshot-based checkpoint.

StephanEwen:
@carp84 The checkpointing from FLIP-158 is a specific state backend that is used to wrap another state backend. So with that change, we can use the existing RocksDB incremental snapshots and the RocksDB log-based snapshots.

We need to see in practice how much it costs to replay one write buffer from the log. With a default write buffer size, that would be around 60 MB. I would assume that doing it in one batch insert would be quite fast, given that there is no processing or serialization overhead (the log just stored bytes as well).

carp84:
@StephanEwen It seems to me the current RocksDB incremental snapshots are time-based instead of size-based, so I'm not sure whether we could directly reuse it through wrapping in change-log backend. However, this is another topic and let's watch and discuss more when implementing the checkpointing part for change-log backend in FLINK-21356.

It's true that if we could prevent generating small files then we don't need the solution here to handle the small file problem, and although I'm still not fully convinced we could get rid of the problem with FLIP-158, I agree we focus on implementing FLIP-158 first and revisit the issue here later if still in need. :-)

klion26:
@StephanEwen sorry for the late reply.
IIUC, maybe we still need this feature after FLIP-158 has been implemented(even if we cloud handle not to generate small sst files -- 64MB). -- before we always set the file threshold in this pr to 128(or 256MB) in production, the job which has a large checkpoint may not generate too much small sst files(sure, it will generate some small sst files also) -- the checkpoint interval will set to ~ 10 min.

Jiayi-Liao:
Hi @StephanEwen @carp84 , our team also implements the feature at Bytedance recently, but in a different way without complex changes on JobMaster's side. The core idea is, 

**Snapshot:**
* Use `RocksDBIncrementalPackageStrategy` to package the incremental files into several `SegmentFile` logically (we implement a default strategy based on the `dfs.blocksize` config on Hdfs)
* For each `SegmentFile`, write the incremental files to a single file on Hdfs and return a `SegmentFileStateHandle` which also maintains the incremental files' state handles inside
* `RocksIncrementalSnapshotStrategy` maintains a mapping between sstable file and `SegmentFile`, to build the `PlaceholderStreamStateHandle` in `IncrementalKeyedStateHandle`
* Shared state management use `SegmentFileStateHandle` instead of `StreamStateHandle` to manage the states 

**Restore:**
* Download all the sstable files with the metadata stored in `SegmentFileStateHandle`
* Restore the mapping between sstable file and `SegmentFile` in `RocksIncrementalSnapshotStrategy`

The good thing is that, we keep the mapping between sstable file and `SegmentFile` on TaskManager side instead of JobMaster side, and there is no change on the shared state management. 

About the FLIP-158, IMO it can help relieve the impact of small file problem but may not solve it. The number of sstable files can be too many even the write buffer is full every time when being flushed, for example it's normal in our scenarios that several GB incremental state are produced with 20+ sstable files per task, especially in cases like feeding samples to online training models with multiple-join in Flink.

flink-state-backends/flink-statebackend-rocksdb/src/main/java/org/apache/flink/contrib/streaming/state/snapshot/RocksIncrementalSnapshotStrategy.java
```diff
@@ -177,6 +187,11 @@ public void notifyCheckpointComplete(long completedCheckpointId) {
 				materializedSstFiles.keySet().removeIf(checkpointId -> checkpointId < completedCheckpointId);
 				lastCompletedCheckpointId = completedCheckpointId;
 			}
+
+			// We need to close the file for current completed checkpoint.
+			if (checkpointStreamFactory instanceof FsSegmentCheckpointStreamFactory) {
```
zlzhang0122, line 192: I think maybe we also need to close the file for current aborted checkpoint too.

PR #8756, opened 2019-06-17
[FLINK-12406] [Runtime] Report BLOCKING_PERSISTENT result partition meta back to client

Xpray:
This PR follows FLINK-12405. 
When a Job finishes, all result partition locations of `BLOCKING_PERSISTENT` type will be reported back from JM to client(`ExecutionEnvironment`), and the `TableEnvironment` can access these locations in later PRs.

## Description
- *The client use `ExecutionEnvironment` to submit a batch job and wait for the `JobResult` from `JM`*
- *When a job finishes, `BLOCKING_PERSISTENT` `ResultPartition` locations will be collected from each `ExecutionVertex` in `ExecutionGraph`, these locations are captured by `ShuffleDescriptor`s*
- *On JM side, the `BLOCKING_PERSISTENT` `ResultPartition` locations flow in this path (`ExecutionGraph` -> `ArchivedExecutionGraph` -> `JobResult`)*
- *On client side, A `JobExecutionResult` will be created with the returned `JobResult`*
- *The `ExecutionEnvironment` sees the `JobExecutionResult` containing the locations, and merge these new locations and stores in itself*

## Failure Handling
- *If any error occurs during collecting locations of `BLOCKING_PERSISTENT` `ResultPartition`, 
we do not terminate the process but leave incomplete locations of some `IntermediateDataSet`, we keep record of these IntermediateDataSetIDs and report back to client* 
- *So the Client can decide what to do with these complete and incomplete locations, generally
the data can be read if locations are complete, or a delete request will be proposed(in later PRs) if the locations are incomplete*

## Brief change log

[217e7fc](https://github.com/apache/flink/pull/8756/commits/217e7fc9470eaa21e070f8863d803d38e24e8d86
) 
  - *The location of a `ResultPartition` is captured by a `ShuffleDescriptor`*
  - *After Job finishes, a `ShuffleDescriptor` will be created for each `BLOCKING_PERSISTENT` `ResultPartition`, and these `ShuffleDescriptor`s may be used in following job.*
  - *Add a new interface `PersistentIntermediateResultDescriptor`, and its implementation `DefaultPersistentIntermediateResultDescriptor` which contains several `ShuffleDescriptor`s*
  - *Since `ShuffleDescritor` is a runtime class, `DefaultPersistentIntermediateResultDescriptor` use a `SerializedValue<Object> to keep the serialized value of a `ShuffleDescritor`*
  - *Add a new method `getPersistentIntermediateResultDescriptor()` in `AccessExecutionGraph`, which returns a `PersistentIntermediateResultDescriptor`, `ExecutionGraph` and `ArchivedExecutionGraph` will implement this new method*
  - *Add an instance of `PersistentIntermediateResultDescriptor` in `JobExecutionResult`, `JobResult`, `ArchivedExecutionGraph` and `ExecutionEnvironment`*

[733f19e](https://github.com/apache/flink/pull/8756/commits/733f19e9a0020267c59013c9d11e643abd792c27
)  
- *`ExecutionEnvironment` keeps track of all previous descriptors*

[17e4fd9](https://github.com/apache/flink/pull/8756/commits/17e4fd95fc3f54d8a7b3349b03ea7785ef3fbcb0
) 
- *Add test cases*

## Verifying this change

org.apache.flink.test.interactive.PersistentShuffleResultITCase#testReportingPersistentIntermediateResultDescriptorToClient
org.apache.flink.runtime.executiongraph.ExecutionVertexTest#testCreateShuffleDescriptorFromExecutionVertex

  - The serializers: no

## Documentation

     [FLIP-36](https://cwiki.apache.org/confluence/display/FLINK/FLIP-36%3A+Support+Interactive+Programming+in+Flink)


flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/ExecutionGraph.java
```diff
@@ -814,6 +820,69 @@ public Executor getFutureExecutor() {
...
+		Set<AbstractID> failedIntermediateDataSetIds = new HashSet<>();
+
+		for (ExecutionVertex executionVertex : getAllExecutionVertices()) {
+			for (IntermediateResultPartition intermediateResultPartition : executionVertex.getProducedPartitions().values()) {
+				if (intermediateResultPartition.getResultType() == ResultPartitionType.BLOCKING_PERSISTENT) {
+					try {
+						addLocation(resultPartitionDescriptors, intermediateResultPartition);
+					} catch (Throwable throwable) {
+						LOG.error("Failed to Failed to get location of ResultPartition: " + intermediateResultPartition.getPartitionId(), throwable);
+						failedIntermediateDataSetIds.add(
```
shuiqiangchen: When getting FailedIntermediateDataSets, we should also return the corresponding error messages so that the client side can be aware of the reason why they can not use the cache.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/ExecutionGraph.java
```diff
@@ -814,6 +820,69 @@ public Executor getFutureExecutor() {
...
+		// keep record of all failed IntermediateDataSetID
+		Set<AbstractID> failedIntermediateDataSetIds = new HashSet<>();
+
+		for (ExecutionVertex executionVertex : getAllExecutionVertices()) {
+			for (IntermediateResultPartition intermediateResultPartition : executionVertex.getProducedPartitions().values()) {
+				if (intermediateResultPartition.getResultType() == ResultPartitionType.BLOCKING_PERSISTENT) {
+					try {
+						addLocation(resultPartitionDescriptors, intermediateResultPartition);
+					} catch (Throwable throwable) {
+						LOG.error("Failed to Failed to get location of ResultPartition: " + intermediateResultPartition.getPartitionId(), throwable);
```
shuiqiangchen: Duplicated "Failed to".

shuiqiangchen: Duplicated "Failed to".

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/ExecutionGraph.java
```diff
@@ -814,6 +820,69 @@ public Executor getFutureExecutor() {
...
+						failedIntermediateDataSetIds.add(
+							new AbstractID(intermediateResultPartition.getIntermediateResult().getId()));
```
shuiqiangchen: When getting FailedIntermediateDataSets, we should also return the corresponding error messages so that the client side can be aware of the reason why they can not use the cache.

Xpray: Thanks for the review @shuiqiangchen. I'll see what I can do to deliver these causes to the client.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/ExecutionGraph.java
Xpray: I'll take care of this.

flink-java/src/main/java/org/apache/flink/api/java/ExecutionEnvironment.java
```diff
@@ -118,6 +119,9 @@
 
 	private final ExecutionConfig config = new ExecutionConfig();
 
+	private final BlockingPersistentResultPartitionMeta blockingPersistentResultPartitionMeta =
```
StephanEwen: This should probably not be in a separate field, but in the `lastJobExecutionResult`.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/AccessExecutionGraph.java
```diff
@@ -158,6 +159,14 @@
 	 */
 	Map<String, SerializedValue<OptionalFailure<Object>>> getAccumulatorsSerialized();
 
+	/**
+	 * Returns a {@link BlockingPersistentResultPartitionMeta}.
+	 * @return BlockingPersistentResultPartitionMeta contains ResultPartition locations
+	 */
+	default BlockingPersistentResultPartitionMeta getBlockingPersistentResultPartitionMeta() {
```
StephanEwen: I think this is a tricky use of a default methods, because this is not a working implementation for all subclasses, which is how default methods on interfaces should only be used.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/ExecutionGraph.java
```diff
@@ -814,6 +820,69 @@ public Executor getFutureExecutor() {
...
+		return new BlockingPersistentResultPartitionMeta(resultPartitionDescriptors, failedIntermediateDataSetIds);
+	}
+
+	/**
+	 *
+	 * @param resultPartitionDescriptors
+	 * @param intermediateResultPartition
+	 * throw exception if any error occurs.
+	 */
+	public void addLocation(
```
StephanEwen: This utility method should not be in the ExecutionGraph - it just blows up the public signature.
It should be private static or moved to a utility class.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/ExecutionGraph.java
```diff
@@ -814,6 +820,69 @@ public Executor getFutureExecutor() {
...
+		Map<AbstractID, ResultPartitionDescriptor> map = resultPartitionDescriptors.computeIfAbsent(
+			new AbstractID(dataSetID), key -> new HashMap<>()
+		);
+
+		TaskManagerLocation taskManagerLocation = null;
+
+		// The taskManagerLocation should be ready already since the previous job is done.
+		try {
+			taskManagerLocation = intermediateResultPartition
+				.getProducer().getCurrentExecutionAttempt().getTaskManagerLocationFuture().get(1, TimeUnit.SECONDS);
```
StephanEwen: This is a blocking waiting call, which cannot be used in a non blocking data structure like the execution graph.
The call to the future needs to complete or fail instantly.

flink-tests/src/test/java/org/apache/flink/test/operators/ExecutionEnvironmentITCase.java
```diff
@@ -66,6 +74,49 @@ public void mapPartition(Iterable<Integer> values, Collector<Integer> out) throw
...
+		// IntermediateDataSetID should be the same
+		Assert.assertEquals(intermediateDataSetID, intermediateDataSetID);
+
+		Map<AbstractID, ResultPartitionDescriptor> descriptors = meta.getResultPartitionDescriptors().get(intermediateDataSetID);
+
+		Assert.assertEquals(1, descriptors.size());
+
+		ResultPartitionDescriptor descriptor = descriptors.values().iterator().next();
+
+		Assert.assertTrue(
```
StephanEwen: Separate conditions should have separate assertions, so that it is visible from a test failure which of the conditions failed.

flink-tests/src/test/java/org/apache/flink/test/operators/ExecutionEnvironmentITCase.java
```diff
@@ -66,6 +74,49 @@ public void mapPartition(Iterable<Integer> values, Collector<Integer> out) throw
 		assertEquals(PARALLELISM, resultCollection.size());
 	}
 
+	@Test
+	public void testAccessingBlockingPersistentResultPartition() throws Exception {
```
StephanEwen: I am not sure if this test is well placed in the ExecutionEnvironmentITCase.
It does not test the ExecutionEnvironment, it tests a different feature.

Xpray:
@StephanEwen , Thanks for the review, I'll try to catch up with the lasted changes on shuffle service with @zhijiangW first.

zhijiangW:
Thanks for opening this PR @Xpray !

Just quickly looked through the codes, the new proposed class `ResultPartitionDescriptor` actually provides the same information with current `ShuffleDescriptor`.  I think the `ShuffleMaster` might provide a method for querying `ShuffleDescriptor` for your use in `BlockingPersistentResultPartitionMeta`.

azagrebin:
I agree with @zhijiangW. We should use `ShuffleMaster#registerPartitionWithProducer` to obtain `ShuffleDescriptor` for cached partitions as we do it for any others. The idea is that concrete implementation of `ShuffleDescriptor` contains enough information to consume the cached partition later using the same shuffle service which produced it. Execution graph and client code should be completely oblivious to the internals of `ShuffleDescriptor`.

flink-java/src/main/java/org/apache/flink/api/java/ExecutionEnvironment.java
Xpray: I think this should be a separate field, since it keeps track of all descriptors from all previous jobs rather than the latest one.

flink-tests/src/test/java/org/apache/flink/test/operators/ExecutionEnvironmentITCase.java
Xpray: I'll move this test to a new ITCase and try to add some unit tests.

zhijiangW:
@Xpray what is the status of this PR?
I found it inactive for more than a year. Do we need to close it for cleanup?

zhijiangW:
After syncing with related guys offline, this PR is not needed. I close it now.

PR #8757, opened 2019-06-17
[FLINK-12850][core] Introduce TypeInfo for LocalDate/LocalTime/LocalDateTime

JingsongLi:

Now in the new type system of table, the default class of time type is LocalDate and so on.
There are some situations that need to be converted to TypeInformation, such as toDataStream, so we need to provide TypeInformation support such as LocalDate.

## Brief change log

Introduce LocalTimeTypeInfo
Introduce LocalDateSerializer, LocalTimeSerializer, LocalDateTimeSerializer
Introduce LocalDateComparator, LocalTimeComparator, LocalDateTimeComparator

## Verifying this change

ut

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-core/src/main/java/org/apache/flink/api/common/typeutils/base/LocalDateComparator.java
+++ b/flink-core/src/main/java/org/apache/flink/api/common/typeutils/base/LocalDateComparator.java
@@ -1,0 +1,187 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.api.common.typeutils.base;
+
+import org.apache.flink.annotation.Internal;
+import org.apache.flink.api.common.typeutils.TypeComparator;
+import org.apache.flink.core.memory.DataInputView;
+import org.apache.flink.core.memory.DataOutputView;
+import org.apache.flink.core.memory.MemorySegment;
+
+import java.io.IOException;
+import java.io.Serializable;
+import java.time.LocalDate;
+
+/**
+ * This class can not extend {@link BasicTypeComparator}, because LocalDate is a
+ * Comparable of ChronoLocalDate instead of Comparable of LocalDate.
+ */
+@Internal
+public final class LocalDateComparator extends TypeComparator<LocalDate> implements Serializable {
+
+	private transient LocalDate reference;
+
+	protected final boolean ascendingComparison;
+
+	// For use by getComparators
+	@SuppressWarnings("rawtypes")
+	private final LocalDateComparator[] comparators = new LocalDateComparator[] {this};
+
+	public LocalDateComparator(boolean ascending) {
+		this.ascendingComparison = ascending;
+	}
+
+	@Override
+	public int hash(LocalDate value) {
+		return value.hashCode();
+	}
+
+	@Override
+	public void setReference(LocalDate toCompare) {
+		this.reference = toCompare;
+	}
+
+	@Override
+	public boolean equalToReference(LocalDate candidate) {
+		return candidate.equals(reference);
+	}
+
+	@Override
+	public int compareToReference(TypeComparator<LocalDate> referencedComparator) {
+		int comp = ((LocalDateComparator) referencedComparator).reference.compareTo(reference);
+		return ascendingComparison ? comp : -comp;
+	}
+
+	@Override
+	public int compare(LocalDate first, LocalDate second) {
+		int cmp = first.compareTo(second);
+		return ascendingComparison ? cmp : -cmp;
+	}
+
+	@Override
+	public boolean invertNormalizedKey() {
+		return !ascendingComparison;
+	}
+
+	@Override
+	public boolean supportsSerializationWithKeyNormalization() {
+		return false;
+	}
+
+	@Override
+	public void writeWithKeyNormalization(LocalDate record, DataOutputView target) throws IOException {
+		throw new UnsupportedOperationException();
+	}
+
+	@Override
+	public int extractKeys(Object record, Object[] target, int index) {
+		target[index] = record;
+		return 1;
+	}
+
+	@SuppressWarnings("rawtypes")
+	@Override
+	public TypeComparator[] getFlatComparators() {
+		return comparators;
+	}
+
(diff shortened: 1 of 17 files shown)


flink-core/src/main/java/org/apache/flink/api/common/typeutils/base/LocalDateComparator.java
```diff
@@ -0,0 +1,187 @@
...
+	}
+
+	@Override
+	public void setReference(LocalDate toCompare) {
+		this.reference = toCompare;
+	}
+
+	@Override
+	public boolean equalToReference(LocalDate candidate) {
+		return candidate.equals(reference);
```
liyafan82: use Objects.equals to avid NPE?

flink-core/src/main/java/org/apache/flink/api/common/typeutils/base/LocalDateComparator.java
```diff
@@ -0,0 +1,187 @@
...
+
+	public static int compareSerializedLocalDate(DataInputView firstSource, DataInputView secondSource,
+			boolean ascendingComparison) throws IOException {
+		int cmp = firstSource.readInt() - secondSource.readInt();
+		if (cmp == 0) {
+			cmp = firstSource.readByte() - secondSource.readByte();
+			if (cmp == 0) {
+				cmp = firstSource.readByte() - secondSource.readByte();
+			}
+		}
```
liyafan82: This logic is confusing for me. Internally, the LocalDate takes up 4 bytes, but we are using 6 bytes for comparison?

flink-core/src/main/java/org/apache/flink/api/common/typeutils/base/LocalDateSerializer.java
```diff
@@ -0,0 +1,113 @@
...
+	@Override
+	public int getLength() {
+		return 6;
+	}
+
+	@Override
+	public void serialize(LocalDate record, DataOutputView target) throws IOException {
+		if (record == null) {
+			target.writeInt(Integer.MIN_VALUE);
+			target.writeShort(Short.MIN_VALUE);
```
liyafan82: should call target.writeByte(Byte.MIN_VALUE) twice

flink-core/src/main/java/org/apache/flink/api/common/typeutils/base/LocalDateComparator.java
JingsongLi: `candidate` of `equalToReference` should not null.
Just like other comparator do.

flink-core/src/main/java/org/apache/flink/api/common/typeutils/base/LocalDateTimeComparator.java
```diff
@@ -0,0 +1,153 @@
...
+		throw new UnsupportedOperationException();
+	}
+
+	@Override
+	public int compareSerialized(DataInputView firstSource, DataInputView secondSource) throws IOException {
+		int cmp = dateComparator.compareSerialized(firstSource, secondSource);
+		if (cmp == 0) {
+			cmp = timeComparator.compareSerialized(firstSource, secondSource);
+		}
+		return cmp;
```
liyafan82: should we consider ascendingComparison?

JingsongLi: `this.dateComparator = new LocalDateComparator(ascending);`
The sub comparator has considered the `ascendingComparison`.

flink-core/src/main/java/org/apache/flink/api/common/typeutils/base/LocalDateTimeSerializer.java
```diff
@@ -0,0 +1,118 @@
...
+	}
+
+	@Override
+	public LocalDateTime deserialize(DataInputView source) throws IOException {
+		LocalDate localDate = LocalDateSerializer.INSTANCE.deserialize(source);
+		LocalTime localTime = LocalTimeSerializer.INSTANCE.deserialize(source);
+		if (localDate == null && localTime == null) {
+			return null;
+		} else if (localDate == null || localTime == null) {
+			throw new IOException("LocalDate and LocalTime are either null or not null together.");
```
liyafan82: Exactly one of LocalDate and LocalTime is null.

flink-core/src/main/java/org/apache/flink/api/common/typeutils/base/LocalDateSerializer.java
```diff
@@ -0,0 +1,113 @@
...
+@Internal
+public final class LocalDateSerializer extends TypeSerializerSingleton<LocalDate> {
+
+	private static final long serialVersionUID = 1L;
+
+	public static final LocalDateSerializer INSTANCE = new LocalDateSerializer();
+
+	@Override
+	public boolean isImmutableType() {
+		return false;
```
wuchong: return true?  LocalDate is immutable, right?

flink-core/src/main/java/org/apache/flink/api/common/typeutils/base/LocalDateSerializer.java
```diff
@@ -0,0 +1,113 @@
...
+	}
+
+	@Override
+	public LocalDate createInstance() {
+		return LocalDate.of(1970, 1, 1);
```
wuchong: I would suggest to add a `private static final LocalDate EPOCH = LocalDate.ofEpochDay(0L);`  and returns `EPOCH` here.

flink-core/src/main/java/org/apache/flink/api/common/typeutils/base/LocalDateTimeSerializer.java
```diff
@@ -0,0 +1,118 @@
...
+@Internal
+public final class LocalDateTimeSerializer extends TypeSerializerSingleton<LocalDateTime> {
+
+	private static final long serialVersionUID = 1L;
+
+	public static final LocalDateTimeSerializer INSTANCE = new LocalDateTimeSerializer();
+
+	@Override
+	public boolean isImmutableType() {
+		return false;
```
wuchong: return true?

flink-core/src/main/java/org/apache/flink/api/common/typeutils/base/LocalTimeSerializer.java
```diff
@@ -0,0 +1,117 @@
...
+@Internal
+public final class LocalTimeSerializer extends TypeSerializerSingleton<LocalTime> {
+
+	private static final long serialVersionUID = 1L;
+
+	public static final LocalTimeSerializer INSTANCE = new LocalTimeSerializer();
+
+	@Override
+	public boolean isImmutableType() {
+		return false;
```
wuchong: return true?

flink-core/src/main/java/org/apache/flink/api/common/typeutils/base/LocalTimeSerializer.java
```diff
@@ -0,0 +1,117 @@
...
+	}
+
+	@Override
+	public LocalTime createInstance() {
+		return LocalTime.of(0, 0);
```
wuchong: use LocalTime.MIDNIGHT directly?

flink-core/src/main/java/org/apache/flink/api/common/typeutils/base/LocalDateSerializer.java
JingsongLi: Good catch.

PR #8759, opened 2019-06-17
[FLINK-12868][yarn] Fix yarn cluster can not be deployed if plugins dir does not exist

pnowojski:
Plugins dir should be optional and cluster deployment should not fail if it doesn't exist.

## Verifying this change

This PR adds a new unit test to add better coverage (apart of end to end/integration tests) for this bug.

  - Dependencies (does it add or upgrade a dependency): (yes / no**)**
  - The serializers: (yes / no ****/ don't know)

## Documentation

### DIFF
--- a/flink-dist/src/main/assemblies/bin.xml
+++ b/flink-dist/src/main/assemblies/bin.xml
@@ -85,7 +85,14 @@ under the License.
 			<outputDirectory>bin</outputDirectory>
 			<fileMode>0755</fileMode>
 		</fileSet>
-		
+
+		<!-- copy plugins directory -->
+		<fileSet>
+			<directory>src/main/flink-bin/plugins</directory>
+			<outputDirectory>plugins</outputDirectory>
+			<fileMode>0755</fileMode>
+		</fileSet>
+
 		<!-- flink scala shell-->
 		<fileSet>
 			<directory>../flink-scala-shell/start-script/</directory>
--- a/flink-dist/src/main/flink-bin/plugins/README.txt
+++ b/flink-dist/src/main/flink-bin/plugins/README.txt
@@ -1,0 +1,18 @@
+The expected structure is as follows: the given plugins root folder,
+containing the plugins folder. One plugin folder contains all
+resources (jar files) belonging to a plugin. The name of the plugin
+folder becomes the plugin id. For example:
+
+plugins/ (root folder)
+    |------------plugin-a/ (folder of plugin a)
+    |                |-plugin-a-1.jar
+    |                |-plugin-a-2.jar
+    |                |-...
+    |                |-(the jars containing the classes of plugin a)
+    |
+    |------------plugin-b/
+    |                |-plugin-b-1.jar
+    |                |-...
+    ...
+
+Read more in the documentation about how to create a plugin.
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/clusterframework/overlays/FlinkDistributionOverlay.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/clusterframework/overlays/FlinkDistributionOverlay.java
@@ -75,7 +75,12 @@ public class FlinkDistributionOverlay extends AbstractContainerOverlay {
 		addPathRecursively(flinkBinPath, TARGET_ROOT, container);
 		addPathRecursively(flinkConfPath, TARGET_ROOT, container);
 		addPathRecursively(flinkLibPath, TARGET_ROOT, container);
-		addPathRecursively(flinkPluginsPath, TARGET_ROOT, container);
+		if (flinkPluginsPath.isDirectory()) {
+			addPathRecursively(flinkPluginsPath, TARGET_ROOT, container);
+		}
+		else {
+			LOG.warn("The plugins directory '" + flinkPluginsPath + "' doesn't exist.");
+		}
 	}
 
 	public static Builder newBuilder() {
--- a/flink-yarn/src/main/java/org/apache/flink/yarn/AbstractYarnClusterDescriptor.java
+++ b/flink-yarn/src/main/java/org/apache/flink/yarn/AbstractYarnClusterDescriptor.java
@@ -1515,30 +1515,39 @@ public abstract class AbstractYarnClusterDescriptor implements ClusterDescriptor
 	}
 
 	protected void addEnvironmentFoldersToShipFiles(Collection<File> effectiveShipFiles) {
+		addLibFoldersToShipFiles(effectiveShipFiles);
+		addPluginsFoldersToShipFiles(effectiveShipFiles);
+	}
+
+	private void addLibFoldersToShipFiles(Collection<File> effectiveShipFiles) {
 		// Add lib folder to the ship files if the environment variable is set.
 		// This is for convenience when running from the command-line.
 		// (for other files users explicitly set the ship files)
 		String libDir = System.getenv().get(ENV_FLINK_LIB_DIR);
 		if (libDir != null) {
-			addEnvFolderToShipFiles(effectiveShipFiles, libDir, ENV_FLINK_LIB_DIR);
+			File directoryFile = new File(libDir);
+			if (directoryFile.isDirectory()) {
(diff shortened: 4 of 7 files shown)


flink-runtime/src/main/java/org/apache/flink/runtime/clusterframework/overlays/FlinkDistributionOverlay.java
```diff
@@ -75,7 +75,12 @@ public void configure(ContainerSpecification container) throws IOException {
 		addPathRecursively(flinkBinPath, TARGET_ROOT, container);
 		addPathRecursively(flinkConfPath, TARGET_ROOT, container);
 		addPathRecursively(flinkLibPath, TARGET_ROOT, container);
-		addPathRecursively(flinkPluginsPath, TARGET_ROOT, container);
+		if (flinkPluginsPath.isDirectory()) {
```
azagrebin: Why is it obligatory to set `ENV_FLINK_PLUGINS_DIR` for mesos but not for yarn? If I do not miss something `FlinkDistributionOverlay.Builder.fromEnvironment.getObligatoryFileFromEnvironment(ENV_FLINK_PLUGINS_DIR)` will fail if `ENV_FLINK_PLUGINS_DIR` is not set.

flink-yarn/src/main/java/org/apache/flink/yarn/AbstractYarnClusterDescriptor.java
```diff
@@ -1520,25 +1520,27 @@ protected void addEnvironmentFoldersToShipFiles(Collection<File> effectiveShipFi
...
-	}
-
-	private void addEnvFolderToShipFiles(Collection<File> effectiveShipFiles, String directory, String environmentVariableName) {
-		File directoryFile = new File(directory);
-		if (directoryFile.isDirectory()) {
-			effectiveShipFiles.add(directoryFile);
-		} else {
-			throw new YarnDeploymentException("The environment variable '" + environmentVariableName +
-				"' is set to '" + directory + "' but the directory doesn't exist.");
+			File directoryFile = new File(pluginsDir);
```
azagrebin: nit: I would move `ENV_FLINK_LIB_DIR `/ `ENV_FLINK_PLUGINS_DIR ` cases to separate methods even if deduplication is under question

flink-runtime/src/main/java/org/apache/flink/runtime/clusterframework/overlays/FlinkDistributionOverlay.java
pnowojski: Hmm, good question. I was just repeating the same pattern as with `ENV_FLINK_LIB_DIR`. So asking the correct question: why is it obligatory to set `ENV_FLINK_LIB_DIR` for mesos but not for yarn? I have no idea. 

If you want, we can create a ticket to unify those two things - which first step would be to find out which way they should be unified.

azagrebin: True, we can create a separate ticket. The reason, I brought up it here, is that the PR basically addresses the problem (or at least some of them) that the previous deployments (w/o plugins) are not broken after plugins have been introduced. I guess the previous mesos deployments (same as yarn) do not have `ENV_FLINK_PLUGINS_DIR` and this case should be probably also tolerated, can be a separate issue though if you think it needs more discussion.

TisonKun:
Hi @pnowojski , as you said "plugins dir should be optional", isn't this option decided by whether user set "ENV_FLINK_PLUGINS_DIR"? Why do we suppress an exception caused by user explicitly set the path and dir not exist? I would say one who sets the path expects there is really a plugin directory and with this exception thrown we can fail fast.

flink-runtime/src/main/java/org/apache/flink/runtime/clusterframework/overlays/FlinkDistributionOverlay.java
azagrebin: As discussed this concern might be not an issue currently because as long as the user runs our `flink-dist` scripts to submit mesos jobs, `config.sh` has a default value for `FLINK_PLUGINS_DIR` which might not exist though but should work overall with this fix. The plugins shipping can only break deployments with custom scripts, same as with `lib` and other directories.

azagrebin:
We should also at least add a similar warning if the configured directory does not exist where it is actually used in `PluginConfig.getPluginsDirPath`.

But thinking more about it, as @TisonKun pointed out, ideally we should fail fast like it happens for `lib` atm. The `lib` checks do not fail because our dist build result always contains `lib` hard-set in `config.sh` (see `flink-dist/src/main/assemblies/bin.xml`). We should do the same for plugins and the docs should mention this directory as a standard place to put custom plugins in client dist. Then all existing and modified in this PR code paths should fail fast if the directory is set but does not exist including `PluginConfig.getPluginsDirPath`.

pnowojski:
Thanks for the reviews/comments @azagrebin and @TisonKun.

I would prefer to avoid treating anything related to this plugin directory as "obligatory". Some arguments:
- `lib` can not be empty for Flink to work properly, while `plugins` can
- if we make the contract of directory must exist when env variable is set, that would be cleaner, however it is much more risky. We have pretty weak test coverage when it comes to the deployments methods, so we could easily miss some place that mishandles the directory. If that happens, we brake existing setups, even when users didn't want to use `plugins` at all. 
- there are only 4 places in the code which are touching the env variable/plugins directory and we can make sure that they correctly handle non existing directory. While on the other hand, there are many many more code paths that are leading us to those 4 places in which we would have to be 100% sure that env variable and plugins directory is handled correctly, **even if the plugins directory is empty** - even if it's empty, it would still have to be shipped/copied/created
- "the failing fast" would improve user experience only if a user by passed our shell scripts and used a custom value for the plugin env variable - that's super rare
- "failing later" (as it is implemented now) means that user gets a warning that directory doesn't exist and he later would get an exception "FileSystem X was not found" or "MetricReporter Y was not found" 

So to sum up. "failing fast" is nicer, but it is outweighed in my opinion by too high risk that we brake some existing deployment (because of poor existing test coverage). With "failing later", true the failure happens later but the error is still easily understandable plus we are logging a warning.

After discussing this with @zentol and @azagrebin, I agree that one more missing thing here, is that we should create an empty `plugins` directory in `flink-dist`. In that way as long as users are using provided by us shell scripts (vast majority), they would be not affected by this "fail fast" vs "fail late" issue. Also it would be much more explicit and better user experience if he doesn't have to figure out how to name the root plugins directory, because it will already exist.

TisonKun:
>we should create an empty `plugins` directory in `flink-dist`

I think this is the right way :-)

pnowojski:
@TisonKun and @azagrebin, could you take a look at the two latest commits?

flink-dist/src/main/flink-bin/plugins/README.txt
```diff
@@ -0,0 +1,18 @@
...
+    |                |-plugin-a-2.jar
+    |                |-...
+    |                |-(the jars containing the classes of plugin a)
+    |
+    |------------plugin-b/
+    |                |-plugin-b-1.jar
+    |                |-...
+    ...
+
+Read more in the documentation about how to create a plugin.
```
TisonKun: Is there a link to related section in our documents? It would be better if we put it here.

flink-dist/src/main/assemblies/bin.xml
```diff
@@ -85,7 +85,14 @@ under the License.
 			<outputDirectory>bin</outputDirectory>
 			<fileMode>0755</fileMode>
 		</fileSet>
-		
+
+		<!-- copy plugins directory -->
+		<fileSet>
+			<directory>src/main/flink-bin/plugins</directory>
```
azagrebin: Side note: this should go to the final user doc where to put the plugins

azagrebin:
Thanks for addressing the comments @pnowojski ! LGTM 👍

pnowojski:
Test failure looks unrelated, merging.

PR #8760, opened 2019-06-17
[FLINK-12869] Add yarn acls capability to flink containers

ashangit:

*Provide yarn application acls mechanism  on flink containers  to be able to provide specific rights to other users than the one running the job (view logs through the resourcemanager/job history, kill the application)*

## Brief change log

  - *Add 2 parameter (yarn.view.acls and yarn.admin.acls and set ApplicationACLs for all containers*

## Verifying this change

  - *Manually verified the change by running a WordCount job on a yarn cluster with acls enabled. Verifying that without yarn.admin.acls a user not running the app can't kill the app. Then relaunching the app providing rights to an other user (setting yarn.admin.acls) and check that the specified user can kill the app. The same can be done for yarn.view.acls by checking that the user can now access the application overview from the resourcemanager (<resourcemanager_url>/cluster/app/<app_id>)*

  - The serializers: no
  - Anything that affects deployment or recovery: JobManager (and its components), Checkpointing, Yarn/Mesos, ZooKeeper: Affect acls on YARN (currently not managed so no rights can be added)

## Documentation


GJL:
@ashangit Do you still want to work on this?

ashangit:
@GJL will look at how I can add some test this week or the next

GJL:
@ashangit Alright, thanks for following up.

GJL:
Thanks for updating the PR. I'll have another look

GJL:
@flinkbot run travis

flink-yarn-tests/src/test/java/org/apache/flink/yarn/YarnTestBase.java
```diff
@@ -578,6 +579,17 @@ public static int getRunningContainers() {
 		return count;
 	}
 
+	public static HashMap<ContainerId, Map<ApplicationAccessType, String>> getRunningContainersAcls() {
+		HashMap<ContainerId, Map<ApplicationAccessType, String>> containers = new HashMap<>();
+		for (int nmId = 0; nmId < NUM_NODEMANAGERS; nmId++) {
+			NodeManager nm = yarnCluster.getNodeManager(nmId);
+			nm.getNMContext().getContainers().forEach((k, v) -> {
+				containers.put(k, v.getLaunchContext().getApplicationACLs());
```
GJL: I think this function should go into `YARNSessionFIFOSecuredITCase`. It's not likely that it will be used elsewhere.

Moreover, you are using a functional style (`forEach`) with side effects. However, I think when programming in a functional style, one should just return the result. For example:

```
	private static Map<ContainerId, Map<ApplicationAccessType, String>> getRunningContainersAcls() {
		return nodeManagersStream()
			.flatMap(toContainersStream())
			.collect(Collectors.toMap(
				Map.Entry::getKey,
				entry -> getApplicationACLs(entry.getValue())));
	}

	private static Stream<NodeManager> nodeManagersStream() {
		return IntStream
			.range(0, NUM_NODEMANAGERS)
			.mapToObj(i -> yarnCluster.getNodeManager(i));
	}

	private static Function<NodeManager, Stream<Map.Entry<ContainerId, Container>>> toContainersStream() {
		return nodeManager -> nodeManager.getNMContext().getContainers().entrySet().stream();
	}

	private static Map<ApplicationAccessType, String> getApplicationACLs(final Container container) {
		return container.getLaunchContext().getApplicationACLs();
	}
```

flink-yarn/src/main/java/org/apache/flink/yarn/Utils.java
```diff
@@ -263,6 +265,13 @@ private static LocalResource registerLocalResource(FileSystem fs, Path remoteRsr
 		return localResource;
 	}
 
+	public static void setAclsFor(ContainerLaunchContext amContainer, org.apache.flink.configuration.Configuration flinkConfig) {
+		amContainer.setApplicationACLs(new HashMap<ApplicationAccessType, String>(){{
```
GJL: Because double brace initialization [has some caveats](https://stackoverflow.com/a/1958961), I wouldn't use it.

GJL: I would feel more comfortable if we only set `ApplicationAccessType.VIEW_APP` and `ApplicationAccessType.MODIFY_APP`, if the user actually configured application ACLs. Also, can we make the default values of the new config options `null`? What do you think?

flink-yarn/src/main/java/org/apache/flink/yarn/configuration/YarnConfigOptions.java
```diff
@@ -204,6 +204,25 @@
...
+	public static final ConfigOption<String> APPLICATION_VIEW_ACLS =
+		key("yarn.view.acls")
+		.defaultValue("")
+		.withDescription("Users and groups to give VIEW acess. The ACLs are of for" +
+			" comma-separated-usersspacecomma-separated-groups");
+
+	/**
+	 * Users and groups to give MODIFY access.
+	 */
+	public static final ConfigOption<String> APPLICATION_ADMIN_ACLS =
```
GJL: I was thinking about renaming this to `APPLICATION_MODIFY_ACLS` so that it is aligned with Hadoop. What do you think?

flink-yarn-tests/src/test/java/org/apache/flink/yarn/YarnTestBase.java
```diff
@@ -578,6 +579,17 @@ public static int getRunningContainers() {
 		return count;
 	}
 
+	public static HashMap<ContainerId, Map<ApplicationAccessType, String>> getRunningContainersAcls() {
```
GJL: It's better to declare `Map<...>` as return type.

GJL:
@ashangit Can I get feedback on the comments I have left? I can finish the remaining work, if you do not have time to work on it.

GJL:
Closing due to inactivity. Feel free to reopen if you find time to work on this. As written above, I can also finalize it.

PR #8761, opened 2019-06-17
[FLINK-12842][network] Fix invalid check released state during ResultPartition#createSubpartitionView

zhijiangW:

*Currently in `ResultPartition#createSubpartitionView` it would check whether this partition is released before creating view. But this check is based on `refCnt != -1` which seems invalid, because the reference counter would not always reflect the released state.*

*In the case of `ResultPartition#release/fail`, the reference counter is not set to -1. Even if in the case of `ResultPartition#onConsumedSubpartition`, the reference counter seems also no chance to be -1.
So we could check the real `isReleased` state during creating view instead of reference counter.*

## Brief change log

  - *Remove check reference counter while `ReleaseOnConsumptionResultPartition#createSubpartitionView`*
  - *Add check `isReleased` state while `ResultPartition#createSubpartitionView`*
  - *Add tests in `ResultPartitionTest` and `ResultPartitionManagerTest`*

## Verifying this change

Adds new tests in `ResultPartitionTest#testCreateSubpartitionOnFailingPartition` and `ResultPartitionManagerTest#testCreateViewForReleasePartition`.

## Documentation

### DIFF
--- a/flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/PartitionTestUtils.java
+++ b/flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/PartitionTestUtils.java
@@ -20,6 +20,13 @@ package org.apache.flink.runtime.io.network.partition;
 
 import org.apache.flink.runtime.io.network.NettyShuffleEnvironment;
 
+import org.hamcrest.Matchers;
+
+import java.io.IOException;
+
+import static org.junit.Assert.assertThat;
+import static org.junit.Assert.fail;
+
 /**
  * This class should consolidate all mocking logic for ResultPartitions.
  * While using Mockito internally (for now), the use of Mockito should not
@@ -44,5 +51,17 @@ public class PartitionTestUtils {
 			.setResultPartitionType(partitionType)
 			.setNumberOfSubpartitions(numChannels)
 			.build();
+	}
+
+	static void verifyCreateSubpartitionViewThrowsException(
+			ResultPartitionManager partitionManager,
+			ResultPartitionID partitionId) throws IOException {
+		try {
+			partitionManager.createSubpartitionView(partitionId, 0, new NoOpBufferAvailablityListener());
+
+			fail("Should throw a PartitionNotFoundException.");
+		} catch (PartitionNotFoundException notFound) {
+			assertThat(partitionId, Matchers.is(notFound.getPartitionId()));
+		}
 	}
 }
--- a/flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/ResultPartitionManagerTest.java
+++ b/flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/ResultPartitionManagerTest.java
@@ -23,6 +23,8 @@ import org.apache.flink.util.TestLogger;
 import org.hamcrest.Matchers;
 import org.junit.Test;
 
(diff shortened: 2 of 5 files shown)


zhijiangW:
@flinkbot attention @zentol @pnowojski

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/ResultPartitionManagerTest.java
```diff
@@ -55,9 +49,24 @@ public void testThrowPartitionNotFoundException() throws Exception {
...
 		partitionManager.registerResultPartition(partition);
 		partitionManager.createSubpartitionView(partition.getPartitionId(), 0, new NoOpBufferAvailablityListener());
 	}
+
+	/**
+	 * Tests {@link ResultPartitionManager#createSubpartitionView(ResultPartitionID, int, BufferAvailabilityListener)}
+	 * would throw a {@link PartitionNotFoundException} if this partition was already released before.
+	 */
+	@Test
+	public void testCreateViewForReleasePartition() throws Exception {
```
zentol: ´...ForReleasedPartition`

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/ResultPartitionTest.java
```diff
@@ -217,6 +218,24 @@ public void testAddOnBlockingPartition() throws Exception {
 		testAddOnPartition(ResultPartitionType.BLOCKING);
 	}
 
+	/**
+	 * Tests {@link ResultPartition#createSubpartitionView(int, BufferAvailabilityListener)} would throw
```
zentol: this is incorrect. `ResultPartition#createSubpartitionView` is not throwing the `PartitionNotFoundException`, `ResultPartitionManager#createSubpartitionView` does since the partition is no longer registered.

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/PartitionTestUtils.java
```diff
@@ -45,4 +52,14 @@ public static ResultPartition createPartition(
 			.setNumberOfSubpartitions(numChannels)
 			.build();
 	}
+
+	static void verifyCreateSubpartitionView(ResultPartitionManager partitionManager, ResultPartition partition) throws IOException {
```
zentol: rename to `verifyCreateSubpartitionViewThrowsException`; consider passing in the result of `partition#getPartitionId()` instead.
The partition itself is not used, and as it stands the method is a a bit ambiguous since it could just as well call `ResultPartition#createSubpartitionView`

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/ResultPartitionTest.java
zhijiangW: Yes, my initial thought was when the registered partition was released to remove from `ResultPartitionManager`, then it would also throw `PartitionNotFoundException` during creating view in manger. I should adjust the comment properly.

zhijiangW:
Thanks for the reviews @zentol !
I have updated the codes for addressing above comments.

zhijiangW:
Thanks @zentol , it has already green.

PR #8763, opened 2019-06-17
[FLINK-12843][network] Refactor the pin logic in ReleaseOnConsumptionResultPartition

zhijiangW:

*The pin logic is for adding the reference counter based on number of subpartitions for `ReleaseOnConsumptionResultPartition`. It seems not necessary to do it in while loop as now, because the atomic counter would not be accessed by other threads during pin. If the `ReleaseOnConsumptionResultPartition` was not created yet, the `createSubpartitionView` would not be called actually resulting in`PartitionNotFoundException`. So we could simple increase the reference counter in `ReleaseOnConsumptionResultPartition` constructor directly.*

## Brief change log

  - *Remove `pin` method in `ResultPartition`*
  - *Remove `pin` method in `ReleaseOnConsumptionResultPartition` and adds the ref counter in constructor*

## Verifying this change

*Covered by `ReleaseOnConsumptionResultPartitionTest`.

## Documentation


zhijiangW:
@flinkbot attention @zentol @pnowojski

zhijiangW:
Thanks for review @zentol , the travis is already green.

PR #8765, opened 2019-06-17
[hotfix][streaming] Report correct value of currentProcessingTime in the exception

pnowojski:
This is a trivial logging change without test coverage.

## Documentation

### DIFF
--- a/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/operators/windowing/WindowOperator.java
+++ b/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/operators/windowing/WindowOperator.java
@@ -319,11 +319,14 @@ public class WindowOperator<K, IN, ACC, OUT, W extends Window>
 									"event-time window cannot become earlier than the current watermark " +
 									"by merging. Current watermark: " + internalTimerService.currentWatermark() +
 									" window: " + mergeResult);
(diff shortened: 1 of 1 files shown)


flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/operators/windowing/WindowOperator.java
```diff
@@ -319,11 +319,14 @@ public void merge(W mergeResult,
 									"event-time window cannot become earlier than the current watermark " +
 									"by merging. Current watermark: " + internalTimerService.currentWatermark() +
 									" window: " + mergeResult);
-						} else if (!windowAssigner.isEventTime() && mergeResult.maxTimestamp() <= internalTimerService.currentProcessingTime()) {
-							throw new UnsupportedOperationException("The end timestamp of a " +
+						} else {
```
aljoscha: This introduces an additional call to `internalTimerService.currentProcessingTime()`, which does a syscall, if I'm not mistaken, for the current time every time the else branch is taken.

We should probably do the `!windowAssigner.isEventTime()` check in the else and then retrieve the current processing time within the else block.

pnowojski:
Yes, good catch. Thanks for spotting this.

pnowojski:
Thanks for the quick review @aljoscha. Will merge it with green travis.

pnowojski:
Merged

PR #8766, opened 2019-06-17
[FLINK-12664][hive] Implement TableSink to write Hive tables

lirui-apache:

To implement a table sink to write Hive tables.

## Brief change log

  - Implemented HiveTableSink extending `OutputFormatTableSink`
  - Since some interfaces are still missing, HiveTableSink currently doesn't support static partition, partial dynamic partition, or insert overwrite.
  - Moved some test cases from HiveTableOutputFormatTest to HiveTableSinkTest. Once we have all the interfaces needed, we'll be able to move all the remaining cases.

## Verifying this change

Reuse existing test cases with some modifications.

  - The serializers: no

## Documentation

### DIFF
--- a/flink-connectors/flink-connector-hive/pom.xml
+++ b/flink-connectors/flink-connector-hive/pom.xml
@@ -48,7 +48,7 @@ under the License.
 
 		<dependency>
 			<groupId>org.apache.flink</groupId>
-			<artifactId>flink-table-api-java</artifactId>
+			<artifactId>flink-table-api-java-bridge_${scala.binary.version}</artifactId>
 			<version>${project.version}</version>
 			<scope>provided</scope>
 		</dependency>
@@ -365,6 +365,10 @@ under the License.
 					<artifactId>avatica</artifactId>
 				</exclusion>
 				<exclusion>
+					<groupId>org.apache.calcite</groupId>
+					<artifactId>calcite-avatica</artifactId>
+				</exclusion>
+				<exclusion>
 					<groupId>com.google.code.gson</groupId>
 					<artifactId>gson</artifactId>
 				</exclusion>
@@ -390,6 +394,13 @@ under the License.
 			<artifactId>flink-table-api-java</artifactId>
 			<version>${project.version}</version>
 			<type>test-jar</type>
+			<scope>test</scope>
+		</dependency>
+
+		<dependency>
+			<groupId>org.apache.flink</groupId>
+			<artifactId>flink-table-planner_${scala.binary.version}</artifactId>
+			<version>${project.version}</version>
 			<scope>test</scope>
 		</dependency>
 
--- a/flink-connectors/flink-connector-hive/pom.xml
+++ b/flink-connectors/flink-connector-hive/pom.xml
@@ -376,6 +376,10 @@ under the License.
 					<groupId>stax</groupId>
 					<artifactId>stax-api</artifactId>
 				</exclusion>
+				<exclusion>
+					<groupId>com.google.guava</groupId>
+					<artifactId>guava</artifactId>
+				</exclusion>
 			</exclusions>
 		</dependency>
 
--- a/flink-connectors/flink-connector-hive/pom.xml
+++ b/flink-connectors/flink-connector-hive/pom.xml
@@ -41,13 +41,6 @@ under the License.
 
 		<dependency>
 			<groupId>org.apache.flink</groupId>
-			<artifactId>flink-table-common</artifactId>
-			<version>${project.version}</version>
-			<scope>provided</scope>
-		</dependency>
-
-		<dependency>
-			<groupId>org.apache.flink</groupId>
 			<artifactId>flink-table-api-java-bridge_${scala.binary.version}</artifactId>
 			<version>${project.version}</version>
 			<scope>provided</scope>
--- a/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/batch/connectors/hive/HiveTableSink.java
+++ b/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/batch/connectors/hive/HiveTableSink.java
@@ -1,0 +1,159 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.batch.connectors.hive;
+
+import org.apache.flink.api.common.io.OutputFormat;
+import org.apache.flink.api.common.typeinfo.TypeInformation;
+import org.apache.flink.api.java.typeutils.RowTypeInfo;
+import org.apache.flink.table.catalog.exceptions.CatalogException;
+import org.apache.flink.table.catalog.hive.client.HiveMetastoreClientFactory;
+import org.apache.flink.table.catalog.hive.client.HiveMetastoreClientWrapper;
+import org.apache.flink.table.catalog.hive.client.HiveShimLoader;
+import org.apache.flink.table.catalog.hive.descriptors.HiveCatalogValidator;
+import org.apache.flink.table.sinks.OutputFormatTableSink;
+import org.apache.flink.table.sinks.TableSink;
+import org.apache.flink.types.Row;
+import org.apache.flink.util.FlinkRuntimeException;
+import org.apache.flink.util.Preconditions;
+
+import org.apache.flink.shaded.guava18.com.google.common.base.Joiner;
+
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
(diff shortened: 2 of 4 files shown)


lirui-apache:
cc @xuefuz @bowenli86 @zjuwangg for review
I don't think the CI build failure is related. Is there any way to rerun the build?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/batch/connectors/hive/HiveTableSink.java
```diff
@@ -0,0 +1,159 @@
...
+		hiveVersion = jobConf.get(HiveCatalogValidator.CATALOG_HIVE_VERSION, HiveShimLoader.getHiveVersion());
+	}
+
+	@Override
+	public OutputFormat<Row> getOutputFormat() {
+		boolean isPartitioned = partitionColumns != null && !partitionColumns.isEmpty();
+		// TODO: need PartitionableTableSink to decide
+		boolean isDynamicPartition = isPartitioned;
+		HiveTablePartition hiveTablePartition;
+		HiveTableOutputFormat outputFormat;
```
xuefuz: Nit: Can we move these to variable close to where they are instantiated to make the code more readable?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/batch/connectors/hive/HiveTableSink.java
```diff
@@ -0,0 +1,159 @@
...
+				hiveTablePartition = new HiveTablePartition(sd, null);
+			}
+			outputFormat = new HiveTableOutputFormat(jobConf, dbName, tableName,
+					partitionColumns, rowTypeInfo, hiveTablePartition, MetaStoreUtils.getTableMetadata(table), overwrite);
+		} catch (TException e) {
+			throw new CatalogException("Failed to query Hive metaStore", e);
+		} catch (IOException e) {
+			throw new FlinkRuntimeException("Failed to create staging dir", e);
+		}
+		return outputFormat;
```
xuefuz: Nit: we can move the return statment as the finally statement in above try block.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/batch/connectors/hive/HiveTableSink.java
```diff
@@ -0,0 +1,159 @@
...
+		this.tableName = tableName;
+		this.partitionColumns = partitionColumns;
+		hiveVersion = jobConf.get(HiveCatalogValidator.CATALOG_HIVE_VERSION, HiveShimLoader.getHiveVersion());
+	}
+
+	@Override
+	public OutputFormat<Row> getOutputFormat() {
+		boolean isPartitioned = partitionColumns != null && !partitionColumns.isEmpty();
+		// TODO: need PartitionableTableSink to decide
+		boolean isDynamicPartition = isPartitioned;
```
xuefuz: this means the table will be dynamically partitioned as long as it's partitioned table. Is this right? If that's not true, then we should set it to false until it's supported, right?

flink-connectors/flink-connector-hive/pom.xml
```diff
@@ -364,6 +364,10 @@ under the License.
 					<groupId>org.apache.calcite.avatica</groupId>
 					<artifactId>avatica</artifactId>
 				</exclusion>
+				<exclusion>
```
bowenli86: can you add these exclusions to pom of flink-sql-client too to keep them in sync?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/batch/connectors/hive/HiveTableSink.java
```diff
@@ -0,0 +1,159 @@
...
+import org.apache.flink.table.catalog.hive.client.HiveMetastoreClientWrapper;
+import org.apache.flink.table.catalog.hive.client.HiveShimLoader;
+import org.apache.flink.table.catalog.hive.descriptors.HiveCatalogValidator;
+import org.apache.flink.table.sinks.OutputFormatTableSink;
+import org.apache.flink.table.sinks.TableSink;
+import org.apache.flink.types.Row;
+import org.apache.flink.util.FlinkRuntimeException;
+import org.apache.flink.util.Preconditions;
+
+import org.apache.flink.shaded.guava18.com.google.common.base.Joiner;
```
bowenli86: we should try not depend on guava. Use `String.join()` instead?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/batch/connectors/hive/HiveTableSink.java
lirui-apache: That's true. Currently we don't have the interface to set static partition spec. Therefore we only support either non-partitioned or fully dynamic partition insert.

lirui-apache:
Thanks @xuefuz and @bowenli86 for the review. PR updated accordingly.

Regarding when `getOutputFormat()` is called, I think it's called in `TableEnvImpl::writeToSink`, which is not task-level.

bowenli86:
I don't have any other concerns. @xuefuz do you?

bowenli86:
@lirui-apache Thanks for your contribution.

Will merge once the CI passes

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/batch/connectors/hive/HiveTableSink.java
```diff
@@ -0,0 +1,155 @@
...
+
+	private final JobConf jobConf;
+	private final RowTypeInfo rowTypeInfo;
+	private final String dbName;
+	private final String tableName;
+	private final List<String> partitionColumns;
+	private final String hiveVersion;
+
+	// TODO: need OverwritableTableSink to configure this
+	private boolean overwrite = false;
```
synckey: lint: would you move the initiation of `overwrite` to constructor?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/batch/connectors/hive/HiveTableSink.java
```diff
@@ -0,0 +1,155 @@
...
+
+	public HiveTableSink(JobConf jobConf, RowTypeInfo rowTypeInfo, String dbName, String tableName,
+			List<String> partitionColumns) {
+		this.jobConf = jobConf;
+		this.rowTypeInfo = rowTypeInfo;
+		this.dbName = dbName;
+		this.tableName = tableName;
+		this.partitionColumns = partitionColumns;
+		hiveVersion = jobConf.get(HiveCatalogValidator.CATALOG_HIVE_VERSION, HiveShimLoader.getHiveVersion());
```
synckey: lint: would you also add a `this` before `hiveVersion`?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/batch/connectors/hive/HiveTableSink.java
```diff
@@ -0,0 +1,155 @@
...
+		try (HiveMetastoreClientWrapper client = HiveMetastoreClientFactory.create(new HiveConf(jobConf, HiveConf.class), hiveVersion)) {
+			Table table = client.getTable(dbName, tableName);
+			StorageDescriptor sd = table.getSd();
+			// here we use the sdLocation to store the output path of the job, which is always a staging dir
+			String sdLocation = sd.getLocation();
+			HiveTablePartition hiveTablePartition;
+			if (isPartitioned) {
+				// TODO: validate partition spec
+				// TODO: strip quotes in partition values
+				LinkedHashMap<String, String> strippedPartSpec = new LinkedHashMap<>();
```
synckey: Is there a reason to make strippedPartSpec `LinkedHashMap` not `Map`?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/batch/connectors/hive/HiveTableSink.java
lirui-apache: As the comment said, `overwrite ` is intended to be configured via the `OverwritableTableSink ` interface. Therefore it shouldn't be initialized in the constructor.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/batch/connectors/hive/HiveTableSink.java
lirui-apache: Could you explain what's the benefit of doing that?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/batch/connectors/hive/HiveTableSink.java
lirui-apache: Partition spec requires the partition columns in a specific order. So it's better to use `LinkedHashMap ` to explicitly indicate we need an ordered map here.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/batch/connectors/hive/HiveTableSink.java
synckey: `this` is used on the initialization of all the other fields except for hiveVersion, just for code lint.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/batch/connectors/hive/HiveTableSink.java
synckey: Makes sense.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/batch/connectors/hive/HiveTableSink.java
synckey: OK.

bowenli86:
Merging

PR #8768, opened 2019-06-17
[FLINK-12143] Fix file system dynamic class loading on create

1u0:

This PR adds a workaround for file system components that may involve dynamic class loading during `FileSystem` instantiation (via `FileSystemFactory.create()`) using plugins mechanism.
The plugin's class loader is set as thread's context class loader during invocation of `FileSystemFactory.create()`.

Note: an alternative approach can be to set the plugin class loader inside of concrete file system component's `FileSystemFactory.create()` implementation.

## Brief change log

  - The `Plugin` interface is extended to provide the class loader used to load the plugin;
  - For non-fallback file system factory,  the `factory.careate()` call in `FileSystem` is wrapped with the plugin's class loader set as thread's one;
  - The end-to-end fs tests are reverted to use the plugin mechanism.

## Verifying this change

This change is already covered by existing tests, such as `test_batch_wordcount.sh hadoop`
and can be verified as follows:

  - by successfully running the `test_batch_wordcount.sh hadoop|dummy-fs` using the `lib` loading mechanism;
  - by successfully running the `test_batch_wordcount.sh hadoop|dummy-fs` using the plugins' loading mechanism.

## Documentation

### DIFF
--- a/flink-core/src/main/java/org/apache/flink/core/plugin/Plugin.java
+++ b/flink-core/src/main/java/org/apache/flink/core/plugin/Plugin.java
@@ -20,6 +20,7 @@ package org.apache.flink.core.plugin;
 
 import org.apache.flink.annotation.PublicEvolving;
 import org.apache.flink.configuration.Configuration;
+import org.apache.flink.util.Preconditions;
 
 /**
  * Interface for plugins. Plugins typically extend this interface in their SPI and the concrete implementations of a
@@ -27,6 +28,18 @@ import org.apache.flink.configuration.Configuration;
  */
 @PublicEvolving
 public interface Plugin {
+
+	/**
+	 * Helper method to get the class loader used to load the plugin. This may be needed for some plugins that use
+	 * dynamic class loading afterwards the plugin was loaded.
+	 *
+	 * @return the class loader used to load the plugin.
+	 */
+	default ClassLoader getClassLoader() {
+		return Preconditions.checkNotNull(
+			this.getClass().getClassLoader(),
+			"%s plugin with null class loader", this.getClass().getName());
+	}
 
 	/**
 	 * Optional method for plugins to pick up settings from the configuration.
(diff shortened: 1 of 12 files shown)


flink-core/src/main/java/org/apache/flink/core/plugin/Plugin.java
```diff
@@ -21,13 +21,26 @@
...
  * service then implement the SPI contract.
  */
 @PublicEvolving
 public interface Plugin {
 
+	/**
+	 * Helper method to get the class loader used to load the plugin. This may be needed for some plugins that use
+	 * dynamic class loading afterwards the plugin was loaded.
+	 *
+	 * @return the class loader used to load the plugin. {@code null} value may represent the bootstrap class loader.
```
pnowojski: (re `null`) - I think this is impossible to happen. How could bootstrap class loader be responsible for loading a `Plugin` class? It would have to be provided by rt.jar

If I'm right, I would remove `@Nullable` annotation and add a checkState here `this.getClass().getClassLoader() != null`, to avoid `null` handling branches.

1u0:
> Have you checked if the tests are passing now?

Unfortunately, I can run only some end-to-end tests on my local machine. I've successfully run the `test_batch_wordcount` with `hadoop` and `dummy-fs` using `lib` and `plugins` loading mechanisms (you can read it in the PR description). The same test with `presto` fails on my local machine, but due to some other reasons.

> Travis build couldn't pass because of problems with our infrastructure at the moment.

Yep, but I'd still wait for a CI run. Also, the s3 end-to-end tests may pass on CI, if corresponding S3 credentials are not set in environment variables. If I understand correctly, this holds true for runs on PR hook, while on nightly, they would be run properly.

pnowojski:
Travis builds seem to be still constantly failing. I've restarted the builds one more time.

1u0:
@pnowojski, thanks for re-triggering CI run. I think that really helped.

If the changes in this PR look ok for you, I propose to still **postpone merging this** and wait until we have green nightly run on `master`.

In #8660 I wasn't expecting that other changes would effect cluster spin up. Merging this early, may conceal backwards compatibility issues of clusters with extended arguments without `plugins` dir.

PR #8769, opened 2019-06-17
[FLINK-12875][hive] support converting input args of char, varchar, bytes, timestamp, date for Hive functions

bowenli86:

This PR adds support for converting  input args of char, varchar, bytes, timestamp, date for Hive functions

## Brief change log

- adds support for converting  input args of char, varchar, bytes, timestamp, date for Hive functions
- adds corresponding unit tests

## Verifying this change

This change added tests and can be verified with new tests added in `HiveSimpleUDFTest`

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/functions/hive/conversion/HiveInspectors.java
+++ b/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/functions/hive/conversion/HiveInspectors.java
@@ -22,39 +22,61 @@ import org.apache.flink.annotation.Internal;
 import org.apache.flink.table.catalog.hive.util.HiveTypeUtil;
 import org.apache.flink.table.functions.hive.FlinkHiveUDFException;
 import org.apache.flink.table.types.DataType;
+import org.apache.flink.table.types.logical.CharType;
+import org.apache.flink.table.types.logical.VarCharType;
 
+import org.apache.hadoop.hive.common.type.HiveChar;
+import org.apache.hadoop.hive.common.type.HiveVarchar;
 import org.apache.hadoop.hive.serde2.io.ByteWritable;
+import org.apache.hadoop.hive.serde2.io.DateWritable;
 import org.apache.hadoop.hive.serde2.io.DoubleWritable;
+import org.apache.hadoop.hive.serde2.io.HiveCharWritable;
+import org.apache.hadoop.hive.serde2.io.HiveVarcharWritable;
 import org.apache.hadoop.hive.serde2.io.ShortWritable;
+import org.apache.hadoop.hive.serde2.io.TimestampWritable;
 import org.apache.hadoop.hive.serde2.objectinspector.ObjectInspector;
 import org.apache.hadoop.hive.serde2.objectinspector.PrimitiveObjectInspector;
+import org.apache.hadoop.hive.serde2.objectinspector.primitive.BinaryObjectInspector;
 import org.apache.hadoop.hive.serde2.objectinspector.primitive.BooleanObjectInspector;
 import org.apache.hadoop.hive.serde2.objectinspector.primitive.ByteObjectInspector;
+import org.apache.hadoop.hive.serde2.objectinspector.primitive.DateObjectInspector;
 import org.apache.hadoop.hive.serde2.objectinspector.primitive.DoubleObjectInspector;
 import org.apache.hadoop.hive.serde2.objectinspector.primitive.FloatObjectInspector;
+import org.apache.hadoop.hive.serde2.objectinspector.primitive.HiveCharObjectInspector;
+import org.apache.hadoop.hive.serde2.objectinspector.primitive.HiveVarcharObjectInspector;
 import org.apache.hadoop.hive.serde2.objectinspector.primitive.IntObjectInspector;
+import org.apache.hadoop.hive.serde2.objectinspector.primitive.JavaBinaryObjectInspector;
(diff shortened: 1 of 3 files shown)


bowenli86:
cc @JingsongLi @xuefuz @lirui-apache @zjuwangg

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/functions/hive/conversion/HiveInspectors.java
```diff
@@ -119,9 +141,45 @@ public static HiveObjectConversion getConversion(ObjectInspector inspector) {
...
+					return o -> new BytesWritable((byte[]) o);
+				} else {
+					return IdentityConversion.INSTANCE;
+				}
+			} else if (inspector instanceof JavaHiveCharObjectInspector) {
+				if (((JavaHiveCharObjectInspector) inspector).preferWritable()) {
+					return o -> new HiveCharWritable(
+						new HiveChar((String) o, ((CharType) dataType.getLogicalType()).getLength()));
+				} else {
+					return IdentityConversion.INSTANCE;
```
xuefuz: If o is a String (as the code shows), then identity conversion might not be expected because for char(p) type, Hive expects a HiveChar object instead of a string. However, I'm not 100% sure.

Same for varchar.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/functions/hive/conversion/HiveInspectors.java
```diff
@@ -162,34 +220,64 @@ public static Object toFlinkObject(ObjectInspector inspector, Object data) {
 
 				return oi.preferWritable() ?
 					oi.get(data) :
-					oi.getPrimitiveWritableObject(data);
+					oi.getPrimitiveJavaObject(data);
```
xuefuz: I'm not sure if we need to call oi.preferWritable() because we are converting Hive data to Flink data. Flink is always expecting java object. Thus, oi.getPrimitiveJavaObject() might be enough. I'm not 100% sure, though.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/functions/hive/conversion/HiveInspectors.java
bowenli86: You are right, I think it should be HiveChar instead

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/functions/hive/conversion/HiveInspectors.java
bowenli86: that's a good point. Let me read Hive source code one more time for this

bowenli86: I read the source code and did some more testing. HiveSimpleUDF and HiveGenericUDF don't need writables as we explicitly requires primitive return type in both of them. I'm not fully sure about UDTF and UDAF as we don't have them yet, nor does tests for them. 

I recommend keep them for now as they are not doing any harms. I've created "[FLINK-12891] evaluate converting hadoop/hive writable between Hive functions and Flink" to re-evaluate this issue once we have support for all hive functions in place in 1.9.

bowenli86:
@xuefuz I've updated the PR. Can you take another look?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/functions/hive/conversion/HiveInspectors.java
```diff
@@ -119,9 +141,45 @@ public static HiveObjectConversion getConversion(ObjectInspector inspector) {
 				} else {
 					return IdentityConversion.INSTANCE;
 				}
+			} else if (inspector instanceof JavaDateObjectInspector) {
+				if (((JavaDateObjectInspector) inspector).preferWritable()) {
```
xuefuz: Nit: preferWritable() is defined in PrimitiveObjectInspector. So, in all those if-else if branches, we don't need those casts if we just cast inspector to PrimitiveObjectInspector once on the top.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/functions/hive/conversion/HiveInspectors.java
xuefuz: Okay. I took a little deeper look and believe we only need to call getPrimitiveJavaObject() to get the java object, which can be further converted to the object that Flink expects. Even if we call getPrimitiveWritableObject() to get a writable object, we still need to convert to its java equivalent and then further convert it to the java object that Flink expects.

However, this can be taken care of in a separate PR.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/functions/hive/conversion/HiveInspectors.java
bowenli86: After doing some research,I think actually there is no need to check `preferWritable()`at all given input will be of primitive typed inspector. I will have a immediate following PR to take care of this

xuefuz: okay. you're probably right.

bowenli86:
Merging

PR #8771, opened 2019-06-18
[FLINK-12778][table] Fix derive row type bug for table aggregate

hequn8128:

This pull request fix derives row type bug for table aggregate.

Currently, we call `aggCalls.get(0).type.getFieldList.foreach(builder.add)` when derive row type for table aggregate. However, for types which are not composite types, the field list would be null. Table Aggregate should, of course, support non-composite types.

To solve the problem, we should judge whether types are structured. This is because a composite type will be converted to a RelDataType which contains field list and is structured.

## Brief change log

  - Fix `deriveTableAggRowType` method in `CommonTableAggregate`. When deriving row type, judge whether the result type of the table aggregate function is structured. Use `FieldInfoUtils.getFieldNames()` to get the field name when the result type is not structured.
  - In `AggregateOperationFactory.createAggregate()`, also use `FieldInfoUtils.getFieldNames()` to get the field name. 
  - Add tests to verify the results.

## Verifying this change

  - Add plan test for table aggregate(both for window and non-window table aggregate).

  - The serializers: (no)

## Documentation


sunjincheng121:
LGTM +1 to merged after CI change to green. 👍 
@flinkbot approve all

sunjincheng121:
I rebase the code just for the CI test. :) @hequn8128

PR #8687, opened 2019-06-11
[FLINK-12612][coordination] Track stored partition on the TaskExecutor 

zentol:
Based on ~~#8630~~, ~~#8654~~ and ~~#8680~~.

## What is the purpose of the change

Introduces partition tracking on the TaskExecutor.

We start tracking partitions after a Task has been started. Partitions
are only tracked iff.
a) they are not released on consumption (in this case the ShuffleEnvironment takes care of that automatically)
b) they have local resources (as otherwise there's nothing to release)

Partitions are tracked using a PartitionTable.

If a task fails we stop tracking any partitions for this task. No release calls will be issued, as
the task will fail their partition, which will automatically release them.

If task reaches the state FINISHED, no additional action is required. Partitions will continue to be tracked
until they are released at a later point.

The TaskExecutor is informed about the task termination via the tasks termination future, which also contains
the final execution state.

Partitions for finished tasks are only released in 2 cases:
a) An external call to TaskExecutor#releasePartitions is made.
b) The connection to a JobManager terminates for whatever reason.

For a) we stop tracking the given partitions and forward a release call to the ShuffleEnvironment.
No sanity checks are made; we assume that the whoever issued the release call has a reason to do so.

For b) we determine all partitions associated with the job corresponding to the terminated connection,
stop tracking them and issue a release call to the ShuffleEnvironment.

No additional cleanup is performed on shutdown, as the ShuffleEnvironment implicitly releases all
partitions requiring it.

Whenever a slot is freed the TaskExecutor was checking whether any other slots were allocated for a given job,
and if this isn't the case disconnects from the corresponding JobMaster.
It now additionally takes into account whether we still have any tracked partitions for the given job.

This check is also performed whenever an partition was released due to TaskExecutor#releasePartitions being called.

PR #8778, opened 2019-06-18
[FLINK-12615][coordination] Track partitions on JM

zentol:
Based on #8687 . When reviewing this PR, please ignore all commits preceding 
"[FLINK-12615][coordination] Support generic key in PartitionTable".

With this PR we track on the JobMaster side which partitions are still on task executors, maintain the connection to a task executor until all partitions on it have been released, and issue a release call for all partitions when a job terminates.

**Attention:** The release logic in PR only serves as a MVP. It ignores the shuffle master (and thus will not work with other shuffle service implementation) and is not viable for high-volume jobs (because it requires all blocking partitions to be persisted on disk until the job is finished).

## Brief changelog

6009f00 modifies the PartitionTable to support arbitrary keys; so far it was hard-wired to use JobIDs as keys for use on the TE side, but now we want to organize them by ResourceID (== TE ID) instead.

c7193dc introduces a PartitionTable into the JM, Scheduler and EG. This commit only modifies constructor-related methods and does not contain any actual logic. This commit only exists to make reviewing easier.

ea418da contains the tracking logic. Partitions are added to the table, grouped by task executor, by an Execution when it reaches the state FINISHED, and removed again by the JM when disassociating from a TE.

3087792 introduces a call to TE#releasePartitions for when a job is finished.

d5a9f3c finally modifies the check for empty taskmanagers to take partitions into account.

### DIFF
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/Execution.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/Execution.java
@@ -30,6 +30,7 @@ import org.apache.flink.runtime.checkpoint.CheckpointOptions;
 import org.apache.flink.runtime.checkpoint.CheckpointType;
 import org.apache.flink.runtime.checkpoint.JobManagerTaskRestore;
 import org.apache.flink.runtime.clusterframework.types.AllocationID;
+import org.apache.flink.runtime.clusterframework.types.ResourceID;
 import org.apache.flink.runtime.clusterframework.types.ResourceProfile;
 import org.apache.flink.runtime.clusterframework.types.SlotProfile;
 import org.apache.flink.runtime.concurrent.ComponentMainThreadExecutor;
@@ -39,6 +40,7 @@ import org.apache.flink.runtime.deployment.TaskDeploymentDescriptor;
 import org.apache.flink.runtime.deployment.TaskDeploymentDescriptorFactory;
 import org.apache.flink.runtime.execution.ExecutionState;
 import org.apache.flink.runtime.instance.SlotSharingGroupId;
+import org.apache.flink.runtime.io.network.partition.PartitionTracker;
 import org.apache.flink.runtime.io.network.partition.ResultPartitionID;
 import org.apache.flink.runtime.jobgraph.IntermediateDataSetID;
 import org.apache.flink.runtime.jobgraph.IntermediateResultPartitionID;
@@ -626,6 +628,7 @@ public class Execution implements AccessExecution, Archiveable<ArchivedExecution
 			vertex.getExecutionGraph().getJobMasterMainThreadExecutor(),
 			producedPartitionsCache -> {
 				producedPartitions = producedPartitionsCache;
+				startTrackingPartitions(location.getResourceID(), producedPartitionsCache.values());
 				return this;
 			});
 	}
@@ -811,7 +814,6 @@ public class Execution implements AccessExecution, Archiveable<ArchivedExecution
 			else if (current == FINISHED || current == FAILED) {
 				// nothing to do any more. finished/failed before it could be cancelled.
 				// in any case, the task is removed from the TaskManager already
-				sendReleaseIntermediateResultPartitionsRpcCall();
 
 				return;
 			}
@@ -844,8 +846,6 @@ public class Execution implements AccessExecution, Archiveable<ArchivedExecution
 				break;
 			case FINISHED:
 			case FAILED:
-				sendReleaseIntermediateResultPartitionsRpcCall();
-				break;
 			case CANCELED:
 				break;
 			default:
@@ -1170,6 +1170,8 @@ public class Execution implements AccessExecution, Archiveable<ArchivedExecution
 	private void finishCancellation() {
 		releaseAssignedResource(new FlinkException("Execution " + this + " was cancelled."));
 		vertex.getExecutionGraph().deregisterExecution(this);
+		// release partitions on TM in case the Task finished while we where already CANCELING
+		stopTrackingAndReleasePartitions();
 	}
 
 	void cachePartitionInfo(PartitionInfo partitionInfo) {
@@ -1228,6 +1230,7 @@ public class Execution implements AccessExecution, Archiveable<ArchivedExecution
 
 				releaseAssignedResource(t);
 				vertex.getExecutionGraph().deregisterExecution(this);
+				stopTrackingAndReleasePartitions();
 
 				if (!isCallback && (current == RUNNING || current == DEPLOYING)) {
 					if (LOG.isDebugEnabled()) {
@@ -1323,23 +1326,25 @@ public class Execution implements AccessExecution, Archiveable<ArchivedExecution
 		}
 	}
 
-	private void sendReleaseIntermediateResultPartitionsRpcCall() {
+	private void startTrackingPartitions(final ResourceID taskExecutorId, final Collection<ResultPartitionDeploymentDescriptor> partitions) {
+		PartitionTracker partitionTracker = vertex.getExecutionGraph().getPartitionTracker();
+		for (ResultPartitionDeploymentDescriptor partition : partitions) {
+			partitionTracker.startTrackingPartition(
+				taskExecutorId,
+				partition);
+		}
+	}
+
+	private void stopTrackingAndReleasePartitions() {
 		LOG.info("Discarding the results produced by task execution {}.", attemptId);
-		final LogicalSlot slot = assignedResource;
+		if (producedPartitions != null && producedPartitions.size() > 0) {
+			final PartitionTracker partitionTracker = getVertex().getExecutionGraph().getPartitionTracker();
+			final List<ResultPartitionID> producedPartitionIds = producedPartitions.values().stream()
+				.map(ResultPartitionDeploymentDescriptor::getShuffleDescriptor)
+				.map(ShuffleDescriptor::getResultPartitionID)
+				.collect(Collectors.toList());
 
-		if (slot != null) {
-			final TaskManagerGateway taskManagerGateway = slot.getTaskManagerGateway();
-
-			Collection<IntermediateResultPartition> partitions = vertex.getProducedPartitions().values();
-			Collection<ResultPartitionID> partitionIds = new ArrayList<>(partitions.size());
-			for (IntermediateResultPartition partition : partitions) {
-				partitionIds.add(new ResultPartitionID(partition.getPartitionId(), attemptId));
-			}
-
-			if (!partitionIds.isEmpty()) {
-				// TODO For some tests this could be a problem when querying too early if all resources were released
-				taskManagerGateway.releasePartitions(getVertex().getJobId(), partitionIds);
-			}
+			partitionTracker.stopTrackingAndReleasePartitions(producedPartitionIds);
 		}
 	}
 
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/ExecutionGraph.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/ExecutionGraph.java
@@ -51,6 +51,8 @@ import org.apache.flink.runtime.executiongraph.failover.RestartAllStrategy;
 import org.apache.flink.runtime.executiongraph.restart.ExecutionGraphRestartCallback;
 import org.apache.flink.runtime.executiongraph.restart.RestartCallback;
 import org.apache.flink.runtime.executiongraph.restart.RestartStrategy;
+import org.apache.flink.runtime.io.network.partition.PartitionTracker;
+import org.apache.flink.runtime.io.network.partition.PartitionTrackerImpl;
 import org.apache.flink.runtime.io.network.partition.ResultPartitionID;
 import org.apache.flink.runtime.jobgraph.IntermediateDataSetID;
 import org.apache.flink.runtime.jobgraph.JobStatus;
@@ -96,6 +98,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.NoSuchElementException;
+import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.CancellationException;
 import java.util.concurrent.CompletableFuture;
@@ -284,6 +287,8 @@ public class ExecutionGraph implements AccessExecutionGraph {
 	 * strong reference to any user-defined classes.*/
 	private volatile ErrorInfo failureInfo;
 
+	private final PartitionTracker partitionTracker;
+
 	/**
 	 * Future for an ongoing or completed scheduling action.
 	 */
@@ -409,7 +414,11 @@ public class ExecutionGraph implements AccessExecutionGraph {
 			blobWriter,
 			allocationTimeout,
 			NettyShuffleMaster.INSTANCE,
-			true);
+			true,
+			new PartitionTrackerImpl(
+				jobInformation.getJobId(),
+				NettyShuffleMaster.INSTANCE,
+				ignored -> Optional.empty()));
 	}
 
 	public ExecutionGraph(
@@ -425,7 +434,8 @@ public class ExecutionGraph implements AccessExecutionGraph {
 			BlobWriter blobWriter,
 			Time allocationTimeout,
 			ShuffleMaster<?> shuffleMaster,
-			boolean forcePartitionReleaseOnConsumption) throws IOException {
+			boolean forcePartitionReleaseOnConsumption,
+			PartitionTracker partitionTracker) throws IOException {
 
 		checkNotNull(futureExecutor);
 
@@ -476,6 +486,8 @@ public class ExecutionGraph implements AccessExecutionGraph {
 		this.shuffleMaster = checkNotNull(shuffleMaster);
 
 		this.forcePartitionReleaseOnConsumption = forcePartitionReleaseOnConsumption;
+
+		this.partitionTracker = checkNotNull(partitionTracker);
 
 		LOG.info("Job recovers via failover strategy: {}", failoverStrategy.getStrategyName());
 	}
@@ -1818,5 +1830,9 @@ public class ExecutionGraph implements AccessExecutionGraph {
 
 	ShuffleMaster<?> getShuffleMaster() {
 		return shuffleMaster;
+	}
+
+	public PartitionTracker getPartitionTracker() {
+		return partitionTracker;
 	}
 }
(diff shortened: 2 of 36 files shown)


flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/Execution.java
```diff
@@ -1061,6 +1063,20 @@ void markFinished(Map<String, Accumulator<?, ?>> userAccumulators, IOMetrics met
 							IntermediateResultPartition[] allPartitions = finishedPartition
 									.getIntermediateResult().getPartitions();
 
+							PartitionTable<ResourceID> partitionTable = vertex.getExecutionGraph().getPartitionTable();
```
azagrebin: We could consider introducing a separate component something like `PartitionTracker` which would also encapsulate `ShuffleMaster` (like `JobAwareShuffleEnvironmentImpl`). It could be responsible for doing what we have in this PR and maybe even register with the `ShuffleMaster` and track `producedPartitions` by execution id as well.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/Execution.java
```diff
@@ -1061,6 +1063,20 @@ void markFinished(Map<String, Accumulator<?, ?>> userAccumulators, IOMetrics met
...
+							for (IntermediateResultPartition partition : allPartitions) {
+								ResultPartitionDeploymentDescriptor descriptor = producedPartitions.get(partition.getPartitionId());
+								if (!descriptor.isReleasedOnConsumption()) {
```
azagrebin: At the end we should also take into account whether this partition actually occupies local resources (see https://issues.apache.org/jira/browse/FLINK-12890) otherwise TM and its connection can be released independently.

flink-runtime/src/main/java/org/apache/flink/runtime/jobmaster/JobMaster.java
```diff
@@ -861,6 +861,14 @@ private void jobStatusChanged(
 		validateRunsInMainThread();
 
 		if (newJobStatus.isGloballyTerminalState()) {
+			scheduledExecutorService.execute(() -> {
+				for (Map.Entry<ResourceID, Tuple2<TaskManagerLocation, TaskExecutorGateway>> entry : registeredTaskManagers.entrySet()) {
+					Collection<ResultPartitionID> storedPartitions = partitionTable.stopTrackingPartitions(entry.getKey());
+					// if this call fails TaskExecutors will cleanup partitions regardless once we close the connections
+					entry.getValue().f1.releasePartitions(jobGraph.getJobID(), storedPartitions);
```
azagrebin: This might be not needed then for cancel/suspend any more (Execution.sendReleaseIntermediateResultPartitionsRpcCall).

zentol:
@zhijiangW I don't see any comment of yours, did something go wrong when submitting the review?

zhijiangW:
I am not sure why they are not shown outside, but inline codes we could see these comments. I would double check and submit them again if necessary.

zentol:
I can only see andreys comments :(

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/ExecutionGraph.java
```diff
@@ -467,6 +473,8 @@ public ExecutionGraph(
 				"ExecutionGraph is not initialized with proper main thread executor. " +
 					"Call to ExecutionGraph.start(...) required.");
 
+		this.partitionTable = partitionTable;
```
zhijiangW: checkNotNull(partitionTable)

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/LegacyScheduler.java
```diff
@@ -169,12 +172,14 @@ public LegacyScheduler(
 		this.blobWriter = checkNotNull(blobWriter);
 		this.slotRequestTimeout = checkNotNull(slotRequestTimeout);
 
-		this.executionGraph = createAndRestoreExecutionGraph(jobManagerJobMetricGroup);
+		this.executionGraph = createAndRestoreExecutionGraph(jobManagerJobMetricGroup, partitionTable);
```
zhijiangW: checkNotNull(partitionTable)

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/Execution.java
zhijiangW: Maybe better to put the following logics in a separate method to make markFinished shorter.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/Execution.java
```diff
@@ -1061,6 +1063,20 @@ void markFinished(Map<String, Accumulator<?, ?>> userAccumulators, IOMetrics met
...
+									if (!taskManagerLocationFuture.isDone()) {
```
zhijiangW: The execution was already finished then it must have a assigned slot before, so I think it is not necessary here. Otherwise we should make this check in the beginning of markFinished, not in the process of starting tracking partitions.

flink-runtime/src/main/java/org/apache/flink/runtime/jobmaster/JobMaster.java
```diff
@@ -528,7 +535,11 @@ public void failSlot(
 
 	private void internalFailAllocation(AllocationID allocationId, Exception cause) {
 		final Optional<ResourceID> resourceIdOptional = slotPool.failAllocation(allocationId, cause);
-		resourceIdOptional.ifPresent(this::releaseEmptyTaskManager);
+		resourceIdOptional.ifPresent(taskManagerId -> {
+			if (!partitionTable.hasTrackedPartitions(taskManagerId)) {
```
zhijiangW: nit: if we put the if (!partitionTable.hasTrackedPartitions(taskManagerId)) inside the method of releaseEmptyTaskManager, we could reuse it if release might be reused in other parts future.

zhijiangW:
I entered the commit ids in above brief change log and left inline comments before. So these comments would only be seen inside them. I just re-submitted them in common way.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/Execution.java
zentol: yes we could do that. Not sure whether we should do this now or defer this to a later point. Currently we'd just end up wrapping the partition table and mirroring the API.

zentol:
rebase on #8789.

flink-runtime/src/test/java/org/apache/flink/runtime/jobmaster/JobMasterTest.java
```diff
@@ -1837,14 +1880,45 @@ private JobMaster createJobMaster(
 			new TestingOnCompletionActions());
 	}
 
+	@Nonnull
+	private JobMaster createJobMaster(
```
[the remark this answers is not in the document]
zentol: we actually do on master since 17.6 :D

flink-runtime/src/main/java/org/apache/flink/runtime/taskexecutor/partition/PartitionTable.java
```diff
@@ -0,0 +1,108 @@
...
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * Thread-safe Utility for tracking partitions.
+ */
+
```
gaoyunhaii: nit: this empty line can be removed

flink-runtime/src/test/java/org/apache/flink/runtime/jobmaster/JobMasterTest.java
```diff
@@ -1788,6 +1848,108 @@ private TestingResourceManagerGateway createAndRegisterTestingResourceManagerGat
...
+		final TestingTaskExecutorGateway testingTaskExecutorGateway = new TestingTaskExecutorGatewayBuilder()
+			.setDisconnectJobManagerConsumer((jobID, throwable) -> disconnectTaskExecutorFuture.complete(jobID))
+			.createTestingTaskExecutorGateway();
+
+		try {
+			jobMaster.start(jobMasterId).get();
+
+			final JobMasterGateway jobMasterGateway = jobMaster.getSelfGateway(JobMasterGateway.class);
+
+			// register a slow to establish a connection
```
gaoyunhaii: Is it "register a _slot_ to ... ?"

flink-runtime/src/main/java/org/apache/flink/runtime/jobmaster/JobMaster.java
```diff
@@ -876,6 +887,17 @@ private void jobStatusChanged(
 		validateRunsInMainThread();
 
 		if (newJobStatus.isGloballyTerminalState()) {
+			// other terminal job states are handled by the executions
+			if (newJobStatus == JobStatus.FINISHED) {
+				scheduledExecutorService.execute(() -> {
+					for (Map.Entry<ResourceID, Tuple2<TaskManagerLocation, TaskExecutorGateway>> entry : registeredTaskManagers.entrySet()) {
```
gaoyunhaii: Do we have concurrent modification problem on the `registeredTaskManagers` ? Here we iterates the map in the scheduler executor, while main thread of JM will remove entries from this map in the method calls of `onStop` -> `disConnectTaskManger`.

flink-runtime/src/main/java/org/apache/flink/runtime/jobmaster/JobMaster.java
```diff
@@ -876,6 +887,17 @@ private void jobStatusChanged(
 		validateRunsInMainThread();
 
 		if (newJobStatus.isGloballyTerminalState()) {
+			// other terminal job states are handled by the executions
+			if (newJobStatus == JobStatus.FINISHED) {
```
gaoyunhaii: From my understanding, I think we can only consider FINISHED due to that when transit to CANCELED and FAILED , all execution vertices will be canceled and the thus the result partition will be canceled via `sendReleaseIntermediateResultPartitionsRpcCall`. 

However, with region failover, an execution may need to re-run after it turns into FINISHED. This happens for cases like `A -> C, B->C`, A and B are all finished but C fails to read data from A, then B also need to re-execute if the partition is random (like RebalancePartition). When the execution vertex is `resetForNewExecution`, the result partition of previous execution is not released. Then when we `cancel` or `failGlobal` on the ExecutionGraph, it will also only consider the last execution, then some result partitions may be leaked.

zentol: failover scenarios are not covered by this PR; it only adds a bare-bones implementation for cleaning up after the job terminates.

You are correct that we have to release partitions for reset vertices, which will be added in a follow-up.

flink-runtime/src/main/java/org/apache/flink/runtime/jobmaster/JobMaster.java
zentol: yes, we should run this in the JM main thread instead.

flink-runtime/src/main/java/org/apache/flink/runtime/jobmaster/JobMaster.java
```diff
@@ -883,6 +883,17 @@ private void jobStatusChanged(
...
+				runAsync(() -> {
+					for (Map.Entry<ResourceID, Tuple2<TaskManagerLocation, TaskExecutorGateway>> entry : registeredTaskManagers.entrySet()) {
+						Collection<ResultPartitionID> storedPartitions = partitionTable.stopTrackingPartitions(entry.getKey());
+						// if this call fails TaskExecutors will cleanup partitions regardless once we close the connections
+						entry.getValue().f1.releasePartitions(jobGraph.getJobID(), storedPartitions);
```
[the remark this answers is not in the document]
zentol: ah yes, I got stuck here since the SM requires a ShuffleDescriptor, but so far we've only been tracking ResultPartitionIDs.
I have updated the PR to include a PartitionTracker which keeps track of shuffle descriptor, result partition ids etc. and acts as a central point for issuing release calls.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/PartitionTracker.java
```diff
@@ -0,0 +1,229 @@
...
+import java.util.Optional;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+import static java.util.stream.Collectors.toList;
+
+/**
+ * Utility for tracking partitions and issuing release calls to task executors and shuffle masters.
+ */
+public class PartitionTracker {
```
zentol: should introduce an interface for this

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/Execution.java
```diff
@@ -285,7 +281,13 @@ public LogicalSlot getAssignedResource() {
 
 	public Optional<ResultPartitionDeploymentDescriptor> getResultPartitionDeploymentDescriptor(
 			IntermediateResultPartitionID id) {
-		return Optional.ofNullable(producedPartitions.get(id));
+		Collection<ResultPartitionDeploymentDescriptor> trackedDeploymentDescriptors = vertex.getExecutionGraph().getPartitionTracker().getTrackedDeploymentDescriptors(getAttemptId());
+		for (ResultPartitionDeploymentDescriptor resultPartitionDeploymentDescriptor : trackedDeploymentDescriptors) {
```
zentol: I hate this bit, but really don't want to introduce an additional map _just_ to support this lookup. Any ideas?

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/PartitionTrackerFactory.java
```diff
@@ -0,0 +1,51 @@
...
+	 */
+	interface TaskExecutorGatewayLookup {
+
+		/**
+		 * Returns a {@link TaskExecutorGateway} corresponding to the given ResourceID.
+		 *
+		 * @param taskExecutorId id of the task executor to look up.
+		 * @return optional task executor gateway
+		 */
+		Optional<TaskExecutorGateway> lookup(ResourceID taskExecutorId);
```
azagrebin: I would prefer a partition releaser:
```
interface TaskExecutorPartitionReleaser {
    void releasePartitions(ResourceID taskExecutorId, JobID jobId, Collection<ResultPartitionID> partitionIds);
}
```
could also return boolean whether TE found, although not used anyways
do we need the full `TaskExecutorGateway` for something?

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/Execution.java
azagrebin: anyways I think the method rather belongs to the tracker with the exec id added to the signature

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/PartitionTrackerImpl.java
```diff
@@ -0,0 +1,231 @@
...
+		final ProducerDescriptor producerDescriptor,
+		final ResultPartitionDeploymentDescriptorFactory resultPartitionDeploymentDescriptorFactory) {
+
+		CompletableFuture<? extends ShuffleDescriptor> shuffleDescriptorFuture = shuffleMaster.registerPartitionWithProducer(partitionDescriptor, producerDescriptor);
+
+		return shuffleDescriptorFuture
+			.thenApply(shuffleDescriptor ->
+				resultPartitionDeploymentDescriptorFactory.createResultPartitionDeploymentDescriptor(partitionDescriptor, shuffleDescriptor))
+			.thenApply(deploymentDescriptor -> {
+				startTrackingPartition(producerDescriptor.getProducerLocation(), deploymentDescriptor);
```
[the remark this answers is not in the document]
azagrebin: We should have thought about a batch registration in shuffle master probably.. as we are joining on all of them anyways

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/ExecutionGraphBuilder.java
```diff
@@ -94,7 +94,8 @@ public static ExecutionGraph buildGraph(
 			BlobWriter blobWriter,
 			Time allocationTimeout,
 			Logger log,
-			ShuffleMaster<?> shuffleMaster) throws JobExecutionException, JobException {
+			PartitionTracker partitionTracker)
+		throws JobExecutionException, JobException {
```
azagrebin: keep formatting

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/Execution.java
```diff
@@ -1325,22 +1324,8 @@ private void sendCancelRpcCall(int numberRetries) {
...
-				partitionIds.add(new ResultPartitionID(partition.getPartitionId(), attemptId));
-			}
-
-			if (!partitionIds.isEmpty()) {
-				// TODO For some tests this could be a problem when querying too early if all resources were released
-				taskManagerGateway.releasePartitions(getVertex().getJobId(), partitionIds);
-			}
-		}
+		final PartitionTracker partitionTracker = getVertex().getExecutionGraph().getPartitionTracker();
+		partitionTracker.stopTrackingAndReleasePartitions(getAttemptId());
```
azagrebin: looks like we are still doing it in cancel and and suspend

flink-runtime/src/main/java/org/apache/flink/runtime/jobmaster/JobMaster.java
```diff
@@ -876,6 +889,15 @@ private void jobStatusChanged(
 		validateRunsInMainThread();
 
 		if (newJobStatus.isGloballyTerminalState()) {
+			// other terminal job states are handled by the executions
+			if (newJobStatus == JobStatus.FINISHED) {
+				runAsync(() -> {
+					for (Map.Entry<ResourceID, Tuple2<TaskManagerLocation, TaskExecutorGateway>> entry : registeredTaskManagers.entrySet()) {
```
azagrebin: `registeredTaskManagers.keySet().forEach(partitionTracker::stopTrackingAndReleaseAllPartitions);`

PR #8780, opened 2019-06-18
[travis] Increase travis cache timeout to 1200

sunjincheng121:
Increase Travis cache timeout to 1200, try to fix the follows error:

```
Cached flink dir /home/travis/flink_cache/37943/flink does not exist. Exiting build.
The command "./tools/travis_controller.sh core" exited with 1.
```
Detail can be found here: https://api.travis-ci.org/v3/job/547116795/log.txt

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ dist: xenial
 
 cache:
   # default timeout is too low
-  timeout: 600
+  timeout: 1200
   directories:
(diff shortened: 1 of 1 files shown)


zentol:
Increasing the cache timeout should have no effect on this problem; the download does succeed after all. It just downloads stuff that isn't up-to-date.

PR #8781, opened 2019-06-18
[FLINK-12874][table-common] Improve the semantics of zero length strings

twalthr:

See the JIRA issue and JavaDocs for more information.

## Brief change log

- Allow all kinds of zero-length strings
- Forbid serializability

## Verifying this change

- Updated `LogicalTypesTest`.

  - The serializers: no

## Documentation

### DIFF
--- a/flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/logical/BinaryType.java
+++ b/flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/logical/BinaryType.java
@@ -19,6 +19,7 @@
 package org.apache.flink.table.types.logical;
 
 import org.apache.flink.annotation.PublicEvolving;
+import org.apache.flink.table.api.TableException;
 import org.apache.flink.table.api.ValidationException;
 
 import java.util.Collections;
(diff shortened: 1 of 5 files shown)


flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/logical/BinaryType.java
```diff
@@ -91,9 +92,11 @@ private BinaryType(int length, boolean isNullable) {
 	 * For consistent behavior, the same logic applies to binary strings.
 	 *
 	 * <p>This method enables this special kind of binary string.
+	 *
+	 * <p>Zero-length binary strings have no serializable string representation.
 	 */
 	public static BinaryType ofEmptyLiteral() {
-		return new BinaryType(EMPTY_LITERAL_LENGTH, false);
+		return new BinaryType(EMPTY_LITERAL_LENGTH, true);
```
dawidwys: Is it possible to create a nullable empty literal? I think empty string literals by the contract of our API will always be non null.

twalthr: Agreed. I did this change because by default types in SQL are always nullable, but we don't need to do it for this special case.

JingsongLi: I prefer `ofEmptyLiteral(boolean isNullable)`.
Literal empty string will be empty literal non-nullable CharType.
But there are some nullable cases too.
Consider `substring('', 1, 3)`, I think we should return an empty literal nullable VarCharType in blink runner, Unlike Flink runner, Blink runner tries not to throw exceptions, which leads to task failure. Blink runner chooses return null to when error in calculation. (be consistent with Hive and Spark)

PR #3330, opened 2017-02-16
[FLINK-5795][TableAPI&SQL] Improve UDTF to support constructor with p…

sunjincheng121:
…arameter.

Thanks for contributing to Apache Flink. Before you open your pull request, please take the following check list into consideration.
If your changes take all of the items into account, feel free to open your pull request. For more information and/or questions please refer to the [How To Contribute guide](http://flink.apache.org/how-to-contribute.html).
In addition to going through the list, please provide a meaningful description of your changes.

- [×] General
  - The pull request references the related JIRA issue ("[FLINK-5795] Improve “UDTF" to support constructor with parameter.")
  - The pull request addresses only one issue
  - Each commit in the PR has a meaningful commit message (including the JIRA id)

- [ ] Documentation
  - Documentation has been added for new functionality
  - Old documentation affected by the pull request has been updated
  - JavaDoc for public methods has been added

- [×] Tests & Build
  - Functionality added by the pull request is covered by tests
  - `mvn clean verify` has been executed successfully locally or a Travis build has passed

PR #8782, opened 2019-06-18
[FLINK-12888] [table-planner-blink] Introduce planner rule to push filter into TableSource

godfreyhe:

*Introduce planner rule to push filter into TableSource*

## Brief change log

  - *remove FunctionCatalog in blink planner and use FunctionCatalog in flink-table-api-java*
  - *Added PushFilterIntoTableSourceScanRule that pushes filter into TableSource*
  - *Added RexNodeExtractor to extract conjunctive predicates to Expressions*

## Verifying this change

  - *Added PushFilterIntoTableSourceScanRuleTest that validates logical plan after PushFilterIntoTableSourceScanRule is applied*
  - *Added TableSourceTest that validates the physical plan for the filterable table source*
  - *Added TableSourceITCase that validates the execution result for the filterable table source*

## Documentation


flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/expressions/RexNodeConverter.java
```diff
@@ -168,6 +170,12 @@ private RexNode visitScalarFunc(UnresolvedCallExpression unresolvedCall) {
 			return relBuilder.call(FlinkSqlOperatorTable.MULTIPLY, child);
 		} else if (BuiltInFunctionDefinitions.MOD.equals(def)) {
 			return relBuilder.call(FlinkSqlOperatorTable.MOD, child);
+		} else if (def instanceof ScalarFunctionDefinition) {
+			// TODO add getName for ScalarFunctionDefinition ???
```
JingsongLi: It looks wrong code, should be:
```
ScalarFunction scalaFunc = ((ScalarFunctionDefinition) func).getScalarFunction();
			List<RexNode> child = convertCallChildren(call);
			SqlFunction sqlFunction = UserDefinedFunctionUtils.createScalarSqlFunction(
					scalaFunc.functionIdentifier(),
					scalaFunc.toString(),
					scalaFunc,
					typeFactory);
			return relBuilder.call(sqlFunction, child);
```

godfreyhe: thanks for the correction. 
`List<RexNode> child = convertCallChildren(call);` is defined at line 122

JingsongLi: Now we just use `functionIdentifier` to be name.
The detail reason is in https://github.com/apache/flink/pull/3330

godfreyhe: we can use `functionIdentifier` now, and fix it in another issue

godfreyhe:
@KurtYoung, @JingsongLi,  i have updated this pr based on comments. do you have any other suggestions? thanks

ps: i add another minor commit into this pr: remove FunctionCatalog in blink planner and use FunctionCatalog in flink-table-api-java

PR #8783, opened 2019-06-18
[FLINK-12863][FLINK-12865] Remove concurrency from HeartbeatManager(Sender)Impl

tillrohrmann:

This commit makes the HeartbeatManager implementations to use the RpcEndpoint's
main thread executor in order to remove concurrency. Furthermore, this commit changes the HeartbeatListener interface to directly return a payload instead of returning a future when
HeartbeatListener#retrievePayload is called.

Since the HeartbeatManager implementations now use the RpcEndpoint's main thread,
we remove the splicing into the RpcEndpoint's main thread in the implementations
of the HeartbeatListeners in the TaskExecutor, JobMaster and ResourceManager
components.

cc @azagrebin

## Brief change log

* Remove future executor from HeartbeatManager implementations
* Use the components main thread executor as the `ScheduledExecutor` for the `HeartbeatManagerImpl` in the `JobMaster`, `TaskExecutor` and `ResourceManager` component.
* Change return type of `HeartbeatListener#retrievePayload` to synchronously return the payload. This avoids another source of concurrency.

## Verifying this change

Exclude the last two commits and run `JobMasterTest#testAllocatedSlotReportDoesNotContainStaleInformation` as well as `TaskExecutorTest#testSlotReportDoesNotContainStaleInformation` in a loop. They should eventually fail.

  - The serializers: (no)

## Documentation


flink-runtime/src/test/java/org/apache/flink/runtime/heartbeat/HeartbeatManagerTest.java
```diff
@@ -491,24 +508,24 @@ public void reportPayload(ResourceID resourceID, Object payload) {
...
+				return specialResponse;
 			} else {
-				return CompletableFuture.completedFuture(defaultResponse);
+				return defaultResponse;
 			}
 		}
 	}
 
-	static class TestingHeartbeatListener implements HeartbeatListener<Object, Object> {
+	static class SimpleTestingHeartbeatListener implements HeartbeatListener<Object, Object> {
```
TisonKun: maybe de-duplicate by `extends TestingHeartbeatListener` and constructed by

```
super(
  resourceId -> future.complete(resourceID),
  (ignore1, ignore2) -> numberHeartbeatReports++,
  (ignore) -> CompletableFuture.completedFuture(payload)
);
this.payload = payload;
```

sunjincheng121:
I restart the CI, duo to the Travis EVN problem!

flink-runtime/src/test/java/org/apache/flink/runtime/heartbeat/TestingHeartbeatServices.java
```diff
@@ -18,43 +18,18 @@
...
-		Logger log) {
-
-		return new HeartbeatManagerSenderImpl<>(
-			heartbeatInterval,
-			heartbeatTimeout,
-			resourceId,
-			heartbeatListener,
-			scheduledExecutorToUse,
-			log);
+		this(1000L, 10000L);
```
azagrebin: seems this class is not needed any more or at least the comment is outdated

flink-runtime/src/test/java/org/apache/flink/runtime/taskexecutor/TaskExecutorTest.java
```diff
@@ -1898,4 +1998,22 @@ public SlotReport createSlotReport(ResourceID resourceId) {
 			return slotReports.poll();
 		}
 	}
+
+	private static final class AllocateSlotNotifyingTaskSlotTable extends TaskSlotTable {
+
+		private final OneShotLatch allocateSlotLatch;
+
+		private AllocateSlotNotifyingTaskSlotTable(Collection<ResourceProfile> resourceProfiles, TimerService<AllocationID> timerService, OneShotLatch allocateSlotLatch) {
```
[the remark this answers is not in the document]
tillrohrmann: True, will break the parameter list.

flink-runtime/src/main/java/org/apache/flink/runtime/heartbeat/HeartbeatManagerSenderImpl.java
```diff
@@ -36,54 +34,43 @@
...
-			HeartbeatListener<I, O> heartbeatListener,
-			ScheduledExecutor mainThreadExecutor,
-			Logger log) {
+	HeartbeatManagerSenderImpl(
+		long heartbeatPeriod,
+		long heartbeatTimeout,
+		ResourceID ownResourceID,
+		HeartbeatListener<I, O> heartbeatListener,
+		ScheduledExecutor mainThreadExecutor,
+		Logger log) {
```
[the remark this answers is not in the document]
tillrohrmann: Good point. Will correct it.

flink-runtime/src/test/java/org/apache/flink/runtime/heartbeat/TestingHeartbeatServices.java
tillrohrmann: I'll double check.

flink-runtime/src/test/java/org/apache/flink/runtime/heartbeat/HeartbeatManagerTest.java
tillrohrmann: This is not so easily possible because one cannot reference sub class fields when calling `super`.

TisonKun: stupid mistake :P

tillrohrmann: I removed this class by using the `TestingHeartbeatListener` directly.

tillrohrmann:
There is still a failing test case caused by my PR: https://api.travis-ci.org/v3/job/548180557/log.txt (see also https://issues.apache.org/jira/browse/FLINK-12895)

sunjincheng121:
Hi @tillrohrmann, Thanks for work on this PR!

Although this is the blocker that we released in 1.8.1, we still should not rush to merged. I think the update you submitted shows that you do not have a rest yesterday night. take care!  

I think it's also perfect If we finished the 1.8.1 release before the 1.9 feature freeze. So our time is still quite abundant! :)

Best, Jincheng

PR #8784, opened 2019-06-18
[FLINK-12892][table][hive] serialize catalog table to properties for table discovery service

bowenli86:

This PR enables serialization of catalog table to properties for table discovery service and table factory to initiate source/sink from those properties

## Brief change log

- implemented `toProperties()` in `AbstractCatalogTable` to serialize tables to properties
- added unit tests 

## Verifying this change

- `AbstractCatalogTableTest`

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogTable.java
+++ b/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogTable.java
@@ -65,11 +65,4 @@ public class HiveCatalogTable extends AbstractCatalogTable {
 		return Optional.ofNullable(getComment());
 	}
 
-	@Override
-	public Map<String, String> toProperties() {
-		// TODO: output properties that are used to auto-discover TableFactory for Hive tables.
-		Map<String, String> properties = new HashMap<>();
-		return properties;
-	}
-
 }
--- a/flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/GenericCatalogTable.java
+++ b/flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/GenericCatalogTable.java
@@ -54,13 +54,6 @@ public class GenericCatalogTable extends AbstractCatalogTable {
 	}
 
 	@Override
-	public Map<String, String> toProperties() {
-		// TODO: Filter out ANY properties that are not needed for table discovery.
-		Map<String, String> properties = new HashMap<>();
-		return properties;
-	}
-
-	@Override
 	public Optional<String> getDescription() {
 		return Optional.of(getComment());
 	}
--- a/flink-table/flink-table-common/src/main/java/org/apache/flink/table/descriptors/DescriptorProperties.java
+++ b/flink-table/flink-table-common/src/main/java/org/apache/flink/table/descriptors/DescriptorProperties.java
@@ -104,7 +104,10 @@ public class DescriptorProperties {
 	}
 
 	/**
-	 * Adds a properties map by appending the given prefix to element keys.
+	 * Adds a properties map by appending the given prefix to element keys with a dot.
+	 *
+	 * <p>For example: for prefix "flink" and a map of a single property with key "k" and value "v".
+	 * The added property will be as key "flink.k" and value "v".
 	 */
 	public void putPropertiesWithPrefix(String prefix, Map<String, String> prop) {
 		checkNotNull(prefix);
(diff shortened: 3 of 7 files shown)


bowenli86:

flink-table/flink-table-common/src/main/java/org/apache/flink/table/descriptors/DescriptorProperties.java
```diff
@@ -103,6 +103,18 @@ public void putProperties(DescriptorProperties otherProperties) {
 		}
 	}
 
+	/**
+	 * Adds a properties map by appending the given prefix to element keys.
+	 */
```
xuefuz: Complete the javadoc as this is going to be a public method. Same for get.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/descriptors/DescriptorProperties.java
```diff
@@ -103,6 +103,18 @@ public void putProperties(DescriptorProperties otherProperties) {
...
+	public void putPropertiesWithPrefix(String prefix, Map<String, String> prop) {
```
xuefuz: Maybe the '.' should be part of the prefix rather than being hardcoded here. Doing this should simplify both doc and code.

bowenli86: '.' is widely used in the descriptor. I'll add an example to the doc

flink-table/flink-table-common/src/main/java/org/apache/flink/table/descriptors/DescriptorProperties.java
bowenli86: You mean `@param`? The style of existing javadoc in this class seems not favor it

flink-table/flink-table-common/src/main/java/org/apache/flink/table/descriptors/DescriptorProperties.java
bowenli86: '.' is widely used in this class. I'll add an example to the doc to make it clearer

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/GenericCatalogTable.java
```diff
@@ -53,13 +53,6 @@ public GenericCatalogTable copy() {
 			getSchema().copy(), new ArrayList<>(getPartitionKeys()), new HashMap<>(getProperties()), getComment());
 	}
 
-	@Override
-	public Map<String, String> toProperties() {
-		// TODO: Filter out ANY properties that are not needed for table discovery.
```
xuefuz: Here TODO item says that we need to filter out any properties that are not needed for table factory discovery. One example that I can think of is GenericInMemoryCatalog.FLINK_IS_GENERIC_KEY.

I also remember that we prefixed some properties before storing them in HMS. Should we remove the prefix in toProperties()?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogTable.java
```diff
@@ -65,11 +65,4 @@ public CatalogBaseTable copy() {
 		return Optional.ofNullable(getComment());
 	}
 
-	@Override
-	public Map<String, String> toProperties() {
-		// TODO: output properties that are used to auto-discover TableFactory for Hive tables.
```
xuefuz: Since we will not use auto table factory discovery for HiveCatalogTable, we don't need to output anything. In other words, an empty map is sufficient.

We probably should keep this for now, until we combine HiveCatalogTable and GenericCatalogTable.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/GenericCatalogTable.java
bowenli86: the property prefix is taken care of already.

I will remove FLINK_IS_GENERIC_KEY. I don't think we need to come up with a sound list of properties to keep or remove right now, any changes can be made later on demand.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/descriptors/DescriptorProperties.java
xuefuz: then it's okay.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogTable.java
bowenli86: right, I'm actually working on combining the two tables now. Maybe we can just keep this PR as-is? there's not much difference to make it happen sooner than later

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/GenericCatalogTable.java
bowenli86: btw, I think FLINK_IS_GENERIC_KEY may be gone soon as we are going to use "type" field to define type of the table, like "hive", "csv", "kafka", etc.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalogTable.java
xuefuz: That's fine as long as you remember the point here. :)

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/GenericCatalogTable.java
xuefuz: well, probably not. I don't think there is going to be a generic property called "type"(, which is different from Blink). It's likely we will retain this key even after combining the two.

bowenli86:
@xuefuz Thanks for your review!

Merging

PR #8789, opened 2019-06-19
[FLINK-12890] Add partition lifecycle related Shuffle API

azagrebin:

At the moment we have `ShuffleEnvironment.releasePartitions` which is used to release locally occupied resources of partition. JM can also use it by calling `TaskExecutorGateway.releasePartitions`.

To support lifecycle management of partitions ([FLINK-12069](https://issues.apache.org/jira/browse/FLINK-12069), relevant mostly for batch and blocking partitions), we need to extend Shuffle API:

  - `ShuffleDescriptor.hasLocalResources` indicates that this partition occupies local resources on TM and requires TM running to consume the produced data (e.g. true for default `NettyShuffleEnviroment` and false for externally stored partitions). If a partition needs external lifecycle management and is not released after the first consumption is done (`ResultPartitionDeploymentDescriptor.isReleasedOnConsumption`), then RM/JM should keep TMs, which produce these partitions, running until partition still needs to be consumed. The connection to these TMs should also to be kept to issue the RPC call `TaskExecutorGateway.releasePartitions` once partition is not needed any more.

  - `ShuffleMaster.removePartitionExternally`: JM should call this whenever the partition does not need to be consumed any more. This call releases partition resources possibly occupied externally outside of TM and should not depend on `ShuffleDescriptor.hasLocalResources`.

## Brief change log

  - Introduce `ShuffleDescriptor.hasLocalResources` and default netty shuffle implementation
  - Introduce `ShuffleMaster.removePartitionExternally` and default netty shuffle implementation

## Verifying this change

Trivial shuffle interface extension.

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/deployment/ResultPartitionDeploymentDescriptor.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/deployment/ResultPartitionDeploymentDescriptor.java
@@ -24,9 +24,12 @@ import org.apache.flink.runtime.jobgraph.IntermediateDataSetID;
 import org.apache.flink.runtime.jobgraph.IntermediateResultPartitionID;
 import org.apache.flink.runtime.shuffle.PartitionDescriptor;
 import org.apache.flink.runtime.shuffle.ShuffleDescriptor;
+import org.apache.flink.runtime.shuffle.ShuffleEnvironment;
+import org.apache.flink.runtime.shuffle.ShuffleMaster;
 import org.apache.flink.runtime.state.KeyGroupRangeAssignment;
 
 import java.io.Serializable;
+import java.util.Collection;
 
 import static org.apache.flink.util.Preconditions.checkNotNull;
 
@@ -96,7 +99,9 @@ public class ResultPartitionDeploymentDescriptor implements Serializable {
 	 * Returns whether to release the partition after having been fully consumed once.
 	 *
 	 * <p>Indicates whether the shuffle service should automatically release all partition resources after
-	 * the first full consumption has been acknowledged.
+	 * the first full consumption has been acknowledged. This kind of partition does not need to be explicitly released
+	 * by {@link ShuffleMaster#releasePartitionExternally(ShuffleDescriptor)}
+	 * and {@link ShuffleEnvironment#releasePartitionsLocally(Collection)}.
 	 *
 	 * @return whether to release the partition after having been fully consumed once.
 	 */
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/deployment/ResultPartitionDeploymentDescriptor.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/deployment/ResultPartitionDeploymentDescriptor.java
@@ -92,6 +92,14 @@ public class ResultPartitionDeploymentDescriptor implements Serializable {
 		return sendScheduleOrUpdateConsumersMessage;
 	}
 
+	/**
+	 * Returns whether to release the partition after having been fully consumed once.
+	 *
+	 * <p>Indicates whether the shuffle service should automatically release all partition resources after
+	 * the first full consumption has been acknowledged.
+	 *
+	 * @return whether to release the partition after having been fully consumed once.
+	 */
 	public boolean isReleasedOnConsumption() {
 		return releasedOnConsumption;
 	}
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/NettyShuffleDescriptor.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/NettyShuffleDescriptor.java
@@ -25,6 +25,7 @@ import org.apache.flink.runtime.io.network.partition.ResultPartitionID;
 
 import java.io.Serializable;
 import java.net.InetSocketAddress;
+import java.util.Optional;
 
 /**
  * Default implementation of {@link ShuffleDescriptor} for {@link NettyShuffleMaster}.
@@ -55,6 +56,11 @@ public class NettyShuffleDescriptor implements ShuffleDescriptor {
 	@Override
 	public ResultPartitionID getResultPartitionID() {
 		return resultPartitionID;
+	}
+
+	@Override
+	public Optional<ResourceID> storesLocalResourcesOn() {
+		return Optional.of(producerLocation);
 	}
 
 	public boolean isLocalTo(ResourceID consumerLocation) {
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleDescriptor.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleDescriptor.java
@@ -62,7 +62,7 @@ public interface ShuffleDescriptor extends Serializable {
 	 * that the task executor is running and being connected to be able to consume the produced data. This is mostly
 	 * relevant for the batch jobs and blocking result partitions which should outlive the producer lifetime and
 	 * be released externally: {@link ResultPartitionDeploymentDescriptor#isReleasedOnConsumption()} is {@code false}.
-	 * {@link ShuffleEnvironment#releasePartitions(Collection)} can be used to release such kind of partitions locally.
+	 * {@link ShuffleEnvironment#releasePartitionsLocally(Collection)} can be used to release such kind of partitions locally.
 	 *
 	 * @return the resource id of the producing task executor if the partition occupies local resources there
 	 */
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleDescriptor.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleDescriptor.java
@@ -18,9 +18,13 @@
 
 package org.apache.flink.runtime.shuffle;
 
+import org.apache.flink.runtime.clusterframework.types.ResourceID;
+import org.apache.flink.runtime.deployment.ResultPartitionDeploymentDescriptor;
 import org.apache.flink.runtime.io.network.partition.ResultPartitionID;
 
 import java.io.Serializable;
+import java.util.Collection;
+import java.util.Optional;
 
 /**
  * Interface for shuffle deployment descriptor of result partition resource.
@@ -50,4 +54,17 @@ public interface ShuffleDescriptor extends Serializable {
 	default boolean isUnknown() {
 		return false;
 	}
+
+	/**
+	 * Returns the location of the producing task executor if the partition occupies local resources there.
+	 *
+	 * <p>Indicates that this partition occupies local resources in the producing task executor. Such partition requires
+	 * that the task executor is running and being connected to be able to consume the produced data. This is mostly
+	 * relevant for the batch jobs and blocking result partitions which should outlive the producer lifetime and
+	 * be released externally: {@link ResultPartitionDeploymentDescriptor#isReleasedOnConsumption()} is {@code false}.
+	 * {@link ShuffleEnvironment#releasePartitions(Collection)} can be used to release such kind of partitions locally.
+	 *
+	 * @return the resource id of the producing task executor if the partition occupies local resources there
+	 */
+	Optional<ResourceID> storesLocalResourcesOn();
 }
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleEnvironment.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleEnvironment.java
@@ -115,6 +115,9 @@ public interface ShuffleEnvironment<P extends ResultPartitionWriter, G extends I
 	/**
 	 * Release local resources occupied by the given partitions.
 	 *
+	 * <p>Relevant for partitions which occupy resources locally
+	 * (can be checked by {@link ShuffleDescriptor#storesLocalResourcesOn()}).
+	 *
 	 * @param partitionIds identifying the partitions to be released
 	 */
 	void releasePartitions(Collection<ResultPartitionID> partitionIds);
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleEnvironment.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleEnvironment.java
@@ -55,17 +55,20 @@ import java.util.Collection;
  * {@link ShuffleEnvironment#createResultPartitionWriters}. The created writers are grouped per owner.
  * The owner is responsible for the writers' lifecycle from the moment of creation.
  *
- * <p>Partitions are released in the following cases:
+ * <p>Partitions are fully released in the following cases:
  * <ol>
  *     <li>{@link ResultPartitionWriter#fail(Throwable)} and {@link ResultPartitionWriter#close()} are called
  *     if the production has failed.
  *     </li>
- *     <li>{@link ResultPartitionWriter#finish()} and {@link ResultPartitionWriter#close()} are called
- *     if the production is done. The actual release can take some time depending on implementation details,
- *     e.g. if the `end of consumption' confirmation from the consumer is being awaited implicitly
- *     or the partition is later released by {@link ShuffleEnvironment#releasePartitions(Collection)}.</li>
- *     <li>{@link ShuffleEnvironment#releasePartitions(Collection)} is called outside of the producer thread,
- *     e.g. to manage the lifecycle of BLOCKING result partitions which can outlive their producers.</li>
+ *     <li>if {@link ResultPartitionDeploymentDescriptor#isReleasedOnConsumption()} is {@code true} and
+ *     {@link ResultPartitionWriter#finish()} and {@link ResultPartitionWriter#close()} are called when the production is done.
+ *     The actual release can take some time depending on implementation details,
+ *     e.g. if the `end of consumption' confirmation from the consumer is being awaited implicitly.</li>
+ *     <li>if {@link ResultPartitionDeploymentDescriptor#isReleasedOnConsumption()} is {@code false} and
+ *     {@link ShuffleMaster#releasePartitionExternally(ShuffleDescriptor)} and {@link ShuffleEnvironment#releasePartitionsLocally(Collection)},
+ *     if it occupies any producer local resources ({@link ShuffleDescriptor#storesLocalResourcesOn()}),
+ *     are called outside of the producer thread, e.g. to manage the lifecycle of BLOCKING result partitions
+ *     which can outlive their producers.</li>
  * </ol>
  * The partitions, which currently still occupy local resources, can be queried with
  * {@link ShuffleEnvironment#getPartitionsOccupyingLocalResources}.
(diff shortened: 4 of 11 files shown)


azagrebin:
@flinkbot attention @tillrohrmann @zentol @zhijiangW

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/NettyShuffleDescriptor.java
```diff
@@ -57,6 +58,11 @@ public ResultPartitionID getResultPartitionID() {
 		return resultPartitionID;
 	}
 
+	@Override
+	public Optional<ResourceID> hasLocalResources() {
```
zentol: you know, whenever I see this method I always first assume that it returns a boolean.

tillrohrmann: maybe we could rename it to `storesLocalResourcesOn()`

flink-runtime/src/main/java/org/apache/flink/runtime/deployment/ResultPartitionDeploymentDescriptor.java
```diff
@@ -92,6 +92,14 @@ public boolean sendScheduleOrUpdateConsumersMessage() {
 		return sendScheduleOrUpdateConsumersMessage;
 	}
 
+	/**
+	 * Returns whether to release the partition after having been fully consumed once.
+	 *
+	 * <p>Indicates whether the shuffle service should automatically release all partition resources after
+	 * the first full consumption has been acknowledged.
+	 *
+	 * @return whether to release the partition after having been fully consumed once.
```
zhijiangW: It has the same information with above `Returns whether to release the partition after having been fully consumed once.`, so might be removed.

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/NettyShuffleDescriptor.java
zhijiangW: agree, it might bring confusing here.

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleEnvironment.java
```diff
@@ -115,6 +115,8 @@
 	/**
 	 * Release local resources occupied by the given partitions.
 	 *
+	 * <p>Relevant for partitions which have {@link ShuffleDescriptor#hasLocalResources()} returning {@code true}.
```
zhijiangW: `hasLocalResources` not return boolean value.

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleMaster.java
```diff
@@ -41,4 +44,17 @@
...
+	 *
+	 * <p>This call triggers release of any resources which are occupied by the given partition in the external systems
+	 * outside of the producer executor. This is mostly relevant for the batch jobs and blocking result partitions
+	 * for which {@link ResultPartitionDeploymentDescriptor#isReleasedOnConsumption()} returns {@code false}.
+	 * The producer local resources are managed by {@link ShuffleDescriptor#hasLocalResources()} and
+	 * {@link ShuffleEnvironment#releasePartitions(Collection)}.
+	 *
+	 * @param shuffleDescriptor shuffle descriptor of the result partition to release externally.
+	 */
+	void releasePartitionExternally(T shuffleDescriptor);
```
zhijiangW: I am a bit confused about the keyword `externally`.

IMO, we provide different ways for releasing partitions. One is based on consumption once which already exists. The other is based on JM's decision when to release a partition in batch job/blocking partition, so the `ShuffleMater` provides an interface for JM calling, then `ShuffleMater` communicates with relevant `ShuffleEnvironment` which already provides `releasePartition` to do so.

No matter which way we provide, the target is to release partition, the resources would be naturally freed as a result of partition release. For blocking partition case, it might also free some local resources in TaskExecutor, so I think we do not need to distinguish the external/internal resources here. Might only call `ShuffleMaster#releasePartition` and `ShuffleEnvironment#releasePartition` separately.

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleDescriptor.java
```diff
@@ -50,4 +54,17 @@
...
+	 *
+	 * <p>Indicates that this partition occupies local resources in the producing task executor. Such partition requires
+	 * that the task executor is running and being connected to be able to consume the produced data. This is mostly
+	 * relevant for the batch jobs and blocking result partitions which should outlive the producer lifetime and
+	 * be released externally: {@link ResultPartitionDeploymentDescriptor#isReleasedOnConsumption()} is {@code false}.
+	 * {@link ShuffleEnvironment#releasePartitions(Collection)} can be used to release such kind of partitions locally.
+	 *
+	 * @return the resource id of the producing task executor if the partition occupies local resources there
+	 */
+	Optional<ResourceID> hasLocalResources();
```
zentol: if there are local resources, does this imply that there aren't _any_ external resources? (Basically, I want to double-check when we have to issue release calls to the ShuffleMaster)

azagrebin: I would define it in a way that in general every shuffle service can potentially have some external resources if not then this is just an empty implementation. I would always call the external release in JM when the partition is not needed unless `releaseOnConsumption` is set in `ResultPartitionDeploymentDescriptor`.

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleMaster.java
zentol: I'm not sure if this signature will work due to generics.

For this to work one has to maintain the generic shuffle descriptor type from `registerPartitionWithProducer` until `releasePartitionExternally`. As it stands this isn't possible. The shuffle master in the EG is currently defined as `ShuffleMaster<?> shuffleMaster`.

zentol: An alternative could be to simply accept a `ResultPartitionID` instead. This would force some book-keeping into the implementations, but i don't see a solution right now.

azagrebin: True, we have to switch to `ShuffleDescriptor` type there because we do not keep static type information about concrete `ShuffleDescriptor` in JM.

flink-runtime/src/main/java/org/apache/flink/runtime/deployment/ResultPartitionDeploymentDescriptor.java
azagrebin: I wanted to keep jdoc format consistent.

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/NettyShuffleDescriptor.java
azagrebin: True, `storesLocalResourcesOn` fits better

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleMaster.java
azagrebin: Having generified `ShuffleMaster` is under question then, we can consider removing generics later..

azagrebin: We already explicitly assume that there can be some local TM resources to release for which we have to keep TM connection and special bookkeeping in TM (#8778 and `JobAwareShuffleEnvironment` in #8687). The way we are solving currently tracking of partitions to be released from JM in TM, we have to do the RPC local release anyways to update the `JobAwareShuffleEnvironment`. Then there is no need for extra internal communication between `ShuffleMaster` and `ShuffleEnvironment` for this purpose atm (this would also require a lot of effort w/o having `TaskManagerGateway`).

And we still have to be able to do the external release in future optimisation where we do not keep TM connection (no need for local release) in case of external shuffle service.

In this regard, I think it might be even better to rename `ShuffleEnvironment#releasePartitions` to `ShuffleEnvironment#releasePartitionsLocally` because this way it reflects how we intend to use it in JM/TM (final users of shuffle service). Of course, eventually shuffle service can do anything internally including extra communication between `ShuffleMater` and `ShuffleEnvironment` or doing local/external cleanup at the same time.

azagrebin:
Thanks for the reviews @zhijiangW @zentol ! I addressed comments. We can further discuss `releasePartitionsXXX` naming.

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleMaster.java
zhijiangW: Thanks for the explanation! 
If I understood correctly, the local release is for RPC JM->TM and external release is for RPC `ShuffleMaster`->`ShuffleEnvironment`. ATM we only have one internal shuffle implementation, then the `ShuffleEnvironment#releasePartitionsLocally` is actually used but `ShuffleMaster#releasePartitionsExternally` might never be used.  But for extending external shuffle implementation future, I am not sure how these two release methods would be used then.

My only concern is wondering this abstraction might bring confusing for future extending implementations, because we seem give some specific tags/limitations in general interface. We only need to provide the semantic/ability for the function, no need to reflect the backend detail implementations (`TaskManagerGateway`, `TM connection`).  

We already provide one simple implementation via JM/TM for release and also define the interface between `ShuffleMaster/ShuffleEnvironment` for release. The general `ShuffleMaster#releasePartition` is only for indicating that `ShuffleMaster` should release given partitions, internal/external shuffle implementations could both rely on it or not.

E.g. the current `releaseOnConsumption` could also be implemented like this: when JM receives the notification of finished consumer task, it notifies `ShuffleMaster` to release corresponding producer's partitions.

azagrebin: The current idea is that (both for `not RPD.releaseOnConsumption`):

- the local release is for RPC JM->TM (only if `SD.storesLocalResourcesOn`): cleans up `JobAwareShuffleEnvironment` wrapper and calls `ShuffleEnvironment#releasePartitionsLocally`.

- the external release is for RPC `ShuffleMaster` -> external system when the consumption is done, e.g. auxiliary service on yarn node. It is not for `ShuffleEnvironment` because TM yarn container can be already released after the production (but not consumption) is done for external shuffle service (does not `SD.storesLocalResourcesOn`). It is not needed for Netty implementation (empty) but has to be called for external shuffle service w/o JM/TM RPC (no TM/ShuffleEnvironment connection).

So in both cases we call the methods for certain semantic purpose.

We need the semantic difference in these methods because we already assume that there are TM local and external systems, and JM/TM treat the cases differently. It would be nice to have only one call, e.g. `ShuffleMaster.releasePartitions` and let it decide what to call/release but then it would have to implement RPC internally and also we need to do the JM/TM RPC anyways to manage `JobAwareShuffleEnvironment` wrapper. Having declared the intentions in method names should actually simplify the shuffle implementations.

`releaseOnConsumption` currently means that shuffle service internally monitors end of consumption (like Netty implementation always did) and takes care about release automatically without any action required by JM/TM. True, in future we could change it that it is managed only by JM/TM directly, I think we keep it for safety until the fine-grained release is stable.

azagrebin: | SD.storesLocalResourcesOn TM -> | Yes | No
---------------------------|----|----
| **PD.releaseOnConsumption \\\|/** |-|-
|**True**|Auto internal release, no action required|Auto internal release, no action required
|**False**|Track partitions in JM/TM, keep JM/TM connection until consumption is done, call JM/TM RPC for TM local and ShuffleMaster for external release when consumption is done|Track partitions in JM, close JM/TM connection when production is done, call only ShuffleMaster to release when consumption is done

zhijiangW: Thanks for the above detail explanations and I could understand the functions/semantics we want to provide. I think we provide three ways for releasing partitions atm as below table:

Partition release | Interface dependencies
-- | --
1.release once consumed  |  PD#releaseOnConsumption
2.TM internal shuffle RPC-based  | JM->TM RPC,  ShuffleEnvironment#releasePartitionsLocally,  SD#storesLocalResourcesOn
3.external shuffle RPC-base  | ShuffleMaster#releasePartitionsExternally, SD#storesLocalResourcesOn

My previous thought was considering whether the interface and proposed methods are clear enough for the third-party implementation of `ShuffleManager`. We could see from above table that there are four interface methods involved in partition release (PD, SD, ShuffleMaster, ShuffleEnvironment) and some methods are coupled with each other.

E.g. we assume if `SD#storesLocalResourcesOn` is true, `ShuffleEnvironment#releasePartitionsLocally` should be used for releasing partition, otherwise `ShuffleMaster#releasePartitionsExternally` should be used. Actually we both agree that for the above case 2 in table, we could also rely on `ShuffleMaster#releasePartitionsExternally` to release partition if we want, and only we rely on JM/TM RPC as shortcut default implementation. 

But if somebody wants to implement another `ShuffleManager` to replace the current one for case 2, he might be confused of why `ShuffleMaster#releasePartitionsExternally` was not suggested using and what is `Externally` indicating for. If the interface was only defined as `ShuffleMaster#releasePartitions`, it does not need to explain the differences between externally and locally which might bring confusing sometimes. It only indicates there triggers a release call from `ShuffleMaster`, the third-party could decide whether to implement it or empty via other ways.

Also for `ShuffleEnvironment#releasePartition` it only indicates to trigger a partition release call from `ShuffleEnvironment`, no need to limit/emphasize it for a certain specific implementation, although the external case 3 might never call it in implementation.

Actually from above table we could see we need two boolean tag infos atm. One is for describing the partition release mechanism, and the other is for indicating TM internal shuffle or external shuffle. I think these two infos might be both tagged in `ShuffleDescriptor` in future. Because we assume the pipelined partition could be consumed only once limited by current implementation, so we tag it in PD#releaseOnConsumption. But whether a partition could be consumed multiple times is also up to the implementation of `ShuffleEnvironment#createResultPartitionWriters`, we could also implement a pipelined partition which could be consumed multi times. So the SD could consider the partition type with internal implementation to give a final tag.

azagrebin: Thanks for more explanation! I am still not sure that I fully understand the suggested final way to define this part of interface.

If we define both methods as `ShuffleMaster#releasePartitions` and `ShuffleEnvironment#releasePartitions` without stating the purpose, it can look for an implementer like any of them can be used for full release and it has to be implemented this way which is currently not true, e.g. for the default netty implementation. Basically, the existing netty implementation will not comply with the interface or we will have to make an implicit assumption in JM/TM about how to use shuffle service just for Netty which does a partial release in case of each method call. True, the methods are coupled at the moment which is not ideal but to make them more flexible, all implementations have to be flexible.

`PD#releaseOnConsumption` is an information from JM (user of shuffle service) about how the partition is intended to be used, I think we should actually move it to the `PartitionDescriptor`, not `ShuffleDescriptor`. If `releaseOnConsumption` is not supported by shuffle service for certain partition type, it should actually throw an exception already in `ShuffleMaster#registerPartitionWithProducer`.

Do I understand correctly your suggestion or it should be defined differently?

zhijiangW: Yes, my previous suggestion was defined the methods as `ShuffleMaster#releasePartitions` and `ShuffleEnvironment#releasePartitions`. In default netty implementation we actually do not use this path for releasing partitions.

For other extending new implementations(no matter external or other internal), users could decide whether to use this path or not. If use, it should adjust to let JM call `ShuffleMaster#releasePartition`, not communicate directly with TM as now.

I am not strongly insisting on this suggestion, just pointed out some personal concerns from first impression. And I think this summary-up could help deeply understand the full stack via discussion. :)

Also agree with your issue of  `PD#releaseOnConsumption`,  the partition release strategy could be abstracted to pass into `PartitionDescriptor` and `ShuffleMaster` could check whether the current implementation supports it.

flink-runtime/src/test/java/org/apache/flink/runtime/deployment/ResultPartitionDeploymentDescriptorTest.java
```diff
@@ -109,7 +110,7 @@ public void testReleasedOnConsumptionFlag() {
 		for (ResultPartitionType partitionType : ResultPartitionType.values()) {
 			ResultPartitionDeploymentDescriptor partitionDescriptor = new ResultPartitionDeploymentDescriptor(
 				new PartitionDescriptor(resultId, partitionId, partitionType, numberOfSubpartitions, connectionIndex),
-				ResultPartitionID::new,
+				NettyShuffleDescriptorBuilder.newBuilder().buildLocal(),
```
zhijiangW: it might be a separate hotfix commit.

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/ResultPartitionFactoryTest.java
```diff
@@ -64,7 +65,7 @@ private static void testForceConsumptionOnRelease(boolean forceConsumptionOnRele
 				ResultPartitionType.BLOCKING,
 				1,
 				0),
-			ResultPartitionID::new,
+			NettyShuffleDescriptorBuilder.newBuilder().buildLocal(),
```
zhijiangW: ditto

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleMaster.java
azagrebin: I agree the problem is not easy and the resulting tables in our discussion should really help the understanding :)

We could also define that the shuffle user (JM/TM) has to guarantee to 
always call both methods `ShuffleMaster#releasePartitions` and `ShuffleEnvironment#releasePartitions` then an implementer could decide which cleanup to do in which method but this is again not true because once the TM is gone, `ShuffleEnvironment#releasePartitions` cannot be called.

`ShuffleMaster` can already communicate directly to `ShuffleEnvironment`. It just does not have to for only `ShuffleEnvironment#releasePartitionsLocally` and the naming states it clearly. If `SD#storesLocalResourcesOn` then the shuffle user (JM/TM) guarantees doing the RPC.

flink-runtime/src/test/java/org/apache/flink/runtime/deployment/ResultPartitionDeploymentDescriptorTest.java
azagrebin: The target change is not possible without this small change.

zentol:
As far as I can tell the only remaining discussion here is about the exact naming of `releasePartition` methods.

Given that the _existence_ of these methods is required for other PRs (#8687, #8778) I would still like to go ahead with the merge, and defer further discussions to a later date.

Overall it looks to me like we need better documentation about how shuffle services can expect Flink to behave, and vice-versa.

azagrebin:
I added a bit more details to docs

flink-runtime/src/test/java/org/apache/flink/runtime/deployment/ResultPartitionDeploymentDescriptorTest.java
zhijiangW: yes, I misunderstood before.

zhijiangW:
Thanks for the further reviews and I have rebased to solve the conflicts @azagrebin !

PR #8791, opened 2019-06-19
[FLINK-12903][py] Remove old Python APIs

zentol:
This PR removes all traces of the old batch/streaming Python APIs.


zentol:
opened from the wrong branch.

PR #8792, opened 2019-06-19
[FLINK-12743][table-runtime-blink] Introduce unbounded streaming anti/semi join operator

wuchong:

This pull request supports unbounded streaming semi/anti join.

## Brief change log

- Introduce `StreamingSemiAntiJoinOperator` operator to support semi/anti join
- Introduce `AbstractStreamingJoinOperator` to put shared member fields and utility for `StreamingSemiAntiJoinOperator` and `StreamingJoinOperator`

## Verifying this change

- Added integration tests for unbounded semi/anti join

## Documentation

### DIFF
--- a/flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/runtime/join/stream/StreamingSemiAntiJoinOperator.java
+++ b/flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/runtime/join/stream/StreamingSemiAntiJoinOperator.java
@@ -160,13 +160,15 @@ public class StreamingSemiAntiJoinOperator extends AbstractStreamingJoinOperator
 						if (isAntiJoin) {
 							// send -[other]
 							other.setHeader(BaseRowUtil.RETRACT_MSG);
+							collector.collect(other);
+							// set header back to ACCUMULATE_MSG, because we will update the other row to state
+							other.setHeader(BaseRowUtil.ACCUMULATE_MSG);
 						} else {
 							// send +[other]
-							other.setHeader(BaseRowUtil.ACCUMULATE_MSG);
+							// the header of other is ACCUMULATE_MSG
+							collector.collect(other);
 						}
-						collector.collect(other);
 					} // ignore when number > 0
-					other.setHeader(BaseRowUtil.ACCUMULATE_MSG);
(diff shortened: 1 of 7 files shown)


flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/runtime/join/stream/StreamingSemiAntiJoinOperator.java
```diff
@@ -0,0 +1,196 @@
...
+						if (isAntiJoin) {
+							// send -[other]
+							other.setHeader(BaseRowUtil.RETRACT_MSG);
+						} else {
+							// send +[other]
+							other.setHeader(BaseRowUtil.ACCUMULATE_MSG);
+						}
+						collector.collect(other);
+					} // ignore when number > 0
+					other.setHeader(BaseRowUtil.ACCUMULATE_MSG);
```
JingsongLi: why set acc to other? Will it be used again?

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/runtime/join/stream/StreamingSemiAntiJoinOperator.java
```diff
@@ -0,0 +1,196 @@
...
+						if (!isAntiJoin) {
+							// send -[other]
+							other.setHeader(BaseRowUtil.RETRACT_MSG);
+						} else {
+							// send +[other]
+							other.setHeader(BaseRowUtil.ACCUMULATE_MSG);
+						}
+						collector.collect(other);
+					} // ignore when number > 0
+					other.setHeader(BaseRowUtil.ACCUMULATE_MSG);
```
JingsongLi: same question as above.

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/runtime/join/stream/StreamingSemiAntiJoinOperator.java
wuchong: Header of `other` may be set to `RETRACT` before, so we have to set it back when storing it in state. All rows stored in state are all accumulate BaseRows.

I did a minor adjust for this to move this line close to setting retract header, and add a comment, to make it more readable.

JingsongLi:
LGTM +1

PR #8804, opened 2019-06-20
[FLINK-12883][WIP][runtime] Add elaborated partition release logic

GJL:

*WIP*

## Brief change log

  - *See commits*

## Verifying this change

  - *Added unit tests, such as `RegionPartitionReleaseStrategyTest`*

## Documentation

### DIFF
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/ExecutionGraph.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/ExecutionGraph.java
@@ -48,6 +48,9 @@ import org.apache.flink.runtime.execution.ExecutionState;
 import org.apache.flink.runtime.execution.SuppressRestartsException;
 import org.apache.flink.runtime.executiongraph.failover.FailoverStrategy;
 import org.apache.flink.runtime.executiongraph.failover.RestartAllStrategy;
+import org.apache.flink.runtime.executiongraph.failover.adapter.DefaultFailoverTopology;
+import org.apache.flink.runtime.executiongraph.failover.flip1.partitionrelease.NotReleasingPartitionReleaseStrategy;
+import org.apache.flink.runtime.executiongraph.failover.flip1.partitionrelease.PartitionReleaseStrategy;
 import org.apache.flink.runtime.executiongraph.restart.ExecutionGraphRestartCallback;
 import org.apache.flink.runtime.executiongraph.restart.RestartCallback;
 import org.apache.flink.runtime.executiongraph.restart.RestartStrategy;
@@ -55,6 +58,7 @@ import org.apache.flink.runtime.io.network.partition.PartitionTracker;
 import org.apache.flink.runtime.io.network.partition.PartitionTrackerImpl;
 import org.apache.flink.runtime.io.network.partition.ResultPartitionID;
 import org.apache.flink.runtime.jobgraph.IntermediateDataSetID;
+import org.apache.flink.runtime.jobgraph.IntermediateResultPartitionID;
 import org.apache.flink.runtime.jobgraph.JobStatus;
 import org.apache.flink.runtime.jobgraph.JobVertex;
 import org.apache.flink.runtime.jobgraph.JobVertexID;
@@ -66,6 +70,11 @@ import org.apache.flink.runtime.jobmanager.scheduler.NoResourceAvailableExceptio
 import org.apache.flink.runtime.jobmaster.slotpool.Scheduler;
 import org.apache.flink.runtime.jobmaster.slotpool.SlotProvider;
 import org.apache.flink.runtime.query.KvStateLocationRegistry;
+import org.apache.flink.runtime.scheduler.adapter.ExecutionGraphToSchedulingTopologyAdapter;
+import org.apache.flink.runtime.scheduler.strategy.ExecutionVertexID;
+import org.apache.flink.runtime.scheduler.strategy.SchedulingExecutionVertex;
+import org.apache.flink.runtime.scheduler.strategy.SchedulingResultPartition;
+import org.apache.flink.runtime.scheduler.strategy.SchedulingTopology;
 import org.apache.flink.runtime.shuffle.NettyShuffleMaster;
 import org.apache.flink.runtime.shuffle.ShuffleMaster;
 import org.apache.flink.runtime.state.SharedStateRegistry;
@@ -250,6 +259,12 @@ public class ExecutionGraph implements AccessExecutionGraph {
 	/** The total number of vertices currently in the execution graph. */
 	private int numVerticesTotal;
 
+	private final PartitionReleaseStrategy.Factory partitionReleaseStrategyFactory;
+
+	private PartitionReleaseStrategy partitionReleaseStrategy;
+
+	private SchedulingTopology schedulingTopology;
+
 	// ------ Configuration of the Execution -------
 
 	/** Flag to indicate whether the scheduler may queue tasks for execution, or needs to be able
@@ -413,6 +428,7 @@ public class ExecutionGraph implements AccessExecutionGraph {
 			userClassLoader,
 			blobWriter,
 			allocationTimeout,
+			new NotReleasingPartitionReleaseStrategy.Factory(),
 			NettyShuffleMaster.INSTANCE,
 			true,
 			new PartitionTrackerImpl(
@@ -433,6 +449,7 @@ public class ExecutionGraph implements AccessExecutionGraph {
 			ClassLoader userClassLoader,
 			BlobWriter blobWriter,
 			Time allocationTimeout,
+			PartitionReleaseStrategy.Factory partitionReleaseStrategyFactory,
 			ShuffleMaster<?> shuffleMaster,
 			boolean forcePartitionReleaseOnConsumption,
 			PartitionTracker partitionTracker) throws IOException {
@@ -463,6 +480,8 @@ public class ExecutionGraph implements AccessExecutionGraph {
 
 		this.rpcTimeout = checkNotNull(rpcTimeout);
 		this.allocationTimeout = checkNotNull(allocationTimeout);
+
+		this.partitionReleaseStrategyFactory = checkNotNull(partitionReleaseStrategyFactory);
 
 		this.restartStrategy = restartStrategy;
 		this.kvStateLocationRegistry = new KvStateLocationRegistry(jobInformation.getJobId(), getAllVertices());
@@ -913,6 +932,11 @@ public class ExecutionGraph implements AccessExecutionGraph {
 		}
 
 		failoverStrategy.notifyNewVertices(newExecJobVertices);
+
+		schedulingTopology = new ExecutionGraphToSchedulingTopologyAdapter(this);
+		partitionReleaseStrategy = partitionReleaseStrategyFactory.createInstance(
+			schedulingTopology,
+			new DefaultFailoverTopology(this));
 	}
 
 	public void scheduleForExecution() throws JobException {
@@ -1605,36 +1629,9 @@ public class ExecutionGraph implements AccessExecutionGraph {
 
 		if (attempt != null) {
 			try {
-				Map<String, Accumulator<?, ?>> accumulators;
-
-				switch (state.getExecutionState()) {
-					case RUNNING:
-						return attempt.switchToRunning();
-
-					case FINISHED:
-						// this deserialization is exception-free
-						accumulators = deserializeAccumulators(state);
-						attempt.markFinished(accumulators, state.getIOMetrics());
-						return true;
-
-					case CANCELED:
-						// this deserialization is exception-free
-						accumulators = deserializeAccumulators(state);
-						attempt.completeCancelling(accumulators, state.getIOMetrics());
-						return true;
-
-					case FAILED:
-						// this deserialization is exception-free
-						accumulators = deserializeAccumulators(state);
-						attempt.markFailed(state.getError(userClassLoader), accumulators, state.getIOMetrics());
-						return true;
-
-					default:
-						// we mark as failed and return false, which triggers the TaskManager
-						// to remove the task
-						attempt.fail(new Exception("TaskManager sent illegal state update: " + state.getExecutionState()));
-						return false;
-				}
+				final boolean stateUpdated = updateStateInternal(state, attempt);
+				maybeReleasePartitions(attempt);
+				return stateUpdated;
 			}
 			catch (Throwable t) {
 				ExceptionUtils.rethrowIfFatalErrorOrOOM(t);
@@ -1647,6 +1644,77 @@ public class ExecutionGraph implements AccessExecutionGraph {
 		else {
 			return false;
 		}
+	}
+
+	private boolean updateStateInternal(final TaskExecutionState state, final Execution attempt) {
+		Map<String, Accumulator<?, ?>> accumulators;
+
+		switch (state.getExecutionState()) {
+			case RUNNING:
+				return attempt.switchToRunning();
+
+			case FINISHED:
+				// this deserialization is exception-free
+				accumulators = deserializeAccumulators(state);
+				attempt.markFinished(accumulators, state.getIOMetrics());
+				return true;
+
+			case CANCELED:
+				// this deserialization is exception-free
+				accumulators = deserializeAccumulators(state);
+				attempt.completeCancelling(accumulators, state.getIOMetrics());
+				return true;
+
+			case FAILED:
+				// this deserialization is exception-free
+				accumulators = deserializeAccumulators(state);
+				attempt.markFailed(state.getError(userClassLoader), accumulators, state.getIOMetrics());
+				return true;
+
+			default:
+				// we mark as failed and return false, which triggers the TaskManager
+				// to remove the task
+				attempt.fail(new Exception("TaskManager sent illegal state update: " + state.getExecutionState()));
+				return false;
+		}
+	}
+
+	private void maybeReleasePartitions(final Execution attempt) {
+		final ExecutionVertexID finishedExecutionVertex = attempt.getVertex().getID();
+
+		if (attempt.getState() == ExecutionState.FINISHED) {
+			final List<IntermediateResultPartitionID> releasablePartitions = partitionReleaseStrategy.vertexFinished(finishedExecutionVertex);
+			releasePartitions(releasablePartitions);
+		} else {
+			partitionReleaseStrategy.vertexUnfinished(finishedExecutionVertex);
+		}
+	}
+
+	private void releasePartitions(final List<IntermediateResultPartitionID> releasablePartitions) {
+		if (releasablePartitions.size() > 0) {
+			final List<ResultPartitionID> partitionIds = releasablePartitions.stream()
+				.map(this::createResultPartitionId)
+				.collect(Collectors.toList());
+
+			partitionTracker.stopTrackingAndReleasePartitions(partitionIds);
+		}
+	}
+
+	private ResultPartitionID createResultPartitionId(final IntermediateResultPartitionID resultPartitionId) {
+		final SchedulingResultPartition schedulingResultPartition = schedulingTopology.getResultPartitionOrThrow(resultPartitionId);
+		final SchedulingExecutionVertex producer = schedulingResultPartition.getProducer();
+		final ExecutionVertexID producerId = producer.getId();
+		final JobVertexID jobVertexId = producerId.getJobVertexId();
+		final ExecutionJobVertex jobVertex = getJobVertex(jobVertexId);
+		checkNotNull(jobVertex, "Unknown job vertex %s", jobVertexId);
+
+		final ExecutionVertex[] taskVertices = jobVertex.getTaskVertices();
+		final int subtaskIndex = producerId.getSubtaskIndex();
+		checkState(subtaskIndex < taskVertices.length, "Invalid subtask index %d for job vertex %s", subtaskIndex, jobVertexId);
+
+		final ExecutionVertex taskVertex = taskVertices[subtaskIndex];
+		final Execution execution = taskVertex.getCurrentExecutionAttempt();
+		return new ResultPartitionID(resultPartitionId, execution.getAttemptId());
 	}
 
 	/**
@@ -1834,5 +1902,9 @@ public class ExecutionGraph implements AccessExecutionGraph {
 
 	public PartitionTracker getPartitionTracker() {
 		return partitionTracker;
+	}
+
+	PartitionReleaseStrategy getPartitionReleaseStrategy() {
+		return partitionReleaseStrategy;
 	}
 }
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/ExecutionVertex.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/ExecutionVertex.java
@@ -601,6 +601,7 @@ public class ExecutionVertex implements AccessExecutionVertex, Archiveable<Archi
 			if (oldState.isTerminal()) {
 				if (oldState == FINISHED) {
 					oldExecution.stopTrackingAndReleasePartitions();
+					getExecutionGraph().getPartitionReleaseStrategy().vertexUnfinished(executionVertexId);
 				}
 
 				priorExecutions.add(oldExecution.archive());
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/ExecutionVertex.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/ExecutionVertex.java
@@ -39,6 +39,7 @@ import org.apache.flink.runtime.jobmanager.scheduler.CoLocationGroup;
 import org.apache.flink.runtime.jobmanager.scheduler.LocationPreferenceConstraint;
 import org.apache.flink.runtime.jobmaster.LogicalSlot;
 import org.apache.flink.runtime.jobmaster.slotpool.SlotProvider;
+import org.apache.flink.runtime.scheduler.strategy.ExecutionVertexID;
 import org.apache.flink.runtime.taskmanager.TaskManagerLocation;
 import org.apache.flink.runtime.util.EvictingBoundedList;
 import org.apache.flink.util.ExceptionUtils;
@@ -82,6 +83,8 @@ public class ExecutionVertex implements AccessExecutionVertex, Archiveable<Archi
 	private final ExecutionEdge[][] inputEdges;
 
 	private final int subTaskIndex;
+
+	private final ExecutionVertexID executionVertexId;
 
 	private final EvictingBoundedList<ArchivedExecution> priorExecutions;
 
@@ -142,6 +145,7 @@ public class ExecutionVertex implements AccessExecutionVertex, Archiveable<Archi
 
 		this.jobVertex = jobVertex;
 		this.subTaskIndex = subTaskIndex;
+		this.executionVertexId = new ExecutionVertexID(jobVertex.getJobVertexId(), subTaskIndex);
 		this.taskNameWithSubtask = String.format("%s (%d/%d)",
 				jobVertex.getJobVertex().getName(), subTaskIndex + 1, jobVertex.getParallelism());
 
@@ -226,6 +230,10 @@ public class ExecutionVertex implements AccessExecutionVertex, Archiveable<Archi
 	@Override
 	public int getParallelSubtaskIndex() {
 		return this.subTaskIndex;
+	}
+
+	public ExecutionVertexID getID() {
+		return executionVertexId;
 	}
 
 	public int getNumberOfInputs() {
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/PipelinedRegionComputeUtil.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/PipelinedRegionComputeUtil.java
@@ -19,6 +19,9 @@
 
 package org.apache.flink.runtime.executiongraph.failover.flip1;
 
+import org.apache.flink.runtime.executiongraph.failover.flip1.partitionrelease.PipelinedRegion;
+import org.apache.flink.runtime.scheduler.strategy.ExecutionVertexID;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -27,13 +30,28 @@ import java.util.HashSet;
 import java.util.IdentityHashMap;
 import java.util.Map;
 import java.util.Set;
+import java.util.function.Function;
+import java.util.stream.Collectors;
 
 /**
- * Utility for computing pipeliend regions.
+ * Utility for computing pipelined regions.
  */
 public final class PipelinedRegionComputeUtil {
 
 	private static final Logger LOG = LoggerFactory.getLogger(PipelinedRegionComputeUtil.class);
+
+	public static Set<PipelinedRegion> toPipelinedRegionsSet(final Set<Set<FailoverVertex>> distinctRegions) {
+		return distinctRegions.stream()
+			.map(toExecutionVertexIdSet())
+			.map(PipelinedRegion::from)
+			.collect(Collectors.toSet());
+	}
+
+	private static Function<Set<FailoverVertex>, Set<ExecutionVertexID>> toExecutionVertexIdSet() {
+		return failoverVertices -> failoverVertices.stream()
+			.map(FailoverVertex::getExecutionVertexID)
+			.collect(Collectors.toSet());
+	}
 
 	public static Set<Set<FailoverVertex>> computePipelinedRegions(final FailoverTopology topology) {
 		// currently we let a job with co-location constraints fail as one region
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/PipelinedRegionComputeUtil.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/PipelinedRegionComputeUtil.java
@@ -22,6 +22,7 @@ package org.apache.flink.runtime.executiongraph.failover.flip1;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.IdentityHashMap;
 import java.util.Map;
@@ -41,8 +42,6 @@ public final class PipelinedRegionComputeUtil {
 			return uniqueRegions(buildOneRegionForAllVertices(topology));
 		}
 
-		// we use the map (list -> null) to imitate an IdentityHashSet (which does not exist)
-		// this helps to optimize the building performance as it uses reference equality
 		final Map<FailoverVertex, Set<FailoverVertex>> vertexToRegion = new IdentityHashMap<>();
 
 		// iterate all the vertices which are topologically sorted
@@ -105,11 +104,9 @@ public final class PipelinedRegionComputeUtil {
 
 	private static Set<Set<FailoverVertex>> uniqueRegions(final Map<FailoverVertex, Set<FailoverVertex>> vertexToRegion) {
 		// find out all the distinct regions
-		final IdentityHashMap<Set<FailoverVertex>, Object> distinctRegions = new IdentityHashMap<>();
-		for (Set<FailoverVertex> regionVertices : vertexToRegion.values()) {
-			distinctRegions.put(regionVertices, null);
-		}
-		return distinctRegions.keySet();
+		final Set<Set<FailoverVertex>> distinctRegions = Collections.newSetFromMap(new IdentityHashMap<>());
+		distinctRegions.addAll(vertexToRegion.values());
+		return distinctRegions;
 	}
 
 	private PipelinedRegionComputeUtil() {
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/PipelinedRegionComputeUtil.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/PipelinedRegionComputeUtil.java
@@ -103,7 +103,6 @@ public final class PipelinedRegionComputeUtil {
 	}
 
 	private static Set<Set<FailoverVertex>> uniqueRegions(final Map<FailoverVertex, Set<FailoverVertex>> vertexToRegion) {
-		// find out all the distinct regions
 		final Set<Set<FailoverVertex>> distinctRegions = Collections.newSetFromMap(new IdentityHashMap<>());
 		distinctRegions.addAll(vertexToRegion.values());
 		return distinctRegions;
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/PipelinedRegionComputeUtil.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/PipelinedRegionComputeUtil.java
@@ -1,0 +1,117 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.flink.runtime.executiongraph.failover.flip1;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.HashSet;
+import java.util.IdentityHashMap;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * Utility for computing pipeliend regions.
+ */
+public final class PipelinedRegionComputeUtil {
+
+	private static final Logger LOG = LoggerFactory.getLogger(PipelinedRegionComputeUtil.class);
+
+	public static Set<Set<FailoverVertex>> computePipelinedRegions(final FailoverTopology topology) {
+		// currently we let a job with co-location constraints fail as one region
+		// putting co-located vertices in the same region with each other can be a future improvement
+		if (topology.containsCoLocationConstraints()) {
+			return uniqueRegions(buildOneRegionForAllVertices(topology));
+		}
+
+		// we use the map (list -> null) to imitate an IdentityHashSet (which does not exist)
+		// this helps to optimize the building performance as it uses reference equality
+		final Map<FailoverVertex, Set<FailoverVertex>> vertexToRegion = new IdentityHashMap<>();
+
+		// iterate all the vertices which are topologically sorted
+		for (FailoverVertex vertex : topology.getFailoverVertices()) {
+			Set<FailoverVertex> currentRegion = new HashSet<>(1);
+			currentRegion.add(vertex);
+			vertexToRegion.put(vertex, currentRegion);
+
+			for (FailoverEdge inputEdge : vertex.getInputEdges()) {
+				if (inputEdge.getResultPartitionType().isPipelined()) {
+					final FailoverVertex producerVertex = inputEdge.getSourceVertex();
+					final Set<FailoverVertex> producerRegion = vertexToRegion.get(producerVertex);
+
+					if (producerRegion == null) {
+						throw new IllegalStateException("Producer task " + producerVertex.getExecutionVertexName()
+							+ " failover region is null while calculating failover region for the consumer task "
+							+ vertex.getExecutionVertexName() + ". This should be a failover region building bug.");
+					}
+
+					// check if it is the same as the producer region, if so skip the merge
+					// this check can significantly reduce compute complexity in All-to-All PIPELINED edge case
+					if (currentRegion != producerRegion) {
+						// merge current region and producer region
+						// merge the smaller region into the larger one to reduce the cost
+						final Set<FailoverVertex> smallerSet;
+						final Set<FailoverVertex> largerSet;
+						if (currentRegion.size() < producerRegion.size()) {
+							smallerSet = currentRegion;
+							largerSet = producerRegion;
+						} else {
+							smallerSet = producerRegion;
+							largerSet = currentRegion;
+						}
+						for (FailoverVertex v : smallerSet) {
+							vertexToRegion.put(v, largerSet);
+						}
+						largerSet.addAll(smallerSet);
+						currentRegion = largerSet;
+					}
+				}
+			}
+		}
+
+		return uniqueRegions(vertexToRegion);
+	}
+
+	private static Map<FailoverVertex, Set<FailoverVertex>> buildOneRegionForAllVertices(final FailoverTopology topology) {
+		LOG.warn("Cannot decompose the topology into individual failover regions due to use of " +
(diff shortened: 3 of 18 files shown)


flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/ExecutionGraph.java
```diff
@@ -1634,6 +1641,28 @@ public boolean updateState(TaskExecutionState state) {
...
+		}
+	}
+
+	private void releasePartitions(final List<? extends IntermediateResultPartitionID> releasablePartitions) {
+		for (final IntermediateResultPartitionID resultPartition : releasablePartitions) {
+			releasePartitions(resultPartition);
+		}
+	}
+
+	private void releasePartitions(final IntermediateResultPartitionID resultPartition) {
```
zentol: JM/TM track/release partitions by their `ResultPartitionID`. You likely will have to use that here as well.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/RegionPartitionReleaseStrategyTest.java
```diff
@@ -0,0 +1,147 @@
...
+		final ExecutionVertexID onlyConsumerVertexId = consumers.get(0).getId();
+		final IntermediateResultPartitionID onlyResultPartitionId = resultPartitions.get(0).getId();
+
+		final Set<PipelinedRegion> pipelinedRegions = pipelinedRegionsSet(
+			PipelinedRegion.from(onlyProducerVertexId),
+			PipelinedRegion.from(onlyConsumerVertexId));
+		final RegionPartitionReleaseStrategy regionPartitionReleaseStrategy = new RegionPartitionReleaseStrategy(testingSchedulingTopology, pipelinedRegions);
+
+		regionPartitionReleaseStrategy.vertexFinished(onlyConsumerVertexId);
+		regionPartitionReleaseStrategy.vertexUnfinished(onlyConsumerVertexId);
```
zentol: This test would pass even if vertexUnfinished is a no-op.

Also, shouldn't we at some point mark the producer vertex as finished? A scenario where a consumer is finished without the producer being finished sounds a bit like undefined behavior.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/partitionrelease/PipelinedRegionExecutionView.java
```diff
@@ -0,0 +1,64 @@
...
+
+/**
+ * Provides a virtual execution state of a {@link PipelinedRegion}.
+ *
+ * <p>A pipelined region can be either finished or unfinished. It is finished iff. all its
+ * execution have reached the finished state.
+ */
+class PipelinedRegionExecutionView {
+
+	private final PipelinedRegion pipelinedRegion;
```
zentol: Do we only have this field to ensure we don't unfinish add arbitrary vertices? You could have the same behavior by maintaining a Map<EVID, Boolean>. You would then only pass in a set of vertices through the constructor.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/RegionPartitionReleaseStrategyTest.java
```diff
@@ -0,0 +1,147 @@
...
+
+import static org.hamcrest.Matchers.contains;
+import static org.hamcrest.Matchers.empty;
+import static org.hamcrest.Matchers.is;
+import static org.junit.Assert.assertThat;
+
+/**
+ * Tests for {@link RegionPartitionReleaseStrategy}.
+ */
+public class RegionPartitionReleaseStrategyTest {
```
GJL: `extends TestLogger`

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/failover/flip1/partitionrelease/PipelinedRegionExecutionViewTest.java
```diff
@@ -0,0 +1,74 @@
...
+
+import org.junit.Test;
+
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+/**
+ * Test for {@link PipelinedRegionExecutionView}.
+ */
+public class PipelinedRegionExecutionViewTest {
```
GJL: `extends TestLogger`

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/ExecutionGraph.java
zentol: or let me ask it differently, how do you intend to map the `IntermediateResultPartitionID` to the `ResultPartitionID`?

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/partitionrelease/RegionPartitionReleaseStrategy.java
```diff
@@ -0,0 +1,170 @@
...
+			}
+		}
+	}
+
+	private PipelinedRegionConsumedBlockingPartitions computeConsumedPartitionsOfVertexRegion(final PipelinedRegion pipelinedRegion) {
+		final Set<IntermediateResultPartitionID> resultPartitionsOutsideOfRegion = findResultPartitionsOutsideOfRegion(pipelinedRegion);
+		return new PipelinedRegionConsumedBlockingPartitions(pipelinedRegion, resultPartitionsOutsideOfRegion);
+	}
+
+	private Set<IntermediateResultPartitionID> findResultPartitionsOutsideOfRegion(final PipelinedRegion pipelinedRegion) {
```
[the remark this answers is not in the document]
GJL: This method calls `.map(schedulingTopology::getVertexOrThrow)` so it cannot be static.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/partitionrelease/PartitionReleaseStrategy.java
```diff
@@ -0,0 +1,49 @@
...
+	 * Calling this method informs the strategy that a vertex finished.
+	 *
+	 * @param finishedVertex Id of the vertex that finished the execution
+	 * @return A list of result partition that can be released
+	 */
+	List<IntermediateResultPartitionID> vertexFinished(ExecutionVertexID finishedVertex);
+
+	/**
+	 * Calling this method informs the strategy that a vertex is no longer in finished state, e.g.,
+	 * when an execution is restarted.
```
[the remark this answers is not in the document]
GJL: I can change it to _"[...]e.g., when a vertex is re-executed"_

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/partitionrelease/RegionPartitionReleaseStrategy.java
```diff
@@ -0,0 +1,170 @@
...
+	private void initRegionExecutionViewByVertex(final Set<PipelinedRegion> pipelinedRegions) {
+		for (PipelinedRegion pipelinedRegion : pipelinedRegions) {
+			final PipelinedRegionExecutionView regionExecutionView = new PipelinedRegionExecutionView(pipelinedRegion);
+			for (ExecutionVertexID executionVertexId : pipelinedRegion) {
+				regionExecutionViewByVertex.put(executionVertexId, regionExecutionView);
+			}
+		}
+	}
+
+	private PipelinedRegionConsumedBlockingPartitions computeConsumedPartitionsOfVertexRegion(final PipelinedRegion pipelinedRegion) {
```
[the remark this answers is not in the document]
GJL: I don't think I can make `findResultPartitionsOutsideOfRegion` static

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/RegionPartitionReleaseStrategyTest.java
GJL: Good catch.

> Also, shouldn't we at some point mark the producer vertex as finished? A scenario where a consumer is finished without the producer being finished sounds a bit like undefined behavior.

We can but:
-  calls to `ExecutionGraph#updateState(TaskExecutionState state)` can overtake each other. Therefore it is to be expected that a consumer is finished (in our bookkeeping) before the producer.
-  I don't think it adds to the readability of the test

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/ExecutionGraph.java
```diff
@@ -1790,6 +1819,8 @@ void notifyExecutionChange(
 			// by filtering out late failure calls, we can save some work in
 			// avoiding redundant local failover
 			if (execution.getGlobalModVersion() == globalModVersion) {
+				partitionReleaseStrategy.vertexUnfinished(execution.getVertex().getID());
```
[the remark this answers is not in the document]
GJL: I think you are right.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/partitionrelease/PipelinedRegionExecutionView.java
GJL: There is a getter for the `pipelinedRegion` which is actually used. Also, I don't see anything wrong with object composition here. Using `Map` would require additional processing of the input.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/ExecutionGraph.java
GJL: We will have to look up the `ExecutionVertex` of the `IntermediateResultPartitionID`, which could be done via the `SchedulingTopology`.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/PipelinedRegionComputeUtil.java
```diff
@@ -0,0 +1,117 @@
...
+		for (FailoverVertex vertex : topology.getFailoverVertices()) {
+			allVertices.add(vertex);
+			vertexToRegion.put(vertex, allVertices);
+		}
+		return vertexToRegion;
+	}
+
+	private static Set<Set<FailoverVertex>> uniqueRegions(final Map<FailoverVertex, Set<FailoverVertex>> vertexToRegion) {
+		// find out all the distinct regions
+		final IdentityHashMap<Set<FailoverVertex>, Object> distinctRegions = new IdentityHashMap<>();
```
GJL: Consider `Collections.newSetFromMap(new IdentityHashMap<>())`

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/partitionrelease/PipelinedRegionExecutionView.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.Set;
+
+import static org.apache.flink.util.Preconditions.checkArgument;
+import static org.apache.flink.util.Preconditions.checkNotNull;
+
+/**
+ * Provides a virtual execution state of a {@link PipelinedRegion}.
+ *
+ * <p>A pipelined region can be either finished or unfinished. It is finished iff. all its
+ * execution have reached the finished state.
```
GJL: _executions_

GJL:
Rebased to master

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/partitionrelease/RegionPartitionReleaseStrategy.java
```diff
@@ -0,0 +1,170 @@
...
+
+		checkNotNull(pipelinedRegions);
+		initConsumedBlockingPartitionsByRegion(pipelinedRegions);
+		initRegionExecutionViewByVertex(pipelinedRegions);
+	}
+
+	private void initConsumedBlockingPartitionsByRegion(final Set<PipelinedRegion> pipelinedRegions) {
+		for (PipelinedRegion pipelinedRegion : pipelinedRegions) {
+			final PipelinedRegionConsumedBlockingPartitions consumedPartitions = computeConsumedPartitionsOfVertexRegion(pipelinedRegion);
+			consumedBlockingPartitionsByRegion.put(pipelinedRegion, consumedPartitions);
```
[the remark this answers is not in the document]
GJL: `computeConsumedPartitionsOfVertexRegion` is not static. I would have to pass `schedulingTopology` to make it static.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/RestartPipelinedRegionStrategy.java
```diff
@@ -94,66 +94,10 @@ public RestartPipelinedRegionStrategy(
...
-				}
-			}
-		}
-
-		// find out all the distinct regions
-		final IdentityHashMap<HashSet<FailoverVertex>, Object> distinctRegions = new IdentityHashMap<>();
-		for (HashSet<FailoverVertex> regionVertices : vertexToRegion.values()) {
-			distinctRegions.put(regionVertices, null);
-		}
+		final Set<Set<FailoverVertex>> distinctRegions = PipelinedRegionComputeUtil.computePipelinedRegions(topology);
```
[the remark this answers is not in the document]
GJL: That's not so obvious for me. `RestartPipelinedRegionStrategy` still needs `FailoverVertex` at the end of the day. Mapping from `PipelinedRegion` to `Set<FailoverVertex>` is tedious. 

To me It's also not clear whether it is the right abstraction that `PipelinedRegionComputeUtil#computePipelinedRegions()` receives `FailoverTopology` as input. 

Lastly, I think that this comment competes with [that one](https://github.com/apache/flink/pull/8804/files#r295769220).

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/ExecutionGraph.java
```diff
@@ -1586,6 +1610,8 @@ public boolean updateState(TaskExecutionState state) {
 
 		if (attempt != null) {
 			try {
+				maybeReleasePartitions(attempt, state);
```
zentol: this must be done after the attempt is marked as finished. If the finishing fails we might still need these partitions.

GJL:
@zentol If a blocking partition is finished and its producer being reset, are there cases where we do not want to release that partition? I wonder if the release strategy should be responsible to release partitions in cases of errors.

zentol:
> If a blocking partition is finished and its producer being reset, are there cases where we do not want to release that partition?

I don't think so.

> I wonder if the release strategy should be responsible to release partitions in cases of errors.

I'm not sure either. We'd have to handle it in the ExecutionVertex otherwise, which would be unfortunate as it adds yet another place from where release calls can originate.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/partitionrelease/PartitionReleaseStrategy.java
```diff
@@ -0,0 +1,57 @@
...
+	 *
+	 * @param finishedVertex Id of the vertex that finished the execution
+	 * @return A list of result partition that can be released
+	 */
+	List<IntermediateResultPartitionID> vertexFinished(ExecutionVertexID finishedVertex);
+
+	/**
+	 * Calling this method informs the strategy that a vertex is no longer in finished state, e.g.,
+	 * when a vertex is re-executed.
+	 * @param executionVertexID Id of the vertex that is no longer in finished state.
```
zhijiangW: add an empty line before @param to keep consistency with above `vertexFinished `

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/partitionrelease/PartitionReleaseStrategy.java
```diff
@@ -0,0 +1,57 @@
...
+ * Interface for strategies that decide when to release
+ * {@link IntermediateResultPartition IntermediateResultPartitions}.
+ */
+public interface PartitionReleaseStrategy {
+
+	/**
+	 * Calling this method informs the strategy that a vertex finished.
+	 *
+	 * @param finishedVertex Id of the vertex that finished the execution
+	 * @return A list of result partition that can be released
```
zhijiangW: partition->partitions

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/partitionrelease/PipelinedRegionExecutionView.java
```diff
@@ -0,0 +1,64 @@
...
+	public boolean isFinished() {
+		return unfinishedVertices.isEmpty();
+	}
+
+	public void finish(final ExecutionVertexID executionVertexId) {
+		checkArgument(pipelinedRegion.contains(executionVertexId));
+		unfinishedVertices.remove(executionVertexId);
+	}
+
+	public void unfinish(final ExecutionVertexID executionVertexId) {
```
zhijiangW: could be package private

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/partitionrelease/PipelinedRegionExecutionView.java
```diff
@@ -0,0 +1,64 @@
...
+		checkArgument(pipelinedRegion.contains(executionVertexId));
+		unfinishedVertices.add(executionVertexId);
+	}
+
+	public PipelinedRegion getPipelinedRegion() {
```
zhijiangW: ditto: package private

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/partitionrelease/PipelinedRegionExecutionView.java
```diff
@@ -0,0 +1,64 @@
...
+import static org.apache.flink.util.Preconditions.checkArgument;
+import static org.apache.flink.util.Preconditions.checkNotNull;
+
+/**
+ * Provides a virtual execution state of a {@link PipelinedRegion}.
+ *
+ * <p>A pipelined region can be either finished or unfinished. It is finished iff. all its
+ * executions have reached the finished state.
+ */
+class PipelinedRegionExecutionView {
```
zhijiangW: PipelinedRegionExecutionView -> PipelinedRegionStateView?

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/partitionrelease/PipelinedRegionExecutionView.java
```diff
@@ -0,0 +1,64 @@
...
+	PipelinedRegionExecutionView(final PipelinedRegion pipelinedRegion) {
+		this.pipelinedRegion = checkNotNull(pipelinedRegion);
+		this.unfinishedVertices = new HashSet<>(pipelinedRegion.getExecutionVertexIds());
+	}
+
+	public boolean isFinished() {
+		return unfinishedVertices.isEmpty();
+	}
+
+	public void finish(final ExecutionVertexID executionVertexId) {
```
zhijiangW: better to call `vertexFinished` as `PartitionReleaseStrategy#vertexFinished`, also for the following `vertexUnfinished`

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/partitionrelease/PipelinedRegion.java
```diff
@@ -0,0 +1,67 @@
...
+
+	public static PipelinedRegion from(final Set<ExecutionVertexID> executionVertexIds) {
+		return new PipelinedRegion(executionVertexIds);
+	}
+
+	public static PipelinedRegion from(final ExecutionVertexID... executionVertexIds) {
+		return new PipelinedRegion(new HashSet<>(Arrays.asList(executionVertexIds)));
+	}
+
+	public Set<ExecutionVertexID> getExecutionVertexIds() {
```
zhijiangW: could be package private

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/partitionrelease/PipelinedRegion.java
```diff
@@ -0,0 +1,67 @@
...
+public class PipelinedRegion implements Iterable<ExecutionVertexID> {
+
+	private final Set<ExecutionVertexID> executionVertexIds;
+
+	private PipelinedRegion(final Set<ExecutionVertexID> executionVertexIds) {
+		this.executionVertexIds = new HashSet<>(executionVertexIds);
+	}
+
+	public static PipelinedRegion from(final Set<ExecutionVertexID> executionVertexIds) {
+		return new PipelinedRegion(executionVertexIds);
```
zhijiangW: checkNotNull(executionVertexIds)

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/partitionrelease/PipelinedRegion.java
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	public static PipelinedRegion from(final ExecutionVertexID... executionVertexIds) {
+		return new PipelinedRegion(new HashSet<>(Arrays.asList(executionVertexIds)));
```
zhijiangW: ditto: checkNotNull(executionVertexIds)

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/partitionrelease/PipelinedRegionConsumedBlockingPartitions.java
```diff
@@ -0,0 +1,51 @@
...
+	private final Set<IntermediateResultPartitionID> consumedBlockingPartitions;
+
+	PipelinedRegionConsumedBlockingPartitions(
+			final PipelinedRegion pipelinedRegion,
+			final Set<IntermediateResultPartitionID> consumedBlockingPartitions) {
+		this.pipelinedRegion = checkNotNull(pipelinedRegion);
+		this.consumedBlockingPartitions = checkNotNull(consumedBlockingPartitions);
+	}
+
+	public Set<IntermediateResultPartitionID> getConsumedBlockingPartitions() {
```
zhijiangW: package private method

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/partitionrelease/PipelinedRegionConsumedBlockingPartitions.java
```diff
@@ -0,0 +1,51 @@
...
+		return consumedBlockingPartitions;
+	}
+
+	public PipelinedRegion getPipelinedRegion() {
```
zhijiangW: ditto: package private

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/partitionrelease/RegionPartitionReleaseStrategy.java
```diff
@@ -0,0 +1,190 @@
...
+
+	public RegionPartitionReleaseStrategy(
+			final SchedulingTopology schedulingTopology,
+			final Set<PipelinedRegion> pipelinedRegions) {
+
+		this.schedulingTopology = checkNotNull(schedulingTopology);
+
+		checkNotNull(pipelinedRegions);
+		initConsumedBlockingPartitionsByRegion(pipelinedRegions);
+		initRegionExecutionViewByVertex(pipelinedRegions);
```
zhijiangW: Could loop `pipelinedRegions` only once for both inits.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/partitionrelease/RegionPartitionReleaseStrategy.java
```diff
@@ -0,0 +1,190 @@
...
+ * Releases blocking intermediate result partitions that are incident to a {@link PipelinedRegion},
+ * as soon as the region's execution vertices are finished.
+ */
+public class RegionPartitionReleaseStrategy implements PartitionReleaseStrategy {
+
+	private final SchedulingTopology schedulingTopology;
+
+	private final Map<PipelinedRegion, PipelinedRegionConsumedBlockingPartitions> consumedBlockingPartitionsByRegion = new IdentityHashMap<>();
+
+	private final Map<ExecutionVertexID, PipelinedRegionExecutionView> regionExecutionViewByVertex = new HashMap<>();
```
zhijiangW: might give an initial size for above maps

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/ExecutionGraph.java
```diff
@@ -1630,6 +1627,79 @@ public boolean updateState(TaskExecutionState state) {
...
+
+		if (state.getExecutionState() == ExecutionState.FINISHED) {
+			final List<IntermediateResultPartitionID> releasablePartitions = partitionReleaseStrategy.vertexFinished(finishedExecutionVertex);
+			releasePartitions(releasablePartitions);
+		} else {
+			partitionReleaseStrategy.vertexUnfinished(finishedExecutionVertex);
+		}
+	}
+
+	private void releasePartitions(final List<? extends IntermediateResultPartitionID> releasablePartitions) {
```
zhijiangW: why need `? extends IntermediateResultPartitionID` here?

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/ExecutionGraph.java
```diff
@@ -1630,6 +1627,79 @@ public boolean updateState(TaskExecutionState state) {
...
+				return false;
+		}
+	}
+
+	private void maybeReleasePartitions(final TaskExecutionState state, final Execution attempt) {
+		final ExecutionVertexID finishedExecutionVertex = attempt.getVertex().getID();
+
+		if (state.getExecutionState() == ExecutionState.FINISHED) {
+			final List<IntermediateResultPartitionID> releasablePartitions = partitionReleaseStrategy.vertexFinished(finishedExecutionVertex);
+			releasePartitions(releasablePartitions);
```
zhijiangW: It might be easy for constructing `ResultPartitionID` here, because we already have `attempt` info in parameters, then we could avoid the complex logic in following `createResultPartitionId`.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/ExecutionVertex.java
```diff
@@ -353,6 +354,10 @@ public InputDependencyConstraint getInputDependencyConstraint() {
 		return getJobVertex().getInputDependencyConstraint();
 	}
 
+	public ExecutionVertexID getID() {
```
zhijiangW: It seems more proper to construct the `ExecutionVertexID` in the constructor only once, then all the other usages could get it directly. And `DefaultFailoverTopology` does not need to create `ExecutionVertexID` as well.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/ExecutionGraph.java
```diff
@@ -247,6 +256,12 @@
 	/** The total number of vertices currently in the execution graph. */
 	private int numVerticesTotal;
 
+	private final PartitionReleaseStrategy.Factory partitionReleaseStrategyFactory;
+
+	private PartitionReleaseStrategy partitionReleaseStrategy;
+
+	private SchedulingTopology schedulingTopology;
```
zhijiangW: How about initializing `schedulingTopology` and `partitionReleaseStrategy` in the constructor, then we could make them final, and no need to maintain class-level `partitionReleaseStrategyFactory`.

GJL: I think this is not possible because the `JobGraph` is attached to the `ExecutionGraph` in an extra step. `ExecutionGraphToSchedulingTopologyAdapter` adapts the `ExecutionGraph` eagerly.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/ExecutionGraph.java
GJL: `attempt` here is the finishing vertex. However, we need the producer of the partition that is incident to the pipelined region.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/ExecutionGraph.java
GJL: https://stackoverflow.com/questions/2723397/what-is-pecs-producer-extends-consumer-super

It's not needed here. I will remove it.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/ExecutionGraph.java
zhijiangW: Yes, I have not looked through the details in `ExecutionGraphToSchedulingTopologyAdapter` before, only saw it relies `ExecutionGraph`, actually it relies some infos after attaching `JobGraph`.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/partitionrelease/RegionPartitionReleaseStrategy.java
GJL: Size needs to be `numberOfExecutionVertexIds / loadFactor` to avoid rehashing. I don't think it's worth it here.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/ExecutionGraph.java
zhijiangW: Yes, I missed it. The current attempt is consumer, and we need the correspond producer attempt for the partition.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/partitionrelease/PartitionReleaseStrategy.java
```diff
@@ -0,0 +1,57 @@
...
+import java.util.List;
+
+/**
+ * Interface for strategies that decide when to release
+ * {@link IntermediateResultPartition IntermediateResultPartitions}.
+ */
+public interface PartitionReleaseStrategy {
+
+	/**
+	 * Calling this method informs the strategy that a vertex finished.
```
gaoyunhaii: I think `that` -> `when` may be better.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/partitionrelease/PartitionReleaseStrategy.java
```diff
@@ -0,0 +1,57 @@
...
+	 *
+	 * @param finishedVertex Id of the vertex that finished the execution
+	 * @return A list of result partition that can be released
+	 */
+	List<IntermediateResultPartitionID> vertexFinished(ExecutionVertexID finishedVertex);
+
+	/**
+	 * Calling this method informs the strategy that a vertex is no longer in finished state, e.g.,
```
gaoyunhaii: Similarly, I think `that` -> `when` may be better.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/partitionrelease/PipelinedRegion.java
GJL: I think here it is ok to omit `checkNotNull`. 

- `PipelinedRegion.from(null)` is an ambiguous  call, and does not compile currently
- `PipelinedRegion.from(null, null)` cannot be caught by `checkNotNull`.
- `Arrays.asList()` already eagerly checks for a `null` argument. From the stacktrace it should be obvious where the NPE originates from.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/partitionrelease/RegionPartitionReleaseStrategy.java
GJL: True but I think a method should only do one thing. For performance reasons only, it's not worth it imo.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/partitionrelease/PipelinedRegionConsumedBlockingPartitions.java
GJL: The class already has default visibilty. We could reduce visibility for all it's methods but that makes harder to move the class to a different package if it is needed in the future. Moreover, if a class can be extended,`public` makes it clear what the interface of this class is):

https://stackoverflow.com/questions/5260467/public-methods-in-package-private-classes

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/partitionrelease/PipelinedRegion.java
GJL: check added

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/partitionrelease/PipelinedRegionExecutionView.java
GJL: good point, changed it

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/partitionrelease/PartitionReleaseStrategy.java
GJL: good catch, fixed

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/partitionrelease/PartitionReleaseStrategy.java
GJL: fixed

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/ExecutionGraph.java
zhijiangW: I still felt the following `createResultPartitionId` method seems complex because of involving in many components, I tried to find a way for constructing it easily.

Another options is that `IntermediateResultPartition` could get `ResultPartitionID` directly via internal `IntermediateResultPartitionID` and producer `ExecutionVertex`. Further `SchedulingResultPartition` could also provide the getter method for `ResultPartitionID` via constructing from `IntermediateResultPartition`. I am not sure whether it is worth doing so.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/partitionrelease/PipelinedRegionExecutionView.java
GJL: `PipelinedRegionExecutionStateView` would be most exact. I'll think about it.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/ExecutionGraph.java
```diff
@@ -1630,6 +1627,79 @@ public boolean updateState(TaskExecutionState state) {
...
+		}
+	}
+
+	private void releasePartitions(final IntermediateResultPartitionID resultPartition) {
+		LOG.info("Release partition {}", resultPartition);
+		final ResultPartitionID resultPartitionId = createResultPartitionId(resultPartition);
+		// TODO: finish implementation
+	}
+
+	private ResultPartitionID createResultPartitionId(final IntermediateResultPartitionID resultPartition) {
```
zhijiangW: add `Id` suffix for the parameter naming

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/partitionrelease/RegionPartitionReleaseStrategy.java
zhijiangW: We could still have two individual methods via adjusting the parameter as `PipelinedRegion` instead of `Set` in `initConsumedBlockingPartitionsByRegion`, and make the loop in constructor.
I believe it has no performance issue here, so you could keep the way as now. :)

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/failover/flip1/partitionrelease/PipelinedRegionExecutionViewTest.java
```diff
@@ -0,0 +1,75 @@
...
+		assertTrue(pipelinedRegionExecutionView.isFinished());
+	}
+
+	@Test
+	public void vertexCanBeUnfinished() {
+		final PipelinedRegion pipelinedRegion = PipelinedRegion.from(TEST_EXECUTION_VERTEX_ID);
+		final PipelinedRegionExecutionView pipelinedRegionExecutionView = new PipelinedRegionExecutionView(pipelinedRegion);
+
+		pipelinedRegionExecutionView.vertexFinished(TEST_EXECUTION_VERTEX_ID);
+		pipelinedRegionExecutionView.vertexUnfinished(TEST_EXECUTION_VERTEX_ID);
```
zhijiangW: Only confirm does it exist in real process that one finished vertex would be unfinished again?
If not, I wonder whether we need some addition checks in detail implementations of `vertexUnfinished/vertexFinished`.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/RegionPartitionReleaseStrategyTest.java
```diff
@@ -0,0 +1,149 @@
...
+
+		regionPartitionReleaseStrategy.vertexFinished(consumerVertex1);
+		regionPartitionReleaseStrategy.vertexFinished(consumerVertex2);
+		regionPartitionReleaseStrategy.vertexUnfinished(consumerVertex2);
+
+		final List<IntermediateResultPartitionID> partitionsToRelease = regionPartitionReleaseStrategy.vertexFinished(consumerVertex1);
+		assertThat(partitionsToRelease, is(empty()));
+	}
+
+	private static Set<PipelinedRegion> pipelinedRegionsSet(final PipelinedRegion... pipelinedRegions) {
```
zhijiangW: might be further deduplication via 
`pipelinedRegionsSet(ExecutionVertexID producerIds, ExecutionVertexID... consumerIds)`

zhijiangW:
I have finished the whole review @GJL !

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/failover/flip1/partitionrelease/PipelinedRegionExecutionViewTest.java
zentol: yes, it can happen in production that a finished vertex is reset. The simplest example is a global failover where some vertex was already finished.

There are also other cases connected to the new failover strategy, which may reset finished vertices if an upstream partition is required for a failover but has disappeared. Then this partitions producer is restarted, along with all downstream consumers, finished or not.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/ExecutionVertex.java
GJL: Fixing `DefaultFailverTopology` can be a separate PR

GJL: done

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/RegionPartitionReleaseStrategyTest.java
GJL: I prefer the current abstraction. 


In this signature it is unclear where the pipelined region boundaries are:

```
pipelinedRegionsSet(ExecutionVertexID producerIds, ExecutionVertexID... consumerIds)
```

```
pipelinedRegionsSet(onlyProducerVertexId, consumerVertex1, consumerVertex2);
```

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/partitionrelease/PartitionReleaseStrategy.java
GJL: Current Javadoc is comprehensible imo.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/ExecutionGraph.java
GJL: Not worth it at the moment because it is not even clear whether it is a good thing to use both `SchedulingTopology` and `FailoverTopology` in the release strategy.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/ExecutionGraph.java
GJL: done

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/PipelinedRegionComputeUtil.java
```diff
@@ -0,0 +1,133 @@
...
+
+	public static Set<Set<FailoverVertex>> computePipelinedRegions(final FailoverTopology topology) {
+		// currently we let a job with co-location constraints fail as one region
+		// putting co-located vertices in the same region with each other can be a future improvement
+		if (topology.containsCoLocationConstraints()) {
+			return uniqueRegions(buildOneRegionForAllVertices(topology));
+		}
+
+		// we use the map (list -> null) to imitate an IdentityHashSet (which does not exist)
+		// this helps to optimize the building performance as it uses reference equality
```
gaoyunhaii: This comment seems to be outdated since the result is now simply saved in a set backend by a IdentityHashMap.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/ExecutionGraph.java
zhijiangW: We might only need `SchedulingTopology` which was already used in below `createResultPartitionId` method.  It might be like this in `createResultPartitionId`.
`
SchedulingResultPartition schedulingResultPartition = schedulingTopology.getResultPartitionOrThrow(resultPartition);

ResultPartitionID resultPartitonId = schedulingResultPartition.getResultPartitionId();
`
If we add the `ResultPartitionID` info in the constructor of `DefaultSchedulingResultPartition`.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/PipelinedRegionComputeUtil.java
GJL: good catch, fixed

GJL:
rebased to master

GJL:
Squashed commits

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/ExecutionGraph.java
```diff
@@ -1605,36 +1629,9 @@ public boolean updateState(TaskExecutionState state) {
...
-						return true;
-
-					default:
-						// we mark as failed and return false, which triggers the TaskManager
-						// to remove the task
-						attempt.fail(new Exception("TaskManager sent illegal state update: " + state.getExecutionState()));
-						return false;
-				}
+				final boolean stateUpdated = updateStateInternal(state, attempt);
+				maybeReleasePartitions(state, attempt);
```
zentol: We should either take `stateUpdated` into account here, or check the actual execution state opposed to what is in the task state update.

If you have a vertex in a CANCELING state, which then receives a FINISHED state updated, the vertex completes the canceling and switches to CANCELED. In this situation we will already issue release calls for all partitions this vertex has produced.
The strategy however would still consider this vertex as finished, and potentially release preceding partitions.

zentol: pushed a fix or this

GJL: makes sense to me

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/ExecutionGraphPartitionReleaseTest.java
```diff
@@ -76,52 +81,78 @@ public void testStrategyNotifiedOfFinishedVerticesAndResultsRespected() throws E
...
 			NettyShuffleMaster.INSTANCE,
 			partitionTracker);
+
+		executionGraph.start(mainThreadExecutor);
+		run (executionGraph::scheduleForExecution);
+
+		return executionGraph;
+	}
+
+	private void run(RunnableWithException r) throws Exception {
```
GJL: I think you can use `TestingComponentMainThreadExecutor` and `TestingComponentMainThreadExecutor.Resource`

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/ExecutionGraphPartitionReleaseTest.java
```diff
@@ -110,6 +111,68 @@ public void testStrategyNotifiedOfFinishedVerticesAndResultsRespected() throws E
 		});
 	}
 
+	@Test
+	public void testStrategyNotifiedOfUnFinishedVertices() throws Exception {
+		// setup a pipeline of 3 failover regions (f1 -> f2 -> f3), where
```
GJL: What is `f3`? According to the description `sourceVertex` is in `f1`, `o1`, `o2`, `o3` are in `f2`. There are no more operators.

zentol: the comment is outdated, in a previous version there was another operator at the end, but this wasn't actually necessary for the test.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/ExecutionGraphPartitionReleaseTest.java
zentol: the resource does not provide access to the ScheduledExecutorService, which I need for the ExecutionGraph constructor.

zentol: The `TestingComponentMainThreadExecutor` can be used though.

GJL:
LGTM

PR #8786, opened 2019-06-18
[FLINK-12877][table][hive] Unify catalog database implementations

bowenli86:
## What is the purpose of the change

This PR unifies catalog database implementations.

## Brief change log

- unifies `GenericCatalogDatabase` and `HiveCatalogDatabase` into `CatalogDatabaseImpl`
- updates `HiveCatalog` accordingly
- removes `GenericCatalogDatabase`, `HiveCatalogDatabase`, and `AbstractCatalogDatabase`
- updates unit tests

## Verifying this change

This change is already covered by existing tests.

## Does this pull request potentially affect one of the following parts:

  - Dependencies (does it add or upgrade a dependency): (no)
  - The public API, i.e., is any changed class annotated with `@Public(Evolving)`: (no)
  - The serializers: (no)
  - The runtime per-record code paths (performance sensitive): (no)
  - Anything that affects deployment or recovery: JobManager (and its components), Checkpointing, Yarn/Mesos, ZooKeeper: (no)
  - The S3 file system connector: (no)

## Documentation

  - Does this pull request introduce a new feature? (no)
  - If yes, how is the feature documented? (not applicable)

PR #8805, opened 2019-06-20
 [FLINK-12902][hadoop] Remove flink-shaded-yarn-tests 

zentol:
Overall, this PR cleans up some dependency issues in `flink-yarn-tests`.

The dependency situation in this module was just insane; you have a few unshaded hadoop dependencies from `flink-yarn-tests`, some shaded hadoop dependencies from `flink-shaded-yarn-tests` and _yet another_ set of shaded hadoop dependencies from `flink-shaded-hadoop`. All of these were available on the test classpath, resulting in what is effectively an incomprehensible classpath.

This was highlighted in the shading patterns in `flink-shaded-yarn-tests`, which absolutely had to be in sync with `flink-shaded-hadoop` as otherwise you ended up with strange `VerifyErrors`.

This PR
a) removes `flink-shaded-yarn-tests` and moves a few of it's (now unshaded) dependencies to `flink-yarn-tests`
b) isolates `flink-yarn-tests`' test classpath from `flink-shaded-hadoop` via exclusions

As such, the test classpath now _only_ contains a few unshaded hadoop dependencies required for the test. `flink-shaded-hadoop` is only used for starting when submitting the applications to YARN, presumably to mirror what would be happening in production.

Additionally, this PR adds missing `test` scopes and removes an unused jackson dependency.


zentol:
green build: https://travis-ci.org/zentol/flink/builds/545701715

sunjincheng121:
The test fails caused by the PR #8786, And It fixed by #8807, So I think rebase the code can get the success CI test. :)

StephanEwen:
How about rethinking the Hadoop/Yarn dependencies a bit?

If we want a hadoop-free Flink, then no module should bundle and `flink-shaded-hadoop` dependency in *compile* scope. These dependencies should always be *provided*, meaning used for compilation, but would have to be dropped in externally during runtime.

Changing that in the `flink-yarn` modules allows us to drop all the exclusion rules downstream, like in `flink-yarn-tests` or in `flink-dist`, which seems the right way.

zentol:
@StephanEwen It is true that flink-yarn should have a `provided` dependency, but this looks like an orthogonal issue to me. We may need the excluion in flink-yarn-tests in any case though, given that provided dependencies are still available on the test classpath.

StephanEwen:
We can treat it as an orthogonal fix.

Are transitive provided dependencies added to the test class path in Maven? I did not expect that. The mysteries of maven...

zentol:
I double-checked how transitive provided dependencies are handled for tests, and they are indeed placed on the test classpath.

PR #8813, opened 2019-06-20
[FLINK-12891][hive] remove hadoop/hive writable from boundaries of Hive functions and Flink

bowenli86:

This PR removes hadoop/hive writable from boundaries of Hive functions and Flink because Flink only deals with java objects rather than hadoop/hive writables. Data is passed from Flink to Hive functions and from Hive functions back to Flink will always be simple java objects.

## Brief change log

- remove hadoop/hive writables conversions from `getConversion()` and `toFlinkObject()` in `HiveInspector`

## Verifying this change

This change is already covered by existing tests, such as `HiveSimpleUDFTest`, `HiveGenericUDFTest`, `HiveGenericUDTFTest`.

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/functions/hive/conversion/HiveInspectors.java
+++ b/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/functions/hive/conversion/HiveInspectors.java
@@ -50,9 +50,6 @@ import org.apache.hadoop.hive.serde2.objectinspector.primitive.FloatObjectInspec
 import org.apache.hadoop.hive.serde2.objectinspector.primitive.HiveCharObjectInspector;
 import org.apache.hadoop.hive.serde2.objectinspector.primitive.HiveVarcharObjectInspector;
 import org.apache.hadoop.hive.serde2.objectinspector.primitive.IntObjectInspector;
-import org.apache.hadoop.hive.serde2.objectinspector.primitive.JavaBinaryObjectInspector;
-import org.apache.hadoop.hive.serde2.objectinspector.primitive.JavaBooleanObjectInspector;
-import org.apache.hadoop.hive.serde2.objectinspector.primitive.JavaByteObjectInspector;
 import org.apache.hadoop.hive.serde2.objectinspector.primitive.JavaConstantBinaryObjectInspector;
 import org.apache.hadoop.hive.serde2.objectinspector.primitive.JavaConstantBooleanObjectInspector;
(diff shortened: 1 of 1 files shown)


bowenli86:
cc @xuefuz @JingsongLi @lirui-apache @zjuwangg

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/functions/hive/conversion/HiveInspectors.java
```diff
@@ -181,86 +168,22 @@ private static ConstantObjectInspector getPrimitiveJavaConstantObjectInspector(P
...
+					inspector instanceof IntObjectInspector ||
+					inspector instanceof LongObjectInspector ||
+					inspector instanceof FloatObjectInspector ||
+					inspector instanceof DoubleObjectInspector ||
+					inspector instanceof DateObjectInspector ||
+					inspector instanceof TimestampObjectInspector ||
+					inspector instanceof BinaryObjectInspector) {
+				return IdentityConversion.INSTANCE;
+			} else if (inspector instanceof HiveCharObjectInspector) {
+				return o -> new HiveChar((String) o, ((CharType) dataType.getLogicalType()).getLength());
```
xuefuz: Here we are converting Flink data to Hive data, and it seems that we probably should use the precision specified by Hive (in HiveCharObjectInspector) instead. The same for HiveVarchar. This might make a difference in certain cases.

bowenli86: the precision in HiveChar/VarcharObjectInspector here is actually set from the LogicalType, thus Flink's LogicalType is the single source of truth

bowenli86:
@xuefuz Thanks for your review!

Merging

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/functions/hive/conversion/HiveInspectors.java
xuefuz: I don't think this is about a single source of truth. It's a data conversion. We are converting data from Flink to Hive. While data type is char on both end, they may come with different precision. For instance, the data could be char(10), while hive's data type is char(5). During the conversion, the data should be enforced with precision 5, not 10.

bowenli86: discussed with Xuefu offline, we will leave this as-is

PR #8821, opened 2019-06-21
 [FLINK-12922][Table SQL / Planner] Remove method parameter from OperatorCodeGenerator

liyafan82:

The TableConfig parameter of OperatorCodGenerator#generateOneInputStreamOperator should be removed, because:

1. This parameter is never actually used.
2. If it is ever used in the future, we can use ctx.getConfig to get the same object
3. The method signature should be consistent. The method generateTwoInputStreamOperator does not have this parameter. So this parameter should also be removed.

## Brief change log

  - Remove the parameter from the method
  - Remove the parameter from method callers.
  

## Verifying this change

  - The serializers: (no)

## Documentation


JingsongLi:
LGTM +1, @liyafan82 can you squash these commits to a one?

liyafan82:
@JingsongLi commits squashed. Please take a look. Thank you.

PR #8822, opened 2019-06-21
[FLINK-12899][table] Introduce a resolved expression with data type

twalthr:

This separates calls to functions into resolved and unresolved call
expressions. Unresolved expressions are created in the API and are
translated to resolved expression. Resolved expressions are validated
and contain a return type. Table operations and planner should work
with resolved expressions only.

## Brief change log

- Introduction of {{ResolvedExpression}}
- Introduction of {{CallExpression}} with implements the above interface
- Ensure that only {{ResolvedExpression}}s leave the API

## Verifying this change

  - The serializers: (no

## Documentation


flink-table/flink-table-planner/src/test/scala/org/apache/flink/table/api/stream/table/ColumnFunctionsTest.scala
```diff
@@ -323,40 +323,6 @@ class ColumnFunctionsTest extends TableTestBase {
...
-        "DataStreamCalc",
-        streamTableNode(t),
-        term("select", "a AS d", "b")
-      )
-
-    verifyAll(tab1, tab2, expected)
-  }
-
-  @Test
-  def testDropColumns(): Unit = {
```
dawidwys: Why did you drop this test? I think it is a valid test to verify that we can drop a range of columns.

twalthr: Very good point. It was more of an accident.

twalthr:
Thanks @dawidwys. Merging...

PR #8824, opened 2019-06-21
[FLINK-12923][runtime] Introduce Task termination future

zentol:
Introduces a termination future into the task, that is completed once `run()` exits.

Don't be afraid of the diff, I simply wrapped the entire `run()` method in try-finally block.

### DIFF
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/Task.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/Task.java
@@ -233,6 +233,9 @@ public class Task implements Runnable, TaskActions, PartitionProducerStateProvid
 	/** Executor to run future callbacks. */
 	private final Executor executor;
 
+	/** Future that is completed once {@link #run()} exits. */
+	private final CompletableFuture<ExecutionState> terminationFuture = new CompletableFuture<>();
+
 	// ------------------------------------------------------------------------
 	//  Fields that control the task execution. All these fields are volatile
 	//  (which means that they introduce memory barriers), to establish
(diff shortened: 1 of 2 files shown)


flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/Task.java
```diff
@@ -509,317 +516,309 @@ public void startTaskThread() {
 	 */
 	@Override
 	public void run() {
+		try {
```
azagrebin:
```
@Override
public void run() {
    try {
        doRun();
    } finally {
         terminationFuture.complete(executionState);
    }
}

private void doRun() {
    // .... what we have in run now
}
```

flink-runtime/src/test/java/org/apache/flink/runtime/taskmanager/TaskTest.java
```diff
@@ -879,6 +879,62 @@ public void testTaskConfig() throws Exception {
 		task.getExecutingThread().join();
 	}
 
+	@Test
+	public void testTerminationFutureCompletesOnNormalExecution() throws Exception {
+		final Task task = createTaskBuilder()
+			.setInvokable(InvokableBlockingWithTrigger.class)
+			.setTaskManagerActions(new NoOpTaskManagerActions())
```
azagrebin: would it hurt to have `NoOpTaskManagerActions` as a default value for `TaskBuilder.taskManagerActions` instead of mock? The tests pass.

zentol: Shouldn't hurt, let's do this in a follow-up.

PR #8826, opened 2019-06-21
[FLINK-12479][operators] Integrate StreamInputProcessor(s) with mailbox

pnowojski:
This PR makes StreamInputProcessors truly non blocking and integrate availibility/finished states with the mailbox.

This PR depends on https://github.com/apache/flink/pull/8692

## Documentation

### DIFF
--- a/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamInputProcessor.java
+++ b/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamInputProcessor.java
@@ -19,6 +19,7 @@
 package org.apache.flink.streaming.runtime.io;
 
 import org.apache.flink.annotation.Internal;
+import org.apache.flink.runtime.io.AvailabilityListener;
 
 import java.io.Closeable;
 
@@ -26,9 +27,11 @@ import java.io.Closeable;
  * Interface for processing records by {@link org.apache.flink.streaming.runtime.tasks.StreamTask}.
  */
 @Internal
-public interface StreamInputProcessor extends Closeable {
+public interface StreamInputProcessor extends AvailabilityListener, Closeable {
 	/**
-	 * @return true if {@link StreamTaskInput} is finished.
+	 * @return true if {@link StreamInputProcessor} estimates that more records can be processed
+	 * immediately. Otherwise false, which means that there are no more records available at the
+	 * moment and the caller should check {@link #isFinished()} and/or {@link #isAvailable()}.
 	 */
 	boolean processInput() throws Exception;
 }
--- a/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamOneInputProcessor.java
+++ b/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamOneInputProcessor.java
@@ -43,6 +43,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
+import java.util.concurrent.CompletableFuture;
 
 import static org.apache.flink.util.Preconditions.checkNotNull;
 import static org.apache.flink.util.Preconditions.checkState;
@@ -129,29 +130,29 @@ public final class StreamOneInputProcessor<IN> implements StreamInputProcessor {
 	}
 
 	@Override
+	public boolean isFinished() {
+		return input.isFinished();
+	}
+
+	@Override
+	public CompletableFuture<?> isAvailable() {
+		return input.isAvailable();
+	}
+
+	@Override
 	public boolean processInput() throws Exception {
 		initializeNumRecordsIn();
 
 		StreamElement recordOrMark = input.pollNextNullable();
-		if (recordOrMark == null) {
-			input.isAvailable().get();
-			return !checkFinished();
-		}
-		int channel = input.getLastChannel();
-		checkState(channel != StreamTaskInput.UNSPECIFIED);
+		if (recordOrMark != null) {
+			int channel = input.getLastChannel();
+			checkState(channel != StreamTaskInput.UNSPECIFIED);
 
-		processElement(recordOrMark, channel);
-		return true;
-	}
-
-	private boolean checkFinished() throws Exception {
-		boolean isFinished = input.isFinished();
-		if (isFinished) {
-			synchronized (lock) {
-				operatorChain.endInput(1);
-			}
+			processElement(recordOrMark, channel);
 		}
-		return isFinished;
+		checkFinished();
+
+		return recordOrMark != null;
 	}
 
 	private void processElement(StreamElement recordOrMark, int channel) throws Exception {
@@ -177,6 +178,14 @@ public final class StreamOneInputProcessor<IN> implements StreamInputProcessor {
 			}
 		} else {
 			throw new UnsupportedOperationException("Unknown type of StreamElement");
+		}
+	}
+
+	private void checkFinished() throws Exception {
+		if (input.isFinished()) {
+			synchronized (lock) {
+				operatorChain.endInput(1);
+			}
 		}
 	}
 
--- a/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamTwoInputSelectableProcessor.java
+++ b/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamTwoInputSelectableProcessor.java
@@ -47,7 +47,6 @@ import org.slf4j.LoggerFactory;
 import java.io.IOException;
 import java.util.Collection;
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.ExecutionException;
 
 import static org.apache.flink.util.Preconditions.checkNotNull;
 import static org.apache.flink.util.Preconditions.checkState;
@@ -156,7 +155,21 @@ public final class StreamTwoInputSelectableProcessor<IN1, IN2> implements Stream
 		this.lastReadInputIndex = 1; // always try to read from the first input
 
 		this.isPrepared = false;
+	}
 
+	@Override
+	public boolean isFinished() {
+		return input1.isFinished() && input2.isFinished();
+	}
+
+	@Override
+	public CompletableFuture<?> isAvailable() {
+		if (inputSelection.isALLMaskOf2()) {
+			return isAnyInputAvailable();
+		} else {
+			StreamTaskInput input = (inputSelection.getInputMask() == InputSelection.FIRST.getInputMask()) ? input1 : input2;
+			return input.isAvailable();
+		}
 	}
 
 	@Override
@@ -179,18 +192,29 @@ public final class StreamTwoInputSelectableProcessor<IN1, IN2> implements Stream
 			if (recordOrMark != null) {
 				processElement1(recordOrMark, input1.getLastChannel());
 			}
+			checkFinished(input1, lastReadInputIndex);
 		} else {
 			recordOrMark = input2.pollNextNullable();
 			if (recordOrMark != null) {
 				processElement2(recordOrMark, input2.getLastChannel());
 			}
+			checkFinished(input2, lastReadInputIndex);
 		}
 
 		if (recordOrMark == null) {
 			setUnavailableInput(readingInputIndex);
 		}
 
-		return !checkFinished();
+		return recordOrMark != null;
+	}
+
+	private void checkFinished(StreamTaskInput input, int inputIndex) throws Exception {
+		if (input.isFinished()) {
+			synchronized (lock) {
+				operatorChain.endInput(getInputId(inputIndex));
+				inputSelection = inputSelector.nextSelection();
+			}
+		}
 	}
 
 	@Override
@@ -213,14 +237,12 @@ public final class StreamTwoInputSelectableProcessor<IN1, IN2> implements Stream
 		}
 	}
 
-	private int selectNextReadingInputIndex()
-		throws InterruptedException, ExecutionException, IOException {
-
-		int readingInputIndex;
-		while ((readingInputIndex = inputSelection.fairSelectNextIndexOutOf2(availableInputsMask, lastReadInputIndex)) == -1) {
-			if (!waitForAvailableInput(inputSelection)) {
-				return -1;
-			}
+	private int selectNextReadingInputIndex() throws IOException {
+		updateAvailability();
+		checkInputSelectionAgainstIsFinished();
+		int readingInputIndex = inputSelection.fairSelectNextIndexOutOf2(availableInputsMask, lastReadInputIndex);
+		if (readingInputIndex == -1) {
+			return -1;
 		}
 
 		// to avoid starvation, if the input selection is ALL and availableInputsMask is not ALL,
@@ -232,6 +254,27 @@ public final class StreamTwoInputSelectableProcessor<IN1, IN2> implements Stream
 		}
 
 		return readingInputIndex;
+	}
+
+	private void checkInputSelectionAgainstIsFinished() throws IOException {
+		if (inputSelection.isALLMaskOf2()) {
+			return;
+		}
+		if (inputSelection.isInputSelected(1) && input1.isFinished()) {
+			throw new IOException("Can not make a progress: only first input is selected but it is already finished");
+		}
+		if (inputSelection.isInputSelected(2) && input2.isFinished()) {
+			throw new IOException("Can not make a progress: only second input is selected but it is already finished");
+		}
+	}
+
+	private void updateAvailability() {
+		if (!input1.isFinished() && input1.isAvailable() == AVAILABLE) {
+			setAvailableInput(input1.getInputIndex());
+		}
+		if (!input2.isFinished() && input2.isAvailable() == AVAILABLE) {
+			setAvailableInput(input2.getInputIndex());
+		}
 	}
 
 	private void processElement1(StreamElement recordOrMark, int channel) throws Exception {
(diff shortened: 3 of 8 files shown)


pnowojski:
> Suggestion: you can rename old StreamInputProcessor to StreamOneInputProcessor first, in a separate commit. And introduce a StreamInputProcessor interface as follow up commit. This way it should preserve git history of the newly named StreamOneInputProcessor.

I didn't know that it can work like that. Will do :)

> Question: do I get it right that StreamTwoInputProcessor.processInput is still blocking, namely in barrierHandler.isAvailable().get()?

ops. Missed that. I'm not sure if it's worth adding some tests for that, it might be an overkill, especially that this will get a pretty good test coverage once we merge in progress/pending mailbox PRs.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamTwoInputSelectableProcessor.java
```diff
@@ -195,14 +216,11 @@ public void cleanup() throws Exception {
...
-		throws InterruptedException, ExecutionException, IOException {
-
+	private int selectNextReadingInputIndex() {
+		updateAvailability();
 		int readingInputIndex;
-		while ((readingInputIndex = inputSelection.fairSelectNextIndexOutOf2(availableInputsMask, lastReadInputIndex)) == -1) {
-			if (!waitForAvailableInput(inputSelection)) {
-				return -1;
-			}
+		if ((readingInputIndex = inputSelection.fairSelectNextIndexOutOf2(availableInputsMask, lastReadInputIndex)) == -1) {
```
1u0: Now the `readingInputIndex` doesn't need to be set in a loop. You can change this code to:
```java
int readingInputIndex = inputSelection.fairSelectNextIndexOutOf2(availableInputsMask, lastReadInputIndex);
if (readingInputIndex == -1) { // OR even delegate this check to next `if`
    return -1;
}
...
```

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamTwoInputSelectableProcessor.java
```diff
@@ -286,64 +316,33 @@ private void checkAndSetAvailable(int inputIndex) {
 		}
 	}
 
-	/**
-	 * @return false if both of the inputs are finished, true otherwise.
-	 */
-	private boolean waitForAvailableInput(InputSelection inputSelection)
-		throws ExecutionException, InterruptedException, IOException {
-
+	private CompletableFuture<?> waitForAvailableInput(InputSelection inputSelection) {
```
1u0: What about renaming all old `StreamTwoInputSelectable.waitFor...Input(...)` methods?
Now they are not actually waiting, but return "availability" future.
For example, they can be named as `getInputAvailable[Future](...), getOneInputAvailable[Future](...)`.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamTwoInputSelectableProcessor.java
```diff
@@ -286,64 +316,33 @@ private void checkAndSetAvailable(int inputIndex) {
...
-		throws ExecutionException, InterruptedException {
-
-		CompletableFuture<?> future1 = input1.isFinished() ? UNAVAILABLE : input1.isAvailable();
-		CompletableFuture<?> future2 = input2.isFinished() ? UNAVAILABLE : input2.isAvailable();
-
-		if (future1 == UNAVAILABLE && future2 == UNAVAILABLE) {
-			return false;
+	private CompletableFuture<?> waitForAvailableEitherInput() {
+		if (input1.isFinished() && input2.isFinished()) {
+			return AVAILABLE;
```
1u0: I think this should not happen, but if it's just to cover all cases, I'd prefer to return here `UNAVAILABLE`.

Additional note about renaming this method: "either" means **mutually exclusive** (out of two options) in English. I'd name this method something like `getAnyInputAvailableFuture()`.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamTwoInputSelectableProcessor.java
```diff
@@ -286,64 +316,33 @@ private void checkAndSetAvailable(int inputIndex) {
...
+		return CompletableFuture.anyOf(input1.isAvailable(), input2.isAvailable());
 	}
 
-	private void waitForOneInput(StreamTaskInput input)
-		throws IOException, ExecutionException, InterruptedException {
-
+	private CompletableFuture<?> waitForOneInput(StreamTaskInput input) {
 		if (input.isFinished()) {
-			throw new IOException("Could not read the finished input: input" + (input.getInputIndex() + 1) +  ".");
+			throw new IllegalStateException("Could not read the finished input: input" + (input.getInputIndex() + 1) +  ".");
```
1u0: Personally, I'd inline this method.
Otherwise:
 * `Preconditions.checkState`? 😃
 * Same here, suggestion to give it a new name.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
```diff
@@ -246,7 +258,18 @@ protected StreamTask(
...
-	protected abstract void performDefaultAction(ActionContext context) throws Exception;
+	protected void performDefaultAction(ActionContext context) throws Exception {
+		if (!inputProcessor.processInput()) {
+			if (inputProcessor.isFinished()) {
+				context.allActionsCompleted();
+			}
+			else {
+				context.actionsUnavailable();
+				inputProcessor.isAvailable().thenRun(
+					() -> ExceptionUtils.suppressExceptions(context::actionsAvailable));
```
1u0: This probably would be addressed by @StefanRRichter's changes. But just in case, what about the following version:
```java
inputProcessor.isAvailable().thenRunAsync(
    () -> ExceptionUtils.suppressExceptions(context::actionsAvailable), workerExecutor);
context.actionsUnavailable();
```

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamTwoInputSelectableProcessor.java
pnowojski: I have also realised, that this shouldn't be a `IllegalStateException`, since this is a user error - I have refactored the code accordingly.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
pnowojski: Yes, I've already changed this accordingly to the @StefanRRichter changes.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/consumer/SingleInputGate.java
```diff
@@ -211,6 +211,34 @@ public void setup() throws IOException {
 
 		BufferPool bufferPool = bufferPoolFactory.get();
 		setBufferPool(bufferPool);
+
+		requestPartitions();
+	}
+
+	private void requestPartitions() throws IOException, InterruptedException {
+		synchronized (requestLock) {
+			if (!requestedPartitionsFlag) {
```
StefanRRichter: Have you considered removing the `requestedPartitionsFlag`. It seems like this was just needed for lazy initialization and can potentially go now.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/consumer/UnionInputGate.java
```diff
@@ -141,17 +141,6 @@ public boolean isFinished() {
 		return inputGatesWithRemainingData.isEmpty();
 	}
 
-	@Override
-	public void requestPartitions() throws IOException, InterruptedException {
-		if (!requestedPartitionsFlag) {
```
StefanRRichter: Seems like this field can be removed completely from the class now.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamTaskNetworkInput.java
```diff
@@ -148,6 +148,9 @@ public boolean isFinished() {
 
 	@Override
 	public CompletableFuture<?> isAvailable() {
+		if (currentRecordDeserializer != null) {
```
StefanRRichter: Nit: I would use ternary operator instead of `if` here

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamOneInputProcessor.java
```diff
@@ -0,0 +1,241 @@
...
+		initializeNumRecordsIn();
+
+		StreamElement recordOrMark = input.pollNextNullable();
+		if (recordOrMark != null) {
+			int channel = input.getLastChannel();
+			checkState(channel != StreamTaskInput.UNSPECIFIED);
+
+			processElement(recordOrMark, channel);
+		}
+		checkFinished();
```
StefanRRichter: Any reason why we need to check this in both cases and not just `recordOrMark == null`? It uses `synchronized` so saving calls might be good.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamTwoInputSelectableProcessor.java
```diff
@@ -286,64 +334,17 @@ private void checkAndSetAvailable(int inputIndex) {
...
-	private boolean checkFinished() throws Exception {
-		if (getInput(lastReadInputIndex).isFinished()) {
-			synchronized (lock) {
-				operatorChain.endInput(getInputId(lastReadInputIndex));
-				inputSelection = inputSelector.nextSelection();
-			}
-		}
-
-		return input1.isFinished() && input2.isFinished();
+		return CompletableFuture.anyOf(input1.isAvailable(), input2.isAvailable());
```
StefanRRichter: You could strongly consider checking if both returned `AVAILABLE` and then also return `AVAILABLE`, and use the `anyOf` only if there is really something to wait for. Some code might only check for the referencial equality and then the method breaks the contract.

StefanRRichter: I would suggest this instead:
<code>
	private CompletableFuture<?> isAnyInputAvailable() {

		if (input1.isFinished()) {
			return input2.isFinished() ? AVAILABLE : input2.isAvailable();
		}

		if (input2.isFinished()) {
			return input1.isAvailable();
		}

		CompletableFuture<?> input1Available = input1.isAvailable();
		CompletableFuture<?> input2Available = input2.isAvailable();

		return (input1Available == AVAILABLE || input2Available == AVAILABLE) ?
			AVAILABLE : CompletableFuture.anyOf(input1Available, input2Available);
	}
<code>

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamTwoInputSelectableProcessor.java
```diff
@@ -162,31 +176,43 @@ public boolean processInput() throws Exception {
 			if (recordOrMark != null) {
 				processElement1(recordOrMark, input1.getLastChannel());
 			}
+			checkFinished(input1, lastReadInputIndex);
```
StefanRRichter: Similar question, could this be done only for the `recordOrMark == null` case?

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamTwoInputSelectableProcessor.java
```diff
@@ -162,31 +176,43 @@ public boolean processInput() throws Exception {
...
 		} else {
 			recordOrMark = input2.pollNextNullable();
 			if (recordOrMark != null) {
 				processElement2(recordOrMark, input2.getLastChannel());
 			}
+			checkFinished(input2, lastReadInputIndex);
```
StefanRRichter: Dito

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamTwoInputSelectableProcessor.java
```diff
@@ -195,14 +221,12 @@ public void cleanup() throws Exception {
...
-	private int selectNextReadingInputIndex()
-		throws InterruptedException, ExecutionException, IOException {
-
-		int readingInputIndex;
-		while ((readingInputIndex = inputSelection.fairSelectNextIndexOutOf2(availableInputsMask, lastReadInputIndex)) == -1) {
-			if (!waitForAvailableInput(inputSelection)) {
-				return -1;
-			}
+	private int selectNextReadingInputIndex() throws IOException {
+		updateAvailability();
```
StefanRRichter: Looks weird to me to have this check (and maybe also the one on the next line) in this method. I would call this as part of `processInput` instead, e.g. before this method is called. I know it does the same, but somehow it feels like the check belong into the caller method and not down here.

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/consumer/SingleInputGateTest.java
```diff
@@ -189,51 +189,55 @@ public void testBackwardsEventWithUninitializedChannel() throws Exception {
...
 		// Setup reader with one local and one unknown input channel
 
-		final SingleInputGate inputGate = createInputGate();
-		final BufferPool bufferPool = mock(BufferPool.class);
-		when(bufferPool.getNumberOfRequiredMemorySegments()).thenReturn(2);
+		NettyShuffleEnvironment environment = createNettyShuffleEnvironment();
+		final SingleInputGate inputGate = createInputGate(environment, 2, ResultPartitionType.PIPELINED);
+		try {
+			final BufferPool bufferPool = mock(BufferPool.class);
+			when(bufferPool.getNumberOfRequiredMemorySegments()).thenReturn(2);
```
1u0: `bufferPool` is not used here anymore and can be removed.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamInputProcessor.java
```diff
@@ -18,213 +18,17 @@
...
- * {@link StreamStatus} events, and forwards them to event subscribers once the
- * {@link StatusWatermarkValve} determines the {@link Watermark} from all inputs has advanced, or
- * that a {@link StreamStatus} needs to be propagated downstream to denote a status change.
- *
- * <p>Forwarding elements, watermarks, or status status elements must be protected by synchronizing
- * on the given lock object. This ensures that we don't call methods on a
- * {@link OneInputStreamOperator} concurrently with the timer callback or other things.
- *
- * @param <IN> The type of the record that can be read with this record reader.
+ * Interface for processing records by StreamTask.
```
1u0: Suggestion, use proper class link:
`Interface for processing records by {@link org.apache.flink.streaming.runtime.tasks.StreamTask}.`

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamInputProcessor.java
```diff
@@ -18,213 +18,17 @@
...
  */
-@Internal
```
1u0: Keep the `@Internal` annotation?

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamInputProcessor.java
```diff
@@ -18,213 +18,17 @@
...
-				}
-			} catch (Exception e) {
-				throw new RuntimeException("Exception occurred while processing valve output stream status: ", e);
-			}
-		}
-	}
+public interface StreamInputProcessor extends AvailabilityListener, Closeable {
+	/**
+	 * @return false if probably there are not more records available and caller should check
+	 * {@link #isAvailable()}. return true if {@link #isAvailable()} check can be safely omitted.
```
1u0: I'd rephrase this paragraph. Maybe something like:
```
Return true if more records can be processed.
False, if there are no more records available at the moment and the caller should check {@link #isFinished()} and/or {@link #isAvailable()}.
```

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
```diff
@@ -243,26 +249,15 @@ protected StreamTask(
...
-					letter.run();
-				}
+	protected void performDefaultAction(DefaultActionContext context) throws Exception {
+		if (!inputProcessor.processInput()) {
+			if (inputProcessor.isFinished()) {
+				context.allActionsCompleted();
+			}
+			else {
+				SuspendedMailboxDefaultAction suspendedDefaultAction = context.suspendDefaultAction();
+				inputProcessor.isAvailable().thenRun(suspendedDefaultAction::resume);
```
StefanRRichter: I would suggest to change this line to `inputProcessor.isAvailable().thenRunAsync(suspendedDefaultAction::resume, getTaskMailboxExecutor());` so that we ensure the resume is executed in the mailbox thread and not by another thread that might get interrupted before it is able to deliver the resume message to the mailbox.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/consumer/SingleInputGate.java
pnowojski: That might be not that simple, since this flag is also used during `updateInputChannel` from unknown channel. I'm not sure, but update probably can happen concurrently to the setup.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamTaskNetworkInput.java
pnowojski: Personally I'm not a fan of them, I hope that won't be a review blocker ;)

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamOneInputProcessor.java
pnowojski: But it only does it if the input `isFinished()`, so that should be a non issue.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamTwoInputSelectableProcessor.java
pnowojski: Hmmm, I would lightly disagree. `updateAvailability()` is (I think) only relevant for `selectNextReadingInputIndex()` so it should be tight to it in case of:
- someones refactor the code and moves `selectNextReadingInputIndex` somewhere else
- someones adds another code path to `selectNextReadingInputIndex`

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamInputProcessor.java
pnowojski: I've rephrased your rephrased version to:
```
	/**
	 * @return true if {@link StreamInputProcessor} estimates that more records can be processed
	 * immediately. Otherwise false, which means that there are no more records available at the
	 * moment and the caller should check {@link #isFinished()} and/or {@link #isAvailable()}.
	 */
```

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/execution/MailboxProcessor.java
```diff
@@ -286,12 +288,20 @@ public void resume() {
 			if (taskMailboxExecutor.isMailboxThread()) {
 				resumeInternal();
 			} else {
-				try {
-					mailbox.putMail(this::resumeInternal);
-				} catch (InterruptedException ie) {
+				boolean delivered = false;
+				boolean interrupted = false;
+				while (!delivered) {
```
StefanRRichter: With the latest update to my PR, making the mailbox unbounded, this whole fix should no longer be required.

pnowojski:
I have applied the requested changes and rebased the PR on the latest master branch.

pnowojski:
Travis is green, modulo one timeout in end-to-end tests when connecting to apache.org. Merging.

PR #8829, opened 2019-06-21
[FLINK-12932][table][sql client] support 'show catalogs' and 'show databases' end-2-end in TableEnvironment and SQL CLI

bowenli86:
## What is the purpose of the change

This PR adds 'show catalogs' and 'show databases' end-2-end in TableEnvironment and SQL CLI.

## Brief change log

- added `listCatalogs` and `listDatabases` API to `TableEnvironment` and implementations in subclasses
- added 'show catalogs' and 'show databases'  to SQL CLI

## Verifying this change

This change added tests and can be verified as follows:

added unit tests in `ExecutionContextTest` and `LocalExecutorITCase`

## Does this pull request potentially affect one of the following parts:

  - Dependencies (does it add or upgrade a dependency): (no)
  - The public API, i.e., is any changed class annotated with `@Public(Evolving)`: (yes)
  - The serializers: (yes)
  - The runtime per-record code paths (performance sensitive): (no)
  - Anything that affects deployment or recovery: JobManager (and its components), Checkpointing, Yarn/Mesos, ZooKeeper: (no)
  - The S3 file system connector: (no)

## Documentation

  - Does this pull request introduce a new feature? (yes)
  - If yes, how is the feature documented? (docs)

Docs will added later separately

PR #8830, opened 2019-06-22
[FLINK-12933][sql client] support 'use catalog' and 'use database' in SQL CLI

bowenli86:

This PR adds `USE CATALOG` and `USE DATABASE` commands to SQL CLI.

This PR depends on https://github.com/apache/flink/pull/8829.

## Brief change log

- added `USE CATALOG` and `USE DATABASE` commands to SQL CLI.
- updated unit tests in `ExecutionContextTest` and `LocalExecutorITCase` to cover these changes

## Verifying this change

This change added tests and can be verified as follows: `ExecutionContextTest` and `LocalExecutorITCase` 

  - The serializers: (no)

## Documentation

Documentation will be added later

### DIFF
--- a/flink-python/pyflink/table/table_environment.py
+++ b/flink-python/pyflink/table/table_environment.py
@@ -183,11 +183,29 @@ class TableEnvironment(object):
         j_table_path = utils.to_jarray(gateway.jvm.String, table_path_continued)
         self._j_tenv.insertInto(table._j_table, table_path, j_table_path)
 
+    def list_catalogs(self):
+        """
+        Gets the names of all catalogs registered in this environment.
+
+        :return: List of catalog names.
+        """
+        j_catalog_name_array = self._j_tenv.listCatalogs()
+        return [item for item in j_catalog_name_array]
+
+    def list_databases(self):
+            """
+            Gets the names of all databases in the current catalog.
+
+            :return: List of database names in the current catalog.
+            """
+            j_database_name_array = self._j_tenv.listDatabases()
+            return [item for item in j_database_name_array]
+
     def list_tables(self):
         """
-        Gets the names of all tables registered in this environment.
+        Gets the names of all tables in the current database of the current catalog.
 
-        :return: List of table names.
+        :return: List of table names in the current database of the current catalog.
         """
         j_table_name_array = self._j_tenv.listTables()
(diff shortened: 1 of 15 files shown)


bowenli86:
@KurtYoung Extremely simple commands like this can be on sql cli itself IMO. DDL/DML is better to go thru sql parser.

I also not against to migrating these simple commands to sql parser *later* when it is ready. Right now it's about making end-to-end user experience complete for catalogs in sql cli.

danny0405:
@KurtYoung @bowenli86 Since we have a new module named `flink-sql-parser` aimed to handle all the sql parse work including DDL/DML/DQL, which means the parser module is the only entrance for any sql parse work in the future, the `Use catalog` and `Use database` are also expected to put there.

IMHO, sql client could have its own commands but should exclude the DDL/DML/DQL,  which is the duty of sql parser.

bowenli86:
@danny0405 Yeah, as I mentioned, I think it's ok to move DQL like https://github.com/apache/flink/pull/8829 and "USE" to sql parser when sql client officially depend on sql parser (which is not yet). IIRC, there are still some community concerns on sql parser, right?

Thus, I propose to add the functionality first so we can 1) enable end-to-end test, both automatic and manual, for SQL CLI, catalogs, and hive integration, and 2) guarantee the feature is out in 1.9 for users, independent of the progress of sql-parser. We can migrate these commands to sql parser later (shouldn't be hard).

bowenli86:

flink-python/pyflink/table/table_environment.py
```diff
@@ -183,11 +183,29 @@ def insert_into(self, table, table_path, *table_path_continued):
...
+        """
+        Gets the names of all catalogs registered in this environment.
+
+        :return: List of catalog names.
+        """
+        j_catalog_name_array = self._j_tenv.listCatalogs()
+        return [item for item in j_catalog_name_array]
+
+    def list_databases(self):
+            """
```
xuefuz: I'm wondering why the indention is different from others here. Not sure if it's a github showing rendering issue.

bowenli86: this time it's actually a real indentation issue. I've updated https://github.com/apache/flink/pull/8829 and rebased this PR to #8829

PR #8831, opened 2019-06-22
[FLINK-12935][hive] package flink-connector-hive into flink distribution for HiveCatalog

bowenli86:

This PR adds `flink-connector-hive` jar into Flink distribution to `/opt` dir. This is mainly to include `HiveCatalog` rather than the hive data connector.

## Brief change log

adds `flink-connector-hive` jar into Flink distribution to`/opt` dir.

## Verifying this change

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-dist/src/main/assemblies/opt.xml
+++ b/flink-dist/src/main/assemblies/opt.xml
@@ -192,6 +192,14 @@
 			<destName>flink-python-${project.version}-java-binding.jar</destName>
 			<fileMode>0644</fileMode>
 		</file>
+
+		<!-- HiveCatalog and Hive connector -->
+		<file>
+			<source>../flink-connectors/flink-connector-hive/target/flink-connector-hive_${scala.binary.version}-${project.version}.jar</source>
+			<outputDirectory>opt</outputDirectory>
+			<destName>flink-connector-hive_${scala.binary.version}-${project.version}.jar</destName>
+			<fileMode>0644</fileMode>
+		</file>
 	</files>
 	<fileSets>
 		<fileSet>


zentol:
we don't have _any_ other connector in the distribution, what's the justification for doing so now?

bowenli86:
> we don't have _any_ other connector in the distribution, what's the justification for doing so now?

Hi @zentol, its purpose is mainly to include `HiveCatalog` for users to manage their metadata, rather than including the Hive data connector (source/sink/format)

bowenli86:

bowenli86:

Merging

zentol:
I'm really worried that we're setting a precedent here and soon afterwards will get requests to add more connectors to the distribution.

IMO the distribution is overloaded as is, and I'd prefer if users had to download the jar manually.

I'm also wondering whether it wouldn't make sense to separate the connector from the catalog if they serve such different purposes.

bowenli86:
Hi @zentol , we decided a long time ago with Fabian and Timo to have HiveCatalog in flink-connector-hive because it's a metadata connector. And all components (metadata connector, source, sink, hive udf) in flink-connector-hive shares lots of code. Moving it to its own module just because of the concern of Flink  distribution size seems a bit overkill to me.

One critical functionality of HiveCatalog is to enable users to persist their Flink metadata (table, view, partition, udf, etc) in Hive metastore, and I think including it in Flink distribution is important to the user experience completeness for Table API and SQL. flink-connector-hive jar is only 117kb.

Yeah, I also resonate with you on the concern that other connectors may try to follow and squeeze in. How about adding a `/catalog` dir within `/opt` to bring up attention that adding flink-connector-hive is not about data connector but catalog/metadata connecter? Not many other connectors would fit in this requirement.

zentol:
Can we leave it out for 1.9 and revisit this for 1.10? There are some ideas floating around for splitting the repository (I believe Stephan wants to open a ML thread soonish), and this may or may not have repercussions on what is bundled in flink-dist.
If this results in separate release cycles for connectors and the flink core (TBD), there may be no real point in bundling anything in /opt since it's likely out-dated before the next release of the connector. In this model, copying jars into dist would be a common upgrade pattern.

I'd rather not add a new jar to /opt in one release if there is a possibility that we remove it in the next one again.

This is admittedly an unfair argument given that this wasn't really brought up so far anywhere and there's lots of uncertainty.

As for the proposal to add a `catalog` sub-directory, I had a similar discussion when it came to the plugin directory recently. My take was that it should have sub-directories for different categories, like `reporters`, `filesystems` etc., but it was rightfully pointed out to me that a given jar may fit into multiple categories.
I'm wondering whether the same logic doesn't apply here as well; ideally we'd follow the same logic for `/opt` and `/plugins`.

bowenli86:
@zentol ok, let's postpone this to 1.10. Closing this PR for now

PR #8839, opened 2019-06-22
[FLINK-12952][docs] Minor doc correction to incremental window functions

mans2singh:

Minor correction to documentation regarding incremental windows functions

## Brief change log

The Flink documentation [Window Functions](https://ci.apache.org/projects/flink/flink-docs-release-1.8/dev/stream/operators/windows.html#window-functions) mentions that

> The window function can be one of ReduceFunction, AggregateFunction, FoldFunction or ProcessWindowFunction. The first two can be executed more efficiently

It should perhaps state (since FoldFunction, though deprecated, is also incremental):

> The first three can be executed more efficiently


mans2singh:
@klion26 @rmetzger  - Do you have any thoughts on this doc correction ?  Thanks

zentol:
Seems incorrect based on the information in FLINK-5720.

```
Folding is an operation that cannot be done incrementally in a distributed way and that also cannot be done on merging windows.
```

PR #8840, opened 2019-06-23
[FLINK-12931][python] Fix lint-python.sh cannot find flake8

dianfu:
## What is the purpose of the change

*The `lint-python.sh` will execute failed during installing `flake8` when there is `anaconda2/anaconda3` installed. We need to solve this issue because `anaconda` is a pretty common package that developers install and use. 
The reason to this issue is that `conda install flake8` will try to install `flake8` into the conda home which may be the conda installed by user, not the `miniconda` installed by lint-python.sh.*

## Brief change log

  - *Updates lint-python.sh to specify the conda home explicitly when executing `conda remove` and `conda install`*

## Verifying this change

This change can be verified as follows:

  - *Executes ./dev/lint-python.sh and it will succeed*
  - *Installs anaconda2/anaconda3, starts a new shell and makes sure the newly installed anaconda2/anaconda3 is used.*
  - *Executes ./dev/lint-python.sh -f and it will fail without the changes of this PR*

## Does this pull request potentially affect one of the following parts:

  - Dependencies (does it add or upgrade a dependency): (no)
  - The public API, i.e., is any changed class annotated with `@Public(Evolving)`: (no)
  - The serializers: (no)
  - The runtime per-record code paths (performance sensitive): (no)
  - Anything that affects deployment or recovery: JobManager (and its components), Checkpointing, Yarn/Mesos, ZooKeeper: (no)
  - The S3 file system connector: (no)

## Documentation

  - Does this pull request introduce a new feature? (yes / no)
  - If yes, how is the feature documented? (not applicable / docs / JavaDocs / not documented)

PR #8841, opened 2019-06-23
[FLINK-12765][coordinator] Bookkeeping of available resources of allocated slots in SlotPool

gaoyunhaii:

This PR is to introduce the bookkeeping logic for the shared slots and colocated slots. It is a part of introducing the fine-grained resource management for Flink and is based on #8704 #8840 and #8846 . Based on the current design, a task will always request the resource according to its own resource need, and the returned slot may be larger than requested resource. Therefore, it leaves chance for slot sharing and colocation group.

To considering resources during slot sharing and colocation, the resource already requested must be recorded to ensure the resource requests of the assigned tasks do not exceed the total resource of the slot.  

The first step is to bookkeep the resources inside the slot sharing tree structure. A variable is added to each MultiTaskSlot to record the resources already requests by all its descendants. 

For the allocating, there are in general two scenarios:
1. The underlying slot has been allocated. 
2. The underlying slot is requested but not allocated.

For the first type of slots, the allocation is modified to use the remaining resource of the slot instead of the total resource. For slot sharing, if no allocated slots can be fulfilled, it will turn to slots in requesting or request new slot as before. For colocation, if the slot has been allocated but it cannot fulfill the current request, then the request will never be fulfilled and will fail directly.

For the second type of slots, since the total resource of the slots are not known before they get allocated, the requests will be recorded temporarily. When the underlying slot get allocated, the requests will be checked whether they can be fulfilled. Some requests may be failed due to over-allocation. For the failed tests:
1. If only a part of requests are failed, they may succeed if requests without others, therefore these requests can be retry. The colocation requests are considered as a whole instead of the single task requests.
2. If all the requests can not be fulfilled, this case can only happen when we request to RM according to the resource of a task in a colocation group, but the return slot can not fulfill the colocation group and any other requests in the same sharing group. Since for static resource matching this exception cannot be recovered, and to avoid infinite retrying, currently we do not allow the requests to retry. 

3. If the request cannot be fulfilled by RM, RM will return an error directly and the requests will be failed as before.

## Brief change log

1. cfc25388147495a33ba9897a439db44dfc8b8c59 introduces merge/subtract on `ResourceProfile` and 122399a7caa515c1b4390cac7d092e897f0aebe0 enhance the tests.
2. 4f81f937ba1d0900bdfacf9b6d0253cc931535cf changes the default resource in ResourceSpec from DEFAULT to UNKNOWN.
3. 62388dba3fb3e36b5f2fa7bcd473dc998c642639 disables jobs who only have parts of resources configured. This ensure the resources of job vertices are either all UNKNOWN or all configured.
4. 869e800e12ffc33cf6585f94194277076c2009b9 revises the codes of the class of Resource.java.

Based on the above commits,

5. c26aa1763187221961b76a68635a8fc75947eff7 maintains the resource requested in the MutiTaskSlot, which indicates the total resources requested by all its descendant.
6. 2f6add017d250c520265f98ba21b938d699cff9e adds the resource checking logic when the underlying slot is resolved. The over-allocated requests will be marked fail. The over-allocated requests will iff some requests are fulfilled by the underlying slot in `SchedulerImpl`. For the co-located requests, it also adds the checking of whether the remaining resource is possible to fulfill the requests if the underlying slot is already resolved.
7. c9cf151a68344a2aac328983593a7dd07a267ac5 modifies the interface of `SlotSelectionStrategy` to also pass the remaining resource of the underlying slot. The implementation of the strategies should use the remaining resource instead of the total resource.

## Verifying this change

  - Added tests that validate the calculated of requested resource for the hierarchical structure of MultiTaskSlot/SingleTaskSlot. 
  - Added tests that validate the routine to fail the over-allocated requests when the underlying slot is resolved.
  - Added tests that validate the retry logic after failing due to over-allocation.
  - Added tests that validate the failure of the co-located requests if the slot is not enough for all the co-located tasks.

  - The serializers: **no**
  - Anything that affects deployment or recovery: JobManager (and its components), Checkpointing, Yarn/Mesos, ZooKeeper: **yes**

## Documentation

  - If yes, how is the feature documented? [Doc](https://docs.google.com/document/d/1UR3vYsLOPXMGVyXHXYg3b5yNZcvSTvot4wela8knVAY/edit?usp=sharing)

### DIFF
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/clusterframework/types/ResourceProfile.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/clusterframework/types/ResourceProfile.java
@@ -20,6 +20,7 @@ package org.apache.flink.runtime.clusterframework.types;
 
 import org.apache.flink.api.common.operators.ResourceSpec;
 import org.apache.flink.api.common.resources.Resource;
+import org.apache.flink.util.Preconditions;
 
 import javax.annotation.Nonnull;
 
@@ -29,6 +30,8 @@ import java.util.HashMap;
 import java.util.Iterator;
 import java.util.Map;
 import java.util.Objects;
+
+import static org.apache.flink.util.Preconditions.checkArgument;
 
 /**
  * Describe the immutable resource profile of the slot, either when requiring or offering it. The profile can be
@@ -48,7 +51,7 @@ public class ResourceProfile implements Serializable, Comparable<ResourceProfile
 
 	private static final long serialVersionUID = 1L;
 
-	public static final ResourceProfile UNKNOWN = new ResourceProfile(-1.0, -1);
+	public static final ResourceProfile UNKNOWN = new ResourceProfile();
 
 	/** ResourceProfile which matches any other ResourceProfile. */
 	public static final ResourceProfile ANY = new ResourceProfile(Double.MAX_VALUE, Integer.MAX_VALUE, Integer.MAX_VALUE, Integer.MAX_VALUE, Integer.MAX_VALUE, Integer.MAX_VALUE, Collections.emptyMap());
@@ -97,6 +100,12 @@ public class ResourceProfile implements Serializable, Comparable<ResourceProfile
 			int networkMemoryInMB,
 			int managedMemoryInMB,
 			Map<String, Resource> extendedResources) {
+		Preconditions.checkArgument(cpuCores >= 0);
+		Preconditions.checkArgument(heapMemoryInMB >= 0);
+		Preconditions.checkArgument(directMemoryInMB >= 0);
+		Preconditions.checkArgument(nativeMemoryInMB >= 0);
+		Preconditions.checkArgument(networkMemoryInMB >= 0);
+		Preconditions.checkArgument(managedMemoryInMB >= 0);
 		this.cpuCores = cpuCores;
 		this.heapMemoryInMB = heapMemoryInMB;
 		this.directMemoryInMB = directMemoryInMB;
@@ -116,6 +125,18 @@ public class ResourceProfile implements Serializable, Comparable<ResourceProfile
 	 */
 	public ResourceProfile(double cpuCores, int heapMemoryInMB) {
 		this(cpuCores, heapMemoryInMB, 0, 0, 0, 0, Collections.emptyMap());
+	}
+
+	/**
+	 * Creates a special ResourceProfile with negative values, indicating resources are unspecified.
+	 */
+	private ResourceProfile() {
+		this.cpuCores = -1.0;
+		this.heapMemoryInMB = -1;
+		this.directMemoryInMB = -1;
+		this.nativeMemoryInMB = -1;
+		this.networkMemoryInMB = -1;
+		this.managedMemoryInMB = -1;
 	}
 
 	/**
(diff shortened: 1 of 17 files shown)


flink-runtime/src/main/java/org/apache/flink/runtime/clusterframework/types/ResourceProfile.java
```diff
@@ -260,6 +265,102 @@ public int compareTo(@Nonnull ResourceProfile other) {
...
+		}
+
+		return result;
+	}
+
+	private int addIntegersConsideringOverflow(int first, int second) {
+		int result = first + second;
+
+		if (result < 0) {
+			return Integer.MAX_VALUE;
```
xintongsong: If we assume both integers to be added are non-negative, we should make this assumption explicit. Maybe change the method name to addNonNegtiveIntegersConsideringOverflow?

flink-runtime/src/main/java/org/apache/flink/runtime/jobmaster/slotpool/SlotSharingManager.java
```diff
@@ -177,14 +181,18 @@ MultiTaskSlot createRootSlot(
...
+	public Collection<SlotSelectionStrategy.SlotInfoAndRemainingResource> listResolvedRootSlotInfo(@Nullable AbstractID groupId) {
 		return resolvedRootSlots
 			.values()
 			.stream()
-			.flatMap((Map<AllocationID, MultiTaskSlot> map) -> map.values().stream())
-			.filter((MultiTaskSlot multiTaskSlot) -> !multiTaskSlot.contains(groupId))
-			.map((MultiTaskSlot multiTaskSlot) -> (SlotInfo) multiTaskSlot.getSlotContextFuture().join())
-			.collect(Collectors.toList());
+				.flatMap((Map<AllocationID, MultiTaskSlot> map) -> map.values().stream())
+				.filter((MultiTaskSlot multiTaskSlot) -> !multiTaskSlot.contains(groupId))
```
xintongsong: indent

flink-runtime/src/main/java/org/apache/flink/runtime/jobmaster/slotpool/SlotSharingManager.java
```diff
@@ -316,6 +331,9 @@ public boolean contains(AbstractID groupId) {
 		// true if we are currently releasing our children
 		private boolean releasingChildren;
 
+		// the total requested by all the descendants.
+		private ResourceProfile requestedResources;
```
xintongsong: I think this is a little confusing. Do you mean the resource requested by tasks that are already assigned into the slot? May be use "occupied" or "allocated" to avoid ambiguity?

flink-runtime/src/main/java/org/apache/flink/runtime/jobmaster/slotpool/SlotSharingManager.java
```diff
@@ -347,20 +365,65 @@ private MultiTaskSlot(
...
+								childrenToEvict.size(),
+								slotContext.getResourceProfile(),
+								allocated,
+								childrenToEvict);
 					}
-				});
+
+					if (childrenToEvict.size() == children.size()) {
+						// This only happens when we request to RM using the resource profile of a task
+						// who is belonging to a CoLocationGroup. Similar to dealing with the fail of
```
xintongsong: It's not clear to me why this only happens for CoLocationGroup.

flink-runtime/src/main/java/org/apache/flink/runtime/jobmaster/slotpool/SlotSharingManager.java
```diff
@@ -509,6 +608,22 @@ private void releaseChild(AbstractID childGroupId) {
 			}
 		}
 
+		private void onResourceRequested(ResourceProfile resourceProfile) {
```
xintongsong: requested -> allocated/occupied

flink-runtime/src/main/java/org/apache/flink/runtime/jobmaster/slotpool/SlotSharingManager.java
gaoyunhaii: I still thought that requested might be better since the resources may have been requested before the underlying slot has been allocated for real.

flink-runtime/src/main/java/org/apache/flink/runtime/jobmaster/slotpool/SlotSharingManager.java
gaoyunhaii: This is because RM always returns a slot whose resource is larger than the requested one, without co-location, there should be at least the one who triggers the request to RM get fulfilled.

flink-runtime/src/main/java/org/apache/flink/runtime/jobmaster/slotpool/SlotSharingManager.java
StephanEwen: I agree that "requested" sounds confusing. I would like to change this to "reserved" if there are no objections.

StephanEwen:
Manually merged in dac02696a9bdb3ecdfd084b784a43e74ca3d9098

PR #8815, opened 2019-06-21
[FLINK-12918][table][hive] unify GenericCatalogTable, HiveCatalogTable and AbstractCatalogTable into CatalogTableImpl

bowenli86:
## What is the purpose of the change

This PR unifies implementations of CatalogTable by combining GenericCatalogTable, HiveCatalogTable and AbstractCatalogTable into CatalogTableImpl.

## Brief change log

- combined GenericCatalogTable, HiveCatalogTable and AbstractCatalogTable into CatalogTableImpl
- moved IS_GENERIC flag from GenericInMemoryCatalog to CatalogConfig
- updated unit tests

## Verifying this change

This change added tests and can be verified as follows: `GenericInMemoryCatalogTest`, `HiveCatalogHiveMetadataTest`, `HiveCatalogGenericMetadataTest`

## Does this pull request potentially affect one of the following parts:

  - Dependencies (does it add or upgrade a dependency): (no)
  - The public API, i.e., is any changed class annotated with `@Public(Evolving)`: (no)
  - The serializers: (no)
  - The runtime per-record code paths (performance sensitive): (no)
  - Anything that affects deployment or recovery: JobManager (and its components), Checkpointing, Yarn/Mesos, ZooKeeper: (no)
  - The S3 file system connector:(no)

## Documentation

  - Does this pull request introduce a new feature? (yes)
  - If yes, how is the feature documented? (not applicable)

PR #3829, opened 2017-05-05
[FLINK-6442] [table] Extend TableAPI Support Sink Table Registration …

lincoln-lil:
Support Sink Table Registration and ‘insert into’ Clause in SQL:
1. support registering a sink table (like source table registration, and will do validation according to the registered table)
2. support ‘insert into’ clause in SQL
3. do not support insert partial columns of the target table due to table columns haven’t nullable property definition for now

Tried to implement translating the whole SQL string via converter rules, while in Flink,  sink operation has no output but all RelNodes' translation should return a DataSet/DataStream, and Calcite's TableModify defines output for DML(INSERT,DELETE,UPDATE) is a RowCount column of BigInt type.
So separate source translation if DML insert,  this a workaround since we have no conclusion for the output type definition of sink operation(compare to traditional SQL insert).

detail design see: https://goo.gl/n3phK5

PR #8844, opened 2019-06-24
[FLINK-12951][table-planner] Add logic to bridge DDL to table source(…

danny0405:

This pull request add logic to convert a sql DDL to table source/sink.

## Brief change log

  - Add `TableEnvironment#sql()` and `TableEnvironment#registerTable()`
  - Add transform logic for all kinds of sql node in SqlExecutableStatements to table Operations
  - Add new table Operation `CreateOperation` for all the DDLs
  - Add computed columns/PKs/UKs support for `TableSchema`

## Verifying this change

This change is covered by test cases `CatalogTableITCase` and `SqlExecutableStatementsTest`.

  - The public API, i.e., is any changed class annotated with `@Public(Evolving)`: yes, the TableSchema supports PK/UK/Computed Column
  - The serializers: no

## Documentation


twalthr:
Maybe @dawidwys can take a look at this changes around the catalog managers and table environment. We should make sure that this change fits in the overall architecture. I saw some `CatalogTable` interface exposed at the API level, we should definitely not do that.

danny0405:
@twalthr Thanks, i have replace the `CatalogTable` with `CatalogBaseTable` in `TableEnvironment`.

bowenli86:
I saw that there are tests added for `CatalogTable`, just an FYI that we are unifying catalog table implementations in https://github.com/apache/flink/pull/8815 and maybe better to rebase this PR to #8815

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/TableEnvironment.java
```diff
@@ -114,6 +116,24 @@
...
+	 *   <li>`table`: Only the table name, with the current catalog name and database  name.</li>
+	 * </ol>
+	 * The registered tables then can be referenced in Sql queries.
+	 *
+	 * @param path           The path under wich the table will be registered
+	 * @param catalogTable   The table to register
+	 * @param ignoreIfExists If true, do nothing if there is already same table name under
+	 *                       the {@code path}. If false, a TableAlreadyExistException throws.
+	 */
+	void registerTable(String[] path, CatalogBaseTable catalogTable, boolean ignoreIfExists);
```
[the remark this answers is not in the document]
danny0405: Agree and removed, thanks

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/operations/ddl/CreateOperation.java
```diff
@@ -0,0 +1,34 @@
...
+import org.apache.flink.annotation.Internal;
+import org.apache.flink.table.operations.Operation;
+
+/**
+ * A {@link Operation} that describes the DDL statements, e.g. CREATE TABLE or CREATE FUNCTION.
+ *
+ * <p>Different sub operations can have their special instances. For example, a
+ * create table operation will have a {@link org.apache.flink.table.catalog.CatalogTable} instance,
+ * while a create function operation will have a
+ * {@link org.apache.flink.api.common.functions.Function} instance.
```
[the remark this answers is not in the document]
danny0405: Yep, thanks

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/operations/ddl/CreateTableOperation.java
```diff
@@ -0,0 +1,68 @@
...
+	public String[] getTablePath() {
+		return tablePath;
+	}
+
+	public boolean isIgnoreIfExists() {
+		return ignoreIfExists;
+	}
+
+	@Override
+	public String asSummaryString() {
```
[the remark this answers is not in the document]
danny0405: Added

flink-table/flink-table-common/src/main/java/org/apache/flink/table/api/ComputedColumn.java
```diff
@@ -0,0 +1,93 @@
...
+
+	public String getName() {
+		return name;
+	}
+
+	public String getExpression() {
+		return expression;
+	}
+
+	public String getProjectExpression() {
```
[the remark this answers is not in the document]
danny0405: removed

flink-table/flink-table-common/src/main/java/org/apache/flink/table/api/TableSchema.java
```diff
@@ -56,9 +57,28 @@
 
 	private final Map<String, Integer> fieldNameToIndex;
 
-	private TableSchema(String[] fieldNames, DataType[] fieldDataTypes) {
+	private final String[] primaryKeys;
+
+	private final String[][] uniqueKeys;
```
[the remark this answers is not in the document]
danny0405: Removed

flink-table/flink-table-planner/src/main/scala/org/apache/flink/table/planner/StreamPlanner.scala
```diff
@@ -100,22 +100,21 @@ class StreamPlanner(
 
     parsed match {
       case insert: SqlInsert =>
-        val targetColumnList = insert.getTargetColumnList
```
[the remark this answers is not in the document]
danny0405: Removed computed columns and add this code snippet back.

flink-table/flink-table-planner/pom.xml
```diff
@@ -88,26 +88,26 @@ under the License.
 
 		<dependency>
 			<groupId>org.apache.flink</groupId>
-			<artifactId>flink-table-api-java-bridge_${scala.binary.version}</artifactId>
+			<artifactId>flink-sql-parser</artifactId>
```
KurtYoung: Seems you don't change anything meaningful for this file?

flink-table/flink-table-common/src/main/java/org/apache/flink/table/api/TableSchema.java
```diff
@@ -304,6 +304,22 @@ public Builder() {
...
+		 */
+		public Builder proto(TableSchema other) {
+			Builder builder = instance();
+			builder.fieldNames.addAll(Arrays.asList(other.getFieldNames()));
+			builder.fieldDataTypes.addAll(Arrays.asList(other.getFieldDataTypes()));
+			return builder;
+		}
+
+		/** Returns a new builder instance. **/
+		public static Builder instance() {
```
KurtYoung: why don't you just new?

flink-table/flink-table-common/src/main/java/org/apache/flink/table/api/TableSchema.java
```diff
@@ -304,6 +304,22 @@ public Builder() {
 			fieldDataTypes = new ArrayList<>();
 		}
 
+		/** Create a proto builder from an existing schema.
+		 *
+		 * <p>Caution: This will invoke {@link #instance()}} to fetch a new builder first.
+		 */
+		public Builder proto(TableSchema other) {
```
KurtYoung: what's this used for?

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/operations/ddl/CreateOperation.java
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * A {@link Operation} that describes the DDL statements, e.g. CREATE TABLE or CREATE FUNCTION.
+ *
+ * <p>Different sub operations can have their special instances. For example, a
+ * create table operation will have a {@link org.apache.flink.table.catalog.CatalogTable} instance,
+ * while a create function operation will have a
+ * {@link org.apache.flink.table.catalog.CatalogFunction} instance.
+ */
+@Internal
+public interface CreateOperation extends Operation {
```
KurtYoung: Do you plan to share something in this interface? Otherwise it will be useless

flink-table/flink-table-planner/src/main/scala/org/apache/flink/table/api/internal/TableEnvImpl.scala
```diff
@@ -385,6 +389,67 @@ abstract class TableEnvImpl(
     planner.getCompletionHints(statement, position)
   }
 
+  override def sql(statement: String): Unit = {
```
godfreyhe: why the behavior of `sql` method in this class is different from the behavior of `sql` method in `TableEnvironmentImpl` ?

flink-table/flink-table-planner/src/main/java/org/apache/flink/table/sqlexec/SqlExecutableStatements.java
```diff
@@ -18,62 +18,197 @@
...
+ * {@link org.apache.flink.table.delegation.Planner}.
  */
 public class SqlExecutableStatements implements ReflectiveVisitor {
-	private TableEnvironment tableEnv;
+	private FlinkPlannerImpl flinkPlanner;
 
-	private final ReflectUtil.MethodDispatcher<Void> dispatcher =
-		ReflectUtil.createMethodDispatcher(Void.class,
+	private final ReflectUtil.MethodDispatcher<Operation> dispatcher =
+		ReflectUtil.createMethodDispatcher(Operation.class,
```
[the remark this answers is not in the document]
KurtYoung: You can have your own visitor which can return `Operation`

flink-table/flink-table-common/src/main/java/org/apache/flink/table/api/TableSchema.java
KurtYoung: But it's not used right now

flink-table/flink-table-planner/src/main/java/org/apache/flink/table/sqlexec/SqlExecutableStatements.java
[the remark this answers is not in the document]
wuchong: 1. I don't see a necessary to have a visitor here. Because we have know the specific SqlNode type (DDL, DML or Query) before invoke these methods. I think some static utility method will be more clear and enough here. For example: `QueryOperation convertToQueryOperation(..)`, `CreateOperation convertToCreateOperation(..)`, `ModifyOperation convertToModifyOperation(..)`
2. The method name and class name is not appropriate, because it doesn't **execute** the statement (e.g. register table to catalog, insert data to table), it only converts `SqlNode` to `Operation`s.

flink-table/flink-table-planner/src/test/java/org/apache/flink/table/sqlexec/SqlExecutableStatementsTest.java
```diff
@@ -0,0 +1,108 @@
...
+	private static final FlinkPlannerImpl planner = batchEnv.getFlinkPlanner();
+
+	@Test
+	public void testCreateTable() {
+		final String sql = "CREATE TABLE tbl1 (\n" +
+			"  a bigint,\n" +
+			"  b varchar, \n" +
+			"  c int, \n" +
+			"  d varchar,\n" +
+			"  PRIMARY KEY (a, b)\n" +
```
wuchong: I would suggest to throw syntax exception if we don't support primary key and computed column the this version. Otherwise, we may have compatibility problem if users use these syntax.

flink-table/flink-table-planner/src/test/java/org/apache/flink/table/sqlexec/SqlExecutableStatementsTest.java
```diff
@@ -0,0 +1,108 @@
...
+			new String[] {"a", "b", "c", "d"});
+		assertArrayEquals(catalogTable.getSchema().getFieldDataTypes(),
+			new DataType[]{
+				DataTypes.BIGINT(),
+				DataTypes.VARCHAR(Integer.MAX_VALUE),
+				DataTypes.INT(),
+				DataTypes.VARCHAR(Integer.MAX_VALUE)});
+	}
+
+	@Ignore // need to implement
```
wuchong: I would like to remove this test instead of ignore this.

flink-table/flink-table-planner/src/test/scala/org/apache/flink/table/catalog/CatalogTableITCase.scala
```diff
@@ -0,0 +1,450 @@
...
+    tableEnv.sql(sinkDDL)
+    tableEnv.sql(query)
+    execJob("testJob")
+    assertEquals(expected.sorted, TestCollectionTableFactory.RESULT.sorted)
+  }
+
+  @Test
+  def testInsertWithAggregateSource(): Unit = {
+    if (isStreaming) {
+      return
```
wuchong: We can create a upserting collection table sink to support this case.

flink-table/flink-table-planner/src/main/scala/org/apache/flink/table/api/internal/TableEnvImpl.scala
wuchong: I also have the same concern. It seems that it doesn't support multiple statements in this env. I think we should keep the same behavior.

flink-table/flink-table-planner/src/test/scala/org/apache/flink/table/catalog/CatalogTableITCase.scala
```diff
@@ -0,0 +1,450 @@
...
+        |)
+      """.stripMargin
+    val query =
+      """
+        |insert into t2
+        |select t1.a, t1.b, (t1.a + 1) as c from t1
+      """.stripMargin
+    tableEnv.sql(sourceDDL)
+    tableEnv.sql(sinkDDL)
+    tableEnv.sql(query)
```
wuchong: Can we add some tests that covers multiple statements in one sql with ';' delimiter ? 

Btw, IT cases are very heavy (the community is trying to reduce travis time). I think one for ETL and one for aggregation is enough. We can add some unit tests to verify the plan which is much more lightweight.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/TableEnvironment.java
```diff
@@ -289,6 +289,69 @@
...
+	 *                     ) with (
+	 *                       connector = 'kafka',
+	 *                       kafka.topic = 'xxx',
+	 *                       kafka.endpoint = 'x.x.x'
+	 *                     )";
+	 *
+	 *  String query = "INSERT INTO sinkTable SELECT * FROM sourceTable";
+	 *
+	 *  tEnv.sql(sourceDDL);
+	 *  tEnv.sql()sinkDDL);
```
KurtYoung: sql(sinkDDL)

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/TableEnvironment.java
```diff
@@ -355,9 +424,12 @@
 	 * }
 	 * </pre>
 	 *
+	 * @deprecated Use {@link #sql(String)}.
+	 *
 	 * @param stmt The SQL statement to evaluate.
 	 * @param config The {@link QueryConfig} to use.
 	 */
+	@Deprecated
```
KurtYoung: Looks like this functionality can't be fully replaced with `sql()` because of the lack of `QueryConfig`

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/internal/TableEnvironmentImpl.java
```diff
@@ -337,17 +372,35 @@ public TableConfig getConfig() {
...
+	 *
+	 * @param path           The path under which the table will be registered
+	 * @param catalogTable   The table to register
+	 * @param ignoreIfExists If true, do nothing if there is already same table name under
+	 *                       the {@code path}. If false, a TableAlreadyExistException throws.
+	 */
+	private void registerTableInternal(String[] path,
+			CatalogBaseTable catalogTable,
+			boolean ignoreIfExists) {
+		String[] fullName = catalogManager.paddedTableName(Arrays.asList(path));
```
KurtYoung: Looks like you are trying to pad table names in different places, one is here, another one is `getDefaultTablePath`, could we make it unified? It seems unnecessary to call `getDefaultTablePath` every time before calling this function.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/internal/TableEnvironmentImpl.java
```diff
@@ -245,6 +249,37 @@ public String explain(Table table) {
...
 	}
 
+	@Override
+	public Table sql(String statement) {
+		List<Operation> operations = planner.parse(statement);
+		Preconditions.checkState(operations.size() <= 1,
+			"sql() only accepts a single SQL statement a time.");
+		Operation operation = operations.get(0);
+		if (operation instanceof CreateTableOperation) {
+			CreateTableOperation operation1 = (CreateTableOperation) operation;
```
KurtYoung: change variable name

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/internal/TableEnvironmentImpl.java
```diff
@@ -445,4 +500,36 @@ protected TableImpl createTable(QueryOperation tableOperation) {
...
+			public List<QueryOperation> getChildren() {
+				return new ArrayList<>();
+			}
+
+			@Override
+			public String asSummaryString() {
+				return "Query result operation: " + result;
+			}
+		};
+		return TableImpl.createTable(
```
KurtYoung: I'm not sure this is a valid way to create a workable table...

flink-table/flink-table-planner/src/main/java/org/apache/flink/table/sqlexec/SqlExecutableStatements.java
```diff
@@ -18,62 +18,197 @@
...
-	public void execute(SqlCreateTable sqlCreateTable) {
-		// need to implement.
+	public Operation execute(SqlCreateTable sqlCreateTable) {
+		// set with properties
+		SqlNodeList propertyList = sqlCreateTable.getPropertyList();
+		Map<String, String> properties = new HashMap<>();
+		if (propertyList != null) {
+			propertyList.getList().forEach(p ->
+				properties.put(((SqlProperty) p).getKeyString().toLowerCase(),
+					((SqlProperty) p).getValueString()));
```
KurtYoung: why toLowerCase for key but not for value?

flink-table/flink-table-planner/src/main/java/org/apache/flink/table/sqlexec/SqlExecutableStatements.java
```diff
@@ -18,62 +18,197 @@
...
+		}
+		String tableType = properties.get("connector");
```
KurtYoung: Is this a necessary check? Not sure if here is the right place to valid the property

flink-table/flink-table-planner/src/main/java/org/apache/flink/table/sqlexec/SqlExecutableStatements.java
```diff
@@ -18,62 +18,197 @@
...
+		if (StringUtils.isNullOrWhitespaceOnly(tableType)) {
+			throw new SqlExecutionException("Option [connector] is required for"
+				+ " CREATE TABLE statement.");
+		}
+
+		TableSchema tableSchema = createTableSchema(sqlCreateTable,
+			new FlinkTypeFactory(new FlinkTypeSystem())); // need to make type factory singleton ?
```
KurtYoung: I don't think you can use `FlinkTypeFactory` anymore, once the tableEnv is brought up to api module. Use another way to replace it?

flink-table/flink-table-planner/src/main/java/org/apache/flink/table/sqlexec/SqlExecutableStatements.java
```diff
@@ -18,62 +18,197 @@
...
+				.getList()
+				.stream()
+				.map(p -> ((SqlIdentifier) p).getSimple())
+				.collect(Collectors.toList());
+		}
+		CatalogTable catalogTable = new CatalogTableImpl(tableSchema,
+			partitionKeys,
+			properties,
+			tableComment);
+		return new CreateTableOperation(sqlCreateTable.fullTableName(), catalogTable, false);
```
KurtYoung: `SqlCreate` from calcite has such information for the boolean flag `ignoreIfExists`

flink-table/flink-table-planner/src/test/java/org/apache/flink/table/sqlexec/SqlExecutableStatementsTest.java
danny0405: I agree, add check and tests.

flink-table/flink-table-planner/src/test/scala/org/apache/flink/table/catalog/CatalogTableITCase.scala
danny0405: We do not multiple statements now, i think we do not need plan test now, because this change only affect table source and sink but not the plan, an ITCase is enough, we need an end-to-end test for catalog table now.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/TableEnvironment.java
```diff
@@ -320,6 +320,69 @@ static TableEnvironment create(EnvironmentSettings settings) {
...
+	 *     c varchar
+	 *   ) with (
+	 *     connector = 'csv',
+	 *     csv.path = 'xxx'
+	 *   )
+	 * </pre></blockquote>
+	 *
+	 * <p>The returns table format for different kind of statement:
+	 * <ul>
+	 *   <li>DDL: a table with one column of VARCHAR type to describe if this operation
```
KurtYoung: Please update this comment

flink-table/flink-table-planner/src/main/java/org/apache/flink/table/catalog/CatalogCalciteSchema.java
```diff
@@ -43,10 +43,12 @@
 
 	private final String catalogName;
 	private final Catalog catalog;
+	private final boolean isBatch;
 
-	public CatalogCalciteSchema(String catalogName, Catalog catalog) {
+	public CatalogCalciteSchema(boolean isBatch, String catalogName, Catalog catalog) {
```
KurtYoung: align the order?

flink-table/flink-table-planner/src/main/scala/org/apache/flink/table/catalog/CalciteCatalogTable.scala
```diff
@@ -0,0 +1,156 @@
...
+
+  override def copy(): CatalogBaseTable = new CalciteCatalogTable(
+      logicalSchema.copy(),
+      new JArrayList[String](getPartitionKeys),
+      new JHashMap[String, String](getProperties),
+      getComment)
+
+  override def getDescription: Optional[String] = Optional.of(getComment)
+
+  override def getDetailedDescription: Optional[String] = Optional.of("This is a" +
```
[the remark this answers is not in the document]
danny0405: Removed

flink-table/flink-table-common/src/main/java/org/apache/flink/table/api/TableSchema.java
```diff
@@ -56,9 +57,28 @@
 
 	private final Map<String, Integer> fieldNameToIndex;
 
-	private TableSchema(String[] fieldNames, DataType[] fieldDataTypes) {
+	private final String[] primaryKeys;
+
+	private final String[][] uniqueKeys;
+
+	private final ComputedColumn[] computedColumns;
```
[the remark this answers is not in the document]
danny0405: Removed

flink-table/flink-table-planner/pom.xml
danny0405: reverted

flink-table/flink-table-planner/src/test/scala/org/apache/flink/table/catalog/CatalogTableITCase.scala
danny0405: I think only test append sink is enough.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/TableEnvironment.java
danny0405: Fixed

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/internal/TableEnvironmentImpl.java
danny0405: Changed

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/internal/TableEnvironmentImpl.java
danny0405: Returns null for insert and create table now

flink-table/flink-table-planner/src/main/java/org/apache/flink/table/sqlexec/SqlExecutableStatements.java
danny0405: Removed `toLowerCase`

flink-table/flink-table-planner/src/main/java/org/apache/flink/table/sqlexec/SqlExecutableStatements.java
danny0405: `connector` is a required property for DDL now.

flink-table/flink-table-planner/src/main/java/org/apache/flink/table/sqlexec/SqlExecutableStatements.java
danny0405: We still can use `FlinkTypeFactory` because this code snippet is not shared.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/TableEnvironment.java
danny0405: Update what ?

flink-table/flink-table-planner/src/main/java/org/apache/flink/table/catalog/CatalogCalciteSchema.java
danny0405: Sorry i didn't get what you mean.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/internal/TableEnvironmentImpl.java
danny0405: `getDefaultTablePath` is just a tool method

flink-table/flink-table-planner/src/main/java/org/apache/flink/table/sqlexec/SqlExecutableStatements.java
danny0405: We actually hard code it to be false when create the `SqlCreateTable` instance, are we planning to support this now ?

flink-table/flink-table-planner/src/main/java/org/apache/flink/table/sqlexec/SqlToOperationConverter.java
```diff
@@ -0,0 +1,198 @@
...
+			return converter.convert(validated);
+		}
+	}
+
+	/**
+	 * Convert the {@link SqlCreateTable} node.
+	 */
+	private Operation convert(SqlCreateTable sqlCreateTable) {
+		// primary key and unique keys are not supported
+		if (sqlCreateTable.getPrimaryKeyList() != null
```
KurtYoung: kuohao

flink-table/flink-table-planner/src/main/scala/org/apache/flink/table/calcite/FlinkPlannerImpl.scala
```diff
@@ -18,6 +18,8 @@
 
 package org.apache.flink.table.calcite
 
+import org.apache.flink.sql.parser.ExtendedSqlNode
```
KurtYoung: import order

flink-table/flink-table-planner/src/main/scala/org/apache/flink/table/calcite/FlinkPlannerImpl.scala
```diff
@@ -98,13 +102,23 @@ class FlinkPlannerImpl(
 
   def validate(sqlNode: SqlNode): SqlNode = {
     val catalogReader = catalogReaderSupplier.apply(false)
+    // do pre-validate rewrite.
+    sqlNode.accept(new PreValidateReWriter(catalogReader, typeFactory))
+    sqlNode match {
+      case node: ExtendedSqlNode =>
+        node.validate()
+        if (sqlNode.getKind.belongsTo(SqlKind.DDL)) {
```
KurtYoung: move this out of pattern match

flink-python/pyflink/table/table_environment.py
```diff
@@ -235,6 +235,69 @@ def explain(self, table):
         """
         return self._j_tenv.explain(table._j_table)
 
+    def sql(self, query):
+        """
```
KurtYoung: please align the comment with java

flink-python/pyflink/table/table_environment.py
```diff
@@ -235,6 +235,69 @@ def explain(self, table):
...
+        tEnv.sql()sinkDDL)
+        tEnv.sql(query)
+        tEnv.execute("MyJob")
+
+        This code snippet creates a job to read data from Kafka source into a CSV sink.
+
+        :param query: The SQL statement to evaluate.
+        """
+        j_table = self._j_tenv.sql(query)
+        return Table(j_table)
```
KurtYoung: Will this cause any problem if `self._j_tenv.sql(query)` returns NULL?

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/internal/TableEnvironmentImpl.java
```diff
@@ -290,6 +292,40 @@ public String explain(boolean extended) {
 		return planner.getCompletionHints(statement, position);
 	}
 
+	@Override
+	public Table sql(String statement) {
+		List<Operation> operations = planner.parse(statement);
+		Preconditions.checkState(operations.size() <= 1,
```
KurtYoung: change to operations.size() == 1? Otherwise it will be unsafe for your next line: `Operation operation = operations.get(0);`

fhueske:
Hi,

I noticed that this PR introduces significant changes to the public API of `TableEnvironment`, i.e., deprecating `sqlQuery()` and `sqlUpdate()` and introducing `sql()`. AFAIK, there was no consensus to remove `sqlQuery()` and `sqlUpdate()`.

Originally we had only `sql()` and moved to `sqlQuery()` and `sqlUpdate()` when introducing support for `INSERT INTO` queries and table sinks. I remember these interfaces were the result and consensus of a long discussion. See #3829 and this shorter thread on the [dev ML](https://lists.apache.org/thread.html/844a9b8a009706de801b0d9746f9a193fce4a197cc11632a4f4fd39d@%3Cdev.flink.apache.org%3E). 

Since many users are relying on these interfaces, we should have a discussion on the dev mailing list before deprecating them. Oftentimes, public interfaces are the result of community discussions which become somewhat pointless when the APIs are changed afterwards.

I'm not saying that these are bad changes. Maybe the requirements for the APIs changed, but this should definitely be discussed with the community. Also, I'm not a big fan of rushing changes to core APIs into the master so close a to a feature freeze.

Best, Fabian

twalthr:
@KurtYoung I completely share Fabian's opinion here. The PR is marked as `[table-planner]` but this contribution changes different modules and should have been marked `[table]`. The API breaking changes should be added to a separate commit and more people should have been pinged about this change.

Regarding the change of `sql()`, returning null is always bad API design and leads to potential `NullPointerException`s. I would like to reintroduce the old interfaces. Both can share the same code but validate if the last statement (in a multiline query) is a `SELECT` query or not.

KurtYoung:
Hi @fhueske and @twalthr , sorry we didn't do detail investigation about these APIs. And also changes like this should definitely have a broader and visible discussions, like in mailing list. 

For this particular API, it was intend to ease downstream users to the usage of current API. So they don't have to judge the sql statement's type to decide whether it's a query or update. 

It was planned and we just don't have enough resource to finish it earlier, sorry to cause the impression that we are rushing changes. But consider this API will having some concerns from you guys, I will discuss with @danny0405 if all functionality can be finished without this API change.

Sorry for the inconvenience, and discussions like this should definitely happen before opening PR next time.

Regarding to returning NULL of `sql()`, we discussed another approach is return a `Optional<Table>`, what do you think about this? @twalthr

KurtYoung:
Sorry, just realized ddl is based on this interface. How about we deleting the deprecated flag for the old APIs and only introducing a new API here. We still need to reach a consensus to deal with returning NULL.

wuchong:
How about supporting DDL & DML both in `sqlUpdate`? This can avoid to introduce a new `sql()` interface right now. 

JDBC has a similar design about this, the `java.sql.Statement` interface has [`ResultSet executeQuery(String sql)`](https://docs.oracle.com/javase/8/docs/api/java/sql/Statement.html#executeQuery-java.lang.String-) method to execute only SELECT statement and return a query result (aka. Table), and also has a [`int executeUpdate(String sql)`](https://docs.oracle.com/javase/8/docs/api/java/sql/Statement.html#executeUpdate-java.lang.String-) method to execute DMLs and DDLs and returns effected rows. 

So I think it makes sense to support DDL in `sqlUpdate` because it is also an update statement.

KurtYoung:
> How about supporting DDL & DML both in `sqlUpdate`? This can avoid to introduce a new `sql()` interface right now.
> 
> JDBC has a similar design about this, the `java.sql.Statement` interface has [`ResultSet executeQuery(String sql)`](https://docs.oracle.com/javase/8/docs/api/java/sql/Statement.html#executeQuery-java.lang.String-) method to execute only SELECT statement and return a query result (aka. Table), and also has a [`int executeUpdate(String sql)`](https://docs.oracle.com/javase/8/docs/api/java/sql/Statement.html#executeUpdate-java.lang.String-) method to execute DMLs and DDLs and returns effected rows.
> 
> So I think it makes sense to support DDL in `sqlUpdate` because it is also an update statement.

Sounds like a reasonable solution

twalthr:
Yes, this was also my intention in my previous comment. `sqlUpdate()` can run multiple statements (DDL and DML) and ensure that no `SELECT` statement is part of the SQL (that would be lost otherwise). `sqlQuery()` can also run multiple statements (DDL and DML) but the last statement must be a `SELECT` such that the method has a return `Table`.

fhueske:
Thanks for the replies and good discussion @KurtYoung, @wuchong, and @twalthr!
I think @wuchong's proposal ("So I think it makes sense to support DDL in sqlUpdate because it is also an update statement.") would be a good idea for this release. But let's also see what @danny0405 thinks. If he agrees, let's revert some of the changes to undeprecate `sqlQuery()` and `sqlUpdate()` and move the DDL logic from `sql()` to `sqlUpdate()`.

We can of course discuss whether it is necessary to evolve the interface and if how it can be improved. However, I'd suggest to have this discussion early on in the development phase of 1.10 to give other users a chance to participate.

Thanks, Fabian

danny0405:
Thanks @fhueske for your nice review comments, i also think we should be caution for adding a pretty new interface and also deprecate the old one(which are most commonly used).

I definitely agree with your point that we should fire a fresh new discussion early on in the development phase of 1.10 about the relationship of `sql` `sqlUpdate` and `sqlQuery`, especially when we introduce other DDLs grammar like `Alter table` or `Show tables` etc.

Broadly i agree the idea to put `CREATE TABLE` ddl in `sqlUpdate`, also i have fired another JIRA issue: https://issues.apache.org/jira/browse/FLINK-13209

fhueske:
Great, thank you @danny0405!

flink-table/flink-table-planner/src/main/java/org/apache/flink/table/sqlexec/SqlToOperationConverter.java
```diff
@@ -0,0 +1,193 @@
...
+		TableSchema.Builder builder = new TableSchema.Builder();
+		// collect the physical table schema first.
+		final List<SqlNode> physicalColumns = columnList.getList().stream()
+			.filter(n -> n instanceof SqlTableColumn).collect(Collectors.toList());
+		for (SqlNode node : physicalColumns) {
+			SqlTableColumn column = (SqlTableColumn) node;
+			final RelDataType relType = column.getType().deriveType(factory,
+				column.getType().getNullable());
+			builder.field(column.getName().getSimple(),
+				TypeConversions.fromLegacyInfoToDataType(FlinkTypeFactory.toTypeInfo(relType)));
```
KarlManong: 这里会导致varchar的precision丢失。和blink的表现不一致。是bug吗？
@danny0405

danny0405: This is not a bug, it's the Flink table planner that does not support type coercion very well, you can take SqlToOperationConverterTest as a reference.

PR #8846, opened 2019-06-24
[FLINK-12766][runtime] Dynamically allocate TaskExecutor's managed memory to slots.

xintongsong:

This pull request is base on #8704 and #8740. It dynamically allocates the TaskExecutor's resources to the slots on it. FTM, managed memory in slot resource profiles is set to the actual size, the rest are set to `ANY` that matches any request (see #8704).

## Brief change log

- 0d3245b7462d157895257750988c46416feb4bb0: Introduce `merge` and `subtract` calculations in `ResourceProfile`, and corresponding tests. These calculations are needed for bookkeeping task executors' resources.
- 4b822d6f68eae559c07220949da894c7811762c7 - 11262902755a455ca7ff0926ca7e8afd323ee69d: Changes on the task executor side.
  - 4b822d6f68eae559c07220949da894c7811762c7: Add argument `allocationResourceProfile` in `TaskExecutorGateway#requestSlot` which describes the dynamically allocated slot resources. Task executors need to record the allocation resource profiles so they can:
    1. offer slots with requested resources to job masters, and
    2. report allocated resources to the resource manager so it can recover the task executor resource bookkeepings after a failover.
  - 1b09722d03d64025e4e184845cfc0d43218fa285: `TaskExecutor` checks the requested allocation resource profiles and rejects those that exceed its current available resources. This could happen due to asynchronous information on RM/TM. 
    - E.g., RM requests slot1 from the TE, then receives an outdated slot report from the TE showing that slot1 is free. After that RM requests slot2 from the TE. At this moment, RM1 believes that the TE has enough available for slot2 because it thinks slot1 is free, but the TE may not have enough resources because slot1 is actually allocated.
  - e90d462c530814bee19efe0203a895d78ad1d2f9: `TaskExecutor` offers slot with allocation resource profile.
  - 231d5691bddbcf5c207d3edcf6d86fc00d52b5ac: `TaskExecutor` report slot status with allocation resource profile.
  - 11262902755a455ca7ff0926ca7e8afd323ee69d: Add `TaskExecutorTest#testAllocationResourceProfile` that validates `TaskExecutor` maintains allocation resource profiles of slots properly.
- 29d44b40f5c245ad606e06cddcd71ea3e8ff3b7b - e2ee9de9847b219a3ad1f07d7a023f4dc1746e1b: Changes on the resource manager side.
  - 29d44b40f5c245ad606e06cddcd71ea3e8ff3b7b: `SlotManager` groups `PendingTaskManagerSlots` according to whether they should be on the same `TaskExecutor`, and matches registered slots accordingly. We need this change to bookkeep available resources for pending task executors in subsequent commit.
  - 4e05a878f37ceba81e65a3ae6c3d9e28f5251dfa: Update `SlotManagerTest#testPendingTaskManagerSlotCompletion`. The original assumption that pending task slots are completed independently from each other no longer holds.
  - 85d58f2641fdbae4d1d349010552fe11b8d723bf: Introduce the data structure `TaskExecutorResourceBookkeeper` for bookkeeping total/available/allocated resources of registered and pending task executors.
  - 35591c5efe5f52c1cfa111a959a744a06c4ce9e9: `SlotManager` bookkeeps task executors' resources.
  - 9b6cac6d719e04d07a169930635e36f8661a8674: With dynamic slot resource allocation, resource managers on Yarn/Mesos should be able to start new worker for resource requests that is larger than the default resources of one slot but not larger than the total resources of slots per task executor.
  - e2ee9de9847b219a3ad1f07d7a023f4dc1746e1b: Add `SlotManagerTest#testBookkeepingTaskExecutorResources` that validates `SlotManager` bookkeeps resources of task executors properly.

## Verifying this change
  - Add `TaskExecutorTest#testAllocationResourceProfile` that validates `TaskExecutor` maintains allocation resource profiles of slots properly.
  - Add `SlotManagerTest#testBookkeepingTaskExecutorResources` that validates `SlotManager` bookkeeps resources of task executors properly.

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-core/src/main/java/org/apache/flink/api/common/resources/Resource.java
+++ b/flink-core/src/main/java/org/apache/flink/api/common/resources/Resource.java
@@ -79,6 +79,26 @@ public abstract class Resource implements Serializable {
 		return create(aggregatedValue, resourceAggregateType);
 	}
 
+	public Resource minus(Resource other) {
+		Preconditions.checkArgument(getClass() == other.getClass(), "Minus with different resource resourceAggregateType");
+		Preconditions.checkArgument(this.name.equals(other.name), "Minus with different resource name");
+		Preconditions.checkArgument(this.resourceAggregateType == other.resourceAggregateType, "Minus with different aggregate resourceAggregateType");
+
+		final double aggregatedValue;
+		switch (resourceAggregateType) {
+			case AGGREGATE_TYPE_MAX :
+				// TODO: For max, should check if the latest max item is removed and change accordingly.
+				aggregatedValue = this.value;
+				break;
+
+			case AGGREGATE_TYPE_SUM:
+			default:
+				aggregatedValue = this.value - other.value;
+		}
+
+		return create(aggregatedValue, resourceAggregateType);
+	}
+
 	@Override
 	public boolean equals(Object o) {
 		if (this == o) {
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/clusterframework/types/ResourceProfile.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/clusterframework/types/ResourceProfile.java
@@ -20,6 +20,7 @@ package org.apache.flink.runtime.clusterframework.types;
 
 import org.apache.flink.api.common.operators.ResourceSpec;
 import org.apache.flink.api.common.resources.Resource;
+import org.apache.flink.util.Preconditions;
 
 import javax.annotation.Nonnull;
 
@@ -29,6 +30,8 @@ import java.util.HashMap;
 import java.util.Iterator;
 import java.util.Map;
 import java.util.Objects;
+
+import static org.apache.flink.util.Preconditions.checkArgument;
 
 /**
  * Describe the immutable resource profile of the slot, either when requiring or offering it. The profile can be
@@ -48,7 +51,7 @@ public class ResourceProfile implements Serializable, Comparable<ResourceProfile
 
 	private static final long serialVersionUID = 1L;
 
-	public static final ResourceProfile UNKNOWN = new ResourceProfile(-1.0, -1);
+	public static final ResourceProfile UNKNOWN = new ResourceProfile();
 
 	/** ResourceProfile which matches any other ResourceProfile. */
 	public static final ResourceProfile ANY = new ResourceProfile(Double.MAX_VALUE, Integer.MAX_VALUE, Integer.MAX_VALUE, Integer.MAX_VALUE, Integer.MAX_VALUE, Integer.MAX_VALUE, Collections.emptyMap());
@@ -97,6 +100,12 @@ public class ResourceProfile implements Serializable, Comparable<ResourceProfile
 			int networkMemoryInMB,
 			int managedMemoryInMB,
 			Map<String, Resource> extendedResources) {
+		Preconditions.checkArgument(cpuCores >= 0);
+		Preconditions.checkArgument(heapMemoryInMB >= 0);
+		Preconditions.checkArgument(directMemoryInMB >= 0);
+		Preconditions.checkArgument(nativeMemoryInMB >= 0);
+		Preconditions.checkArgument(networkMemoryInMB >= 0);
+		Preconditions.checkArgument(managedMemoryInMB >= 0);
 		this.cpuCores = cpuCores;
 		this.heapMemoryInMB = heapMemoryInMB;
 		this.directMemoryInMB = directMemoryInMB;
@@ -116,6 +125,18 @@ public class ResourceProfile implements Serializable, Comparable<ResourceProfile
 	 */
 	public ResourceProfile(double cpuCores, int heapMemoryInMB) {
 		this(cpuCores, heapMemoryInMB, 0, 0, 0, 0, Collections.emptyMap());
+	}
+
+	/**
+	 * Creates a special ResourceProfile with negative values, indicating resources are unspecified.
+	 */
+	private ResourceProfile() {
+		this.cpuCores = -1.0;
+		this.heapMemoryInMB = -1;
+		this.directMemoryInMB = -1;
+		this.nativeMemoryInMB = -1;
+		this.networkMemoryInMB = -1;
+		this.managedMemoryInMB = -1;
 	}
 
 	/**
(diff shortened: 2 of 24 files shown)


flink-runtime/src/main/java/org/apache/flink/runtime/resourcemanager/slotmanager/SlotManager.java
```diff
@@ -128,6 +130,20 @@
 	/** Resource profiles of slots that can be acquired. */
 	private final HashSet<ResourceProfile> availableSlotResourceProfiles = new HashSet<>();
 
+	/**
+	 * All allocated slots's allocation resource profile.
```
gaoyunhaii: slots's -> slots'

flink-core/src/main/java/org/apache/flink/api/common/resources/Resource.java
```diff
@@ -79,6 +79,26 @@ public Resource merge(Resource other) {
 		return create(aggregatedValue, resourceAggregateType);
 	}
 
+	public Resource minus(Resource other) {
```
gaoyunhaii: The minus could be merged with the same methods in other diff.

flink-runtime/src/main/java/org/apache/flink/runtime/clusterframework/types/ResourceProfile.java
```diff
@@ -334,6 +334,72 @@ public boolean approximate(ResourceProfile that) {
 		return Objects.equals(this.extendedResources, that.extendedResources);
 	}
 
+	public ResourceProfile merge(ResourceProfile that) {
```
gaoyunhaii: The `merge` method could be merged with the same methods in other diff.

flink-runtime/src/main/java/org/apache/flink/runtime/resourcemanager/slotmanager/SlotManager.java
```diff
@@ -128,6 +130,20 @@
...
+	 * If the task manager has no allocated slots(all slots are free), it does not exist in this map.
+	 * ResourceID : The id of each task manager
+	 * Tuple2 : The allocated slots and allocation resources of the task manager
+	 *    f0 : The allocated slots map, keyed by slotId and value is allocation resource profile of the slot
+	 *    f1 : The remaining resource of the task manager identified by resourceId
+	 */
+	private final Map<ResourceID, Tuple2<Map<SlotID, ResourceProfile>, ResourceProfile>>
```
gaoyunhaii: I think it may be better to have a distinct class to manage the total resource, the free resource, the resource of the allocated slot. Then we could merge these two map with a simpler map.

flink-runtime/src/main/java/org/apache/flink/runtime/resourcemanager/slotmanager/SlotManager.java
```diff
@@ -363,9 +382,12 @@ public void registerTaskManager(final TaskExecutorConnection taskExecutorConnect
 			// first register the TaskManager
 			ArrayList<SlotID> reportedSlots = new ArrayList<>();
 
+			ResourceProfile taskExecutorResource = new ResourceProfile(0.0, 0);
```
gaoyunhaii: Can be replaced with ResourceProfile.EMPTY after merging code.

flink-runtime/src/main/java/org/apache/flink/runtime/resourcemanager/slotmanager/SlotManager.java
```diff
@@ -949,11 +982,73 @@ private void removeSlot(SlotID slotId) {
...
 					new FlinkException("The assigned slot " + slot.getSlotId() + " was removed."));
 			}
+
+			removeSlotFromAllocatedResources(slotId);
 		} else {
 			LOG.debug("There was no slot registered with slot id {}.", slotId);
 		}
 	}
 
+	private void recordAllocatedSlotAndResource(TaskManagerSlot slot, ResourceProfile resourceProfile) {
```
gaoyunhaii: I think this method might only be called when
1. in `allocateSlot`.
2. in `registerSlot` and found that the allocationId is not null.

flink-runtime/src/main/java/org/apache/flink/runtime/resourcemanager/slotmanager/SlotManager.java
```diff
@@ -949,11 +982,73 @@ private void removeSlot(SlotID slotId) {
...
+
+		ResourceID resourceID = targetSlot.getSlotId().getResourceID();
+		Tuple2<Map<SlotID, ResourceProfile>, ResourceProfile> allocatedResources =
+			taskExecutorAllocationSlotsResource.get(resourceID);
+		ResourceProfile remain = (allocatedResources == null) ?
+			taskExecutorResources.get(resourceID) : allocatedResources.f1;
+
+		// for now, we only do dynamic resource allocation for managed memory
+		return new ResourceProfile(
+			targetSlotResource.getCpuCores(),
```
gaoyunhaii: I think here might be `requiredSlotResource` instead of `targetSlotResource`, since the `targetSlotResource` may be less than the required one but there are still remaining resources.

flink-runtime/src/main/java/org/apache/flink/runtime/taskexecutor/SlotStatus.java
```diff
@@ -88,6 +93,15 @@ public AllocationID getAllocationID() {
 		return allocationID;
 	}
 
+	/**
+	 * Get allocated resource in this slot.
```
gaoyunhaii: Gets

flink-runtime/src/test/java/org/apache/flink/runtime/taskexecutor/TaskExecutorTest.java
```diff
@@ -1897,6 +1906,128 @@ public void testSlotReportResourceProfile() throws Throwable {
 		assertEquals(numTaskSlots, slotNum);
 	}
 
+	@Test
+	public void testAllocationResourceProfile() throws Throwable {
+		final TaskSlotTable taskSlotTable = new TaskSlotTable(Arrays.asList(ResourceProfile.ANY, ResourceProfile.ANY), timerService);
```
gaoyunhaii: This line is too long and might be split

xintongsong:
Hi @StephanEwen , would you please take a look at this PR at your convenience? Thank you.

xintongsong:
Rebased to master after #8704 is merged.

StephanEwen:
I am currently going through this PR.

About the changes to the `ResourceProfile`:
  - The code could use some tests
  - The code seems to assume that all values are non-negative, but the constructor does not actually enforce this. What do you think about adding checks to the constructor and using a private constructor for the UNKNOWN profile (which needs negative values)?

xintongsong:
Rebased to master after the previous PR #8740 is merged, and addressed comments regarding `ResourceProfile`.

StephanEwen:
The code is overall in decent shape.
I am a bit unsure, though, about this approach.

For example all the changes on the TaskManager result in something quite unintuitive to me.
  - there are slots, they have a resource profile
  - then an allocated resource profile, which is possibly a subset of that.
  - some operations (for example to identify idempotent requests) now take the new resource profile into account. I cannot even say whether that is correct or not - probably not breaking anything immediately, but breaks with the design of how requests are identified and can cause tricky bugs later.
  - the TM may now request requests for free slots based on its bookkeeping. While this should typically not happen when the RM bookkeeping is in sync, what happens when the views diverge, due to a request timeout? The slot allocation protocol is so sensitive, do we really want to add this additional complexity?

I assume that the reason for all the TM changes is a combination of 
  - when the JM allocates a slot, it actually gets that slot offered from the TM directly, and not from the RM. So the profile somehow needs to go through the TM to end up in the Slot on the JM
  - some safety net bookkeeping

This solution is a temporary solution, an in the long run we probably want to either have fully dynamically sized slots, allocated and sized on demand, or keep the slots are static and change the operators in the APIs to adjust their resource consumption to what is there in the slots.

Having a temporary solution, of this big complexity, touching something so sensitive as the slot allocation, and for a new SQL runtime that is in its first release and might change quite a bit in the next release... I simply do not feel comfortable to merge this. The risk of bugs and regression, and the added complexity seems so high compared to the benefit.

I would be happy to double down on a FLIP about how resource management should look like in the next versions and make sure we have something really well working in the following release.

KurtYoung:
We can try dividing managed memory evenly and make it belong to slot. Enabling slot sharing can ease the memory waste. I think this is also a reasonable starting point.

xintongsong:
Thank you for the comments, @StephanEwen, @KurtYoung.

This truly is an imperfect temporal solution. If you guys both think we should not have this for 1.9, I respect your opinions. Will not insist on this one.

I think we still need to work on #8841, to make sure that slot sharing works properly with fine grained resource profiles, if we decide that batch jobs need slot sharing to ease the memory waste.

PR #8850, opened 2019-06-24
[FLINK-12954] Supports create(drop) view grammar for sql parser

danny0405:

- Add create view grammar like `CREATE VIEW view_name(col1, col2) AS select col3, col4 from tbl`

## Brief change log

  - Add create/drop view

## Verifying this change
See tests in FlinkSqlParserimplTest.

  - The serializers: no

## Documentation

  - If yes, how is the feature documented?  not documented

### DIFF
--- a/flink-table/flink-sql-parser/src/main/codegen/includes/parserImpls.ftl
+++ b/flink-table/flink-sql-parser/src/main/codegen/includes/parserImpls.ftl
@@ -122,6 +122,31 @@ SqlNode PropertyValue() :
     }
 }
 
+/** Parse a table properties. */
+SqlNodeList TableProperties():
+{
+    SqlNode property;
+    final List<SqlNode> proList = new ArrayList<SqlNode>();
+    final Span span;
+}
+{
+    <LPAREN> { span = span(); }
+    [
+        property = PropertyValue()
+        {
+            proList.add(property);
+        }
+        (
+            <COMMA> property = PropertyValue()
+            {
+                proList.add(property);
+            }
+        )*
+    ]
+    <RPAREN>
+    {  return new SqlNodeList(proList, span.end(this)); }
+}
+
 SqlCreate SqlCreateTable(Span s, boolean replace) :
 {
     final SqlParserPos startPos = s.pos();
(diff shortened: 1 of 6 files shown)


flink-table/flink-sql-parser/src/main/java/org/apache/flink/sql/parser/ddl/SqlCreateView.java
```diff
@@ -0,0 +1,122 @@
...
+		this.fieldList = fieldList;
+		this.query = query;
+		this.comment = comment;
+	}
+
+	@Override
+	public List<SqlNode> getOperandList() {
+		List<SqlNode> ops = Lists.newArrayList();
+		ops.add(viewName);
+		ops.add(fieldList);
```
wuchong: Do we want to support field list in the first version? If not, I would suggest to remove it to keep compatible.

flink-table/flink-sql-parser/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -334,6 +317,50 @@ void PartitionSpecCommaList(SqlNodeList list) :
...
+*/
+SqlCreate SqlCreateView(Span s, boolean replace) : {
+    boolean replaceView = false;
+    SqlIdentifier viewName;
+    SqlCharStringLiteral comment = null;
+    SqlNode query;
+    SqlNodeList fieldList = SqlNodeList.EMPTY;
+}
+{
+    [ <OR> <REPLACE> { replaceView = true; } ]
```
wuchong: We can remove this, the parameter `replace` already do this.
`SqlCreateTable` is the same.

danny0405: Yep, thanks.

flink-table/flink-sql-parser/src/main/java/org/apache/flink/sql/parser/ddl/SqlCreateView.java
danny0405: Yes, we want to support this. So use can rename the field.

flink-table/flink-sql-parser/src/test/java/org/apache/flink/sql/parser/FlinkSqlParserImplTest.java
```diff
@@ -583,6 +583,25 @@ public void testInvalidUpsertOverwrite() {
 			"OVERWRITE expression is only used with INSERT mode");
 	}
 
+	@Test
+	public void testCreateViewWithProperty() {
```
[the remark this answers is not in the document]
danny0405: done

wuchong:
LGTM
+1 to merge

PR #8851, opened 2019-06-24
[FLINK-12876] [runtime] Add an adapter of region failover NG for legacy scheduler

zhuzhurk:

*This PR adds an adapter to adapt flip1.RestartPipelinedRegionStrategy for legacy scheduler, so that the legacy scheduler can support fine grained recovery.*

## Brief change log

  - *An adapter of flip1.RestartPipelinedRegionStrategy. It implements interfaces of the old FailoverStrategy to support the legacy scheduler.*
  - *A vertex versioner from Gary to maintain vertex versions. This helps to deal with concurrent region failovers.*
 - *A scheduling util that implements similar scheduling logic to ExecutionGraph for failover purposes.*

## Verifying this change

  - *Tests for strategy loading and building*
  - *Tests for basic failover handling*
  - *Tests for concurrent failovers handling*

  - The serializers: (no)

## Documentation

  - If yes, how is the feature documented? (not documented yet)


flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/AdaptedRestartPipelinedRegionStrategyNG.java
```diff
@@ -0,0 +1,302 @@
...
+					// fail globally if any error happens
+					if (t != null) {
+						LOG.info("Unexpected error happens in region failover. Fail globally.", t);
+						failGlobal(t);
+					}
+				});
+	}
+
+	@VisibleForTesting
+	protected CompletableFuture<?> cancelTasks(final Set<ExecutionVertexID> verticesToRestart) {
```
eaglewatcherwb: argument should be `verticesToCancel`

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/AdaptedRestartPipelinedRegionStrategyNG.java
```diff
@@ -0,0 +1,302 @@
...
+	}
+
+	private CompletableFuture<?> cancelExecutionVertex(final ExecutionVertexID executionVertexId) {
+		return getExecutionVertex(executionVertexId).cancel();
+	}
+
+	private Map<JobVertexID, ExecutionJobVertex> getInvolvedExecutionJobVertices(
+		final Set<ExecutionVertex> executionVertices) {
+
+		Map<JobVertexID, ExecutionJobVertex> tasks = new HashMap<>(executionVertices.size());
```
eaglewatcherwb: initial size for map is not necessary. see: https://stackoverflow.com/questions/564899/choosing-the-initial-capacity-of-a-hashset-with-an-expected-number-of-unique-val

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/AdaptedRestartPipelinedRegionStrategyNG.java
```diff
@@ -0,0 +1,302 @@
...
+
+		for (ExecutionVertex ev : verticesToRestart) {
+			CoLocationGroup cgroup = ev.getJobVertex().getCoLocationGroup();
+			if (cgroup != null && !colGroups.contains(cgroup)){
+				cgroup.resetConstraints();
+				colGroups.add(cgroup);
+			}
+
+			ev.resetForNewExecution(restartTimestamp, globalModVersion);
+		}
```
eaglewatcherwb: code in line 167~178 is a duplicate of that in `FailoverRegion#reset`, which could be encapsulated as a static function and called by both `AdaptedRestartPipelinedRegionStrategyNG` and `FailoverRegion`.

	static void resetForNewRestartVertices(final Collection<ExecutionVertex> vertices, long globalModVersionOfFailover) throws Exception {
		final Collection<CoLocationGroup> colGroups = new HashSet<>();
		final long restartTimestamp = System.currentTimeMillis();

		for (ExecutionVertex ev : vertices) {
			CoLocationGroup cgroup = ev.getJobVertex().getCoLocationGroup();
			if (cgroup != null && !colGroups.contains(cgroup)){
				cgroup.resetConstraints();
				colGroups.add(cgroup);
			}

			ev.resetForNewExecution(restartTimestamp, globalModVersionOfFailover);
		}
	}

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/AdaptedRestartPipelinedRegionStrategyNG.java
```diff
@@ -0,0 +1,302 @@
...
+	private Map<JobVertexID, ExecutionJobVertex> getInvolvedExecutionJobVertices(
+		final Set<ExecutionVertex> executionVertices) {
+
+		Map<JobVertexID, ExecutionJobVertex> tasks = new HashMap<>(executionVertices.size());
+		for (ExecutionVertex executionVertex : executionVertices) {
+			JobVertexID jobvertexId = executionVertex.getJobvertexId();
+			ExecutionJobVertex jobVertex = executionVertex.getJobVertex();
+			tasks.putIfAbsent(jobvertexId, jobVertex);
+		}
+		return tasks;
```
eaglewatcherwb: code in line 245~251 is a duplicate of code in line 233~239 in `RestartPipelinedRegionStrategy#initTasks`.

eaglewatcherwb: code in line 245~ 251 is a duplicate of code in line 233~239 in `RestartPipelinedRegionStrategy#initTasks`.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/AdaptedRestartPipelinedRegionStrategyNGConcurrentFailoverTest.java
```diff
@@ -0,0 +1,416 @@
...
+		failoverStrategy.setBlockerFuture(new CompletableFuture<>());
+
+		final Iterator<ExecutionVertex> vertexIterator = eg.getAllExecutionVertices().iterator();
+		final ExecutionVertex ev11 = vertexIterator.next();
+		final ExecutionVertex ev12 = vertexIterator.next();
+		final ExecutionVertex ev21 = vertexIterator.next();
+		final ExecutionVertex ev22 = vertexIterator.next();
+
+		// start job scheduling
+		testMainThreadUtil.execute(() -> eg.scheduleForExecution());
```
eaglewatcherwb:
```suggestion
		testMainThreadUtil.execute(eg::scheduleForExecution);
```

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/AdaptedRestartPipelinedRegionStrategyNGConcurrentFailoverTest.java
```diff
@@ -0,0 +1,416 @@
...
+		final SimpleSlotProvider slotProvider = new SimpleSlotProvider(jid, DEFAULT_PARALLELISM);
+		final TestRestartStrategy restartStrategy = TestRestartStrategy.manuallyTriggered();
+
+		final ExecutionGraph eg = createExecutionGraph(
+			jid,
+			TestAdaptedRestartPipelinedRegionStrategyNG::new,
+			restartStrategy,
+			slotProvider);
+
+		final TestAdaptedRestartPipelinedRegionStrategyNG failoverStrategy =
```
eaglewatcherwb: `failoverStrategy` not used

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/AdaptedRestartPipelinedRegionStrategyNGConcurrentFailoverTest.java
```diff
@@ -0,0 +1,416 @@
...
+			(TestAdaptedRestartPipelinedRegionStrategyNG) eg.getFailoverStrategy();
+
+		final Iterator<ExecutionVertex> vertexIterator = eg.getAllExecutionVertices().iterator();
+		final ExecutionVertex ev11 = vertexIterator.next();
+		final ExecutionVertex ev12 = vertexIterator.next();
+		final ExecutionVertex ev21 = vertexIterator.next();
+		final ExecutionVertex ev22 = vertexIterator.next();
+
+		// start job scheduling
+		testMainThreadUtil.execute(() -> eg.scheduleForExecution());
```
eaglewatcherwb:
```suggestion
		testMainThreadUtil.execute(eg::scheduleForExecution);
```

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/AdaptedRestartPipelinedRegionStrategyNGConcurrentFailoverTest.java
```diff
@@ -0,0 +1,416 @@
...
+
+		// verify that no task is restarted
+		assertEquals(JobStatus.RESTARTING, eg.getState());
+		assertEquals(ExecutionState.CANCELED, ev11.getExecutionState());
+		assertEquals(ExecutionState.CANCELED, ev12.getExecutionState());
+		assertEquals(ExecutionState.CANCELED, ev21.getExecutionState());
+		assertEquals(ExecutionState.CANCELED, ev22.getExecutionState());
+
+		// trigger global recovery
+		testMainThreadUtil.execute(() -> restartStrategy.triggerNextAction());
```
eaglewatcherwb:
```suggestion
		testMainThreadUtil.execute(restartStrategy::triggerNextAction);
```

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/AdaptedRestartPipelinedRegionStrategyNGConcurrentFailoverTest.java
```diff
@@ -0,0 +1,416 @@
...
+		failoverStrategy.setBlockerFuture(new CompletableFuture<>());
+
+		final Iterator<ExecutionVertex> vertexIterator = eg.getAllExecutionVertices().iterator();
+		final ExecutionVertex ev11 = vertexIterator.next();
+		final ExecutionVertex ev12 = vertexIterator.next();
+		final ExecutionVertex ev21 = vertexIterator.next();
+		final ExecutionVertex ev22 = vertexIterator.next();
+
+		// start job scheduling
+		testMainThreadUtil.execute(() -> eg.scheduleForExecution());
```
eaglewatcherwb:
```suggestion
		testMainThreadUtil.execute(eg::scheduleForExecution);
```

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/AdaptedRestartPipelinedRegionStrategyNGConcurrentFailoverTest.java
```diff
@@ -0,0 +1,416 @@
...
+	/**
+	 * Test implementation of the {@link AdaptedRestartPipelinedRegionStrategyNG} that makes it possible
+	 * to control when the failover action is performed via {@link CompletableFuture}.
+	 */
+	static class TestAdaptedRestartPipelinedRegionStrategyNG extends AdaptedRestartPipelinedRegionStrategyNG {
+
+		@Nonnull
+		CompletableFuture<?> blockerFuture;
+
+		public TestAdaptedRestartPipelinedRegionStrategyNG(ExecutionGraph executionGraph) {
```
eaglewatcherwb: package-private is ok

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/AdaptedRestartPipelinedRegionStrategyNGConcurrentFailoverTest.java
```diff
@@ -0,0 +1,416 @@
...
+			super(executionGraph);
+			this.blockerFuture = CompletableFuture.completedFuture(null);
+		}
+
+		public void setBlockerFuture(@Nonnull CompletableFuture<?> blockerFuture) {
```
eaglewatcherwb: package-private is ok

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/AdaptedRestartPipelinedRegionStrategyNGFailoverTest.java
```diff
@@ -0,0 +1,771 @@
...
+
+	private static ExecutionVertex getExecutionVertex(
+		final ExecutionVertexID vertexID,
+		final ExecutionGraph executionGraph) {
+
+		return executionGraph.getJobVertex(vertexID.getJobVertexId())
+			.getTaskVertices()[vertexID.getSubtaskIndex()];
+	}
+
+	private static Set<ExecutionVertex> getExecutionVertices(
```
eaglewatcherwb: never used

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/AdaptedRestartPipelinedRegionStrategyNG.java
```diff
@@ -0,0 +1,302 @@
...
+					}
+
+					// found out vertices which are still valid to restart.
+					// some vertices involved in this failover may be modified if another region
+					// failover happens during the cancellation stage of this failover.
+					// Will ignore the modified vertices as the other failover will deal with them.
+					final Set<ExecutionVertex> unmodifiedVertices = executionVertexVersioner
+						.getUnmodifiedExecutionVertices(vertexVersions)
+						.stream()
+						.map(id -> getExecutionVertex(id))
```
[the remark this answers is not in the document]
zhuzhurk: Ok

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/AdaptedRestartPipelinedRegionStrategyNG.java
```diff
@@ -0,0 +1,302 @@
...
+
+						// reset tasks to CREATED state and reload state
+						resetTasks(unmodifiedVertices, globalModVersion);
+
+						// re-schedule tasks
+						rescheduleTasks(unmodifiedVertices, globalModVersion);
+					} catch (GlobalModVersionMismatch e) {
+						// happens when a global recovery happens concurrently to the regional recovery
+						// just stop this local failover
+						return;
```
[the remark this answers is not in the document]
zhuzhurk: No necessary log here. It should not cause any issue. Concurrent failover should not be a big issue.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/AdaptedRestartPipelinedRegionStrategyNG.java
zhuzhurk: Ok.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/AdaptedRestartPipelinedRegionStrategyNG.java
zhuzhurk: Ok.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/AdaptedRestartPipelinedRegionStrategyNG.java
zhuzhurk: This strategy should be a replacement of the legacy RestartPipelinedStrategy, as well as its sub components like FailoverRegion. And it's a temporary solution as well.
In the next few versions, I think we may gradually remove legacy RestartPipelinedStrategy, then remove this adaptor, and then use the new generation FailoverStrategy.

To keep simplicity, I think we'd better not changing existing logic, or as less as possible.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/AdaptedRestartPipelinedRegionStrategyNGConcurrentFailoverTest.java
zhuzhurk: Ok.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/AdaptedRestartPipelinedRegionStrategyNGConcurrentFailoverTest.java
zhuzhurk: Ok.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/AdaptedRestartPipelinedRegionStrategyNGConcurrentFailoverTest.java
zhuzhurk: Ok.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/AdaptedRestartPipelinedRegionStrategyNGConcurrentFailoverTest.java
zhuzhurk: Ok.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/AdaptedRestartPipelinedRegionStrategyNGFailoverTest.java
zhuzhurk: Ok.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/AdaptedRestartPipelinedRegionStrategyNG.java
zhuzhurk: Same as above.
To keep simplicity, I think we'd better make changes to existing logic as less as possible.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/AdaptedRestartPipelinedRegionStrategyNGConcurrentFailoverTest.java
zhuzhurk: Ok.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/AdaptedRestartPipelinedRegionStrategyNGConcurrentFailoverTest.java
zhuzhurk: Ok

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/AdaptedRestartPipelinedRegionStrategyNGConcurrentFailoverTest.java
zhuzhurk: Ok.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/AdaptedSchedulingUtils.java
```diff
@@ -0,0 +1,177 @@
...
+					previousAllocations);
+
+				schedulingFutures.add(schedulingVertexFuture);
+			}
+		}
+
+		return FutureUtils.waitForAll(schedulingFutures);
+	}
+
+	static CompletableFuture<Void> scheduleEager(
```
GJL: I don't feel comfortable duplicating all this code from `ExecutionGraph`. I think at least, the `ExecutionGraph` can call this function: 

```
final Set<ExecutionVertex> executionVertices = StreamSupport
			.stream(getAllExecutionVertices().spliterator(), false)
			.collect(Collectors.toSet());

		return AdaptedSchedulingUtils.scheduleEager(executionVertices, this);
```


(This is not a review)

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/FailoverStrategy.java
```diff
@@ -34,22 +34,4 @@
...
 	Set<ExecutionVertexID> getTasksNeedingRestart(ExecutionVertexID executionVertexId, Throwable cause);
-
-	// ------------------------------------------------------------------------
-	//  factory
-	// ------------------------------------------------------------------------
-
-	/**
-	 * The factory to instantiate {@link FailoverStrategy}.
-	 */
-	interface Factory {
```
GJL: Don't we need this interface again in the future?

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/AdaptedRestartPipelinedRegionStrategyNG.java
```diff
@@ -0,0 +1,295 @@
...
+				newSchedulingFuture = AdaptedSchedulingUtils.scheduleEager(vertices, executionGraph);
+				break;
+
+			default:
+				throw new JobException("Schedule mode is invalid.");
+		}
+
+		// if no global failover is triggered in the scheduling process,
+		// register a failure handling callback to the scheduling
+		if (isLocalFailoverValid(globalModVersion)) {
```
GJL: I think between `AdaptedSchedulingUtils.schedule[...]()` and this line, there cannot be modifications of the ExecutionGraph.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/AdaptedRestartPipelinedRegionStrategyNGConcurrentFailoverTest.java
```diff
@@ -0,0 +1,411 @@
...
+import org.apache.flink.runtime.jobgraph.DistributionPattern;
+import org.apache.flink.runtime.jobgraph.JobGraph;
+import org.apache.flink.runtime.jobgraph.JobStatus;
+import org.apache.flink.runtime.jobgraph.JobVertex;
+import org.apache.flink.runtime.jobmaster.slotpool.SlotProvider;
+import org.apache.flink.runtime.scheduler.strategy.ExecutionVertexID;
+import org.apache.flink.runtime.testingUtils.TestingUtils;
+import org.apache.flink.runtime.testtasks.NoOpInvokable;
+import org.apache.flink.util.TestLogger;
+import org.junit.ClassRule;
```
GJL: image

```
    Import order. Imports must be ordered alphabetically, grouped into the following blocks, with each block separated by an empty line:
        <imports from org.apache.flink.*>
        <imports from org.apache.flink.shaded.*>
        <imports from other libraries>
        <imports from javax.*>
        <imports from java.*>
        <imports from scala.*>
        <static imports>

```

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/AdaptedSchedulingUtils.java
```diff
@@ -0,0 +1,177 @@
...
+public class AdaptedSchedulingUtils {
+
+	static CompletableFuture<Void> scheduleLazy(
+		final Set<ExecutionVertex> vertices,
+		final ExecutionGraph executionGraph) {
+
+		final Set<AllocationID> previousAllocations = getAllPriorAllocationIds(vertices);
+
+		final ArrayList<CompletableFuture<Void>> schedulingFutures = new ArrayList<>(vertices.size());
+		for (ExecutionVertex executionVertex : vertices) {
```
GJL: Are there downsides to giving up the topological sorting?

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/FailoverStrategyLoader.java
```diff
@@ -63,7 +63,7 @@
 					return new RestartAllStrategy.Factory();
 
 				case PIPELINED_REGION_RESTART_STRATEGY_NAME:
```
GJL: I think the user should be able to keep using the legacy `RestartPipelinedRegionStrategy` for now.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/ExecutionVertex.java
```diff
@@ -576,7 +576,7 @@ else if (numSources < parallelism) {
 	 */
 	public Execution resetForNewExecution(final long timestamp, final long originatingGlobalModVersion)
 			throws GlobalModVersionMismatch {
-		LOG.debug("Resetting execution vertex {} for new execution.", getTaskNameWithSubtaskIndex());
+		LOG.info("Resetting execution vertex {} to CREATED state for new execution.", getTaskNameWithSubtaskIndex());
```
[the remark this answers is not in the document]
zhuzhurk: Currently we have log tracking vertex state transitions in most cases(through Execution::transitionState), but lack log of transition to CREATED state(it is directly set). I think this could be helpful to track vertex lifecycle. 
WDYT?

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/FailoverStrategyLoader.java
zhuzhurk: Shall we make AdaptedRestartPipelinedRegionStrategyNG as default "region" failover strategy and add a "region_legacy" for RestartPipelinedRegionStrategy?
As the legacy RestartPipelinedRegionStrategy is not applicable for batch jobs.

Or adding  a "region_new" for AdaptedRestartPipelinedRegionStrategyNG and let users to select it when running batch jobs?

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/AdaptedSchedulingUtils.java
zhuzhurk: Theoretically, it should not cause any issue as these vertices are scheduled individually. And only vertex satisfying the input constraint will be scheduled. 

In implementation, however, there is a bit higher chance that downstream task get launched earlier than its source tasks, and causes a resource deadlock. But even doing it topologically does not prevent this to happen. 
To avoid resource deadlock, users need to set `InputDependencyConstraint` to be ALL. In this case, any schedule order should be OK.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/AdaptedRestartPipelinedRegionStrategyNGConcurrentFailoverTest.java
zhuzhurk: Sorry for missing checking it. Will fix it.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/AdaptedRestartPipelinedRegionStrategyNG.java
zhuzhurk: This is to ensure that no global failure happens during the sync part of the schedule logic. Usually this should not happen. It's a insurance without much cost.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/FailoverStrategy.java
zhuzhurk: So far I think there is no need to create a JM shared FailoverStrategy factory. Each JM/Scheduler can create their own failoverStrategy directly.
We should implement a loader util for creating failoverStrategy from ExecutionGraph though.
WDYT?

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/AdaptedSchedulingUtils.java
zhuzhurk: I think its a balance between simplicity, risk and maintainability.
I'd be fine to have EG to call these methods to do scheduling.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/AdaptedSchedulingUtils.java
zhuzhurk: Above discusses for LAZY scheduling. For EAGER mode, unordered scheduling increase the chance that downstream tasks get launched earlier. In this case downstream tasks will fail to request partition and will trigger lots of partition state check RPCs to JM, which may result in JM main thread get jammed.

So I think it can be beneficial to order the vertices before scheduling, both for eager or lazy mode. Will do it in the next commit.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/ExecutionVertex.java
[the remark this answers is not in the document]
GJL: I find `Resetting [...] to CREATED [...] for new execution` to be a bit verbose. I would be ok to change the log level to `INFO` so that it is consistent with. 
```
LOG.info("{} ({}) switched from {} to {}.", getVertex().getTaskNameWithSubtaskIndex(), getAttemptId(), currentState, targetState);
```

I would make it  a separate `[hotfix]` commit.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/AdaptedRestartPipelinedRegionStrategyNG.java
GJL: Does it _usually not happen_, or is it forbidden to happen by contract?

I would rather add a check asserting that `rescheduleTasks()` is executed in the main thread. The code should be written so that it relies on well-defined contracts/invariants otherwise the next person who has to maintain the code might end up adding more paranoid `isLocalFailoverValid()` checks.

Here is more food for thought:
- Is it correct to add a null check after calling a method that is annotated with `@Nonnull`?
- Is it correct to add a null check after calling a method that returns `Optional`?

I think the same principle applies here.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/ExecutionVertex.java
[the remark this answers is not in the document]
zhuzhurk: Ok. Let's revert this change and do it later in a separate fix.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/FailoverStrategy.java
GJL: I cannot give you a definite answer right now but it seems to be that deleting the `Factory` interface is not needed for the adapter.  I think we can make a decision once the default `SchedulerNG` implementation is done.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/FailoverStrategyLoader.java
GJL: Here is what I think:
- default should be `AdaptedRestartPipelinedRegionStrategyNG`
- `region-legacy` will use `RestartPipelinedRegionStrategy`
- `region` will use `AdaptedRestartPipelinedRegionStrategyNG`

User will get the best experience out of the box but can switch to old behavior in case of bugs.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/AdaptedRestartPipelinedRegionStrategyNG.java
zhuzhurk: To clarify, this check is not for concurrent failovers. So main thread check does not help here. 

The main concern here is that some synchronize part of scheduling might cause task failure immediately. In this case, for example, before invoking `scheduleEager` the job is RUNNING and globalModVersion==1, after `scheduleEager ` returns, the job status is FAILING and globalModVersion==2. 

Although we can have contract that there should be no direct failing in the scheduling process, it can be hard to guarantee that. That's why I'd add a check here as an insurance.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/FailoverStrategy.java
zhuzhurk: Ok. Let's exclude this change for now.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/FailoverStrategyLoader.java
zhuzhurk: Ok. Let's take this way.

GJL: If it's too much work to get all tests running with `AdaptedRestartPipelinedRegionStrategyNG` as the default, I would be ok with leaving it to "restart all". We will have to fix tests in a separate PR.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/AdaptedRestartPipelinedRegionStrategyNGTest.java
```diff
@@ -0,0 +1,171 @@
...
+	 *     (v12) --> (v22) -+-> (v32) --> (v42)
+	 *
+	 *                      ^
+	 *                      |
+	 *                  (blocking)
+	 * </pre>
+	 * 4 regions. Each region has 2 pipelined connected vertices.
+	 */
+	@Test
+	public void testRegionBuilding() throws Exception {
```
GJL: This test doesn't add code coverage because we are redunantly unit testing the behavior of `RestartPipelinedRegionStrategy`. We should test the behavior that is added by `AdaptedRestartPipelinedRegionStrategyNG`.

I think we can delete it together with the method `AdaptedRestartPipelinedRegionStrategyNG#getFailoverRegion()`.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/AdaptedRestartPipelinedRegionStrategyNGTest.java
```diff
@@ -0,0 +1,171 @@
...
+	/**
+	 * Test that {@link AdaptedRestartPipelinedRegionStrategyNG} is loaded
+	 * as failover strategy if {@link JobManagerOptions#EXECUTION_FAILOVER_STRATEGY}
+	 * value is "region".
+	 */
+	@Test
+	public void testAdaptedRestartPipelinedRegionStrategyNGLoading() throws Exception {
+		final ExecutionGraph eg = createExecutionGraph(new JobGraph());
+
+		assertTrue(eg.getFailoverStrategy() instanceof AdaptedRestartPipelinedRegionStrategyNG);
```
GJL: I think we can delete the whole class. This test should not be in this class because it is not testing any behavior of `FailoverStrategyLoader`. We could test `FailoverStrategyLoader` explicitly or omit this test entirely.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/AdaptedRestartPipelinedRegionStrategyNGFailoverTest.java
```diff
@@ -0,0 +1,763 @@
...
+	 *     (v12) --> (v22) -+-> (v32) --> (v42)
+	 *
+	 *                      ^
+	 *                      |
+	 *                  (blocking)
+	 * </pre>
+	 * 4 regions. Each region has 2 pipelined connected vertices.
+	 */
+	@Test
+	public void testRegionFailoverForRegionInternalErrorsInLazyMode() throws Exception {
```
GJL: Compared to `testRegionFailoverInEagerMode()`, what does this test add? As far as I can see, additional lines that are covered are:

```
case LAZY_FROM_SOURCES:
	newSchedulingFuture = AdaptedSchedulingUtils.scheduleLazy(vertices, executionGraph);
	break;
```

Is it justified to complicate the testing job topology for these additional lines, or can these be tested differently?

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/AdaptedRestartPipelinedRegionStrategyNGFailoverTest.java
```diff
@@ -0,0 +1,763 @@
...
+		final ExecutionGraph eg = createExecutionGraph(jobGraph, checkpointId);
+
+		final Iterator<ExecutionVertex> vertexIterator = eg.getAllExecutionVertices().iterator();
+		final ExecutionVertex ev11 = vertexIterator.next();
+		final ExecutionVertex ev12 = vertexIterator.next();
+		final ExecutionVertex ev21 = vertexIterator.next();
+		final ExecutionVertex ev22 = vertexIterator.next();
+
+		// verify initial state
+		assertVertexInState(ExecutionState.DEPLOYING, ev11);
```
GJL: This test is called `AdaptedRestartPipelinedRegionStrategyNGFailoverTest`. At this point in the test, no methods have been invoked on the instance under test (`AdaptedRestartPipelinedRegionStrategyNG`). Are these assertions really adding to the test coverage? Imo, we are redundantly unit testing `ExecutionVertex` here.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/AdaptedRestartPipelinedRegionStrategyNGFailoverTest.java
```diff
@@ -0,0 +1,763 @@
...
+		assertEquals(0, ev21.getCurrentExecutionAttempt().getAttemptNumber());
+		assertEquals(0, ev22.getCurrentExecutionAttempt().getAttemptNumber());
+		assertEquals(0, ev31.getCurrentExecutionAttempt().getAttemptNumber());
+		assertEquals(0, ev32.getCurrentExecutionAttempt().getAttemptNumber());
+		assertEquals(0, ev41.getCurrentExecutionAttempt().getAttemptNumber());
+		assertEquals(0, ev42.getCurrentExecutionAttempt().getAttemptNumber());
+
+		// trigger task failure of ev31 on consuming data from ev21
+		// vertices { ev11, ev21, ev31, ev32, ev41, ev42} should be affected
+		testMainThreadUtil.execute(() -> ev31.getCurrentExecutionAttempt().fail(new PartitionConnectionException(
```
GJL: We are mostly testing the logic of `RestartPipelinedRegionStrategy`. A test like this is not justified if we want to test that `cause` is correctly propagated to `getTasksNeedingRestart()`.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/AdaptedRestartPipelinedRegionStrategyNGFailoverTest.java
```diff
@@ -0,0 +1,763 @@
...
+	private final TestingComponentMainThreadExecutor testMainThreadUtil =
+		EXECUTOR_RESOURCE.getComponentMainThreadTestExecutor();
+
+	/**
+	 * Tests for region failover for job in EAGER mode.
+	 * This applies to streaming job, with no BLOCKING edge.
+	 * <pre>
+	 *     (v11) -+-> (v21)
+	 *
+	 *     (v12) -+-> (v22)
```
GJL: What is the difference between `-->` and `-+->`?

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/AdaptedRestartPipelinedRegionStrategyNG.java
GJL: Ok, if it can happen, then we can leave it.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/AdaptedRestartPipelinedRegionStrategyNGTest.java
zhuzhurk: Ok.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/AdaptedRestartPipelinedRegionStrategyNGTest.java
zhuzhurk: Fine. Let's omit this test as the loading behavior is actually covered by some other AdaptedRestartPipelinedRegionStrategyNG tests.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/AdaptedRestartPipelinedRegionStrategyNGFailoverTest.java
zhuzhurk: No difference here. I will change it to ---> to avoid confusion.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/AdaptedRestartPipelinedRegionStrategyNGConcurrentFailoverTest.java
```diff
@@ -0,0 +1,413 @@
...
+	 *            x
+	 *     (v12) -+-> (v22)
+	 *
+	 *            ^
+	 *            |
+	 *       (blocking)
+	 * </pre>
+	 */
+	@Test
+	public void testRegionFailoverWhenInGlobalFailover() throws Exception {
```
GJL: Do we need this test? It does not invoke methods on the adapter.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/AdaptedRestartPipelinedRegionStrategyNGConcurrentFailoverTest.java
```diff
@@ -0,0 +1,413 @@
...
+		JobID jid,
+		Factory failoverStrategy,
+		RestartStrategy restartStrategy,
+		SlotProvider slotProvider) throws Exception {
+
+		final JobInformation jobInformation = new DummyJobInformation(
+			jid,
+			"test job");
+
+		// build a simple execution graph with on job vertex, parallelism 2
```
zentol: typo: on -> one

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/AdaptedRestartPipelinedRegionStrategyNGFailoverTest.java
```diff
@@ -0,0 +1,763 @@
...
+		testMainThreadUtil.execute(() ->ev21.getCurrentExecutionAttempt().completeCancelling());
+
+		verifyCheckpointRestoredAsExpected(checkpointId, eg);
+	}
+
+	/**
+	 * Tests that the vertex partition state is reset property during failover.
+	 */
+	@Test
+	public void testStatusResettingOnRegionFailover() throws Exception {
```
GJL: Is this test only to validate that `resetForNewExecution()` is called on the `ExecutionVertex`?

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/AdaptedRestartPipelinedRegionStrategyNG.java
```diff
@@ -0,0 +1,296 @@
...
+			if (cgroup != null && !colGroups.contains(cgroup)){
+				cgroup.resetConstraints();
+				colGroups.add(cgroup);
+			}
+
+			ev.resetForNewExecution(restartTimestamp, globalModVersion);
+		}
+
+		// if there is checkpointed state, reload it into the executions
+		if (executionGraph.getCheckpointCoordinator() != null) {
```
GJL: What are the chances to move this into `resetForNewExecution()`? Now this code is duplicated.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/FailoverStrategyLoader.java
zhuzhurk: Ok. Let's make the adapted region failover strategy default in another PR.

flink-tests/src/test/java/org/apache/flink/test/runtime/RegionFailoverITCase.java
```diff
@@ -0,0 +1,216 @@
...
+
+import static org.apache.flink.configuration.JobManagerOptions.EXECUTION_FAILOVER_STRATEGY;
+import static org.apache.flink.util.Preconditions.checkNotNull;
+import static org.hamcrest.Matchers.is;
+import static org.junit.Assert.assertThat;
+
+/**
+ * IT case for testing region failover strategy.
+ */
+public class RegionFailoverITCase extends TestLogger {
```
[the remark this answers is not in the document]
zhuzhurk: I take it as E2E tests that ensures the job can succeed even when task failures happens when using region failover strategy.
IT cases do blackbox verifications so it does not verify internal states(like which tasks are re-scheduled).
It just gives us confidence that Flink runtime is working well in this scenario.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/AdaptedRestartPipelinedRegionStrategyNGConcurrentFailoverTest.java
```diff
@@ -0,0 +1,413 @@
...
+ */
+public class AdaptedRestartPipelinedRegionStrategyNGConcurrentFailoverTest extends TestLogger {
+
+	public final int DEFAULT_PARALLELISM = 2;
+
+	@ClassRule
+	public static final TestingComponentMainThreadExecutor.Resource EXECUTOR_RESOURCE =
+		new TestingComponentMainThreadExecutor.Resource();
+
+	private final TestingComponentMainThreadExecutor testMainThreadUtil =
```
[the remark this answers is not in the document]
zhuzhurk: Ok.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/AdaptedRestartPipelinedRegionStrategyNGConcurrentFailoverTest.java
zhuzhurk: legacy log. Will remove it.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/AdaptedRestartPipelinedRegionStrategyNGConcurrentFailoverTest.java
```diff
@@ -0,0 +1,413 @@
...
+			assertEquals(ExecutionState.DEPLOYING, ev11.getExecutionState());
+			assertEquals(ExecutionState.DEPLOYING, ev12.getExecutionState());
+			assertEquals(ExecutionState.CREATED, ev21.getExecutionState());
+			assertEquals(ExecutionState.CREATED, ev22.getExecutionState());
+		});
+
+		// complete region failover blocker to trigger region failover
+		testMainThreadUtil.execute(() -> failoverStrategy.blockerFuture.complete(null));
+
+		// verify that no task is restarted by region failover
```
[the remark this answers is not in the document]
zhuzhurk: Yes. Some vertices will get reset twice and their attempt number will be larger than 1.

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/ExecutionVertexVersioner.java
```diff
@@ -0,0 +1,76 @@
...
+ *
+ * <p>Examples for modifications include:
+ * <ul>
+ *     <li>cancellation of the underlying execution
+ *     <li>deployment of the execution vertex
+ * </ul>
+ *
+ * @see DefaultScheduler
+ */
+public class ExecutionVertexVersioner {
```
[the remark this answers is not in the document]
zhuzhurk: Yes. Will add tests for it.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/AdaptedRestartPipelinedRegionStrategyNGConcurrentFailoverTest.java
```diff
@@ -0,0 +1,413 @@
...
+		final SimpleSlotProvider slotProvider = new SimpleSlotProvider(jid, DEFAULT_PARALLELISM);
+		final TestRestartStrategy restartStrategy = TestRestartStrategy.manuallyTriggered();
+
+		final ExecutionGraph eg = createExecutionGraph(
+			jid,
+			TestAdaptedRestartPipelinedRegionStrategyNG::new,
+			restartStrategy,
+			slotProvider);
+
+		final TestAdaptedRestartPipelinedRegionStrategyNG failoverStrategy =
```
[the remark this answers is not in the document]
zhuzhurk: The TestAdaptedRestartPipelinedRegionStrategyNG need EG as the constructor param.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/AdaptedRestartPipelinedRegionStrategyNGConcurrentFailoverTest.java
```diff
@@ -0,0 +1,413 @@
...
+		assertEquals(ExecutionState.CANCELED, ev12.getExecutionState());
+		assertEquals(ExecutionState.CANCELED, ev21.getExecutionState());
+		assertEquals(ExecutionState.CANCELED, ev22.getExecutionState());
+
+		// trigger global recovery
+		testMainThreadUtil.execute(restartStrategy::triggerNextAction);
+
+		// verify the state and attempt number
+		// do in main thread as the global recovery process is scheduled async to this thread
+		testMainThreadUtil.execute(() -> {
```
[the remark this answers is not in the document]
zhuzhurk: In current implementation it should work.
But you are right, there is risk here, if the restart delay is not set 0 or if there are multiple chained main thread scheduled actions.

Will make the check more stable for all cases.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/AdaptedRestartPipelinedRegionStrategyNG.java
zhuzhurk: This is not ExecutionVertex level operation so it cannot be merged into ExecutionVertex::resetForNewExecution.
If you are really concerning about the code duplication of partial checkpoint restore between the new strategy and the legacy strategy, I can merge them.

IMO, we keep the legacy RestartPipelinedStrategy in case that the new strategy has some bugs. Making changes to legacy RestartPipelinedStrategy is adding bug risk to it. I think it would be better to leave it as is, since it is actually in deprecated state.

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/ExecutionVertexVersioner.java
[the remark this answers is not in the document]
zhuzhurk: ExecutionVertexVersioner is from @GJL . He has also implemented tests for them https://github.com/GJL/flink/blob/FLINK-12433/flink-runtime/src/test/java/org/apache/flink/runtime/scheduler/ExecutionVertexVersionerTest.java.

I've ported these tests and made some minor changes.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/AdaptedRestartPipelinedRegionStrategyNGFailoverTest.java
zhuzhurk: Will simplify the test to test the cause propagation only.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/AdaptedRestartPipelinedRegionStrategyNGFailoverTest.java
zhuzhurk: It is to verify whether failover resetting partition state is correctly.
Will remove it as vertex resetting in adapter is covered by other cases. And resetting partition when resetting vertex is not testing changes in this PR.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/AdaptedRestartPipelinedRegionStrategyNGFailoverTest.java
zhuzhurk: Will simplify the topology to 4 regions with one vertex each.
There were 2 purposes to complicate the topology:
1. cover the case that each region has more than one vertices. It's not necessary anyhow
2. make it easier to control the cancelling termination. Yet we have other approaches to do this

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/SchedulingUtils.java
```diff
@@ -0,0 +1,219 @@
...
+			return computePriorAllocationIds(vertices);
+		} else {
+			return Collections.emptySet();
+		}
+	}
+
+	/**
+	 * Computes and returns a set with the prior allocation ids for given execution vertices.
+	 */
+	static private Set<AllocationID> computePriorAllocationIds(final Collection<ExecutionVertex> vertices) {
```
GJL: image

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/AdaptedRestartPipelinedRegionStrategyNGFailoverTest.java
```diff
@@ -0,0 +1,641 @@
...
+			AkkaUtils.getDefaultTimeout(),
+			restartStrategy,
+			TestAdaptedRestartPipelinedRegionStrategyNG::new,
+			slotProvider);
+		try {
+			eg.attachJobGraph(jobGraph.getVerticesSortedTopologicallyFromSources());
+		}
+		catch (JobException e) {
+			e.printStackTrace();
+			fail("Job failed with exception: " + e.getMessage());
```
GJL: This is an anti-pattern. The `try-catch` should be removed:

- createExecutionGraph() already declares `throws Exception`
- Test should fail if an exception is thrown

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/SchedulingUtils.java
zhuzhurk: Will fix it.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/AdaptedRestartPipelinedRegionStrategyNGFailoverTest.java
zhuzhurk: Will fix it.

GJL:
Superseded by #8922

PR #8852, opened 2019-06-24
 [FLINK-12798][table-api][table-planner] Add a proper discover mechanism that will  enable switching between Flink & Blink Planner/Executor

dawidwys:

This PR improves the lookup mechanism for `Planner` & `Executor` to enable switching between Flink & Blink planner components/

## Brief change log

  - Improved type handling in `TableFactoryService`
  - Enabled returning multiple matching `TableFactory`s from `TableFactoryService`
  - Added `ExecutorFactory` & `PlannerFactory` that can be discovered based on string properties
  - Added `PlannerDescriptor` to enable switching between different planner components in the API.

## Verifying this change

Added tests for disambiguating factories based on `optionalContext`: `org.apache.flink.table.factories.ComponentFactoryServiceTest`

## Documentation

  - If yes, how is the feature documented? (not applicable / docs / JavaDocs / **not documented**): should be added as a follow-up as part of major rework of docs for the updated Table API


flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/factories/ComponentTableFactory.java
```diff
@@ -0,0 +1,61 @@
...
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * A common inteface for {@link PlannerFactory} and
+ * {@link org.apache.flink.table.executor.ExecutorFactory}. It enables further disambiguating if there are mutliple
+ * implementations present.
```
godfreyhe: please correct package of `ExecutorFactory`

nit: inteface -> interface, mutliple -> multiple

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/delegation/ExecutorFactory.java
```diff
@@ -0,0 +1,50 @@
...
+/**
+ * Factory that creates {@link Executor}.
+ *
+ * <p>This factory is used with Java's Service Provider Interfaces (SPI) for discovering. A factory is
+ * called with a set of normalized properties that describe the desired configuration. Those properties
+ * may include execution configurations such as watermark interval, max parallelism etc., table specific
+ * initialization configuration such as if the queries should be executed in batch mode.
+ *
+ * <p><b>Important:</b> The implementations of this interface should also implement method
+ * <pre>{@code public Executor create(Map<String, String> propertis, StreamExecutionEnvironment executionEnvironmet);}
```
godfreyhe: nit: propertis -> properties, executionEnvironmet -> executionEnvironment

flink-table/flink-table-planner/src/main/java/org/apache/flink/table/planner/StreamPlannerFactory.java
```diff
@@ -43,27 +46,26 @@
...
-			FunctionCatalog functionCatalog,
-			CatalogManager catalogManager) {
-		try {
-			Class<?> clazz = Class.forName("org.apache.flink.table.planner.StreamPlanner");
-			Constructor con = clazz.getConstructor(
-				Executor.class,
-				TableConfig.class,
-				FunctionCatalog.class,
-				CatalogManager.class);
+	public Planner create(
```
godfreyhe: please update the javadoc

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/factories/ComponentTableFactory.java
```diff
@@ -0,0 +1,61 @@
...
+ */
+@Internal
+public interface ComponentTableFactory extends TableFactory {
```
aljoscha: This way round it sounds like a Table factory for components. Maybe it could be `TableComponentFactory`, or `ComponentFactory`.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/descriptors/PlannerDescriptor.java
```diff
@@ -0,0 +1,102 @@
...
+
+import javax.annotation.Nullable;
+
+import java.util.Map;
+
+/**
+ * Common class that defines necessary properties to choose and create both
+ * {@link Executor} and {@link Planner}.
+ */
+public class PlannerDescriptor {
```
aljoscha: Do we have `@PublicEvolcing` annotations in the rest of the new api package? Then we should probably also have them here and in the other new classes.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/descriptors/PlannerDescriptor.java
```diff
@@ -0,0 +1,102 @@
...
+	@VisibleForTesting
+	private PlannerDescriptor(@Nullable String plannerClass, @Nullable String executorClass) {
+		this.plannerClass = plannerClass;
+		this.executorClass = executorClass;
+	}
+
+	/**
+	 * Sets the legacy planner as the required module.
+	 */
+	public static PlannerDescriptor legacy() {
```
aljoscha: I don't like the hardcoded methods to much because it basically links the descriptor tightly to the planners at the time when this was created. `legacy()` will be outdated soon, and then `blink()` will be outdated as well. An alternative could be a method `planner(Class<? extends Planner> planner)` and `planner(String plannerClass)`. 

What do you think? With this, `any()` could also be renamed to `create()`, which would then roughly yield:
```
PlannerDescriptor desscriptor = PlannerDescriptor.create()
        .planner(StreamPlannerFactory.class)
        .executor(...)
        .batch()
```

I realize that calling planner and executor is a bit wordy, though... 😓

dawidwys: I started with the approach you described. The problem with it is that we require that users now the class names of planner and executor, plus as you noticed they have to specify both. 

With the `legacy()`/`blink()` approach they are preconfigured and user does not need to know about the underlying classes.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/descriptors/PlannerDescriptor.java
dawidwys: Good point!

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/factories/ComponentTableFactory.java
dawidwys: I will rename it to `ComponentFactory`

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/descriptors/PlannerDescriptor.java
dawidwys: What other classes did you have in mind? I think this is the only class missing annotation.

aljoscha: I didn't have any in mind, I just noticed it here 😅

flink-table/flink-table-common/src/main/java/org/apache/flink/table/api/AmbiguousTableFactoryException.java
```diff
@@ -31,7 +31,7 @@
 public class AmbiguousTableFactoryException extends RuntimeException {
 
 	// factories that match the properties
-	private final List<TableFactory> matchingFactories;
+	private final List<?> matchingFactories;
```
twalthr: Why do we need a wildcard here? All objects in this list are still table factories, no?

flink-table/flink-table-common/src/main/java/org/apache/flink/table/factories/TableFactoryService.java
```diff
@@ -103,9 +106,37 @@
...
+			Class<T> factoryClass,
+			Map<String, String> propertyMap,
+			ClassLoader classLoader) {
 		Preconditions.checkNotNull(classLoader);
-		return findInternal(factoryClass, propertyMap, Optional.of(classLoader));
+		return findSingleInternal(factoryClass, propertyMap, Optional.of(classLoader));
+	}
+
+	/**
+	 * Finds a table factory of the given class and descriptor.
```
twalthr: Update doc. It find more than one table factory.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/factories/TableFactoryService.java
```diff
@@ -117,19 +148,65 @@
...
+			Class<T> factoryClass,
+			Map<String, String> properties,
+			Optional<ClassLoader> classLoader) {
+
+		List<TableFactory> tableFactories = discoverFactories(classLoader);
+		return filter(tableFactories, factoryClass, properties);
+	}
+
+	/**
+	 * Finds a table factory of the given class, property map, and classloader.
```
twalthr: update doc, there is no classloader?

flink-table/flink-table-planner/src/main/java/org/apache/flink/table/executor/StreamExecutorFactory.java
```diff
@@ -21,35 +21,51 @@
...
-	 * Creates a {@link StreamExecutor} that is backed by given {@link StreamExecutionEnvironment}.
+	 * Creates a corresponding {@link StreamExecutor}.
 	 *
+	 * @param properties Static properties of the {@link Executor}, the same that were used for factory lookup.
 	 * @param executionEnvironment a {@link StreamExecutionEnvironment} to use while executing Table programs.
-	 * @return {@link StreamExecutor}
+	 * @return instance of a {@link Executor}
 	 */
-	public static Executor create(StreamExecutionEnvironment executionEnvironment) {
+	public Executor create(Map<String, String> properties, StreamExecutionEnvironment executionEnvironment) {
```
twalthr: nit: the required parameters first?

flink-table/flink-table-common/src/main/java/org/apache/flink/table/api/AmbiguousTableFactoryException.java
dawidwys: This makes it easier to use with the subtypes of `TableFactory`, e.g. you can pass `List<PlannerFactory>`. I changed it though to `List<? extends TableFactory>`

flink-table/flink-table-planner/src/main/java/org/apache/flink/table/executor/StreamExecutorFactory.java
dawidwys: Agree. That's why the `StreamExecutionEnvironment` is second, as this is optional. See `create(Map<String, String> properties)`.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/EnvironmentSettings.java
```diff
@@ -0,0 +1,217 @@
...
+
+/**
+ * All properties that define and initialize a table environment. Those properties are used only
+ * during instantiation of a {@link TableEnvironment}.
+ *
+ * <p>Example:
+ * <pre>{@code
+ *    EnvironmentSettings.newInstance()
+ *      .useOldPlanner()
+ *      .inStreamMode()
```
twalthr: nit: batch and streaming? so `inStreamingMode`?

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/EnvironmentSettings.java
```diff
@@ -0,0 +1,217 @@
...
+	 */
+	public static class Builder {
+		private String plannerClass = null;
+		private String executorClass = null;
+		private String builtInCatalogName = "default_catalog";
+		private String builtInDatabaseName = "default_database";
+		private boolean isBatchMode = false;
+
+		/**
+		 * Sets the old Flink planner as the required module.
```
twalthr: Mentioned what the default behavior is.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/EnvironmentSettings.java
```diff
@@ -0,0 +1,217 @@
...
+		/**
+		 * Sets that the components should work in a stream mode.
+		 */
+		public Builder inStreamMode() {
+			this.isBatchMode = false;
+			return this;
+		}
+
+		/**
+		 * Specifies the name of the initial catalog to be created when instantiating
```
twalthr: Mention what the default value is?

flink-table/flink-table-api-scala-bridge/src/main/scala/org/apache/flink/table/api/scala/BatchTableEnvironment.scala
```diff
@@ -233,11 +233,12 @@ object BatchTableEnvironment {
           classOf[ExecutionEnvironment],
           classOf[TableConfig],
           classOf[CatalogManager])
+      val defaultCatalog = "default_catalog"
```
twalthr: Could we just use `EnvironmentSettings.newInstance().build()` to have just a single location that defines those constants?

flink-table/flink-table-api-java-bridge/src/main/java/org/apache/flink/table/api/java/internal/StreamTableEnvironmentImpl.java
```diff
@@ -114,11 +118,35 @@ public static StreamTableEnvironmentImpl create(
...
+	/**
+	 * Creates an instance of a {@link StreamTableEnvironment}. It uses the {@link StreamExecutionEnvironment} for
+	 * executing queries. This is also the {@link StreamExecutionEnvironment} that will be used when converting
+	 * from/to {@link DataStream}.
+	 *
+	 * @param catalogManager The {@link CatalogManager} to use for storing and looking up {@link Table}s.
+	 * @param tableConfig The configuration of the TableEnvironment.
+	 * @param executionEnvironment The {@link StreamExecutionEnvironment} of the TableEnvironment.
+	 */
+	public static StreamTableEnvironmentImpl create(
```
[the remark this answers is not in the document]
dawidwys: I tried to "move/remove" most of them. I had to leave one here, as we need a common root for both: `create(StreamExec, TableConfig)` & `create(StreamExec, EnvSettings)`

flink-table/flink-table-api-scala-bridge/src/main/scala/org/apache/flink/table/api/scala/StreamTableEnvironment.scala
```diff
@@ -247,40 +247,50 @@ object StreamTableEnvironment {
...
+        .useAnyPlanner()
+        .inStreamMode()
+        .build())
   }
 
   /**
     * The [[TableEnvironment]] for a Scala [[StreamExecutionEnvironment]] that works with
     * [[DataStream]]s.
     *
-    * A TableEnvironment can be used to:
```
twalthr: Why did you remove this documentation?

dawidwys: It's duplicated on the class level, but I'm ok with reverting that.

dawidwys: I reverted that

twalthr:
Thank you @dawidwys. I will take care of merging this.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/factories/TableFactoryService.java
```diff
@@ -103,9 +106,52 @@
...
+	/**
+	 * Finds all table factories of the given class and property map.
+	 *
+	 * @param factoryClass desired factory class
+	 * @param propertyMap properties that describe the factory configuration
+	 * @param <T> factory class type
+	 * @return all the matching factories
+	 */
+	public static <T extends TableFactory> List<T> findAll(Class<T> factoryClass, Map<String, String> propertyMap) {
+		return findAllInternal(factoryClass, propertyMap, Optional.empty());
```
ssquan: Hi, should we use Flink user classloader instead?  I package BlinkExecutorFactory into a fat  JAR, but it can't be found by Flink cluster.

dawidwys: Hi @ssquan What is the exception you are getting? The `defaultLoader` actually already uses the user classloader. Could you make sure you have proper entries in the `META-INF/services/org.apache.flink.table.factories.TableFactory`. You might need a `https://maven.apache.org/plugins/maven-shade-plugin/examples/resource-transformers.html#ServicesResourceTransformer` if you have services coming from multiple dependencies.

PR #8857, opened 2019-06-24
[FLINK-12960] Introduce ShuffleDescriptor#ReleaseType and ShuffleDescriptor#getSupportedReleaseTypes

azagrebin:

`ResultPartitionDeploymentDescriptor#releasedOnConsumption` shows the intention how the partition is going to be used by the shuffle user and released. The `ShuffleDescriptor` should provide a way to query which release type is supported by shuffle service for this partition. If the requested release type is not supported by the shuffle service for a certain type of partition, the job should fail fast.

## Brief change log

  - Introduce `ShuffleDescriptor#ReleaseType.AUTO and MANUAL`
  - Introduce `ShuffleDescriptor#getSupportedReleaseTypes`
  - Add assertion whether the auto-release on consumption or manual are supported for the requested `ResultPartitionDeploymentDescriptor#releaseType`
  - adjust tests add test for the assertion

## Verifying this change

Simple refactoring covered by existing tests, plus `ResultPartitionDeploymentDescriptorTest#testIncompatibleReleaseTypeManual`

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/deployment/ResultPartitionDeploymentDescriptor.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/deployment/ResultPartitionDeploymentDescriptor.java
@@ -24,6 +24,7 @@ import org.apache.flink.runtime.jobgraph.IntermediateDataSetID;
 import org.apache.flink.runtime.jobgraph.IntermediateResultPartitionID;
 import org.apache.flink.runtime.shuffle.PartitionDescriptor;
 import org.apache.flink.runtime.shuffle.ShuffleDescriptor;
+import org.apache.flink.runtime.shuffle.ShuffleDescriptor.ReleaseType;
 import org.apache.flink.runtime.shuffle.ShuffleEnvironment;
 import org.apache.flink.runtime.shuffle.ShuffleMaster;
 import org.apache.flink.runtime.state.KeyGroupRangeAssignment;
@@ -31,6 +32,7 @@ import org.apache.flink.runtime.state.KeyGroupRangeAssignment;
 import java.io.Serializable;
 import java.util.Collection;
 
+import static org.apache.flink.util.Preconditions.checkArgument;
 import static org.apache.flink.util.Preconditions.checkNotNull;
 
 /**
@@ -51,20 +53,49 @@ public class ResultPartitionDeploymentDescriptor implements Serializable {
 	/** Flag whether the result partition should send scheduleOrUpdateConsumer messages. */
 	private final boolean sendScheduleOrUpdateConsumersMessage;
 
-	/** Whether the result partition is released on consumption. */
-	private final boolean releasedOnConsumption;
+	private final ReleaseType releaseType;
 
 	public ResultPartitionDeploymentDescriptor(
 			PartitionDescriptor partitionDescriptor,
 			ShuffleDescriptor shuffleDescriptor,
 			int maxParallelism,
 			boolean sendScheduleOrUpdateConsumersMessage) {
+		this(
+			checkNotNull(partitionDescriptor),
+			shuffleDescriptor,
+			maxParallelism,
+			sendScheduleOrUpdateConsumersMessage,
+			// Later we might have to make the scheduling adjust automatically
+			// if certain release type is not supported by shuffle service implementation at hand
+			partitionDescriptor.getPartitionType() == ResultPartitionType.BLOCKING ? ReleaseType.MANUAL : ReleaseType.AUTO);
+	}
+
+	public ResultPartitionDeploymentDescriptor(
+			PartitionDescriptor partitionDescriptor,
+			ShuffleDescriptor shuffleDescriptor,
+			int maxParallelism,
+			boolean sendScheduleOrUpdateConsumersMessage,
+			ReleaseType releaseType) {
+		checkReleaseOnConsumptionIsSupportedForPartition(shuffleDescriptor, releaseType);
 		this.partitionDescriptor = checkNotNull(partitionDescriptor);
 		this.shuffleDescriptor = checkNotNull(shuffleDescriptor);
 		KeyGroupRangeAssignment.checkParallelismPreconditions(maxParallelism);
 		this.maxParallelism = maxParallelism;
 		this.sendScheduleOrUpdateConsumersMessage = sendScheduleOrUpdateConsumersMessage;
-		this.releasedOnConsumption = partitionDescriptor.getPartitionType() != ResultPartitionType.BLOCKING;
+		this.releaseType = releaseType;
+	}
+
+	private static void checkReleaseOnConsumptionIsSupportedForPartition(
+			ShuffleDescriptor shuffleDescriptor,
+			ReleaseType releaseType) {
+		checkNotNull(shuffleDescriptor);
+		checkArgument(
+			shuffleDescriptor.getSupportedReleaseTypes().contains(releaseType),
+			"Release type %s is not supported by the shuffle service for this partition" +
+				"(id: %s), supported release types: %s",
+			releaseType,
+			shuffleDescriptor.getResultPartitionID(),
+			shuffleDescriptor.getSupportedReleaseTypes());
 	}
 
 	public IntermediateDataSetID getResultId() {
@@ -103,10 +134,15 @@ public class ResultPartitionDeploymentDescriptor implements Serializable {
 	 * by {@link ShuffleMaster#releasePartitionExternally(ShuffleDescriptor)}
 	 * and {@link ShuffleEnvironment#releasePartitionsLocally(Collection)}.
 	 *
+	 * <p>The partition has to support the corresponding {@link ReleaseType} in
+	 * {@link ShuffleDescriptor#getSupportedReleaseTypes()}:
+	 * {@link ReleaseType#AUTO} for {@code isReleasedOnConsumption()} to return {@code true} and
+	 * {@link ReleaseType#MANUAL} for {@code isReleasedOnConsumption()} to return {@code false}.
+	 *
 	 * @return whether to release the partition after having been fully consumed once.
 	 */
 	public boolean isReleasedOnConsumption() {
-		return releasedOnConsumption;
+		return releaseType == ReleaseType.AUTO;
 	}
 
 	@Override
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleDescriptor.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleDescriptor.java
@@ -24,6 +24,7 @@ import org.apache.flink.runtime.io.network.partition.ResultPartitionID;
 
 import java.io.Serializable;
 import java.util.Collection;
+import java.util.EnumSet;
 import java.util.Optional;
 
 /**
@@ -67,4 +68,32 @@ public interface ShuffleDescriptor extends Serializable {
 	 * @return the resource id of the producing task executor if the partition occupies local resources there
 	 */
 	Optional<ResourceID> storesLocalResourcesOn();
+
+	/**
+	 * Return release types supported by Shuffle Service for this partition.
+	 */
+	EnumSet<ReleaseType> getSupportedReleaseTypes();
+
+	/**
+	 * Partition release type.
+	 */
+	enum ReleaseType {
+		/**
+		 * Auto-release the partition after having been fully consumed once.
+		 *
+		 * <p>No additional actions required, like {@link ShuffleMaster#releasePartitionExternally(ShuffleDescriptor)}
+		 * or {@link ShuffleEnvironment#releasePartitionsLocally(Collection)}
+		 */
+		AUTO,
+
+		/**
+		 * Manually release the partition, the partition has to support consumption multiple times.
+		 *
+		 * <p>The partition requires manual release once all consumption is done:
+		 * {@link ShuffleMaster#releasePartitionExternally(ShuffleDescriptor)} and
+		 * if the partition occupies producer local resources ({@link #storesLocalResourcesOn()}) then also
+		 * {@link ShuffleEnvironment#releasePartitionsLocally(Collection)}.
+		 */
+		MANUAL
+	}
 }
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleMaster.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleMaster.java
@@ -19,6 +19,7 @@
 package org.apache.flink.runtime.shuffle;
 
 import org.apache.flink.runtime.deployment.ResultPartitionDeploymentDescriptor;
+import org.apache.flink.runtime.shuffle.ShuffleDescriptor.ReleaseType;
 
 import java.util.Collection;
 import java.util.concurrent.CompletableFuture;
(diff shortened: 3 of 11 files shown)


azagrebin:
@flinkbot attention @zentol @zhijiangW

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/NettyShuffleUtils.java
```diff
@@ -0,0 +1,50 @@
...
+		checkReleaseOnConsumptionIsSupportedForPartition(
+			resultPartitionDeploymentDescriptor.isReleasedOnConsumption(),
+			resultPartitionDeploymentDescriptor.getPartitionType().isBlocking());
+	}
+
+	public static void checkReleaseOnConsumptionIsSupportedForPartition(
+			boolean isReleasedOnConsumption,
+			boolean isBlockingPartition) {
+		Preconditions.checkArgument(
+			!isReleasedOnConsumption || isBlockingPartition,
```
zentol: Long-term we probably want to change this a bit; this kind of setup forces users of the shuffle service (i.e. the ones creating a PartitionDescriptor) to be fully aware of it's capabilities.

Basically, both the partition and shuffle descriptor describe an is-state; this partition must be released on consumption, and we expect the shuffle service to fail if it can't fulfill that.

I'd like to see this changed at some point so that the partition descriptor (or some other form of input) describes a wish instead, to facilitate something like:

Scheduler: "hey shuffle service, if you could persist this partition, that would be great."
ShuffleService: "Sorry kiddo, this one must be released on consumption."
Scheduler: "oh alright then, I'll just adjust my scheduling."

The usefulness of this can be easily shown with a simple though experiment: assume there exists a shuffle service implementation that supports pipelined partitions that are consumable multiple times.
Currently, it would be impossible for the runtime to ever make use of this flag, so long as we don't resort to type checks.

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/NettyShuffleUtils.java
```diff
@@ -0,0 +1,50 @@
...
+			"Auto-release on consumption is supported only for blocking partition types");
```
zentol: I'd invert this to say "Auto-release on consumption must not be disabled for pipelined partition types." The current message is a bit confusing; after all auto-release on consumption is _effectively_ supported for pipelined partitions.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NettyShuffleEnvironment.java
```diff
@@ -178,6 +179,7 @@ public void releasePartitionsLocally(Collection<ResultPartitionID> partitionIds)
 			ResultPartition[] resultPartitions = new ResultPartition[resultPartitionDeploymentDescriptors.size()];
 			int counter = 0;
 			for (ResultPartitionDeploymentDescriptor rpdd : resultPartitionDeploymentDescriptors) {
+				checkReleaseOnConsumptionIsSupportedForPartition(rpdd);
```
zentol: have you considered moving this into the factory instead?

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/NettyShuffleUtils.java
```diff
@@ -0,0 +1,50 @@
...
+import org.apache.flink.runtime.deployment.ResultPartitionDeploymentDescriptor;
+import org.apache.flink.util.Preconditions;
+
+/**
+ * Common utility methods for default Netty implementation of Shuffle Service.
+ */
+public enum NettyShuffleUtils {
+	;
+
+	public static void checkReleaseOnConsumptionIsSupportedForPartition(PartitionDescriptor partitionDescriptor) {
```
zentol: missing tests

flink-runtime/src/main/java/org/apache/flink/runtime/deployment/ResultPartitionDeploymentDescriptor.java
```diff
@@ -59,12 +61,42 @@ public ResultPartitionDeploymentDescriptor(
...
+	}
+
+	private static void checkReleaseOnConsumptionIsSupportedForPartition(
+			ShuffleDescriptor shuffleDescriptor,
+			boolean releasedOnConsumption) {
+		checkArgument(
+			releasedOnConsumption ?
+				shuffleDescriptor.getSupportedReleaseTypes().contains(ReleaseType.AUTO) :
+				shuffleDescriptor.getSupportedReleaseTypes().contains(ReleaseType.MANUAL),
+			"Release on consumption <%s> is not supported by the shuffle service for this partition, " +
```
zentol: let's include the result partition ID as well

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleDescriptor.java
```diff
@@ -67,4 +68,31 @@ default boolean isUnknown() {
...
+		 * <p>No additional actions required, like {@link ShuffleMaster#releasePartitionExternally(ShuffleDescriptor)}
+		 * or {@link ShuffleEnvironment#releasePartitionsLocally(Collection)}
+		 */
+		AUTO,
+
+		/**
+		 * Manually release the partition, the partition has to support consumption multiple times.
+		 *
+		 * <p>The partition requires manual actions to release it once all consumption is done:
+		 * {@link ShuffleMaster#releasePartitionExternally(ShuffleDescriptor)} and {@link ShuffleEnvironment#releasePartitionsLocally(Collection)}
```
zentol: this is slight ambiguous as it isn't whether the local resource condition applies to just the ShuffleEnvironment call or both of them.
I'd suggest something along the lines of `ShuffleMaster[...] and, if [...], ShuffleEnvironment [...]`

flink-runtime/src/main/java/org/apache/flink/runtime/deployment/ResultPartitionDeploymentDescriptor.java
```diff
@@ -59,12 +61,42 @@ public ResultPartitionDeploymentDescriptor(
...
 			int maxParallelism,
 			boolean sendScheduleOrUpdateConsumersMessage) {
+		this(
+			partitionDescriptor,
+			shuffleDescriptor,
+			maxParallelism,
+			sendScheduleOrUpdateConsumersMessage,
+			// Later we might have to make the scheduling adjust automatically
+			// if certain release type is not supported by shuffle service implementation at hand
+			partitionDescriptor.getPartitionType() != ResultPartitionType.BLOCKING);
```
zhijiangW: I think in future the `releasedOnConsumption` could be refactored into one `ReleaseStrategy` as proposed in #8804 , then it could be configured which strategy is used in practice and check whether this config is consistent with shuffle implementation.

flink-runtime/src/main/java/org/apache/flink/runtime/deployment/ResultPartitionDeploymentDescriptor.java
```diff
@@ -59,12 +61,42 @@ public ResultPartitionDeploymentDescriptor(
...
+	}
+
+	public ResultPartitionDeploymentDescriptor(
+			PartitionDescriptor partitionDescriptor,
+			ShuffleDescriptor shuffleDescriptor,
+			int maxParallelism,
+			boolean sendScheduleOrUpdateConsumersMessage,
+			boolean releasedOnConsumption) {
+		checkReleaseOnConsumptionIsSupportedForPartition(shuffleDescriptor, releasedOnConsumption);
```
zhijiangW: `checkNotNull(partitionDescriptor)` should be put here.

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleDescriptor.java
```diff
@@ -67,4 +68,31 @@ default boolean isUnknown() {
 	 * @return the resource id of the producing task executor if the partition occupies local resources there
 	 */
 	Optional<ResourceID> storesLocalResourcesOn();
+
+	/**
+	 * Return release types supported by Shuffle Service for this partition.
```
zhijiangW: remove `for this partition`.
`by this shuffle service`.

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleMaster.java
```diff
@@ -51,6 +52,7 @@
 	 * <p>This call triggers release of any resources which are occupied by the given partition in the external systems
 	 * outside of the producer executor. This is mostly relevant for the batch jobs and blocking result partitions.
 	 * This method is not called if {@link ResultPartitionDeploymentDescriptor#isReleasedOnConsumption()} is {@code true}.
+	 * The partition has to support the {@link ReleaseType#MANUAL} in {@link ShuffleDescriptor#getSupportedReleaseTypes()}.
```
zhijiangW: In theory different `ShuffleManager` implementations could choose  which release strategy is supported, we should not limit all the implementations must support one type, otherwise we could make this limitation reflected in interface methods.

flink-runtime/src/main/java/org/apache/flink/runtime/deployment/ResultPartitionDeploymentDescriptor.java
azagrebin: True, the PR just leaves it as it was before at the moment

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleMaster.java
azagrebin: The goal of PR is to let shuffle services decide what to support in `ShuffleDescriptor#getSupportedReleaseTypes`. This doc comment line is related to when the manual release is possible (releasePartitionLocally/releasePartitionExternally).

Or do you mean, that we implicitly assume at the moment that `ReleaseType#MANUAL` has to be always supported for blocking partitions? For this, we have assertion `ResultPartitionDeploymentDescriptor#checkReleaseOnConsumptionIsSupportedForPartition` which will fail the job fast. It is true, we have to make scheduling/release strategies automatically adjust later if something is not supported as written in the comment to the hardcoded `releasedOnConsumption = partitionDescriptor.getPartitionType() != ResultPartitionType.BLOCKING`.

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleDescriptor.java
azagrebin: why? if we leave `by this shuffle service` then it might sound like for all partitions but this method returns supported release types for certain partition.

azagrebin:
Thanks for the review @zentol @zhijiangW ! I updated the PR.

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleDescriptor.java
zhijiangW: There might exist two options:
- One `ShuffleService` implementation might provide different `ShuffleDescriptor` implementations for different partitions with different release types. Then the previous comment makes sense.

- One `ShuffleService` implementation would only have one kind of `ShuffleDescriptor` and provide one enum of release types for all the partitions. Just like atm we only have `NettyShuffleDescriptor` implementation which provides the same release types for all partitions. So my above comment suggestion was based on this. The release type is `ShuffleService` global level suitable for all partitions.

zhijiangW: There might exist two options:

One ShuffleService implementation might provide different ShuffleDescriptor implementations for different partitions with different release types. Then the previous comment makes sense.

One ShuffleService implementation would only have one kind of ShuffleDescriptor and provide one enum of release types for all the partitions. Just like atm we only have NettyShuffleDescriptor implementation which provides the same release types for all partitions. So my above comment suggestion was based on this. The release type is ShuffleService global level suitable for all partitions.

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleMaster.java
zhijiangW: Sorry I have not noticed that this javadoc change was for the method of `releasePartitionsExternally`, if so I agree it makes sense that this method implementation is  for MANUL release.
Before I thought the new info was added on the class javadoc, that means all the `ShuffleMaster` implementation must support MANUL release type, so I pointed out that concern.

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleDescriptor.java
zentol: The `NettyShuffleDescriptor` dopes not have the same release types for all partitions. For non-blocking partitions we do not support `ReleaseType.MANUAL`.

flink-runtime/src/test/java/org/apache/flink/runtime/deployment/ResultPartitionDeploymentDescriptorTest.java
```diff
@@ -72,13 +72,9 @@
...
 	@Test
-	public void testSerializationWithUnknownShuffleDescriptor() throws Exception {
+	public void testSerializationOfUnknownShuffleDescriptor() throws Exception {
 		ShuffleDescriptor shuffleDescriptor = new UnknownShuffleDescriptor(resultPartitionID);
-
-		ResultPartitionDeploymentDescriptor copy =
-			createCopyAndVerifyResultPartitionDeploymentDescriptor(shuffleDescriptor);
-
-		ShuffleDescriptor shuffleDescriptorCopy = copy.getShuffleDescriptor();
+		ShuffleDescriptor shuffleDescriptorCopy = CommonTestUtils.createCopySerializable(shuffleDescriptor);
```
zhijiangW: this change should be a separate hotfix

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/ResultPartitionFactoryTest.java
```diff
@@ -65,7 +65,7 @@ private static void testForceConsumptionOnRelease(boolean forceConsumptionOnRele
 				ResultPartitionType.BLOCKING,
 				1,
 				0),
-			NettyShuffleDescriptorBuilder.newBuilder().buildLocal(),
+			NettyShuffleDescriptorBuilder.newBuilder().setBlocking(true).buildLocal(),
```
zhijiangW: might be proper for using `partitionType.isBlocking()` instead of direct `true`

flink-runtime/src/test/java/org/apache/flink/runtime/util/NettyShuffleDescriptorBuilder.java
```diff
@@ -73,19 +74,26 @@ public NettyShuffleDescriptorBuilder setConnectionIndex(int connectionIndex) {
 		return this;
 	}
 
+	public NettyShuffleDescriptorBuilder setBlocking(boolean blocking) {
```
zhijiangW: `setIsBlocking(boolean isBlocking)` ?

flink-runtime/src/main/java/org/apache/flink/runtime/deployment/ResultPartitionDeploymentDescriptor.java
```diff
@@ -103,6 +137,9 @@ public boolean sendScheduleOrUpdateConsumersMessage() {
 	 * by {@link ShuffleMaster#releasePartitionExternally(ShuffleDescriptor)}
 	 * and {@link ShuffleEnvironment#releasePartitionsLocally(Collection)}.
 	 *
+	 * <p>The partition has to support the corresponding {@link ReleaseType} in
```
zhijiangW: I think it might be better to indicate which value true/false should support which `releaseType`, otherwise it seems not clear the mapping between this value and release type.

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleDescriptor.java
zhijiangW: Yes,  I have no other concerns here now.

zhijiangW:
@azagrebin I have no other concerns except some nit comments.

PR #8858, opened 2019-06-24
[hotfix][tests] Change some StreamTask tests to create a test task in the task's thread

1u0:

The changes introduced in #8692 PR have an invariant, that the `StreamTask`'s mailbox loop is running in the same thread that set in the `taskMailboxExecutor`. Currently, the thread reference is set in `StreamTask` instantiation time, this conflicts with some `StreamTask` tests, as they were creating the task instance in a different (non task's) thread.

## Brief change log

  - `StreamTaskTest` is rewritten to properly run a task under test, instead of overriding some fields via `Whitebox`;
  - the rest of tests in `flink-streaming-java` are rewritten to delegate task instance creation to the task's main thread.

## Verifying this change

Running the tests in `flink-streaming-java` and `flink-streaming-scala` should be green. Also, running the same tests with changes in #8692 should be green.

## Documentation

### DIFF
--- a/flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/StreamTaskTest.java
+++ b/flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/StreamTaskTest.java
@@ -20,13 +20,10 @@ package org.apache.flink.streaming.runtime.tasks;
 
 import org.apache.flink.api.common.ExecutionConfig;
 import org.apache.flink.api.common.JobID;
-import org.apache.flink.api.common.TaskInfo;
 import org.apache.flink.api.common.typeutils.TypeSerializer;
 import org.apache.flink.configuration.CheckpointingOptions;
 import org.apache.flink.configuration.Configuration;
-import org.apache.flink.core.fs.CloseableRegistry;
 import org.apache.flink.core.testutils.OneShotLatch;
-import org.apache.flink.mock.Whitebox;
 import org.apache.flink.runtime.blob.BlobCacheService;
 import org.apache.flink.runtime.blob.PermanentBlobCache;
 import org.apache.flink.runtime.blob.TransientBlobCache;
@@ -41,6 +38,7 @@ import org.apache.flink.runtime.checkpoint.TaskStateSnapshot;
 import org.apache.flink.runtime.clusterframework.types.AllocationID;
 import org.apache.flink.runtime.deployment.InputGateDeploymentDescriptor;
 import org.apache.flink.runtime.deployment.ResultPartitionDeploymentDescriptor;
+import org.apache.flink.runtime.execution.CancelTaskException;
 import org.apache.flink.runtime.execution.Environment;
 import org.apache.flink.runtime.execution.ExecutionState;
 import org.apache.flink.runtime.execution.librarycache.LibraryCacheManager;
@@ -53,7 +51,6 @@ import org.apache.flink.runtime.io.network.NettyShuffleEnvironmentBuilder;
 import org.apache.flink.runtime.io.network.TaskEventDispatcher;
 import org.apache.flink.runtime.io.network.partition.NoOpResultPartitionConsumableNotifier;
 import org.apache.flink.runtime.io.network.partition.ResultPartitionConsumableNotifier;
-import org.apache.flink.runtime.io.network.partition.ResultPartitionManager;
 import org.apache.flink.runtime.jobgraph.JobVertexID;
 import org.apache.flink.runtime.jobgraph.OperatorID;
 import org.apache.flink.runtime.jobgraph.tasks.AbstractInvokable;
@@ -67,7 +64,6 @@ import org.apache.flink.runtime.query.KvStateRegistry;
 import org.apache.flink.runtime.shuffle.ShuffleEnvironment;
 import org.apache.flink.runtime.state.AbstractKeyedStateBackend;
 import org.apache.flink.runtime.state.AbstractStateBackend;
-import org.apache.flink.runtime.state.CheckpointStorage;
 import org.apache.flink.runtime.state.CheckpointStreamFactory;
 import org.apache.flink.runtime.state.DoneFuture;
 import org.apache.flink.runtime.state.KeyGroupStatePartitionStreamProvider;
@@ -83,7 +79,6 @@ import org.apache.flink.runtime.state.TaskLocalStateStoreImpl;
 import org.apache.flink.runtime.state.TaskStateManager;
 import org.apache.flink.runtime.state.TaskStateManagerImpl;
 import org.apache.flink.runtime.state.TestTaskStateManager;
-import org.apache.flink.runtime.state.memory.MemoryBackendCheckpointStorage;
 import org.apache.flink.runtime.state.memory.MemoryStateBackend;
 import org.apache.flink.runtime.taskexecutor.KvStateService;
 import org.apache.flink.runtime.taskexecutor.PartitionProducerStateChecker;
@@ -93,7 +88,6 @@ import org.apache.flink.runtime.taskmanager.NoOpTaskManagerActions;
 import org.apache.flink.runtime.taskmanager.Task;
 import org.apache.flink.runtime.taskmanager.TaskExecutionState;
 import org.apache.flink.runtime.taskmanager.TaskManagerActions;
-import org.apache.flink.runtime.testingUtils.TestingUtils;
 import org.apache.flink.runtime.util.TestingTaskManagerRuntimeInfo;
 import org.apache.flink.streaming.api.TimeCharacteristic;
 import org.apache.flink.streaming.api.functions.source.SourceFunction;
@@ -112,9 +106,12 @@ import org.apache.flink.streaming.runtime.streamstatus.StreamStatusMaintainer;
 import org.apache.flink.util.CloseableIterable;
 import org.apache.flink.util.SerializedValue;
 import org.apache.flink.util.TestLogger;
+import org.apache.flink.util.function.SupplierWithException;
 
 import org.junit.Assert;
+import org.junit.Rule;
 import org.junit.Test;
+import org.junit.rules.Timeout;
 import org.junit.runner.RunWith;
 import org.mockito.ArgumentCaptor;
 import org.mockito.invocation.InvocationOnMock;
@@ -130,20 +127,19 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.CompletionException;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executor;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.RunnableFuture;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicReference;
 
-import static org.apache.flink.runtime.concurrent.Executors.newDirectExecutorService;
+import static org.apache.flink.streaming.util.StreamTaskUtil.waitTaskIsRunning;
 import static org.hamcrest.Matchers.everyItem;
 import static org.hamcrest.Matchers.greaterThanOrEqualTo;
 import static org.hamcrest.Matchers.hasSize;
 import static org.hamcrest.Matchers.instanceOf;
+import static org.hamcrest.Matchers.is;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertThat;
@@ -152,7 +148,6 @@ import static org.junit.Assert.fail;
 import static org.mockito.ArgumentMatchers.nullable;
 import static org.mockito.Matchers.any;
 import static org.mockito.Matchers.anyLong;
-import static org.mockito.Matchers.anyString;
 import static org.mockito.Matchers.eq;
 import static org.mockito.Mockito.doAnswer;
 import static org.mockito.Mockito.mock;
@@ -173,6 +168,9 @@ import static org.powermock.api.mockito.PowerMockito.whenNew;
 public class StreamTaskTest extends TestLogger {
 
 	private static OneShotLatch syncLatch;
+
+	@Rule
+	public final Timeout timeoutPerTest = Timeout.seconds(30);
 
 	/**
 	 * This test checks that cancel calls that are issued before the operator is
@@ -311,21 +309,7 @@ public class StreamTaskTest extends TestLogger {
 
 	@Test
 	public void testDecliningCheckpointStreamOperator() throws Exception {
-		final long checkpointId = 42L;
-		final long timestamp = 1L;
-
-		TaskInfo mockTaskInfo = mock(TaskInfo.class);
-		when(mockTaskInfo.getTaskNameWithSubtasks()).thenReturn("foobar");
-		when(mockTaskInfo.getIndexOfThisSubtask()).thenReturn(0);
 		CheckpointExceptionHandlerTest.DeclineDummyEnvironment declineDummyEnvironment = new CheckpointExceptionHandlerTest.DeclineDummyEnvironment();
-
-		StreamTask<?, ?> streamTask = new EmptyStreamTask(declineDummyEnvironment);
-		CheckpointMetaData checkpointMetaData = new CheckpointMetaData(checkpointId, timestamp);
-
-		// mock the operators
-		StreamOperator<?> streamOperator1 = mock(StreamOperator.class);
-		StreamOperator<?> streamOperator2 = mock(StreamOperator.class);
-		StreamOperator<?> streamOperator3 = mock(StreamOperator.class);
 
 		// mock the returned snapshots
 		OperatorSnapshotFutures operatorSnapshotResult1 = mock(OperatorSnapshotFutures.class);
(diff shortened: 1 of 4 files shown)


1u0:
CC @StefanRRichter, @pnowojski.

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/StreamTaskTest.java
```diff
@@ -805,29 +723,102 @@ public void testSetsUserCodeClassLoaderForTimerThreadFactory() throws Throwable
...
+	private static class RunningTask<T> {
+		final T streamTask;
+		final CompletableFuture<Void> invocationFuture;
+
+		RunningTask(T streamTask, CompletableFuture<Void> invocationFuture) {
+			this.streamTask = streamTask;
+			this.invocationFuture = invocationFuture;
+		}
+
+		void waitForTaskCompletion(boolean cancelled) throws Exception {
```
pnowojski: I would split this method into two along `boolean cancelled` parameter: `waitForTaskCompletion(false)` -> `waitForTaskCompletion()` and `waitForTaskCompletion(true)` -> `waitForCancelledTaskCompletion()`. They do not share any code and you are not using this parameter in any generic way (it's always a constant). Also by splitting into two you can deduplicate the code by a little bit:
```
waitForCancelledTaskCompletion() {
  try {
    waitForTaskCompletion();
  } catch (...) {
     assertThat(...);
  }
}
```
}

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/StreamTaskTest.java
```diff
@@ -805,29 +723,102 @@ public void testSetsUserCodeClassLoaderForTimerThreadFactory() throws Throwable
...
+				T task;
+				try {
+					task = taskFactory.get();
+					taskFuture.complete(task);
+				} catch (Exception e) {
+					taskFuture.completeExceptionally(e);
+					return;
+				}
+				try {
+					task.invoke();
```
pnowojski: why failures here are not being forwarded to `taskFuture.completeExceptionally(e);`?

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/StreamTaskTest.java
```diff
@@ -805,29 +723,102 @@ public void testSetsUserCodeClassLoaderForTimerThreadFactory() throws Throwable
...
+					throw e;
+				} catch (Exception e) {
+					throw new RuntimeException(e);
+				}
+			}, Executors.newSingleThreadExecutor());
+
+		return new RunningTask<>(taskFuture.get(), invocationFuture);
+	}
+
+	private static void waitForTaskCompletion(CompletableFuture<Void> taskInvocation, boolean cancelled) throws Exception {
```
pnowojski: Is this method used anywhere?

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/StreamTaskTest.java
```diff
@@ -1023,25 +1014,63 @@ protected AbstractStateBackend createInnerBackend(Configuration config) {
 	// ------------------------------------------------------------------------
 	// ------------------------------------------------------------------------
 
+	private enum Event {
+		TASK_IS_RUNNING,
+	}
+
 	private static class EmptyStreamTask extends StreamTask<String, AbstractStreamOperator<String>> {
```
pnowojski: I think this is no longer an "empty" `StreamTask` so it should be renamed. Maybe to `AwaitableStreamTask`? `MockStreamTask`? `TestingStreamTask`?

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/StreamTaskTest.java
```diff
@@ -1023,25 +1014,63 @@ protected AbstractStateBackend createInnerBackend(Configuration config) {
...
+			if (overrideOperatorChain != null) {
+				super.operatorChain = this.overrideOperatorChain;
+				super.headOperator = super.operatorChain.getHeadOperator();
+			}
+		}
 
 		@Override
-		protected void performDefaultAction(ActionContext context) throws Exception {
-			context.allActionsCompleted();
+		protected void performDefaultAction(ActionContext context) {
```
pnowojski: I'm not sure if overriding this method is a good way to perform the test checks. I think it might be better to rebase this on top of
https://github.com/apache/flink/pull/8826
and implement `sourceFinished` and `emitEvent(Event.TASK_IS_RUNNING);` as an `StreamInputProcessor` there.

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/StreamTaskTest.java
```diff
@@ -311,71 +310,45 @@ public void testCancellationFailsWithBlockingLock() throws Exception {
...
-		Whitebox.setInternalState(streamTask, "asynchronousCheckpointExceptionHandler", asyncCheckpointExceptionHandler);
+			verify(operatorSnapshotResult1).cancel();
+			verify(operatorSnapshotResult2).cancel();
 
-		try {
-			streamTask.triggerCheckpoint(checkpointMetaData, CheckpointOptions.forCheckpointWithDefaultLocation(), false);
-			fail("Expected test exception here.");
-		} catch (Exception e) {
-			assertEquals(testException, e.getCause());
+			task.streamTask.setSourceFinished();
```
pnowojski: Why are you calling `setSourceFinished` only 2 times? In other places task is failing exceptionally/being cancelled? 

Also I think better name would be "input" not "source".

If so maybe rename the method to `finishInput()`? `finishTaskGracefully()`? Or at least add a java doc explaining what this method is doing.

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/StreamTaskTestHarness.java
```diff
@@ -288,31 +290,24 @@ public void waitForTaskRunning() throws Exception {
 	 * @throws Exception
 	 */
 	public void waitForTaskRunning(long timeout) throws Exception {
-		if (taskThread == null) {
+		if (taskThread == null || taskThread.task == null) {
```
pnowojski: nit: change to `checkState`?

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/StreamTaskTestHarness.java
```diff
@@ -232,15 +231,18 @@ public Thread invoke() throws Exception {
...
+		checkState(this.taskThread == null);
 		this.mockEnv = checkNotNull(mockEnv);
 
 		initializeInputs();
 		initializeOutput();
 
-		this.task = taskFactory.apply(mockEnv);
-
-		taskThread = new TaskThread(task);
+		taskThread = new TaskThread(() -> taskFactory.apply(mockEnv));
```
pnowojski: I'm missing the information why do we need this change. Could you explain more in the commit message and maybe optionally somewhere in the code as a comment?

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/StreamTaskTest.java
```diff
@@ -1023,25 +1014,63 @@ protected AbstractStateBackend createInnerBackend(Configuration config) {
...
+		EmptyStreamTask(Environment env, LinkedBlockingQueue<Event> eventQueue, OperatorChain<String, AbstractStreamOperator<String>> operatorChain) {
 			super(env, null);
+			this.eventQueue = eventQueue;
+			this.overrideOperatorChain = operatorChain;
 		}
 
 		@Override
-		protected void init() throws Exception {}
+		protected void init() {
+			if (overrideOperatorChain != null) {
```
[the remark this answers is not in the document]
1u0: This is leftover, while I was transitioning and rewriting tests from the old constructor.

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/StreamTaskTest.java
```diff
@@ -1023,25 +1014,63 @@ protected AbstractStateBackend createInnerBackend(Configuration config) {
...
-		public EmptyStreamTask(Environment env) {
+		private boolean isRunningFoo;
+		private final LinkedBlockingQueue<Event> eventQueue;
+		private final OperatorChain<String, AbstractStreamOperator<String>> overrideOperatorChain;
+		private volatile boolean sourceFinished;
+
+		EmptyStreamTask(Environment env, LinkedBlockingQueue<Event> eventQueue, OperatorChain<String, AbstractStreamOperator<String>> operatorChain) {
 			super(env, null);
+			this.eventQueue = eventQueue;
+			this.overrideOperatorChain = operatorChain;
```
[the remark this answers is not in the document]
1u0: The `operatorChain` is re-set here in such strange way, exactly because of reasons you have listed
and it's needed for the tests, that we actually running proper `StreamTask.invoke()` (and `run()`).

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/StreamTaskTest.java
1u0: My interpretation, that this task was called "empty" due to not emitting/producing anything.

But I haven't considered renaming it, to keep changes minimum. I can rename it if that not an issue.

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/StreamTaskTest.java
1u0: The idea that `taskFuture` is used here only to synchronize until a task instance have been created.
Note, that `runTask` doesn't return until we pass that point.

The rest of execution should be tracked by `invocationFuture`.

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/StreamTaskTest.java
1u0: This is just preserving old tests semantics as much as possible, although, it may not really matter for the provided tests.

+1 for renaming.

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/StreamTaskTest.java
```diff
@@ -805,29 +723,102 @@ public void testSetsUserCodeClassLoaderForTimerThreadFactory() throws Throwable
...
 		}
 	}
 
 	// ------------------------------------------------------------------------
 	//  Test Utilities
 	// ------------------------------------------------------------------------
 
+	private static StreamOperator<?> streamOperatorWithSnapshot(OperatorSnapshotFutures operatorSnapshotResult) throws Exception {
+		StreamOperator<?> operator = mock(StreamOperator.class);
+		when(operator.getOperatorID()).thenReturn(new OperatorID());
```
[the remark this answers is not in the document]
1u0: I can look into it, although I don't know how long it would take to properly refactor all those duplication.

Basically, there is a risk that this would unnecessary bloat this PR and take more time to refactor.

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/StreamTaskTest.java
pnowojski: I'm not sure. I think that `EmptyStreamTask` is not the best, but also I don't think that my proposal are much better. I would lean towards `AwaitableStreamTask` or `AwaitableEmptyStreamTask`, but if you don't like it, I'm fine with leaving it as it is.

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/StreamTaskTest.java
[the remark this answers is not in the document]
pnowojski: Ok, I see it now. It's pretty fragile setup and it boils down to kind of broken life cycle of the tasks and operators :( Of course it's still better compared to what it used to be.

The proper solution would be to refactor the lifecycle, so that the classes are initialised in the constructor, but that's beyond the scope. In that case maybe:
1. either leave a comment in your `#init()` method that we are overriding an `operatorChain` instance created in `StreamTask#invoke()`.
2. or maybe as a stop gap solution we could extract construction of an `OperatorChain` in the `StreamTask` to some overloadable method `StreamTask#createOperatorChain()`?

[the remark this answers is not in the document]
1u0: I have added a comment in the `EmptyStreamTask.init()` method.

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/StreamTaskTest.java
1u0: I have updated this code a little bit.

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/StreamTaskTest.java
1u0: I have changed the way how the tests are waiting for task to be running. That way it should not conflict with the future changes in your PR.

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/StreamTaskTest.java
```diff
@@ -805,29 +723,102 @@ public void testSetsUserCodeClassLoaderForTimerThreadFactory() throws Throwable
...
+
+		void waitForTaskCompletion(boolean cancelled) throws Exception {
+			if (!cancelled) {
+				invocationFuture.get();
+				return;
+			}
+			try {
+				invocationFuture.get();
+			} catch (Exception e) {
+				assertThat(e.getCause(), is(instanceOf(CancelTaskException.class)));
```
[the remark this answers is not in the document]
1u0: In this case, the `CancelTaskException` is the original exception that thrown by `task.invoke()`.
It's not expected to be wrapped by more exceptions.

The only wrapping exception comes from the `invocationFuture`.

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/StreamTaskTest.java
1u0: I have renamed to `MockStreamTask`.

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/StreamTaskTestHarness.java
1u0: I've added a commit message (as commit description).

pnowojski:
I tried to drop  the `mock(StreamOperator.class)` and `mock(OperatorChain.class)` calls, but it's not that easy :/ It would require more sophisticated lifecycle simulation in `StreamTaskTest` to make it work, so I'm dropping this issue for now.

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/StreamTaskTestHarness.java
```diff
@@ -231,15 +230,20 @@ public Thread invoke() throws Exception {
...
 		taskThread.start();
+		// Wait until the task is set
+		while (taskThread.task == null) {
+			Thread.sleep(10L);
```
StefanRRichter: Nit: Why not use a latch like `OneShotLatch` here to wait-block until the field is set and the thread then signals?

1u0: Can be done as you propose by using a `OneShotLatch`, but this would add a dedicated field in `TaskThread` in that case.

1u0:
CI fails on some other test:
```
01:36:14.687 [ERROR] testMigrationAndRestore[Migrate Savepoint: 1.2](org.apache.flink.test.state.operator.restore.unkeyed.ChainLengthIncreaseTest)  Time elapsed: 7.853 s  <<< ERROR!
java.util.concurrent.ExecutionException: java.util.concurrent.CompletionException: org.apache.flink.runtime.checkpoint.CheckpointException: Task received cancellation from one of its inputs
Caused by: java.util.concurrent.CompletionException: org.apache.flink.runtime.checkpoint.CheckpointException: Task received cancellation from one of its inputs
Caused by: org.apache.flink.runtime.checkpoint.CheckpointException: Task received cancellation from one of its inputs
Caused by: org.apache.flink.runtime.checkpoint.CheckpointException: Task received cancellation from one of its inputs
```

Another, unrelated PR with the same failure:
https://travis-ci.org/apache/flink/jobs/551265900

pnowojski:
Thanks. Indeed failure seems unrelated. I'm merging it.

PR #8859, opened 2019-06-24
[FLINK-12905][table-planner] Enable querying CatalogViews in legacy planner

dawidwys:

Enable querying catalog views in legacy planner

## Brief change log

  - Enabled table access conversion in SqlToRelConverter
  - Enabled `ExpandingTableScanFactory` in `FlinkRelBuilder` to use `FlinkPlannerImpl` for expanding views
  - Added `SqlDialect` to `CatalogView`
  - Added conversion from Flink's `CatalogView` to `FlinkViewTable`, where we check if the View's 
dialect is compatible with the dialect currently used in FlinkPlanner.
  - Added `TableConfig` to Calcite's `Context`

## Verifying this change

* org.apache.flink.table.catalog.ViewExpansionTest

## Documentation


dawidwys:
cc @bowenli86 @xuefuz @twalthr

dawidwys:
Hi @bowenli86 this is a good point. I will address it soon.

One somewhat related question. Does the hive catalog (not necessarily hive metastore) support cross catalog views now?

bowenli86:
> Hi @bowenli86 this is a good point. I will address it soon.
> 
> One somewhat related question. Does the hive catalog (not necessarily hive metastore) support cross catalog views now?

`HiveCatalog` itself doesn't have constraints on that, because views and their expanded query are just data passing thru `HiveCatalog` on read/write and `HiveCatalog` won't manipulate them, so it's totally up to the planner that whether it can parse and support those cross-catalog views. I'm not sure whether users would find cross-catalog views useful, maybe we can listen to user feedbacks after releasing 1.9? I don't think it need to happen in 1.9.

@lirui-apache can help and share how we build a `HiveAnalyzer` in Blink to support Hive view expanded query on view creation and parsing it upon read. Maybe you guys can talk offline on that.

dawidwys:
From the planner perspective I am sure it is possible. From the planner perspective catalog is just a single part of a schema. I was asking strictly about the HiveCatalog.

I think this is a really common use case. Usually you have schema registry for kafka topics and hive metastore for static/batch tables. I think it will be one of the most common use case to support e.g. joins between those catalogs.

But if you are saying that there is no problem to store a view that hive does not understand than it should be fine.

bowenli86:
cross-catalog view would be Flink generic view, so yes, `HiveCatalog` is able to read/write it, just that Hive cannot understand it from Hive side.

lirui-apache:
For Hive to access (non-generic) views created by Flink, we need to 1) remove the catalog part from the object paths and 2) make sure the expanded query is valid HiveQL syntax.
To achieve the above in Blink, we used Hive's analyzer to analyze the original query in order to get a Hive-compatible expanded query. Blink's expanded query is stored as a property of the view, and will be restored if the view is accessed from Blink side.

If we're to do the same in Flink, I guess it can be done in a separate PR. And the prerequisite is that `CatalogView` preserves the original query when it reaches `HiveCatalog::createTable`.

bowenli86:
Hi @dawidwys , is there any update on this?

bowenli86:
sorry, mistakenly closed this PR..... reopened it

dawidwys:
@twalthr Could you check the PR, when you have some time?

Also we should think how we can we share tests such as ViewExpansionTest or PathResolutionTest between planners. Currently the path resolution is not tested in the blink planner.

flink-table/flink-table-planner/src/main/java/org/apache/flink/table/planner/FlinkViewTable.java
```diff
@@ -0,0 +1,65 @@
...
+			List<String> viewPath) {
+		super(null, rowType, viewSql, schemaPath, viewPath);
+		this.viewDialect = checkNotNull(viewDialect);
+	}
+
+	@Override
+	public RelNode toRel(RelOptTable.ToRelContext context, RelOptTable relOptTable) {
+		TableConfig tableConfig = context.getCluster().getPlanner().getContext().unwrap(TableConfig.class);
+		SqlDialect sqlDialect = tableConfig.getSqlDialect();
+		if (!viewDialect.equals(sqlDialect)) {
```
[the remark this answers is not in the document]
lirui-apache: Perhaps a more general question is what're the implications of setting sql dialect. According to the JavaDoc of `SqlDialect`, HIVE dialect is required only if users want to use some Hive-specific features e.g. `INSERT OVERWRITE`. Enforcing a dialect check for views might confuse users. For example, users need to worry about dialect even if the query is common enough to be supported by both DEFAULT and HIVE dialects. Besides, it prevents mixed usage of generic and hive views, which doesn't seem very user friendly.

dawidwys:
After some discussions with @twalthr over @lirui-apache comment I think we should not limit the support of views to the ones that have a corresponding dialect to the one of SqlParser. In a longer run we should have a SqlDialect flag in a View definition, but it should be a self contained information independent of the configuration of the SqlParser. (E.g. add Hive specific operators when translating the view to a relational tree, reconfiguring/creating a parser compatible with the view etc.).

I reverted the PR to the state where we perform only best effort parsing of the view.

dawidwys:
@danny0405 @lirui-apache Could you have a look?

danny0405:
Thanks @dawidwys for the nice job, reviewing now ~

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/calcite/FlinkRelBuilder.scala
```diff
@@ -167,19 +165,9 @@ object FlinkRelBuilder {
 
   def proto(context: Context): RelBuilderFactory = new RelBuilderFactory() {
     def create(cluster: RelOptCluster, schema: RelOptSchema): RelBuilder = {
+      val clusterContext = cluster.getPlanner.getContext.unwrap(classOf[FlinkContext])
+      val mergedContext = Contexts.chain(clusterContext, context)
```
danny0405: Can we need to chain the contexts here ? Have a quick look at the `ChainContext#unwrap`, it seems that the `Contexts.chain(clusterContext, context)` always returns `clusterContext ` if it is not null, this logic is just inconsistent with the original, because the old code use `context` to unwrap.

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/catalog/DatabaseCalciteSchema.java
```diff
@@ -68,25 +68,28 @@ public Table getTable(String tableName) {
...
-				Catalog catalog = catalogManager.getCatalog(catalogName).get();
-				FlinkStatistic statistic = getStatistic(result.isTemporary(),
-					catalog, table, identifier);
+				FlinkStatistic statistic = getStatistic(result.isTemporary(), table, identifier);
 				return new CatalogSchemaTable(identifier,
 					table,
 					statistic,
-					catalog.getTableFactory().orElse(null),
+					catalogManager.getCatalog(catalogName)
+						.flatMap(Catalog::getTableFactory)
```
danny0405: Did you mean a temporal object could exist without a catalog ? For example, the view ?

flink-table/flink-table-planner/src/main/scala/org/apache/flink/table/calcite/FlinkPlannerImpl.scala
```diff
@@ -86,34 +84,43 @@ class FlinkPlannerImpl(
...
-    if (sqlNode.getKind.belongsTo(SqlKind.DDL)
-      || sqlNode.getKind == SqlKind.INSERT) {
-      return sqlNode
-    }
+    val validator = getOrCreateSqlValidator()
+    validateInternal(sqlNode, validator)
+  }
+
+  private def validateInternal(sqlNode: SqlNode, validator: FlinkCalciteSqlValidator): SqlNode = {
     try {
```
danny0405: Still name it `validate` ? Just like the 2 `rel` methods.

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/plan/schema/ExpandingPreparingTable.java
```diff
@@ -0,0 +1,73 @@
...
+		@Nullable RelOptSchema relOptSchema,
+		RelDataType rowType,
+		Iterable<String> names,
+		FlinkStatistic statistic) {
+		super(relOptSchema, rowType, names, statistic);
+	}
+
+	/**
+	 * Expands at least the most outer level of a relational tree. Does not need to expand
+	 * any nested scans of an {@link ExpandingPreparingTable}. Those will be expanded recursively.
```
danny0405: at least -> at most ?

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/catalog/SqlCatalogViewTable.java
```diff
@@ -0,0 +1,61 @@
...
+		super(relOptSchema, rowType, names, statistic);
+		this.view = view;
+		this.viewPath = viewPath;
+	}
+
+	@Override
+	public RelNode expandTopLevel(ToRelContext context) {
+		RelRoot root = context.expandView(rowType, view.getExpandedQuery(), viewPath, names);
+		return RelOptUtil.createCastRel(root.rel, rowType, true);
+	}
```
danny0405: We should avoid to introduce the redundant cast nodes, if you want to fix the nullability, just fix it from the source type inference.

danny0405:
Thanks @dawidwys , i have left some review comments, i'm glad that we can register the view through the catalog API now, but how about the SQL DDL ? It would be very nice if we can create a view through DDL: `tableEnv.sqlUpdate("create view as select ...")`.

flink-table/flink-table-planner-blink/src/test/scala/org/apache/flink/table/planner/plan/common/ViewsExpanding.scala
```diff
@@ -0,0 +1,117 @@
...
+
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.junit.runners.Parameterized
+import org.junit.runners.Parameterized.Parameters
+
+import java.util
+
+@RunWith(classOf[Parameterized])
+class ViewsExpanding(tableTestUtil: TableTestBase => TableTestUtil) extends TableTestBase {
```
lirui-apache: I thought UT class names need the "Test" suffix to be picked up by the surefire plugin, no?

lirui-apache:
@dawidwys Thanks for the update. +1 to have self contained dialect flag in view definition. The PR looks good to me overall. Just left some minor comment.
I agree with @danny0405 we need to support views via DDL, which can be done as follow ups. My understanding is this PR only enables the planner to handle CatalogViews. But we shouldn't expect end users to directly create CatalogView instances because users wouldn't know how to generate expanded query strings, right?

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/catalog/DatabaseCalciteSchema.java
dawidwys: Not a temporal, but temporary. Yes, tables, views, functions, etc. That's what we discussed and agreed upon in FLIP-64 and FLIP-57.

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/plan/schema/ExpandingPreparingTable.java
dawidwys: This is correct. The implementation is free to expand more than the outer level. Than the `ExpandingPreparingTable` will make an empty pass over the nested nodes, cause inner levels were already expanded. This is absolutely fine though.

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/catalog/SqlCatalogViewTable.java
dawidwys: TBH, I am not sure of whats the purpose of this line. I added it, because the same is done in Calcite in `ViewTable`:

```
  private RelRoot expandView(RelOptTable.ToRelContext context,
      RelDataType rowType, String queryString) {
    try {
      final RelRoot root =
          context.expandView(rowType, queryString, schemaPath, viewPath);
      final RelNode rel = RelOptUtil.createCastRel(root.rel, rowType, true);
      ...
  }
```

If you are sure it is safe to remove it, I'm fine doing it, but could you explain why it was introduced in `ViewTable`?

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/calcite/FlinkRelBuilder.scala
dawidwys: I think the current behavior is correct whereas the old one was rather wrong or not necessarily wrong, but was throwing away the benefits of having the `unwrap` method.

What do you mean that it returns always `clusterContext`? It always returns `ChainContext` with sub contexts as provided, than you can access particular context through `unwrap` method. The `Contexts.chain` method is exactly for the purpose of merging multiple different contexts.

Let's have a look at an example:
```
val calciteConfig = Contexts.of(
      FLINK_LOGICAL_PROJECT_FACTORY, 
      ...
      )
val flinkConfig = new FlinkConfig(...)

val chainedConfig = Contexts.chain(calciteConfig, flinkConfig)

chainedConfig.unwrap(classOf[FlinkConfig]) // returns flinkConfig
chainedConfig.unwrap(classOf[ProjectFactory]) // returns FLINK_LOGICAL_PROJECT_FACTORY
```

Of course we can talk about which of the contexts should take precedence. You are probably right(though not 100% sure as the merging seems a bit odd to me) that the `context` should be put first. Will change that.

flink-table/flink-table-planner-blink/src/test/scala/org/apache/flink/table/planner/plan/common/ViewsExpanding.scala
dawidwys: Good catch. Will fix it.

dawidwys:
Thanks @lirui-apache @danny0405 for your reviews. The DDL and view creation in general is not subject of this PR at all. @lirui-apache understanding of this PR is absolutely valid. 

The view creation is absolutely different story. As far as I know it is covered in FLIP-71.

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/catalog/SqlCatalogViewTable.java
danny0405: The only reason i can see is to fix the nullability, or rename the node (e.g. create view with column alias), for the nullability, we should fix it from the source, we do not support the view alias yet, so there is no need to add the casts.

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/plan/schema/ExpandingPreparingTable.java
danny0405: Yes, the logic is right, but a little conflict with the method name, which is a little confusing.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/calcite/FlinkRelBuilder.scala
danny0405: The `ChainContext#unwap` is easy to use but a little hard to understand, just like you said, the context returned is very relevant with the sequence of the contexts(for same class contexts), which is not that intuitive. I'm not against to use `ChainContext`, we just need to be caution with the usage.

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/plan/schema/ExpandingPreparingTable.java
dawidwys: Is it though? The method says expand the top level. That's what the implementer should do. That's the contract of the method. The comment just says its fine if you do more.

dawidwys: I'm fine changing the name to sth different if you have a good suggestion. The change `at least` to `at most` is not valid though.

danny0405: Or, better, the sub-class just return an un-expand relational node, let's do the expanding work all in the parent class, which would make the interface duty/function more clear.

dawidwys: Yes, that's what the method does. By expanding I meant:
- in case of sql string view it means converting sql to relNode
- in case of Operation view it means converting SqlOperation to relNode (this will actually expand all nested scans because we are using expanding scan factory - that's why I added the comment)

If it makes it more understandable how about renaming it to `convertToRel`? Any other ideas for a name?

dawidwys: Yes, that's what the method does. By expanding I meant:
- in case of sql string view it means converting sql to relNode
- in case of Operation view it means converting SqlOperation to relNode (this will actually expand all nested scans because we are using expanding scan factory - that's why I added the comment)

If it makes it more understandable how about renaming it to `convertToRel`? Any other ideas for a name?

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/catalog/SqlCatalogViewTable.java
dawidwys: Ok, if you are sure about it, I'm fine removing it. I would add a check in the PlannerImpl that verifies the types match instead. WDYT?

danny0405: That make sense, thanks.

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/plan/schema/ExpandingPreparingTable.java
danny0405: I'm okey with the new name.

PR #8618, opened 2019-06-04
 [FLINK-12732][state-processor-api] Add savepoint reader for consuming partitioned operator state

sjwiesman:
## What is the purpose of the change

This is the second PR for FLIP-43 adding the functionality to read partitioned operator state from a state snapshot. It is based on #8615 and only the final 4 commits should be reviewed.

## Brief change log

62dd050 - Some basic refactoring, this pulls the `NeverCompleteFuture` into a reusable utility

38eaa66 - Implements a shim `Environment` on-top of the available user `RuntimeContext`

9689001 - Adds the new functionality to read keyed state out of a snapshot

be0552c - Adds new documentation

## Verifying this change

*(Please pick either of the following options)*

This change added tests and can be verified as follows:
UT and IT tests

## Does this pull request potentially affect one of the following parts:

  - Dependencies (does it add or upgrade a dependency): (yes / **no**)
  - The public API, i.e., is any changed class annotated with `@Public(Evolving)`: (yes / **no**)
  - The serializers: (yes / no / don't know)
  - The runtime per-record code paths (performance sensitive): (yes / **no** / don't know)
  - Anything that affects deployment or recovery: JobManager (and its components), Checkpointing, Yarn/Mesos, ZooKeeper: (yes / **no** / don't know)
  - The S3 file system connector: (yes / **no** / don't know)

## Documentation

  - Does this pull request introduce a new feature? (**yes** / no)
  - If yes, how is the feature documented? (not applicable / **docs** / JavaDocs / not documented)

PR #8861, opened 2019-06-24
[FLINK-12963][state-processor-api] Add savepoint writer for bootstrapping new savepoints

sjwiesman:

Introduce a savepoint writer for bootstrapping state. 

## Brief change log
This is based on #8618 and only the last 4 commits are relevant. 

b341e91  Introduces the new runtime components for running a bounded stream task and snapshotting its contents after processing all elements. **Note** this is the most important commit. 

830ac6a Add user-facing interfaces and corresponding stream operators. 

f12439f Adds the user api 

b0f176e  Update the travis profile so that the state processor api is run during the libraries stage

## Verifying this change

UT and IT test

  - The serializers: no

## Documentation

  - If yes, how is the feature documented? documented


sjwiesman:
cc @tzulitai

docs/dev/libs/state_processor_api.md
```diff
@@ -0,0 +1,459 @@
...
+</div>
+<div data-lang="scala" markdown="1">
+{% highlight scala %}
+case class Account(id: Int, amount: Double, timestamp: Long)
+ 
+class AccountBootstrapper extends KeyedStateBootstrapFunction[Integer, Account] {
+    var state: ValueState[Double]
+
+    override def open(parameters: Configuration): Unit = {
+        val descriptor = new ValueStateDescriptor[Double]("total",Types.DOUBLE)
```
Myasuka: I wonder could this be acceptable for scala? Should the `Double` be `JDouble` after we `import java.lang.{Double => JDouble}`

docs/dev/libs/state_processor_api.md
```diff
@@ -0,0 +1,459 @@
...
+CREATE TABLE currency_converter.rates (
+   value Tuple2<Integer, Double>
+);
+ 
+CREATE NAMESPACE summarize;
+ 
+CREATE TABLE summarize.keyed_state (
+   key   INTEGER PRIMARY KEY,
+   total DOUBLE,
+   count INTEGER
```
Myasuka: `count` would be treated as a key word for SQL and showed as bold in web UI, shall we rename this to another name?

sjwiesman: I put it and key in backticks, it's more about expressing the idea.

docs/dev/libs/state_processor_api.md
sjwiesman: Nice catch

flink-libraries/flink-state-processing-api/src/main/java/org/apache/flink/state/api/BootstrapTransformation.java
```diff
@@ -0,0 +1,129 @@
...
+			config,
+			metadata
+		);
+
+		MapPartitionOperator<T, TaggedOperatorSubtaskState> subtaskStates = input
+			.mapPartition(operatorRunner)
+			.name(uid);
+
+		if (operator instanceof BroadcastStateBootstrapOperator) {
+			subtaskStates = subtaskStates.setParallelism(1);
```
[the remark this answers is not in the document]
sjwiesman: Because the DataSet api does not support broadcast partitioning. This is the closest approximation; if the parallelism is one than all elements will be shipped to the same task slot. In practice it's fine because we only need a single copy of the broadcast state which can then be restored at any parallelism.

sjwiesman:
Thanks for taking a look @Myasuka 

Parallelism does not matter since it does not affect the user's ability to restore, only max parallelism which is used to define key groups. This implementation assumes the parallelism of the provided data set was set appropriately by the user for the data. The only times we change it are when 

1) Writing broadcast state
2) The parallelism is greater than the specified max parallelism

As outlined in the docs, max parallelism is set either when creating a new savepoint `Savepoint.create` or when using an existing savepoint as a base `Savepoint.load` its max parallelism is used.

I do agree, the ability to set max functionality at a per operator level, in addition to the global max parallelism would be useful.

sjwiesman:
@tzulitai I've rebased on master and made some misc updates to align with the changes you made on the reader api when merging.

flink-libraries/flink-state-processing-api/src/main/java/org/apache/flink/state/api/WritableSavepoint.java
```diff
@@ -0,0 +1,129 @@
...
+
+		DataSet<OperatorState> finalOperatorStates;
+		if (existingOperators == null) {
+			finalOperatorStates = newOperatorStates;
+		} else {
+			finalOperatorStates = newOperatorStates.union(existingOperators);
+		}
+
+		finalOperatorStates
+			.reduceGroup(new OperatorStateReducer(metadata))
```
tzulitai: Following my comment regarding how we should use the `savepointMetadata` to maintain an index of `Map<String(uid), Either<OperatorState, BootstrapTransformation>>`, I think this reduce can even be avoided.

The main logic for `write` would essentially become:
1. For each entry in the index, write the operator state data. If it is an `OperatorState`, then we simply re-write the data. If it is a `BootstrapTransformation`, then we perform the logic in `transformation.getOperatorSubtaskStates(...)`.
2. By now, all values in the `SavepointMetadata` index should be `OperatorState`s, and no longer `BootstrapTransformation`s. Simply output the `SavepointMetadata` using `SavepointOutputFormat`. The output format internall transforms our `SavepointMetadata` back to Flink's `Savepoint`, just for the sake of API usage.

tzulitai:
@flinkbot approve description
@flinkbot approve consensus

flink-libraries/flink-state-processing-api/src/main/java/org/apache/flink/state/api/WritableSavepoint.java
```diff
@@ -0,0 +1,129 @@
...
+	}
+
+	/**
+	 * Drop an existing operator from the savepoint.
+	 * @param uid The uid of the operator.
+	 * @return A modified savepoint.
+	 */
+	@SuppressWarnings("unchecked")
+	public F removeOperator(String uid) {
+		droppedOperators.add(uid);
```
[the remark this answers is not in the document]
sjwiesman: Yes, that is much better than what I have.

flink-libraries/flink-state-processing-api/src/main/java/org/apache/flink/state/api/WritableSavepoint.java
```diff
@@ -0,0 +1,129 @@
...
+			.stream()
+			.map(entry -> getOperatorStates(savepointPath, entry.getKey(), stateBackend, entry.getValue(), metadata))
+			.reduce(DataSet::union)
+			.orElseThrow(() -> new IllegalStateException("Savepoint's must contain at least one operator"));
+
+		DataSet<OperatorState> finalOperatorStates;
+		if (existingOperators == null) {
+			finalOperatorStates = newOperatorStates;
+		} else {
+			finalOperatorStates = newOperatorStates.union(existingOperators);
```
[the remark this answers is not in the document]
sjwiesman: Currently no, it is just a shallow copy of pointers.

flink-libraries/flink-state-processing-api/src/main/java/org/apache/flink/state/api/WritableSavepoint.java
```diff
@@ -0,0 +1,129 @@
...
+			DataSet<OperatorState> wrappedCollection = newOperatorStates
+				.getExecutionEnvironment()
+				.fromCollection(existingOperators);
+
+			finalOperatorStates = newOperatorStates.union(wrappedCollection);
+		}
+		return finalOperatorStates;
+	}
+
+	private List<OperatorState> getExistingOperatorStates() {
```
tzulitai: I think this method as well as `getOperatorStates(Path)` should be refactored into `ModifiableSavepointMetadata`.
The concern of filtering out the operator state index to return only the existing / new operator states can be encapsulated there.

flink-libraries/flink-state-processing-api/src/main/java/org/apache/flink/state/api/runtime/metadata/ModifiableSavepointMetadata.java
```diff
@@ -0,0 +1,85 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Savepoint metadata that can be modified.
+ */
+@Internal
+public class ModifiableSavepointMetadata extends SavepointMetadata {
+
+	private transient Map<OperatorID, Either<OperatorState, BootstrapTransformation<?>>> operatorStateIndex;
```
tzulitai: Just for the sake of readability, I would introduce a dedicated wrapper class, say `OperatorStateSpec` to replace this `Either`. `Either.left(...)` / `Either.isLeft()` is harder to understand at first glance than `OperatorStateSpec.existing(...) / OperatorStateSpec.isExistingState()`.

tzulitai: Concerning `transient`:

I think generally speaking, the `ModifiableSavepointMetadata` should not need to be serializable at all, since its main concern is bookkeeping what existing / new operators there are to facilitate building the job graph, and therefore do not need to be part of the serialized job graph.

I think the only reason it needs to be serializable at the moment is because some runtime classes, e.g. `MergeOperatorStates` reduce operator, holds a reference to it.
It should be possible (as IMO also more reasonable) to reduce the dependencies on `SavepointMetadata` in these runtime classes, and only pass in the necessary information. This should allow us to not make `SavepointMetadata` a `Serializable`.

flink-libraries/flink-state-processing-api/src/main/java/org/apache/flink/state/api/runtime/metadata/SavepointMetadata.java
```diff
@@ -20,32 +20,38 @@
...
-import java.io.IOException;
 import java.io.Serializable;
 import java.util.Collection;
 
 /**
  * Returns metadata about a savepoint.
  */
 @Internal
-public interface SavepointMetadata extends Serializable {
+public class SavepointMetadata implements Serializable {
```
tzulitai: I don't think we need this base class anymore (at least from the current state of the code). It's only subclass is the `ModifiableSavepointMetadata`.

Moreover, the contents of this base class doesn't seem to justify as a proper savepoint metadata anyways, since it only maintains the master states.

Therefore, I would suggest to simplify the hierarchy and just have a single `SavepointMetadata` class for now.

tzulitai:
@flinkbot approve all

PR #8862, opened 2019-06-24
[FLINK-12964][sql client] add commented-out defaults to sql client yaml file to make it easier for users to adopt

bowenli86:

This PR adds commented-out defaults for catalogs to sql client yaml file, to make it easier for users to adopt. This is done similar to how existing defaults of tables and functions is written, e.g.

```
#==============================================================================
# Tables
#==============================================================================

# Define tables here such as sources, sinks, views, or temporal tables.

tables: [] # empty list
# A typical table source definition looks like:
# - name: ...
#   type: source-table
#   connector: ...
#   format: ...
#   schema: ...

...

#==============================================================================
# User-defined functions
#==============================================================================

# Define scalar, aggregate, or table functions here.

functions: [] # empty list
# A typical function definition looks like:
# - name: ...
#   from: class
#   class: ...
#   constructor: ...
```

## Brief change log

*(for example:)*

## Verifying this change

This change is a trivial rework / code cleanup without any test coverage. And verified manually

  - The public API, i.e., is any changed class annotated with `@Public(Evolving)`:  (no)
  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-table/flink-sql-client/conf/sql-client-defaults.yaml
+++ b/flink-table/flink-sql-client/conf/sql-client-defaults.yaml
@@ -119,6 +119,8 @@ deployment:
 # Catalogs
 #==============================================================================
 
+# Define catalogs here.
+
 catalogs: [] # empty list
 #  - name: myhive
 #    type: hive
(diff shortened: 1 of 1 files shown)


bowenli86:

PR #8865, opened 2019-06-25
 [FLINK-12924][table] Introduce basic type inference interfaces

twalthr:

This PR introduces an easly version of the type inference logic. It introduces the basic interfaces and first logic for inferrence in `TypeInferenceUtil`.

More tests will follow once the first type strategies and validators are added.

This PR exclude the following features:
- DEFAULT expression
- NULL input type inference
- assignment operators

## Brief change log

See commit messages.

## Verifying this change

Not tested yet.

  - The serializers: no

## Documentation

### DIFF
--- a/flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/inference/CallContextBase.java
+++ b/flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/inference/CallContextBase.java
@@ -42,12 +42,16 @@ public interface CallContextBase {
 	/**
 	 * Returns {@code true} if the argument at the given position is a literal and {@code null},
 	 * {@code false} otherwise.
+	 *
+	 * <p>Use {@link #isArgumentLiteral(int)} before to check if the argument is actually a literal.
 	 */
 	boolean isArgumentNull(int pos);
 
 	/**
 	 * Returns the literal value of the argument at the given position, given that the argument is a
 	 * literal, is not null, and can be expressed as an instance of the provided class.
+	 *
+	 * <p>Use {@link #isArgumentLiteral(int)} before to check if the argument is actually a literal.
 	 */
 	<T> Optional<T> getArgumentValue(int pos, Class<T> clazz);
 
--- a/flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/inference/CallContextBase.java
+++ b/flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/inference/CallContextBase.java
@@ -1,0 +1,60 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.table.types.inference;
+
+import org.apache.flink.annotation.PublicEvolving;
+import org.apache.flink.table.functions.FunctionDefinition;
+
+import java.util.Optional;
+
+/**
+ * Provides details about the function call for which type inference is performed.
+ */
+@PublicEvolving
+public interface CallContextBase {
+
+	/**
+	 * Returns the function definition that defines the function currently being called.
+	 */
+	FunctionDefinition getFunctionDefinition();
+
+	/**
+	 * Returns whether the argument at the given position is a value literal.
+	 */
+	boolean isArgumentLiteral(int pos);
+
+	/**
+	 * Returns {@code true} if the argument at the given position is a literal and {@code null},
+	 * {@code false} otherwise.
+	 */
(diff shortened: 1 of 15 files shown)


flink-table/flink-table-planner/src/main/java/org/apache/flink/table/expressions/rules/ResolveCallByArgumentsRule.java
```diff
@@ -160,10 +234,70 @@ private ResolvedExpression castIfNeeded(
...
+		}
+
+		@Override
+		public boolean isArgumentLiteral(int pos) {
+			return getArgument(pos) instanceof ValueLiteralExpression;
+		}
+
+		@Override
+		public boolean isArgumentNull(int pos) {
+			final ValueLiteralExpression literal = (ValueLiteralExpression) getArgument(pos);
```
dawidwys: nit: Maybe check `isArgumentLiteral` just to have a nicer exception?

flink-table/flink-table-planner/src/main/java/org/apache/flink/table/expressions/rules/ResolveCallByArgumentsRule.java
```diff
@@ -160,10 +234,70 @@ private ResolvedExpression castIfNeeded(
...
+			return literal.isNull();
+		}
+
+		@Override
+		public <T> Optional<T> getArgumentValue(int pos, Class<T> clazz) {
+			final ValueLiteralExpression literal = (ValueLiteralExpression) getArgument(pos);
```
dawidwys: nit: Maybe check `isArgumentLiteral` just to have a nicer exception?

flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/inference/CallContextBase.java
```diff
@@ -0,0 +1,60 @@
...
+	/**
+	 * Returns whether the argument at the given position is a value literal.
+	 */
+	boolean isArgumentLiteral(int pos);
+
+	/**
+	 * Returns {@code true} if the argument at the given position is a literal and {@code null},
+	 * {@code false} otherwise.
+	 */
+	boolean isArgumentNull(int pos);
```
dawidwys: nit: Maybe add a note the `isArgumentLiteral` should be called first?

flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/inference/CallContextBase.java
```diff
@@ -0,0 +1,60 @@
...
+
+	/**
+	 * Returns the literal value of the argument at the given position, given that the argument is a
+	 * literal, is not null, and can be expressed as an instance of the provided class.
```
dawidwys: nit: Maybe add a note the `isArgumentLiteral` should be called first?

flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/inference/TypeInference.java
```diff
@@ -0,0 +1,183 @@
...
+				Preconditions.checkNotNull(accumulatorTypeStrategy, "Accumulator type strategy must not be null.");
+			return this;
+		}
+
+		/**
+		 * Sets the strategy for inferring the final output data type of a function call.
+		 *
+		 * <p>Required.
+		 */
+		public Builder outputTypeStrategy(TypeStrategy outputTypeStrategy) {
```
dawidwys: Personal taste, so feel free to ignore as we never had a proper discussion on design patterns standards.

Personally, in most cases, I prefer to have a static factory method with the required parameter(s). In this case I would prefer `public static Builder TypeInference#outputTypeStrategy(TypeStrategy outputTypeStrategy)`, but as I said this is just a personal taste.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/inference/TypeTransformation.java
```diff
@@ -0,0 +1,34 @@
...
+package org.apache.flink.table.types.inference;
+
+import org.apache.flink.annotation.PublicEvolving;
+import org.apache.flink.table.types.DataType;
+
+/**
+ * Transforms one data type to another.
+ */
+@PublicEvolving
+public interface TypeTransformation {
```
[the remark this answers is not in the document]
twalthr: This interface is similar to Calcite's `org.apache.calcite.sql.type.SqlTypeTransform`. I could image transforms such as `toNullable`, `bridgedToInternalConversions`, etc.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/inference/TypeInferenceUtil.java
```diff
@@ -0,0 +1,307 @@
...
+			return outputDataType;
+		}
+	}
+
+	// --------------------------------------------------------------------------------------------
+
+	private static Result runTypeInferenceInternal(TypeInference typeInference, CallContext callContext) {
+		final List<DataType> argumentTypes = callContext.getArgumentDataTypes();
+
+		try {
```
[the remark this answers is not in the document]
twalthr: The signature is only an interesting information if the input arguments were not correct. If something is going wrong in the accumulator or return type inference this information is not interesting.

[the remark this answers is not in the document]
twalthr: The second exception also uses the adapted call context.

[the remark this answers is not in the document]
dawidwys: Could we then merge the two `try ... catch` blocks? I saw that the second call uses the `adaptedCallContext` but actually it makes no difference. Internally you use only `getName`, `getFunctionDefinition`, which are the same for both contexts.

PR #8872, opened 2019-06-25
[FLINK-12977][table]Port CsvTableSource to api-java-bridge

c4emmmm:

This PR is currently tied to 2 jiras:
FLINK-12977
Port CsvTableSource and CsvTableSink to api-java-bridge
FLINK-12978
Support LookupableTableSource for CsvTableSource
These changes can be separated into 2 PRs if it's required.

Currently, CsvTableSource and CsvTableSink are in flink-planner. If they can be moved to `flink-table-api-java-bridge`, so that blink planner can reuse them.

This should be a straightforward move, no API should be changed. And we should rewrite them in Java, because api-java-bridge doesn't depends on Scala.

CsvTableSource is a highly used source in demo and testing. It will be great if it also supports LookupableTableSource so that it can be used in temporal table join (act like a external dimension table).

The lookupable csv source implementation should be tested in blink planner, because currently only blink support lookup.

## Brief change log

Port CsvTableSource to flink-table-api-java-bridge and rewrite it in Java.
Support LookupableTableSource in the new CsvTableSource.
Modify two test cases since they use the constructor of CsvTableSource which is exposed in the Scala version but now hidden in the new Java one.

## Verifying this change

This change is already covered by existing tests, such as TableSourceTest, TableSourceValidationTest, ExternalCatalogSchemaTest.

  - The serializers: (no)

## Documentation


flink-python/pyflink/table/sinks.py
```diff
@@ -49,24 +49,20 @@ class CsvTableSink(TableSink):
     :param write_mode: The write mode to specify whether existing files are overwritten or not.
     """
 
-    def __init__(self, field_names, field_types, path, field_delimiter=',', num_files=1,
+    def __init__(self, field_names, field_types, path, field_delimiter=',', num_files=-1,
```
wuchong: @dianfu could you have a look at the pyflink changes?

c4emmmm:
> Thanks @c4emmmm .
> 
> I think we should also add a `CsvTableSourceITCase` in blink-planner to verify the lookup and scan can work well.
> 
> You can refer to `org.apache.flink.table.sources.csv.CsvTableSourceITCase#testTemporalJoinCsv`.

@wuchong Thanks for your feedback. I have push a commit with most points fixed. Test case would be a follow-up commit.

c4emmmm:
@wuchong 
A test case for lookup join has been added, please take a look when you have time.

Before this is merged, I shall merge these commits into reasonable fewer ones if everything seems ok. Please let me know when you think it's ready. Thanks.

flink-table/flink-table-api-java-bridge/src/main/java/org/apache/flink/table/sources/CsvTableSource.java
```diff
@@ -0,0 +1,541 @@
...
+		this(new CsvInputFormatConfig(path, fieldNames, fieldTypes, selectedFields,
+			fieldDelim, lineDelim, quoteCharacter, ignoreFirstLine, ignoreComments, lenient));
+	}
+
+	private CsvTableSource(CsvInputFormatConfig config) {
+		this.config = config;
+	}
+
+	/**
+	 * Return a new builder that builds a CsvTableSource. For example: {{{ val source:
```
wuchong: `{{{` is a scaladoc, please use `<pre></pre>` in javadoc, and please pay attention to the line break.

flink-table/flink-table-api-java-bridge/src/main/java/org/apache/flink/table/sinks/CsvTableSink.java
```diff
@@ -0,0 +1,176 @@
...
+
+	@Override
+	public TypeInformation<?>[] getFieldTypes() {
+		return fieldTypes;
+	}
+
+	/**
+	 * Formats a [[Row]] into a [[String]] with fields separated by the field delimiter.
+	 */
+	public static class CsvFormatter implements MapFunction<Row, String> {
```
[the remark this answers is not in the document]
wuchong: Please add a serialVersionUID for this class

flink-python/pyflink/table/sinks.py
dianfu: The pyflink related changes LGTM.

PR #8877, opened 2019-06-25
[FLINK-12984][metrics] only call Histogram#getStatistics() once where possible

NicoK:

In some occasions, `Histogram#getStatistics()` was called multiple times to retrieve different statistics. However, at least the Dropwizard implementation has some constant overhead per call and we should maybe rather interpret this method as returning a point-in-time snapshot of the histogram in order to get consistent values when querying them.

Please note that this PR is based on #8886.
-> follow-up PR: #8884

## Brief change log

- adapt prometheus reporter
- adapt JMX and dropwizward unit tests

## Verifying this change

  - The serializers: **no**

## Documentation

### DIFF
--- a/flink-metrics/flink-metrics-core/pom.xml
+++ b/flink-metrics/flink-metrics-core/pom.xml
@@ -34,6 +34,15 @@ under the License.
 
 	<packaging>jar</packaging>
 
+	<dependencies>
+		<!-- ================== test dependencies ================== -->
+
+		<dependency>
+			<groupId>org.apache.flink</groupId>
+			<artifactId>flink-test-utils-junit</artifactId>
+		</dependency>
+	</dependencies>
+
 	<build>
 		<plugins>
 			<!-- activate API compatibility checks -->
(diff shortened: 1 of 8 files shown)


StephanEwen:
Looks good in general.

What is a bit strange is that the dropwizard metrics now have dependencies on flink-runtime.
Is that because `DescriptiveStatisticsHistogram` is in the wrong place (`flink-runtime`)? Should that implementation be in the metrics projects instead?

NicoK:
@StephanEwen yes, indirectly: I needed the (test-scope, test-jar) dependency to have a common test class `AbstractHistogramTest` and this needed to be in `flink-runtime` because `DescriptiveStatisticsHistogram` lives there.

FYI: There was already a test-scope (normal jar) dependency on `flink-runtime`, presumably for similar reasons. I guess, the reason to put it into `flink-runtime` was to have some default to be available at any time without extra dependencies for the user, and also because our latency markers use the `DescriptiveStatisticsHistogram`.

StephanEwen:
Couldn't we move these classes into the metrics modules? Both the test base and the `DescriptiveStatisticsHistogram`? It seems just wrong to couple the metrics to the runtime modules for these dependencies.

Sorry for being nit-picking here, but we need to push for more clean and lean dependency structure. Too many times that we had to work around issues with out dependency jungle...

NicoK:
I moved the new `AbstractHistogramTest` to `flink-metrics-core` which should cover everything that was touched during this PR. I propose to do move `DescriptiveStatisticsHistogram` in a separate PR (this would add a dependency form `runtime` to `metrics-core` though).

zentol:
You wanna squash the move of the test into #8886?

PR #8880, opened 2019-06-25
[FLINK-12986] [network] Fix instability of BoundedBlockingSubpartition under memory pressure.

StephanEwen:

This PR changes the implementation of the `BoundedBlockingResultpartition` and adds alternative implementations to evaluate.

The current implementation (based on memory mapped files) turned out to have stability issues under large scale and memory pressure. Big thanks to Yingjie Cao for testing/benchmarking this at scale.

The theory for the cause of the stability issue is that processes may stall when trying to allocate memory, because first memory would have to be freed by evicting a page from the memory mapped file and writing it to disk. This lazy persistence is the problem.

We implemented and tested different approaches:
  - `MMAP`: Current implementation
  - `FILE`: Directly writes to a file and reads from the file, so no lazy writing causing stalls
  - `FILE_MMAP`: Directly writes to a file and maps the file only for reading. This also has no lazy write (on page eviction)

Tests concluded that the `FILE_MMAP` variant is stable and the implementation of choice.
The PR still includes the other implementations (with tests) for future experiments.

## Brief change log

  - Factors out the implementation of the bounded data storage from the `BoundedBlockingResultpartition`
  - Add the implementations for `FILE`, `MMAP`, `FILE_MMAP`
  - Generify tests to cover all three implementations (through test bases / parameterized tests)
  - Change `ResultPartitionFactory` to instantiate the `FILE_MMAP` variant for blocking result partitions.

## Verifying this change

  - Any batch job (DataSet API) with blocking shuffles can be used to verify this. Set the `ExecutionMode` to `BATCH_FORCED`.
  - The batch job should have multiple TaskManagers to trigger the Netty shuffle code paths

  - The serializers: **no**

## Documentation


flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BufferReaderWriterUtil.java
```diff
@@ -0,0 +1,227 @@
...
+
+		final ByteBuffer headerBuffer = arrayWithHeaderBuffer[0];
+		headerBuffer.clear();
+		headerBuffer.putInt(buffer.isBuffer() ? HEADER_VALUE_IS_BUFFER : HEADER_VALUE_IS_EVENT);
+		headerBuffer.putInt(buffer.getSize());
+		headerBuffer.flip();
+
+		arrayWithHeaderBuffer[1] = buffer.getNioBufferReadable();
+
+		return channel.write(arrayWithHeaderBuffer);
```
wsry: I am not sure if this function call can guarantee to write all data out, at least the java doc of GatheringByteChannel#write says that it may return zero. Maybe it is more safe to do the write in a while loop just like read.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/FileChannelMemoryMappedBoundedData.java
```diff
@@ -0,0 +1,190 @@
...
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.StandardOpenOption;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
+
+import static org.apache.flink.util.Preconditions.checkState;
+
+final class FileChannelMemoryMappedBoundedData implements BoundedData {
```
wsry: It is better to supplement some java doc.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/FileChannelMemoryMappedBoundedData.java
```diff
@@ -0,0 +1,190 @@
...
+	private boolean tryWriteBuffer(Buffer buffer) throws IOException {
+		final long spaceLeft = endOfCurrentRegion - pos;
+		final long bytesWritten = BufferReaderWriterUtil.writeToByteChannelIfBelowSize(
+				fileChannel, buffer, headerAndBufferArray, spaceLeft);
+
+		if (bytesWritten >= 0) {
+			pos += bytesWritten;
+			return true;
+		}
+		else {
```
wsry: else can be removed

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/FileChannelMemoryMappedBoundedData.java
```diff
@@ -0,0 +1,190 @@
...
+				"The buffer (%d bytes) is larger than the maximum size of a memory buffer (%d bytes)",
+				buffer.getSize(), maxRegionSize));
+	}
+
+	// ------------------------------------------------------------------------
+	//  Factories
+	// ------------------------------------------------------------------------
+
+	/**
+	 * Creates new MemoryMappedBuffers, creating a memory mapped file at the given path.
```
wsry: MemoryMappedBuffers -> FileChannelMemoryMappedBoundedData

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/FileChannelMemoryMappedBoundedData.java
```diff
@@ -0,0 +1,190 @@
...
+	 */
+	public static FileChannelMemoryMappedBoundedData create(Path memMappedFilePath) throws IOException {
+		return createWithRegionSize(memMappedFilePath, Integer.MAX_VALUE);
+	}
+
+	/**
+	 * Creates new MemoryMappedBuffers, creating a memory mapped file at the given path.
```
wsry: MemoryMappedBuffers -> FileChannelMemoryMappedBoundedData

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/MemoryMappedBoundedData.java
```diff
@@ -44,10 +43,10 @@
 /**
  * This class is largely a workaround for the fact that a memory mapped region in Java can cannot
```
wsry: can cannot -> cannot

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/MemoryMappedBoundedData.java
```diff
@@ -70,15 +69,15 @@
  * and {@link BoundedBlockingSubpartitionReader}, such as writing first and rading after.
```
wsry: rading after -> reading after

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/MemoryMappedBoundedData.java
```diff
@@ -70,15 +69,15 @@
...
  * Not obeying these contracts throws NullPointerExceptions.
  */
-class MemoryMappedBuffers implements Closeable {
+final class MemoryMappedBoundedData implements BoundedData {
 
 	/** Memory mappings should be at the granularity of page sizes, for efficiency. */
 	private static final int PAGE_SIZE = PageSizeUtil.getSystemPageSizeOrConservativeMultiple();
 
-	/** The encoder to the current memory mapped region we are writing to.
+	/** The the current memory mapped region we are writing to.
```
wsry: The the -> The

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartition.java
```diff
@@ -41,9 +40,8 @@
  * in a blocking manner: The result is first produced, then consumed.
  * The result can be consumed possibly multiple times.
  *
- * <p>The implementation creates a temporary memory mapped file and writes all buffers to that
- * memory and serves the result from that memory. The kernel backs the mapped memory region
- * with physical memory and file space incrementally as new pages are filled.
+ * <p>Depending on the upplied implementation of {@link BoundedData}, the actual data is stored
```
wsry: upplied -> supplied?

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartition.java
```diff
@@ -91,28 +89,16 @@
 	private boolean isReleased;
 
 	/**
-	 * Common constructor.
+	 * Main constructor.
```
wsry: the comment can be deleted (no useful information)

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/FileChannelMemoryMappedBoundedData.java
```diff
@@ -0,0 +1,190 @@
...
+
+	// ------------------------------------------------------------------------
+	//  Factories
+	// ------------------------------------------------------------------------
+
+	/**
+	 * Creates new MemoryMappedBuffers, creating a memory mapped file at the given path.
+	 */
+	public static FileChannelMemoryMappedBoundedData create(Path memMappedFilePath) throws IOException {
+		return createWithRegionSize(memMappedFilePath, Integer.MAX_VALUE);
```
zhijiangW: checkNotNull(memMappedFilePath)?

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/FileChannelMemoryMappedBoundedData.java
```diff
@@ -0,0 +1,190 @@
...
+	}
+
+	/**
+	 * Creates new MemoryMappedBuffers, creating a memory mapped file at the given path.
+	 * Each mapped region (= ByteBuffer) will be of the given size.
+	 */
+	public static FileChannelMemoryMappedBoundedData createWithRegionSize(Path memMappedFilePath, int regionSize) throws IOException {
```
zhijiangW: `checkNotNull(memMappedFilePath)`, considering this factory might be used in other places, not only for current `Integer.MAX_VALUE` as default size. Is it worth also checking the region size?

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/FileChannelMemoryMappedBoundedData.java
```diff
@@ -0,0 +1,190 @@
...
+	/** The position where the current memory mapped started. */
+	private long startOfCurrentRegion;
+
+	/** The maximum size of each mapped region. */
+	private final long maxRegionSize;
+
+	FileChannelMemoryMappedBoundedData(
+			Path filePath,
+			FileChannel fileChannel,
+			int maxSizePerMappedRegion) throws IOException {
```
zhijiangW: nit: `IOException` would never be thrown atm

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/FileChannelMemoryMappedBoundedData.java
```diff
@@ -0,0 +1,190 @@
...
+	/** The position where the current memory mapped region must end. */
+	private long endOfCurrentRegion;
+
+	/** The position where the current memory mapped started. */
+	private long startOfCurrentRegion;
+
+	/** The maximum size of each mapped region. */
+	private final long maxRegionSize;
+
+	FileChannelMemoryMappedBoundedData(
```
zhijiangW: could be private if it is constructed only from factory atm

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/FileChannelMemoryMappedBoundedData.java
```diff
@@ -0,0 +1,190 @@
...
+	 * After calling this method, further calls to {@link #writeBuffer(Buffer)} will fail.
+	 */
+	@Override
+	public void finishWrite() throws IOException {
+		mapRegionAndStartNext();
+		fileChannel.close();
+	}
+
+	/**
+	 * Closes the file and unmaps all memory mapped regions
```
zhijiangW: missing sentence end mark

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartitionReader.java
```diff
@@ -44,7 +44,7 @@
 	/** The reader/decoder to the memory mapped region with the data we currently read from.
 	 * Null once the reader empty or disposed.*/
 	@Nullable
-	private BufferSlicer memory;
+	private BoundedData.Reader data;
```
wsry: Maybe “dataReader” is better than “data” to distinguish from instance of  BoundedData.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/ResultPartitionFactory.java
```diff
@@ -157,8 +159,9 @@ private static void initializeBoundedBlockingPartitions(
 		int i = 0;
 		try {
 			for (; i < subpartitions.length; i++) {
-				subpartitions[i] = new BoundedBlockingSubpartition(
-					i, parent, ioManager.createChannel().getPathFile().toPath());
+				final File spillFile = ioManager.createChannel().getPathFile();
+				subpartitions[i] =
+						BoundedBlockingSubpartition.createWithFileAndMemoryMappedReader(i, parent, spillFile);
```
wsry: Now that three ways of file writing/reading are implemented, why not make it configurable?

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/FileChannelBoundedData.java
```diff
@@ -0,0 +1,160 @@
...
+				if (next != null) {
+					return next;
+				}
+				else {
+					recycle(memory);
+					return null;
+				}
+			}
+
+			throw new IOException("Bug in BoundedBlockingSubpartition with FILE data: " +
```
wsry: Though, theoretically, two read buffers is enough. But there could be still problems, because the writeAndFlush method of netty channel dose not work synchronously and the reader can be scheduled not only when the buffer of the same reader is sent out but also when the downstream adding credit or when buffers of other readers multiplex the same channel are sent out. So it is unsafe to make the assumption that when the reader is scheduled there must be a buffer for reading ahead.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/MemoryMappedBoundedData.java
```diff
@@ -220,28 +225,29 @@ private static int alignSize(int maxRegionSize) {
...
 	 * sequence of mapped ByteBuffers.
 	 */
-	static final class BufferSlicer {
+	static final class BufferSlicer implements BoundedData.Reader {
 
-		/** The reader/decoder to the memory mapped region with the data we currently read from.
+		/** The memory mapped region we currently read from.
 		 * Max 2GB large. Further regions may be in the {@link #furtherData} field. */
-		private BufferToByteBuffer.Reader data;
+		private ByteBuffer data;
```
zhijiangW: it might be better to name `currentData` comparing with `furtherData`

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BufferReaderWriterUtil.java
StephanEwen: I was looking into this as well while writing. I think I found the docs saying that the file channel implementation does guarantee this, but I can double check and add a comment.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/FileChannelMemoryMappedBoundedData.java
StephanEwen: True, could be private at the moment.
My idea was to have it accessible for tests.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/ResultPartitionFactory.java
StephanEwen: I am not confident enough in any other implementation, yet.
I only added them for our testing and benchmarking.

I am also always unsure about adding too many configurable parts for users. For API / behavior, we need customizability to fit the application and setup needs.

For runtime and performance, I tend to favor one opinionated way that works well across the board. If there is another option that is 10% faster in 10% of the cases, I am not sure users benefit too much from that, but the added configuration makes it overall harder to understand and predict.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/FileChannelBoundedData.java
StephanEwen: The next read should happen when the async write call is done (via channel future in netty).
The Flink 1.8 implementation has a similar assumption.
Also, without this assumption we need a theoretically unbounded buffer pool, which is not nice.

It should work in theory, but I am not 100% confident in practice. 
One more reason that I want to not expose this to users just now.

StephanEwen:
@zhijiangW Thanks, these are very good comments. Some answers in line.

> Maybe we should not always use the FILE_MMAP type as default. It could give an option to config or system auto selects the proper type internally based on some factors. E.g. via automatically checking whether use 64 bit JRE, if not only small size file fitting into virtual address space could use MMAP. For 32 bit JRE, FILE_FILE might be a better choice.

That is a fair point. We could do that.
Just out of curiosity: Do you know how many people use 32bit JVMs still?

>n Java has currently the limitation of not being able to unmap files from user code. Due to this bug ("http://bugs.sun.com/bugdatabase/view_bug.do?bug_id=4724038") in Sun's JRE, the files are unmapped when GC releases the byte buffers. We could supply the workaround mentioned in the bug report (see {@link #setUseUnmap}), which may fail on non-Oracle/OpenJDK JVMs. It forcefully unmaps the buffer on close by using an undocumented internal cleanup functionality. We could also check automatically whether this way is supported, then it might also be another factor to decide whether use FILE_MMAP by default.

The code explicitly unmaps files using a utility from Netty, see https://github.com/apache/flink/pull/8880/files#diff-94d133764660d5d8372ca24c08257e25R138
This should work, the JVM actually segfaults if you use the buffer after that.

> Currently we use Integer.MAX_VALUE as every mmap region size. If we do not want to make it configurable because users might also do not know how to tune it, we might automatically select different size based on whether 32 or 64 bit JRE. Because on 32 bit platform the address space can be very fragmented, so large size might not be mapped. Using a lower region size makes the implementation a little bit slower, but the chance is slower that mmap fails. As I know, lucene uses the size of 1 << 30 for 64 bit, and 1 << 28 for 32 bit.

If we use `FILE` mode for 32 bit JVMs, it seems that this would be solved already.

> In current FILE_MMAP implementation, we would always mmap a region after writing Integer.MAX_VALUE data size. This is simple way because we do not need to consider the partial data margin. But if the produced partitions could not be consumed in time limited by resource, I am not sure whether this pre-mmap has other downsides. E.g. one map produces 5000 subpartitions, but only one reduce task could be deployed at a time limited by resource. Another option is lazy mmap that means establishing the mmap after the partition is requested. But it seems complicated for handling the region margin to avoid partial header or data.

Do we know if the eager mapping has downsides? There is no eager pre-fetching, so there should be no wasted I/O. It would keep the virtual memory page table occupied for longer (is that an issue?) and it might incentivize the kernel to keep that data cached in memory longer, and evict other non-mapped files from the OS file cache instead (sounds like an advantage, could it have a downside?).

Eager mapping simplifies the implementation because it avoids the need to synchronize the lazy mapping when creating multiple readers concurrently.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BufferReaderWriterUtil.java
StephanEwen: I am adding a safety net. Seems cheap enough, and will put us on the safe side.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/FileChannelMemoryMappedBoundedData.java
zhijiangW: I also thought so before, so it might add @VisiableForTesting

zhijiangW:
Thanks for the explanations @StephanEwen !

As for how many people use 32bit JVM, I have not surveyed on it to give a ratio. I think the 64bit should be the mainstream in common sense. But we also do not want to bring a potential bad experience for 32bit users.

As for the downsides in eager mapping, I think it might be based on the assumption that cached data is hot to be read soon, so it seems reasonable to do so. But if this assumption is not true,  then the priority for caching these data in OS level might be misleading. I am not very sure whether it would bring downsides here, just point out a reminder here. And I agree that the current eager way would simplify the implementation much.

StephanEwen:
Addressed the comments and added the distinction based on memory architecture.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/FileChannelBoundedData.java
wsry: There should be problems in practice. However, now that the FILE mode is not offered to users, we can solve the problem latter.

StephanEwen:
@zhijiangW and @wsry I would merge this now unless you have further comments I should address first.

zhijiangW:
No other concerns on my side atm. LGTM!

StephanEwen:
All right, merging now...
Thanks @wsry and @zhijiangW for checking and spotting the issues.

StephanEwen:
Manually merged in d88882cc6db523a4fe2d70adad99bb66dc5d48c1

PR #8882, opened 2019-06-25
[FLINK-12965][table][hive] unify catalog view implementations

bowenli86:

This PR unified implementations of `CatalogView`.

## Brief change log

- unifies `GenericCatalogView`, `HiveCatalogView`, `AbstractCatalogView` into `CatalogViewImpl`
- updates unit tests

## Verifying this change

This change added tests and can be verified as follows: `GenericInMemoryCatalogTest`, `HiveCatalogGenericMetadataTest`, `HiveCatalogHiveMetadataTest`

  - The public API, i.e., is any changed class annotated with `@Public(Evolving)`:(no)
  - The serializers: (no)
  - The S3 file system connector:(no)

## Documentation


bowenli86:

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/QueryOperationCatalogView.java
```diff
@@ -28,7 +28,7 @@
  * A view created from a {@link QueryOperation} via operations on {@link org.apache.flink.table.api.Table}.
  */
 @Internal
-public class QueryOperationCatalogView extends AbstractCatalogView {
+public class QueryOperationCatalogView extends CatalogViewImpl {
```
xuefuz: Previously QueryOperationCatalogView, HiveCatalogView, and GenericCatalogView were children of AbstractCatalogView. Now we combine the latter two, and make the first one as a child of the combo, which might be a concern. I'm not sure if this posts any issue, but maybe we can keep the relationship unchanged?

bowenli86: Good point! I missed this since its counterpart of table is named `ConnectorCatalogTable`, maybe this class should be named `ConnectorCatalogView` :) .... nvm

I will restore AbstractCatalogView

bowenli86:

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/AbstractCatalogView.java
```diff
@@ -43,8 +43,12 @@
 	private final Map<String, String> properties;
 	private final String comment;
 
-	public AbstractCatalogView(String originalQuery, String expandedQuery, TableSchema schema,
-			Map<String, String> properties, String comment) {
+	public AbstractCatalogView(
```
xuefuz: nit: maybe we can restore the file completely.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/QueryOperationCatalogView.java
xuefuz: I think we need to revert changes to this file.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/AbstractCatalogView.java
bowenli86: you are right, I should have reverted unrelated changes

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/QueryOperationCatalogView.java
bowenli86: right, how could I forget that...

PR #8883, opened 2019-06-25
[FLINK-12988][table] restore AbstractCatalogTable as parent for Catal…

bowenli86:
…ogTableImpl and ConnectorCatalogTable

After doing #8815, Xuefu brought up a concern that it may not reflect the correct inheritance structure of CatalogTableImpl and ConnectorCatalogTable (which inherits from CatalogTableImpl now) anymore.

Thus we propose to add back AbstractCatalogTable from FLINK-12918. But the unification of catalog table implementations remain untouch

## Brief change log

- add back AbstractCatalogTable 

## Verifying this change

This change added tests and can be verified as follows:  GenericInMemoryCatalogTest, HiveCatalogGenericMetadataTest, HiveCatalogHiveMetadataTest

  - The serializers: (no)
  - Anything that affects deployment or recovery: JobManager (and its components), Checkpointing, Yarn/Mesos, ZooKeeper:(no)
  - The S3 file system connector:(no)

## Documentation

### DIFF
--- a/flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/AbstractCatalogTable.java
+++ b/flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/AbstractCatalogTable.java
@@ -19,63 +19,66 @@
 package org.apache.flink.table.catalog;
 
 import org.apache.flink.table.api.TableSchema;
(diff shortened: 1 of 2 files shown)


bowenli86:
cc @xuefuz

PR #8886, opened 2019-06-25
[FLINK-12987][metrics] fix DescriptiveStatisticsHistogram#getCount() not returning the number of elements seen

NicoK:

`DescriptiveStatisticsHistogram#getCount()` only returned the number of currently
stored elements, not the total number seen. This is different from the docs in `Histogram` and the `DropwizardHistogramWrapper` implementation and fixed by this PR.

-> follow-up PR: #8877

## Brief change log

- extend `HistogramStatistics` with an additional `AtomicLong` to count the number of elements seen

## Verifying this change

- add an abstract unit test for verifying any `Histogram` implementation based on
the previous test from `DropwizardFlinkHistogramWrapperTest`
- create `DescriptiveStatisticsHistogramTest` to run this test with `DescriptiveStatisticsHistogram`

  - Dependencies (does it add or upgrade a dependency): **yes** (test-jar, test-scope `flink-metrics` -> `flink-runtime`)
  - The serializers: **no**

## Documentation

### DIFF
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/metrics/DescriptiveStatisticsHistogram.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/metrics/DescriptiveStatisticsHistogram.java
@@ -29,18 +29,21 @@ public class DescriptiveStatisticsHistogram implements org.apache.flink.metrics.
 
 	private final DescriptiveStatistics descriptiveStatistics;
 
+	private long elementsSeen = 0L;
+
 	public DescriptiveStatisticsHistogram(int windowSize) {
 		this.descriptiveStatistics = new DescriptiveStatistics(windowSize);
 	}
 
 	@Override
 	public void update(long value) {
+		elementsSeen += 1L;
 		this.descriptiveStatistics.addValue(value);
 	}
 
 	@Override
 	public long getCount() {
-		return this.descriptiveStatistics.getN();
+		return this.elementsSeen;
 	}
 
 	@Override
--- a/flink-metrics/flink-metrics-core/pom.xml
+++ b/flink-metrics/flink-metrics-core/pom.xml
@@ -34,6 +34,15 @@ under the License.
 
 	<packaging>jar</packaging>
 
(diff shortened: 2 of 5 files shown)


flink-runtime/src/main/java/org/apache/flink/runtime/metrics/DescriptiveStatisticsHistogram.java
```diff
@@ -22,25 +22,30 @@
...
+import java.util.concurrent.atomic.AtomicLong;
+
 /**
  * The {@link DescriptiveStatisticsHistogram} use a DescriptiveStatistics {@link DescriptiveStatistics} as a Flink {@link Histogram}.
  */
 public class DescriptiveStatisticsHistogram implements org.apache.flink.metrics.Histogram {
 
 	private final DescriptiveStatistics descriptiveStatistics;
 
+	private final AtomicLong elementsSeen = new AtomicLong();
```
StephanEwen: Is this class assumed to be use in a multi-threaded context?

NicoK: that depends on how users use it - usually, there should be >=1 writer thread(s) (tasks) and 1 reader (the metrics reporter)

zentol: All metrics are read in a multi-thread context; the reporter thread and MetricQueryService can run concurrently.

zentol: I would not try to handle concurrent write access btw. (users are responsible for ensuring single-threaded access)

NicoK: Seems like the underlying code is not thread-safe anyways and we provide only best-effort metrics. I reverted to just using a `long` now.

PR #8887, opened 2019-06-25
[FLINK-12982][metrics] improve DescriptiveStatisticsHistogramStatistics performance

NicoK:

Instead of redirecting `DescriptiveStatisticsHistogramStatistics` calls to `DescriptiveStatistics`, it now takes a point-in-time snapshot using an own `UnivariateStatistic` implementation that
1. calculates min, max, mean, and standard deviation in one go (as opposed to four iterations over the values array!)
2. caches pivots for the percentile calculation to speed up retrieval of multiple percentiles/quartiles

As a result, this roughly increases value retrieval performance by 120% when accessing typical statistics in a metrics reporter, e.g. the InfluxDB reporter: count, min, max, mean, stddev, p50, p75, p95, p98, p99, p999.

Performance results:
1. only adding values to the histogram
```
    Benchmark                                       Mode  Cnt      Score        Error   Units
    HistogramBenchmarks.dropwizardHistogramAdd     thrpt   30   47985.359 ±    25.847  ops/ms
    HistogramBenchmarks.descriptiveHistogramAdd    thrpt   30   70158.792 ±   276.858  ops/ms
-> this PR:
    HistogramBenchmarks.descriptiveHistogramAdd    thrpt   30   75303.040 ±   475.355  ops/ms
```
2. after adding each value, also retrieving a common set of metrics:
```
    Benchmark                                       Mode  Cnt      Score        Error   Units
    HistogramBenchmarks.dropwizardHistogram        thrpt   30     400.274 ±     4.930  ops/ms
    HistogramBenchmarks.descriptiveHistogram       thrpt   30     124.533 ±     1.060  ops/ms
-> this PR:
    HistogramBenchmarks.descriptiveHistogram       thrpt   30     251.895 ±     1.809  ops/ms
```

Please note that this PR is based on #8884.
-> follow-up PR: #8888 

## Brief change log

- create a custom `CommonMetricsSnapshot implements UnivariateStatistic`
-- calculates min, max, mean, and stddev in one go
-- caches values inside a percentile implementation for optimised retrieval of multiple percentiles 
- create a point-in-time histogram snapshot during creation of `DescriptiveStatisticsHistogramStatistics`

## Verifying this change

  - The serializers: **no**

## Documentation

### DIFF
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/metrics/DescriptiveStatisticsHistogramStatistics.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/metrics/DescriptiveStatisticsHistogramStatistics.java
@@ -19,53 +19,183 @@ package org.apache.flink.runtime.metrics;
 
 import org.apache.flink.metrics.HistogramStatistics;
 
+import org.apache.commons.math3.exception.MathIllegalArgumentException;
 import org.apache.commons.math3.stat.descriptive.DescriptiveStatistics;
+import org.apache.commons.math3.stat.descriptive.UnivariateStatistic;
+import org.apache.commons.math3.stat.descriptive.moment.SecondMoment;
+import org.apache.commons.math3.stat.descriptive.moment.StandardDeviation;
+import org.apache.commons.math3.stat.descriptive.rank.Percentile;
+import org.apache.commons.math3.stat.ranking.NaNStrategy;
 
 import java.util.Arrays;
 
 /**
  * DescriptiveStatistics histogram statistics implementation returned by {@link DescriptiveStatisticsHistogram}.
- * The statistics class wraps a {@link DescriptiveStatistics} instance and forwards the method calls accordingly.
+ *
+ * <p>The statistics takes a point-in-time snapshot of a {@link DescriptiveStatistics} instance and
+ * allows optimised metrics retrieval from this.
  */
 public class DescriptiveStatisticsHistogramStatistics extends HistogramStatistics {
-	private final DescriptiveStatistics descriptiveStatistics;
+	private final CommonMetricsSnapshot statisticsSummary = new CommonMetricsSnapshot();
 
 	public DescriptiveStatisticsHistogramStatistics(DescriptiveStatistics latencyHistogram) {
-		this.descriptiveStatistics = latencyHistogram;
+		latencyHistogram.apply(statisticsSummary);
 	}
 
 	@Override
 	public double getQuantile(double quantile) {
-		return descriptiveStatistics.getPercentile(quantile * 100);
+		return statisticsSummary.getPercentile(quantile * 100);
 	}
 
 	@Override
 	public long[] getValues() {
-		return Arrays.stream(descriptiveStatistics.getValues()).mapToLong(i -> (long) i).toArray();
+		return Arrays.stream(statisticsSummary.getValues()).mapToLong(i -> (long) i).toArray();
 	}
 
 	@Override
 	public int size() {
-		return (int) descriptiveStatistics.getN();
+		return (int) statisticsSummary.getCount();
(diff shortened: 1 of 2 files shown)


flink-runtime/src/main/java/org/apache/flink/runtime/metrics/DescriptiveStatisticsHistogramStatistics.java
```diff
@@ -19,53 +19,174 @@
...
+	/**
+	 * Function to extract several commonly used metrics in an optimised way, i.e. with as few runs
+	 * over the data / calculations as possible.
+	 *
+	 * <p>Note that calls to {@link #evaluate(double[])} or {@link #evaluate(double[], int, int)}
+	 * will not return a value but instead populate this class so that further values can be
+	 * retrieved from it.
+	 */
+	private static class CommonMetricsSnapshot implements UnivariateStatistic {
+		long count = 0;
```
zentol: these should all be private since we access them through a getter anyway

flink-runtime/src/main/java/org/apache/flink/runtime/metrics/DescriptiveStatisticsHistogramStatistics.java
```diff
@@ -19,53 +19,174 @@
...
+		double min = Double.NaN;
+		double max = Double.NaN;
+		double mean = Double.NaN;
+		double stddev = Double.NaN;
+		private Percentile percentilesImpl = new Percentile().withNaNStrategy(NaNStrategy.FIXED);
```
zentol: can be final

flink-runtime/src/main/java/org/apache/flink/runtime/metrics/DescriptiveStatisticsHistogramStatistics.java
```diff
@@ -19,53 +19,174 @@
...
+	}
+
+	/**
+	 * Calculates min, max, mean (first moment), as well as the second moment in one go over
+	 * the value array.
+	 */
+	private static class SimpleStats extends SecondMoment {
+		private static final long serialVersionUID = 1L;
+
+		double min = Double.NaN;
```
zentol: private

flink-runtime/src/main/java/org/apache/flink/runtime/metrics/DescriptiveStatisticsHistogramStatistics.java
```diff
@@ -19,53 +19,174 @@
...
+			this.min = secondMoment.getMin();
+			this.max = secondMoment.getMax();
+
+			this.stddev = new StandardDeviation(secondMoment).getResult();
+
+			return Double.NaN;
+		}
+
+		@Override
+		public CommonMetricsSnapshot copy() {
```
zentol: is this method actually used somewhere?

NicoK: We need to implement this from the interface and honestly don't know whether this maybe used in, for example this:
```
	public DescriptiveStatisticsHistogramStatistics(DescriptiveStatistics latencyHistogram) {
		latencyHistogram.apply(statisticsSummary);
	}
```

But good point, I forgot to override this for `SimpleStats`.

flink-runtime/src/main/java/org/apache/flink/runtime/metrics/DescriptiveStatisticsHistogramStatistics.java
NicoK: cannot be due to the `copy()` method

NicoK:
FYI: I updated the code according to your comments and my findings

PR #8888, opened 2019-06-25
[FLINK-12983][metrics] replace descriptive histogram's storage back-end

NicoK:

Instead of using `DescriptiveStatistics` and its `ResizableDoubleArray` storage back-end, implement an own circular array based on a fixed-size double array that wraps around. This significantly improves the time needed to add new values. It also ensures that access is synchronized which `ResizableDoubleArray` does not guarantee.

Performance results:
1. only adding values to the histogram
```
    Benchmark                                       Mode  Cnt      Score        Error   Units
    HistogramBenchmarks.dropwizardHistogramAdd     thrpt   30   47985.359 ±    25.847  ops/ms
    HistogramBenchmarks.descriptiveHistogramAdd    thrpt   30   70158.792 ±   276.858  ops/ms
-> #8887:
    HistogramBenchmarks.descriptiveHistogramAdd    thrpt   30   75303.040 ±   475.355  ops/ms
-> this PR:
    HistogramBenchmarks.descrHistogramCircularAdd  thrpt   30  200906.902 ±   384.483  ops/ms
-> using a linear array of twice the histogram size (and moving values back to the start once the window reaches the end), for reference:
    HistogramBenchmarks.descrHistogramLinearAdd    thrpt   30  189788.728 ±   233.283  ops/ms
```
2. after adding each value, also retrieving a common set of metrics:
```
    Benchmark                                       Mode  Cnt      Score        Error   Units
    HistogramBenchmarks.dropwizardHistogram        thrpt   30     400.274 ±     4.930  ops/ms
    HistogramBenchmarks.descriptiveHistogram       thrpt   30     124.533 ±     1.060  ops/ms
-> #8887:
    HistogramBenchmarks.descriptiveHistogram       thrpt   30     251.895 ±     1.809  ops/ms
-> this PR:
    HistogramBenchmarks.descrHistogramCircular     thrpt   30     301.068 ±     2.077  ops/ms
-> using a linear array as above:
    HistogramBenchmarks.descrHistogramLinear       thrpt   30     234.050 ±     5.485  ops/ms
```

Please note that this PR is based on #8887.

## Brief change log

- add `CircularDoubleArray` to `DescriptiveStatisticsHistogram`
- remove extra `AtomicLong elementsSeen` (can be done more efficiently in `CircularDoubleArray`)

## Verifying this change

  - The serializers: **no**
  - The runtime per-record code paths (performance sensitive): **maybe** (histograms are used by latency markers only but may be used by users)

## Documentation

### DIFF
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/metrics/DescriptiveStatisticsHistogram.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/metrics/DescriptiveStatisticsHistogram.java
@@ -27,27 +27,63 @@ import org.apache.commons.math3.stat.descriptive.DescriptiveStatistics;
  */
 public class DescriptiveStatisticsHistogram implements org.apache.flink.metrics.Histogram {
 
-	private final DescriptiveStatistics descriptiveStatistics;
-
-	private long elementsSeen = 0L;
+	private final CircularDoubleArray descriptiveStatistics;
 
 	public DescriptiveStatisticsHistogram(int windowSize) {
-		this.descriptiveStatistics = new DescriptiveStatistics(windowSize);
+		this.descriptiveStatistics = new CircularDoubleArray(windowSize);
 	}
 
 	@Override
 	public void update(long value) {
-		elementsSeen += 1L;
 		this.descriptiveStatistics.addValue(value);
 	}
 
 	@Override
 	public long getCount() {
-		return this.elementsSeen;
+		return this.descriptiveStatistics.getElementsSeen();
 	}
 
 	@Override
 	public HistogramStatistics getStatistics() {
 		return new DescriptiveStatisticsHistogramStatistics(this.descriptiveStatistics);
+	}
+
+	/**
+	 * Fixed-size array that wraps around at the end and has a dynamic start position.
+	 */
+	static class CircularDoubleArray {
+		private final double[] backingArray;
+		private int nextPos = 0;
+		private boolean fullSize = false;
+		private long elementsSeen = 0;
+
+		CircularDoubleArray(int windowSize) {
+			this.backingArray = new double[windowSize];
+		}
+
+		synchronized void addValue(double value) {
+			backingArray[nextPos] = value;
+			++elementsSeen;
+			++nextPos;
+			if (nextPos == backingArray.length) {
+				nextPos = 0;
+				fullSize = true;
+			}
+		}
+
+		synchronized double[] toUnsortedArray() {
+			final int size = getSize();
+			double[] result = new double[size];
+			System.arraycopy(backingArray, 0, result, 0, result.length);
+			return result;
+		}
+
+		private synchronized int getSize() {
+			return fullSize ? backingArray.length : nextPos;
+		}
+
+		private synchronized long getElementsSeen() {
+			return elementsSeen;
+		}
 	}
 }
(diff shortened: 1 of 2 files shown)


flink-runtime/src/main/java/org/apache/flink/runtime/metrics/DescriptiveStatisticsHistogram.java
```diff
@@ -27,27 +27,63 @@
...
+			if (nextPos == backingArray.length) {
+				nextPos = 0;
+				fullSize = true;
+			}
+		}
+
+		synchronized double[] toUnsortedArray() {
+			final int size = getSize();
+			double[] result = new double[size];
+			System.arraycopy(backingArray, 0, result, 0, result.length);
```
zentol: I'm wondering whether the order could be relevant to _someone_. How expensive would it be to keep the insertion order (i.e. by adding a second arraycopy call)?

NicoK: `CircularDoubleArray` is a package-private class - I'm wondering who/which component would require a sorted array then. This circular array has special APIs (making clear that we return an unsorted array) and is only used by `org.apache.flink.runtime.metrics.DescriptiveStatisticsHistogramStatistics.CommonMetricsSnapshot` which does not require sorting. Whether sorting is required is basically defined via `org.apache.flink.runtime.metrics.DescriptiveStatisticsHistogramStatistics.CommonMetricsSnapshot`.

Out of curiosity, I added this method and used it instead:
```
		public synchronized double[] toArray() {
			if (fullSize) {
				final double[] result = new double[backingArray.length];
				final int firstLength = result.length - nextPos;
				System.arraycopy(backingArray, nextPos, result, 0, firstLength);
				System.arraycopy(backingArray, 0, result, firstLength, nextPos);
				return result;
			} else {
				final double[] result = new double[nextPos];
				System.arraycopy(backingArray, 0, result, 0, nextPos);
				return result;
			}
		}
```

First quick results are as follows and show the benefit for not working with a sorted array if we don't need it (and are requesting the histogram a lot):
```
Flink 1.9
Benchmark                                     Mode  Cnt      Score     Error   Units
HistogramBenchmarks.descriptiveHistogram     thrpt   30     89.224 ±   1.974  ops/ms
HistogramBenchmarks.descriptiveHistogramAdd  thrpt   30  56034.903 ± 939.263  ops/ms

Flink 1.10 + FLINK-12983
Benchmark                                     Mode  Cnt       Score      Error   Units
HistogramBenchmarks.descriptiveHistogram     thrpt   30     280.240 ±    3.747  ops/ms
HistogramBenchmarks.descriptiveHistogramAdd  thrpt   30  207176.894 ± 2417.831  ops/ms

Flink 1.10 + FLINK-12983 + sortedArray
Benchmark                                     Mode  Cnt       Score      Error   Units
HistogramBenchmarks.descriptiveHistogram     thrpt   30     239.000 ±    1.506  ops/ms
HistogramBenchmarks.descriptiveHistogramAdd  thrpt   30  210352.180 ± 1135.394  ops/ms
```

zentol: I was more thinking of users who use `HistogramStatistics#getValues`; but we don't provide any ordering guarantees there anyway.

PR #8889, opened 2019-06-25
[FLINK-12934][hive] add hadoop dependencies as 'flink-shaded-hadoop-2-uber' for flink-connector-hive to connect to remote hive metastore service

bowenli86:

Now, for all hadoop related dependencies, we just use the single jar of flink-shaded-hadoop-2-uber, instead of depending on individual hadoop jars. It's much more convenient for users to use now.

To summarize all flink-connector-hive's dependencies, which are all of provided scope:

hive-metastore
hive-exec
flink-shaded-hadoop-2-uber

## Brief change log

- adds a few more dependencies for HiveCatalog in flink-connector-hive to connect to a remote hive metastore service

## Verifying this change

We need to add end-to-end test in order to cover such work. cc @lirui-apache @zjuwangg

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-connectors/flink-connector-hive/pom.xml
+++ b/flink-connectors/flink-connector-hive/pom.xml
@@ -35,6 +35,13 @@ under the License.
 
 	<packaging>jar</packaging>
 
+	<properties>
+		<servlet-api.version>2.5</servlet-api.version>
+		<commons-configuration.version>1.6</commons-configuration.version>
+		<commons-logging.version>1.2</commons-logging.version>
+	</properties>
+
+
 	<dependencies>
 
 		<!-- core dependencies -->
@@ -62,6 +69,13 @@ under the License.
 
 		<!-- Hadoop dependency -->
 		<!-- Hadoop as provided dependencies, so we can depend on them without pulling in Hadoop -->
+
+		<dependency>
+			<groupId>org.apache.hadoop</groupId>
+			<artifactId>hadoop-auth</artifactId>
+			<version>${hivemetastore.hadoop.version}</version>
+			<scope>provided</scope>
+		</dependency>
 
 		<dependency>
 			<groupId>org.apache.hadoop</groupId>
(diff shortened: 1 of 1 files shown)


bowenli86:

bowenli86:
@xuefuz good catch! Will add them

bowenli86:
@xuefuz after considering it carefully, I decided to make them as `provided` dependency for now as flink-connector-hive doesn't necessarily need them for either compilation or testing. I added them as dependencies more as a reminder to users and developers that they need to add them into their classpath when running in production

flink-connectors/flink-connector-hive/pom.xml
```diff
@@ -452,29 +493,15 @@ under the License.
...
+									<include>*:*</include>
 								</includes>
 							</artifactSet>
 							<promoteTransitiveDependencies>true</promoteTransitiveDependencies>
 							<!-- DO NOT RELOCATE GUAVA IN THIS PACKAGE -->
 							<filters>
 								<filter>
-									<!-- some dependencies bring their own LICENSE.txt which we don't need -->
 									<artifact>*:*</artifact>
-									<excludes>
```
xuefuz: How come we don't need those any more?

bowenli86: because we don't package any hive/hadoop dependencies anymore a long time ago. They are all provided, so we don't need to filter anything for them. This is a kind of historical leftover, and can be on its own PR if necessary.

xuefuz:
Thanks for the update. Looking at the required libraries, I'm wondering if it's reasonable to expect user to find those very specific libraries and put them in the classpath. The other question would be why only those are missing. 

For usability, it seems fine to me if we ask user to put flink-hive jar in the classpath. However, asking for a very specific jar like commons-logging seems having too much usability overhead.

flink-connectors/flink-connector-hive/pom.xml
xuefuz: This is fine.

bowenli86:
@xuefuz I've updated the dependency. Now, for all hadoop related dependencies, we just using the single jar of flink-shaded-hadoop-2-uber, instead of depending on individual hadoop jars. It's much more convenient for users to use now.

To summarize all flink-connector-hive's dependencies, which are all of `provided` scope:

- hive-metastore
- hive-exec
- flink-shaded-hadoop-2-uber

lirui-apache:
Do we know what's inside the shaded jar with this change? I think we'd better check the included transitive deps to see if any of them needs relocation.

bowenli86:
> Do we know what's inside the shaded jar with this change? I think we'd better check the included transitive deps to see if any of them needs relocation.

you can check it here https://github.com/apache/flink-shaded/blob/master/flink-shaded-hadoop-2-uber/pom.xml

PR #8890, opened 2019-06-26
[FLINK-12989][hive] Generate HiveTableSink from from a Hive table

xuefuz:

Add the generation of HiveTableSink from from a Hive table

## Brief change log

*(for example:)*
  - Added the logic of generation of HiveTableSink from from a Hive table
  - Refactored and simplified HiveTableSink related code
  - Added test for the logic

## Verifying this change

This change is verified with existing and new test cases.

  - Dependencies (does it add or upgrade a dependency): (yno)
  - The serializers: (yes )

## Documentation

### DIFF
--- a/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/batch/connectors/hive/HiveTableFactory.java
+++ b/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/batch/connectors/hive/HiveTableFactory.java
@@ -20,6 +20,7 @@ package org.apache.flink.batch.connectors.hive;
 
 import org.apache.flink.table.catalog.CatalogTable;
 import org.apache.flink.table.catalog.CatalogTableImpl;
+import org.apache.flink.table.catalog.ObjectPath;
 import org.apache.flink.table.catalog.config.CatalogConfig;
 import org.apache.flink.table.factories.TableFactoryUtil;
 import org.apache.flink.table.factories.TableSinkFactory;
@@ -70,14 +71,14 @@ public class HiveTableFactory implements TableSourceFactory<Row>, TableSinkFacto
 	}
 
 	@Override
-	public TableSource<Row> createTableSource(CatalogTable table) {
+	public TableSource<Row> createTableSource(ObjectPath tablePath, CatalogTable table) {
 		Preconditions.checkNotNull(table);
 		Preconditions.checkArgument(table instanceof CatalogTableImpl);
 
 		boolean isGeneric = Boolean.valueOf(table.getProperties().get(CatalogConfig.IS_GENERIC));
 
 		if (!isGeneric) {
-			return createInputFormatTableSource(table);
+			return createInputFormatTableSource(tablePath, table);
 		} else {
 			return TableFactoryUtil.findAndCreateTableSource(table);
 		}
@@ -86,20 +87,20 @@ public class HiveTableFactory implements TableSourceFactory<Row>, TableSinkFacto
 	/**
 	 * Creates and configures a {@link org.apache.flink.table.sources.InputFormatTableSource} using the given {@link CatalogTable}.
 	 */
-	private InputFormatTableSource<Row> createInputFormatTableSource(CatalogTable table) {
+	private InputFormatTableSource<Row> createInputFormatTableSource(ObjectPath tablePath, CatalogTable table) {
 		// TODO: create an InputFormatTableSource from a HiveCatalogTable instance.
 		return null;
 	}
 
 	@Override
-	public TableSink<Row> createTableSink(CatalogTable table) {
+	public TableSink<Row> createTableSink(ObjectPath tablePath, CatalogTable table) {
 		Preconditions.checkNotNull(table);
 		Preconditions.checkArgument(table instanceof CatalogTableImpl);
 
 		boolean isGeneric = Boolean.valueOf(table.getProperties().get(CatalogConfig.IS_GENERIC));
 
 		if (!isGeneric) {
-			return createOutputFormatTableSink((CatalogTableImpl) table);
+			return createOutputFormatTableSink(tablePath, (CatalogTableImpl) table);
 		} else {
 			return TableFactoryUtil.findAndCreateTableSink(table);
 		}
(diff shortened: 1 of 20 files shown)


xuefuz:
Please note that this PR included the changes in PR #8883. cc: @bowenli86

zjuwangg:
better to rename the pr title

flink-table/flink-table-common/src/main/java/org/apache/flink/table/catalog/ObjectPath.java
```diff
@@ -28,8 +31,19 @@
  * A database name and object (table/view/function) name combo in a catalog.
  */
 public class ObjectPath {
-	private final String databaseName;
-	private final String objectName;
+	// Default database name
```
lirui-apache: How about just make ObjectPath a Serializable?

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/CatalogTableImpl.java
```diff
@@ -32,65 +32,39 @@
...
-		this(tableSchema, new ArrayList<>(), properties, comment);
+			ObjectPath tablePath,
+			TableSchema tableSchema,
+			Map<String, String> properties,
+			String comment) {
+		this(tablePath, tableSchema, new ArrayList<>(), properties, comment);
 	}
 
 	public CatalogTableImpl(
+			ObjectPath tablePath,
```
lirui-apache: If a CatalogTable can have an ObjectPath field, do we need some validation in APIs like `HiveCatalog::createTable`?
And btw, why did we decide to separate CatalogTable and ObjectPath in the first place?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/batch/connectors/hive/HiveTableFactory.java
```diff
@@ -31,13 +32,22 @@
...
 import java.util.List;
 import java.util.Map;
 
+import static org.apache.flink.util.Preconditions.checkNotNull;
+
 /**
  * A table factory implementation for tables stored in Hive catalog.
  */
 public class HiveTableFactory implements TableSourceFactory<Row>, TableSinkFactory<Row> {
+	private HiveCatalog hiveCatalog;
```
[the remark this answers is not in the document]
xuefuz: True. In that case, I think we can just pass on HiveConf instead.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/CatalogTableImpl.java
xuefuz: I added some validation.

As to separation, CatalogTable as API doesn't require ObjectPath, as it's consistent how objects are organized in databases. However, as an impl class such as CatalogTableImpl, it can has anything it needs without violating the API. One implementation might not need this, and we don't want to enforce this at the API level.

xuefuz:
Based on feedback, I have changed the design a bit so we don't store table path in catalog table any more. Instead, we added ObjectPath in TableSource/SinkFactory APIs. This simplified and reduced the changes.

@lirui-apache @bowenli86 pls take another look. thanks.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/batch/connectors/hive/HiveTableSink.java
```diff
@@ -55,30 +58,31 @@
...
-	private final String tableName;
-	private final List<String> partitionColumns;
 	private final String hiveVersion;
 
 	// TODO: need OverwritableTableSink to configure this
 	private boolean overwrite = false;
 
-	public HiveTableSink(JobConf jobConf, RowTypeInfo rowTypeInfo, String dbName, String tableName,
-			List<String> partitionColumns) {
+	public HiveTableSink(JobConf jobConf, ObjectPath tablePath, CatalogTableImpl table) {
```
lirui-apache: Seems we just need a CatalogTable here too

xuefuz:
> @xuefuz thanks for the update!
> 
> Just want to confirm if the approach also works for in-memory catalog, because currently users can create Hive tables in in-memory catalog. Would be better to work for both HiveCatalog (HiveTableFactory) and in-memory catalog. If we need extra work for it, it can be in the different PR.
> 
> Otherwise LGTM with a few minor comments.

Good point. I expect that GenericInMemoryCatalog will have its own table factory, which will need to process the request according to the actual table type in a similar fashion as in HiveTableFactory. However, that's a different thing with lower priority. We can do that as a followup.

flink-connectors/flink-connector-hive/src/test/java/org/apache/flink/batch/connectors/hive/HiveTableSinkTest.java
```diff
@@ -113,8 +114,8 @@ public void testInsertIntoDynamicPartition() throws Exception {
 		tableEnv.registerDataSet("src", execEnv.fromCollection(toWrite, rowTypeInfo));
 
 		Table hiveTable = hiveCatalog.getHiveTable(tablePath);
-		tableEnv.registerTableSink("destSink", new HiveTableSink(new JobConf(hiveConf), rowTypeInfo,
-				"default", "dest", HiveCatalog.getFieldNames(hiveTable.getPartitionKeys())));
+		CatalogBaseTable table = hiveCatalog.getTable(tablePath);
+		tableEnv.registerTableSink("destSink", new HiveTableSink(new JobConf(hiveConf), tablePath, (CatalogTableImpl) table));
```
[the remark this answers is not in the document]
xuefuz: HiveTableSink constructor requires a CatalogTableImpl instance.

flink-connectors/flink-connector-hive/src/test/java/org/apache/flink/batch/connectors/hive/HiveTableSinkTest.java
```diff
@@ -190,8 +191,8 @@ public void testWriteComplexType() throws Exception {
 
 		tableEnv.registerDataSet("nestedSrc", execEnv.fromCollection(toWrite, rowTypeInfo));
 		hiveTable = hiveCatalog.getHiveTable(tablePath);
-		tableEnv.registerTableSink("nestedSink", new HiveTableSink(new JobConf(hiveConf), rowTypeInfo,
-				"default", "dest", HiveCatalog.getFieldNames(hiveTable.getPartitionKeys())));
+		catalogTable = (CatalogTableImpl) hiveCatalog.getTable(tablePath);
```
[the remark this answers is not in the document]
xuefuz: Same as above.

bowenli86:
Merging

PR #8892, opened 2019-06-26
[FLINK-12990][python] Date type doesn't consider the local TimeZone

dianfu:

*Currently, the python DateType is converted by an `int` which indicates the days passed since 1970-1-1 and then the Java side will create a Java Date by call `new Date(days * 86400)`. As we know that the Date constructor expected milliseconds since 1970-1-1 00:00:00 GMT and so we should convert `days * 86400` to GMT milliseconds. This PR will fix this issue.*

## Brief change log

  - *Add method getOffsetFromLocalMillis to deal with the offset calculation of the local timezone and use the calculated offset when creating the Date object.*

## Verifying this change

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-table/flink-table-planner/src/main/scala/org/apache/flink/table/util/python/PythonTableUtils.scala
+++ b/flink-table/flink-table-planner/src/main/scala/org/apache/flink/table/util/python/PythonTableUtils.scala
@@ -20,6 +20,8 @@ package org.apache.flink.table.util.python
 
 import java.nio.charset.StandardCharsets
 import java.sql.{Date, Time, Timestamp}
+import java.time.{LocalDate, LocalDateTime, LocalTime}
+import java.util.TimeZone
 import java.util.function.BiConsumer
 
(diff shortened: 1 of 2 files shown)


flink-table/flink-table-planner/src/main/scala/org/apache/flink/table/util/python/PythonTableUtils.scala
```diff
@@ -389,4 +394,26 @@ object PythonTableUtils {
         array
     }
   }
+
+  private def getOffsetFromLocalMillis(millisLocal: Long): Int = {
```
sunjincheng121: It's better to add the test case for this utils method, What do you think?

dianfu:
@sunjincheng121 Makes sense and I will add some test cases.

dianfu:
@sunjincheng121 Have updated the PR adding the test cases. :)

PR #8893, opened 2019-06-26
[hotfix][python] Align the signature of type utility methods with Java

dianfu:

*Currently, the python method `DataTypes.VARCHAR(length)` assumes default value `1` for argument `length` and we should align with Java and don't provide default value for it as currently the length is not respected in Planner and users will easily write programs which can run in the current version and won't run in the future versions after the argument `length` is respected.*

## Brief change log

  - *Remove the default value for the utilities methods of DataTypes*

## Verifying this change

This change is already covered by existing tests*.

  - The serializers: (no)

## Documentation


sunjincheng121:
@flinkbot approve all

PR #8894, opened 2019-06-26
 [FLINK-12961][datastream] Providing an internal execution method of StreamExecutionEnvironment accepting StreamGraph as input parameter

ifndef-SleePy:

* Expose an internal method `execute(StreamGraph)` of `StreamExecutionEnvironment`. So the pluggable runner could get a chance to set properties of `StreamGraph`.

## Brief change log

* Add a new abstract method `execute(StreamGraph)` of `StreamExecutionEnvironment`
* Make `execute(jobName)` of `StreamExecutionEnvironment` as an implementation since all subclasses have same logic

## Verifying this change

  - The serializers: no

## Documentation

### DIFF
--- a/flink-streaming-java/src/main/java/org/apache/flink/streaming/api/environment/StreamExecutionEnvironment.java
+++ b/flink-streaming-java/src/main/java/org/apache/flink/streaming/api/environment/StreamExecutionEnvironment.java
@@ -1502,7 +1502,23 @@ public abstract class StreamExecutionEnvironment {
 	 * @return The result of the job execution, containing elapsed time and accumulators.
 	 * @throws Exception which occurs during job execution.
 	 */
-	public abstract JobExecutionResult execute(String jobName) throws Exception;
+	public JobExecutionResult execute(String jobName) throws Exception {
+		Preconditions.checkNotNull(jobName, "Streaming Job name should not be null.");
+
+		return execute(getStreamGraph(jobName));
+	}
+
+	/**
+	 * Triggers the program execution. The environment will execute all parts of
+	 * the program that have resulted in a "sink" operation. Sink operations are
+	 * for example printing results or forwarding them to a message queue.
+	 *
+	 * @param streamGraph the stream graph representing the transformations
+	 * @return The result of the job execution, containing elapsed time and accumulators.
+	 * @throws Exception which occurs during job execution.
+	 */
+	@Internal
+	public abstract JobExecutionResult execute(StreamGraph streamGraph) throws Exception;
 
 	/**
 	 * Getter of the {@link org.apache.flink.streaming.api.graph.StreamGraph} of the streaming job.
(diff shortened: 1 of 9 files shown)


ifndef-SleePy:
@flinkbot attention @aljoscha

aljoscha:
Ah, wait a second, I might have discovered a new thing. I have to check how this will work together with a `StreamExecutionEnvironment` where a user also applied transformations, those would be ignored in the new `execute(StreamGraph)` call. This scenario can happen when a user mixes a Table API and DataStream program. Let me think.

aljoscha:
Ah, wait a second, I might have discovered a new thing. I have to check how this will work together with a `StreamExecutionEnvironment` where a user also applied transformations, those would be ignored in the new `execute(StreamGraph)` call. This scenario can happen when a user mixes a Table API and DataStream program. Let me think.

ifndef-SleePy:
@aljoscha
I think you have raised a good question!

AFAIK, Blink runner generates `StreamGraph` itself. And it also keeps transformations itself. When a `DataStream` is translated to a `Table`, the Blink batch runner back tracks the all transformations of `StreamExecutionEnvironment` through transformations kept by itself.

Although there is a corner case that if there exists a disconnected sub graph in `StreamExecutionEnvironment` which might not be touched by back tracking. I have not reached an agreement with my colleague about this case.

I'm wondering whether there is a better solution for Blink batch runner?

aljoscha:
I have this alternative proposal, which makes the schedule mode configurable on the `ExecutionConfig`.  This would let everything integrate, reuse the same `StreamGraph` generation logic. I think the schedule mode was the only thing that is not configurable that the Blink Planner would need.

ifndef-SleePy:
Abandon this PR since we think alternative approach is better.

ifndef-SleePy:
We have changed idea, will continue working on this. Rebase master to resolve conflicts.

ifndef-SleePy:
Address review comments.
BTW, trigger Travis checking again.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/environment/StreamExecutionEnvironment.java
```diff
@@ -1502,7 +1502,25 @@ public JobExecutionResult execute() throws Exception {
...
+
+		return execute(getStreamGraph(jobName));
+	}
+
+	/**
+	 * Triggers the program execution. The environment will execute all parts of
+	 * the program that have resulted in a "sink" operation. Sink operations are
+	 * for example printing results or forwarding them to a message queue.
+	 *
+	 * <p>The program execution will be logged and displayed with the provided name
```
aljoscha: Ah, whoever merged this should remove this line, because it refers to the name parameter which is not there anymore.

ifndef-SleePy: Ah, that's my fault. I will remove this line before merging since Travis checking is not finished yet.

aljoscha:
I merged this. 👌

PR #8895, opened 2019-06-26
[FLINK-12536][Runtime/Network]Make BufferOrEventSequence#getNext non blocking

klion26:

Currently it is non-blocking in case of credit-based flow control (default), however for SpilledBufferOrEventSequence it is blocking on reading from file. We might want to consider reimplementing it to be non blocking with CompletableFuture<?> isAvailable() method.

Otherwise we will block mailbox processing for the duration of reading from file - for example we will block processing time timers and potentially in the future network flushes.

## Brief change log

Use a ByteBuffer pool to read the BufferOrEvent asynchronous.
- the pool size if default 2, can be configurated by key `taskmanager.memory.async-load.buffer-count`
- will reuse the read thread in `IOManager`

## Verifying this change

This change is already covered by existing tests, such as:
- `SpilledBufferOrEventSequenceTest.java`

  - The serializers: (**no**)

## Documentation

### DIFF
--- a/flink-core/src/main/java/org/apache/flink/util/IOUtils.java
+++ b/flink-core/src/main/java/org/apache/flink/util/IOUtils.java
@@ -25,6 +25,8 @@ import java.io.InputStream;
 import java.io.OutputStream;
 import java.io.PrintStream;
 import java.net.Socket;
+import java.nio.ByteBuffer;
+import java.nio.channels.FileChannel;
 
 /**
  * An utility class for I/O related functionality.
@@ -136,6 +138,24 @@ public final class IOUtils {
 			}
 			toRead -= ret;
 			off += ret;
+		}
+	}
+
+	/**
+	 * Reads bytes into ByteBuffer, blocking until all ByteBuffer is full.
+	 *
+	 * @param fileChannel The source where we read from
+	 * @param buffer The target we'll write the bytes to.
+	 * @throws IOException If we could not read requested number of bytes for any reason (including EOF)
+	 */
+	public static void readFully(FileChannel fileChannel, ByteBuffer buffer) throws IOException {
+		int toRead = buffer.limit() - buffer.position();
+		while (toRead > 0) {
+			final int ret = fileChannel.read(buffer);
+			if (ret < 0) {
+				throw new IOException("Premeture EOF from FileChannel");
+			}
+			toRead -= ret;
 		}
 	}
 
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/io/disk/iomanager/BufferOrEventFileChannelReader.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/io/disk/iomanager/BufferOrEventFileChannelReader.java
@@ -1,0 +1,91 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.runtime.io.disk.iomanager;
+
+import org.apache.flink.runtime.io.network.buffer.Buffer;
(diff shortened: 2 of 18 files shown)


klion26:
Travis build failed seems irrelevant, have filed a [issue](https://issues.apache.org/jira/browse/FLINK-13009) to track it.

```
Test testVCoresAreSetCorrectlyAndJobManagerHostnameAreShownInWebInterfaceAndDynamicPropertiesAndYarnApplicationNameAndTaskManagerSlots(org.apache.flink.yarn.YARNSessionCapacitySchedulerITCase) failed with:
java.lang.NullPointerException: java.lang.NullPointerException
	at org.apache.hadoop.yarn.server.resourcemanager.rmapp.attempt.RMAppAttemptMetrics.getAggregateAppResourceUsage(RMAppAttemptMetrics.java:128)
	at org.apache.hadoop.yarn.server.resourcemanager.rmapp.attempt.RMAppAttemptImpl.getApplicationResourceUsageReport(RMAppAttemptImpl.java:900)
	at org.apache.hadoop.yarn.server.resourcemanager.rmapp.RMAppImpl.createAndGetApplicationReport(RMAppImpl.java:660)
	at org.apache.hadoop.yarn.server.resourcemanager.ClientRMService.getApplications(ClientRMService.java:930)
	at org.apache.hadoop.yarn.api.impl.pb.service.ApplicationClientProtocolPBServiceImpl.getApplications(ApplicationClientProtocolPBServiceImpl.java:273)
	at org.apache.hadoop.yarn.proto.ApplicationClientProtocol$ApplicationClientProtocolService$2.callBlockingMethod(ApplicationClientProtocol.java:507)
	at org.apache.hadoop.ipc.ProtobufRpcEngine$Server$ProtoBufRpcInvoker.call(ProtobufRpcEngine.java:447)
	at org.apache.hadoop.ipc.RPC$Server.call(RPC.java:989)
	at org.apache.hadoop.ipc.Server$RpcCall.run(Server.java:847)
	at org.apache.hadoop.ipc.Server$RpcCall.run(Server.java:790)
	at java.security.AccessController.doPrivileged(Native Method)
	at javax.security.auth.Subject.doAs(Subject.java:422)
	at org.apache.hadoop.security.UserGroupInformation.doAs(UserGroupInformation.java:1836)
	at org.apache.hadoop.ipc.Server$Handler.run(Server.java:2486)
```

flink-core/src/main/java/org/apache/flink/util/IOUtils.java
```diff
@@ -139,6 +141,24 @@ public static void readFully(final InputStream in, final byte[] buf, int off, fi
...
+	/**
+	 * Reads bytes into ByteBuffer, blocking until all ByteBuffer is full.
+	 *
+	 * @param fileChannel The source where we read from
+	 * @param buffer The target we'll write the bytes to.
+	 * @throws IOException If we could not read requested number of bytes for any reason (including EOF)
+	 */
+	public static void readFully(FileChannel fileChannel, ByteBuffer buffer) throws IOException {
+		int toRead = buffer.limit() - buffer.position();
+		while (toRead > 0) {
```
StephanEwen: This can be simplified to `while (buffer.hasRemaining())`

flink-runtime/src/main/java/org/apache/flink/runtime/io/disk/iomanager/IOManager.java
```diff
@@ -224,6 +238,8 @@ public void deleteChannel(FileIOChannel.ID channel) throws IOException {
 
 	public abstract BufferFileReader createBufferFileReader(FileIOChannel.ID channelID, RequestDoneCallback<Buffer> callback) throws IOException;
 
+	public abstract BufferFileReader createBufferOrEventFileReader(FileIOChannel.ID channelID, RequestDoneCallback<Buffer> callback) throws IOException;
```
StephanEwen: If we go forward, I would suggest to factor this out into an interface (that is also implemented by `IOManager`) so we do not pass such a big dependency. This component does not need to know about the IOManager as a whole if it is only interested in this sub-functionality.

flink-runtime/src/main/java/org/apache/flink/runtime/io/disk/iomanager/BufferOrEventFileChannelReader.java
```diff
@@ -0,0 +1,91 @@
...
+
+	public static void writeBufferOrEventMeta(ByteBuffer buffer, int channel, int size, byte isBuffer) {
+		buffer.putInt(channel);
+		buffer.putInt(size);
+		buffer.put(isBuffer);
+	}
+
+	private void readBufferOrEventMeta(ByteBuffer buffer) {
+		// ignore channel.
+		buffer.order(ByteOrder.LITTLE_ENDIAN);
```
StephanEwen: Passing the values through fields seems like passing information sideways or through side effects.
If you want to avoid object creation for the return type, I would suggest to look at splitting the methods differently.

klion26:
@StephanEwen thanks for the comments, I'll continue after the discussion has been solved.

klion26:
closed the pr, as this issue won't do.

PR #8896, opened 2019-06-26
[FLINK-12993][runtime] Refactor forceReleaseOnConsumption to JM concept

zentol:
The `forceReleaseOnConsumption` flag allows us to switch between the current behavior (all partitions are released on consumption) and the planned new behavior (only pipelined partitions are released on consumption).

This PR moves the processing of this flag from the TM side (where it was handled in the netty shuffle environment) to the JM side (now handled in the Execution when creating the ResultPartitionDeploymentDescriptor).

The corresponding config option is evaluated by the `ExecutionGraphBuilder`.

### DIFF
--- a/flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/ResultPartitionFactoryTest.java
+++ b/flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/ResultPartitionFactoryTest.java
@@ -39,23 +39,22 @@ import static org.hamcrest.MatcherAssert.assertThat;
 public class ResultPartitionFactoryTest extends TestLogger {
 
 	@Test
-	public void testForceConsumptionOnReleaseEnabled() {
-		testForceConsumptionOnRelease(true);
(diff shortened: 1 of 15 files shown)


flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/ResultPartitionFactoryTest.java
```diff
@@ -39,23 +40,22 @@
 public class ResultPartitionFactoryTest extends TestLogger {
 
 	@Test
-	public void testForceConsumptionOnReleaseEnabled() {
-		testForceConsumptionOnRelease(true);
+	public void testConsumptionOnReleaseEnabled() {
+		testConsumptionOnRelease(true);
```
azagrebin: if `testConsumptionOnRelease` returned `ResultPartition`, you could do the assertion here.
Also `testConsumptionOnRelease ` could directly accept the `ReleaseType` then there would be less ifs.

azagrebin:
@flinkbot approve all

PR #8904, opened 2019-06-26
[FLINK-13005][hive] HiveCatalog should not add 'flink.is_generic' key for Hive table

bowenli86:

This PR fixes a bug that 'flink.is_generic' key should not add for metadata for a non-generic catalog table (a.k.a a Hive table). 

Currently when we create a CatalogTable from a Hive table, if there is no 'flink.is_generic' key in the properties map, we will first add it with a default value as 'false'.

## Brief change log

- made 'flink.' a catalog-wise reserved prefix key for properties
- changed behavior to not add 'flink.is_generic' key when it's not in hive table's properties map
- updated unit tests to cover this case

## Verifying this change

This change added tests and can be verified as follows: `HiveCatalogHiveMetadataTest`

  - The serializers: (no)

## Documentation


bowenli86:

bowenli86:
@xuefuz I've updated the PR with new design as we discussed yesterday. Can you take another look?

bowenli86:

Merging

PR #8909, opened 2019-06-27
[FLINK-13004][table-runtime-blink] Correct the logic of needToCleanupState in KeyedProcessFunctionWithCleanupState

Myasuka:

Current implementation of `#needToCleanupState` in `KeyedProcessFunctionWithCleanupState` used incorrect way to judge whether the time should cleaned up. We should sue `equals` instead of `==` to compare wrapper type of Long. 

## Brief change log

  - Correct the logic of `#needToCleanupState` in `KeyedProcessFunctionWithCleanupState`.
  - Correct the check logic of  whether there are still records which have not been processed yet in `AbstractRowTimeUnboundedPrecedingOver`.

## Verifying this change

  - Updated test of `OverWindowHarnessTest`

  - The serializers: **no**

## Documentation

### DIFF
--- a/flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/runtime/over/AbstractRowTimeUnboundedPrecedingOver.java
+++ b/flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/runtime/over/AbstractRowTimeUnboundedPrecedingOver.java
@@ -154,7 +154,7 @@ public abstract class AbstractRowTimeUnboundedPrecedingOver<K> extends KeyedProc
 			KeyedProcessFunction<K, BaseRow, BaseRow>.OnTimerContext ctx,
 			Collector<BaseRow> out) throws Exception {
 		if (isProcessingTimeTimer(ctx)) {
-			if (needToCleanupState(timestamp)) {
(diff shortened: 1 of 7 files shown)


Myasuka:
@flinkbot attention @wuchong

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/runtime/over/AbstractRowTimeUnboundedPrecedingOver.java
```diff
@@ -157,7 +157,7 @@ public void onTimer(
 			if (needToCleanupState(timestamp)) {
 
 				// we check whether there are still records which have not been processed yet
-				boolean noRecordsToProcess = !inputState.contains(timestamp);
+				boolean noRecordsToProcess = !inputState.keys().iterator().hasNext();
```
wuchong: good catch!

wuchong:
Thanks @Myasuka .

LGTM.

+1 to merge

Myasuka:
@wuchong Thanks for your review, rebase with latest code to fix build error due to symbol not found.

PR #8911, opened 2019-06-27
[FLINK-12995][hive] Add Hive-1.2.1 build to Travis

lirui-apache:

To add Hive-1.2.1 build to Travis.

## Brief change log

  - Added a Travis job to build and test against Hive-1.2.1

## Verifying this change

Existing tests

  - The serializers: no

## Documentation

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,6 +65,7 @@ notifications:
 stages:
   - name: compile
   - name: test
+  - name: connector_hive_1
   - name: E2E
     if: type = cron
   - name: cleanup
@@ -111,6 +112,11 @@ jobs:
       script: ./tools/travis_controller.sh misc
       env: PROFILE="-Dhadoop.version=2.8.3 -Pinclude-kinesis -Dinclude_hadoop_aws -Dscala-2.11"
       name: misc
+    - if: type in (pull_request, push)
+      stage: connector_hive_1
+      script: ./tools/travis_controller.sh connector_hive_1
+      env: PROFILE="-Dhadoop.version=2.8.3 -Pinclude-kinesis -Dinclude_hadoop_aws -Dscala-2.11"
+      name: connector_hive_1
     - if: type in (pull_request, push)
       stage: cleanup
       script: ./tools/travis_controller.sh cleanup
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,7 +65,6 @@ notifications:
 stages:
   - name: compile
   - name: test
-  - name: connector_hive_1
   - name: E2E
     if: type = cron
   - name: cleanup
@@ -113,7 +112,6 @@ jobs:
       env: PROFILE="-Dhadoop.version=2.8.3 -Pinclude-kinesis -Dinclude_hadoop_aws -Dscala-2.11"
       name: misc
     - if: type in (pull_request, push)
-      stage: connector_hive_1
       script: ./tools/travis_controller.sh connector_hive_1
       env: PROFILE="-Dhadoop.version=2.8.3 -Pinclude-kinesis -Dinclude_hadoop_aws -Dscala-2.11"
       name: connector_hive_1
--- a/.travis.yml
+++ b/.travis.yml
@@ -192,9 +192,9 @@ jobs:
       env: PROFILE="-Dhadoop.version=2.8.3 -Pinclude-kinesis -Dinclude_hadoop_aws -Dscala-2.12"
       name: misc - scala 2.12
     - if: type = cron
-      script: ./tools/travis_controller.sh connector_hive_1
+      script: ./tools/travis_controller.sh connector_hive_1_2_1
       env: PROFILE="-Dhadoop.version=2.8.3 -Pinclude-kinesis -Dinclude_hadoop_aws -Dscala-2.12"
-      name: connector_hive_1 - scala 2.12
+      name: connector_hive_1_2_1 - scala 2.12
     - if: type = cron
       stage: cleanup
       script: ./tools/travis_controller.sh cleanup
--- a/.travis.yml
+++ b/.travis.yml
@@ -160,45 +160,41 @@ jobs:
     - if: type = cron
       stage: compile
       script: ./tools/travis_controller.sh compile
-      env: PROFILE="-Dhadoop.version=2.8.3 -Pinclude-kinesis -Dinclude_hadoop_aws -Dscala-2.12"
+      env: PROFILE="-Dhadoop.version=2.8.3 -Pinclude-kinesis -Dinclude_hadoop_aws -Dscala-2.12 -Phive-1.2.1"
       name: compile - scala 2.12
     - if: type = cron
       stage: test
       script: ./tools/travis_controller.sh core
-      env: PROFILE="-Dhadoop.version=2.8.3 -Pinclude-kinesis -Dinclude_hadoop_aws -Dscala-2.12"
+      env: PROFILE="-Dhadoop.version=2.8.3 -Pinclude-kinesis -Dinclude_hadoop_aws -Dscala-2.12 -Phive-1.2.1"
       name: core - scala 2.12
     - if: type = cron
       script: ./tools/travis_controller.sh libraries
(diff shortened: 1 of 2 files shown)


lirui-apache:
Here's [build result](https://travis-ci.org/lirui-apache/flink/builds/551257514) of my own repo. The connector_hive_1 failed as expected.

lirui-apache:
cc @bowenli86 @xuefuz @zjuwangg

tools/travis/stage.sh
```diff
@@ -164,6 +168,9 @@ function get_compile_modules_for_stage() {
         (${STAGE_TESTS})
             echo "-pl $MODULES_TESTS -am"
         ;;
+        (${STAGE_CONNECTOR_HIVE_1})
+            echo "-pl $MODULES_CONNECTOR_HIVE -am -Phive-1.2.1 clean"
```
zentol: My guess is you're using clean so it is recompiled with a different hive version?

zentol:
hold on, do we _really_ need another profile just for hive? How long do the tests need? With the forced recompilation, how much time are we investing here?

zentol:
As an alternative, could we combine this with the hadoop 2.4 or scala 2.12 profile?

tools/travis/stage.sh
lirui-apache: Yes

lirui-apache:
Hi @zentol, the `connector_hive_1` in the build of my own repo took about 13 min and failed, which is expected because we're having some issue with compiling with Hive-1.2.1. So I suppose that's the overhead of the forced recompilation. Once we fix the build, I expect the testing to take another 2~3 min. Of course the testing time will increase as we add more test cases.

lirui-apache:
> As an alternative, could we combine this with the hadoop 2.4 or scala 2.12 profile?

Would you mind explain how to combine it with hadoop 2.4 or scala 2.12? Does it mean the test will run with different hadoop or scala versions? And will the test run for each PR if we do that?

bowenli86:
sorry, should have asked @zentol to help review.

@lirui-apache seems like I might misunderstood how the stage is run? Shouldn't the profile only recompile and retest flink-connector-hive module?

lirui-apache:
> sorry, should have asked @zentol to help review.
> 
> @lirui-apache seems like I might misunderstood how the stage is run? Shouldn't the profile only recompile and retest flink-connector-hive module?

The approach in this PR will recompile flink-connector-hive and all the depended modules. To only recompile flink-connector-hive, we can remove the `-am` option. My understanding is if we remove `-am`, the depended modules will be unavailable (or downloaded somewhere which doesn't reflect changes in a PR) because local maven repo is cleared for each Travis job. @zentol please correct me if I misunderstand.

Another alternative is to completely skip the recompile and only run tests against Hive-1.2.1. So that we can get rid of most of the overhead while we're still able to catch the issues with Hive-1.2.1. Drawbacks of this approach is that errors that happen during tests are probably harder to understand than errors that happen during compiling. For example, we'll get `NoClassDefFoundError` instead of `cannot find symbol`.

zentol:
To integrate it into the hadoop profile, modify the profile of these [builds](https://github.com/apache/flink/blob/master/.travis.yml#L119) to include the hive version.

tools/travis/stage.sh
```diff
@@ -212,6 +219,9 @@ function get_test_modules_for_stage() {
         (${STAGE_TESTS})
             echo "-pl $modules_tests"
         ;;
+        (${STAGE_CONNECTOR_HIVE_1})
+            echo "-pl $MODULES_CONNECTOR_HIVE -Phive-1.2.1"
```
zentol: on a side-note, if you would include clean here you'd only recompile the hive module, but it's quite a hack.

lirui-apache: This sounds like a good hack to me. Do you think we can do it?

lirui-apache:
> To integrate it into the hadoop profile, modify the profile of these [builds](https://github.com/apache/flink/blob/master/.travis.yml#L119) to include the hive version.

Thanks @zentol for the explanation. If combining into hadoop means the test is only triggered by cron jobs, I'd prefer not to do that, at least not while we actively develop hive connector features. Is it acceptable to keep the job in main profile for now?

bowenli86:
@flinkbot attention @zentol

zentol:
I can understand that you'd like to have it per build.

However, with the current Travis situation in mind I would still opt for the cron job. Additionally, this introduces another special case into the Travis builds, and this is something that in the long run we simply cannot maintain.
The CI has already failed due to a compile error. Whether this is an actual problem with the Hive compatibility, or just a bug in the travis logic, I cannot tell. And that's a bit of a problem.

lirui-apache:
Hi @zentol, thanks for your comments. The compile error is expected because previous commits broke build against Hive-1.2.1. And that's exactly why I wanted to add this to Travis.
I'm fine if you think it has to go to cron job. I guess I can add it to the scala 2.12 profile since Hive-1.2.1 doesn't require a specific scala version.

zentol:
CI report for commit 887e6044d73ad3c74a58e175544263aa268f9abf: FAILURE [Travis](https://travis-ci.org/flink-ci/flink-ci/builds/554872310?utm_source=github_status&utm_medium=notification)

lirui-apache:
Updated to move the build to cron job. This build on my own repo verifies it's working properly, https://travis-ci.org/lirui-apache/flink/builds/556152415
@bowenli86 @zentol please take another look, thanks.

tools/travis/stage.sh
```diff
@@ -25,6 +25,7 @@ STAGE_BLINK_PLANNER="blink_planner"
 STAGE_CONNECTORS="connectors"
 STAGE_KAFKA_GELLY="kafka/gelly"
 STAGE_TESTS="tests"
+STAGE_CONNECTOR_HIVE_1="connector_hive_1"
```
bowenli86: can we rename to "..._HIVE/hive_1.2.1" or "..._HIVE/hive_1_2_1"  to make it clean which hive 1.x version we are using?

.travis.yml
```diff
@@ -191,6 +191,10 @@ jobs:
       script: ./tools/travis_controller.sh misc
       env: PROFILE="-Dhadoop.version=2.8.3 -Pinclude-kinesis -Dinclude_hadoop_aws -Dscala-2.12"
       name: misc - scala 2.12
+    - if: type = cron
+      script: ./tools/travis_controller.sh connector_hive_1
+      env: PROFILE="-Dhadoop.version=2.8.3 -Pinclude-kinesis -Dinclude_hadoop_aws -Dscala-2.12"
```
bowenli86: should we use hadoop 2.6.5 which hive 1.2.1 is built on,  rather than 2.8.3?

tools/travis/stage.sh
```diff
@@ -213,6 +220,9 @@ function get_test_modules_for_stage() {
         (${STAGE_TESTS})
             echo "-pl $modules_tests"
         ;;
+        (${STAGE_CONNECTOR_HIVE_1})
+            echo "-pl $MODULES_CONNECTOR_HIVE -Phive-1.2.1 clean"
```
bowenli86: nit: can you add a comment here to note that 'clean' will only recompile flink-connector-hive module?

.travis.yml
lirui-apache: We can't because that'll be a different profile. And profile is used to get the Flink cache between jobs. Since we're using a higher version, hopefully that'll be fine.

lirui-apache:
Update to address comments.
@bowenli86 please take a look, thanks.

bowenli86:
LGTM. Can you retrigger Travis tests? I saw lots of failures in current build.

@flinkbot attention @zentol 

Hopefully Chesnay can take another look. Otherwise, I will merge this PR once its build passes.

lirui-apache:
I don't think the blink planner failures are related. Seems some of them can be reproduced w/o this patch.

.travis.yml
zentol: I would suggest to add `-Phive-1.2.1` to every scala 2.12 profile; then we get the test coverage without requiring another profile that uses funky hacks.

lirui-apache:
Thanks for @zentol 's suggestions. PR updated accordingly.

PR #8912, opened 2019-06-27
[FLINK-12709] [runtime] Implement new generation restart strategy loader which also respects legacy…

zhuzhurk:
… restart strategy configs

*A loader is need to load factories of RestartBackoffTimeStrategy. In order to be backwards compatible, the loader should respect legacy RestartStrategy configurations. The legacy configuration can be in cluster config format (https://ci.apache.org/projects/flink/flink-docs-stable/dev/restart_strategies.html) or in RestartStrategies.RestartStrategyConfiguration format.*

## Brief change log

  - *The loader coverts legacy configuration into new generation format*
  - *The loader creates RestartBackoffTimeStrategy factory from new format configs*

## Verifying this change

  - *Added unit tests RestartBackoffTimeStrategyFactoryLoader*

  - The serializers: (no)

## Documentation


flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/RestartBackoffTimeStrategyFactoryLoader.java
```diff
@@ -0,0 +1,398 @@
...
+			ConfigConstants.RESTART_STRATEGY_FIXED_DELAY_DELAY);
+
+		final long delay;
+		try {
+			delay = Duration.apply(delayString).toMillis();
+		} catch (NumberFormatException nfe) {
+			throw new Exception("Invalid config value for " +
+				ConfigConstants.RESTART_STRATEGY_FIXED_DELAY_DELAY + ": " + delayString +
+				". Value must be a valid duration (such as '100 milli' or '10 s')");
+		}
```
eaglewatcherwb: code in line 284~ 291 is a duplicate to that in line 312~ 327, which could be extracted into a static function.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/failover/flip1/RestartBackoffTimeStrategyFactoryLoaderTest.java
```diff
@@ -0,0 +1,210 @@
...
+	}
+
+	// ------------------------------------------------------------------------
+	//  utilities
+	// ------------------------------------------------------------------------
+
+	/**
+	 * A test implementation of {@link RestartBackoffTimeStrategy}.
+	 */
+	private static class TestRestartBackoffTimeStrategy implements RestartBackoffTimeStrategy {
```
eaglewatcherwb: Seems a duplicated class of `TestRestartBackoffTimeStrategy` with that in TestRestartBackoffTimeStrategy.java

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/RestartBackoffTimeStrategyFactoryLoader.java
```diff
@@ -0,0 +1,398 @@
...
+			LOG.warn("Illegal access while calling method {} from class {}.", CREATE_METHOD, restartStrategyClassName);
+		}
+
+		// fallback in case of an error
+		LOG.info("Configured restart strategy {} is not valid. Fall back to no restart strategy.",
+			restartStrategyClassName);
+		return new NoRestartBackoffTimeStrategy.NoRestartBackoffTimeStrategyFactory();
+	}
+
+	private static boolean tryAdaptingJobRestartStrategyConfiguration(
```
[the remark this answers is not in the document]
zhuzhurk: true. fixed.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/failover/flip1/RestartBackoffTimeStrategyFactoryLoaderTest.java
zhuzhurk: You are right. Will modify the standalone TestRestartBackoffTimeStrategy class and use it for testing.

docs/_includes/generated/blob_server_configuration.html
```diff
@@ -7,16 +7,16 @@
...
-            <td><h5>blob.client.socket.timeout</h5></td>
-            <td style="word-wrap: break-word;">300000</td>
-            <td>The socket timeout in milliseconds for the blob client.</td>
-        </tr>
         <tr>
             <td><h5>blob.client.connect.timeout</h5></td>
             <td style="word-wrap: break-word;">0</td>
             <td>The connection timeout in milliseconds for the blob client.</td>
         </tr>
+        <tr>
```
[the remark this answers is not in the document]
zhuzhurk: The changes for config html is automatically generated. I guess it's due to some previous changes did not generate docs properly.
I have notified zhijiang about it and he has already fixed it. After rebasing master, this diff would be gone.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/RestartBackoffTimeStrategyFactoryLoader.java
zhuzhurk: Yes. we can extract the duration parsing logic.

zhuzhurk:
Forced pushing after rebased onto master.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/RestartBackoffTimeStrategyFactoryLoader.java
```diff
@@ -0,0 +1,393 @@
...
+		final int maxFailuresPerInterval = clusterConfiguration.getInteger(
+			ConfigConstants.RESTART_STRATEGY_FAILURE_RATE_MAX_FAILURES_PER_INTERVAL,
+			1);
+
+		final long failuresInterval = parseDurationFromConfig(
+			clusterConfiguration,
+			ConfigConstants.RESTART_STRATEGY_FAILURE_RATE_FAILURE_RATE_INTERVAL,
+			Duration.apply(1, TimeUnit.MINUTES).toString()).toMillis();
+
+		final Duration heartbeatPause = parseDurationFromConfig(
```
zhuzhurk: In legacy FailureRateRestartStrategy it uses `AkkaOptions.WATCH_HEARTBEAT_INTERVAL` as the default delay. But I think `AkkaOptions.WATCH_HEARTBEAT_PAUSE` should be the proper one.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/RestartBackoffTimeStrategyFactoryLoader.java
```diff
@@ -0,0 +1,393 @@
...
+	private static boolean tryAdaptingClusterConfigurationOfDeprecatedValues(
+		final Configuration clusterConfiguration,
+		final Configuration adaptedConfiguration) throws Exception {
+
+		// support deprecated ConfigConstants values
+		final int maxAttempts = clusterConfiguration.getInteger(
+			ConfigConstants.RESTART_STRATEGY_FIXED_DELAY_ATTEMPTS,
+			ConfigConstants.DEFAULT_EXECUTION_RETRIES);
+
+		final Duration heartbeatPause = parseDurationFromConfig(
```
zhuzhurk: The behavior is a bit different here.
Previously if user has configured a valid `RESTART_STRATEGY_FIXED_DELAY_DELAY` value and an invalid `AkkaOptions.WATCH_HEARTBEAT_PAUSE` value. It would be OK.
In this diff, if the `AkkaOptions.WATCH_HEARTBEAT_PAUSE` is invalid it will always fail this process.
I think it should accepted, since a valid `AkkaOptions.WATCH_HEARTBEAT_PAUSE` is required in some other critical path.

flink-core/src/main/java/org/apache/flink/configuration/RestartBackoffTimeStrategyOptions.java
```diff
@@ -25,6 +25,16 @@
  */
 @PublicEvolving
 public class RestartBackoffTimeStrategyOptions {
+
+	/**
+	 * Class name of the RestartBackoffTimeStrategy implementation to use.
+	 */
+	@PublicEvolving
+	public static final ConfigOption<String> RESTART_BACKOFF_TIME_STRATEGY_CLASS_NAME = ConfigOptions
+		.key("restart-backoff-time-strategy.class-name")
```
GJL: With the introduction of the new scheduler, users will not be able to use custom implementations of the `RestartStrategy` interface. Strictly speaking this is a breaking change. However, it is not clear to me if there are any Flink users at all that require a custom `RestartStrategy`. I think it makes sense to start a discussion on the dev mailing list. Until it is clear that users require this level of customization, we should not offer users to write their own back off time strategy for now.

flink-core/src/main/java/org/apache/flink/configuration/RestartBackoffTimeStrategyOptions.java
```diff
@@ -46,7 +57,7 @@
 		.withDescription("Time interval in milliseconds for measuring failure rate.");
 
 	/**
-	 * Backoff time between two consecutive restart attempts in FailureRateRestartBackoffTimeStrategy.
+	 * Backoff time (milli-seconds) between two consecutive restart attempts in FailureRateRestartBackoffTimeStrategy.
 	 */
 	@PublicEvolving
 	public static final ConfigOption<Long> RESTART_BACKOFF_TIME_STRATEGY_FAILURE_RATE_FAILURE_RATE_BACKOFF_TIME = ConfigOptions
```
GJL: I know this discussion is out of scope for this PR but do we need new config keys? For backwards compatibility we should respect the existing [config keys](https://github.com/apache/flink/blob/0a405251b297109fde1f9a155eff14be4d943887/flink-core/src/main/java/org/apache/flink/configuration/RestartStrategyOptions.java) anyways.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/RestartBackoffTimeStrategyFactoryLoader.java
```diff
@@ -0,0 +1,393 @@
...
+		final String restartStrategyClassName = configuration.getString(
+			RestartBackoffTimeStrategyOptions.RESTART_BACKOFF_TIME_STRATEGY_CLASS_NAME);
+
+		if (restartStrategyClassName == null) {
+			LOG.info("No restart strategy is configured. Using no restart strategy.");
+			return new NoRestartBackoffTimeStrategy.NoRestartBackoffTimeStrategyFactory();
+		}
+
+		try {
+			final Class<?> clazz = Class.forName(restartStrategyClassName);
```
GJL: If we do not offer users to provide their own strategy (see one of my other comments), we can rethink whether it is _"correct"_ to always instantiate the strategies by reflection.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/RestartBackoffTimeStrategyFactoryLoader.java
```diff
@@ -0,0 +1,393 @@
...
+	}
+
+	private static RestartBackoffTimeStrategy.Factory createRestartStrategyFactoryInternal(
+			final Configuration configuration) {
+
+		final String restartStrategyClassName = configuration.getString(
+			RestartBackoffTimeStrategyOptions.RESTART_BACKOFF_TIME_STRATEGY_CLASS_NAME);
+
+		if (restartStrategyClassName == null) {
+			LOG.info("No restart strategy is configured. Using no restart strategy.");
```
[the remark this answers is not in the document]
zhuzhurk: Ok.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/RestartBackoffTimeStrategyFactoryLoader.java
```diff
@@ -0,0 +1,393 @@
...
+
+					if (result != null) {
+						return (RestartBackoffTimeStrategy.Factory) result;
+					}
+				}
+			}
+		} catch (ClassNotFoundException cnfe) {
+			LOG.warn("Could not find restart strategy class {}.", restartStrategyClassName);
+		} catch (NoSuchMethodException nsme) {
+			LOG.warn("Class {} does not has static method {}.", restartStrategyClassName, CREATE_METHOD);
```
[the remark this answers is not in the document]
zhuzhurk: Ok/

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/RestartBackoffTimeStrategyFactoryLoader.java
```diff
@@ -0,0 +1,393 @@
...
+		} catch (InvocationTargetException ite) {
+			LOG.warn("Cannot call static method {} from class {}.", CREATE_METHOD, restartStrategyClassName);
+		} catch (IllegalAccessException iae) {
+			LOG.warn("Illegal access while calling method {} from class {}.", CREATE_METHOD, restartStrategyClassName);
+		}
+
+		// fallback in case of an error
+		LOG.info("Configured restart strategy {} is not valid. Fall back to no restart strategy.",
```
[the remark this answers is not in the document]
zhuzhurk: Ok.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/failover/flip1/RestartBackoffTimeStrategyFactoryLoaderTest.java
```diff
@@ -0,0 +1,174 @@
...
+	}
+
+	// ------------------------------------------------------------------------
+	//  utilities
+	// ------------------------------------------------------------------------
+
+	/**
+	 * A test class that has not implemented {@link RestartBackoffTimeStrategy}.
+	 */
+	private static class InvalidTestRestartBackoffTimeStrategy {}
```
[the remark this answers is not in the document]
zhuzhurk: Ok.

flink-core/src/main/java/org/apache/flink/configuration/RestartBackoffTimeStrategyOptions.java
zhuzhurk: Good question. Previously the configs for strategy strategies are quite a mess in ConfigConstants and I wanted to deprecate them. But with FLIP 61 the configs are better organized in RestartStrategyOptions.java.

I think now we can use the configs from RestartStrategyOptions.java directly. And I'm also thinking to rename RestartBackoffTimeStrategy to RestartStrategy once we have removed the legacy RestartStrategy.

I will open a hotfix commit to migrate these options and their usages in related strategy.

flink-core/src/main/java/org/apache/flink/configuration/RestartBackoffTimeStrategyOptions.java
zhuzhurk: Yes it is a breaking change if a user was using a custom RestartStrategy.
I will open a discussion to
1. notify users that previous custom RestartStrategy does not work in the new scheduler anymore
2. discuss whether we should keep supporting custom RestartBackoffTimeStrategy for existing custom RestartStrategy migration or future use

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/RestartBackoffTimeStrategyFactoryLoader.java
zhuzhurk: Yes. I will remove the customization logic from this PR. 
Will add it in another PR if the discussion result is that customizable restart strategies are still required.

GJL: As long as there are unit tests, it is probably reasonably safe to keep instantiating by reflection.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/RestartBackoffTimeStrategyFactoryLoader.java
```diff
@@ -0,0 +1,393 @@
...
+
+import scala.concurrent.duration.Duration;
+
+import static org.apache.flink.util.Preconditions.checkArgument;
+import static org.apache.flink.util.Preconditions.checkNotNull;
+
+/**
+ * A utility class to load {@link RestartBackoffTimeStrategy.Factory} from the configuration.
+ */
+public class RestartBackoffTimeStrategyFactoryLoader {
```
GJL: Since there are static members only, I'd make this class final with a private constructor.

flink-core/src/main/java/org/apache/flink/configuration/RestartBackoffTimeStrategyOptions.java
zhuzhurk: The survey is concluded that we do not support restart strategy customization. 
I will also remove the description of restart strategy customization in a hotfix commit, so that we do not make it a public interface.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/RestartBackoffTimeStrategyFactoryLoader.java
GJL: Actually I changed my mind on this matter. If we remove support for custom restart strategies/back-off time strategies, there is no apparent reason to instantiate the strategy reflectively. I would propose to do the simplest thing possible to avoid over-engineering the code. Also the fact that we require reflective invocation of the `createFactory` method means to me that something is wrong. Should we need custom strategies, we can consider one of the following alternatives for loading them:

- http://mail-archives.apache.org/mod_mbox/flink-dev/201904.mbox/%3C02763EDD-4806-4649-91ED-D7E8205C5D64@ververica.com%3E
- https://docs.oracle.com/javase/9/docs/api/java/util/ServiceLoader.html

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/RestartBackoffTimeStrategyFactoryLoader.java
```diff
@@ -0,0 +1,312 @@
...
+	private static Configuration getAdaptedConfiguration(
+			final RestartStrategies.RestartStrategyConfiguration jobRestartStrategyConfiguration,
+			final Configuration clusterConfiguration,
+			final boolean isCheckpointingEnabled) {
+
+		// clone a new config to not modify the existing config
+		final Configuration adaptedConfiguration = new Configuration();
+
+		// try determining restart strategy based on job level restart strategy config first
+		if (!tryAdaptingJobRestartStrategyConfiguration(jobRestartStrategyConfiguration, adaptedConfiguration)) {
```
GJL: I think we should drop the reflective instantiation (see my other comment) and return the `RestartBackoffTimeStrategy.Factory` immediately.  Moreover, I would restructure the code a bit to avoid output arguments (`adaptedConfiguration`), which are usually discouraged:

```
		final Optional<RestartBackoffTimeStrategy.Factory> jobRestartStrategy = [...];
		final Optional<RestartBackoffTimeStrategy.Factory> clusterRestartStrategy = [...];

		return Stream.of(jobRestartStrategy, clusterRestartStrategy)
			.filter(Optional::isPresent)
			.findFirst()
			.map(Optional::get)
			.orElse(defaultRestartStrategy(checkpointingEnabled);
```

Let me know what you think.

zhuzhurk: Agreed. We need to reorganize the code if we do not base it on the reflective way.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/FixedDelayRestartBackoffTimeStrategy.java
```diff
@@ -57,6 +58,11 @@ public long getBackoffTime() {
 		return backoffTimeMS;
 	}
 
+	@VisibleForTesting
+	int getMaxNumberRestartAttempts() {
```
GJL: I think it's reasonable to make this part of the public API, i.e., remove `@VisibleForTesting` and add `public`. Generally speaking, if a value is exposed by `toString()`, the value should be accessible by a getter or other methods.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/RestartBackoffTimeStrategyFactoryLoader.java
```diff
@@ -0,0 +1,135 @@
...
+
+		return getJobRestartStrategyFactory(jobRestartStrategyConfiguration)
+			.orElse(getClusterRestartStrategyFactory(clusterConfiguration)
+				.orElse(getDefaultRestartStrategyFactory(isCheckpointingEnabled)));
+	}
+
+	private static Optional<RestartBackoffTimeStrategy.Factory> getJobRestartStrategyFactory(
+			final RestartStrategies.RestartStrategyConfiguration restartStrategyConfiguration) {
+
+		if (restartStrategyConfiguration instanceof RestartStrategies.NoRestartStrategyConfiguration) {
```
GJL: I would add the following imports:

```
import org.apache.flink.api.common.restartstrategy.RestartStrategies.FailureRateRestartStrategyConfiguration;
import org.apache.flink.api.common.restartstrategy.RestartStrategies.FallbackRestartStrategyConfiguration;
import org.apache.flink.api.common.restartstrategy.RestartStrategies.FixedDelayRestartStrategyConfiguration;
import org.apache.flink.api.common.restartstrategy.RestartStrategies.NoRestartStrategyConfiguration;
```

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/RestartBackoffTimeStrategyFactoryLoader.java
```diff
@@ -0,0 +1,135 @@
...
+			return Optional.of(NoRestartBackoffTimeStrategy.NoRestartBackoffTimeStrategyFactory.INSTANCE);
+		} else if (restartStrategyConfiguration instanceof RestartStrategies.FixedDelayRestartStrategyConfiguration) {
+			final RestartStrategies.FixedDelayRestartStrategyConfiguration fixedDelayConfig =
```
GJL: I think this branch and the one before is not tested.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/RestartBackoffTimeStrategyFactoryLoader.java
```diff
@@ -0,0 +1,135 @@
...
+		final String restartStrategyName = clusterConfiguration.getString(RestartStrategyOptions.RESTART_STRATEGY);
+		if (restartStrategyName == null) {
+			return Optional.empty();
+		}
+
+		switch (restartStrategyName.toLowerCase()) {
+			case "none":
+			case "off":
+			case "disable":
+				return Optional.of(NoRestartBackoffTimeStrategy.NoRestartBackoffTimeStrategyFactory.INSTANCE);
```
GJL: I think this branch is not tested

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/RestartBackoffTimeStrategyFactoryLoader.java
```diff
@@ -0,0 +1,135 @@
...
+				(RestartStrategies.FailureRateRestartStrategyConfiguration) restartStrategyConfiguration;
+
+			return Optional.of(new FailureRateRestartBackoffTimeStrategy.FailureRateRestartBackoffTimeStrategyFactory(
+				failureRateConfig.getMaxFailureRate(),
+				failureRateConfig.getFailureInterval().toMilliseconds(),
+				failureRateConfig.getDelayBetweenAttemptsInterval().toMilliseconds()));
+		} else if (restartStrategyConfiguration instanceof RestartStrategies.FallbackRestartStrategyConfiguration) {
+			return Optional.empty();
+		} else {
+			throw new IllegalArgumentException("Unknown restart strategy configuration " +
```
GJL: Branch is not tested but it's acceptable.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/failover/flip1/RestartBackoffTimeStrategyFactoryLoaderTest.java
```diff
@@ -0,0 +1,124 @@
...
+import static junit.framework.TestCase.assertEquals;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.core.IsInstanceOf.instanceOf;
+
+/**
+ * Unit tests for {@link RestartBackoffTimeStrategyFactoryLoader}.
+ */
+public class RestartBackoffTimeStrategyFactoryLoaderTest extends TestLogger {
+
+	private static final RestartStrategies.RestartStrategyConfiguration DEFAULT_RESTART_STRATEGY_CONFIGURATION =
```
GJL: Maybe `DEFAULT_JOB_LEVEL_RESTART_CONFIGURATION`. If you don't want to change it, it's also fine.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/RestartBackoffTimeStrategyFactoryLoader.java
zhuzhurk: Yes. I'd like to avoid adding a testing RestartStrategyConfiguration just for testing this branch.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/FixedDelayRestartBackoffTimeStrategy.java
zhuzhurk: Ok.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/RestartBackoffTimeStrategyFactoryLoader.java
zhuzhurk: Good suggestion!

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/failover/flip1/RestartBackoffTimeStrategyFactoryLoaderTest.java
zhuzhurk: I will change it to make the tests easier to understand.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/RestartBackoffTimeStrategyFactoryLoader.java
zhuzhurk: Done.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/flip1/RestartBackoffTimeStrategyFactoryLoader.java
zhuzhurk: Done.

PR #8914, opened 2019-06-27
[FLINK-12929] Pass TypeInformation in addSource

lombarde:
Co-authored-by: Georg Rollinger <georg.rollinger@posteo.net>

Fixes https://issues.apache.org/jira/browse/FLINK-12929:
Changed scala.StreamExecutionEnvironment.addSource to pass the type info for a SourceFunction that does not extend ResultTypeQueryable.

## Brief change log

see above

## Verifying this change

No test added as scala/StreamExecutionEnvironment.scala seems to be not covered yet

  - The serializers: no

## Documentation

### DIFF
--- a/flink-streaming-scala/src/main/scala/org/apache/flink/streaming/api/scala/StreamExecutionEnvironment.scala
+++ b/flink-streaming-scala/src/main/scala/org/apache/flink/streaming/api/scala/StreamExecutionEnvironment.scala
@@ -615,7 +615,7 @@ class StreamExecutionEnvironment(javaEnv: JavaEnv) {
     
     val cleanFun = scalaClean(function)
     val typeInfo = implicitly[TypeInformation[T]]
-    asScalaStream(javaEnv.addSource(cleanFun).returns(typeInfo))
+    asScalaStream(javaEnv.addSource(cleanFun, typeInfo))
   }
 
   /**
--- a/flink-streaming-java/src/main/java/org/apache/flink/streaming/api/environment/StreamExecutionEnvironment.java
+++ b/flink-streaming-java/src/main/java/org/apache/flink/streaming/api/environment/StreamExecutionEnvironment.java
@@ -1453,17 +1453,16 @@ public abstract class StreamExecutionEnvironment {
 	@SuppressWarnings("unchecked")
 	public <OUT> DataStreamSource<OUT> addSource(SourceFunction<OUT> function, String sourceName, TypeInformation<OUT> typeInfo) {
 
+		if (function instanceof ResultTypeQueryable) {
+			typeInfo = ((ResultTypeQueryable<OUT>) function).getProducedType();
+		}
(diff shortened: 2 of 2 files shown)


StephanEwen:
@aljoscha What is the overall contract in the Scala API for which type information is used:

Consider this case:
```scala
env.addSource[StringTypeInfo.INSTANCE](sourceThatImplementsResultTypeQueryable);
```
Which info will be used / should be used in the end? Is this well defined?

aljoscha:
I think it was somewhat well defined until this change: https://github.com/apache/flink/commit/e39699f3eb41517a7035a5e626c6bb20c6b70501#diff-8c6a04f65f5a61e021e96322e8b06432R1000

Before, the idea was that if the source implements `ResultTypeQueryable`, then that should take precedence. Otherwise we try to deduce a `TypeInformation`. I think we should go back to that, i.e. `ResultTypeQueryable` takes precedence (in `addSource()`).

The fact that the Scala API uses `returns()` comes from the fact that there was no `StreamExecutionEnvironment.addSource(SourceFunction, TypeInformation)`. Nowadays, I think no internal methods should use `returns()`.

For this PR, it means that we have to not call `returns()`, but always use the constructor. And then in the constructor we have to do the correct change to make `ResultTypeQueryable` take precedence.

StephanEwen:
@lombarde Are you interested in picking this up, following @aljoscha's suggestion?

lombarde:
> @lombarde Are you interested in picking this up, following @aljoscha's suggestion?

We updated our code following your suggestions, but we are not completely sure we understood what you meant. Please double check.

aljoscha:
I merged this, thanks for your work @lombarde. 😃

flink-streaming-scala/src/main/scala/org/apache/flink/streaming/api/scala/StreamExecutionEnvironment.scala
```diff
@@ -615,7 +615,7 @@ class StreamExecutionEnvironment(javaEnv: JavaEnv) {
     
     val cleanFun = scalaClean(function)
     val typeInfo = implicitly[TypeInformation[T]]
-    asScalaStream(javaEnv.addSource(cleanFun).returns(typeInfo))
+    asScalaStream(javaEnv.addSource(cleanFun, typeInfo))
```
tdbgamer: Is there a reason why the `returns` call was removed here? This has broken our microservices we have upgraded to 1.9.1 since the output type of the transformer doesn't get set and later fails when the resulting DataStream is used.

aljoscha: This change should actually still set the output type because we now pass the type directly in the function call which also sets it on the transformer? Could you maybe step through the code using a debugger and see if (or why not) the output type is set or not set.

PR #3751, opened 2017-04-21
[FLINK-6330] [docs] Add basic Docker, K8s docs

patricklucas:
I may do some more work on these before 1.2.2/1.3, but I'd like to at least get this basic information into master.

PR #8917, opened 2019-06-27
[FLINK-13017][docs] do not mount local $HOME into docker

NicoK:

Remove the (writable!) mount of a user's $HOME into the dockerized documentation build container in order to
- make the builds independent from the host system (making them reproducible)
- not have the commands in the container affect the host

## Brief change log

- remove mounting user $HOME

## Verifying this change

I verified building the docs inside the new environment.

  - The serializers: **no**

## Documentation

### DIFF
--- a/docs/docker/run.sh
+++ b/docs/docker/run.sh
@@ -31,12 +31,10 @@ if [ "$(uname -s)" == "Linux" ]; then
   USER_NAME=${SUDO_USER:=$USER}
   USER_ID=$(id -u "${USER_NAME}")
   GROUP_ID=$(id -g "${USER_NAME}")
-  LOCAL_HOME="/home/${USER_NAME}"
 else # boot2docker uid and gid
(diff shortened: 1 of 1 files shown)


sjwiesman:
I noticed that the run script did not originally mount`$HOME` and then that was added by @patricklucas in #3751. Patrick can you just verify that removing this won't break something.

NicoK:
originally, it pointed to `/home/$USER` which (on Linux) is basically the same - Patrick only made it work on MacOS as well

sjwiesman:
Gotcha, then +1

NicoK:
@sjwiesman Did you try building the docs with these changes? Just to be on the safe side and have someone else verify with a different OS at least (I'm on Linux)

sjwiesman:
Yes. I successfully built with these changes on a Mac

StephanEwen:
Looks good to me, merging this...

PR #8920, opened 2019-06-27
[FLINK-13024][table] integrate FunctionCatalog with CatalogManager

bowenli86:
What is the purpose of the change

This PR is a first step to integrate FunctionCatalog with Catalog APIs. 

Please note that this PR is not ready to be merged because I found some gaps that prevent it from functioning. Please see my comments below.

## Brief change log

- migrated `FunctionCatalog`'s constructors and API implementations to be based on `CatalogManager`

## Verifying this change

This change is already covered by existing tests:
TODO: ideally, existing tests should pass with minimal changes, and some new tests should be added

  - The serializers: (no)

## Documentation

  - Does this pull request introduce a new feature? (yes )


flink-table/flink-table-planner/src/main/scala/org/apache/flink/table/plan/rules/logical/PushFilterIntoTableSourceScanRule.scala
```diff
@@ -68,7 +68,7 @@ class PushFilterIntoTableSourceScanRule extends RelOptRule(
       RexProgramExtractor.extractConjunctiveConditions(
         program,
         call.builder().getRexBuilder,
-        new FunctionCatalog("default_catalog", "default_database"))
+        new FunctionCatalog(null))
```
bowenli86: are we able to pass in CatalogManager here?

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/FunctionCatalog.java
```diff
@@ -116,15 +119,48 @@ public void registerScalarFunction(String name, ScalarFunction function) {
...
+		String functionName = normalizeName(name);
+
+		Catalog catalog = catalogManager.getCatalog(catalogManager.getCurrentCatalog()).get();
+
+		FunctionDefinition userCandidate = null;
+		try {
+			CatalogFunction catalogFunction = catalog.getFunction(
+				new ObjectPath(catalogManager.getCurrentDatabase(), functionName));
+
+			// TODO: use FunctionDefintionFactory to initiate a FunctionDefinition from CatalogFunction
```
bowenli86: we need FunctionDefintionFactory here

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/FunctionCatalog.java
```diff
@@ -116,15 +119,48 @@ public void registerScalarFunction(String name, ScalarFunction function) {
...
+			// Ignore since there will always be a current database of the current catalog
+		}
+
+		// Get built-in functions
+		result.addAll(
+			BuiltInFunctionDefinitions.getDefinitions()
+				.stream()
+				.map(f -> f.getName())
+				.collect(Collectors.toList()));
+
```
zjuwangg: minor:
Since this method also returns the built-in functions, we may change the method name to another like `getDefinedFunctions` or other?

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/FunctionCatalog.java
```diff
@@ -136,19 +172,66 @@ public void registerScalarFunction(String name, ScalarFunction function) {
...
-		userFunctions.put(normalizeName(name), functionDefinition);
+
+		if (functionDefinition instanceof BuiltInFunctionDefinition) {
+			throw new TableException("Cannot register a built-in Flink function");
+		} else {
+			Catalog catalog = catalogManager.getCatalog(catalogManager.getCurrentCatalog()).get();
+
+			String functionIdentifier;
+
+			if (functionDefinition.getKind() == FunctionKind.SCALAR) {
```
zjuwangg: minor:
Can we use the switch/case syntax to make it more readable?

bowenli86:
@flinkbot attention @twalthr @dawidwys

flink-table/flink-table-planner/src/main/scala/org/apache/flink/table/plan/rules/logical/PushFilterIntoTableSourceScanRule.scala
bowenli86: found that this is taken care of as part of https://github.com/apache/flink/pull/8782

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/FunctionCatalog.java
bowenli86: I agree. That can be on its own PR

bowenli86:
Hi @twalthr , thanks for your feedback!

I believe cutting the scope to only support Hive UDFs makes sense given the limited time left for 1.9 feature freeze. I'm not sure about merging FunctionCatalog into CatalogManager, CatalogManager has many APIs and responsibilities that have nothing to do with functions, merging FunctionCatalog and CatalogManager seems will expose many parts of the planner with a lot of information that they don't need to know.

The latest push incorporates changes that adds `FunctionDefinitionFactory`, `HiveFunctionDefinitionFactory`, and support of only reading Hive UDFs in `FunctionCatalog`. The most critical part that's missing now is `FunctionLooup.lookupFunction` API should pass in an array of Expressions to get literals and input arg types, in order	to instantiate Hive functions, compute the result/accumulator type, and create corresponding FunctionDefinition. Are you able to fill this gap anytime soon, hopefully in 1.9?

cc @xuefuz @KurtYoung @JingsongLi @dawidwys

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/FunctionCatalog.java
```diff
@@ -116,14 +125,60 @@ public void registerScalarFunction(String name, ScalarFunction function) {
...
+
+		// First, check Hive UDFs since HiveCatalog provides its own FunctionDefinitionFactory
+		// TODO: may consider built-in FunctionDefinitionFactory
+		if (catalog.getFunctionDefinitionFactory().isPresent()) {
+			try {
+				CatalogFunction catalogFunction = catalog.getFunction(
+					new ObjectPath(catalogManager.getCurrentDatabase(), functionName));
+
+				userCandidate = catalog.getFunctionDefinitionFactory().get()
+					.createFunctionDefinition(name, catalogFunction, null, null);
```
xuefuz: name => functionName?

flink-table/flink-table-common/src/main/java/org/apache/flink/table/factories/FunctionDefinitionFactory.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+package org.apache.flink.table.factories;
+
+import org.apache.flink.table.catalog.CatalogFunction;
+import org.apache.flink.table.functions.FunctionDefinition;
+import org.apache.flink.table.types.DataType;
+
+/**
+ * A factory to create {@link FunctionDefinition} based on string-based properties.
```
xuefuz: "based on string-based properties" might be inaccurate as it's not apparent from the only interface. (Though it's possible we might introduce additional interface for that.)

flink-table/flink-table-common/src/main/java/org/apache/flink/table/factories/FunctionDefinitionFactory.java
```diff
@@ -0,0 +1,43 @@
...
+	 *
+	 * @param name name of the {@link CatalogFunction}
+	 * @param catalogFunction the catalog function
+	 * @param constantArguments arguments of a function call (only literal arguments
+	 *                  are passed, nulls for non-literal ones)
+	 * @param argTypes types of arguments
+	 * @return a {@link FunctionDefinition}
+	 */
+	FunctionDefinition createFunctionDefinition(
+		String name, CatalogFunction catalogFunction, Object[] constantArguments, DataType[] argTypes);
```
xuefuz: Should we change the first parameter to ObjectPath functionPath? It's possible to have functions with the same name but in different databases in a catalog.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/FunctionCatalog.java
```diff
@@ -116,14 +125,60 @@ public void registerScalarFunction(String name, ScalarFunction function) {
...
+				.collect(Collectors.toList()));
+
+		return result.stream()
+			.map(n -> normalizeName(n))
+			.collect(Collectors.toList())
+			.toArray(new String[0]);
 	}
 
 	@Override
 	public Optional<FunctionLookup.Result> lookupFunction(String name) {
```
xuefuz: I'm curious about the format of name here. In complete form, a lookup is for a function like catalog.database.functionName, though it's possible to have database.functionName or just functionName. When it's just functionName, it can be either refer to one in current catalog/db or a built-in function.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/FunctionCatalog.java
```diff
@@ -116,14 +125,60 @@ public void registerScalarFunction(String name, ScalarFunction function) {
 	}
 
 	public String[] getUserDefinedFunctions() {
```
xuefuz: We probably also need to take care of namespace (catalog.database) of functions.

bowenli86:
@xuefuz tahsnk for your review!

A few of your comments on function path are valid. I thought about that previously too, and I think that should reside in a different PR since it requires refactoring upstream code in udf/expression framework. As this PR is already such a huge change, it's better for us to keep its scope small.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/factories/FunctionDefinitionFactory.java
bowenli86: removed the "string-properties" part

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/FunctionCatalog.java
```diff
@@ -116,14 +120,60 @@ public void registerScalarFunction(String name, ScalarFunction function) {
...
 	@Override
 	public Optional<FunctionLookup.Result> lookupFunction(String name) {
-		final FunctionDefinition userCandidate = userFunctions.get(normalizeName(name));
+		String functionName = normalizeName(name);
+
+		FunctionDefinition userCandidate = null;
+
+		Catalog catalog = catalogManager.getCatalog(catalogManager.getCurrentCatalog()).get();
+
+		// First, check Hive UDFs since HiveCatalog provides its own FunctionDefinitionFactory
```
twalthr: Don't mention Hive here. This is general API without any Hive specifics.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/FunctionCatalog.java
twalthr: I have the same question here. I guess the signature needs to be changed to `String[] path`. And we cannot simply assume the current catalog but must consider all catalogs available.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/FunctionLookup.java
```diff
@@ -33,6 +33,10 @@
 
 	/**
 	 * Lookup a function by name. The lookup is case insensitive.
+	 *
+	 * TODO: 1) replace string with ObjectIdentifier so users can reference functions across catalogs
+	 * TODO: 2) should pass in an array of Expressions to get literals and input arg types, in order
+	 * 	to instantiate Hive functions, compute the result/accumulator type, and create corresponding FunctionDefinition
```
twalthr: The list of literals/input arg types should not be part of the lookup as they are not known at this point.

btw: can we remove all these TODOs from the PR. TODO's make sense if one functionality is missing in a class. But the function lookup is under heavy refactoring and we could add a lot of TODOs everywhere that would make the code unreadable.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/factories/FunctionDefinitionFactory.java
twalthr: Why does the factory needs information about the database? Actually, I also wonder why the `TableSourceFactory` has access to `ObjectPath`. It mixes two separate layers.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/factories/FunctionDefinitionFactory.java
```diff
@@ -0,0 +1,43 @@
...
+	 * Creates a {@link FunctionDefinition} from given {@link CatalogFunction}.
+	 *
+	 * @param name name of the {@link CatalogFunction}
+	 * @param catalogFunction the catalog function
+	 * @param constantArguments arguments of a function call (only literal arguments
+	 *                  are passed, nulls for non-literal ones)
+	 * @param argTypes types of arguments
+	 * @return a {@link FunctionDefinition}
+	 */
+	FunctionDefinition createFunctionDefinition(
```
twalthr: `Object[] constantArguments, DataType[] argTypes` should not be part of this interface. The returned function definitions should handle those by returning a `TypeInference`.

JingsongLi: +1 to timo's comment.
`FunctionDefinition` should be a definition instead of selected method.

JingsongLi:
> Thanks @bowenli86. I added some feedback.
> 
> I think my biggest concern is how we pass input types and literals to functions. This should actually happen at a different level in the stack. For simplification, a function catalog should just lookup functions by name. The returned function definition of the factory should take care of receiving literals and input types. Since function definition does not return `TypeInference` yet, we could also use `getParameterType` and `getReturnType` of UDFs. They don't take literals but I think it is acceptable to postpone supporting this to the next release to still get this feature in before the feature freeze.

My suggestion:

Solution I: We only support function with single one method in 1.9. We rewrite `function.getResolver().getEvalMethod(typeInfos)` in hive UDF to find the single method in hive function by ourselves. Then, we can use this return type to override `getReturnType`. The biggest problem with this solution is that support for Hive functions is limited. 

Solution II:
We add `HiveScalarSqlFunction` in blink runner, it extends `ScalarSqlFunction`, override getFunction to set the input types and literals to cloned function. (Use java reflection to invoke `setArgumentTypesAndConstants`) And override `ReturnTypeInference`, to return the correct type with java reflection to `getHiveResultType`.
In blink `FunctionCatalogOperatorTable. convertToSqlFunction`, we judge the function, if it is a hive function(Use java reflection), we decide to generate a `HiveScalarSqlFunction`.
In this way, we can let get the correct return type from hive function and let hive function know its input types and literals.
Advantage: 1.We can support most of hive functions.  2.Just change blink runner.
Disadvantage: Use java reflection will be a little hack, and it's also a temporary solution.
like the following:
https://github.com/JingsongLi/flink/commit/3a74dcebd7342ceb6bb3b4da4f388360918369d5
https://github.com/JingsongLi/flink/commit/6a3f9ecce5343f8cba979809066ef1a17b58d1b3

I prefer Solution II, Hope to get your feedback.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/FunctionCatalog.java
bowenli86: @twalthr Agree. As I mentioned in the comment, I hope to change this API in a followup PR rather than this PR. As such a change would require refactoring upstream code which will complicate this PR itself.

bowenli86:
> Thanks @bowenli86. I added some feedback.
> 
> I think my biggest concern is how we pass input types and literals to functions. This should actually happen at a different level in the stack. For simplification, a function catalog should just lookup functions by name. The returned function definition of the factory should take care of receiving literals and input types. Since function definition does not return `TypeInference` yet, we could also use `getParameterType` and `getReturnType` of UDFs. They don't take literals but I think it is acceptable to postpone supporting this to the next release to still get this feature in before the feature freeze.

@twalthr I'm a bit confused by your comment.

If you've read the code and see its context, the problem I ran into is: `TableFunctionDefinition`'s constructor requires a non-null `resultType`, and `AggregateFunctionDefinition`'s constructor requires non-null `resultType` and `accumulatorType`. For Hive `GenericUDTF`, `UDAF`, `GenericUDAFResolver1/2`, we can only get the `resultType` and `accumulartorType` by initializing these Hive functions with literals and input data types - please see `HiveGenericUDTF.getResultType()` and `HiveGenericUDAF.getResutType()/getAccumularType()`. 

I agree `FunctionDefinitionFactory` should not take literals and input data types, but due to current limitations of the Flink udf framework, I haven't found a good way to bypass it. What do you suggest to pass to constructors of `TableFunctionDefinition` and `AggregateFunctionDefinition` in `HiveFunctionDefinitionFactory#createFunctionDefinition`? A void as `BasicTypeInfo.VOID_TYPE_INFO`?

flink-table/flink-table-common/src/main/java/org/apache/flink/table/factories/FunctionDefinitionFactory.java
xuefuz: Currently HiveSource/Sink need database name and table name to access Hive metastore directly to fetch additional information needed to read/write Hive table, which is why HiveTableFactory provides ObjectPath. Note that CatalogTable itself doesn't carry over this info.

JingsongLi:
Hi bowen:
> `TableFunctionDefinition` and `AggregateFunctionDefinition` need resultType and accumulateType

First, I think the design of `FunctionDefinition` should be optimized, this result type should be an implicit result type instead of real result type, because we can not able to infer the type here, so we can't get the real result type.

In our internal blink we pass `GenericType<Row>` to `TableFunctionDefinition` resultType, and pass `Generic<Object>` to aggregate resultType and pass `GenericType<AggregationBuffer>` to aggregate accumulateType. (See `HiveGenericUDTF` and `HiveGenericUDAF`) I think it's okay to be an implicit result type.

bowenli86:
Hi @JingsongLi , thanks for the explanation!

I think solution II works better for now compared to solution I. I've updated this PR according to your feedback, please take a look. Thanks

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/functions/hive/HiveFunctionDefinitionFactory.java
```diff
@@ -0,0 +1,149 @@
...
+			HiveGenericUDAF udaf = new HiveGenericUDAF(new HiveFunctionWrapper<>(functionClassName), true, hiveVersion);
+
+			return new AggregateFunctionDefinition(
+				name,
+				udaf,
+				GenericTypeInfo.of(Object.class),
+				GenericTypeInfo.of(GenericUDAFEvaluator.AggregationBuffer.class)
+			);
+		} else {
+			throw new IllegalArgumentException(
```
JingsongLi: Minor: Is there any possible to register flink function to hive catalog?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/functions/hive/HiveFunctionDefinitionFactory.java
```diff
@@ -0,0 +1,149 @@
...
+				name,
+				udtf,
+				GenericTypeInfo.of(Row.class)
+			);
+		} else if (GenericUDAFResolver2.class.isAssignableFrom(clazz)) {
+			LOG.info(
+				"Transforming Hive function '{}' into a HiveGenericUDAF with no UDAF bridging and Hive version %s",
+				name, hiveVersion);
+
+			HiveGenericUDAF udaf = new HiveGenericUDAF(new HiveFunctionWrapper<>(functionClassName), false, hiveVersion);
```
JingsongLi: Minor: Extract a method to construct `HiveGenericUDAF` and `AggregateFunctionDefinition`?

flink-table/flink-table-planner/src/main/scala/org/apache/flink/table/plan/rules/logical/PushFilterIntoTableSourceScanRule.scala
JingsongLi: I think there's a problem with the previous code here. There must be no function in the freshly produced `FunctionCatalog`.
But if pass null to FunctionCatalog, there will be a NPE.
How about pass a `new CatalogManager(defaultCatalogName, new GenericInMemoryCatalog)`?

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/FunctionCatalog.java
```diff
@@ -34,26 +36,28 @@
...
 
+	// For simplicity, currently hold registered Flink functions in memory here
+	// TODO: should move to catalog
 	private final Map<String, FunctionDefinition> userFunctions = new LinkedHashMap<>();
 
-	public FunctionCatalog(String defaultCatalogName, String defaultDatabaseName) {
-		this.defaultCatalogName = defaultCatalogName;
-		this.defaultDatabaseName = defaultDatabaseName;
+	public FunctionCatalog(CatalogManager catalogManager) {
+		this.catalogManager = catalogManager;
```
JingsongLi: checkNotNull?

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/api/TableEnvironment.scala
```diff
@@ -0,0 +1,1057 @@
...
+
+/**
+  * The abstract base class for batch and stream TableEnvironments.
+  *
+  * @param streamEnv The [[JavaStreamExecEnv]] which is wrapped in this
+  *                [[StreamTableEnvironment]].
+  * @param config The configuration of the TableEnvironment
+  * @param catalogManager a catalog manager that encapsulates all available catalogs.
+  */
+abstract class TableEnvironment(
```
JingsongLi: What this file for?
I think it has been removed.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/catalog/Catalog.java
```diff
@@ -56,6 +57,16 @@
 		return Optional.empty();
 	}
 
+	/**
+	 * Get an optional {@link FunctionDefinitionFactory} instance that's responsible for generating function definitions
+	 * for catalog functions in this catalog.
+	 *
+	 * @return an optional FunctionDefinitionFactory instance
+	 */
+	default Optional<FunctionDefinitionFactory> getFunctionDefinitionFactory() {
```
twalthr: Do we need an additional method here? Both source/sink and function factories have a common root and we could just have a generic `getTableFactory()` method.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/functions/hive/HiveFunctionDefinitionFactory.java
bowenli86: Timo and I decided to limit the scope of function catalog features in 1.9.0 to be just reading hive functions. Please see our discussions above.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/api/TableEnvironment.scala
bowenli86: Yes, it's due to a bad rebase-to-master. It's been removed

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/FunctionCatalog.java
```diff
@@ -136,18 +185,19 @@ public void registerScalarFunction(String name, ScalarFunction function) {
...
 				.stream()
-				.filter(f -> normalizeName(name).equals(normalizeName(f.getName())))
+				.filter(f -> functionName.equals(normalizeName(f.getName())))
 				.findFirst()
 				.map(Function.identity());
 		}
 
 		return foundDefinition.map(definition -> new FunctionLookup.Result(
-			ObjectIdentifier.of(defaultCatalogName, defaultDatabaseName, name),
+			ObjectIdentifier.of(catalogManager.getCurrentCatalog(), catalogManager.getCurrentDatabase(), name),
```
xuefuz: I'm a little confused/concerned about the identifier created here. From about code it seems that the function can be resolved either in the current catalog (in catalog manager), in "userFunctions", which might conceptually belong to default catalog/database (or built-in catalog), or in BuiltInFunctionDefinitions which doesn't belong to any catalog. As a followup, we might want to unify userFunctions with built-in catalog and clarify the resolution process in multiple catalogs.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/catalog/Catalog.java
bowenli86: make sense.

Side note: I wonder if we should have a better name for `Table`Factory in the near future. As far as I know, it's already used for table, catalog, and function. Though its javadoc describes itself as "A factory to create different table-related instances from string-based properties.", it's still quite confusing to developers - I myself frequently get confused by it and miss points like what you suggested, let alone those who are new and not familiar with this matter

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/FunctionCatalog.java
bowenli86: yes, that's a very good point. I will revert the path of in-memory and built-in functions to its original behavior and follow up to see how to deal with path of built-in functions.

bowenli86:
@twalthr @JingsongLi @xuefuz thanks for your feedbacks. Please take another look

flink-table/flink-table-common/src/main/java/org/apache/flink/table/catalog/Catalog.java
twalthr: Agreed. I would also vote for calling it just `Factory`. But in a backwards compatible-way with proper deprecation.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/FunctionCatalog.java
twalthr: Unifying built-in functions and catalog functions is on our roadmap for 1.10. FunctionCatalog will not be necessary then.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/batch/connectors/hive/HiveTableFactory.java
```diff
@@ -112,4 +141,75 @@ public HiveTableFactory(HiveConf hiveConf) {
 		return new HiveTableSink(new JobConf(hiveConf), tablePath, table);
 	}
 
+	@Override
+	public FunctionDefinition createFunctionDefinition(String name, CatalogFunction catalogFunction) {
```
xuefuz: To have consistent APIs, it might be good to use ObjectPath in place of name.

bowenli86: AFAICT, `createFunctionDefinition()` is a bit different from `createTableSource/Sink()`. Its implementation should be only about how to initialize a func definition from the CatalogFunction. The name is only for building the func definition rather than accessing something from catalog APIs, thus we don't need the database name.

bowenli86:
CI has passed successfully https://travis-ci.com/flink-ci/flink/builds/119001147

@twalthr @JingsongLi @xuefuz can you take another look?

JingsongLi:
LGTM

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/batch/connectors/hive/HiveTableFactory.java
twalthr: Actually, a factory does not need to know the name. `ObjectPath` in `createTableSource/Sink()` is only a temporary solution, right? If we want to pass it for debugging purposes we should pass the entire `ObjectIdentifier`. But I would prefer a clean signature `createXXX(CatalogXXX)`.

bowenli86: @twalthr `HiveTableFactory#createTableSource/Sink()` does need the ObjectPath, not for debugging purpose, but to access partition information via Catalog APIs.

bowenli86:
thank you everyone for the review. rebased to master and running tests now

bowenli86:
CI was successful https://travis-ci.com/flink-ci/flink/builds/119185078
Merging

PR #8925, opened 2019-06-28
[FLINK-12852][network] Fix the deadlock occured when requesting exclusive buffers

gaoyunhaii:

This pull request tries to fix the deadlock problem during requesting exclusive buffers. Since currently the number of maximum buffers and the number of required buffers are not the same for local buffer pools, there may be cases that the local buffer pools of the upstream tasks occupy all the buffers while the downstream tasks fail to acquire exclusive buffers to make progress. Although this problem can be fixed by increasing the number of total buffers, the deadlock may not be acceptable. Therefore, this PR tries to failover the current execution when the deadlock occurs and tips users to increase the number of buffers in the exceptional message.

## Brief change log

The main changes include

1. Add an option for the timeout of `requestMemorySegment` for each channel. The default timeout is 30s. This option is marked as undocumented since it may be removed within the future implementation.
2. Transfer the timeout to `NetworkBufferPool`.
3. `requestMemorySegments` will throw `IOException("Insufficient buffer")`  if not all segments acquired after timeout.

## Verifying this change

1. Added test that validates `requestMemorySegments` end exceptionally if not all segments acquired after timeout.

  - The serializers: **no**

## Documentation

### DIFF
--- a/flink-core/src/main/java/org/apache/flink/configuration/NettyShuffleEnvironmentOptions.java
+++ b/flink-core/src/main/java/org/apache/flink/configuration/NettyShuffleEnvironmentOptions.java
@@ -21,6 +21,7 @@ package org.apache.flink.configuration;
 import org.apache.flink.annotation.PublicEvolving;
 import org.apache.flink.annotation.docs.ConfigGroup;
 import org.apache.flink.annotation.docs.ConfigGroups;
+import org.apache.flink.annotation.docs.Documentation;
 
 import static org.apache.flink.configuration.ConfigOptions.key;
 
@@ -139,6 +140,18 @@ public class NettyShuffleEnvironmentOptions {
 				" The floating buffers are distributed based on backlog (real-time output buffers in the subpartition) feedback, and can" +
 				" help relieve back-pressure caused by unbalanced data distribution among the subpartitions. This value should be" +
 				" increased in case of higher round trip times between nodes and/or larger number of machines in the cluster.");
+
+	/**
+	 * The timeout for requesting exclusive buffers for each channel.
+	 */
+	@Documentation.ExcludeFromDocumentation("This option is purely implementation related, and may be removed as the implementation changes.")
+	public static final ConfigOption<Long> NETWORK_EXCLUSIVE_BUFFERS_REQUEST_TIMEOUT_MILLISECONDS =
+		key("taskmanager.network.memory.exclusive-buffers-request-timeout-ms")
+			.defaultValue(30000L)
+			.withDescription("The timeout for requesting exclusive buffers for each channel. Since the number of maximum buffers and " +
+					"the number of required buffers is not the same for local buffer pools, there may be deadlock cases that the upstream" +
+					"tasks have occupied all the buffers and the downstream tasks are waiting for the exclusive buffers. The timeout breaks" +
+					"the tie by failing the request of exclusive buffers and ask users to increase the number of total buffers.");
 
 	// ------------------------------------------------------------------------
 	//  Netty Options
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/buffer/NetworkBufferPool.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/buffer/NetworkBufferPool.java
@@ -18,6 +18,7 @@
 
 package org.apache.flink.runtime.io.network.buffer;
 
+import org.apache.flink.api.common.time.Deadline;
 import org.apache.flink.configuration.NettyShuffleEnvironmentOptions;
 import org.apache.flink.core.memory.MemorySegment;
 import org.apache.flink.core.memory.MemorySegmentFactory;
@@ -31,6 +32,7 @@ import org.slf4j.LoggerFactory;
 import javax.annotation.Nullable;
 
 import java.io.IOException;
+import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashSet;
@@ -76,7 +78,7 @@ public class NetworkBufferPool implements BufferPoolFactory, MemorySegmentProvid
 	private final long requestSegmentsTimeoutInMillis;
 
 	public NetworkBufferPool(int numberOfSegmentsToAllocate, int segmentSize, int numberOfSegmentsToRequest) {
-		this(numberOfSegmentsToAllocate, segmentSize, numberOfSegmentsToRequest, Long.MAX_VALUE);
+		this(numberOfSegmentsToAllocate, segmentSize, numberOfSegmentsToRequest, Integer.MAX_VALUE);
 	}
 
 	/**
@@ -191,7 +193,7 @@ public class NetworkBufferPool implements BufferPoolFactory, MemorySegmentProvid
 
 		final List<MemorySegment> segments = new ArrayList<>(numberOfSegmentsToRequest);
 		try {
-			long startTimeInNanos = System.nanoTime();
+			final Deadline deadline = Deadline.now().plus(Duration.ofMillis(requestSegmentsTimeoutInMillis));
 			while (true) {
 				if (isDestroyed) {
 					throw new IllegalStateException("Buffer pool is destroyed.");
@@ -206,7 +208,7 @@ public class NetworkBufferPool implements BufferPoolFactory, MemorySegmentProvid
 					break;
 				}
 
-				if ((System.nanoTime() - startTimeInNanos) / 1_000_000 >= requestSegmentsTimeoutInMillis) {
+				if (!deadline.hasTimeLeft()) {
 					throw new IOException(String.format("Insufficient number of network buffers: " +
 									"requesting exclusive buffers timeout. The total number of network " +
 									"buffers is currently set to %d of %d bytes each. You can increase this " +
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/buffer/NetworkBufferPool.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/buffer/NetworkBufferPool.java
@@ -73,16 +73,30 @@ public class NetworkBufferPool implements BufferPoolFactory, MemorySegmentProvid
 
 	private final int numberOfSegmentsToRequest;
 
+	private final long requestSegmentsTimeoutInMillis;
+
+	public NetworkBufferPool(int numberOfSegmentsToAllocate, int segmentSize, int numberOfSegmentsToRequest) {
+		this(numberOfSegmentsToAllocate, segmentSize, numberOfSegmentsToRequest, Long.MAX_VALUE);
+	}
+
 	/**
 	 * Allocates all {@link MemorySegment} instances managed by this pool.
 	 */
-	public NetworkBufferPool(int numberOfSegmentsToAllocate, int segmentSize, int numberOfSegmentsToRequest) {
+	public NetworkBufferPool(
+		int numberOfSegmentsToAllocate,
+		int segmentSize,
+		int numberOfSegmentsToRequest,
+		long requestSegmentsTimeoutInMillis) {
 
 		this.totalNumberOfMemorySegments = numberOfSegmentsToAllocate;
 		this.memorySegmentSize = segmentSize;
 
 		checkArgument(numberOfSegmentsToRequest > 0, "The number of required buffers should be larger than 0.");
 		this.numberOfSegmentsToRequest = numberOfSegmentsToRequest;
+
+		checkArgument(requestSegmentsTimeoutInMillis > 0,
+				"The timeout for requesting exclusive buffers should be larger than 0.");
+		this.requestSegmentsTimeoutInMillis = requestSegmentsTimeoutInMillis;
 
 		final long sizeInLong = (long) segmentSize;
 
@@ -177,6 +191,7 @@ public class NetworkBufferPool implements BufferPoolFactory, MemorySegmentProvid
 
 		final List<MemorySegment> segments = new ArrayList<>(numberOfSegmentsToRequest);
 		try {
+			long startTimeInNanos = System.nanoTime();
 			while (segments.size() < numberOfSegmentsToRequest) {
 				if (isDestroyed) {
 					throw new IllegalStateException("Buffer pool is destroyed.");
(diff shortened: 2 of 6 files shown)


flink-runtime/src/main/java/org/apache/flink/runtime/io/network/buffer/NetworkBufferPool.java
```diff
@@ -181,6 +196,19 @@ public void recycle(MemorySegment segment) {
 				if (segment != null) {
 					segments.add(segment);
 				}
+
+				if ((System.nanoTime() - startTimeInNanos) / 1_000_000 >= requestSegmentsTimeoutInMillis &&
```
Aitozi: The `poll` wait time is in `SECONDS` unit, I think here nanos unit is not necessary. Just in millis?

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/buffer/NetworkBufferPool.java
```diff
@@ -181,6 +196,19 @@ public void recycle(MemorySegment segment) {
...
+						segments.size() < numberOfSegmentsToRequest) {
+					throw new IOException(String.format("Insufficient number of network buffers: " +
```
Aitozi: And before the real request action, it has checked that `numTotalRequiredBuffers + numberOfSegmentsToRequest > totalNumberOfMemorySegments` above, can this situation happen?

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/buffer/NetworkBufferPool.java
gaoyunhaii: The nanos unit is due to that I thought `System.nanoTime()` should have less overhead than `System.currentTimeInMills()`, Therefore, it might be better to use `System.nanoTime()`.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/buffer/NetworkBufferPool.java
gaoyunhaii: This might happen if the configured number of buffers on TM is less than the `required` one.  On TM, the number of buffers in local buffer pool is defined by two values, namely `[required, max)`, and each RemoteInputChannel will also counts its exclusive buffers as `required` buffers. If the sum of the `required` number of buffers has exceeded the number of buffers in the NetworkBufferPool, the above situation happens.

gaoyunhaii:
Very thanks @Aitozi  for the review !   :)

StephanEwen:
This issue is super relevant, but I think we need consensus on which way we want to address this issue first.
Maybe the timeout solution here is correct, but its not yet clear to me.

I think we need to discuss that on the Jira issue first.

StephanEwen:
Some observation at a first glance:
  - It would increase code readability if you used `Duration` (from Java) and `Deadline` (from Flink) rather than working with longs and system time.
  - The test could probably use a much lower timeout (like 50ms) rather than 1s. These comparatively long timeouts eventually add up to long testing times across all tests.

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/buffer/NetworkBufferPoolTest.java
```diff
@@ -477,4 +477,46 @@ public void go() throws Exception {
...
+			@Override
+			public void go() throws Exception {
+				globalPool.requestMemorySegments();
+			}
+		};
+
+		asyncRequest.start();
+
+		try {
+			asyncRequest.trySync(requestSegmentsTimeoutInMillis * 10);
```
StephanEwen: This probably leads to a flaky test. It should be okay to sync "forever" because if the feature is broken, the test will deadlock.

gaoyunhaii: Agree with that the limit might cause flaky test, and changed to use forever sync and detect the exception with `expectedException`.

gaoyunhaii:
Very thanks @StephanEwen for the review! I have modified the PR according to the comments.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/buffer/NetworkBufferPool.java
```diff
@@ -177,7 +193,8 @@ public void recycle(MemorySegment segment) {
 
 		final List<MemorySegment> segments = new ArrayList<>(numberOfSegmentsToRequest);
 		try {
-			while (segments.size() < numberOfSegmentsToRequest) {
+			final Deadline deadline = Deadline.now().plus(Duration.ofMillis(requestSegmentsTimeoutInMillis));
```
pnowojski: This is probably a system call (which might be quite costly), could we restructure the code in a such way, that we make this only after first timeout to `availableMemorySegments.poll(2, TimeUnit.SECONDS);`?

Or even remove it entirely: if `poll(2, TimeUnit.SECONDS)` returns `null` maybe we can assume that 2 seconds have passed and subtract that from the timeout time?

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/buffer/NetworkBufferPool.java
```diff
@@ -186,6 +203,22 @@ public void recycle(MemorySegment segment) {
 				if (segment != null) {
 					segments.add(segment);
 				}
+
+				if (segments.size() >= numberOfSegmentsToRequest) {
+					break;
+				}
+
+				if (!deadline.hasTimeLeft()) {
+					throw new IOException(String.format("Insufficient number of network buffers: " +
```
pnowojski: This exception throwing looks duplicated. If the message is indeed identical everywhere, can you extract it to `throwInsufficientNumberOfNetworkBuffers()` method?

or maybe we extract the construction of the 
> The total number of network buffers is currently set to %d of %d bytes each. You can increase this number by setting the configuration keys '%s', '%s', and '%s'."

message to a common method, and here throw a slightly modified exception with a mention that "timeout was triggered, and you also might want to increase the timeout value"?

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/buffer/NetworkBufferPool.java
```diff
@@ -186,6 +203,22 @@ public void recycle(MemorySegment segment) {
 				if (segment != null) {
 					segments.add(segment);
 				}
+
```
pnowojski: This method has grown a bit. Could we extract some parts of it to separate methods?  Maybe put
```
			if (numTotalRequiredBuffers + numberOfSegmentsToRequest > totalNumberOfMemorySegments) {
				throw new IOException(String.format("Insufficient number of network buffers: " +
								"required %d, but only %d available. The total number of network " +
								"buffers is currently set to %d of %d bytes each. You can increase this " +
								"number by setting the configuration keys '%s', '%s', and '%s'.",
					numberOfSegmentsToRequest,
						totalNumberOfMemorySegments - numTotalRequiredBuffers,
						totalNumberOfMemorySegments,
						memorySegmentSize,
						NettyShuffleEnvironmentOptions.NETWORK_BUFFERS_MEMORY_FRACTION.key(),
						NettyShuffleEnvironmentOptions.NETWORK_BUFFERS_MEMORY_MIN.key(),
						NettyShuffleEnvironmentOptions.NETWORK_BUFFERS_MEMORY_MAX.key()));
			}

			try {
				redistributeBuffers();
			} catch (Throwable t) {
				this.numTotalRequiredBuffers -= numberOfSegmentsToRequest;

				try {
					redistributeBuffers();
				} catch (IOException inner) {
					t.addSuppressed(inner);
				}
				ExceptionUtils.rethrowIOException(t);
			}
``` 
into a separate `tryRedistributeBuffers()` method?

flink-core/src/main/java/org/apache/flink/configuration/NettyShuffleEnvironmentOptions.java
```diff
@@ -140,6 +141,18 @@
 				" help relieve back-pressure caused by unbalanced data distribution among the subpartitions. This value should be" +
 				" increased in case of higher round trip times between nodes and/or larger number of machines in the cluster.");
 
+	/**
+	 * The timeout for requesting exclusive buffers for each channel.
+	 */
+	@Documentation.ExcludeFromDocumentation("This option is purely implementation related, and may be removed as the implementation changes.")
+	public static final ConfigOption<Long> NETWORK_EXCLUSIVE_BUFFERS_REQUEST_TIMEOUT_MILLISECONDS =
+		key("taskmanager.network.memory.exclusive-buffers-request-timeout-ms")
+			.defaultValue(30000L)
```
pnowojski: I don't know what's the good default value for this timeout. @StephanEwen do you have any proposal?

StephanEwen: I think 30s sounds good for a start.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/buffer/NetworkBufferPool.java
```diff
@@ -73,17 +75,31 @@
...
 	/**
 	 * Allocates all {@link MemorySegment} instances managed by this pool.
 	 */
-	public NetworkBufferPool(int numberOfSegmentsToAllocate, int segmentSize, int numberOfSegmentsToRequest) {
+	public NetworkBufferPool(
+		int numberOfSegmentsToAllocate,
+		int segmentSize,
+		int numberOfSegmentsToRequest,
+		long requestSegmentsTimeoutInMillis) {
```
StephanEwen: I would make this a Duration as well. Always simpler, because it encapsulates the unit, and there is no accidental confusion between millis, nanos, secs, etc.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/buffer/NetworkBufferPool.java
StephanEwen: Shortcut: `Deadline.fromNow(...)`

StephanEwen: I would favor simplicity/robustness here. It is a rare call after all.

gaoyunhaii: Changed to `fromNow`

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/buffer/NetworkBufferPool.java
gaoyunhaii: Agree with that an encapsulated object should be more clear and changed the related variables to Duration in `NettyShuffleEnvironmentConfiguration`, `NetworkBufferPool` and `NettyShuffleEnvironmentBuilder`.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/buffer/NetworkBufferPool.java
gaoyunhaii: Agree with that the method seems too long and I have put this part into separated `tryRedistributeBuffers` method.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/buffer/NetworkBufferPool.java
gaoyunhaii: Very thanks for the comments, I preferred to the second method since the timeout should need a different exception message. I have extracted the construction and changed the exception message for requesting timeout.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/buffer/NetworkBufferPool.java
gaoyunhaii: Very thanks for all the comments @pnowojski and @StephanEwen. for the two options to optimize the performance, I prefer to the second option since the number of exclusive buffers might be larger than 1 and factoring out the first `poll` might not reduce the overhead for normal cases. 

For the second one, to not increase much complexity we may reuse the operations of Duration like 

<pre>
//private static final Duration SEGMENT_POLL_TIME_ELAPSED = Duration.ofSeconds(2);

Duration remainTimes = Duration.ofNanos(requestSegmentsTimeout.toNanos());

while (true) {
	if (isDestroyed) {
		throw new IllegalStateException("Buffer pool is destroyed.");
	}

	final MemorySegment segment = availableMemorySegments.poll(2, TimeUnit.SECONDS);
	if (segment != null) {
		segments.add(segment);
	}

	if (segments.size() >= numberOfSegmentsToRequest) {
		break;
	}

	remainTimes = remainTimes.minus(SEGMENT_POLL_TIME_ELAPSED);

	if (remainTimes.toNanos() <= 0) {
		throw new IOException(String.format("Timeout triggered when requesting exclusive buffers: %s, " +
						" or you may increase the timeout which is %dms by setting the key '%s'.",
				getConfigDescription(),
				requestSegmentsTimeout.toMillis(),
				NettyShuffleEnvironmentOptions.NETWORK_EXCLUSIVE_BUFFERS_REQUEST_TIMEOUT_MILLISECONDS.key()));
	}
}
</pre>

On the other side, indeed the method is only called when creating the single input channels during task creation and should not affect the buffer transmission/receive. Reusing `Deadline` should be able to hide the computation logic of the timeout. So do we prefer to the current `Deadline` method or the above `Duration` method ?

gaoyunhaii:
Very thanks for the review @pnowojski @StephanEwen ! I have updated the PR.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/buffer/NetworkBufferPool.java
pnowojski: Let's stick with the `Deadline` as in your proposal there is already a small bug (for the timeout based on subtraction to work, you would have to subtract it only if `segment == null`, otherwise even if get the buffer after only a 1 millisecond of waiting, you subtract 2 seconds of waiting time).

PR #8927, opened 2019-07-01
[FLINK-13036] [elasticsearch] ActionRequestFailureHandler can access RuntimeContext

FaxianZhao:

This pull request makes RuntimeContext is accessable for ActionRequestFailureHandler.
We can count exceptions via accumulator in ActionRequestFailureHandler instead of log, ignore them only.
By the way we can check message count end-to-end consistency.

## Verifying this change

  - The serializers: (no)

## Documentation


SuXingLee:
The PR 's title should be `[FLINK-13036] [elasticsearch]` ,wrong title make  error issue links.

aljoscha:
I'd say we cannot add access to the RuntimeContext here. That interface is quite a bit too "big" and allows access to many things that these simple Elasticsearch handlers shouldn't be able to access.

PR #8928, opened 2019-07-01
[FLINK-12997][coordination] Release partitions when a vertex is reset

zentol:
If an execution finishes it doesn not issue release calls for partitions as they could be re-used later. If the corresponding vertex is reset however we still have to release the partitions that the (now old) execution produced.

This PR implements this behavior.

### DIFF
--- a/flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/ExecutionGraphTestUtils.java
+++ b/flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/ExecutionGraphTestUtils.java
@@ -18,15 +18,19 @@
 
 package org.apache.flink.runtime.executiongraph;
 
-import org.apache.flink.api.common.ExecutionConfig;
 import org.apache.flink.api.common.JobID;
 import org.apache.flink.api.common.time.Deadline;
 import org.apache.flink.api.common.time.Time;
 import org.apache.flink.configuration.Configuration;
+import org.apache.flink.metrics.MetricGroup;
 import org.apache.flink.metrics.groups.UnregisteredMetricsGroup;
+import org.apache.flink.runtime.JobException;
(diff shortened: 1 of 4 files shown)


flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/ExecutionVertexTest.java
```diff
@@ -0,0 +1,85 @@
...
+		producerExecutionJobVertex.resetForNewExecution(1L, 1L);
+
+		final IntermediateResultPartitionID intermediateResultPartitionID = producerExecutionJobVertex
+			.getProducedDataSets()[0]
+			.getPartitions()[0]
+			.getPartitionId();
+		final ResultPartitionID resultPartitionID = execution
+			.getResultPartitionDeploymentDescriptor(intermediateResultPartitionID)
+			.get()
+			.getShuffleDescriptor().getResultPartitionID();
```
GJL: maybe break the line between method calls for consistency

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/ExecutionGraphTestUtils.java
```diff
@@ -537,4 +529,137 @@ public static void verifyGeneratedExecutionJobVertex(
...
+		}
+
+		public TestingExecutionGraphBuilder setPartitionTracker(final PartitionTracker partitionTracker) {
+			this.partitionTracker = partitionTracker;
+			return this;
+		}
+
+		public ExecutionGraph build() throws JobException, JobExecutionException {
+			return ExecutionGraphBuilder.buildGraph(
+				priorExecutionGraph,
```
GJL: Not sure if we need to include `priorExecutionGraph` in the builder. I cannot find a usage where `prior` is non-null.

PR #8931, opened 2019-07-01
[FLINK-13041] Make ScheduleMode configurable via ExecutionConfig

aljoscha:

The new Blink-based Table API Runner currently needs a way of globally setting the `ScheduleMode` (either `LAZY_FROM_SOURCES` or `EAGER`) to support both batch-style and streaming-style execution.

This is a stop-gap solution. Ideally, `Transformations` would know whether they are bounded or not and the scheduler could decide the scheduling mode based on that information but this requires more work on both the Table API Runner and the DAG API.

## Brief change log

*(for example:)*
  - Move `ScheduleMode` from `flink-runtime` to `flink-core`
  - Make `ScheduleMode` configurable via `ExecutionConfig`

## Verifying this change

  - Added a test in `StreamExecutionEnvironmentTest`

  - The serializers: no

## Documentation


aljoscha:
R: @ifndef-SleePy

flink-runtime/src/test/java/org/apache/flink/runtime/jobgraph/ScheduleModeTest.java
```diff
@@ -18,6 +18,8 @@
 
 package org.apache.flink.runtime.jobgraph;
```
ifndef-SleePy: Move the unit test to api.common package?

flink-core/src/main/java/org/apache/flink/api/common/ExecutionConfig.java
```diff
@@ -90,6 +90,8 @@
 	/** Defines how data exchange happens - batch or pipelined */
 	private ExecutionMode executionMode = ExecutionMode.PIPELINED;
 
+	private ScheduleMode scheduleMode = ScheduleMode.LAZY_FROM_SOURCES;
```
zhuzhurk: Does this mean if users do not specify the `ScheduleMode `, even stream jobs will be executed in batch style?

ifndef-SleePy: Good point!
It changes default behavior of `DataStream` job, `EAGER` -> `LAZY_FROM_SOURCE`.
Should we move `ScheduleMode` from `ExecutionConfig` to `StreamExecutionEnvironment` since `ExecutionConfig` is also used by `DataSet`? `DataSet` need `LAZY_FROM_SOURCE`, `DataStream` need `EAGER` by default.

aljoscha: Ha! you got me! I left this in on purpose because I wanted to see if all the tests still pass with this different default. We should of course not have this in the final code.

I put it in `ExecutionConfig` because it is available in `flink-core` while `StreamExecutionEnvironment` is not. I had a feeling that maybe the Table API will need access to this at some point or other. What do you think?

flink-runtime/src/test/java/org/apache/flink/runtime/jobgraph/ScheduleModeTest.java
aljoscha: good point! I'm moving

flink-core/src/main/java/org/apache/flink/api/common/ExecutionConfig.java
zhuzhurk: Changing it to EAGER by default may have `DataSet` job get scheduled in streaming style, which might be unexpected. Shall we override it to be LAZY_FROM_SOURCES by default for DataSet?

flink-core/src/main/java/org/apache/flink/api/common/ExecutionConfig.java
```diff
@@ -90,6 +90,8 @@
 	/** Defines how data exchange happens - batch or pipelined */
 	private ExecutionMode executionMode = ExecutionMode.PIPELINED;
 
+	private ScheduleMode scheduleMode = ScheduleMode.EAGER;
```
zhuzhurk: The new field may need to take part in ExecutionConfig's `equals` and `hashCode` calculation.

flink-core/src/main/java/org/apache/flink/api/common/ExecutionConfig.java
ifndef-SleePy: I have a concern that `DataSet` would ignore this `ScheduleMode` setting. It looks a bit imperfect. Unless you think we should not care about `DataSet` anymore? 

Regard to the `flink-core` part, it's an interesting point that I didn't notice. Yes, I agree with you. I also have a strong feeling that Table API will need it sooner or later. I was thinking Table API may generate `StreamGraph` itself in the future. In that way, Table API could set this property directly. But it's hard to say what it will look like.

aljoscha: So moving it to `StreamExecutionEnvironment` might make sense after all, and not have it in `ExecutionConfig`, I think. (Still thinking ...)

StephanEwen:
I am not a fan of this solution. This exposes the schedule mode to the public DataStream API. Users should not have access to this.

ifndef-SleePy:
Hi @StephanEwen ,
Yes, as Aljoscha described in JIRA issue [1], it's a stop-gap solution. Here are some contexts [2][3] of this solution.
It's quite a long story. In a nutshell, there is no appropriate way to allow Blink planner to set schedule mode at the moment. That would be nice if you could give me some suggestions.

1. https://issues.apache.org/jira/browse/FLINK-13041
2. https://github.com/apache/flink/pull/8894
3. https://docs.google.com/document/d/1uYjO7boR2l4qEBcv34IvRYg5LEYyj9Hr94A3FEnUMis/edit?usp=sharing

aljoscha:
As an alternative, we considered https://issues.apache.org/jira/browse/FLINK-12961: This adds `StreamExecutionEnvironment.execute(StreamGraph)`, adding `StreamGraph` to the public API, which is an even more internal thing that we shouldn't expose.

Another alternative I considered is to enhance `Transformation` (formerly known as `StreamTransformation`) with an `isBounded()` flag that has to be set when creating the transforms. From this, the executing code could derive whether the pipeline is bounded and then pick a scheduling mode. This would require changing the Blink Planner to set this flag properly on sources and to change the DAG classes to properly forward/set this flag and/or the StreamGraphBuilder to understand it and react. I want to tackle this for the DAG representation in the next release but didn't want to go down that route now because:
 - The Blink Runner doesn't work that way currently and therefore needs changes.
 - When adding this to the DAG we need to ensure (properly test) that everything works, which also takes some time to do right.

A third alternative that I thought about is adding a private `setScheduleMode()` method that we invoke via reflection, but that also doesn't seem nice.

@StephanEwen what do you think? We can start on the `isBounded()` solution but it might take some time to implement it both in the API/translation and the Blink Runner. (To me, this is the best solution from an engineering/code quality perspective.)

aljoscha:
Side note: the same is true for https://issues.apache.org/jira/browse/FLINK-13042, where instead of a global flag we can require the Blink Runner to set a unique slot sharing group for each transformation, thereby effectively disabling slot sharing.

aljoscha:
I opened a PR for my last proposal: #8951

aljoscha:
Closing this because we want to do it differently.

PR #8932, opened 2019-07-01
 [FLINK-13029][table-planner] Removed usages of ExpressionBridge in QueryOperation's factories

dawidwys:

The QueryOperation's factories are part of the API, therefore they should not use the `PlannerExpression`s and `ExpressionBridge`.

## Brief change log

  - Ported few type checks to the new type system (window size type, group by key type)
  - Loosen the restrictions for `DataType` -> `TypeInformation` conversion to support value literal types.
  - Removed all usages of `ExpressionBridge` in QueryOperation's

## Verifying this change

* This change is already covered by existing tests.
* Added tests to check value literal types support:
** org.apache.flink.table.api.batch.table.CalcTest#testSelectLiterals
** org.apache.flink.table.api.batch.table.CalcTest#testGroupByLiteral

## Documentation

### DIFF
--- a/flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/utils/LegacyTypeInfoDataTypeConverter.java
+++ b/flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/utils/LegacyTypeInfoDataTypeConverter.java
@@ -48,6 +48,7 @@ import org.apache.flink.table.typeutils.TimeIntervalTypeInfo;
 import org.apache.flink.types.Row;
 import org.apache.flink.util.Preconditions;
 
+import java.time.LocalDateTime;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.stream.IntStream;
@@ -200,7 +201,7 @@ public final class LegacyTypeInfoDataTypeConverter {
 			return Types.STRING;
 		}
 
-		else if (canConvertToTimestampLenient(logicalType)) {
+		else if (canConvertToTimestampTypeInfoLenient(dataType)) {
 			return Types.SQL_TIMESTAMP;
 		}
 
@@ -238,8 +239,11 @@ public final class LegacyTypeInfoDataTypeConverter {
 				dataType.getConversionClass().getName()));
 	}
 
-	private static boolean canConvertToTimestampLenient(LogicalType logicalType) {
-		return hasRoot(logicalType, LogicalTypeRoot.TIMESTAMP_WITHOUT_TIME_ZONE) && LogicalTypeChecks.getPrecision(logicalType) <= 3;
+	private static boolean canConvertToTimestampTypeInfoLenient(DataType dataType) {
+		LogicalType logicalType = dataType.getLogicalType();
+		return hasRoot(logicalType, LogicalTypeRoot.TIMESTAMP_WITHOUT_TIME_ZONE) &&
+			dataType.getConversionClass() != LocalDateTime.class &&
+			LogicalTypeChecks.getPrecision(logicalType) <= 3;
 	}
 
 	private static DataType createLegacyType(LogicalTypeRoot typeRoot, TypeInformation<?> typeInfo) {
--- a/flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/utils/LegacyTypeInfoDataTypeConverter.java
+++ b/flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/utils/LegacyTypeInfoDataTypeConverter.java
@@ -36,11 +36,13 @@ import org.apache.flink.table.types.DataType;
 import org.apache.flink.table.types.FieldsDataType;
 import org.apache.flink.table.types.KeyValueDataType;
 import org.apache.flink.table.types.logical.LegacyTypeInformationType;
+import org.apache.flink.table.types.logical.LogicalType;
 import org.apache.flink.table.types.logical.LogicalTypeRoot;
 import org.apache.flink.table.types.logical.RowType;
 import org.apache.flink.table.types.logical.TimestampKind;
 import org.apache.flink.table.types.logical.TimestampType;
 import org.apache.flink.table.types.logical.TypeInformationAnyType;
+import org.apache.flink.table.types.logical.utils.LogicalTypeChecks;
 import org.apache.flink.table.typeutils.TimeIndicatorTypeInfo;
 import org.apache.flink.table.typeutils.TimeIntervalTypeInfo;
 import org.apache.flink.types.Row;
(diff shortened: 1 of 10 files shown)


flink-table/flink-table-planner/src/main/java/org/apache/flink/table/operations/AggregateOperationFactory.java
```diff
@@ -501,16 +509,38 @@ protected Void defaultMethod(ResolvedExpression expression) {
...
-					"because it's not a valid key type which must be hashable and comparable", expression));
+		@Override
+		public Boolean visit(AtomicDataType atomicDataType) {
+			LogicalType logicalType = atomicDataType.getLogicalType();
+			if (logicalType instanceof LegacyTypeInformationType) {
+				return ((LegacyTypeInformationType) logicalType).getTypeInformation().isKeyType();
+			} else if (logicalType.getTypeRoot() == LogicalTypeRoot.ANY) {
+				// we don't know anything about the ANY type, we don't know if it is comparable and hashable.
+				return false;
+			} else if (logicalType instanceof StructuredType) {
```
twalthr: A structured type is not an AtomicDataType but a FieldsDataType.

flink-table/flink-table-planner/src/main/java/org/apache/flink/table/operations/AggregateOperationFactory.java
```diff
@@ -501,16 +509,38 @@ protected Void defaultMethod(ResolvedExpression expression) {
...
+		}
+
+		@Override
+		public Boolean visit(FieldsDataType fieldsDataType) {
+			return fieldsDataType.getFieldDataTypes().values().stream().allMatch(t -> t.accept(this));
+		}
+
+		@Override
+		public Boolean visit(KeyValueDataType keyValueDataType) {
+			return keyValueDataType.getKeyDataType().accept(this) && keyValueDataType.getValueDataType().accept(this);
```
twalthr: Do we really want to allow grouping on maps and arrays? Do systems like Oracle support this?

flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/utils/LegacyTypeInfoDataTypeConverter.java
```diff
@@ -221,6 +238,10 @@ else if (canConvertToAnyTypeInfo(dataType)) {
 				dataType.getConversionClass().getName()));
 	}
 
+	private static boolean canConvertToTimestampLenient(LogicalType logicalType) {
```
twalthr: match naming scheme: `canConvertToTimestampTypeInfoLenient`

flink-table/flink-table-planner/src/main/java/org/apache/flink/table/operations/JoinOperationFactory.java
```diff
@@ -99,9 +93,8 @@ private void validateCondition(QueryOperation right, JoinType joinType, Resolved
 	}
 
 	private void verifyConditionType(ResolvedExpression condition) {
-		PlannerExpression plannerExpression = expressionBridge.bridge(condition);
-		TypeInformation<?> conditionType = plannerExpression.resultType();
-		if (conditionType != Types.BOOLEAN) {
+		DataType conditionType = condition.getOutputDataType();
+		if (!LogicalTypeChecks.hasRoot(conditionType.getLogicalType(), LogicalTypeRoot.BOOLEAN)) {
```
twalthr: nit: In general I would recommend to store the logical type in a local variable and let the condition look like

```if (hasRoot(conditionType, BOOLEAN))```

for better readability.

File: flink-table/flink-table-planner/src/test/scala/org/apache/flink/table/api/batch/table/CalcTest.scala [hunk 5]
```diff
@@ -117,6 +120,45 @@ class CalcTest extends TableTestBase {
     util.verifyTable(resultTable, expected)
   }
 
+  @Test
+  def testSelectLiterals(): Unit = {
+    val util = batchTestUtil()
+    val sourceTable = util.addTable[Int]("MyTable", 'a)
+    val resultTable = sourceTable.select("ABC", BigDecimal(1234), LocalDateTime.of(1, 1, 1, 1, 1))
```
twalthr: Could you also verify that `LocalDateTime` fields still throw an exception when they are not a literal but a table field?

flink-table/flink-table-planner/src/main/java/org/apache/flink/table/operations/AggregateOperationFactory.java
dawidwys: I did some additional investigation for arrays:
* oracle does not really have ARRAY type. I checked for VARRAY and you cannot group on it.
* PostgreSQL has ARRAY type and you can group on it
* in SQL standard the only comment I could find says that if you try to group on user defined type with UNKNOWN comparison the behavior is implementation specific.

I think it should be fine to allow grouping on array types.

flink-table/flink-table-planner/src/main/java/org/apache/flink/table/operations/AggregateOperationFactory.java
```diff
@@ -501,16 +493,40 @@ protected Void defaultMethod(ResolvedExpression expression) {
 		}
 	}
 
-	private class GroupingExpressionValidator extends ResolvedExpressionDefaultVisitor<Void> {
+	private static class IsKeyTypeChecker implements DataTypeVisitor<Boolean> {
+
 		@Override
-		protected Void defaultMethod(ResolvedExpression expression) {
-			TypeInformation<?> groupingType = expressionBridge.bridge(expression).resultType();
+		public Boolean visit(AtomicDataType atomicDataType) {
```
twalthr: nit: I think we could still simplify this code by just using a `LogicalTypeDefaultVisitor` and iterate over `LogicalType#getChildren`?

flink-table/flink-table-planner/src/main/java/org/apache/flink/table/operations/AggregateOperationFactory.java
twalthr: Thanks for the investigation. Sounds good to me.

flink-table/flink-table-planner/src/main/java/org/apache/flink/table/operations/AggregateOperationFactory.java
dawidwys: Not really. `LegacyTypeInformationType` goes through default method. The same happens with `LegacyTypeInformationType`. We would have exact same code in the `visit` method of the `LogicalTypeDefaultVisitor`

PR #8933, opened 2019-07-01
[FLINK-12968][table-common] Add an utility for finding a common type from a set of types

twalthr:

This adds the last utility required to perform type inference properly. It is meant for finding a common, more general `LogicalType` for a given set of types. If such a type exists, all given types can be casted to this more general type. It aims to be compatible with the SQL standard. It is inspired by Apache Calcite's `SqlTypeFactoryImpl#leastRestrictive` method.

## Brief change log

Utility class added.

## Verifying this change

This change added tests and can be verified as follows: `LogicalTypeGeneralizationTest`

  - The serializers: no

## Documentation

### DIFF
--- a/flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/logical/utils/LogicalTypeGeneralization.java
+++ b/flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/logical/utils/LogicalTypeGeneralization.java
@@ -29,6 +29,7 @@ import org.apache.flink.table.types.logical.DoubleType;
 import org.apache.flink.table.types.logical.LocalZonedTimestampType;
 import org.apache.flink.table.types.logical.LogicalType;
 import org.apache.flink.table.types.logical.LogicalTypeRoot;
+import org.apache.flink.table.types.logical.MapType;
 import org.apache.flink.table.types.logical.MultisetType;
 import org.apache.flink.table.types.logical.NullType;
 import org.apache.flink.table.types.logical.RowType;
@@ -80,6 +81,7 @@ import static org.apache.flink.table.types.logical.LogicalTypeRoot.DECIMAL;
 import static org.apache.flink.table.types.logical.LogicalTypeRoot.DOUBLE;
 import static org.apache.flink.table.types.logical.LogicalTypeRoot.INTERVAL_DAY_TIME;
 import static org.apache.flink.table.types.logical.LogicalTypeRoot.INTERVAL_YEAR_MONTH;
+import static org.apache.flink.table.types.logical.LogicalTypeRoot.MAP;
 import static org.apache.flink.table.types.logical.LogicalTypeRoot.MULTISET;
 import static org.apache.flink.table.types.logical.LogicalTypeRoot.NULL;
 import static org.apache.flink.table.types.logical.LogicalTypeRoot.ROW;
@@ -247,6 +249,8 @@ public final class LogicalTypeGeneralization {
 				return findCommonArrayType(normalizedTypes);
 			} else if (typeRoot == MULTISET) {
 				return findCommonMultisetType(normalizedTypes);
+			} else if (typeRoot == MAP) {
+				return findCommonMapType(normalizedTypes);
 			} else if (typeRoot == ROW) {
 				return findCommonRowType(normalizedTypes);
 			}
@@ -381,6 +385,8 @@ public final class LogicalTypeGeneralization {
 		} else if (hasFamily(resultType, EXACT_NUMERIC) && (hasFamily(type, TIMESTAMP) || hasRoot(type, DATE))) {
 			return type;
 		}
+		// for "DATETIME + EXACT_NUMERIC", EXACT_NUMERIC is always treated as an interval of days
+		// therefore, TIME + EXACT_NUMERIC is not supported
 		return null;
 	}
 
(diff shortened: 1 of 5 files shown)


flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/logical/utils/LogicalTypeGeneralization.java
```diff
@@ -0,0 +1,631 @@
...
+			}
+			// other types are handled by findCommonCastableType
+			else {
+				return null;
+			}
+		}
+
+		// NULL type only
+		if (resultType == null && hasNullType) {
+			return new NullType();
```
dawidwys: I think this branch is unreachable.
`resultType` may be null only if `normalizedTypes` is empty (because of line 234), but then `hasNullType` cannot be `true` as we do not have types at all.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/logical/utils/LogicalTypeGeneralization.java
```diff
@@ -0,0 +1,631 @@
...
+			if (type.getChildren().size() != numberOfChildren) {
+				return null;
+			}
+		}
+
+		// recursively compute column-wise least restrictive
+		final List<LogicalType> resultChildren = new ArrayList<>(numberOfChildren);
+		for (int i = 0; i < numberOfChildren; i++) {
+			final int childPos = i;
+			final Optional<LogicalType> childType = findCommonType(new AbstractList<LogicalType>() {
```
dawidwys: I would extract this anonymous class to give it a better name and maybe a bit of explanation. 

I guess we could name it `ChildTypeView` or sth similar `*View`?

flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/logical/utils/LogicalTypeGeneralization.java
```diff
@@ -0,0 +1,631 @@
...
+			T right) {
+		final List<T> leftBoundaries = resToBoundaries.get(left);
+		final T leftStart = leftBoundaries.get(0);
+		final T leftEnd = leftBoundaries.get(leftBoundaries.size() - 1);
+
+		final List<T> rightBoundaries = resToBoundaries.get(right);
+		final T rightStart = rightBoundaries.get(0);
+		final T rightEnd = rightBoundaries.get(rightBoundaries.size() - 1);
+
+		final T combinedStart = res[Math.min(leftStart.ordinal(), rightStart.ordinal())];
```
dawidwys: If we want to depend on the ordinal of `Resolution`s let's at least add a comment next to those classes that the order has to be maintained.

To be honest though, I would feel safer if those enums implemented `Comparable`.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/logical/utils/LogicalTypeGeneralization.java
```diff
@@ -0,0 +1,631 @@
...
+			}
+
+			// for types of family CONSTRUCTED
+			if (typeRoot == ARRAY) {
+				return findCommonArrayType(normalizedTypes);
+			} else if (typeRoot == MULTISET) {
+				return findCommonMultisetType(normalizedTypes);
+			} else if (typeRoot == ROW) {
+				return findCommonRowType(normalizedTypes);
+			}
```
dawidwys: missing `MAP` typeRoot

flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/logical/utils/LogicalTypeGeneralization.java
```diff
@@ -0,0 +1,631 @@
...
+		}
+		return firstType;
+	}
+
+	private static @Nullable LogicalType findCommonTypePattern(LogicalType resultType, LogicalType type) {
+		if (hasFamily(resultType, DATETIME) && hasFamily(type, INTERVAL)) {
+			return resultType;
+		} else if (hasFamily(resultType, INTERVAL) && hasFamily(type, DATETIME)) {
+			return type;
+		} else if ((hasFamily(resultType, TIMESTAMP) || hasRoot(resultType, DATE)) && hasFamily(type, EXACT_NUMERIC)) {
```
dawidwys: Out of curiosity why doesn't it work for `TIME_WITHOUT_TIME_ZONE`?

If I understand correctly: `DATETIME - TIME_WITHOUT_TIME_ZONE = TIMESTAMP + DATE`

flink-table/flink-table-common/src/test/java/org/apache/flink/table/types/LogicalTypeGeneralizationTest.java
```diff
@@ -0,0 +1,310 @@
...
+				// year-month intervals
+				{
+					Arrays.asList(
+						new YearMonthIntervalType(YearMonthResolution.MONTH),
+						new YearMonthIntervalType(YearMonthResolution.YEAR)),
+					new YearMonthIntervalType(YearMonthResolution.YEAR_TO_MONTH)
+				},
+
+			}
+		);
```
dawidwys: I think we should add test case for Date + Interval

zentol:
CI report for commit b9c2383e2b8a7608f5269e03888f149996ba9f4e: SUCCESS [Travis](https://travis-ci.org/flink-ci/flink-ci/builds/554872233?utm_source=github_status&utm_medium=notification)

flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/logical/utils/LogicalTypeGeneralization.java
twalthr: I added the following comment:
```
// for "DATETIME + EXACT_NUMERIC", EXACT_NUMERIC is always treated as an interval of days
// therefore, TIME + EXACT_NUMERIC is not supported
```

This is also in sync with Oracle's behavior.

twalthr:
Thanks for the review @dawidwys. I addressed your feedback and will merge this...

PR #8934, opened 2019-07-01
[FLINK-12628][Runtime / Coordination] Check test failure if partitionhas no consumers in Execution.getPartitionMaxParallelism

liyafan82:

Resolve issue FLINK-12628 Check test failure if partition has no consumers in Execution.getPartitionMaxParallelism:

Currently, we work around this case in Execution.getPartitionMaxParallelism because of tests:

// TODO consumers.isEmpty() only exists for test, currently there has to be exactly one consumer in real jobs!

though partition is supposed to have always at least one consumer atm.
We should check which test fails and consider fixing it.

According to my investigation, there is no test failure, when we ignore the case for consumers.isEmpty() equals to true.

## Brief change log

  - Change the implementation of Execution.getPartitionMaxParallelism to ignore the case for consumers.isEmpty() equals to true.

## Verifying this change

This change is already covered by existing tests, such as ExecutionTest.

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/Execution.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/Execution.java
@@ -57,7 +57,6 @@ import org.apache.flink.runtime.shuffle.PartitionDescriptor;
 import org.apache.flink.runtime.shuffle.ProducerDescriptor;
 import org.apache.flink.runtime.shuffle.ShuffleDescriptor;
 import org.apache.flink.runtime.shuffle.ShuffleMaster;
-import org.apache.flink.runtime.state.KeyGroupRangeAssignment;
 import org.apache.flink.runtime.taskmanager.TaskManagerLocation;
 import org.apache.flink.util.ExceptionUtils;
 import org.apache.flink.util.FlinkException;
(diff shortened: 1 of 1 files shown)


flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/Execution.java
```diff
@@ -679,14 +678,10 @@ public void setInitialState(@Nullable JobManagerTaskRestore taskRestore) {
...
 	private static int getPartitionMaxParallelism(IntermediateResultPartition partition) {
-		// TODO consumers.isEmpty() only exists for test, currently there has to be exactly one consumer in real jobs!
 		final List<List<ExecutionEdge>> consumers = partition.getConsumers();
-		int maxParallelism = KeyGroupRangeAssignment.UPPER_BOUND_MAX_PARALLELISM;
-		if (!consumers.isEmpty()) {
-			List<ExecutionEdge> consumer = consumers.get(0);
-			ExecutionJobVertex consumerVertex = consumer.get(0).getTarget().getJobVertex();
-			maxParallelism = consumerVertex.getMaxParallelism();
-		}
+		List<ExecutionEdge> consumer = consumers.get(0);
```
klion26: As the previous comment said, `consumers will be empty in test`, so `consumers.get(0)` will throw an exception here?

liyafan82: Maybe not. In my local environment, no exception was thrown when we ignore the case for the consumer being empty. Let's see if we can reproduce it in Travis.

azagrebin: We can also add a check if this assumption holds to better debug it if fails:
`Preconditions.checkArgument(!consumers.isEmpty(), "Currently there has to be exactly one consumer in real jobs");`

azagrebin:
@liyafan82 
I rebased the PR on the latest master and added the precondition check, I mentioned in the comment.
Do you have time to work on this?
We moved recently to other in house Travis CI. It looks like it is triggered only for new PRs. We could close this PR and open a new one to see Travis results.

liyafan82:
> @liyafan82
> I rebased the PR on the latest master and added the precondition check, I mentioned in the comment.
> Do you have time to work on this?
> We moved recently to other in house Travis CI. It looks like it is triggered only for new PRs. We could close this PR and open a new one to see Travis results.

Sure. I will start a new PR for this. Thanks.

azagrebin:
@liyafan82 
the CI actually reported success, we can continue with this PR so I will reopen it, sorry for confusion.

liyafan82:
> @liyafan82
> the CI actually reported success, we can continue with this PR so I will reopen it, sorry for confusion.

Sure. Thanks for your effort.

PR #8936, opened 2019-07-01
[FLINK-13043][Library / CEP] Fix the bug of parsing Dewey number from string

liyafan82:

There is a bug in the current implementation for parsing the Dewey number:

 

String[] splits = deweyNumberString.split("\\.");

if (splits.length == 0){
 return new DeweyNumber(Integer.parseInt(deweyNumberString)); 
}
 

The length in the if condition should be 1 instead of 0.

## Brief change log

  - Fix the problem by modifying the implementation of DeweyNumber#fromString
  

## Verifying this change

This change is already covered by existing tests, such as DeweyNumberTest.

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-libraries/flink-cep/src/main/java/org/apache/flink/cep/nfa/DeweyNumber.java
+++ b/flink-libraries/flink-cep/src/main/java/org/apache/flink/cep/nfa/DeweyNumber.java
@@ -175,9 +175,9 @@ public class DeweyNumber implements Serializable {
 	public static DeweyNumber fromString(final String deweyNumberString) {
 		String[] splits = deweyNumberString.split("\\.");
 
-		if (splits.length == 0) {
+		if (splits.length == 1) {
 			return new DeweyNumber(Integer.parseInt(deweyNumberString));
-		} else {
+		} else if (splits.length > 0) {
 			int[] deweyNumber = new int[splits.length];
 
 			for (int i = 0; i < splits.length; i++) {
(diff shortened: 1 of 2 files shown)


flink-libraries/flink-cep/src/main/java/org/apache/flink/cep/nfa/DeweyNumber.java
```diff
@@ -175,16 +175,18 @@ public String toString() {
 	public static DeweyNumber fromString(final String deweyNumberString) {
 		String[] splits = deweyNumberString.split("\\.");
 
-		if (splits.length == 0) {
+		if (splits.length == 1) {
 			return new DeweyNumber(Integer.parseInt(deweyNumberString));
-		} else {
+		} else if (splits.length > 0) {
```
Myasuka: In which case would the returned `splits` has a length not larger than `0`? I think the condition `if (splits.length > 0)` might be redundant.

liyafan82: @Myasuka thanks for the comments. 
In some rare cases, the length can be 0. So this condition guards against such cases.

StephanEwen:
This fix needs a test before it can be merged...

StephanEwen:

liyafan82:
> This fix needs a test before it can be merged...

Test added. Please take a look. Thanks.

StephanEwen:
@flinkbot approve all

StephanEwen:
Looks good, merging this...

PR #8937, opened 2019-07-01
[FLINK-13040] promote blink table config and add to document

XuPingyong:

1. promote planner and runtime config of blink
2. add anotations to indicate the config works for batch, streaming sql or both.
3. add these configs to document

## Brief change log

1. promote planner and runtime config of blink
2. add anotations to indicate the config works for batch, streaming sql or both.
3. add these configs to document

## Verifying this change

no. Compatible is no need to be considered.

  - The serializers: (no)

## Documentation

The preview screenshot of configuration page looks like:
image

### DIFF
--- a/docs/ops/config.md
+++ b/docs/ops/config.md
@@ -183,12 +183,6 @@ unless user define a `OptionsFactory` and set via `RocksDBStateBackend.setOption
 
 {% include generated/rocks_db_configurable_configuration.html %}
 
-### blink table planner
-{% include generated/planner_config_configuration.html %}
-
-### blink table runtime
-{% include generated/table_config_configuration.html %}
-
 ### Queryable State
 
 {% include generated/queryable_state_configuration.html %}
@@ -214,6 +208,13 @@ You have to configure `jobmanager.archive.fs.dir` in order to archive terminated
 - `jobmanager.archive.fs.dir`: Directory to upload information about terminated jobs to. You have to add this directory to the list of monitored directories of the history server via `historyserver.archive.fs.dir`.
 
 {% include generated/history_server_configuration.html %}
+
+
+### Blink Table Planner
+{% include generated/planner_config_configuration.html %}
+
+### Blink Table Runtime
+{% include generated/table_config_configuration.html %}
 
 ## Legacy
 
--- a/docs/ops/config.md
+++ b/docs/ops/config.md
@@ -183,6 +183,12 @@ unless user define a `OptionsFactory` and set via `RocksDBStateBackend.setOption
 
 {% include generated/rocks_db_configurable_configuration.html %}
 
+### blink table planner
+{% include generated/planner_config_configuration.html %}
+
+### blink table runtime
+{% include generated/table_config_configuration.html %}
+
 ### Queryable State
 
 {% include generated/queryable_state_configuration.html %}
--- a/flink-annotations/src/main/java/org/apache/flink/annotation/docs/Documentation.java
+++ b/flink-annotations/src/main/java/org/apache/flink/annotation/docs/Documentation.java
@@ -60,18 +60,17 @@ public final class Documentation {
 	}
 
 	/**
-	 * Annotation used on config option fields to include them in the "Table Options" section.
+	 * Annotation used on table config to introduce additional table-related meta data.
 	 *
-	 * <p>The {@link TableOption#execMode()} argument indicate which exec mode the config works for,
+	 * <p>The {@link TableMeta#execMode()} argument indicates which exec mode the config works for,
 	 * for batch, streaming or both.
 	 *
 	 */
 	@Target(ElementType.FIELD)
 	@Retention(RetentionPolicy.RUNTIME)
 	@Internal
-	public @interface TableOption {
-
-		ExecMode execMode() default ExecMode.BOTH;
+	public @interface TableMeta {
+		ExecMode execMode();
 	}
 
 	/**
--- a/flink-annotations/src/main/java/org/apache/flink/annotation/docs/Documentation.java
+++ b/flink-annotations/src/main/java/org/apache/flink/annotation/docs/Documentation.java
@@ -60,6 +60,28 @@ public final class Documentation {
 	}
 
 	/**
+	 * Annotation used on config option fields to include them in the "Table Options" section.
+	 *
+	 * <p>The {@link TableOption#execMode()} argument indicate which exec mode the config works for,
+	 * for batch, streaming or both.
+	 *
+	 */
+	@Target(ElementType.FIELD)
+	@Retention(RetentionPolicy.RUNTIME)
+	@Internal
+	public @interface TableOption {
+
+		ExecMode execMode() default ExecMode.BOTH;
+	}
+
+	/**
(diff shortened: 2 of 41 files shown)


flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/api/PlannerConfigOptions.java
```diff
@@ -30,7 +31,8 @@
 	// ------------------------------------------------------------------------
 	//  Optimizer Options
 	// ------------------------------------------------------------------------
-
+	@Documentation.ExcludeFromDocumentation
```
zentol: please specify reasons for why config options are excluded from the documentation. This applies to all options that are excluded, not just this one.

flink-docs/src/main/java/org/apache/flink/docs/configuration/ConfigOptionsDocGenerator.java
```diff
@@ -260,10 +262,26 @@ private static String toHtmlTable(final List<OptionWithMetaInfo> options) {
...
 		String defaultValue = stringifyDefault(optionWithMetaInfo);
+		Documentation.TableOption tableOption = optionWithMetaInfo.field.getAnnotation(Documentation.TableOption.class);
+		StringBuilder sb = new StringBuilder();
+		if (tableOption != null) {
+			Documentation.ExecMode execMode = tableOption.execMode();
+			if (Documentation.ExecMode.BATCH.equals(execMode) || Documentation.ExecMode.STREAMING.equals(execMode)) {
+				sb.append("<br> <span class=\"label label-primary\">")
+						.append(execMode.toString())
+						.append("</span>");
+			} else if (Documentation.ExecMode.BOTH.equals(execMode)) {
```
zentol: unnecessarily complex if statement.
Either use
```
if (BOTH.equals(execMode) {
...
} else {
...
}
```
or introduce a switch statement with a fast-failing default branch.

flink-annotations/src/main/java/org/apache/flink/annotation/docs/Documentation.java
```diff
@@ -59,6 +59,28 @@
 		int position() default Integer.MAX_VALUE;
 	}
 
+	/**
+	 * Annotation used on config option fields to include them in the "Table Options" section.
```
zentol: This doesn't seem accurate. There's no "Table Options" section, and this option doesn't control whether anything is included in any documentation, but only serves to introduce additional table-related meta data.

docs/ops/config.md
```diff
@@ -183,6 +183,12 @@ unless user define a `OptionsFactory` and set via `RocksDBStateBackend.setOption
 
 {% include generated/rocks_db_configurable_configuration.html %}
 
+### blink table planner
```
zentol: - this page is meant as a full reference; I'd assume these should also be placed on the API specific pages
- please follow the naming convention of other sectons, i.e. this should be `Blink Table Planner`
- I'm not sold on the placement, this area here is all about state (checkpointing, rocksdb, QS)

flink-annotations/src/main/java/org/apache/flink/annotation/docs/Documentation.java
```diff
@@ -59,6 +59,28 @@
...
+	 *
+	 * <p>The {@link TableOption#execMode()} argument indicate which exec mode the config works for,
+	 * for batch, streaming or both.
+	 *
+	 */
+	@Target(ElementType.FIELD)
+	@Retention(RetentionPolicy.RUNTIME)
+	@Internal
+	public @interface TableOption {
+
```
zentol: remove this line, following the existing code style in this class

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/api/PlannerConfigOptions.java
```diff
@@ -39,102 +41,134 @@
 							"(only count RexCall node, including leaves and interior nodes). Negative number to" +
 							" use the default threshold: double of number of nodes.");
 
+	@Documentation.TableOption(execMode = Documentation.ExecMode.BOTH)
```
zentol: since BOTH is the default you could omit the explicit execMode. Otherwise, please remove the default.

XuPingyong:
Thanks @zentol , I have fixed according to your comments.

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/api/PlannerConfigOptions.java
XuPingyong: Thanks, I have remove the default.

flink-annotations/src/main/java/org/apache/flink/annotation/docs/Documentation.java
XuPingyong: Fixed it.

docs/ops/config.md
XuPingyong: - Documents about blink will be introduced later, then config will also be placed together with the related API.
- Fixed it.
- I moved these to the end of this page.

flink-annotations/src/main/java/org/apache/flink/annotation/docs/Documentation.java
XuPingyong: Thanks, fixed it.

flink-docs/src/main/java/org/apache/flink/docs/configuration/ConfigOptionsDocGenerator.java
XuPingyong: Fixed it.

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/api/PlannerConfigOptions.java
XuPingyong: General users are not easy to understand and have no need to use these configs in most scenes.  Advanced users can tune performance of some special cases according to them.
So I do not want to place these professional configs to the documentation.

zentol: document the reason accordingly

XuPingyong: Added the reason.

XuPingyong:
> Since you're introducing some custom logic for differentiating between stream/batch, please include a preview screenshot of how the page looks like.

@zentol  I have put a preview screenshot in the description of this pr, is it OK?

zentol:
I can only comment on changes that are directly related to the generation of the configuration docs. Everything else is beyond my expertise; someone else has to review that.

wuchong:
Btw, I also suggest to split this pull request into 2 commits:
1. add documentation annotation for table
2. improve options in blink

XuPingyong:
Thanks @zentol @wuchong . I firstly split it to two prs.

PR #8940, opened 2019-07-01
[FLINK-13047][table] Fix the Optional.orElse() usage issue in Databas…

xuefuz:
…eCalciteSchema

[FLINK-13047][table] Fix the Optional.orElse() usage issue in DatabaseCalciteSchem

## Brief change log

  - Evaluate the expression is orElse() clause only if Optional object is empty.

## Verifying this change

This change is already covered by existing test.

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-table/flink-table-planner/src/main/java/org/apache/flink/table/catalog/DatabaseCalciteSchema.java
+++ b/flink-table/flink-table-planner/src/main/java/org/apache/flink/table/catalog/DatabaseCalciteSchema.java
@@ -110,7 +110,7 @@ class DatabaseCalciteSchema implements Schema {
 			if (tf instanceof TableSourceFactory) {
 				tableSource = ((TableSourceFactory) tf).createTableSource(tablePath, table);
 			} else {
-				throw new TableException(String.format("TableFactory provided by catalog %s must implement TableSourceFactory",
+				throw new TableException(String.format("Cannot query a sink-only table. TableFactory provided by catalog %s must implement TableSourceFactory",
 					catalog.getClass()));
 			}
 		} else {
--- a/flink-table/flink-table-planner/src/main/java/org/apache/flink/table/catalog/DatabaseCalciteSchema.java
+++ b/flink-table/flink-table-planner/src/main/java/org/apache/flink/table/catalog/DatabaseCalciteSchema.java
@@ -29,7 +29,6 @@ import org.apache.flink.table.plan.schema.TableSourceTable;
 import org.apache.flink.table.plan.stats.FlinkStatistic;
 import org.apache.flink.table.sources.StreamTableSource;
 import org.apache.flink.table.sources.TableSource;
-import org.apache.flink.types.Row;
 
 import org.apache.calcite.linq4j.tree.Expression;
 import org.apache.calcite.rel.type.RelProtoDataType;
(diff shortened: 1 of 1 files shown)


xuefuz:
cc: @bowenli86 @lirui-apache @zjuwangg

flink-table/flink-table-planner/src/main/java/org/apache/flink/table/catalog/DatabaseCalciteSchema.java
```diff
@@ -105,8 +105,10 @@ private Table convertConnectorTable(ConnectorCatalogTable<?, ?> table) {
 
 	private Table convertCatalogTable(ObjectPath tablePath, CatalogTable table) {
 		Optional<TableFactory> tableFactory = catalog.getTableFactory();
-		TableSource<Row> tableSource = tableFactory.map(tf -> ((TableSourceFactory) tf).createTableSource(tablePath, table))
-			.orElse(TableFactoryUtil.findAndCreateTableSource(table));
+		TableSource<Row> tableSource = tableFactory.map(tf -> ((TableSourceFactory) tf).createTableSource(tablePath, table)).get();
```
bowenli86: by comparing the logic here and `HiveTableFactory.createTableSource()`, I found there are duplicated calls on `TableFactoryUtil.findAndCreateTableSource(table)`. That made me wonder if any table factory implementation shouldn't call the generic table discovery service (a.k.a `TableFactoryUtil.findAndCreateTableSource`), and should just create tables that is specific to that table factory itself. 

Thus, the `HiveTableFactory.createTableSource` should just be as the following:

```
@Override
	public TableSource<Row> createTableSource(ObjectPath tablePath, CatalogTable table) {
		...

		if (!isGeneric) {
			return createInputFormatTableSource(tablePath, table);
		} else {
			return null;
		}
	}
```

xuefuz: Good point. However, that's not what I meant to do. HiveTableFactory is responsible to generate table source/sink for both generic and hive tables, so it needs to take care of both. In DatabaseCalciteSchema.convertCatalogTable(), it calls TableFactoryUtil.findAndCreateTableSource(table) only if the catalog doesn't return a table factory object. Thus, there is no duplication per se. However, I did realize that the logic here is confusion, especially with the java Optional object. I will refactor the code a bit to make it clearer.

flink-table/flink-table-planner/src/main/java/org/apache/flink/table/catalog/DatabaseCalciteSchema.java
```diff
@@ -104,9 +103,19 @@ private Table convertConnectorTable(ConnectorCatalogTable<?, ?> table) {
...
+		TableSource<?> tableSource;
 		Optional<TableFactory> tableFactory = catalog.getTableFactory();
-		TableSource<Row> tableSource = tableFactory.map(tf -> ((TableSourceFactory) tf).createTableSource(tablePath, table))
-			.orElse(TableFactoryUtil.findAndCreateTableSource(table));
+		if (tableFactory.isPresent()) {
+			TableFactory tf = tableFactory.get();
+			if (tf instanceof TableSourceFactory) {
+				tableSource = ((TableSourceFactory) tf).createTableSource(tablePath, table);
+			} else {
+				throw new TableException(String.format("TableFactory provided by catalog %s must implement TableSourceFactory",
```
bowenli86: minor: shall we use similar error message as that in `convertConnectorTable()`, something like "Cannot query a sink-only table. TableFactory provided by catalog %s must implement ....".

Asking this because when I first saw this exception, I got a bit on confused why it cannot be `TableSinkFactory`.

bowenli86:
since the build queue is so congested now and this PR's build may not finish probably till tomorrow, I built and test locally and made sure it passes.

merging

PR #8945, opened 2019-07-02
[FLINK-13045][table] Move Scala expression DSL to flink-table-api-scala

twalthr:

This PR move the Scala expression DSL to `flink-table-api-scala`.

Users of pure table programs should define there imports like:

```
import org.apache.flink.table.api._

TableEnvironment.create(...)
```

Users of the DataStream API should define their imports like:
```
import org.apache.flink.table.api._
import org.apache.flink.table.api.scala._

StreamTableEnvironment.create(...)
```

This PR did not split the package object `org.apache.flink.table.api.scala._` into two parts yet because we want to give users the chance to update their imports.

## Brief change log

- Updated all Scala implicits
- Move the implicits to their final destination

## Verifying this change

  - The serializers: no

## Documentation


twalthr:
Thanks for your thoughts. I think the purpose of this annotation is to mark code where people need to think twice before they change something. Unfortunately is not possible to annotate package objects. So I think having an annotation there does not harm.

PR #8946, opened 2019-07-02
[FLINK-13042][datastream] Make slot sharing configurable on StreamExecutionEnvironment

ifndef-SleePy:

* There is a requirement of Blink batch planner that providing a global setting that disabling slot sharing. To support that, will expose a `PublicEvolving` method on `StreamExecutionEnvironment` to globally disable slot sharing.

* Note that, this method might be removed if there is a better approach to satisfy Blink batch planner in the future.

* There is a discussion under https://github.com/apache/flink/pull/8931 at the moment
* **This PR should not go forward until there is an agreement reached**

## Brief change log

* Add a `PublicEvolving` method on `StreamExecutionEnvironment` to enable/disable slot sharing

## Verifying this change

* Added integration tests to verify slot sharing behavior

  - The serializers: no

## Documentation

  - Does this pull request introduce a new feature? **yes** maybe
  - If yes, how is the feature documented? **docs & JavaDocs**

### DIFF
--- a/docs/dev/stream/operators/index.md
+++ b/docs/dev/stream/operators/index.md
@@ -1118,6 +1118,16 @@ someStream.filter(...).slotSharingGroup("name");
         </p>
       </td>
     </tr>
+    <tr>
+      <td>Enable or disable slot sharing</td>
+      <td>
+        <p>Enable or disable slot sharing globally
+{% highlight java %}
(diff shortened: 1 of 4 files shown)


ifndef-SleePy:
We don't need this at the moment. So close it.

PR #8951, opened 2019-07-02
[FLINK-13062] Set ScheduleMode based on boundedness of streaming Pipeline

aljoscha:

The new Blink-based Table Runner needs "streaming pipelines" to be executed with `ScheduleMode.LAZY_FROM_SOURCES` if all sources are bounded. The current Blink code base uses a global flag for this and configures the `StreamGraphGenerator` accordingly.

We propose to add an `isBounded()` property to `Transformation` (formerly known as `StreamTransformation`). The property would only be explicitly settable on sources, other transformations inherit the property from their inputs. The `StreamGraphGenerator` must use `ScheduleMode.LAZY_FROM_SOURCES` if all sources are bounded, otherwise, it should use ScheduleMode.EAGER, as is the currently existing behaviour.

## Brief change log

* ccf6f96 refactors `DataStreamSource` as preparation
* 03aa30b adds the `isBounded()` property to `Transformation`, along with tests for the inheritance logic
* 277c947 adds a new *public* method on `StreamExecutionEnvironment` for adding bounded sources
* c37ffb9 changes `StreamGraphGenerator` to set the `ScheduleMode` based on the newly added boundedness property, along with tests

## Verifying this change

* new tests in 03aa30b for the `isBounded()` property on `Transformation`
* new tests in c37ffb9, i.e in `StreamGraphGeneratorTest`

  - The serializers: no

## Documentation

  - Does this pull request introduce a new feature? yes, there is a new public API method for adding bounded sources
  - If yes, how is the feature documented? documented in Javadocs, as for now this is more of an under-the-radar featre

### DIFF
--- a/flink-connectors/flink-connector-elasticsearch6/src/test/java/org/apache/flink/streaming/connectors/elasticsearch6/Elasticsearch6UpsertTableSinkFactoryTest.java
+++ b/flink-connectors/flink-connector-elasticsearch6/src/test/java/org/apache/flink/streaming/connectors/elasticsearch6/Elasticsearch6UpsertTableSinkFactoryTest.java
@@ -206,7 +206,7 @@ public class Elasticsearch6UpsertTableSinkFactoryTest extends ElasticsearchUpser
 	private static class TransformationMock extends Transformation<Tuple2<Boolean, Row>> {
 
 		public TransformationMock(String name, TypeInformation<Tuple2<Boolean, Row>> outputType, int parallelism) {
-			super(name, outputType, parallelism);
+			super(name, outputType, parallelism, false);
 		}
 
 		@Override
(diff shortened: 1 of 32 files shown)


aljoscha:
This is an alternative proposal for #8931 

cc @ifndef-SleePy

ifndef-SleePy:
Wow, It's amazing that you have already finished this proposal!

I have taken a glance at it. I got your point. And I believe it could make planner work for 1.9. Furthermore I need to discuss with the guys who are responsible for planner implementation to make sure it could satisfy their requirements for both 1.9 and the further. Since it's midnight in my timezone, I guess I can't give you feedback until tomorrow :(

aljoscha:
Sure, please take time to sleep! 😃

guoweiM:
Thanks @aljoscha . I think it is ok for BlinkPlanner to choose a proper schedule mode  in 1.9.
I have a little question about how planner set the configuration of runtime.  Sometime a planner might need directly configure the engine to schedule the Job in a specific way. For example there are bounded sources but the planner want use the third scheduling strategy.

My point is we might need a common way to let the planner configure how the engine is execute in the future.  What do you think ?

XuPingyong:
Thanks @aljoscha .  I think boundedness of streaming Pipeline is not related to the streaming/batch behavior. A bounded source(eg. Collection source) can run not only batch job, but also streaming job.

aljoscha:
Another alternative would be to set the `ScheduleMode` based on the `ExecutionMode`, which is already configured in the `ExecutionConfig`.

PR #8956, opened 2019-07-03
[FLINK-12991][python] Correct the implementation of Catalog.get_table_factory

dianfu:

*Currently the implementation of Catalog.get_table_factory returns a Java TableFactory and this is not friendly for Python users.  After taking a look at the implementation and it seems that this method will only be used internally at Java side. So I think it's not necessary to add it to the Python API. This PR fixes this issue by removing this method from the Python Catalog.*

## Brief change log

  - *Removes Catalog.get_table_factory*

## Verifying this change

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-python/pyflink/table/catalog.py
+++ b/flink-python/pyflink/table/catalog.py
@@ -54,15 +54,6 @@ class Catalog(object):
         """
         return self._j_catalog.getDefaultDatabase()
 
-    def get_table_factory(self):
-        """
-        Get an optional TableFactory instance that's responsible for generating source/sink for
-        tables stored in this catalog.
-
-        :return: An optional TableFactory instance.
-        """
-        return self._j_catalog.getTableFactory()
-
     def list_databases(self):
         """
         Get the names of all databases in this catalog.
--- a/flink-python/pyflink/table/tests/test_catalog_completeness.py
+++ b/flink-python/pyflink/table/tests/test_catalog_completeness.py
@@ -40,7 +40,7 @@ class CatalogAPICompletenessTests(PythonAPICompletenessTestCase, unittest.TestCa
     @classmethod
     def excluded_methods(cls):
         # open/close are not needed in Python API as they are used internally
-        return {'open', 'close'}
+        return {'open', 'close', 'getTableFactory'}
 
 
 class CatalogDatabaseAPICompletenessTests(PythonAPICompletenessTestCase, unittest.TestCase):


dianfu:
@bowenli86 Could you help to take a look at if this change makes sense to you?

bowenli86:
I'm wondering how you expect users to use catalog APIs in python - are users able to develop their own catalog implementations by implementing the python catalog interface? Or they can only develop their own ones in Java and then reference them in python?

dianfu:
@bowenli86 Thanks a lot for your comments. The Python catalog API is just a thin wrapper of the Java catalog API and it's not designed to let users implement a Python catalog. So for your questions, it's the latter case. Users should develop their own catalog in Java and reference them in Python.

dianfu:
@bowenli86 Any feedback? We need to fix this issue in 1.9 and the 1.9 release branch will be cut latter today. So appreciate if you can take a look at this PR ASAP.

bowenli86:
sounds good. 

For the PR specifically, I'm fine either way because `open, close, getTableFactory` are actually publicly available to Java/Scala Table API users. Hiding them from Python table api only solve the problem from the surface, not from its root.

zentol:
CI report for commit 59c446d3df87c0405515a4ab7360df3f21e46dd4: FAILURE [Travis](https://travis-ci.org/flink-ci/flink-ci/builds/554872065?utm_source=github_status&utm_medium=notification)

dianfu:
@bowenli86 Thanks a lot for your reply. 

> I'm fine either way because open, close, getTableFactory are actually publicly available to Java/Scala Table API users.
Regarding to `open, close, getTableFactory`, I agree with you that they are public in Java. However, I think they are public for Catalog developers, not for API users. I think it is also the same case for `TableFactory`. It's also public for connector/format/catalog developers, not for API users. I have not found any documentation about how to use `TableFactory` for API users. So the key point here is not only whether the interface is public or not, but also how we want users to use it.

> Hiding them from Python table api only solve the problem from the surface, not from its root.
According to the above, I don't agree with you on this point. We should remove this interface in Python for now as I think this interface is not for API users. Anyway, we can add it back in the future if we found that it's useful and should be supported. However, removing an interface isn't easy to do.

@zentol Thanks a lot for the remind. The failure is not related to this PR and I have rebased the PR to re-trigger the tests.

bowenli86:
it's fine to do this in python temporarily. Fundamentally, we need to think about maybe redefining the `Catalog` interface, as we don't want java/scala users access the three methods too

dianfu:
@bowenli86 Thanks a lot for the clarify. Make much sense to me. :)

PR #8966, opened 2019-07-03
[FLINK-13074][table-planner-blink] PartitionableTableSink didn't work for  flink&blink planner

danny0405:

Add bridge logic of PartitionableTableSink for flink&blink planner, now we only add dynamic partitions for blink-planner and static partitions for flink-planner; after merging of blink and flink planner, these functions would also be merged.

## Brief change log

  - Add PartitionableTableSink bridge logic

## Verifying this change

There are no tests for this patch now, would add test cases when both blink/flink planner upgrade Calcite version to 1.20.0

  - The serializers: no

## Documentation

### DIFF
--- a/flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/catalog/DatabaseCalciteSchema.java
+++ b/flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/catalog/DatabaseCalciteSchema.java
@@ -28,6 +28,7 @@ import org.apache.flink.table.factories.TableSourceFactory;
 import org.apache.flink.table.operations.DataStreamQueryOperation;
 import org.apache.flink.table.operations.QueryOperation;
 import org.apache.flink.table.operations.RichTableSourceQueryOperation;
+import org.apache.flink.table.plan.schema.TableSinkTable;
 import org.apache.flink.table.plan.schema.TableSourceTable;
 import org.apache.flink.table.plan.stats.FlinkStatistic;
 import org.apache.flink.table.sources.LookupableTableSource;
@@ -108,22 +109,36 @@ class DatabaseCalciteSchema extends FlinkSchema {
 	}
 
 	private Table convertConnectorTable(ConnectorCatalogTable<?, ?> table) {
-		return table.getTableSource()
-				.map(tableSource -> {
-					if (!(tableSource instanceof StreamTableSource ||
-							tableSource instanceof LookupableTableSource)) {
-						throw new TableException(
-								"Only StreamTableSource and LookupableTableSource can be used in Blink planner.");
-					}
-					if (!isStreamingMode && tableSource instanceof StreamTableSource &&
-							!((StreamTableSource<?>) tableSource).isBounded()) {
-						throw new TableException("Only bounded StreamTableSource can be used in batch mode.");
-					}
-					return new TableSourceTable<>(
-							tableSource,
-							isStreamingMode,
-							FlinkStatistic.UNKNOWN());
-				}).orElseThrow(() -> new TableException("Cannot query a sink only table."));
+		Optional<TableSourceTable> tableSourceTable = table.getTableSource()
+			.map(tableSource -> {
+				if (!(tableSource instanceof StreamTableSource ||
+					tableSource instanceof LookupableTableSource)) {
+					throw new TableException(
+						"Only StreamTableSource and LookupableTableSource can be used in Blink planner.");
+				}
+				if (!isStreamingMode && tableSource instanceof StreamTableSource &&
+					!((StreamTableSource<?>) tableSource).isBounded()) {
+					throw new TableException("Only bounded StreamTableSource can be used in batch mode.");
+				}
+				return new TableSourceTable<>(
+					tableSource,
+					isStreamingMode,
+					FlinkStatistic.UNKNOWN());
+			});
+		if (tableSourceTable.isPresent()) {
+			return tableSourceTable.get();
+		} else {
+			Optional<TableSinkTable> tableSinkTable = table.getTableSink()
+				.map(tableSink -> new TableSinkTable<>(
+					tableSink,
+					FlinkStatistic.UNKNOWN()));
+			if (tableSinkTable.isPresent()) {
+				return tableSinkTable.get();
+			} else {
+				throw new TableException("Cannot convert a connector table " +
+					"without either source or sink.");
+			}
+		}
 	}
 
 	private Table convertCatalogTable(ObjectPath tablePath, CatalogTable table) {
--- a/flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/calcite/PreValidateReWriter.scala
+++ b/flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/calcite/PreValidateReWriter.scala
@@ -99,15 +99,12 @@ object PreValidateReWriter {
       case t: RelOptTable => t
       case _ => null
     }
-    def tester(idx: Integer): Boolean = {
-      !assignedFields.contains(idx)
-    }
     for (node <- partitions.getList) {
       val sqlProperty = node.asInstanceOf[SqlProperty]
       val id = sqlProperty.getKey
       val targetField = SqlValidatorUtil.getTargetField(targetRowType,
         typeFactory, id, calciteCatalogReader, relOptTable)
-      validateField(tester, id, targetField)
+      validateField(idx => !assignedFields.contains(idx), id, targetField)
       val value = sqlProperty.getValue.asInstanceOf[SqlLiteral]
       assignedFields.put(targetField.getIndex,
         maybeCast(value, value.createSqlType(typeFactory), targetField.getType, typeFactory))
--- a/flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/calcite/PreValidateReWriter.scala
+++ b/flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/calcite/PreValidateReWriter.scala
@@ -41,8 +41,8 @@ import scala.collection.JavaConversions._
 /** Implements [[org.apache.calcite.sql.util.SqlVisitor]]
   * interface to do some rewrite work before sql node validation. */
 class PreValidateReWriter(
-  val catalogReader: CalciteCatalogReader,
-  val typeFactory: RelDataTypeFactory) extends SqlBasicVisitor[Unit] {
+    val catalogReader: CalciteCatalogReader,
+    val typeFactory: RelDataTypeFactory) extends SqlBasicVisitor[Unit] {
   override def visit(call: SqlCall): Unit = {
     call match {
       case r: RichSqlInsert if r.getStaticPartitions.nonEmpty
@@ -80,10 +80,10 @@ object PreValidateReWriter {
     * @param partitions           Static partition statements
     */
   def appendPartitionProjects(sqlInsert: RichSqlInsert,
-    calciteCatalogReader: CalciteCatalogReader,
-    typeFactory: RelDataTypeFactory,
-    select: SqlSelect,
-    partitions: SqlNodeList): Unit = {
+      calciteCatalogReader: CalciteCatalogReader,
+      typeFactory: RelDataTypeFactory,
+      select: SqlSelect,
+      partitions: SqlNodeList): Unit = {
     val names = sqlInsert.getTargetTable.asInstanceOf[SqlIdentifier].names
     val table = calciteCatalogReader.getTable(names)
     if (table == null) {
@@ -99,12 +99,15 @@ object PreValidateReWriter {
       case t: RelOptTable => t
       case _ => null
     }
+    def tester(idx: Integer): Boolean = {
+      !assignedFields.contains(idx)
+    }
     for (node <- partitions.getList) {
       val sqlProperty = node.asInstanceOf[SqlProperty]
       val id = sqlProperty.getKey
       val targetField = SqlValidatorUtil.getTargetField(targetRowType,
         typeFactory, id, calciteCatalogReader, relOptTable)
-      validateField(assignedFields.containsValue, id, targetField)
+      validateField(tester, id, targetField)
       val value = sqlProperty.getValue.asInstanceOf[SqlLiteral]
       assignedFields.put(targetField.getIndex,
         maybeCast(value, value.createSqlType(typeFactory), targetField.getType, typeFactory))
@@ -142,10 +145,10 @@ object PreValidateReWriter {
     * @return Rowtype
     */
   private def createTargetRowType(
-    typeFactory: RelDataTypeFactory,
-    catalogReader: CalciteCatalogReader,
-    table: SqlValidatorTable,
-    targetColumnList: SqlNodeList): RelDataType = {
+      typeFactory: RelDataTypeFactory,
+      catalogReader: CalciteCatalogReader,
+      table: SqlValidatorTable,
+      targetColumnList: SqlNodeList): RelDataType = {
     val baseRowType = table.getRowType
     if (targetColumnList == null) return baseRowType
     val fields = new util.ArrayList[util.Map.Entry[String, RelDataType]]
@@ -166,8 +169,8 @@ object PreValidateReWriter {
 
   /** Check whether the field is valid. **/
   private def validateField(tester: Function[Integer, Boolean],
-    id: SqlIdentifier,
-    targetField: RelDataTypeField): Unit = {
+      id: SqlIdentifier,
+      targetField: RelDataTypeField): Unit = {
     if (targetField == null) {
       throw newValidationError(id, RESOURCE.unknownTargetColumn(id.toString))
     }
--- a/flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/PlannerBase.scala
+++ b/flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/PlannerBase.scala
@@ -21,6 +21,7 @@ package org.apache.flink.table.planner
 import org.apache.flink.annotation.VisibleForTesting
 import org.apache.flink.api.dag.Transformation
 import org.apache.flink.configuration.Configuration
+import org.apache.flink.sql.parser.dml.RichSqlInsert
 import org.apache.flink.streaming.api.environment.StreamExecutionEnvironment
 import org.apache.flink.table.api.{TableConfig, TableEnvironment, TableException}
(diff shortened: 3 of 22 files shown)


flink-table/flink-table-planner-blink/src/test/scala/org/apache/flink/table/runtime/utils/BatchTestBase.scala
```diff
@@ -65,6 +66,12 @@ class BatchTestBase extends BatchAbstractTestBase {
   val LINE_COL_TWICE_PATTERN: Pattern = Pattern.compile("(?s)From line ([0-9]+),"
     + " column ([0-9]+) to line ([0-9]+), column ([0-9]+): (.*)")
 
+  /**
+    * Subclass should overwrite this method if we want to overwrite configuration during
+    * sql parse to sql to rel conversion phrase.
+    */
+  protected def getTableConfig: TableConfig = new TableConfig
```
wuchong: It is confused we expose the create a new TableConfig. Because we use the `tEnv.getConfig` to set a lot of configurations in `before()`. I think if we want to set sql parser configurations, we can also override `before()` method and add sql parse configs in `conf`.

flink-table/flink-table-planner/src/main/scala/org/apache/flink/table/planner/StreamPlanner.scala
```diff
@@ -101,17 +104,19 @@ class StreamPlanner(
...
         }
         // get name of sink table
         val targetTablePath = insert.getTargetTable.asInstanceOf[SqlIdentifier].names
+        val staticPartitions = insert.getStaticPartitionKVs
 
         List(new CatalogSinkModifyOperation(targetTablePath,
           SqlToOperationConverter.convert(planner,
-            insert.getSource).asInstanceOf[PlannerQueryOperation])
+            insert.getSource).asInstanceOf[PlannerQueryOperation],
+          staticPartitions)
```
wuchong: Can we keep sync with blink planner for this?

flink-table/flink-table-planner-blink/src/test/scala/org/apache/flink/table/runtime/batch/sql/PartitionableSinkITCase.scala
```diff
@@ -0,0 +1,278 @@
...
+import java.util.concurrent.LinkedBlockingQueue
+import java.util.{LinkedList => JLinkedList, List => JList, Map => JMap}
+
+import scala.collection.JavaConversions._
+import scala.collection.Seq
+
+/**
+  * Test cases for [[org.apache.flink.table.sinks.PartitionableTableSink]].
+  */
+class PartitionableSinkITCase extends BatchTestBase {
```
wuchong: Why don't we have `PartitionableSinkITCase` in flink planner?

flink-table/flink-table-planner-blink/src/test/scala/org/apache/flink/table/runtime/batch/sql/PartitionableSinkITCase.scala
```diff
@@ -0,0 +1,278 @@
...
+    tEnv.execute("testJob")
+    // this sink should have been set up with static partitions
+    assertEquals(testSink.getStaticPartitions.toMap, Map("a" -> "1"))
+    val resultSet = List(RESULT1, RESULT2, RESULT3)
+    resultSet.foreach(l => assertSortedByFirstNField(l, 2))
+    assertEquals(resultSet.map(l => collectDistinctGroupCount(l, 2)).sum, 4)
+  }
+
+  @Test(expected = classOf[TableException])
+  def testDynamicPartitionInFrontOfStaticPartition(): Unit = {
```
wuchong: please assert the exception message as well.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/calcite/PreValidateReWriter.scala
```diff
@@ -99,12 +99,15 @@ object PreValidateReWriter {
...
+    def tester(idx: Integer): Boolean = {
+      !assignedFields.contains(idx)
+    }
     for (node <- partitions.getList) {
       val sqlProperty = node.asInstanceOf[SqlProperty]
       val id = sqlProperty.getKey
       val targetField = SqlValidatorUtil.getTargetField(targetRowType,
         typeFactory, id, calciteCatalogReader, relOptTable)
-      validateField(assignedFields.containsValue, id, targetField)
+      validateField(tester, id, targetField)
```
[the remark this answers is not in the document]
danny0405: Yeah, we could.

flink-table/flink-table-planner-blink/src/test/scala/org/apache/flink/table/runtime/batch/sql/PartitionableSinkITCase.scala
danny0405: Sorry i think this is overchecked

flink-table/flink-table-planner-blink/src/test/scala/org/apache/flink/table/runtime/batch/sql/PartitionableSinkITCase.scala
danny0405: Added

flink-table/flink-table-planner-blink/src/test/scala/org/apache/flink/table/runtime/batch/sql/PartitionableSinkITCase.scala
danny0405: Added

flink-table/flink-table-planner/src/main/scala/org/apache/flink/table/sinks/TableSinkUtils.scala
```diff
@@ -62,5 +68,18 @@ object TableSinkUtils {
           s"Query result schema: $srcSchema\n" +
           s"TableSink schema:    $sinkSchema")
     }
+    // check partitions are valid
+    sink match {
```
wuchong: Please check `partitionableTableSink.getPartitionFieldNames` is not empty first if `sinkOperation.getStaticPartitions` is not empty. Regarding to the javadoc, if `sink.getPartitionFieldNames` is empty, it is not a partitioned sink.

flink-table/flink-table-planner/src/main/scala/org/apache/flink/table/api/internal/BatchTableEnvImpl.scala
```diff
@@ -133,6 +135,26 @@ abstract class BatchTableEnvImpl(
...
   }
 
+  /**
+    * Key by the partition fields if the sink is a [[PartitionableTableSink]].
+    * @param sink    the table sink
+    * @param dataSet the data set
+    * @tparam R      the data set record type
+    * @return a data set that maybe keyed by.
+    */
+  private def mayKeyByPartitionFields[R](
```
wuchong: What about `applyShuffleByPartitionFieldsIfNeeded` ?

flink-table/flink-table-planner/src/main/scala/org/apache/flink/table/planner/StreamPlanner.scala
```diff
@@ -155,7 +153,13 @@ class StreamPlanner(
       case catalogSink: CatalogSinkModifyOperation =>
         getTableSink(catalogSink.getTablePath)
           .map(sink => {
-            TableSinkUtils.validateSink(catalogSink.getChild, catalogSink.getTablePath, sink)
+            TableSinkUtils.validateSink(catalogSink.getStaticPartitions,
+              catalogSink.getChild, catalogSink.getTablePath, sink)
+            sink match {
+              case partitionableSink: PartitionableTableSink =>
```
wuchong: Only set static partitions when sink has partitioned fields, i.e. is a partitioned sink. Currently, we set a partition map always which may have unexpected effects.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/PlannerBase.scala
```diff
@@ -173,7 +165,12 @@ abstract class PlannerBase(
       case catalogSink: CatalogSinkModifyOperation =>
         val input = getRelBuilder.queryOperation(modifyOperation.getChild).build()
         getTableSink(catalogSink.getTablePath).map(sink => {
-          TableSinkUtils.validateSink(catalogSink.getChild, catalogSink.getTablePath, sink)
+          TableSinkUtils.validateSink(catalogSink, catalogSink.getTablePath, sink)
+          sink match {
+            case partitionableSink: PartitionableTableSink =>
```
wuchong: Only set static partitions when sink has partitioned fields, i.e. is a partitioned sink. Currently, we set a partition map always which may have unexpected effects.

flink-table/flink-table-planner/src/main/scala/org/apache/flink/table/api/internal/TableEnvImpl.scala
```diff
@@ -532,7 +542,13 @@ abstract class TableEnvImpl(
 
       case Some(tableSink) =>
         // validate schema of source table and table sink
-        TableSinkUtils.validateSink(table.getQueryOperation, sinkTablePath.asJava, tableSink)
+        TableSinkUtils.validateSink(insertOptions.staticPartitions, table.getQueryOperation,
+          sinkTablePath.asJava, tableSink)
+        tableSink match {
+          case partitionableSink: PartitionableTableSink =>
```
wuchong: Only set static partitions when sink has partitioned fields, i.e. is a partitioned sink. Currently, we set a partition map always which may have unexpected effects.

flink-table/flink-table-planner/src/main/scala/org/apache/flink/table/api/internal/TableEnvImpl.scala
```diff
@@ -440,10 +441,12 @@ abstract class TableEnvImpl(
     // parse the sql query
     val parsed = planner.parse(stmt)
     parsed match {
-      case insert: SqlInsert =>
-        // validate the SQL query
-        val query = insert.getSource
-        val validatedQuery = planner.validate(query)
+      case insert: RichSqlInsert =>
+        // validate the insert
+        val validatedInsert = planner.validate(insert).asInstanceOf[RichSqlInsert]
```
wuchong: I don't think we need cast to `RichSqlInsert` here.

flink-table/flink-table-planner/src/main/scala/org/apache/flink/table/api/internal/TableEnvImpl.scala
```diff
@@ -453,7 +456,8 @@ abstract class TableEnvImpl(
         val targetTablePath = insert.getTargetTable.asInstanceOf[SqlIdentifier].names
 
         // insert query result into sink table
-        insertInto(queryResult, targetTablePath.asScala:_*)
+        insertInto(queryResult, InsertOptions.apply(insert.getStaticPartitionKVs),
```
wuchong: `insertOptions(insert.getStaticPartitionKVs)`

flink-table/flink-table-planner/src/main/scala/org/apache/flink/table/planner/StreamPlanner.scala
```diff
@@ -355,6 +373,26 @@ class StreamPlanner(
...
   }
 
+  /**
+    * Key by the partition fields if the sink is a [[PartitionableTableSink]].
+    * @param sink       the table sink
+    * @param dataStream the data stream
+    * @tparam R         the data stream record type
+    * @return a data stream that maybe keyed by.
+    */
+  private def mayKeyByPartitionFields[R](
```
wuchong: What about `applyShuffleByPartitionFieldsIfNeeded` ?

flink-table/flink-table-planner-blink/src/test/scala/org/apache/flink/table/runtime/utils/BatchTestBase.scala
wuchong: Can we add a TODO with the JIRA ID above this?  And tag Critical and 1.9 for the JIRA.

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/catalog/DatabaseCalciteSchema.java
```diff
@@ -108,22 +109,36 @@ public Table getTable(String tableName) {
...
+		} else {
+			Optional<TableSinkTable> tableSinkTable = table.getTableSink()
+				.map(tableSink -> new TableSinkTable<>(
+					tableSink,
+					FlinkStatistic.UNKNOWN()));
+			if (tableSinkTable.isPresent()) {
+				return tableSinkTable.get();
+			} else {
+				throw new TableException("Cannot convert a connector table " +
+					"without either source or sink.");
```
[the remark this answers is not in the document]
danny0405: This actually complains the compile error, because the table source and sink table have different java type.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/PlannerBase.scala
danny0405: The `PartitionableTableSink` is the only trait that `Flink` recognize as a partiton sink, if the partitioned fields is empty, the static partitions are always empty(we have checked this in sql node parse). So what exactly is the unexpected effects ?

BTW, i don't think the two methods `#getPartitionFieldNames` and `#setStaticPartition` should affect each other.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/sinks/TableSinkUtils.scala
```diff
@@ -67,5 +71,19 @@ object TableSinkUtils {
           s"Query result schema: $srcSchema\n" +
           s"TableSink schema:    $sinkSchema")
     }
+    // check partitions are valid
+    sink match {
```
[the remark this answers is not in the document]
danny0405: We already check that in the when we generate the sql node, i don't think we should re-check it here.

flink-table/flink-table-planner-blink/src/test/scala/org/apache/flink/table/runtime/batch/sql/PartitionableSinkITCase.scala
```diff
@@ -0,0 +1,300 @@
...
+  }
+}
+
+object PartitionableSinkITCase {
+  val RESULT1 = new JLinkedList[String]()
+  val RESULT2 = new JLinkedList[String]()
+  val RESULT3 = new JLinkedList[String]()
+  val RESULT_QUEUE: JList[JLinkedList[String]] = new JArrayList[JLinkedList[String]]()
+
+  def init(): Unit = {
```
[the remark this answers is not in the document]
danny0405: Ignored

flink-table/flink-table-planner/src/main/scala/org/apache/flink/table/api/internal/TableEnvImpl.scala
danny0405: It is definitely a `RichSqlInsert ` now, i choose the right type as it is.

flink-table/flink-table-planner/src/main/scala/org/apache/flink/table/api/internal/BatchTableEnvImpl.scala
danny0405: Renamed to shuffleByPartitionFieldsIfNeeded.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/sinks/TableSinkUtils.scala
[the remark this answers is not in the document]
wuchong: I don't think so. We can't guarantee sql parser is the only entry for insert partition. Maybe we will support this on Table API in the future. And it doesn't hurt anything to do a validation check when inserting. 

Btw, where is the check when we generate the sql node?

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/catalog/DatabaseCalciteSchema.java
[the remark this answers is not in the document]
wuchong: What compile error do you mean? 
I mean replace the code from L131-L140.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/PlannerBase.scala
wuchong: From user's point of view, if it is not a partitioned sink (`getPartitionFieldNames()` returns empty), the `setStaticPartition` should never be called. However, currently, `setStaticPartition` is always be called (may pass in an empty partition map).

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/catalog/DatabaseCalciteSchema.java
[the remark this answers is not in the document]
danny0405: Yeah, because this code branch is a else branch of the above if which has a java type tableSourceTable.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/PlannerBase.scala
danny0405: So what's the problem here ? Isn't the empty static partitions the right thing here ?

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/sinks/TableSinkUtils.scala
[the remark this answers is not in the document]
danny0405: The `PreValidateReWriter`. I don't know how the `TableAPI` uses this yet, but at least the one who implement this function should make sure this check by themselves.

flink-table/flink-table-planner-blink/src/test/scala/org/apache/flink/table/runtime/utils/BatchTestBase.scala
danny0405: Sure, i have fire a JIRA https://issues.apache.org/jira/browse/FLINK-13338: but without any `TODO` here.

wuchong:
Travis result: https://travis-ci.org/wuchong/flink/builds/562528710

Merging

PR #8922, opened 2019-06-28
[FLINK-12876][runtime] Add an adapter of region failover NG for legacy scheduler

GJL:
## What is the purpose of the change

*Based on #8851*

## Brief change log

  - *See #8851*

## Verifying this change

This change is already covered by existing tests, such as *RegionFailoverITCase*.

This change added tests and can be verified as follows:

  - *Added unit tests for `AdaptedRestartPipelinedRegionStrategyNG`*

## Does this pull request potentially affect one of the following parts:

  - Dependencies (does it add or upgrade a dependency): (yes / **no**)
  - The public API, i.e., is any changed class annotated with `@Public(Evolving)`: (yes / **no**)
  - The serializers: (yes / **no** / don't know)
  - The runtime per-record code paths (performance sensitive): (yes / **no** / don't know)
  - Anything that affects deployment or recovery: JobManager (and its components), Checkpointing, Yarn/Mesos, ZooKeeper: (**yes** / no / don't know)
  - The S3 file system connector: (yes / **no** / don't know)

## Documentation

  - Does this pull request introduce a new feature? (**yes** / no)
  - If yes, how is the feature documented? (not applicable / docs / JavaDocs / **not documented**)

PR #8971, opened 2019-07-03
[FLINK-13019][coordination][tests] IT test for fine-grained recovery

azagrebin:

The simplest IT test is to create a batch job with a sequence of non-parallel mappers.
The each mapper writes to a blocking partition which means the next mapper starts when the previous is done.
The mappers are not chained into one Task which makes them separate failover regions.

## Brief change log

add `BatchFineGrainedRecoveryITCase`

## Verifying this change

run `BatchFineGrainedRecoveryITCase`

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-tests/src/test/java/org/apache/flink/test/recovery/BatchFineGrainedRecoveryITCase.java
+++ b/flink-tests/src/test/java/org/apache/flink/test/recovery/BatchFineGrainedRecoveryITCase.java
@@ -58,6 +58,11 @@ import static org.junit.Assert.assertThat;
  * The test job is a sequence of non-parallel mappers. Each mapper writes to a blocking partition which means
  * the next mapper starts when the previous is done. The mappers are not chained into one task which makes them
  * separate fail-over regions.
+ *
+ * <p>The test verifies that fine-grained recovery works by randomly incuding failures in any of the mappers.
+ * Since all mappers are connected via blocking partitions, which should be re-used on failure, and the consumer
+ * of the mapper wasn't deployed yet, as the consumed partition was not fully produced yet, only the failed mapper
+ * should actually restart.
  */
 public class BatchFineGrainedRecoveryITCase extends TestLogger {
 	private static final int EMITTED_RECORD_NUMBER = 1000;
@@ -232,11 +237,10 @@ public class BatchFineGrainedRecoveryITCase extends TestLogger {
 	private enum StaticFailureCounter {
 		;
 
-		private static final int maxFailureNumber = MAX_FAILURE_NUMBER;
 		private static final AtomicInteger failureNumber = new AtomicInteger(0);
 
 		private static boolean failOrNot() {
-			return failureNumber.incrementAndGet() < maxFailureNumber;
+			return failureNumber.incrementAndGet() < MAX_FAILURE_NUMBER;
 		}
 
 		private static void reset() {
--- a/flink-tests/src/test/java/org/apache/flink/test/recovery/BatchFineGrainedRecoveryITCase.java
+++ b/flink-tests/src/test/java/org/apache/flink/test/recovery/BatchFineGrainedRecoveryITCase.java
@@ -1,0 +1,246 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.test.recovery;
+
+import org.apache.flink.api.common.ExecutionMode;
+import org.apache.flink.api.common.functions.RichMapPartitionFunction;
+import org.apache.flink.api.common.restartstrategy.RestartStrategies;
+import org.apache.flink.api.common.time.Time;
+import org.apache.flink.api.java.DataSet;
+import org.apache.flink.api.java.ExecutionEnvironment;
+import org.apache.flink.configuration.Configuration;
+import org.apache.flink.configuration.JobManagerOptions;
+import org.apache.flink.runtime.minicluster.TestingMiniCluster;
+import org.apache.flink.runtime.minicluster.TestingMiniClusterConfiguration.Builder;
+import org.apache.flink.test.util.TestEnvironment;
+import org.apache.flink.util.Collector;
+import org.apache.flink.util.FlinkRuntimeException;
+import org.apache.flink.util.TestLogger;
+
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.Random;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.stream.Collectors;
+import java.util.stream.LongStream;
+
+import static org.apache.flink.configuration.JobManagerOptions.FORCE_PARTITION_RELEASE_ON_CONSUMPTION;
+import static org.apache.flink.runtime.executiongraph.failover.FailoverStrategyLoader.PIPELINED_REGION_RESTART_STRATEGY_NAME;
+import static org.hamcrest.CoreMatchers.is;
+import static org.junit.Assert.assertThat;
+
+/**
+ * IT case for fine-grained recovery of batch jobs.
+ *
+ * <p>The test activates the region fail-over strategy to restart only failed producers.
+ * The test job is a sequence of non-parallel mappers. Each mapper writes to a blocking partition which means
+ * the next mapper starts when the previous is done. The mappers are not chained into one task which makes them
+ * separate fail-over regions.
+ */
+public class BatchFineGrainedRecoveryITCase extends TestLogger {
+	private static final int EMITTED_RECORD_NUMBER = 1000;
+	private static final int MAX_FAILURE_NUMBER = 10;
+	private static final int MAP_NUMBER = 3;
+	private static final List<Long> EXPECTED_JOB_OUTPUT =
+		LongStream.range(MAP_NUMBER, EMITTED_RECORD_NUMBER + MAP_NUMBER).boxed().collect(Collectors.toList());
+
+	private TestingMiniCluster miniCluster;
+
+	@Before
+	public void setup() throws Exception {
+		Configuration configuration = new Configuration();
+		configuration.setBoolean(FORCE_PARTITION_RELEASE_ON_CONSUMPTION, false);
+		configuration.setString(JobManagerOptions.EXECUTION_FAILOVER_STRATEGY, PIPELINED_REGION_RESTART_STRATEGY_NAME);
+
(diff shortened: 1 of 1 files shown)


azagrebin:
the PR is based on #8922 atm

azagrebin:
@flinkbot attention @zentol

flink-tests/src/test/java/org/apache/flink/test/recovery/BatchFineGrainedRecoveryITCase.java
```diff
@@ -0,0 +1,317 @@
...
+import java.util.Random;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.stream.IntStream;
+
+import static org.apache.flink.configuration.JobManagerOptions.FORCE_PARTITION_RELEASE_ON_CONSUMPTION;
+import static org.apache.flink.runtime.executiongraph.failover.FailoverStrategyLoader.PIPELINED_REGION_RESTART_STRATEGY_NAME;
+import static org.hamcrest.CoreMatchers.is;
+import static org.junit.Assert.assertThat;
+
+public class BatchFineGrainedRecoveryITCase extends TestLogger {
```
zentol: missing javadoc

flink-tests/src/test/java/org/apache/flink/test/recovery/BatchFineGrainedRecoveryITCase.java
```diff
@@ -0,0 +1,317 @@
...
+	}
+
+	private static <T> DataSet<T> disableChaining(
+			DataSet<T> input,
+			@SuppressWarnings("SameParameterValue") Class<T> cl) {
+		return input.mapPartition((MapPartitionFunction<T, T>) (ns, out) -> ns.forEach(out::collect)).returns(cl);
+	}
+
+	private static void verifyResult(JobExecutionResult result) {
+		Collection<Integer> accumulatorResult = result.getAccumulatorResult(CHECK_ACC_NAME);
```
zentol: you could simply this by using `DataSet#collect`. You wouldn't need a separate outputformat class, but the behavior should be the same.

flink-tests/src/test/java/org/apache/flink/test/recovery/BatchFineGrainedRecoveryITCase.java
```diff
@@ -0,0 +1,317 @@
...
+		ExecutionEnvironment env = createExecutionEnvironment();
+
+		StaticFailureCounter.reset();
+		StaticMapFailureTracker.reset();
+
+		FailureStrategy failureStrategy = new RandomExceptionFailureStrategy(1, EMITTED_RECORD_NUMBER);
+
+		DataSet<Integer> input = env.createInput(FixedRecordNumberInputFormat.emitFixedRecordNumber());
+		for (int i = 0; i < MAP_NUMBER; i++) {
+			input = input.flatMap(new TestMapper(StaticMapFailureTracker.addNewMap(), failureStrategy));
```
zentol: why are you not implementing the TestMapper as a MapPartition directly? It shouldn't change the semantics of the test, but you wouldn't need the dedicated `disableChaining` logic.

flink-tests/src/test/java/org/apache/flink/test/recovery/BatchFineGrainedRecoveryITCase.java
```diff
@@ -0,0 +1,317 @@
...
+
+	private static ExecutionEnvironment createExecutionEnvironment() {
+		@SuppressWarnings("StaticVariableUsedBeforeInitialization")
+		ExecutionEnvironment env = new TestEnvironment(miniCluster, 1, true);
+		env.setRestartStrategy(RestartStrategies.fixedDelayRestart(MAX_FAILURE_NUMBER, Time.milliseconds(10)));
+		env.getConfig().setExecutionMode(ExecutionMode.BATCH_FORCED);
+		return env;
+	}
+
+	private static <T> DataSet<T> disableChaining(
```
zentol: given that this is only used with a `DataSet<Integer>` as an input, the generics here could be simplified.

flink-tests/src/test/java/org/apache/flink/test/recovery/BatchFineGrainedRecoveryITCase.java
```diff
@@ -0,0 +1,317 @@
...
+	private enum StaticMapFailureTracker {
+		;
+
+		private static final List<AtomicInteger> mapRestarts = new ArrayList<>(10);
+		private static final List<AtomicInteger> expectedMapRestarts = new ArrayList<>(10);
+		private static final List<AtomicInteger> mapFailures = new ArrayList<>(10);
+
+		private static void reset() {
+			mapRestarts.clear();
+			expectedMapRestarts.clear();
```
zentol: should also reset `mapFailures`

flink-tests/src/test/java/org/apache/flink/test/recovery/BatchFineGrainedRecoveryITCase.java
```diff
@@ -0,0 +1,317 @@
...
+	@Test
+	public void testProgram() throws Exception {
+		ExecutionEnvironment env = createExecutionEnvironment();
+
+		StaticFailureCounter.reset();
+		StaticMapFailureTracker.reset();
+
+		FailureStrategy failureStrategy = new RandomExceptionFailureStrategy(1, EMITTED_RECORD_NUMBER);
+
+		DataSet<Integer> input = env.createInput(FixedRecordNumberInputFormat.emitFixedRecordNumber());
```
zentol: could be simplified by using `env#fromCollection`/`env#fromElements` or even `env#generateSequence`

flink-tests/src/test/java/org/apache/flink/test/recovery/BatchFineGrainedRecoveryITCase.java
```diff
@@ -0,0 +1,317 @@
...
+	private static class RandomExceptionFailureStrategy implements FailureStrategy {
+		private static final long serialVersionUID = 1L;
+
+		private final CoinToss coin;
+
+		private RandomExceptionFailureStrategy(int probFraction, int probBase) {
+			this(new CoinToss(probFraction, probBase));
+		}
+
+		private RandomExceptionFailureStrategy(CoinToss coin) {
```
zentol: This constructor seems redundant; only the other constructor is actually used.

flink-tests/src/test/java/org/apache/flink/test/recovery/BatchFineGrainedRecoveryITCase.java
```diff
@@ -0,0 +1,252 @@
...
+				StaticMapFailureTracker.mapFailure(trackingIndex);
+				throw t;
+			}
+		}
+	}
+
+	private enum StaticFailureCounter {
+		;
+
+		private static int maxFailureNumber;
```
zentol: shouldn't this be final?

flink-tests/src/test/java/org/apache/flink/test/recovery/BatchFineGrainedRecoveryITCase.java
```diff
@@ -0,0 +1,252 @@
...
+		private static final AtomicInteger failureNumber = new AtomicInteger(0);
+
+		private static boolean failOrNot() {
+			return failureNumber.get() < maxFailureNumber &&
```
zentol: the latter condition should be sufficient

flink-tests/src/test/java/org/apache/flink/test/recovery/BatchFineGrainedRecoveryITCase.java
```diff
@@ -0,0 +1,252 @@
...
+			mapRestarts.get(index).incrementAndGet();
+		}
+
+		private static void mapFailure(int index) {
+			mapFailures.get(index).incrementAndGet();
+			expectedMapRestarts.get(index).incrementAndGet();
+		}
+
+		private static void verify() {
+			assertThat(collect(mapRestarts), is(collect(expectedMapRestarts)));
```
zentol: Could we not compare the lists of atomic integers directly?

flink-tests/src/test/java/org/apache/flink/test/recovery/BatchFineGrainedRecoveryITCase.java
```diff
@@ -0,0 +1,252 @@
...
+ * separate fail-over regions.
+ */
+public class BatchFineGrainedRecoveryITCase extends TestLogger {
+	private static final int EMITTED_RECORD_NUMBER = 1000;
+	private static final int MAX_FAILURE_NUMBER = 10;
+	private static final int MAP_NUMBER = 3;
+	private static final List<Long> EXPECTED_JOB_OUTPUT =
+		LongStream.range(MAP_NUMBER, EMITTED_RECORD_NUMBER + MAP_NUMBER).boxed().collect(Collectors.toList());
+
+	private static TestingMiniCluster miniCluster;
```
zentol: it seems a bit odd that this is static. Is this purely for being able to use it in `createExecutionEnvironment`?

flink-tests/src/test/java/org/apache/flink/test/recovery/BatchFineGrainedRecoveryITCase.java
```diff
@@ -0,0 +1,252 @@
...
+		env.getConfig().setExecutionMode(ExecutionMode.BATCH_FORCED);
+		return env;
+	}
+
+	private enum StaticMapFailureTracker {
+		;
+
+		private static final List<AtomicInteger> mapRestarts = new ArrayList<>(10);
+		private static final List<AtomicInteger> expectedMapRestarts = new ArrayList<>(10);
+		private static final List<AtomicInteger> mapFailures = new ArrayList<>(10);
```
zentol: `mapFailures` is never read anywhere and could probably be removed

flink-tests/src/test/java/org/apache/flink/test/recovery/BatchFineGrainedRecoveryITCase.java
```diff
@@ -0,0 +1,252 @@
...
+		assertThat(input.collect(), is(EXPECTED_JOB_OUTPUT));
+
+		StaticMapFailureTracker.verify();
+	}
+
+	private static ExecutionEnvironment createExecutionEnvironment() {
+		@SuppressWarnings("StaticVariableUsedBeforeInitialization")
+		ExecutionEnvironment env = new TestEnvironment(miniCluster, 1, true);
+		env.setRestartStrategy(RestartStrategies.fixedDelayRestart(MAX_FAILURE_NUMBER, Time.milliseconds(10)));
+		env.getConfig().setExecutionMode(ExecutionMode.BATCH_FORCED);
```
zentol: add a comment that this forces all partitions to be blocking; the execution mode is quite an arcane thing.

flink-tests/src/test/java/org/apache/flink/test/recovery/BatchFineGrainedRecoveryITCase.java
```diff
@@ -0,0 +1,252 @@
...
+
+import static org.apache.flink.configuration.JobManagerOptions.FORCE_PARTITION_RELEASE_ON_CONSUMPTION;
+import static org.apache.flink.runtime.executiongraph.failover.FailoverStrategyLoader.PIPELINED_REGION_RESTART_STRATEGY_NAME;
+import static org.hamcrest.CoreMatchers.is;
+import static org.junit.Assert.assertThat;
+
+/**
+ * IT case for fine-grained recovery of batch jobs.
+ *
+ * <p>The test activates the region fail-over strategy to restart only failed producers and their consumers.
```
zentol: for this job we don't restart consumers; I fear the way this is phrased might trip people up.

flink-tests/src/test/java/org/apache/flink/test/recovery/BatchFineGrainedRecoveryITCase.java
azagrebin: does not work for atomic integers

PR #8972, opened 2019-07-03
[FLINK-13060][coordination] Respect restart constraints in new RegionFailover strategy

zentol:
Based on #8922. Ignore everything but my commit.

Extends the added `AdaptedRestartPipelinedRegionStrategyNG` to respect restart constraints imposed by the configured `RestartStrategy`.

Currently, the failover strategy does query `RestartStrategy#canRestart` before executing a failover, but the RS is never informed about this failover happening. As a result it is not taken into account for the maximum number of failure attempts (or max rate), nor is the restart delay honored.

This PR modifies the `AdaptedRestartPipelinedRegionStrategyNG` to call `RestartStrategy#restart` before executing any failover, with a `RestartCallback` that will execute the failover logic.

This change has a few repercussion on existing tests. Since the execution of the failover logic is now dependent on the restart strategy actually using the callback, we may no longer rely on the `InfiniteDelayRestartStrategy` as this one does not do so. Additionally, since restart strategies use the `schedule` method, we now have to additionally call `ManuallyTriggeredScheduledExecutor#triggerScheduledTasks`, as `#triggerAll` (for some reason) ignores scheduled tasks.

`AdaptedRestartPipelinedRegionStrategyNGFailoverTest#testFailoverExecutionDependentOnRestartStrategyRecoveryTrigger` tests that the failover behavior is dependent on the restart strategy. I'm not really happy with the test, as it effectively pauses the failover progress (after canceling the tasks, but before resetting them) and I suppose the exact state we should be in at this point is not well-defined.

### DIFF
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/AdaptedRestartPipelinedRegionStrategyNG.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/AdaptedRestartPipelinedRegionStrategyNG.java
@@ -31,6 +31,8 @@ import org.apache.flink.runtime.executiongraph.GlobalModVersionMismatch;
 import org.apache.flink.runtime.executiongraph.SchedulingUtils;
 import org.apache.flink.runtime.executiongraph.failover.adapter.DefaultFailoverTopology;
 import org.apache.flink.runtime.executiongraph.failover.flip1.RestartPipelinedRegionStrategy;
+import org.apache.flink.runtime.executiongraph.restart.RestartCallback;
+import org.apache.flink.runtime.executiongraph.restart.RestartStrategy;
 import org.apache.flink.runtime.jobgraph.JobStatus;
 import org.apache.flink.runtime.jobgraph.JobVertexID;
 import org.apache.flink.runtime.jobmanager.scheduler.CoLocationGroup;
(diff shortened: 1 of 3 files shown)


flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/AdaptedRestartPipelinedRegionStrategyNGFailoverTest.java
```diff
@@ -248,6 +253,45 @@ public void testNoRestart() throws Exception {
...
+
+		ev.fail(new Exception("Test Exception"));
+
+		manualMainThreadExecutor.triggerAll();
+
+		// the entire failover-procedure is being halted by the restart strategy not doing anything
+		// the only thing the failover strategy should do is cancel tasks that require it
+
+		// sanity check to ensure we actually called into the restart strategy
+		assertEquals(restartCallFuture.isDone(),true);
```
GJL: space after `,` missing

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/AdaptedRestartPipelinedRegionStrategyNGFailoverTest.java
```diff
@@ -248,6 +253,45 @@ public void testNoRestart() throws Exception {
...
 
+	/**
+	 * Tests that the execution of the restart logic of the failover strategy is dependent on the restart strategy
+	 * calling {@link RestartCallback#triggerFullRecovery()}.
+	 */
+	@Test
+	public void testFailoverExecutionDependentOnRestartStrategyRecoveryTrigger() throws Exception {
+		final JobGraph jobGraph = createBatchJobGraph();
+		final CompletableFuture<Void> restartCallFuture = new CompletableFuture<>();
+		final RestartStrategy restartStrategy = new RestartStrategy() {
```
GJL: Is it not possible to use `TestRestartStrategy manuallyTriggered()`? It has the method `getNumberOfQueuedActions()`.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/AdaptedRestartPipelinedRegionStrategyNGFailoverTest.java
GJL: I think we use `restartCallFuture` as a `CountDownLatch` or `AtomicInteger` here. See my other comment at the beginning of this test.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/AdaptedRestartPipelinedRegionStrategyNG.java
```diff
@@ -107,11 +107,16 @@ protected void restartTasks(final Set<ExecutionVertexID> verticesToRestart) {
 
 		FutureUtils.assertNoException(
 			cancelTasks(verticesToRestart)
-				.thenRunAsync(resetAndRescheduleTasks(globalModVersion, vertexVersions), executionGraph.getJobMasterMainThreadExecutor())
+				.thenApply(ignored -> createResetAndRescheduleTasksRunnable(globalModVersion, vertexVersions))
+				.thenAccept(restartAction -> executionGraph.getRestartStrategy()
```
GJL: I don't think the `.thenApply` stage is needed. Also, the `.thenAccept` stage breaks the level of abstraction. I'd extract the body into a new function. Consider this structure:

```
			cancelTasks(verticesToRestart)
				.thenRun(scheduleRestart(globalModVersion, vertexVersions)) // strictly speaking should be thenRunAsync?
				.handle(failGlobalOnError()));
	}

	private Runnable scheduleRestart(final long globalModVersion, final Set<ExecutionVertexVersion> vertexVersions) {
		final RestartStrategy restartStrategy = executionGraph.getRestartStrategy();
		return () -> restartStrategy.restart(
			resetAndRescheduleTasks(globalModVersion, vertexVersions),
			executionGraph.getJobMasterMainThreadExecutor()
		);
	}

	private RestartCallback resetAndRescheduleTasks(final long globalModVersion, final Set<ExecutionVertexVersion> vertexVersions) {
		...
```

PR #8973, opened 2019-07-03
[FLINK-13081][table-planner][table-api-java] Supports explain DAG plan

godfreyhe:

*Supports explain DAG plan*

## Brief change log

  - *add explain(Boolean) into TableEnvironment*
  - *add isLazyOptMode field into EnvironmentSettings, which tells whether the table environment should optimize a query in sqlUpdate&insertInto method or before executing*

## Verifying this change

the existing test cases should be passed

## Documentation

### DIFF
--- a/flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/TableEnvironment.java
+++ b/flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/TableEnvironment.java
@@ -290,6 +290,25 @@ public interface TableEnvironment {
 	String explain(Table table);
 
 	/**
+	 * Returns the AST of the specified Table API and SQL queries and the execution plan to compute
+	 * the result of the given {@link Table}.
+	 *
+	 * @param table The table for which the AST and execution plan will be returned.
+	 * @param extended if the plan should contain additional properties such as
+	 * e.g. estimated cost, traits
+	 */
+	String explain(Table table, boolean extended);
+
+	/**
+	 * Returns the AST of the specified Table API and SQL queries and the execution plan to compute
+	 * the result of multiple-sinks plan.
+	 *
+	 * @param extended if the plan should contain additional properties such as
+	 * e.g. estimated cost, traits
+	 */
+	String explain(boolean extended);
+
+	/**
 	 * Returns completion hints for the given statement at the given cursor position.
 	 * The completion happens case insensitively.
 	 *
--- a/flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/internal/TableEnvironmentImpl.java
+++ b/flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/internal/TableEnvironmentImpl.java
@@ -68,6 +68,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.stream.Collectors;
 
 /**
  * Implementation of {@link TableEnvironment} that works exclusively with Table API interfaces.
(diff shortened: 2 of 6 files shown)


godfreyhe:
cc @dawidwys @twalthr

dawidwys:
Hi @godfreyhe could you split this PR into two separate functionalities (`explain` & `execute`).

I do agree how `explain` is implemented, but the `execute` method should be briefly discussed.

godfreyhe:
sure, split this PR into two commits or two PRs ?

yes, there are some unclear questions about `execute` method: which `execute` methods (`TableEnvironment#execute` or `StreamExecutionEnvironment#execute`) users should use? how to make sure users only use `TableEnvironment#execute` method if their code only depends on planner module instead of bridge module？

dawidwys:
Yes, those are exactly the concerns that we also have. I guess splitting into a separate commits will suffice, but it would be better to split it into separate PRs.

godfreyhe:
I have split original PR into two PRs: [FLINK-13081](https://issues.apache.org/jira/browse/FLINK-13081) & [FLINK-13088](https://issues.apache.org/jira/browse/FLINK-13088). this PR only involves `explain(Boolean)` method

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/EnvironmentSettings.java
```diff
@@ -121,6 +131,16 @@ public boolean isBatchMode() {
 		return isBatchMode;
 	}
 
+	/**
+	 * Tells if the {@link TableEnvironment should trigger planner#translate
+	 * in {@link TableEnvironment#insertInto(Table, String, String...)}/
+	 * {@link TableEnvironment#sqlUpdate(String)} methods or
+	 * in {@link TableEnvironment#execute(String)} method.
+	 */
+	public boolean isLazyOptMode() {
```
dawidwys: This is not necessary for the `explain` method.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/TableEnvironment.java
```diff
@@ -275,8 +275,28 @@
 	 *
 	 * @param table The table for which the AST and execution plan will be returned.
 	 */
+	@Deprecated
```
dawidwys: Why is it deprecated? I think it is valid. I think we can safely assume that by default `extended = false`.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/TableEnvironment.java
```diff
@@ -275,8 +275,28 @@
...
+	String explain(Table table, boolean extended);
+
+	/**
+	 * Returns the AST of the specified Table API and SQL queries and the execution plan to compute
+	 * the result of multiple-sinks plan.
+	 *
+	 * @param extended if the plan should contain additional properties such as
+	 * e.g. estimated cost, traits
+	 */
+	String explain(boolean extended);
```
dawidwys: Let's add this in the second PR. We do not have the concept of lazy evaluation yet.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/TableEnvironment.java
godfreyhe: ok, i will revert it. i want to unify `public String explain(Table table)` into `public String explain(Table table, boolean extended)` before, which looks more simple.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/TableEnvironment.java
godfreyhe: The main goal of this PR is to introduce `explain(Boolean)` method, otherwise this PR is meaningless.

dawidwys: Let's rebase it on top of #8984 then

godfreyhe:
@dawidwys , i have updated this PR

dawidwys:
Thank you for the updated. I fixed small mistake and merged it.

PR #8978, opened 2019-07-04
[FLINK-13089][table-planner-blink] Implement batch nested loop join and add join it cases in blink

JingsongLi:

Nested loop join has two advantages:
1.Nested loop join is quicker when build row size is small.
2.Nested loop join support all kind of joins, include non-key join.

## Brief change log

Introduce NestedLoopJoinCodeGenerator.
Implement BatchExecNestedLoopJoin.
Add join it cases.

## Verifying this change

it cases

## Documentation


flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/codegen/NestedLoopJoinCodeGenerator.scala
```diff
@@ -0,0 +1,369 @@
...
+    }
+  }
+
+  def gen(): CodeGenOperatorFactory[BaseRow] = {
+    val config = ctx.tableConfig
+
+    val exprGenerator = new ExprCodeGenerator(ctx, joinType.isOuter)
+        .bindInput(leftType).bindSecondInput(rightType)
+
+    // otherwise we use ResettableExternalBuffer to prevent OOM
```
KurtYoung: otherwise?

JingsongLi: mistake...

PR #8981, opened 2019-07-04
[FLINK-13097] Make the cause for EOFException explicit (buffer depletion)

cchepelov:

This pull request makes explicit that the cause of EOFException is the inability to further allocate buffers.
The JDK documentation hints but does not specify that EOFException is "mainly" made for input channels, making the choice of EOFException valid to the letter of the specification. 
However, not providing an explanatory text makes it difficult as a newcomer to understand that Flink isn't failing to read (as would be expected of an EOFException) but to allocate a buffer

Alternates solutions considered:
* throw OutOfMemoryError: errors are not made to be thrown by application-level code; furthermore, and contrary to OOME, the depletion of a pre-allocated MemorySegmentSource does **not** mean that the JVM is in an unpredictable state
* throw ClosedChannelException: the definition of this exception may make it more appropriate, however it is also vague enough that explanatory text would still be important to facilitate diagnosis. Furthermore, while the AbstractPagedOutputView contract would allow it ("throws IOException"), this may be a surprising change to existing consumers.

## Brief change log

- throw new EOFException now uses the ctor with an explanatory message.

## Verifying this change

  - The serializers: no

## Documentation

### DIFF
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/io/disk/SimpleCollectingOutputView.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/io/disk/SimpleCollectingOutputView.java
@@ -76,7 +76,7 @@ public class SimpleCollectingOutputView extends AbstractPagedOutputView {
 			this.segmentNum++;
 			return next;
 		} else {
-			throw new EOFException();
+			throw new EOFException("Can't collect further: memorySource depleted");
 		}
 	}
 	


cchepelov:
please disregard until I redo a  proper "push -f".

 This should be a one-liner.

cchepelov:
apologies for the trouble @rmetzger

aljoscha:
Thanks! I merged this.

PR #8984, opened 2019-07-04
 [FLINK-13088][table-api] Support lazy query transformation & execution on TableEnvironment

dawidwys:

In the long run we want to have a single self-contained unified `TableEnvironment` equivalent to `ExecutionEnvironment` & `StreamExecutionEnvironment` with `execute` method. Moreover Blink planner translates queries lazily when the `execute` method is called on `TableEnvironment` This PR enables creating an instance of the unified TableEnvironment. It also implements the lazy translation schema.

## Verifying this change

## Documentation


dawidwys:
cc @godfreyhe @twalthr

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/internal/TableEnvironmentImpl.java
```diff
@@ -337,6 +362,47 @@ public TableConfig getConfig() {
...
+	 * Subclasses can override this method to add additional checks.
+	 *
+	 * @param tableSource tableSource to validate
+	 */
+	protected void validateTableSource(TableSource<?> tableSource) {
+		TableSourceValidation.validateTableSource(tableSource);
+	}
+
+	private void translate(List<ModifyOperation> modifyOperations) {
+		List<Transformation<?>> transformations = planner.translate(modifyOperations);
```
godfreyhe: will `translate` twice if `shouldTranslateEagerly` is true ?

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/internal/TableEnvironmentImpl.java
```diff
@@ -337,6 +362,47 @@ public TableConfig getConfig() {
 		return tableConfig;
 	}
 
+	@Override
+	public JobExecutionResult execute(String jobName) throws Exception {
+		translate(bufferedModifyOperations);
+		return execEnv.execute(jobName);
```
godfreyhe: `bufferedModifyOperations` must be cleared here, otherwise the `bufferedModifyOperations` will be translated again and again if table env is used for multiple job. 
e.g.
val tableEnv = // shouldTranslateEagerly is false
tableEnv.sqlUpdate("insert into table1 select ...");
tableEnv.execute("job1")
tableEnv.sqlUpdate("insert into table2 select ..."); 
tableEnv.execute("job2")

flink-table/flink-table-api-java-bridge/src/main/java/org/apache/flink/table/api/java/internal/StreamTableEnvironmentImpl.java
```diff
@@ -305,6 +327,11 @@ protected void validateTableSource(TableSource<?> tableSource) {
 		validateTimeCharacteristic(TableSourceValidation.hasRowtimeAttribute(tableSource));
 	}
 
+	@Override
+	protected boolean shouldTranslateEagerly() {
+		return true;
```
godfreyhe: blink planner will also use this class, if i understand correctly. so here could not return true deterministically.

flink-table/flink-table-api-scala-bridge/src/main/scala/org/apache/flink/table/api/scala/internal/StreamTableEnvironmentImpl.scala
```diff
@@ -182,6 +187,8 @@ class StreamTableEnvironmentImpl (
     }
   }
 
+  override protected def shouldTranslateEagerly(): Boolean = true
```
godfreyhe: same as `java.StreamTableEnvironmentImpl#shouldTranslateEagerly`

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/internal/TableEnvironmentImpl.java
dawidwys: good catch!

flink-table/flink-table-api-java-bridge/src/main/java/org/apache/flink/table/api/java/internal/StreamTableEnvironmentImpl.java
dawidwys: Correct, we wanted to leave it this way. The thing is if we use this class we should not allow lazy translation. All `toDataStream` conversion must be translated eagerly. If we allowed for other transformations to be executed lazily then some parts would be executed lazily and others eagerly. This poses many problems. How do we handle DML? DCLs commands if some are executed eagerly and others lazily.

Another problem is what happens if the user calls `execute` on `StreamExecutionEnvironment`, which queries will be executed?

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/internal/TableEnvironmentImpl.java
dawidwys: Why? AFAIK just once.

godfreyhe: yes, just once. `modifyOperations` is always empty when `shouldTranslateEagerly` is true.

flink-table/flink-table-api-java-bridge/src/main/java/org/apache/flink/table/api/java/internal/StreamTableEnvironmentImpl.java
godfreyhe: make sense

godfreyhe: `<T, ACC> void registerFunction(String name, AggregateFunction<T, ACC> aggregateFunction);` and `<T> void registerFunction(String name, TableFunction<T> tableFunction);` are only defined in `StreamTableEnvironment`, not in `TableEnvironment`. Should they be moved to  `TableEnvironment` to support lazy translation?

dawidwys: We cannot do that. Those interface depend on type extraction (java/scala specific). We cannot add them to the unified interface. This has to wait until we have the new type inference for udfs in place.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/internal/TableEnvironmentImpl.java
```diff
@@ -266,31 +264,24 @@ public Table sqlQuery(String query) {
...
-				table.getQueryOperation())));
-
-		execEnv.apply(translate);
-	}
+				table.getQueryOperation()));
 
-	@Override
-	public void insertInto(Table table, String path, String... pathContinued) {
-		insertInto(table, new StreamQueryConfig(), path, pathContinued);
+		if (shouldTranslateEagerly()) {
```
twalthr: nit: can we give those methods better names? `isEagerOperationTranslation`, `translateOperation`, `bufferOperation`?

flink-table/flink-table-planner/src/test/scala/org/apache/flink/table/runtime/harness/GroupAggregateHarnessTest.scala
```diff
@@ -185,7 +191,7 @@ class GroupAggregateHarnessTest extends HarnessTestBase {
   @Test
   def testDistinctAggregateWithRetract(): Unit = {
     val env = StreamExecutionEnvironment.getExecutionEnvironment
-    val tEnv = StreamTableEnvironment.create(env)
+    val tEnv = StreamTableEnvironment.create(env, tableConfig)
```
twalthr: don't use deprecated constructor

flink-table/flink-table-planner/src/test/scala/org/apache/flink/table/runtime/harness/TableAggregateHarnessTest.scala
```diff
@@ -17,33 +17,37 @@
...
+import java.lang.{Integer => JInt}
+import java.util.concurrent.ConcurrentLinkedQueue
+
 import scala.collection.mutable
 
 class TableAggregateHarnessTest extends HarnessTestBase {
 
-  protected var queryConfig =
-    new TestStreamQueryConfig(Time.seconds(2), Time.seconds(3))
+  private val tableConfig = new TableConfig {
```
twalthr: use don't extend from table config, this class should be final.

flink-table/flink-table-api-java-bridge/src/main/java/org/apache/flink/table/api/java/internal/StreamTableEnvironmentImpl.java
```diff
@@ -98,6 +98,7 @@ public StreamTableEnvironmentImpl(
 	 * from/to {@link DataStream}.
 	 *
 	 * @param tableConfig The configuration of the TableEnvironment.
+	 * @param settings The settings required to instantiate proper components.
```
twalthr: Remove these docs all together?

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/TableEnvironment.java
```diff
@@ -53,6 +54,26 @@
 @PublicEvolving
 public interface TableEnvironment {
 
+	/**
+	 * Creates a table environment that is the entry point and central context for creating Table & SQL
+	 * API programs.
+	 *
+	 * <p>It is unified for bounded and unbounded data processing.
+	 *
+	 * <p>A stream table environment is responsible for:
```
twalthr: Wrong docs copy. Use the docs of the class.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/internal/TableEnvironmentImpl.java
```diff
@@ -109,6 +115,33 @@ protected TableEnvironmentImpl(
...
 	}
 
+	/**
+	 * Creates an instance of a unified {@link TableEnvironment}..
+	 */
+	public static TableEnvironmentImpl create(EnvironmentSettings settings) {
+		FunctionCatalog functionCatalog = new FunctionCatalog(
+			settings.getBuiltInCatalogName(),
+			settings.getBuiltInDatabaseName());
+		CatalogManager catalogManager = new CatalogManager(
```
twalthr: nit: could we add at least a couple of empty lines to structure the code a bit.

flink-table/flink-table-api-java-bridge/src/main/java/org/apache/flink/table/api/java/internal/StreamTableEnvironmentImpl.java
```diff
@@ -282,6 +287,22 @@ public StreamTableDescriptor connect(ConnectorDescriptor connectorDescriptor) {
 		return (StreamTableDescriptor) super.connect(connectorDescriptor);
 	}
 
+	@Override
+	public void sqlUpdate(String stmt, StreamQueryConfig config) {
```
twalthr: Deprecate all QueryConfig classes and refer to `TableConfig`.

flink-table/flink-table-api-java-bridge/src/main/java/org/apache/flink/table/api/java/internal/StreamTableEnvironmentImpl.java
godfreyhe: maybe the plan result of tests which use `registerFunction(String name, AggregateFunction<T, ACC> aggregateFunction)` and ` registerFunction(String name, TableFunction<T> tableFunction)` will be updated in blink planner,  and be reverted back when new type inference is ready.

flink-table/flink-table-planner/src/test/scala/org/apache/flink/table/runtime/harness/TableAggregateHarnessTest.scala
dawidwys: The problem is that we have there a safety check that disallow values used in those tests. I did not want to rewrite those tests yet.

flink-table/flink-table-planner/src/test/scala/org/apache/flink/table/runtime/harness/GroupAggregateHarnessTest.scala
dawidwys: Unfortunately this is the only way as of now to pass the custom TableConfig, which we have to use because of the incompatible values used in these tests (see comment below)

godfreyhe:
hi @dawidwys , i'm trying to adapt blink planner to unified table environment + this pr. in blink planner, there are many kinds of plan tests, such as: rule test(`LogicalUnnestRuleTest`), logical plan test(`SubQueryAntiJoinTest`), physical plan test(`CalcTest`), and so on. and plan test result contains `planBefore` (the original `RelNode` tree) and `planAfter`(the optimized `RelNode` tree). so, i want to expose `bufferedModifyOperations` as `protected` or add a `get` method for testing.

twalthr:
Hi @godfreyhe, I would like to propose the following solution to keep the reworking efforts of the tests low and avoid changes of the API classes just for tests. As far as I know, all test classes extend from some test base, right? The test base can provide a mocking table environment `TestingTableEnvironment implements TableEnvironment` that allows for arbitrary testing logic. E.g. buffering `ModifyOperation`s. What do you think?

godfreyhe:
@twalthr thanks for your suggestion. yes, all test classes extend from `TableTestBase` in blink planner.      we could add a `TestingTableEnvironment` which extends `TableEnvironmentImpl`, and override `insertInto`, `sqlUpdate`, `execute` methods. (`insertInto` method and `sqlUpdate` method in `TestingTableEnvironment` are same with `TableEnvironment`'s)

zentol:
CI report for commit 0aa1e39f2100670115c4ebf91346e573a55a1762: SUCCESS [Travis](https://travis-ci.org/dataArtisans/flink/builds/554657921?utm_source=github_status&utm_medium=notification)

PR #8982, opened 2019-07-04
[FLINK-13098][datastream] Add a new type UNDEFINED of shuffle mode

ifndef-SleePy:
## What is the purpose of the change

* This new shuffle type leaves some space for optimization later. The optimization might be based on resources or some global settings.

## Brief change log

* The `UNDEFINED` type is the default value of shuffle mode. If there is no specific `PartitionTransformation`, the shuffle mode would be `UNDEFINED`.
* The `UNDEFINED` shuffle mode can be chained

## Verifying this change

* Modify the existing case to cover this

## Does this pull request potentially affect one of the following parts:

  - Dependencies (does it add or upgrade a dependency): no
  - The public API, i.e., is any changed class annotated with `@Public(Evolving)`: **yes**
  - The serializers: no
  - The runtime per-record code paths (performance sensitive): no
  - Anything that affects deployment or recovery: JobManager (and its components), Checkpointing, Yarn/Mesos, ZooKeeper: no
  - The S3 file system connector: no

## Documentation

  - Does this pull request introduce a new feature? no
  - If yes, how is the feature documented? not applicable

PR #8985, opened 2019-07-04
[FLINK-13101][datastream] Introduce blockingConnectionsBetweenChains property of StreamGraph

ifndef-SleePy:

* **This is based on https://github.com/apache/flink/pull/8982**

* The reason of introducing it is to satisfy the requirement of Blink batch planner.
* Because the current scheduling strategy is a bit simple. It can not support some complex scenarios, like a batch job with resources limited.
* To be honest, it's probably a work-around solution. However it's an internal implementation, we can replace it when we are able to support batch job by scheduling strategy.

## Brief change log

* Add a property `blockingConnectionsBetweenChains` of `StreamGraph`
* The property `blockingConnectionsBetweenChains` means, if there are some stream edges that can not be chained and the shuffle mode of edge is not specified, translate these edges into `BLOCKING` result partition type.

## Verifying this change

* This change added unit tests

  - The serializers: no

## Documentation

### DIFF
--- a/flink-streaming-java/src/main/java/org/apache/flink/streaming/api/graph/StreamGraph.java
+++ b/flink-streaming-java/src/main/java/org/apache/flink/streaming/api/graph/StreamGraph.java
@@ -97,6 +97,12 @@ public class StreamGraph extends StreamingPlan {
 
 	private TimeCharacteristic timeCharacteristic;
 
+	/**
+	 * If there are some stream edges that can not be chained and the shuffle mode of edge is not specified,
+	 * translate these edges into `BLOCKING` result partition type.
+	 */
+	private boolean blockingConnectionsBetweenChains;
+
 	private Map<Integer, StreamNode> streamNodes;
 	private Set<Integer> sources;
 	private Set<Integer> sinks;
@@ -182,6 +188,14 @@ public class StreamGraph extends StreamingPlan {
 
 	public void setTimeCharacteristic(TimeCharacteristic timeCharacteristic) {
 		this.timeCharacteristic = timeCharacteristic;
+	}
+
+	public boolean isBlockingConnectionsBetweenChains() {
+		return blockingConnectionsBetweenChains;
+	}
+
+	public void setBlockingConnectionsBetweenChains(boolean blockingConnectionsBetweenChains) {
+		this.blockingConnectionsBetweenChains = blockingConnectionsBetweenChains;
 	}
 
 	// Checkpointing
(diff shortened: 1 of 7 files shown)


ifndef-SleePy:
@flinkbot attention @aljoscha

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/graph/StreamGraph.java
```diff
@@ -97,6 +97,8 @@
 
 	private TimeCharacteristic timeCharacteristic;
 
+	private boolean blockingAfterChainingOff;
```
guoweiM: I think `slotSharingEnableAfterChain` is more accurate?  From the discussion in the slack what my understand is that the planner only controls whether slot is sharing and ShuffleMode is determined by this currently.
What do you think?

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/graph/StreamGraphGenerator.java
```diff
@@ -123,6 +123,12 @@
 
 	private String jobName = DEFAULT_JOB_NAME;
 
+	/**
+	 * If there are some stream edges that can not be chained and the shuffle mode of edge is not specified,
+	 * translate these edges into `BLOCKING` result partition type.
+	 */
+	private boolean blockingAfterChainingOff = false;
```
guoweiM: Why add this property?  Planner can get `StreamGraph` and set the property to it.

ifndef-SleePy: Yes, planner could set `StreamGraph` directly. But that's a work-around way for 1.9. Actually the `StreamGraphGenerator` should be the only entry of `StreamGraph`. So just keep it here.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/graph/StreamGraph.java
ifndef-SleePy: I'm not sure we have reached an agreement of details. I guess we have a divergence here.
Let's wait for the opinion of Aljoscha and Stephan.

ifndef-SleePy:
@flinkbot attention @aljoscha

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/graph/StreamGraph.java
guoweiM: ok. I just remind another thing that Blink Planner might set "slot-sharing after chaining" off and ask ShuffleMode between tasks to be PIPELINE.

ifndef-SleePy:
I have implemented an alternative solution https://github.com/apache/flink/pull/9000. Maybe the other one is more close with the original discussion.

zentol:
CI report for commit 34336fc5898ab8bd74eb30153dcc13dd9611983f: SUCCESS [Travis](https://travis-ci.org/dataArtisans/flink/builds/554657817?utm_source=github_status&utm_medium=notification)

aljoscha:
Merged. 👌

PR #9000, opened 2019-07-05
[FLINK-13101][datastream] Add slotSharingAfterChainingDisabled proper…

ifndef-SleePy:
…ty of StreamGraph

* The reason of introducing it is to satisfy the requirement of Blink batch planner.
* Because the current scheduling strategy is a bit simple. It can not support some complex scenarios, like a batch job with resources limited.
* To be honest, it's probably a work-around solution. However it's an internal implementation, we can replace it when we are able to support batch job by scheduling strategy.
* **This is alternative solution of FLINK-13101**

## Brief change log

* Add a property `slotSharingAfterChainingDisabled` of `StreamGraph`
* The property means, if there are some operators that can not be chained, disable slot sharing and translate these edges into `BLOCKING` result partition type.

## Verifying this change

* This change added unit tests

  - The serializers: no

## Documentation

### DIFF
--- a/flink-streaming-java/src/main/java/org/apache/flink/streaming/api/graph/StreamGraph.java
+++ b/flink-streaming-java/src/main/java/org/apache/flink/streaming/api/graph/StreamGraph.java
@@ -97,6 +97,11 @@ public class StreamGraph extends StreamingPlan {
 
 	private TimeCharacteristic timeCharacteristic;
 
+	/**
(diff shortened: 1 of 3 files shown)


zentol:
CI report for commit 2f861e597ff2056e8012caa6d6fc9d460f117fe8: SUCCESS [Travis](https://travis-ci.org/dataArtisans/flink/builds/554657363?utm_source=github_status&utm_medium=notification)

PR #9006, opened 2019-07-06
[FLINK-13107][table-planner-blink] Copy TableApi IT and UT to Blink planner

beyond1920:

The issue aims to copy the testcases in the following packages from flink-planner and original blink to Blink-planner:
1. org.apache.flink.table.api.batch.table
2. org.apache.flink.table.api.stream.table
3. org.apache.flink.table.runtime.batch.table
4. org.apache.flink.table.runtime.stream.table

## Brief change log

  - Copy UT and ITCase of TableApi
  - Fix some Bugs, commit message contains detail information about those Bugs

## Verifying this change

UT, ITCase

  - The serializers: (no)

## Documentation


flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/expressions/RexNodeConverter.java
```diff
@@ -529,7 +530,8 @@ private RexNode convertOver(List<Expression> children) {
 				.collect(Collectors.toList());
 		// assemble bounds
 		Expression preceding = args.get(2);
-		boolean isPhysical = ((ResolvedExpression) preceding).getOutputDataType().equals(DataTypes.BIGINT());
+		boolean isPhysical = fromDataTypeToTypeInfo(((ResolvedExpression) preceding).getOutputDataType()).equals(
```
KurtYoung: what's the difference between LongTypeInfo and BIGINT datatype?

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/plan/schema/TypedFlinkTableFunction.scala
```diff
@@ -46,8 +47,18 @@ class TypedFlinkTableFunction(
       typeFactory: RelDataTypeFactory,
       arguments: Array[AnyRef],
       argTypes: Array[Class[_]]): RelDataType = {
-    // we have determined the row type before, just convert it to RelDataType
-    typeFactory.asInstanceOf[FlinkTypeFactory].createFieldTypeFromLogicalType(
-      fromDataTypeToLogicalType(externalResultType))
+    val fieldTypes = TableEnvironment.getFieldTypes(externalResultType)
```
KurtYoung: I would suggest to not use `TableEnvironment.getFieldTypes`, it will be deleted soon

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/expressions/RexNodeConverter.java
beyond1920: TypeInformation has no nullable information while DataType has.
There is no need to compare nullable here.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/plan/schema/TypedFlinkTableFunction.scala
beyond1920: OK, I would use FieldInfoUtils in api-module instead.

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/expressions/RexNodeConverter.java
KurtYoung: How about we ignore the nullability during comparison, but not convert to TypeInformation. It seems unnecessary.

JingsongLi:
We need copy all `org.apache.flink.table.api.*`? I think we have these duplicate plan cases in `org.apache.flink.table.table.*`? Why not put cases to `org.apache.flink.table.table.*` to be consistent with previous SQL plan tests?

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/types/ClassDataTypeConverter.java
```diff
@@ -31,6 +35,24 @@
  * Convert unknown types by {@link LegacyTypeInformationType}.
  */
 public class ClassDataTypeConverter {
+	private static final Map<String, DataType> primitiveClassDataTypes = new HashMap<>();
+	static {
+
+		// NOTE: this list explicitly excludes data types that need further parameters
+		// exclusions: DECIMAL, INTERVAL YEAR TO MONTH, MAP, MULTISET, ROW, NULL, ANY
+		addDefaultDataType(boolean.class, DataTypes.BOOLEAN());
```
JingsongLi: -1 about this change, why we need convert a `int` to `DataType(Integer)`?

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/expressions/aggregations.scala
```diff
@@ -57,7 +58,10 @@ case class Sum(child: PlannerExpression) extends Aggregation {
   override private[flink] def children: Seq[PlannerExpression] = Seq(child)
   override def toString = s"sum($child)"
 
-  override private[flink] def resultType = child.resultType
+  override private[flink] def resultType = {
```
JingsongLi: > [FLINK-13107][table-planner-blink] Fix bug when infer resultType of some PlannerExpression.
[FLINK-13107][table-planner-blink] Derive sum, avg, div return type using behavior of blink ?

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/expressions/RexNodeConverter.java
JingsongLi: Here, I think we need use `LogicalTypeRoot`, because we don't want to judge `nullable` and `conversionClass`.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/plan/util/SetOpRewriteUtil.scala
```diff
@@ -81,7 +81,7 @@ object SetOpRewriteUtil {
     val fieldTypes = logicalType.getChildren.map(fromLogicalTypeToTypeInfo).toArray
     val tf = new ReplicateRows(fieldTypes)
     val resultType = fromLegacyInfoToDataType(new RowTypeInfo(fieldTypes, fieldNames))
-    val function = new TypedFlinkTableFunction(tf, resultType)
+    val function = new TypedFlinkTableFunction(tf, fieldNames, resultType)
```
JingsongLi: This should be first commit.

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/types/ClassDataTypeConverter.java
beyond1920: The code is similar with `ClassDataTypeConverter` in table-common module. When UDF has a `eval` method explicitly return `int` type, we need convert a `int` to `DataType`. 
I add code here to fix the Inconsistence between `ClassDataTypeConverter` with `LegacyTypeInfoDataTypeConverter`.
`ClassDataTypeConverter` map Class to DataTypes, However, for primitive class, it explicit do the following logical to binds convention class:
`private static void addDefaultDataType(Class<?> clazz, DataType rootType) {
		final DataType dataType;
		if (clazz.isPrimitive()) {
			dataType = rootType.notNull();
		} else {
			dataType = rootType.nullable();
		}
		defaultDataTypes.put(clazz.getName(), dataType.bridgedTo(clazz));
	}
`
So for the primitive class, for example, int.class, it's DataType binding to int.class, instead of Integer.class.
The logical is not consistent with `LegacyTypeInfoDataTypeConverter`.
the mapping in `LegacyTypeInfoDataTypeConverter` only contains DataTypes.INT().bridgedTo(Integer.class), not including DataTypes.INT().bridgedTo(int.class).

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/plan/util/SetOpRewriteUtil.scala
beyond1920: Good catch. However `SetOpRewriteUtil` is a new class, which I found compile error after rebase code, so The update not in first commit...

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/expressions/aggregations.scala
beyond1920: I think we should use behavior of blink.

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/expressions/RexNodeConverter.java
beyond1920: Ok

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/types/ClassDataTypeConverter.java
JingsongLi: So why not change `TypeInfoDataTypeConverter`? This is the conversion util in blink.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/expressions/aggregations.scala
JingsongLi: Yeah, The change is very good, I think the commit message should be more clear.

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/types/ClassDataTypeConverter.java
beyond1920: Good suggestion.

KurtYoung:
> What are the implications on the test times when we copy so many tests?

According to this: https://travis-ci.org/beyond1920/flink/builds/557233740, the test time increases less than 2 mins.

KurtYoung:
travis passed here: https://travis-ci.org/beyond1920/flink/builds/557233740
I'm merging this

PR #9007, opened 2019-07-06
[FLINK-13131] [Deployment / YARN] add a custom directory for storing configuration files and jar files for user jobs

zhangbinzaifendou:
The user-related jars and configuration files can only be placed under fileSystem.getHomeDirectory(/home/xxx), which is inflexible and increases the configuration so that users can customize the directory.

### DIFF
--- a/flink-connectors/flink-connector-hive/pom.xml
+++ b/flink-connectors/flink-connector-hive/pom.xml
@@ -562,5 +562,24 @@ under the License.
 				</plugins>
 			</build>
 		</profile>
+		<profile>
+			<id>java11</id>
+			<activation>
+				<jdk>11</jdk>
+			</activation>
+
+			<build>
+				<plugins>
+					<plugin>
+						<groupId>org.apache.maven.plugins</groupId>
+						<artifactId>maven-surefire-plugin</artifactId>
+						<configuration>
+							<!-- hive does not support Java 11 -->
+							<skip>true</skip>
+						</configuration>
+					</plugin>
+				</plugins>
+			</build>
+		</profile>
 	</profiles>
 </project>


zhangbinzaifendou:
@Myasuka  Thank you very much for your review and Suggestions
1. In hadoop 2.7.2, if dfs.user.home.dir.prefix=/home is set, but the value returned by getHomeDirectory is /user/xxx, the actual online environment has speciality, so I think flink should provide a flexible configuration. So that the user can set a directory, if not set, use the default (/user/xxx).
2. I have modified the problem of import *
Thank you!

Myasuka:
@zhangbinzaifendou Please correct me if I am wrong, if `dfs.user.home.dir.prefix` has been configured to another value, e.g. `"/home"`. From Hadoop latest code, [`DistributedFileSystem`](https://github.com/apache/hadoop/blob/9c907294860a48f0d1676a31bda15795b7a6771a/hadoop-hdfs-project/hadoop-hdfs-client/src/main/java/org/apache/hadoop/hdfs/DistributedFileSystem.java#L211) would return the home directory as `"/home/userxxx"` via [DFSUtilClient](https://github.com/apache/hadoop/blob/9c907294860a48f0d1676a31bda15795b7a6771a/hadoop-hdfs-project/hadoop-hdfs-client/src/main/java/org/apache/hadoop/hdfs/DFSUtilClient.java#L1006). In other words, there is no need for flink user to configure the staging folder since Hadoop could return the wanted home directory even `dfs.user.home.dir.prefix` changed.

zhangbinzaifendou:
@Myasuka Sorry for not being precise.
The apache community version of hadoop is fine, there is a problem in our hadoop version, I also said that our online environment is very special, for our online environment, so we need to add a configuration that can meet our needs very simply. It also adds flexibility to flink.If the user wants to redefine a new directory, flink can't do it now. So the benefits are obvious.
Thank you!

Myasuka:
@zhangbinzaifendou as you explained, the motivation to create this PR is due to your special online environment and your Hadoop. However, this is not the reason that you have to modify the Flink source code. I think the real solution is to modify your special Hadoop which caused this problem.

As an open source software, Flink should adapt to the standard or widely-used ecosystems not some special customized ones. There would be no problems if we follow the correct logic of `FileSystem.getHomeDirectory()` even `dfs.user.home.dir.prefix` has been configured. And Flink use `FileSystem.getHomeDirectory()` to store its internal jars and files which is totally transparent to users. Exposing this internal implementation to users by adding such a configuration would never be a good idea.

zhangbinzaifendou:
@Myasuka Thanks for your reply.
Our design is based on the needs of the company, the changes are simple, but the impact is huge. This is obviously very flexible if Flink supports user-defined directories. I believe that many open source software must have a lot of unreasonable design from the beginning, and it is gradually matured based on the continuous improvement of many companies' real scenes. It is because we have such a scenario that we find that Flink design is not flexible enough, so we need to increase this feature, so that it will not affect the original use, but also ensure flexibility, is it better?

wangyang0918:
Hi @zhangbinzaifendou, thanks for your contribution.

I agree with @Myasuka. Your private hadoop version should be updated to support this scenario. Maybe you could build your own flink-shaded-hadoop uber jar.

zhangbinzaifendou:
@wangyang0918  Thank you for your answer. 
I'm afraid I don't quite agree with you. User scenarios are very complex, shouldn't there be a flexible configuration? Providing flexible configuration is a principle of good architectural design.

PR #9020, opened 2019-07-08
[FLINK-13078][table-common] Preparation for a type parser utility

twalthr:

This adds some last minute changes to the serialization format of the logical types. These are prerequisites also for a type parser utility.

## Brief change log

See commit messages.

## Verifying this change

This change is already covered by existing tests, such as `LogicalTypesTest`.

  - The serializers: no

## Documentation


aljoscha:
So basically, this is wrapping plain strings in `''` and using proper default values? The former is so that the serialized representation can be uses to parse it again? If yes, then this looks good to merge!

twalthr:
Yes, otherwise the parser tokenization becomes tricky because it needs to be context aware. Strings should be surrounded by `' '`. The synonyms should be documented somewhere. Btw I forgot that `INTEGER` is a synonym for `INT`. I will add that while merging.

twalthr:
I will close this PR and include it in the main PR of FLINK-13078.

PR #9021, opened 2019-07-08
[FLINK-13205][runtime] Make checkpoints injection ordered with stop-with-savepoint

1u0:

This pull request changes implementation of "stop-with-savepoint" on `StreamTask` so, that:
 * the `finishTask()` is carried by `notifyCheckpointComplete()` thread, instead of the thread that has triggered checkpoint;
 * the checkpoint injecting thread (in sources) doesn't block (waits for notify-checkpoint);
 * the dispatcher thread in `Task` is reverted to be single threaded, to preserve checkpoints injection ordering.

## Brief change log

  - the checkpoint injection (at stream task sources) made non blocking for "stop-with-savepoint"; 
  - the Task's dispatcher has been reverted to single threaded execution to preserve injection ordering.

## Verifying this change

This change is already covered by existing tests, such as `SourceTaskTerminationTest.java, SynchronousCheckpointITCase.java, TaskAsyncCallTest.java`.

## Documentation


1u0:
CC @pnowojski, @StefanRRichter, @kl0u.

1u0:
I have updated the PR, so that in case of sync checkpoint confirmation, `StreamTask.finishTask()` happens before the waiting `StreamTask.triggerCheckpointOnBarrier()` returns.

StefanRRichter:
Big +1, I think this approach is a lot easier and cleaner 👍  Would still suggest to wait for @kl0u 's opinion...

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SynchronousSavepointLatch.java
```diff
@@ -59,61 +66,53 @@ void setCheckpointId(final long checkpointId) {
 		}
 	}
 
-	boolean blockUntilCheckpointIsAcknowledged() throws Exception {
+	void blockUntilCheckpointIsAcknowledged() throws Exception {
```
StefanRRichter: I would suggest to change the exception types for some methods like this more precisely to `InterruptedException` because it might be easier to see for caller that there might be a special reaction to perform (e.g. restoring the interrupted status).

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SynchronousSavepointLatch.java
```diff
@@ -59,61 +66,53 @@ void setCheckpointId(final long checkpointId) {
...
 		synchronized (synchronizationPoint) {
-			if (isSet() && !isDone()) {
+			if (completionResult == null && isSet()) {
 				waiting = true;
 				synchronizationPoint.wait();
```
StefanRRichter: This line looks broken. What about spurious wakeups or other parties calling `notifyAll` on the object - in particular if this is the checkpoint lock that is to expect.

kl0u:
Thanks for the work @1u0 ! 
I will have a look at this but I believe that this is a big issue that deserves a proper Jira and not a hotfix.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SynchronousSavepointLatch.java
1u0: Thanks @StefanRRichter for feedback. I've adjusted this code to check waiting condition.

zentol:
I agree with @kl0u, if your hotfix needs a review, it's not a hotfix.

1u0:
I can create a Jira ticket and update commit messages and title in this PR to reference it.

The reason it wasn't done so initially, because this whole change was a hotfix commit in [[FLINK-12482]](https://issues.apache.org/jira/browse/FLINK-12482) effort. The latter won't be included in 1.9 release and the issue addressed here is not that critical imo.

1u0:
I think the `finishTask()` doesn't need to be run with acquired `checkpointLock`. The reason is the same as for `cancelTask()` invocation in `cancel()` - currently, both invocations should have the same expectations in regard of holding the lock. So far, the `cancelTask()` call wasn't called under `checkpointLock`.

What I'm not sure about, if the modified `StreamTask.notifyCheckpointComplete()` should be called within `SafetyNetCloseableRegistry` context or with `Task.closeNetworkResources()` call afterwards.

kl0u:
From the code, I am not sure if you need to call any of these methods because after the `finishTask`, the execution will go to the `finally()` blocks of the `StreamTask` and the `Task` accordingly which should take care of cleaning up resources. 

That said, someone with more familiarity with that part of the code could have a more educated opinion.

StefanRRichter:
@kl0u What was even the reason to introduce `finishTask()`? For all production code usages it seems that it does effectively exactly the same as `cancelTask` and is also just used in one place. Should we not rather drop it and call `cancelTask()` instead? 

@1u0 Can you explain a bit more why you are concerned about `SafetyNetCloseableRegistry` and changed `notifyCheckpointComplete` - what's the connection there?

kl0u:
@StefanRRichter The reason for the `finishTask()` (in addition to the `cancelTask()`) is that we should provide methods that allow us to differentiate Flink's behaviour between normal termination and cancellation whenever we can. Not doing so has already led to some problems (e.g. the situation with UDFs not having a proper `close()` method for normal termination). 

The fact that for now the `finishTask()` only calls the cancelTask is coincidental and simply due to the fact that the `cancel()` is the only method in source function interface that allows us to get out of the run-loop. In fact, there is already a PR (https://github.com/apache/flink/pull/9090) which makes the two methods (`finishTask` and `cancelTask`) diverge.

StefanRRichter:
@kl0u ok, that makes sense. Personally, I would then have suggested to introduce `finishTask` only with the changes from #9090 but I guess it's already there now.

@1u0 I think you should be no reason to worry about `SafetyNetCloseableRegistry` IMO.

I think then this can be merged?

kl0u:
Merged on master and release-1.9

PR #9022, opened 2019-07-08
[FLINK-13127] Fix --yarnship classpath

dmvk:

Fix class loading of resources shipped via `--yarnship <files>`. More details in [jira issue](https://issues.apache.org/jira/browse/FLINK-13127).

## Brief change log

  - *AbstractYarnClusterDescriptor* adds directories to classpath instead of files in case it finds non-jar files

## Verifying this change

Tested on private yarn cluster.

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-yarn/src/main/java/org/apache/flink/yarn/AbstractYarnClusterDescriptor.java
+++ b/flink-yarn/src/main/java/org/apache/flink/yarn/AbstractYarnClusterDescriptor.java
@@ -83,7 +83,6 @@ import java.io.ObjectOutputStream;
 import java.io.PrintStream;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
-import java.net.URISyntaxException;
 import java.nio.file.FileVisitResult;
 import java.nio.file.Files;
 import java.nio.file.SimpleFileVisitor;
@@ -1086,8 +1085,7 @@ public abstract class AbstractYarnClusterDescriptor implements ClusterDescriptor
 			Path localSrcPath,
 			Map<String, LocalResource> localResources,
 			Path targetHomeDir,
-			String relativeTargetPath) throws IOException, URISyntaxException {
-
+			String relativeTargetPath) throws IOException {
 		Tuple2<Path, LocalResource> resource = Utils.setupLocalResource(
 			fs,
 			appId.toString(),
@@ -1098,6 +1096,16 @@ public abstract class AbstractYarnClusterDescriptor implements ClusterDescriptor
 		localResources.put(key, resource.f1);
 
 		return resource.f0;
+	}
+
+	/**
+	 * Match file name for "<tt>flink-dist*.jar</tt>" pattern.
+	 *
+	 * @param fileName file name to check
+	 * @return true if file is a dist jar
+	 */
+	private static boolean isDistJar(String fileName) {
+		return fileName.startsWith("flink-dist") && fileName.endsWith("jar");
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


flink-yarn/src/main/java/org/apache/flink/yarn/AbstractYarnClusterDescriptor.java
```diff
@@ -1162,7 +1172,11 @@ public FileVisitResult visitFile(java.nio.file.Path file, BasicFileAttributes at
 									.append(remotePath).append(",");
 
 								// add files to the classpath
-								classPaths.add(key);
+								if (fileName.endsWith("jar")) {
```
TisonKun: point 2 refers these following 5 lines.

flink-yarn/src/main/java/org/apache/flink/yarn/AbstractYarnClusterDescriptor.java
```diff
@@ -1181,11 +1195,21 @@ public FileVisitResult visitFile(java.nio.file.Path file, BasicFileAttributes at
 					envShipFileList.append(key).append("=").append(remotePath).append(",");
 
 					// add files to the classpath
-					classPaths.add(key);
+					if (key.endsWith("jar")) {
```
TisonKun: and these following 5 lines.

It could be `addFilesToClassPath(classPath, shipFile)`

xintongsong:
Hi @dmvk,
Thanks for opening this PR. I think this is a helpful contribution. A few comments:
1. I think it would be better if you could find a committer assigning this jira ticket to you before starting the implementation, so other contributors know you are working on this and will not start a redundant work. @tillrohrmann ,could you help assigning the jira ticket?
2. Agree with @TisonKun, I think this change is not that trivial. It would be better to have a test case to make sure other people won't accidentally break it in the future.
3. Why do we need to separate and sort archives and resources? I mean, I'm total fine with this because it makes the classpath organized and friendly for people to look into. Just curious whether there is any other reason that makes this necessary.

dmvk:
@xintongsong @TisonKun thanks for the review! For points 1) and 2), I've updated a test case and removed duplicated code in `org.apache.flink.yarn.AbstractYarnClusterDescriptor#uploadAndRegisterFiles` method.

I don't think there is any significant benefit of sorting those, the idea was to make sure that it is deterministic, but that would be probably the case without sorting anyway. I don't have any strong arguments for this, so I can remove it if you want to.

Thanks!

flink-yarn/src/test/java/org/apache/flink/yarn/YarnFileStageTest.java
```diff
@@ -177,14 +179,23 @@ static void testCopyFromLocalRecursive(
...
 				localResources,
 				new StringBuilder());
+
+			assertEquals(
+				Arrays.asList(
+					srcDir.getName(),
+					srcDir.getName() + "/nested",
+					srcDir.getName() + "/nested/4",
+					srcDir.getName() + "/test.jar")
+				, classpath);
```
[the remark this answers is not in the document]
TisonKun: FYI, you can make use of `Assert.assertThat` and `Matchers.containsInAnyOrder`.

Here is an example 

```java
Assert.assertThat(new ArrayList<Integer>() {{
			add(1); add(2); add(3);
		}}, Matchers.containsInAnyOrder(3, 2, 1));
```

xintongsong:
Maybe also rebase to the latest master? There is a travis test failure, which is not caused by this PR and should been fixed on master branch.

dmvk:
@xintongsong done ;) thanks

xintongsong:
@dmvk Thanks for addressing the comments, the PR looks good to me.
@tillrohrmann Could you take a final check on this?

dmvk:
@tillrohrmann did you have a time to take a look at this?

tillrohrmann:
I'm looking at it right now.

dmvk:
@tillrohrmann any progress? thanks

tillrohrmann:
The test failure is unrelated. Thanks a lot for the review @TisonKun and @xintongsong.

PR #9027, opened 2019-07-09
[FLINK-13161][table-blink-runtime]numBuckets calculate wrong in BinaryHashBucketArea

AjaxXu:

*fix numBuckets calculate in BinaryHashBucketArea*

## Documentation


flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/runtime/hashtable/BinaryHashBucketArea.java
```diff
@@ -203,7 +203,7 @@ private void resize(boolean spillingAllowed) throws IOException {
 		int oldNumBuckets = numBuckets;
 		MemorySegment[] oldOverflowSegments = overflowSegments;
 		int newNumSegs = oldBuckets.length * 2;
-		int newNumBuckets = MathUtils.roundDownToPowerOf2(newNumSegs << table.bucketsPerSegmentBits);
```
JingsongLi: Add a `checkState(MathUtils.isPowerOf2(oldBuckets.length))`?

TsReaper: No need. This will be checked in the `setNewBuckets` method called below (line 238).

TsReaper: No need. `oldBuckets.length` will always be a power of 2 according to the implementation. Also this will be checked in the `setNewBuckets` method.

JingsongLi: `checkState` is just for safety, `setNewBuckets` may be misused in future, `numBuckets` is power of 2 not mean `buckets.length` is power of 2.
How about just add `checkArgument(MathUtils.isPowerOf2(buckets.length))` in `setNewBuckets`.

AjaxXu: Actually, we make sure the oldBuckets.length is power of 2. So after resize the the length is still be power of 2. And as @TsReaper mentioned, it will be checked later.

AjaxXu: > `checkState` is just for safety, `setNewBuckets` may be misused in future, `numBuckets` is power of 2 not mean `buckets.length` is power of 2.
> How about just add `checkArgument(MathUtils.isPowerOf2(buckets.length))` in `setNewBuckets`.

The advise is more powerful in check the buckets.length. I will change it.

JingsongLi: Because I've considered changing resize strategy, if resize strategy becomes 1.5 times(Just like `JDK ArrayList.grow`) or something in the future, it will cause unexpected problems.

AjaxXu: I had changed the check condition. Is that OK.  But in my option, setting numBuckets to power of 2 is for fast matching. If this strategy is not change. The buckets.length and numBuckets is always be power of 2. Otherwise, it will result in wasting segment.

JingsongLi: Yeah, 
> the buckets.length and numBuckets should be always power of 2, Otherwise, it will result in wasting segment.

So we need add check condition to avoid inefficient or incorrect code. I think check condition is used for this, it is not an expected way.

AjaxXu: Alright! Good for your explain.

JingsongLi:
LGTM +1

AjaxXu:
@JingsongLi @TsReaper Someone can help me to merge this pr? Thanks.

TsReaper:
@wuchong @KurtYoung please take a look. Thanks.

JingsongLi:
@AjaxXu Sorry for late, we are all busy with the release of 1.9. I've told @wuchong that he'll be merge soon.

AjaxXu:
@JingsongLi @TsReaper Thank you guys. Great job you had done.

zentol:
@JingsongLi @TsReaper Do you intend to merge this to1.9?

wuchong:
Hi @zentol , I think this should go into 1.9 too, because this is an important fix.

PR #9030, opened 2019-07-09
[FLINK-13123] Align Stop/Cancel Commands in CLI and REST Interface and Improve Documentation

knaufk:

Currently, the REST API and CLI around stopping and cancelling jobs are not aligned in terms of terminology and the differences between cancel and job are not as clear as they could be.

I would like to make the following changes to the CLI:

    add deprecation warning for cancel -s command and redirect users to stop
    rename -s of stop command to -p for savepoint location. Emphasize that this is optional, as a savepoint is taken in any case

I would like to make the following changes to the REST API:

    Rename stop-with-savepoint to stop
    Rename "endOfEventTime" to "drain" in accordance with the CLI

## Brief change log

* [hotfix] [docs] add missing YARN options in CLI docs
* [cli] add deperecation warning to "cancel -s"
* [cli] rename "-s" parameter of stop command
* [rest] align terminology of "stop" endpoint with cli
* [hotfix] [docs] fix typo in docs/README.md 

## Verifying this change

* run tests in flink-clients, flink-runtime-web and flink-runtime

  - The serializers: no

## Documentation

### DIFF
--- a/docs/ops/cli.md
+++ b/docs/ops/cli.md
@@ -176,23 +176,7 @@ These examples about how to manage a job in CLI.
 
 -   Gracefully stop a job with a savepoint (streaming jobs only):
 
-        ./bin/flink stop -s [targetDirectory] -d <jobID>
-        
-
-**NOTE**: The difference between cancelling and stopping a (streaming) job is the following:
-
-On a cancel call, the operators in a job immediately receive a `cancel()` method call to cancel them as
-soon as possible.
-If operators are not not stopping after the cancel call, Flink will start interrupting the thread periodically
-until it stops.
-
-A "stop" call is a more graceful way of stopping a running streaming job, as the "stop" signal flows from 
-source to sink. When the user requests to stop a job, all sources will be requested to send the last checkpoint barrier
-that will trigger a savepoint, and after the successful completion of that savepoint, they will finish by calling their
-`cancel()` method. If the `-d` flag is specified, then a `MAX_WATERMARK` will be emitted before the last checkpoint 
-barrier. This will result all registered event-time timers to fire, thus flushing out any state that is waiting for 
-a specific watermark, e.g. windows. The job will keep running until all sources properly shut down. This allows the
- job to finish processing all in-flight data. 
+        ./bin/flink stop [-p targetDirectory] [-d] <jobID>
 
 ### Savepoints
 
@@ -220,6 +204,23 @@ If you don't specify a target directory, you need to have [configured a default 
 This will trigger a savepoint for the job with ID `jobId` and YARN application ID `yarnAppId`, and returns the path of the created savepoint.
 
 Everything else is the same as described in the above **Trigger a Savepoint** section.
+
+#### Stop
+
+Use the `stop` to gracefully stop a running streaming job with a savepoint.
+
+{% highlight bash %}
+./bin/flink stop [-p targetDirectory] [-d] <jobID>
+{% endhighlight %}
+
+A "stop" call is a more graceful way of stopping a running streaming job, as the "stop" signal flows from
+source to sink. When the user requests to stop a job, all sources will be requested to send the last checkpoint barrier
+that will trigger a savepoint, and after the successful completion of that savepoint, they will finish by calling their
+`cancel()` method. If the `-d` flag is specified, then a `MAX_WATERMARK` will be emitted before the last checkpoint
+barrier. This will result all registered event-time timers to fire, thus flushing out any state that is waiting for
+a specific watermark, e.g. windows. The job will keep running until all sources properly shut down. This allows the
+ job to finish processing all in-flight data.
+
 
 #### Cancel with a savepoint (deprecated)
 
@@ -431,7 +432,7 @@ Action "stop" stops a running program with a savepoint (streaming jobs only).
   "stop" action options:
      -d,--drain                           Send MAX_WATERMARK before taking the
                                           savepoint and stopping the pipelne.
-     -s,--withSavepoint <withSavepoint>   Path to the savepoint (for example
+     -p,--savepointPath <savepointPath>   Path to the savepoint (for example
                                           hdfs:///flink/savepoint-1537). If no
                                           directory is specified, the configured
                                           default will be used
--- a/docs/ops/cli.md
+++ b/docs/ops/cli.md
@@ -340,6 +340,8 @@ Action "run" compiles and runs a program.
                                           shutdown when the CLI is terminated
                                           abruptly, e.g., in response to a user
                                           interrupt, such as typing Ctrl + C.
+     -yat,--yarnapplicationType <arg>     Set a custom application type for the
+                                          application on YARN
      -yD <property=value>                 use value for given property
      -yd,--yarndetached                   If present, runs the job in detached
                                           mode (deprecated; use non-YARN
@@ -430,6 +432,15 @@ Action "stop" stops a running program with a savepoint (streaming jobs only).
                                           directory is specified, the configured
                                           default will be used
                                           ("state.savepoints.dir").
+  Options for yarn-cluster mode:
+     -m,--jobmanager <arg>            Address of the JobManager (master) to
+                                      which to connect. Use this flag to connect
+                                      to a different JobManager than the one
+                                      specified in the configuration.
+     -yid,--yarnapplicationId <arg>   Attach to running YARN session
+     -z,--zookeeperNamespace <arg>    Namespace to create the Zookeeper
+                                      sub-paths for high availability mode
+
   Options for default mode:
      -m,--jobmanager <arg>           Address of the JobManager (master) to which
                                      to connect. Use this flag to connect to a
--- a/docs/ops/cli.md
+++ b/docs/ops/cli.md
@@ -170,11 +170,11 @@ These examples about how to manage a job in CLI.
 
         ./bin/flink cancel <jobID>
 
--   Cancel a job with a savepoint:
+-   Cancel a job with a savepoint (deprecated; use "stop" instead):
 
         ./bin/flink cancel -s [targetDirectory] <jobID>
 
--   Stop a job with a savepoint (streaming jobs only):
+-   Gracefully stop a job with a savepoint (streaming jobs only):
 
         ./bin/flink stop -s [targetDirectory] -d <jobID>
         
@@ -221,7 +221,7 @@ This will trigger a savepoint for the job with ID `jobId` and YARN application I
 
 Everything else is the same as described in the above **Trigger a Savepoint** section.
 
-#### Cancel with a savepoint
+#### Cancel with a savepoint (deprecated)
 
 You can atomically trigger a savepoint and cancel a job.
 
@@ -232,6 +232,10 @@ You can atomically trigger a savepoint and cancel a job.
 If no savepoint directory is configured, you need to configure a default savepoint directory for the Flink installation (see [Savepoints]({{site.baseurl}}/ops/state/savepoints.html#configuration)).
 
 The job will only be cancelled if the savepoint succeeds.
+
+<p style="border-radius: 5px; padding: 5px" class="bg-danger">
+    <b>Note</b>: Cancelling a job with savepoint is deprecated. Use "stop" instead.</p>
+{% endunless %}
 
 #### Restore a savepoint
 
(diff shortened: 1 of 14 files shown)


docs/ops/cli.md
```diff
@@ -170,29 +170,13 @@ These examples about how to manage a job in CLI.
 
         ./bin/flink cancel <jobID>
 
--   Cancel a job with a savepoint:
+-   Cancel a job with a savepoint (deprecated; use "stop" instead):
 
         ./bin/flink cancel -s [targetDirectory] <jobID>
 
--   Stop a job with a savepoint (streaming jobs only):
+-   Gracefully stop a job (streaming jobs only):
```
kl0u: I would also add "with a savepoint" to make it clear that we always take a savepoint.

docs/ops/cli.md
```diff
@@ -221,6 +205,23 @@ This will trigger a savepoint for the job with ID `jobId` and YARN application I
 
 Everything else is the same as described in the above **Trigger a Savepoint** section.
 
+#### Stop
+
+Use the `stop` to gracefully stop a running streaming job.
```
kl0u: Again, "with a savepoint".

flink-runtime/src/main/java/org/apache/flink/runtime/rest/handler/job/JobCancellationHandler.java
```diff
@@ -83,7 +83,7 @@ public JobCancellationHandler(
 				terminationFuture = gateway.cancelJob(jobId, timeout);
 				break;
 			case STOP:
-				throw new RestHandlerException("The \"stop\" command has been removed. Please use \"stop-with-savepoint\" instead.", HttpResponseStatus.PERMANENT_REDIRECT);
+				throw new RestHandlerException("The \"stop\" command has been removed. Please use \"stop\" instead.", HttpResponseStatus.PERMANENT_REDIRECT);
```
kl0u: The exception message does not make much sense anymore, so it has to be updated.

flink-runtime/src/main/java/org/apache/flink/runtime/rest/messages/job/savepoints/stop/StopWithSavepointRequestBody.java
```diff
@@ -55,6 +55,6 @@ public String getTargetDirectory() {
 	}
 
 	public boolean shouldAdvanceToEndOfEventTime() {
```
kl0u: Maybe update the method name as well to `shouldDrain()` or sth along those lines.

carp84:
Thanks for noticing this and trying to align the commands @knaufk . Not sure whether it's a little bit off-track but as pointed out in FLIP-45, always stop with a savepoint may still be an unexpected behavior to user, since it slows down the job stop procedure and might block launching new jobs when resource is contended, and IMHO we need to fix it. Wdyt?

knaufk:
@zentol Thanks for the review!

@carp84 I think, the problem is that the old `stop` command never really worked. The only two sources implementing the `StoppableFunction` interface were the `NiFiSource` and the `TwitterSource`. Of course, custom sources might have implemented it. Personally, I would be ok with changing this and adding a warning to the release notes and mention to use `cancel` for a quick ungraceful shutdown.

docs/ops/cli.md
aljoscha: it might not be a "savepoint" in the future, i.e. it would stop with a checkpoint (or snapshot) or whatnot.

knaufk: Done

docs/ops/cli.md
knaufk: Done.

flink-runtime/src/main/java/org/apache/flink/runtime/rest/handler/job/JobCancellationHandler.java
knaufk: Updated.

flink-runtime/src/main/java/org/apache/flink/runtime/rest/messages/job/savepoints/stop/StopWithSavepointRequestBody.java
knaufk: Done.

docs/ops/cli.md
knaufk: @aljoscha I went with "savepoint" for now, because that's what it is as of Flink 1.9. "Savepoint" is just in the docs, not in the Public API for the stop command though. 

BTW: another reason not to use `-s` for the `stop` command, although it could also be `snapshot` :)

carp84:
Thanks for the comments and further explanation @zentol @knaufk .

I agree that there were few implementations of `StoppableFunction` in our code base, while besides the possible user-customized implementations, we actually have removed the necessity of having a "stoppable source" and introduced a new way to stop from the source through FLIP-34, that if it's a `CheckpointType.SYNC_SAVEPOINT` or say a synchronous savepoint, the source task will be finished after the savepoint completes (full source code reference [here](https://github.com/apache/flink/blob/28ac5ef7f1769e9256b3089ed4853736771ae457/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java#L808)):
```
if (isRunning && syncSavepointLatch.isSet()) {

        final boolean checkpointWasAcked =
                        syncSavepointLatch.blockUntilCheckpointIsAcknowledged();

        if (checkpointWasAcked) {
                finishTask();
        }
}
```

And I'm proposing to reinforce the job stop semantic in FLIP-45 after which we will support the normal `stop` command again but will always accompanied by a checkpoint (by introducing a new `CheckpointType.SYNC_CHECKPOINT`), and I'd suggest to consider these together if possible, to prevent back and forth changes. Wdyt? Thanks.

aljoscha:
@carp84 has a good point, we should make the cli/REST API future proof, i.e. it should seamlessly accomodate the new stop mode of FLIP-45 (and whatever comes out of FLIP-47 😅 ).

@carp84 Do you have a concrete proposal for the cli/REST API?

carp84:
I suggest to keep the existing command format for `stop` and fix the path w/o `-s` option (the normal stop process) through FLIP-45. Sounds good to you? Thanks. @aljoscha @knaufk

knaufk:
@carp84 That's what the current PR is doing, correct?

The way I envision FLIP-45/47, we will still have the `stop` command afterwards, but maybe with an additional parameter `-f` or so to specify the snapshot format.

docs/ops/cli.md
```diff
@@ -170,29 +170,13 @@ These examples about how to manage a job in CLI.
 
         ./bin/flink cancel <jobID>
 
--   Cancel a job with a savepoint:
+-   Cancel a job with a savepoint (deprecated; use "stop" instead):
```
carp84: suggest to change to `use "stop -s" command instead`

docs/ops/cli.md
```diff
@@ -170,29 +170,13 @@ These examples about how to manage a job in CLI.
...
 
         ./bin/flink cancel -s [targetDirectory] <jobID>
 
--   Stop a job with a savepoint (streaming jobs only):
+-   Gracefully stop a job with a savepoint (streaming jobs only):
```
carp84: suggest to change to `Gracefully stop a job, optionally with a savepoint (streaming jobs only)`

docs/ops/cli.md
```diff
@@ -170,29 +170,13 @@ These examples about how to manage a job in CLI.
...
 
-        ./bin/flink stop -s [targetDirectory] -d <jobID>
-        
-
-**NOTE**: The difference between cancelling and stopping a (streaming) job is the following:
```
carp84: I'd suggest to keep these descriptions and decide how to refine them after FLIP-45

docs/ops/cli.md
```diff
@@ -221,6 +205,23 @@ This will trigger a savepoint for the job with ID `jobId` and YARN application I
...
+A "stop" call is a more graceful way of stopping a running streaming job, as the "stop" signal flows from
+source to sink. When the user requests to stop a job, all sources will be requested to send the last checkpoint barrier
+that will trigger a savepoint, and after the successful completion of that savepoint, they will finish by calling their
+`cancel()` method. If the `-d` flag is specified, then a `MAX_WATERMARK` will be emitted before the last checkpoint
+barrier. This will result all registered event-time timers to fire, thus flushing out any state that is waiting for
+a specific watermark, e.g. windows. The job will keep running until all sources properly shut down. This allows the
+ job to finish processing all in-flight data.
+
+
 #### Cancel with a savepoint
```
carp84: shall we also add a `(deprecated)` annotation here?

docs/ops/cli.zh.md
```diff
@@ -169,29 +169,13 @@ available.
 
         ./bin/flink cancel <jobID>
 
--   Cancel a job with a savepoint:
+-   Cancel a job with a savepoint (deprecated; use "stop" instead):
```
carp84: Same suggestions as for the english docs.

carp84:
And I'd suggest to remove the below description on JIRA:

> Emphasize that this is optional, as a savepoint is taken in any case

docs/ops/cli.md
knaufk: Yes, makes sense.

docs/ops/cli.md
knaufk: It rather keep it as it is. The `stop` command right now always performs a savepoint. It's not optional. 

I understand, that the stop command in the future might or might not take a savepoint, but the documentation should reflect the current state.

docs/ops/cli.md
knaufk: "stop -s" and "stop" both do the same thing. "-s" alone does not do anything, that's why I changed it to "-p" instead to make it clearer that we are actually only passing the savepoint path.

docs/ops/cli.md
knaufk: I've only moved this part.

docs/ops/cli.md
carp84: Makes sense.

docs/ops/cli.md
carp84: I see, but still I'd suggest to explicitly use `stop -p` here so we could prevent change again after supporting normal stop. Wdyt?

docs/ops/cli.md
carp84: Ah yes, makes sense.

docs/ops/cli.md
kl0u: I would say that @carp84 has a point here. Irrespectively of which option we keep, `-p` or `-s`, I think it makes sense here to propose "use `stop -s/p <location>` so that if the semantics of `stop` or `stop -s/p` (with no further argument) change, the user scripts do not.

knaufk: I think at this point it is not yet clear, where we will converge to in term of FLIP-45/47, so we should try to make as little assumptions about the future features as possible while still keeping the API flexible. My basic assumption is that 

* `stop` always indicates a graceful shutdown
* `cancel` indicates an ungraceful, fast shutdown

I understand we are basically discussing the following two options:

Option 1: `stop [-p savepoint-path] [-d]` (`[]` indicates optional arguments)

If we want to support multiple snapshot formats during graceful shutdowns in the future, we evolve it to:  

`stop [-p snapshot-path] [-d] [-f format]`, while the format defaults to a "savepoint format"`

If we also want to support a graceful shutdown via `stop` without a snapshot at all, we could for example add an additional "format" indicating  "no snapshot" or add an additional flag, which is mutually exclusive to `-f`. 

Option 2: `stop [-s] [-p savepoint-path] [-d]`

The `-s` would indicate that a snapshot should be taken. If `stop` is called without `-s`, we throw an exception in Flink 1.9 (we would need to change the current behavior for this).  

The advantage would be that, if we later want to add a `stop` without a snapshot, it would be more natural in terms of the API.


Personally, I don't see the need for a graceful shutdown without some kind of snapshot (yet). So, I think, Option 1 is a better trade off between flexibility and simplicity for current users.

@carp84 @aljoscha @kl0u: What do you think?

kl0u:
In general my comments are covered. So if @carp84 is also satisfied with the latest changes, I can merge.

docs/ops/cli.md
carp84: I totally agree that a graceful shutdown should be accompanied by some kind of "snapshot" and Option 1 is better, but focusing on this line in the doc, I suggest to explicitly indicate that `cancel with savepoint` could only be replaced by `stop with savepoint`, which will be correct no matter where FLIP-45/47 goes.

It's true that in current master `stop` has the same effect with `stop with savepoint`, but later on this may change. If it did change and now we write the doc as `use "stop" instead`, we still need to change it to `use "stop -p" instead` later on, and if we forgot to, this will mislead our users. What do you think? @knaufk

carp84: After some offline discussion, I agree we keep the doc as is and do necessary refinement later when FLIP-45/47 is done.

kl0u:
Merged.

PR #9034, opened 2019-07-09
[FLINK-13063] Temporary fix for AsyncWaitOperator consistency problems.

StefanRRichter:

This PR is a temporary fix for the consistency problems with the `AsyncWaitOperator` that are described under FLINK-13063. This temporary fix works simply by deactivating all chaining for `AsyncWaitOperator`.

## Brief change log

Set the chaining strategy in the constructor of `AsyncWaitOperator` to `NEVER` by default.

## Verifying this change

This change is already covered by existing tests. I added an additional test that chaining is suppressed by default and also check that users could go back to the old (broken) behavior if they had to, by setting the chaining strategy manually on the operator.

  - The serializers: (no)

## Documentation

  - If yes, how is the feature documented? (change and backwards compatibility)

### DIFF
--- a/flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/AsyncWaitOperator.java
+++ b/flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/AsyncWaitOperator.java
@@ -119,9 +119,9 @@ public class AsyncWaitOperator<IN, OUT>
 			AsyncDataStream.OutputMode outputMode) {
 		super(asyncFunction);
 
-		// TODO this is a temporary fix for the problems described under FLINK-13063 at the cost of never chaining
+		// TODO this is a temporary fix for the problems described under FLINK-13063 at the cost of breaking chains for
 		//  AsyncOperators.
-		setChainingStrategy(ChainingStrategy.NEVER);
+		setChainingStrategy(ChainingStrategy.HEAD);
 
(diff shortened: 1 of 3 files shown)


flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/AsyncWaitOperator.java
```diff
@@ -118,7 +118,10 @@ public AsyncWaitOperator(
 			int capacity,
 			AsyncDataStream.OutputMode outputMode) {
 		super(asyncFunction);
-		chainingStrategy = ChainingStrategy.ALWAYS;
+
+		// TODO this is a temporary fix for the problems described under FLINK-13063 at the cost of never chaining
+		//  AsyncOperators.
+		setChainingStrategy(ChainingStrategy.NEVER);
```
pnowojski: Can we make this `HEAD`? I think there are no known issues that are forcing the `AsyncWaitOperator` to be also a tail of an operator chain?

StefanRRichter: Ok, I think that is possible and was just a communication issue about the requirements.

StefanRRichter:
Thanks for the review @pnowojski ! Merging.

PR #9036, opened 2019-07-09
[FLINK-13112][table-planner-blink] Support LocalZonedTimestampType in blink

JingsongLi:

Now we just support TimestampType, it is without time zone.
We need support LocalZonedTimestampType, and define time zone in config.

Now, for the functions of `TIMESTAMP_WITH_LOCAL_TIME_ZONE`:
1.cast: only support cast to and cast from `Char/VarChar/Date/Time/Timestamp`. (Limitation of calcite)
2.Support calcite functions: EXTRACT, FLOOR, CEIL, UNIX_TIMESTAMP, DATE_FORMAT.
3.Not support functions of blink extension.

## Verifying this change

ut

  - The serializers: no

## Documentation

### DIFF
--- a/flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/TableConfig.java
+++ b/flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/TableConfig.java
@@ -24,7 +24,7 @@ import org.apache.flink.configuration.Configuration;
 import org.apache.flink.util.Preconditions;
 
 import java.math.MathContext;
-import java.util.TimeZone;
+import java.time.ZoneId;
 
 /**
  * A config to define the runtime behavior of the Table API.
@@ -33,9 +33,9 @@ import java.util.TimeZone;
 public class TableConfig {
 
 	/**
-	 * Defines the timezone for date/time/timestamp conversions.
+	 * Defines the zone id for timestamp with local time zone.
 	 */
-	private TimeZone timeZone = TimeZone.getTimeZone("UTC");
+	private ZoneId localZoneId = ZoneId.systemDefault();
 
 	/**
 	 * Defines if all fields need to be checked for NULL first.
@@ -104,17 +104,17 @@ public class TableConfig {
 	}
 
 	/**
-	 * Returns the timezone for date/time/timestamp conversions.
+	 * Returns the zone id for timestamp with local time zone.
 	 */
-	public TimeZone getTimeZone() {
-		return timeZone;
+	public ZoneId getLocalTimeZone() {
+		return localZoneId;
 	}
 
 	/**
-	 * Sets the timezone for date/time/timestamp conversions.
+	 * Sets the zone id for timestamp with local time zone.
 	 */
-	public void setTimeZone(TimeZone timeZone) {
-		this.timeZone = Preconditions.checkNotNull(timeZone);
+	public void setLocalTimeZone(ZoneId zoneId) {
+		this.localZoneId = Preconditions.checkNotNull(zoneId);
 	}
 
 	/**
--- a/flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/codegen/CodeGeneratorContext.scala
+++ b/flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/codegen/CodeGeneratorContext.scala
@@ -35,6 +35,8 @@ import org.apache.flink.util.InstantiationUtil
 
 import org.apache.calcite.avatica.util.DateTimeUtils
 
+import java.util.TimeZone
+
 import scala.collection.mutable
 
 /**
@@ -526,7 +528,7 @@ class CodeGeneratorContext(val tableConfig: TableConfig) {
     * Adds a reusable TimeZone to the member area of the generated class.
     */
   def addReusableTimeZone(): String = {
-    val zoneID = tableConfig.getTimeZone.getID
+    val zoneID = TimeZone.getTimeZone(tableConfig.getLocalTimeZone).getID
     val stmt =
       s"""private static final java.util.TimeZone $DEFAULT_TIMEZONE_TERM =
          |                 java.util.TimeZone.getTimeZone("$zoneID");""".stripMargin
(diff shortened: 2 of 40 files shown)


flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/api/TableConfig.scala
```diff
@@ -102,17 +97,19 @@ class TableConfig {
...
 
   /**
-   * Sets the timezone for date/time/timestamp conversions.
-   */
-  def setTimeZone(timeZone: TimeZone): Unit = {
-    require(timeZone != null, "timeZone must not be null.")
-    this.timeZone = timeZone
+    * Sets the timezone for date/time/timestamp conversions.
+    */
+  def setTimeZoneId(zoneId: ZoneId): Unit = {
```
KurtYoung: Don't think we should add to this class, we are about to remove this TableConfig

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/api/ExecutionConfigOptions.java
```diff
@@ -21,13 +21,24 @@
...
 /**
  * This class holds configuration constants used by Flink's table module.
  */
 public class ExecutionConfigOptions {
 
+	// ------------------------------------------------------------------------
+	//  Timestamp Options
+	// ------------------------------------------------------------------------
+	public static final ConfigOption<String> SQL_EXEC_LOCAL_TIME_ZONE_ID =
+			key("sql.exec.local.time-zone-id")
```
KurtYoung: Is this equivalent to TableConfig's setTimeZoneId?

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/TableConfig.java
```diff
@@ -70,20 +64,6 @@
...
-	 * Returns the timezone for date/time/timestamp conversions.
-	 */
-	public TimeZone getTimeZone() {
-		return timeZone;
-	}
-
-	/**
-	 * Sets the timezone for date/time/timestamp conversions.
-	 */
-	public void setTimeZone(TimeZone timeZone) {
```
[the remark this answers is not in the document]
JingsongLi: I'll change it to setZoneId

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/api/ExecutionConfigOptions.java
JingsongLi: In the future, I think we need set it to configuration, for now, I'll remove it.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/api/TableConfig.scala
JingsongLi: Yeah, I'll keep it be consistent with common TableConfig.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/TableConfig.java
```diff
@@ -32,9 +32,9 @@
 public class TableConfig {
 
 	/**
-	 * Defines the timezone for date/time/timestamp conversions.
+	 * Defines the zone id for timestamp with local time zone.
 	 */
-	private TimeZone timeZone = TimeZone.getTimeZone("UTC");
+	private ZoneId zoneId = ZoneId.systemDefault();
```
KurtYoung: Will this break the old behavior?

JingsongLi: This never be used in release-1.8.

KurtYoung: Another concern is it may caused unstable tests if `systemDefault` is based on some system level settings.

JingsongLi: In theory, tests should pass always, if the test not set zone id, the assert should not expect some time zone dependents.

KurtYoung: Is there any bad experience with time zone set to "UTC" by default?

JingsongLi: TimestampWithLocalTimeZone will be same to TimestampWithoutZone if we set to "UTC" by default.
I take a look to hive, spark, they all are default system time zone by default. Oracle maybe same.
I think user want to default system time zone when use TimestampWithLocalTimeZone at the most of the time.

JingsongLi: I'll change to getLocalTimeZone

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/TableConfig.java
```diff
@@ -33,9 +33,9 @@
 public class TableConfig {
 
 	/**
-	 * Defines the timezone for date/time/timestamp conversions.
+	 * Defines the zone id for timestamp with local time zone.
 	 */
-	private TimeZone timeZone = TimeZone.getTimeZone("UTC");
+	private ZoneId localZoneId = ZoneId.systemDefault();
```
KurtYoung: cc @twalthr, we changed `TimeZone` to `LocalZoneId` to better fit with `TIMESTAMP WITH LOCAL TIME ZONE`

KurtYoung:
LGTM, @twalthr , do you have further comments?

JingsongLi:
travis in: https://travis-ci.org/JingsongLi/flink/builds/557200888

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/codegen/CodeGeneratorContext.scala
```diff
@@ -526,14 +528,24 @@ class CodeGeneratorContext(val tableConfig: TableConfig) {
...
       s"""private static final java.util.TimeZone $DEFAULT_TIMEZONE_TERM =
          |                 java.util.TimeZone.getTimeZone("$zoneID");""".stripMargin
     addReusableMember(stmt)
     DEFAULT_TIMEZONE_TERM
   }
 
+  /**
+    * Adds a reusable Time ZoneId to the member area of the generated class.
+    */
+  def addReusableTimeZoneID(): String = {
```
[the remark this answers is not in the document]
JingsongLi: No, I think it should be used, we now use old time zone, we should use zone id instead.

KurtYoung:
travis passed: https://travis-ci.org/JingsongLi/flink/builds/557282177

KurtYoung:
Merging this..

PR #9038, opened 2019-07-09
[FLINK-13169][tests][coordination] IT test for fine-grained recovery (task executor failures)

azagrebin:

The `BatchFineGrainedRecoveryITCase` can be extended with an additional test failure strategy which abruptly shuts down the task executor. This leads to the loss of all previously completed and the in-progress mapper result partitions. The fail-over strategy should restart the current in-progress mapper which will get the `PartitionNotFoundException` because the previous result becomes unavailable and the previous mapper has to be restarted as well. The same should happen subsequently with all previous mappers. When the source is recomputed, all mappers has to be restarted again to recalculate their lost results.

## Brief change log

Extend `BatchFineGrainedRecoveryITCase` with the `RandomTaskExecutorFailureStrategy`.

## Verifying this change

Run `BatchFineGrainedRecoveryITCase`.

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-tests/src/test/java/org/apache/flink/test/recovery/BatchFineGrainedRecoveryITCase.java
+++ b/flink-tests/src/test/java/org/apache/flink/test/recovery/BatchFineGrainedRecoveryITCase.java
@@ -32,7 +32,7 @@ import org.apache.flink.runtime.minicluster.TestingMiniCluster;
 import org.apache.flink.runtime.minicluster.TestingMiniClusterConfiguration.Builder;
 import org.apache.flink.test.util.TestEnvironment;
 import org.apache.flink.util.Collector;
-import org.apache.flink.util.ExceptionUtils;
+import org.apache.flink.util.FlinkException;
 import org.apache.flink.util.FlinkRuntimeException;
 import org.apache.flink.util.TestLogger;
 
@@ -44,13 +44,13 @@ import org.slf4j.LoggerFactory;
 
 import java.io.Serializable;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Random;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 import java.util.stream.LongStream;
@@ -58,6 +58,7 @@ import java.util.stream.LongStream;
 import static org.apache.flink.configuration.JobManagerOptions.FORCE_PARTITION_RELEASE_ON_CONSUMPTION;
 import static org.apache.flink.runtime.executiongraph.failover.FailoverStrategyLoader.PIPELINED_REGION_RESTART_STRATEGY_NAME;
 import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.CoreMatchers.nullValue;
 import static org.junit.Assert.assertThat;
 
 /**
@@ -90,12 +91,29 @@ public class BatchFineGrainedRecoveryITCase extends TestLogger {
 	private static final int EMITTED_RECORD_NUMBER = 1000;
 	private static final int MAP_NUMBER = 3;
 	private static final int MAX_MAP_FAILURES = 4;
-	private static final int MAX_JOB_RESTART_ATTEMPTS = MAP_NUMBER * (MAP_NUMBER + 1) * MAX_MAP_FAILURES / 2;
+
+	/**
+	 * Number of job failures for all mappers due to backtracking when the produced partitions get lost.
+	 *
+	 * <p>Each mapper failure produces number of backtracking failures (partition not found) which is the mapper index + 1,
+	 * because all previous producers have to be restarted and they firstly will not find the previous result.
+	 */
+	private static final int ALL_MAPPERS_BACKTRACK_FAILURES = IntStream.range(0, MAP_NUMBER + 1).sum();
+
+	/**
+	 * Max number of job failures.
+	 *
+	 * <p>For each possible mapper failure, it is all possible backtracking failures plus the generated failure itself.
+	 */
+	private static final int MAX_JOB_RESTART_ATTEMPTS = (ALL_MAPPERS_BACKTRACK_FAILURES + 1) * MAX_MAP_FAILURES;
+
 	private static final String TASK_NAME_PREFIX = "Test partition mapper ";
+
 	private static final List<Long> EXPECTED_JOB_OUTPUT =
 		LongStream.range(MAP_NUMBER, EMITTED_RECORD_NUMBER + MAP_NUMBER).boxed().collect(Collectors.toList());
 
 	private static TestingMiniCluster miniCluster;
+
 	private static AtomicInteger lastTaskManagerIndexInMiniCluster;
 
 	@Before
@@ -147,9 +165,12 @@ public class BatchFineGrainedRecoveryITCase extends TestLogger {
 
 	private static FailureStrategy createFailureStrategy() {
 		CoinToss coin = new CoinToss(1, EMITTED_RECORD_NUMBER);
-		return new JoinedFailureStrategy(
-			new RandomExceptionFailureStrategy(coin),
-			new RandomTaskExecutorFailureStrategy(coin));
+		return new FixedFailureStrategy(
+			1,
+			new LimitedGlobalFailureStrategy(
+				new JoinedFailureStrategy(
+					new RandomExceptionFailureStrategy(coin),
+					new RandomTaskExecutorFailureStrategy(coin))));
 	}
 
 	private static ExecutionEnvironment createExecutionEnvironment() {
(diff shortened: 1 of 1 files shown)


azagrebin:
@flinkbot attention @zentol

flink-tests/src/test/java/org/apache/flink/test/recovery/BatchFineGrainedRecoveryITCase.java
```diff
@@ -101,84 +132,116 @@ public void teardown() throws Exception {
...
-			this.coin = new CoinToss(probFraction, probBase);
+		private AbstractRandomFailureStrategy(CoinToss coin) {
+			this.coin = coin;
 		}
 
 		@Override
-		public void failOrNot() {
-			if (coin.toss() && StaticFailureCounter.failOrNot()) {
-				throw new FlinkRuntimeException("BAGA-BOOM!!! The user function generated test failure.");
+		public boolean failOrNot(int trackingIndex) {
```
zentol: What's the idea behind returning a boolean? The `RandomExceptionFailureStrategy` throws an error right away (), and the other kills the TM which should also result in a new instance being created (after all, there is only 1 TM).

In which cases would the return value actually be used?

EDIT: OK, so this is used in `JoinedFailureStrategy#failOrNot` to prevent multiple strategies from issuing failures. This is a fairly odd way of accomplishing this imo; I would prefer if we would separate the determination of whether we should fail from the actual failure.
Something like `boolean shouldFail()` and `void fail() throws Exception`, which you could then use in the `JoinedFailureStrategy` using `failureStrategies.anyMatch(FS::shouldFail).ifPresent(FS::fail)`.
You can still have a default `failOrNot` implementation that does both at once for easier use in the functions.

flink-tests/src/test/java/org/apache/flink/test/recovery/BatchFineGrainedRecoveryITCase.java
```diff
@@ -101,84 +132,116 @@ public void teardown() throws Exception {
...
 		env.getConfig().setExecutionMode(ExecutionMode.BATCH_FORCED); // forces all partitions to be blocking
 		return env;
 	}
 
-	private enum StaticMapFailureTracker {
-		;
+	@SuppressWarnings({"StaticVariableUsedBeforeInitialization", "OverlyBroadThrowsClause"})
+	private static void restartTaskManager() throws Exception {
+		int tmi = lastTaskManagerIndexInMiniCluster.getAndIncrement();
+		miniCluster.terminateTaskExecutor(tmi).get();
```
zentol: This could lead to a deadlock once the TE is waiting for the task termination future. The task thread tries to kill the TE and blocks, while the TE waits for the task to shutdown and blocks as well.

flink-tests/src/test/java/org/apache/flink/test/recovery/BatchFineGrainedRecoveryITCase.java
```diff
@@ -101,84 +132,116 @@ public void teardown() throws Exception {
...
+
+		@Override
+		void fail(int trackingIndex) {
+			StaticMapFailureTracker.mapFailureWithBacktracking(trackingIndex);
+			try {
+				restartTaskManager();
+			} catch (InterruptedException e) {
+				Thread.currentThread().interrupt();
+			} catch (Exception e) {
+				ExceptionUtils.rethrow(e);
```
zentol: If the shutdown of the TE fails we likely want to forward this exception directly to the test. As it stands we're just gonna restart the calling task, which may or may not result in a test failure (depending on whether the kill call went through). In any case it likely obfuscates things a bit.

flink-tests/src/test/java/org/apache/flink/test/recovery/BatchFineGrainedRecoveryITCase.java
```diff
@@ -59,19 +68,35 @@
...
+ * </ul>
  */
 public class BatchFineGrainedRecoveryITCase extends TestLogger {
+	private static final Logger LOG = LoggerFactory.getLogger(BatchFineGrainedRecoveryITCase.class);
+
 	private static final int EMITTED_RECORD_NUMBER = 1000;
-	private static final int MAX_FAILURE_NUMBER = 10;
 	private static final int MAP_NUMBER = 3;
+	private static final int MAX_MAP_FAILURES = 4;
+	private static final int MAX_JOB_RESTART_ATTEMPTS = MAP_NUMBER * (MAP_NUMBER + 1) * MAX_MAP_FAILURES / 2;
```
zentol: please add a comment for how you arrived at this formula.

flink-tests/src/test/java/org/apache/flink/test/recovery/BatchFineGrainedRecoveryITCase.java
```diff
@@ -81,13 +106,19 @@ public void setup() throws Exception {
 
 		miniCluster = new TestingMiniCluster(
 			new Builder()
-				.setNumTaskManagers(MAP_NUMBER)
+				.setNumTaskManagers(1)
```
zentol: won't this always result in a full restart?

flink-tests/src/test/java/org/apache/flink/test/recovery/BatchFineGrainedRecoveryITCase.java
azagrebin: I think we do not need 2 methods at the moment, one expresses the same thing.
I will add `throws Exception` to `failOrNot` and explain that it can fail either by throwing an exception or implicitly later and return `true` (TM failure).

flink-tests/src/test/java/org/apache/flink/test/recovery/BatchFineGrainedRecoveryITCase.java
azagrebin: The task should get interrupted at some point in worst case and the `InterruptedException` is ignored in `RandomTaskExecutorFailureStrategy.fail` but JM usually get informed about TM shutdown earlier and fails the task and the job.

flink-tests/src/test/java/org/apache/flink/test/recovery/BatchFineGrainedRecoveryITCase.java
azagrebin: True, I will add tracking and verification of unexpected failures in `StaticMapFailureTracker`.

flink-tests/src/test/java/org/apache/flink/test/recovery/BatchFineGrainedRecoveryITCase.java
azagrebin: Yes, it will. The idea is to test the backtracking with failing the finished producers. The test with loosing only some partitions will be more complicated but more general, of course. We could do it as a followup. Although I am not sure about its value atm.

flink-tests/src/test/java/org/apache/flink/test/recovery/BatchFineGrainedRecoveryITCase.java
```diff
@@ -199,6 +371,76 @@ private boolean toss() {
 		}
 	}
 
+	private enum StaticMapFailureTracker {
+		;
+
+		private static final List<AtomicInteger> mapRestarts = new ArrayList<>(MAP_NUMBER);
+		private static final List<AtomicInteger> expectedMapRestarts = new ArrayList<>(MAP_NUMBER);
+		private static final List<AtomicInteger> mapFailures = new ArrayList<>(MAP_NUMBER);
```
zentol: mapFailures and mapFailuresWithBacktracking  are not used in any assertion

flink-tests/src/test/java/org/apache/flink/test/recovery/BatchFineGrainedRecoveryITCase.java
```diff
@@ -199,6 +371,76 @@ private boolean toss() {
...
+
+		private static void mapFailureWithBacktracking(int index) {
+			mapFailuresWithBacktracking.get(index).incrementAndGet();
+			IntStream.range(0, index + 1).forEach(i -> expectedMapRestarts.get(i).incrementAndGet());
+			IntStream.range(0, index + 1).forEach(i -> expectedMapRestarts.get(i).incrementAndGet());
+		}
+
+		private static void unrelatedFailure(Throwable failure) {
+			unrelatedFailure.getAndUpdate(t -> {
+				// wrap to avoid side effects in the atomic updater (multiple failure.addSuppressed calls)
```
zentol: instead of potentially wrapping this multiple times a simple `synchronized` block may be a better option

flink-tests/src/test/java/org/apache/flink/test/recovery/BatchFineGrainedRecoveryITCase.java
```diff
@@ -199,6 +371,76 @@ private boolean toss() {
...
+
+		private static void mapFailure(int index) {
+			mapFailures.get(index).incrementAndGet();
+			expectedMapRestarts.get(index).incrementAndGet();
+		}
+
+		private static void mapFailureWithBacktracking(int index) {
+			mapFailuresWithBacktracking.get(index).incrementAndGet();
+			IntStream.range(0, index + 1).forEach(i -> expectedMapRestarts.get(i).incrementAndGet());
+			IntStream.range(0, index + 1).forEach(i -> expectedMapRestarts.get(i).incrementAndGet());
```
zentol: why are we doing this twice?

flink-tests/src/test/java/org/apache/flink/test/recovery/BatchFineGrainedRecoveryITCase.java
```diff
@@ -101,84 +150,207 @@ public void teardown() throws Exception {
...
+		public boolean failOrNot(int trackingIndex) throws Exception {
+			if (failureNumber < maxFailureNumber) {
+				try {
+					boolean failed = wrappedFailureStrategy.failOrNot(trackingIndex);
+					if (failed) {
+						failureNumber++;
+					}
+					return failed;
+				} catch (Exception e) {
+					failureNumber++;
```
zentol: what's the benefit of incrementing the count when an exception is thrown? This RS will never be queried again since on the next attempt a fresh instance is used.

My feeling is you could simplify this whole limiting logic by enforcing exactly 1 failure for each mapper across the entire test, which you could trivially track via the static failure counter.

flink-tests/src/test/java/org/apache/flink/test/recovery/BatchFineGrainedRecoveryITCase.java
azagrebin: I wanted to keep them for logs and debuging

flink-tests/src/test/java/org/apache/flink/test/recovery/BatchFineGrainedRecoveryITCase.java
azagrebin: once restart for not finding the partition and one for actual run after the consumed partition is recomputed but true I will use `addAndGet(2)`

flink-tests/src/test/java/org/apache/flink/test/recovery/BatchFineGrainedRecoveryITCase.java
```diff
@@ -101,101 +166,276 @@ public void teardown() throws Exception {
...
 		StaticMapFailureTracker.verify();
 	}
 
-	private ExecutionEnvironment createExecutionEnvironment() {
+	private static FailureStrategy createFailureStrategy() {
+		int failWithExceptionAfterNumberOfProcessedRecords = rnd.nextInt(EMITTED_RECORD_NUMBER) + 1;
+		int failTaskExecutorAfterNumberOfProcessedRecords = rnd.nextInt(EMITTED_RECORD_NUMBER) + 1;
+		// it has to fail only once during one mapper run so that different failure strategies do not mess up each other stats
+		FailureStrategy failureStrategy = new OneTimeFailureStrategy(
+			new JoinedFailureStrategy(
```
zentol: could be further simplified by determining the strategy beforehand for each mapper separately at random and only passing that strategy. We would then have a fully determined test scenario before executing the test; which is functionally identical to the current state but (hopefully) easier to reproduce.
You then also wouldn't need a List<Map> in the failure tracker; it could just be a list, since they could re-use the trackingIndex. The equals/hashcode and UUID in the failure strategies could then be removed.

flink-tests/src/test/java/org/apache/flink/test/recovery/BatchFineGrainedRecoveryITCase.java
```diff
@@ -101,101 +166,276 @@ public void teardown() throws Exception {
...
+				new GloballyTrackingFailureStrategy(
```
zentol: if you accept the above suggestion, then I think we could merge this strategy into the `OneTimeFailureStrategy`.

flink-tests/src/test/java/org/apache/flink/test/recovery/BatchFineGrainedRecoveryITCase.java
```diff
@@ -101,101 +166,276 @@ public void teardown() throws Exception {
...
+
+		private static int addNewMap() {
+			mapRestarts.add(new AtomicInteger(0));
+			mapFailures.add(new HashMap<>(2));
+			return mapRestarts.size() - 1;
+		}
+
+		private static boolean failOrNot(int index, FailureStrategy failureStrategy) throws Exception {
+			Boolean prevFailed = mapFailures.get(index).get(failureStrategy);
+			boolean alreadyFailed = prevFailed != null && prevFailed;
```
zentol: could be simplified using `Map#getOrDefault`

flink-tests/src/test/java/org/apache/flink/test/recovery/BatchFineGrainedRecoveryITCase.java
```diff
@@ -101,101 +166,276 @@ public void teardown() throws Exception {
...
+			return id.hashCode();
+		}
+	}
+
+	@SuppressWarnings("SynchronizationOnStaticField")
+	private enum StaticMapFailureTracker {
+		;
+
+		private static final List<AtomicInteger> mapRestarts = new ArrayList<>(MAP_NUMBER);
+		private static final List<Map<FailureStrategy, Boolean>> mapFailures = new ArrayList<>(MAP_NUMBER);
```
zentol: instead of a map you could use a set, and only add the strategy if it has failed.

flink-tests/src/test/java/org/apache/flink/test/recovery/BatchFineGrainedRecoveryITCase.java
```diff
@@ -101,101 +166,276 @@ public void teardown() throws Exception {
...
+				unrelatedFailure = ExceptionUtils.firstOrSuppressed(failure, unrelatedFailure);
+			}
 		}
 
-		private boolean toss() {
-			int prob = rnd.nextInt(probBase) + 1;
-			return prob <= probFraction;
+		private static void verify() {
+			synchronized (classLock) {
+				assertThat(unrelatedFailure, is(nullValue()));
```
zentol: Id prefer throwing the contained exception; i.e.
```
if (throwable != null) {
	throw new AssertionError("Test failed due to unexpected exception.", throwable);
}
```

flink-tests/src/test/java/org/apache/flink/test/recovery/BatchFineGrainedRecoveryITCase.java
```diff
@@ -332,112 +338,104 @@ void fail(int trackingIndex) throws Exception {
...
-		private boolean toss() {
-			int prob = rnd.nextInt(probBase) + 1;
-			return prob <= probFraction;
+		@Override
+		public int hashCode() {
+			return id.hashCode();
 		}
 	}
 
+	@SuppressWarnings("SynchronizationOnStaticField")
```
zentol: instead of having this class be full of static fields, we could also turn it into a "normal" class and add a static instance in the test that we reference.

So this class would then be:
```
private static class StaticMapFailureTracker {
	private final List<AtomicInteger> mapRestarts = new ArrayList<>(MAP_NUMBER);
...
```

and on `@Before` we setup the instance.  This would make it easier to re-use and you no longer need the reset calls. You could then further simplify it by setting up the data-structures in the constructor.

zentol: th suppression can be removed now.

azagrebin:
Thanks for the review @zentol
CI fails with a known instability
Mine is green: https://travis-ci.org/azagrebin/flink/builds/561012414

PR #9039, opened 2019-07-09
[FLINK-13170][table-planner] Planner should get table factory from ca…

lirui-apache:
…talog when creating sink for CatalogTable

Planner should first try getting table factory from catalog when creating table sinks for `CatalogTable`.

## Brief change log

  - Make planner use `TableSinkFactory` returned by `Catalog::getTableFactory`.

## Verifying this change

Existing tests.

  - The serializers: no

## Documentation

### DIFF
--- a/flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/PlannerBase.scala
+++ b/flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/PlannerBase.scala
@@ -248,7 +248,7 @@ abstract class PlannerBase(
 
         val catalog = catalogManager.getCatalog(s.getTablePath.get(0))
         val catalogTable = s.getCatalogTable.get().asInstanceOf[CatalogTable]
-        if (catalog.isPresent) {
+        if (catalog.isPresent && catalog.get().getTableFactory.isPresent) {
           val dbName = s.getTablePath.get(1)
           val tableName = s.getTablePath.get(2)
           val sink = TableFactoryUtil.createTableSinkForCatalogTable(
--- a/flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/PlannerBase.scala
+++ b/flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/PlannerBase.scala
@@ -24,7 +24,7 @@ import org.apache.flink.configuration.Configuration
 import org.apache.flink.streaming.api.environment.StreamExecutionEnvironment
 import org.apache.flink.table.api.{TableConfig, TableEnvironment, TableException}
 import org.apache.flink.table.calcite.{FlinkPlannerImpl, FlinkRelBuilder, FlinkTypeFactory}
-import org.apache.flink.table.catalog.{CatalogManager, CatalogManagerCalciteSchema, CatalogTable, ConnectorCatalogTable, FunctionCatalog}
+import org.apache.flink.table.catalog.{CatalogManager, CatalogManagerCalciteSchema, CatalogTable, ConnectorCatalogTable, FunctionCatalog, ObjectPath}
 import org.apache.flink.table.delegation.{Executor, Planner}
 import org.apache.flink.table.executor.ExecutorBase
 import org.apache.flink.table.factories.{TableFactoryService, TableFactoryUtil, TableSinkFactory}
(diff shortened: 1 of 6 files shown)


lirui-apache:
cc @xuefuz @bowenli86 @zjuwangg

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/PlannerBase.scala
```diff
@@ -241,7 +241,20 @@ abstract class PlannerBase(
       case Some(s) if JavaScalaConversionUtil.toScala(s.getCatalogTable)
         .exists(_.isInstanceOf[CatalogTable]) =>
 
-        val sinkProperties = s.getCatalogTable.get().asInstanceOf[CatalogTable].toProperties
+        val catalog = catalogManager.getCatalog(s.getTablePath.get(0))
+        val catalogTable = s.getCatalogTable.get().asInstanceOf[CatalogTable]
+        if (catalog.isPresent && catalog.get().getTableFactory.isPresent) {
```
bowenli86: seems to me `catalog` is always present? otherwise line 244 will throw exceptions

bowenli86:
can we add some test coverage?

bowenli86:
@flinkbot attention @KurtYoung @JingsongLi @twalthr @dawidwys  

Can you guys help to take a look?

KurtYoung:
@godfreyhe please take a look at this

lirui-apache:
Moved the logic to a util method and updated some tests to make sure the change is covered.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/PlannerBase.scala
```diff
@@ -241,7 +241,18 @@ abstract class PlannerBase(
       case Some(s) if JavaScalaConversionUtil.toScala(s.getCatalogTable)
         .exists(_.isInstanceOf[CatalogTable]) =>
 
-        val sinkProperties = s.getCatalogTable.get().asInstanceOf[CatalogTable].toProperties
+        val catalog = catalogManager.getCatalog(s.getTablePath.get(0))
+        val catalogTable = s.getCatalogTable.get().asInstanceOf[CatalogTable]
+        if (catalog.isPresent) {
```
bowenli86: minor: we can check if `catalog.getTableFactory().isPresent()` here. If it's not, we don't even need to go into the if clause

bowenli86:
@flinkbot attention @godfreyhe

godfreyhe:
thanks for this pr @lirui-apache. `StreamPlanner` in flink-table-planner module should also be updated, just as `PlannerBase` in flink-table-planner-blink module. and add some tests for also for blink planner?

lirui-apache:
@godfreyhe thanks for your suggestions. I have updated `StreamPlanner ` and added a test for blink planner in `SinkTest`. Please have a look.

danny0405:
Thanks @lirui-apache , overall looks good to me, just reminder that the `TableEnvironment#sql` has been reverted, and we should still use `sqlQuery` for select and `sqlUpdate` for insert into and ddl.

lirui-apache:
Thanks @danny0405 , I've updated accordingly.
@godfreyhe @bowenli86 please take another look, thanks.

flink-table/flink-table-planner-blink/src/test/scala/org/apache/flink/table/plan/batch/sql/SinkTest.scala
```diff
@@ -60,4 +67,25 @@ class SinkTest extends TableTestBase {
     util.verifyPlan()
   }
 
+  @Test
+  def testCatalogTableSink(): Unit = {
```
godfreyhe: actually, I find that the following simplified test will throw `NoMatchingTableFactoryException` before this PR, so I think this test case can be simplified.
```
  @Test
  def testCatalogTableSink(): Unit = {
    val schemaBuilder = new TableSchema.Builder()
    schemaBuilder.fields(Array("i"), Array(DataTypes.INT()))
    val schema = schemaBuilder.build()
    val catalog = util.tableEnv.getCatalog(util.tableEnv.getCurrentCatalog).get()
    val catalogTable = new CatalogTableImpl(schema, Map[String, String]().asJava, "")
    catalog.createTable(new ObjectPath("default_database", "tbl"), catalogTable, false)
    util.tableEnv.sqlQuery("select 1").insertInto("tbl")
    util.tableEnv.explain(false)
  }
```

lirui-apache: It seems this test case fails even with this PR. Because `GenericInMemoryCatalog::getTableFactory` returns Optional.empty(). So it falls back to the service discovery. Could you please verify on your side?

godfreyhe: oops.., i'm wrong before, i debug the case and stop at line 253 (`if (sink.isPresent) {`), i thought it could be succeed. 😂😂 

could you update `Mockito.atLeast(1)` to `Mockito.atLeast(2)` ? the first time is `    Mockito.when[Optional[_]](catalog.getTableFactory).thenReturn(Optional.of(factory))
` and another is in `getTableSink` method.

lirui-apache: I did a little bit test and seems `Mockito.when(...)` doesn't count as an invocation. The `getTableFactory` is called twice because we have `catalog.get().getTableFactory.isPresent` in `PlannerBase.getTableSink` and `catalog.getTableFactory().orElse(null)` in `TableFactoryUtil.createTableSinkForCatalogTable`. From a UT's perspective, these are implementation details. So I prefer not to verify "atLeast(2)". Does that make sense to you?

godfreyhe:
LGTM, +1 to merge

lirui-apache:
Couldn't reproduce the test failure. Rebase to try again.

bowenli86:
I don't think the StandaloneResourceManagerTest test failure is related. It's reported in https://issues.apache.org/jira/browse/FLINK-13242

merging

PR #9045, opened 2019-07-09
[FLINK-12955][hbase] Support LookupableTableSource for HBase

wuchong:

Support LookupableTableSource for HBase.

## Brief change log

- Introduce `HBaseLookupFunction` to support lookup
- Make `HBaseTableSource` supports `LookupableTableSource`.

## Verifying this change

- Added integration tests for lookup function

## Documentation

### DIFF
--- a/flink-connectors/flink-hbase/src/main/java/org/apache/flink/addons/hbase/HBaseLookupFunction.java
+++ b/flink-connectors/flink-hbase/src/main/java/org/apache/flink/addons/hbase/HBaseLookupFunction.java
@@ -1,0 +1,150 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.addons.hbase;
+
+import org.apache.flink.addons.hbase.util.HBaseConfigurationUtil;
+import org.apache.flink.addons.hbase.util.HBaseReadHelper;
+import org.apache.flink.annotation.VisibleForTesting;
+import org.apache.flink.api.common.typeinfo.TypeInformation;
+import org.apache.flink.table.functions.FunctionContext;
+import org.apache.flink.table.functions.TableFunction;
+import org.apache.flink.types.Row;
+import org.apache.flink.util.StringUtils;
+
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.hbase.HBaseConfiguration;
+import org.apache.hadoop.hbase.HConstants;
+import org.apache.hadoop.hbase.TableName;
+import org.apache.hadoop.hbase.TableNotFoundException;
+import org.apache.hadoop.hbase.client.Connection;
+import org.apache.hadoop.hbase.client.ConnectionFactory;
+import org.apache.hadoop.hbase.client.HTable;
+import org.apache.hadoop.hbase.client.Result;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+
+/**
+ * The HBaseLookupFunction is a standard user-defined table function, it can be used in tableAPI
+ * and also useful for temporal table join plan in SQL.
+ */
+public class HBaseLookupFunction extends TableFunction<Row> {
+	private static final Logger LOG = LoggerFactory.getLogger(HBaseLookupFunction.class);
+	private static final long serialVersionUID = 1L;
+
+	private final String hTableName;
+	private final byte[] serializedConfig;
+	private final HBaseTableSchema hbaseTableSchema;
+
+	private transient HBaseReadHelper readHelper;
+	private transient Connection hConnection;
+	private transient HTable table;
+
+	public HBaseLookupFunction(
+			Configuration configuration,
+			String hTableName,
+			HBaseTableSchema hbaseTableSchema) {
+		this.serializedConfig = HBaseConfigurationUtil.serializeConfiguration(configuration);
+		this.hTableName = hTableName;
+		this.hbaseTableSchema = hbaseTableSchema;
+	}
+
+	/**
+	 * The invoke entry point of lookup function.
+	 * @param rowKey the lookup key. Currently only support single rowkey.
+	 */
+	public void eval(Object rowKey) throws IOException {
+		// fetch result
+		Result result = table.get(readHelper.createGet(rowKey));
+		if (!result.isEmpty()) {
+			// parse and collect
+			collect(readHelper.parseToRow(result, rowKey));
+		}
+	}
+
+	@Override
+	public TypeInformation<Row> getResultType() {
+		return hbaseTableSchema.convertsToTableSchema().toRowType();
+	}
+
+	private org.apache.hadoop.conf.Configuration prepareRuntimeConfiguration() {
+		// create default configuration from current runtime env (`hbase-site.xml` in classpath) first,
+		// and overwrite configuration using serialized configuration from client-side env (`hbase-site.xml` in classpath).
+		// user params from client-side have the highest priority
+		org.apache.hadoop.conf.Configuration runtimeConfig = HBaseConfigurationUtil.deserializeConfiguration(
+			serializedConfig,
+			HBaseConfiguration.create());
+
+		// do validation: check key option(s) in final runtime configuration
+		if (StringUtils.isNullOrWhitespaceOnly(runtimeConfig.get(HConstants.ZOOKEEPER_QUORUM))) {
+			LOG.error("can not connect to HBase without {} configuration", HConstants.ZOOKEEPER_QUORUM);
+			throw new IllegalArgumentException("check HBase configuration failed, lost: '" + HConstants.ZOOKEEPER_QUORUM + "'!");
+		}
+
+		return runtimeConfig;
+	}
+
+	@Override
+	public void open(FunctionContext context) {
+		LOG.info("start open ...");
+		org.apache.hadoop.conf.Configuration config = prepareRuntimeConfiguration();
+		try {
+			hConnection = ConnectionFactory.createConnection(config);
+			table = (HTable) hConnection.getTable(TableName.valueOf(hTableName));
+		} catch (TableNotFoundException tnfe) {
+			LOG.error("Table '{}' not found ", hTableName, tnfe);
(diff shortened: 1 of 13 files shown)


flink-connectors/flink-hbase/src/main/java/org/apache/flink/addons/hbase/HBaseLookupFunction.java
```diff
@@ -0,0 +1,150 @@
...
+
+	private final String hTableName;
+	private final byte[] serializedConfig;
+	private final HBaseTableSchema hbaseTableSchema;
+
+	private transient HBaseReadHelper readHelper;
+	private transient Connection hConnection;
+	private transient HTable table;
+
+	public HBaseLookupFunction(
```
carp84: Could be package private, no need to be public.

flink-connectors/flink-hbase/src/main/java/org/apache/flink/addons/hbase/HBaseLookupFunction.java
```diff
@@ -0,0 +1,150 @@
...
+		this.serializedConfig = HBaseConfigurationUtil.serializeConfiguration(configuration);
+		this.hTableName = hTableName;
+		this.hbaseTableSchema = hbaseTableSchema;
+	}
+
+	/**
+	 * The invoke entry point of lookup function.
+	 * @param rowKey the lookup key. Currently only support single rowkey.
+	 */
+	public void eval(Object rowKey) throws IOException {
```
carp84: This method is never used. Is it some reserved method for future usage?

flink-connectors/flink-hbase/src/main/java/org/apache/flink/addons/hbase/HBaseLookupFunction.java
```diff
@@ -0,0 +1,150 @@
...
+			collect(readHelper.parseToRow(result, rowKey));
+		}
+	}
+
+	@Override
+	public TypeInformation<Row> getResultType() {
+		return hbaseTableSchema.convertsToTableSchema().toRowType();
+	}
+
+	private org.apache.hadoop.conf.Configuration prepareRuntimeConfiguration() throws IOException {
```
carp84: No actual exception thrown, consider remove the `throws` declaration from method signature.

flink-connectors/flink-hbase/src/main/java/org/apache/flink/addons/hbase/HBaseRowInputFormat.java
```diff
@@ -53,78 +47,33 @@
...
+	private final HBaseTableSchema schema;
 
-	// row which is returned
-	private Row resultRow;
-	// nested family rows
-	private Row[] familyRows;
+	private transient org.apache.hadoop.conf.Configuration conf;
+	private transient HBaseReadHelper readHelper;
 
 	public HBaseRowInputFormat(org.apache.hadoop.conf.Configuration conf, String tableName, HBaseTableSchema schema) {
```
carp84: Seems no need to be public.

flink-connectors/flink-hbase/src/main/java/org/apache/flink/addons/hbase/HBaseTableSchema.java
```diff
@@ -45,13 +54,13 @@
 	 * @param qualifier the qualifier name
 	 * @param clazz     the data type of the qualifier
 	 */
-	void addColumn(String family, String qualifier, Class<?> clazz) {
+	public void addColumn(String family, String qualifier, Class<?> clazz) {
```
carp84: Why do we need to make the method public? The same question applies to many of the later methods in this class.

flink-connectors/flink-hbase/src/main/java/org/apache/flink/addons/hbase/HBaseTableFactory.java
```diff
@@ -0,0 +1,121 @@
...
+				hbaseSchema.setRowKey(name, type.getTypeClass());
+			}
+		}
+		return hbaseSchema;
+	}
+
+	private DescriptorProperties getValidatedProperties(Map<String, String> properties) {
+		final DescriptorProperties descriptorProperties = new DescriptorProperties(true);
+		descriptorProperties.putProperties(properties);
+		new HBaseValidator().validate(descriptorProperties);
```
carp84: Could we introduce a singleton instance inside `HBaseValidator` instead of create a new one in each invocation?

flink-connectors/flink-hbase/src/main/java/org/apache/flink/addons/hbase/util/HBaseReadHelper.java
```diff
@@ -0,0 +1,166 @@
...
+		}
+		return get;
+	}
+
+	/**
+	 * Returns an instance of Scan that retrieves the required subset of records from the HBase table.
+	 *
+	 * @return The appropriate instance of Scan for this use case.
+	 */
+	public Scan createScanner() {
```
carp84: Consider change the method name to `createScan` to prevent misunderstanding. scan and scanner are two different things in hbase (refer to `HTable.getScanner` for details).

flink-connectors/flink-hbase/src/main/java/org/apache/flink/addons/hbase/HBaseTableSchema.java
```diff
@@ -64,12 +73,31 @@ void addColumn(String family, String qualifier, Class<?> clazz) {
 		familyMap.put(family, qualifierMap);
 	}
 
+	/**
+	 * Sets row key information in the table schema.
+	 * @param rowKeyName the row key field name
+	 * @param clazz the data type of the row key
+	 */
+	public void setRowKey(String rowKeyName, Class<?> clazz) {
```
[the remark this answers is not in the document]
wuchong: This method is used in `HBaseTableSource` and `HBaseTableFactory`.  But I can make the methods as package-private.

flink-connectors/flink-hbase/src/main/java/org/apache/flink/addons/hbase/HBaseTableSource.java
```diff
@@ -90,7 +94,16 @@ private HBaseTableSource(Configuration conf, String tableName, TableSchema table
...
-		this.hBaseSchema.addColumn(family, qualifier, clazz);
+		this.hbaseSchema.addColumn(family, qualifier, clazz);
+	}
+
+	/**
+	 * Sets row key information in the table schema.
+	 * @param rowKeyName the row key field name
+	 * @param clazz the data type of the row key
+	 */
+	public void setRowKey(String rowKeyName, Class<?> clazz) {
```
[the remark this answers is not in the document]
wuchong: This method is a public API to let users have a rowkey in schema.

flink-connectors/flink-hbase/src/main/java/org/apache/flink/addons/hbase/HBaseValidator.java
```diff
@@ -0,0 +1,53 @@
...
+import java.util.List;
+
+/**
+ * The validator for HBase.
+ * More features to be supported, e.g., batch read/write, async api(support from hbase version 2.0.0), Caching for LookupFunction.
+ */
+@Internal
+public class HBaseValidator extends ConnectorDescriptorValidator {
+
+	public static final String CONNECTOR_TYPE_VALUE_HBASE = "hbase";
```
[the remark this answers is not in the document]
wuchong: I just know all the existing connector validators (e.g. KafkaValidator, CsvValidator) are all placed in `org.apache.flink.table.descriptors` package. This is not mandatory, but is suggested to do. So I will move this file to `org.apache.flink.table.descriptors`, and then the static field have to be public.

flink-connectors/flink-hbase/src/main/java/org/apache/flink/addons/hbase/util/HBaseTypeUtils.java
```diff
@@ -0,0 +1,174 @@
...
+ */
+@Internal
+public class HBaseTypeUtils {
+
+	private static final byte[] EMPTY_BYTES = new byte[]{};
+
+	/**
+	 * Deserialize byte array to Java Object with the given type.
+	 */
+	public static Object deserializeToObject(byte[] value, int typeIdx, Charset stringCharset) {
```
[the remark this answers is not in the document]
wuchong: I think methods in utility should be public (otherwise why we put them in a Utility class?), even though they are invoked in the same package. In the other hand, they may used across package in the future when we introduce hbase sink.

flink-connectors/flink-hbase/src/main/java/org/apache/flink/addons/hbase/util/HBaseTypeUtils.java
```diff
@@ -0,0 +1,174 @@
...
+	 * Checks whether the given Class is a supported type in HBase connector.
+	 */
+	public static boolean isSupportedType(Class<?> clazz) {
+		return getTypeIndex(clazz) != -1;
+	}
+
+	/**
+	 * Checks whether the given {@link TypeInformation} is a supported type in HBase connector.
+	 */
+	public static boolean isSupportedType(TypeInformation typeInfo) {
```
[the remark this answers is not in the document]
wuchong: Removed.

flink-connectors/flink-hbase/src/main/java/org/apache/flink/addons/hbase/HBaseTableFactory.java
wuchong: I think it's a good idea. However, all the existing validator keeps this style, for example: KafkaValidator, CsvValidator, etc... So I prefer to follow this for now.

wuchong:
Hi @carp84 , thanks for reviewing. Comments are addressed.

flink-connectors/flink-hbase/src/main/java/org/apache/flink/addons/hbase/HBaseRowInputFormat.java
wuchong: `HBaseRowInputFormat` is a public API to let users construct an `InputFormat` for HBase. They can be used in DataStream or DataSet. 

And it was public before.

flink-connectors/flink-hbase/src/main/java/org/apache/flink/addons/hbase/HBaseLookupFunction.java
wuchong: This is intended. `HBaseLookupFunction` is a public API to let users construct it and use it in Table API directly via `env.registerFunction(TableFunction)`.

flink-connectors/flink-hbase/src/main/java/org/apache/flink/addons/hbase/util/HBaseTypeUtils.java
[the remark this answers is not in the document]
carp84: Not really, since it's under the `org.apache.flink.addons.hbase.util` package in name of `HBaseTypeUtils`, it makes sense if its visibility is within the package. And not all public classes are user facing interface, that's why we have the `@Public` annotation. Wdyt?

Anyway, I don't think this is a blocker though.

flink-connectors/flink-hbase/src/main/java/org/apache/flink/addons/hbase/HBaseTableFactory.java
carp84: You're right, unfortunately... Let's find some time to clean this up, in another PR.

wuchong:
Thanks @carp84 , I will fix the travis.

PR #9046, opened 2019-07-09
[Table]Add a PrintTableSink 

hehuiyuan:

This is very convenient for local mode and it is convenient for  learning the Table & SQL Api 

## Brief change log

Add a PrintTableSink class.

## Use method
```scala
PrintTableSink printsink = new PrintTableSink();
tableEnvironment.registerTableSink("printsinktable",printsink.configure(new String[]{"name"},new TypeInformation[]{Types.STRING} ));
tableEnvironment.sqlUpdate("INSERT INTO printsinktable SELECT name from table_name" );
```


aljoscha:
Hi,
thanks for contributing! However, usually PRs are not accepted if they don't correspond to a Jira Issue. Please also see the contribution guide about that: https://flink.apache.org/contributing/how-to-contribute.html.

Aljoscha

PR #9049, opened 2019-07-09
[FLINK-13176][SQL CLI] remember current catalog and database in SQL CLI SessionContext

bowenli86:

This PR supports remembering current catalog and database that users set in SQL CLI SessionContext

## Brief change log

- added current catalog and database in SessionContext
- updated SQL commands execution logic

## Verifying this change

To be added

  - The public API, i.e., is any changed class annotated with `@Public(Evolving)`:(no)
  - The serializers: (no)
  - Anything that affects deployment or recovery: JobManager (and its components), Checkpointing, Yarn/Mesos, ZooKeeper:(no)

## Documentation

### DIFF
--- a/flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/gateway/SessionContext.java
+++ b/flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/gateway/SessionContext.java
@@ -19,13 +19,18 @@
 package org.apache.flink.table.client.gateway;
 
 import org.apache.flink.table.client.config.Environment;
+import org.apache.flink.table.client.config.entries.ExecutionEntry;
 import org.apache.flink.table.client.config.entries.ViewEntry;
+import org.apache.flink.util.StringUtils;
 
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
+
+import static org.apache.flink.util.Preconditions.checkArgument;
 
 /**
  * Context describing a session.
@@ -71,6 +76,26 @@ public class SessionContext {
 
 	public String getName() {
 		return name;
+	}
+
+	public Optional<String> getCurrentCatalog() {
+		return Optional.ofNullable(sessionProperties.get(ExecutionEntry.EXECUTION_CURRNET_CATALOG));
+	}
+
+	public void setCurrentCatalog(String currentCatalog) {
+		checkArgument(!StringUtils.isNullOrWhitespaceOnly(currentCatalog));
+
+		sessionProperties.put(ExecutionEntry.EXECUTION_CURRNET_CATALOG, currentCatalog);
+	}
+
+	public Optional<String> getCurrentDatabase() {
+		return Optional.ofNullable(sessionProperties.get(ExecutionEntry.EXECUTION_CURRNET_DATABASE));
+	}
+
+	public void setCurrentDatabase(String currentDatabase) {
+		checkArgument(!StringUtils.isNullOrWhitespaceOnly(currentDatabase));
+
+		sessionProperties.put(ExecutionEntry.EXECUTION_CURRNET_DATABASE, currentDatabase);
 	}
 
 	public Environment getEnvironment() {
--- a/flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/gateway/local/ExecutionContext.java
+++ b/flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/gateway/local/ExecutionContext.java
@@ -78,7 +78,6 @@ import java.net.URL;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Optional;
 import java.util.function.Supplier;
 
 /**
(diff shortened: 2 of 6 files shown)


KurtYoung:
cc @docete

bowenli86:
please note that UT will added to this PR once FLINK-13175 is fixed.

flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/gateway/SessionContext.java
```diff
@@ -40,6 +44,9 @@
 
 	private final Map<String, ViewEntry> views;
 
+	private Optional<String> currentCatalog = Optional.empty();
```
twalthr: Could we just put the current catalog and into the `sessionProperties`? Then handling of merging environment files would happen automatically and we could reduce the amount of changes in this PR.

bowenli86:
@twalthr @xuefuz @danny0405 I've updated the PR according to feedback. can you please take a look?

Note that I didn't find a good way to test this change due to the SQL CLI architecture except end-to-end test, for which I opened FLINK-12625.

I still feel the architecture to keep cli and gateway completely separate on state is a bit awkward though, but I'm fine to live with it and have this PR to unblock our progress.

bowenli86:
I adopted Timo's suggestion. Please take a look @xuefuz @twalthr @docete

bowenli86:
built successfully with latest master on my own branch https://travis-ci.org/bowenli86/flink/builds/557967675

bowenli86:
@xuefuz can you take a look?

flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/gateway/local/ExecutionContext.java
```diff
@@ -303,14 +303,24 @@ private EnvironmentInstance() {
...
+			if (sessionContext.getCurrentCatalog().isPresent()) {
+				potentialCatalog = sessionContext.getCurrentCatalog();
+			} else if (mergedEnv.getExecution().getCurrentCatalog().isPresent()) {
+				potentialCatalog = mergedEnv.getExecution().getCurrentCatalog();
 			}
 
-			Optional<String> potentialCurrentDatabase = mergedEnv.getExecution().getCurrentDatabase();
-			if (potentialCurrentDatabase.isPresent()) {
-				tableEnv.useDatabase(potentialCurrentDatabase.get());
+			// set current database to default
```
xuefuz: This line of comment doesn't seem matching the code below.

flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/gateway/local/ExecutionContext.java
```diff
@@ -303,14 +303,24 @@ private EnvironmentInstance() {
 			// register catalogs
 			catalogs.forEach(tableEnv::registerCatalog);
 
-			Optional<String> potentialCurrentCatalog = mergedEnv.getExecution().getCurrentCatalog();
-			if (potentialCurrentCatalog.isPresent()) {
-				tableEnv.useCatalog(potentialCurrentCatalog.get());
+			// set current catalog
+			Optional<String> potentialCatalog = Optional.empty();
```
xuefuz: It might be better if we code similarly for both default catalog and default database.

bowenli86:
@xuefuz thanks for your review! I have addressed feedbacks

flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/gateway/local/LocalExecutor.java
```diff
@@ -232,7 +238,10 @@ public void useCatalog(SessionContext session, String catalogName) throws SqlExe
 			.getTableEnvironment();
```
twalthr: @bowenli86 please remove `useCatalog` and `useDatabase` from the executor interface. The functionality is already covered by `org.apache.flink.table.client.gateway.local.LocalExecutor#validateSession`.

flink-table/flink-sql-client/src/test/java/org/apache/flink/table/client/gateway/local/DependencyTest.java
```diff
@@ -193,7 +200,20 @@ public Catalog createCatalog(String name, Map<String, String> properties) {
 					ADDITIONAL_TEST_DATABASE,
 					new CatalogDatabaseImpl(new HashMap<>(), null),
 					false);
-			} catch (DatabaseAlreadyExistException e) {
+				hiveCatalog.createTable(
```
twalthr: @bowenli86 please remove the Hive dependency from the SQL Client. For tests would should follow a clear separation of concerns. We just want to test catalog functionality so we can implement a dummy catalog. Combining multiple modules to test end-to-end behavior should be done in an end-to-end test. Otherwise we end up with complex module dependency structures that are actually not necessary for testing.

twalthr:
Hi @bowenli86, sorry for the post-merge comments. But I think we should not introduce many Hive-specifics into SQL Client. Would be great if you could address my comments.

flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/gateway/local/LocalExecutor.java
bowenli86: @twalthr can you elaborate a bit more? AFAICT, `org.apache.flink.table.client.gateway.local.LocalExecutor#validateSession` seems only to be responsible for creating an execution context, but cannot set any current catalog and current database?

flink-table/flink-sql-client/src/test/java/org/apache/flink/table/client/gateway/local/DependencyTest.java
bowenli86: yeah, the reason at that time is we don't have a good way to test and cover catalog discovery for `HiveCatalog`. As soon as we have end-2-end test (WIP https://github.com/apache/flink/pull/9149), we should be able to remove hive test dependencies from sql cli

twalthr: But you don't need a Hive catalog to test catalog functionality. You can simply create a test catalog similar to sources and sinks in this test.

flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/gateway/local/LocalExecutor.java
bowenli86: `useCatalog` and `useDatabase` are APIs for CliClient when running SQL statements such as "USE CATALOG XXX". Are we talking about the same thing?

flink-table/flink-sql-client/src/test/java/org/apache/flink/table/client/gateway/local/DependencyTest.java
bowenli86: we have a plan to remove the dependencies

PR #9052, opened 2019-07-10
[FLINK-13175] Fix ExecutionContextTest after FLINK-12951

danny0405:

After FLINK-12951, table will always be registered into current catalog, this fix the behavior for ExecutionContextTest.

## Brief change log

  - Remove defaultCatalog and defaultDataBase totally from TableEnvironmentImpl

## Verifying this change


KurtYoung:
Changes LGTM, waiting travis pass: https://travis-ci.org/danny0405/flink/builds/556648449

PR #9054, opened 2019-07-10
[FLINK-13183][Table]Add a PrintTableSink

hehuiyuan:

This is very convenient for local mode and it is convenient for  learning the Table & SQL Api 

## Brief change log

Add a PrintTableSink class.

## Use method
```scala
PrintTableSink printsink = new PrintTableSink();
tableEnvironment.registerTableSink("printsinktable",printsink.configure(new String[]{"name"},new TypeInformation[]{Types.STRING} ));
tableEnvironment.sqlUpdate("INSERT INTO printsinktable SELECT name from table_name" );
```

### DIFF
--- a/flink-table/flink-table-api-java-bridge/src/main/java/org/apache/flink/table/sinks/PrintTableSink.java
+++ b/flink-table/flink-table-api-java-bridge/src/main/java/org/apache/flink/table/sinks/PrintTableSink.java
@@ -1,6 +1,7 @@
 package org.apache.flink.table.sinks;
 
 import org.apache.flink.api.common.typeinfo.TypeInformation;
+import org.apache.flink.api.java.DataSet;
 import org.apache.flink.api.java.typeutils.RowTypeInfo;
 import org.apache.flink.streaming.api.datastream.DataStream;
 import org.apache.flink.streaming.api.datastream.DataStreamSink;
@@ -10,10 +11,15 @@ import org.apache.flink.table.utils.TableConnectorUtils;
 /**
  * A simple {@link TableSink} to emit data to the standard output or standard error stream.
  */
-public class PrintTableSink implements AppendStreamTableSink {
+public class PrintTableSink implements BatchTableSink , AppendStreamTableSink {
 
 	private String[] fieldNames;
 	private TypeInformation<?>[] fieldTypes;
+
+	@Override
+	public void emitDataSet(DataSet dataSet) throws Exception {
+		dataSet.print();
+	}
 
 	@Override
 	public void emitDataStream(DataStream dataStream) {
--- a/flink-table/flink-table-api-java-bridge/src/main/java/org/apache/flink/table/sinks/PrintTableSink.java
+++ b/flink-table/flink-table-api-java-bridge/src/main/java/org/apache/flink/table/sinks/PrintTableSink.java
@@ -1,0 +1,53 @@
+package org.apache.flink.table.sinks;
+
+import org.apache.flink.api.common.typeinfo.TypeInformation;
+import org.apache.flink.api.java.typeutils.RowTypeInfo;
+import org.apache.flink.streaming.api.datastream.DataStream;
+import org.apache.flink.streaming.api.datastream.DataStreamSink;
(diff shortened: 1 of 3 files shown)


zjffdu:
Is it possible to add unit test ? @hehuiyuan

hehuiyuan:
> Is it possible for infer the schema info ? To me, it is redundant to declare it here.

That is possible to print some fields 

> Is it possible for infer the schema info ? To me, it is redundant to declare it here.


Hi，are you talking about fields and types?

flink-table/flink-table-api-java-bridge/src/main/java/org/apache/flink/table/sinks/PrintTableSink.java
```diff
@@ -0,0 +1,63 @@
...
+ */
+public class PrintTableSink implements BatchTableSink , AppendStreamTableSink {
+
+	private String[] fieldNames;
+	private TypeInformation<?>[] fieldTypes;
+
+	@Override
+	public void emitDataSet(DataSet dataSet) {
+		try {
+			dataSet.print();
```
twalthr: This triggers an execution immediately. And does not only add a sink.

flink-table/flink-table-api-java-bridge/src/main/java/org/apache/flink/table/sinks/PrintTableSink.java
```diff
@@ -0,0 +1,63 @@
...
+		} catch (Exception e) {
+			e.printStackTrace();
```
twalthr: Throw exceptions instead of swallowing them.

flink-table/flink-table-api-java-bridge/src/main/java/org/apache/flink/table/sinks/PrintTableSink.java
```diff
@@ -0,0 +1,63 @@
...
+		return fieldNames;
+	}
+
+	@Override
+	public TypeInformation<?>[] getFieldTypes() {
+		return fieldTypes;
+	}
+
+	@Override
+	public TableSink configure(String[] fieldNames, TypeInformation[] fieldTypes) {
```
twalthr: Using this interface is not recommended anymore. Use a `TableFactory` instead?

flink-table/flink-table-api-java-bridge/src/main/java/org/apache/flink/table/sinks/PrintTableSink.java
hehuiyuan: Yeah , `print`  triggers an execution ,the method `emitDataSet `  is  for DataSet.

hehuiyuan: Hi , do you have any better suggestions?
printtable1

flink-table/flink-table-api-java-bridge/src/main/java/org/apache/flink/table/sinks/PrintTableSink.java
hehuiyuan: If use a TableFactory, it will create 
PrintTableSinkFactoryBase  / PrintAppendTableSinkFactory / CsvBatchTableSinkFactory to create StreamtableSink and BatchTableSink for PrintTableSink.

flink-table/flink-table-api-java-bridge/src/main/java/org/apache/flink/table/sinks/PrintTableSink.java
hehuiyuan: ok

flink-table/flink-table-api-java-bridge/src/main/java/org/apache/flink/table/sinks/PrintTableSink.java
hehuiyuan: @twalthr Hi , do you have any better suggestions for `BatchTableSink`?

aljoscha:
I'm closing this now because there is other work going on about adding proper support for collect()-style methods on TableEnvironment, of which print() is one.

PR #9043, opened 2019-07-09
[FLINK-13165] Complete slot requests in request order

tillrohrmann:
## What is the purpose of the change

This commit makes sure that slot requests enqueued at the SlotPoolImpl will get completed
in the order in which they were requested. This makes sure that the original request order
will be respected, hence preventing deadlock situations where dependent requests get completed
first.

## Brief change log

- Use `LinkedHashMap` for `SlotPoolImpl#waitingForResourceManager`
- Renamed `DualKeyMap` to `DualKeyLinkedMap` and use `LinkedHashMap` for `DualKeyLinkedMap#aMap` and `DualKeyLinkedMap#bMap`

## Verifying this change

- Added `SlotPoolRequestCompletionTest`

## Does this pull request potentially affect one of the following parts:

  - Dependencies (does it add or upgrade a dependency): (no)
  - The public API, i.e., is any changed class annotated with `@Public(Evolving)`: (no)
  - The serializers: (no)
  - The runtime per-record code paths (performance sensitive): (no)
  - Anything that affects deployment or recovery: JobManager (and its components), Checkpointing, Yarn/Mesos, ZooKeeper: (yes)
  - The S3 file system connector: (no)

## Documentation

  - Does this pull request introduce a new feature? (no)
  - If yes, how is the feature documented? (not applicable)

PR #9058, opened 2019-07-10
[FLINK-13166] Add support for batch slot requests to SlotPoolImpl

tillrohrmann:

This PR is based on #9043.

This commit adds a new type of slot request which can be issued to the SlotPoolImpl.
The batch slot request is intended for batch jobs which can be executed with a single
slot (having at least one slot for every requested resource profile equivalence class).
Usually, a job which fulfills this criterion must not contain a pipelined shuffle.

The new batch slot request behaves in the following aspects differently than the normal
slot request:

* Batch slot request don't time out if the SlotPool contains at least one allocated slot
which can fulfill the pending slot request
* Batch slot request don't react to the failAllocation signal from the ResourceManager
* Batch slot request don't fail if the slot request to the resource manager fails

In order to time out batch slot request which cannot be fulfilled with an allocated slot,
the SlotPoolImpl schedules a periodic task which checks for this condition. If a slot cannot
be fulfilled, it is marked as unfulfillable and the current timestamp is recorded. If the
slot cannot be marked as fulfillable until the batch slot timeout has been exceeded, the
slot request will be timed out.

The batch slot request will be requested by calling SlotPool#requestNewAllocatedBatchSlot.

cc @xintongsong @StephanEwen 

## Verifying this change

Added `SlotPoolBatchSlotRequestTest`

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-runtime/src/test/java/org/apache/flink/runtime/jobmaster/slotpool/SlotPoolBatchSlotRequestTest.java
+++ b/flink-runtime/src/test/java/org/apache/flink/runtime/jobmaster/slotpool/SlotPoolBatchSlotRequestTest.java
@@ -1,0 +1,339 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.runtime.jobmaster.slotpool;
+
+import org.apache.flink.api.common.JobID;
+import org.apache.flink.api.common.time.Time;
+import org.apache.flink.runtime.clusterframework.types.AllocationID;
+import org.apache.flink.runtime.clusterframework.types.ResourceID;
+import org.apache.flink.runtime.clusterframework.types.ResourceProfile;
+import org.apache.flink.runtime.concurrent.ComponentMainThreadExecutor;
+import org.apache.flink.runtime.concurrent.ComponentMainThreadExecutorServiceAdapter;
+import org.apache.flink.runtime.concurrent.FutureUtils;
+import org.apache.flink.runtime.executiongraph.utils.SimpleAckingTaskManagerGateway;
+import org.apache.flink.runtime.jobmaster.JobMasterId;
+import org.apache.flink.runtime.jobmaster.SlotRequestId;
+import org.apache.flink.runtime.resourcemanager.ResourceManagerGateway;
+import org.apache.flink.runtime.resourcemanager.utils.TestingResourceManagerGateway;
+import org.apache.flink.runtime.taskexecutor.slot.SlotOffer;
+import org.apache.flink.runtime.taskmanager.LocalTaskManagerLocation;
+import org.apache.flink.runtime.taskmanager.TaskManagerLocation;
+import org.apache.flink.runtime.testingUtils.TestingUtils;
+import org.apache.flink.runtime.util.clock.Clock;
+import org.apache.flink.runtime.util.clock.ManualClock;
+import org.apache.flink.runtime.util.clock.SystemClock;
+import org.apache.flink.util.ExceptionUtils;
+import org.apache.flink.util.FlinkException;
+import org.apache.flink.util.TestLogger;
+
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.instanceOf;
+import static org.hamcrest.Matchers.is;
+import static org.junit.Assert.fail;
+
+/**
+ * Tests for batch slot requests.
+ */
+public class SlotPoolBatchSlotRequestTest extends TestLogger {
+
+	private static final ResourceProfile resourceProfile = new ResourceProfile(1.0, 1024);
+	private static final ResourceProfile smallerResourceProfile = new ResourceProfile(0.5, 512);
+	public static final CompletableFuture[] COMPLETABLE_FUTURES_EMPTY_ARRAY = new CompletableFuture[0];
+	private static ScheduledExecutorService singleThreadScheduledExecutorService;
+	private static ComponentMainThreadExecutor mainThreadExecutor;
+
+	@BeforeClass
+	public static void setupClass() {
+		singleThreadScheduledExecutorService = Executors.newSingleThreadScheduledExecutor();
+		mainThreadExecutor = ComponentMainThreadExecutorServiceAdapter.forSingleThreadExecutor(singleThreadScheduledExecutorService);
+	}
+
+	@AfterClass
+	public static void teardownClass() {
+		if (singleThreadScheduledExecutorService != null) {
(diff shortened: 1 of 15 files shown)


xintongsong:
Hi @tillrohrmann, thanks for opening this PR.

I have one concern about the way we handle batch slot requests. 

> - Batch slot request don't time out if the SlotPool contains at least one allocated slot
which can fulfill the pending slot request
> - Batch slot request don't react to the failAllocation signal from the ResourceManager
> - Batch slot request don't fail if the slot request to the resource manager fails

From above descriptions, it seems to me that you are assuming a batch request the is not fulfilled in the first place can only wait for another slot in the slot pool to be freed. I'm not sure about this.

I think we need to consider the possibility that a pending slot request in slot pool can also be satisfied by the resource manager after the slot request to RM failed in the first place.
- When there are multiple jobs running in the cluster, there might be resources taken by other jobs becoming available. While for streaming jobs there may not be many cases that users run multiple jobs in one cluster, it is probably not that rare for batch jobs.
- If we think of fine grained resource profiles, we might be able to regenerate new slots from the resources returned by existing slots. E.g., if we have 1 slot with 100MB and 2 slots with 50MB managed memory in the slot pool, and lots of tasks that requires 100MB managed memory to run. There is a chance that when tasks in the two 50MB slots finished, the two slots are freed and then the resource manager is able to allocate another 100MB slot from the task executor.

Therefore, I would suggest to make the slot pool retry requesting slot from the resource manager for batch slot requests before their timeouts.

tillrohrmann:
Hi @xintongsong, you are right that we don't retry slot requests from within the `SlotPool`. However, we do so in case of a failover.

I'm actually not so sure whether the `SlotPool` should retry the requests. One could also argue that the `SlotManager` should simply not fail the slot requests if it thinks that it can eventually satisfy it. In both cases you need to define a timeout, though, and it just happens that in the current implementation we don't retry.

Moreover, I think that this retry logic would be an optimization of the existing logic and, thus, not strictly release critical. Since the batch request is anyway only a band aid/temporary bridge, I would like to avoid to add even more logic which we need to rework later on just before the feature freeze.

Please object if you think that this feature is a release blocker and needs to be added.

StephanEwen:
I agree with Till here. The logic is not yet perfect, but should be an improvement over the current state.

Under fine-grained recovery, the current state would lead to failure of a task and individual recovery, re-triggering a request to the RM. That is good, but the downside is that it takes away recovery attempts. I think this is tricky for users to understand, that we rely on failure / recovery to re-request resources. It makes re-try attempts meaningless and brings users to debug jobs (because they see unexpected failures) when really nothing is wrong.

With this change here, we don't rely on failure/recovery any more, but do not re-trigger timed out requests within a stage. It may hence be that a stage does not optimally use its resources. Requests come again in the next stage.

Like Till suggested, for 1.10, we should consider a different model. Requests from the SlotPool to the RM should not time out (unless there is an actual failure) and resources that appear at the RM make it to the SlotPool. Letting the SlotPool periodically request resources seems like a workaround to me.

tillrohrmann:
Another thing which deserves discussion is how we handle failed slot requests to the `ResourceManager`. With this PR, batch requests will simply ignore the failed requests to the `ResourceManager` and rely on the `batchSlotTimeout` to fail the request eventually.

However, with the #8740, we fail the `ResourceManager` slot request if we know that we cannot fulfill the requested `ResourceProfile`. In this case, we won't react to this signal and instead wait on the timeout. If the contract is that a slot request to the `ResourceManager` only fails if the `ResourceManager` is sure that it cannot fulfill the request (never not just in the moment the request arrived), then we could use this signal to immediately fail the slot request.

If we decide to do this, then we should do it as a follow up to this PR.

xintongsong:
Sorry to bring it back here.

> With this change here, we don't rely on failure/recovery any more, but do not re-trigger timed out requests within a stage. It may hence be that a stage does not optimally use its resources. Requests come again in the next stage.

I'm worried about that is is not only a matter of whether optimally use the resources, but may cause jobs fail to execute when the total resource of the cluster is actually enough.

Consider we have two vertex A (2 parallelism, 100MB managed memory) and B (1 parallelism, 200MB managed memory), and the cluster has one TM with 200MB managed memory in total. Ideally, we would expect the two tasks of vertex A run on the TM concurrently, and then one task of vertex B. However, if we first request two 100MB slots, RM can not allocate any more 200MB slot for B, and there is no 200MB in the slot pool neither. As a result, the slot request for B is failed, leading to the job failure.

Please correct me if I was wrong. I'm not very sure about whether the failure of slot request for B would cause the job to fail, or whether this can be resolved by the fine grained recovery.

zhuzhurk:
Hi Xingtong, I think you are right that this improvement cannot handle the case you describes. However, the fine-grained recovery can work as a fallback. It uses re-scheduling as a retry for resources. In this way B will finally get assigned with the resources that is released from A and returned to RM.
As Stephan mentioned, the failover way can be annoying to Flink users. And Till's PR is targeting for improvement this by reducing task failovers caused by slot allocation timeout. It works for most cases, although not all(like the one you mentioned).

xintongsong:
Thank you for the clarification, @zhuzhurk.

If the fine grained recover can work as a fallback, I think we can eventually get jobs in scenarios that I described above to work. As long as we don't fails jobs that should be able to run, then this should be an optimization that we don't necessarily do for this version.

And @tillrohrmann, I want to make one more clarification. 

> However, with the #8740, we fail the ResourceManager slot request if we know that we cannot fulfill the requested ResourceProfile. In this case, we won't react to this signal and instead wait on the timeout. If the contract is that a slot request to the ResourceManager only fails if the ResourceManager is sure that it cannot fulfill the request (never not just in the moment the request arrived), then we could use this signal to immediately fail the slot request.

I think currently we are failing the ResourceManager slot request if it cannot be fulfilled not only for the moment request arrived with #8740. To be more specific, the logic is to fail slot request that can not be fulfilled by neither a registered slot (after the startup period, for standalone), nor a pending slot that can be allocated (for Yarn/Mesos). In other words, we are failing slot requests that can, to the best of our knowledge, never be fulfilled.

I would prefer the current way (fail requests that can never be fulfilled), while @StephanEwen said we could also consider the other way (fail requests that can not be fulfilled at the moment). What do you think?

tillrohrmann:
Yes, I think the fine grained recovery should effectively retrigger the request as part of its failover. This is not nice, but should do the trick.

For the problem with differently sized slots which originate from the same TM, this might actually be an argument to not include the change which allows to share managed memory between slots in this release @StephanEwen. It seems to me that not all implications are clear at the moment. Being able to dynamically size the slots complicates the slot allocation protocol further because with this change requests might become fulfillable depending on what one releases.

flink-runtime/src/test/java/org/apache/flink/runtime/jobmaster/slotpool/SlotPoolBatchSlotRequestTest.java
```diff
@@ -0,0 +1,341 @@
...
+			final List<CompletableFuture<PhysicalSlot>> slotFutures = Arrays.asList(firstSlotFuture, secondSlotFuture, thirdSlotFuture);
+
+			// initial batch slot timeout check
+			advanceTimeAndTriggerCheckBatchSlotTimeout(slotPool, clock, batchSlotTimeout);
+
+			assertThat(CompletableFuture.anyOf(slotFutures.toArray(COMPLETABLE_FUTURES_EMPTY_ARRAY)).isDone(), is(false));
+
+			releaseTaskManager(slotPool, directMainThreadExecutor, taskManagerResourceId);
+
+			clock.advanceTime(1L, TimeUnit.MILLISECONDS);
```
StephanEwen: This line can probably be removed from the test.

flink-runtime/src/test/java/org/apache/flink/runtime/jobmaster/slotpool/SlotPoolPendingRequestFailureTest.java
```diff
@@ -127,13 +133,50 @@ public void testFailingResourceManagerRequestFailsPendingSlotRequestAndCancelsRM
...
+		try {
+			final Time timeout = Time.milliseconds(5L);
+
+			final CompletableFuture<PhysicalSlot> slotFuture = CompletableFuture
+				.supplyAsync(() -> requestNewAllocatedSlot(slotPool, new SlotRequestId(), timeout), componentMainThreadExecutor)
+				.thenCompose(Function.identity());
+
+			try {
+				slotFuture.get();
+				fail("Expected TimeoutFuture.");
```
StephanEwen: That error message seems a bit off.

StephanEwen:
+1 to merge this

flink-runtime/src/test/java/org/apache/flink/runtime/jobmaster/slotpool/SlotPoolBatchSlotRequestTest.java
tillrohrmann: Good point, will remove.

flink-runtime/src/test/java/org/apache/flink/runtime/jobmaster/slotpool/SlotPoolPendingRequestFailureTest.java
tillrohrmann: True, will correct it.

tillrohrmann:
Thanks for the review @StephanEwen. I've addressed your comments. Merging once Travis gives green light.

PR #9060, opened 2019-07-10
[FLINK-13145][tests] Run HA dataset E2E test with new RestartPipelinedRegionStrategy

GJL:

*Run HA dataset end-to-end test (`test-scripts/test_ha_dataset.sh`) with `AdaptedRestartPipelinedRegionStrategyNG` enabled*

## Brief change log

  - *See commits*

## Verifying this change

  - *The change is a test, and it passes.*

## Documentation

### DIFF
--- a/flink-end-to-end-tests/flink-dataset-fine-grained-recovery-test/pom.xml
+++ b/flink-end-to-end-tests/flink-dataset-fine-grained-recovery-test/pom.xml
@@ -1,0 +1,89 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+		xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+
+	<modelVersion>4.0.0</modelVersion>
+
+	<parent>
+		<groupId>org.apache.flink</groupId>
+		<artifactId>flink-end-to-end-tests</artifactId>
+		<version>1.10-SNAPSHOT</version>
+		<relativePath>..</relativePath>
+	</parent>
+
+	<artifactId>flink-dataset-fine-grained-recovery-test</artifactId>
+	<name>flink-dataset-fine-grained-recovery-test</name>
+	<packaging>jar</packaging>
+
+	<dependencies>
+		<dependency>
+			<groupId>org.apache.flink</groupId>
+			<artifactId>flink-java</artifactId>
+			<version>${project.version}</version>
+			<scope>provided</scope>
+		</dependency>
+	</dependencies>
+
+	<build>
+		<plugins>
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-surefire-plugin</artifactId>
+				<executions>
+					<execution>
+						<id>dataset-fine-grained-recovery</id>
+						<phase>test</phase>
+						<goals>
+							<goal>test</goal>
+						</goals>
+						<configuration>
+							<includes>
+								<include>**/*Test.*</include>
+							</includes>
+						</configuration>
+					</execution>
+				</executions>
+			</plugin>
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-shade-plugin</artifactId>
+				<executions>
+					<execution>
+						<id>DataSetFineGrainedRecoveryTestProgram</id>
+						<phase>package</phase>
+						<goals>
+							<goal>shade</goal>
+						</goals>
+						<configuration>
+							<finalName>DataSetFineGrainedRecoveryTestProgram</finalName>
+							<transformers>
+								<transformer implementation="org.apache.maven.plugins.shade.resource.ManifestResourceTransformer">
+									<mainClass>org.apache.flink.batch.tests.DataSetFineGrainedRecoveryTestProgram</mainClass>
+								</transformer>
+							</transformers>
+						</configuration>
+					</execution>
+				</executions>
+			</plugin>
+		</plugins>
+	</build>
+
+</project>
--- a/flink-end-to-end-tests/flink-dataset-fine-grained-recovery-test/src/main/java/org/apache/flink/batch/tests/util/FileBasedOneShotLatch.java
+++ b/flink-end-to-end-tests/flink-dataset-fine-grained-recovery-test/src/main/java/org/apache/flink/batch/tests/util/FileBasedOneShotLatch.java
@@ -1,0 +1,129 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.flink.batch.tests.util;
+
+import com.sun.nio.file.SensitivityWatchEventModifier;
+
+import javax.annotation.concurrent.NotThreadSafe;
+
+import java.io.Closeable;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.StandardWatchEventKinds;
+import java.nio.file.WatchEvent;
+import java.nio.file.WatchKey;
+import java.nio.file.WatchService;
+
+import static org.apache.flink.util.Preconditions.checkNotNull;
+
+/**
+ * A synchronization aid that allows a single thread to wait on the creation of a specified file.
+ */
+@NotThreadSafe
+public class FileBasedOneShotLatch implements Closeable {
+
+	private final Path latchFile;
+
+	private final WatchService watchService;
+
+	private boolean released;
+
+	public FileBasedOneShotLatch(final Path latchFile) {
+		this.latchFile = checkNotNull(latchFile);
+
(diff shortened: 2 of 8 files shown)


flink-end-to-end-tests/test-scripts/test_ha_dataset.sh
```diff
@@ -53,20 +52,51 @@ function run_ha_test() {
...
         # create a new one which will take over
         kill_single 'StandaloneSessionClusterEntrypoint'
         wait_job_running ${JOB_ID}
     done
 
-    cancel_job ${JOB_ID}
+    for (( c=0; c<${TM_KILLS}; c++ )); do
+        sleep $(( ( RANDOM % 10 )  + 1 ))
+        kill_and_replace_random_task_manager
+        wait_job_running ${JOB_ID}
```
zentol: you should check the return value here to ensure the job actually reached a running state

flink-end-to-end-tests/test-scripts/test_ha_dataset.sh
```diff
@@ -53,20 +52,51 @@ function run_ha_test() {
...
+    done
+
+    wait_job_terminal_state ${JOB_ID} "FINISHED"
```
zentol: I'm wondering how useful this test can be if it has a finite runtime. For all we know the job could finish before we even kill the JM, or maybe more likely, any TM.

From what we've seen so far, this can only result in either:
a) The test taking significantly more time than required, as we have to ensure that all these events happen before it terminates
b) The test being unstable.

I like neither option, do admit though that this would make it very difficult (or even impossible) to verify the correctness of the output.

How much longer does the test now run for?

It looks like the test was _supposed_ to run indefinitely (`--source true`), but this isn't in sync with the actual parameters of the job which use `infinite` as the parameter.

flink-end-to-end-tests/test-scripts/test_ha_dataset.sh
GJL: `wait_job_running` will terminate the script if the job does not become running within a timeout (10s). Since we are not launching a new process by invoking the function, the main script will exit. Am I missing something?

zentol: you're right.

flink-end-to-end-tests/test-scripts/test_ha_dataset.sh
```diff
@@ -53,20 +50,51 @@ function run_ha_test() {
...
         # create a new one which will take over
         kill_single 'StandaloneSessionClusterEntrypoint'
         wait_job_running ${JOB_ID}
     done
 
-    cancel_job ${JOB_ID}
+    for (( c=0; c<${TM_KILLS}; c++ )); do
+        sleep $(( ( RANDOM % 10 )  + 1 ))
+        kill_and_replace_random_task_manager
+        wait_job_running ${JOB_ID}
```
GJL: `wait_job_running` can be omitted. In fact it only asserts that the job appears in the `flink list`

flink-end-to-end-tests/test-scripts/test_ha_dataset.sh
GJL: These are valid concerns.

> How much longer does the test now run for?

The test runs 4.5-5 minutes on my machine. It takes around 2 minutes to complete the batch job after the last injected fault (time determined using unscientific methods). 

> I like neither option, do admit though that this would make it very difficult (or even impossible) to verify the correctness of the output.

I don't see a good solution yet. Here are some options:
1. Make job block on external signals (files)
1. Leave it as before, i.e., don't verify correctness of the output

flink-end-to-end-tests/flink-dataset-fine-grained-recovery-test/src/test/java/org/apache/flink/batch/tests/util/FileBasedOneShotLatchTest.java
```diff
@@ -0,0 +1,84 @@
...
+
+import java.io.File;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+import static org.junit.Assert.assertTrue;
+
+/**
+ * Tests for {@link FileBasedOneShotLatch}.
+ */
+public class FileBasedOneShotLatchTest {
```
GJL: Test is not run due to surefire config in the `flink-end-to-end-tests` module. I don't have a solution. Suggestions welcome.

flink-end-to-end-tests/test-scripts/test_ha_dataset.sh
GJL: I added a new job that blocks on an external condition.

flink-end-to-end-tests/flink-dataset-fine-grained-recovery-test/src/test/java/org/apache/flink/batch/tests/util/FileBasedOneShotLatchTest.java
zentol: you can override the surefire configuration for this module to run this test; have a look at the prometheus/metrics-availability e2e test poms. Alternatively, maybe not have this in this module but in flink-test-utils?

flink-end-to-end-tests/flink-dataset-fine-grained-recovery-test/src/main/java/org/apache/flink/batch/tests/util/FileBasedOneShotLatch.java
```diff
@@ -0,0 +1,126 @@
...
+			return;
+		}
+
+		awaitLatchFile(watchService);
+	}
+
+	private void awaitLatchFile(final WatchService watchService) throws InterruptedException {
+		while (true) {
+			WatchKey take = watchService.take();
+			if (isReleasedOrReleasable()) {
```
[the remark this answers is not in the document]
GJL: It could be that other files are created in the directory.

flink-end-to-end-tests/flink-dataset-fine-grained-recovery-test/src/test/java/org/apache/flink/batch/tests/util/FileBasedOneShotLatchTest.java
```diff
@@ -0,0 +1,84 @@
...
+			try {
+				latch.await();
+				awaitCompleted.set(true);
+			} catch (InterruptedException e) {
+				Thread.currentThread().interrupt();
+			}
+		});
+		thread.start();
+
+		latchFile.createNewFile();
```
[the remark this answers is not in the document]
GJL: I will add a new test case

flink-end-to-end-tests/flink-dataset-fine-grained-recovery-test/src/test/java/org/apache/flink/batch/tests/util/FileBasedOneShotLatchTest.java
```diff
@@ -0,0 +1,84 @@
...
+
+import static org.junit.Assert.assertTrue;
+
+/**
+ * Tests for {@link FileBasedOneShotLatch}.
+ */
+public class FileBasedOneShotLatchTest {
+
+	@Rule
+	public TemporaryFolder temporaryFolder = new TemporaryFolder();
```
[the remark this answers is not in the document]
GJL: done

flink-end-to-end-tests/flink-dataset-fine-grained-recovery-test/src/main/java/org/apache/flink/batch/tests/util/FileBasedOneShotLatch.java
```diff
@@ -0,0 +1,126 @@
...
+
+	private static WatchService initWatchService(final Path parentDir) {
+		final WatchService watchService = createWatchService();
+		watchForLatchFile(watchService, parentDir);
+		return watchService;
+	}
+
+	private static WatchService createWatchService() {
+		try {
+			return FileSystems.getDefault().newWatchService();
```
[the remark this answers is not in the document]
GJL: good suggestion, done

flink-end-to-end-tests/flink-dataset-fine-grained-recovery-test/src/test/java/org/apache/flink/batch/tests/util/FileBasedOneShotLatchTest.java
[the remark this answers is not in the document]
GJL: Done

flink-end-to-end-tests/flink-dataset-fine-grained-recovery-test/src/test/java/org/apache/flink/batch/tests/util/FileBasedOneShotLatchTest.java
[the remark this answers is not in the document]
GJL: accidentally used `--amend` when fixing

flink-end-to-end-tests/flink-dataset-fine-grained-recovery-test/pom.xml
```diff
@@ -0,0 +1,76 @@
...
+		<dependency>
+			<groupId>org.apache.flink</groupId>
+			<artifactId>flink-java</artifactId>
+			<version>${project.version}</version>
+			<scope>provided</scope>
+		</dependency>
+
+		<dependency>
+			<groupId>junit</groupId>
+			<artifactId>junit</artifactId>
```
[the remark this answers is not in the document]
GJL: Actually this dependency is not needed since it already comes from `flink-parent`

flink-end-to-end-tests/flink-dataset-fine-grained-recovery-test/src/test/java/org/apache/flink/batch/tests/util/FileBasedOneShotLatchTest.java
GJL: Changed the surefire config. I find it awkward to have a dependency on  `flink-test-utils` with `compile` scope since the job is strictly speaking not a test.

flink-end-to-end-tests/flink-dataset-fine-grained-recovery-test/pom.xml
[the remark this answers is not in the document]
GJL: Removed.

flink-end-to-end-tests/flink-dataset-fine-grained-recovery-test/src/main/java/org/apache/flink/batch/tests/util/FileBasedOneShotLatch.java
```diff
@@ -0,0 +1,125 @@
...
+		if (isReleasedOrReleasable()) {
+			return;
+		}
+
+		awaitLatchFile(watchService);
+	}
+
+	private void awaitLatchFile(final WatchService watchService) throws InterruptedException {
+		while (true) {
+			WatchKey take = watchService.take();
```
GJL: rename to `watchKey`

GJL: done

flink-end-to-end-tests/flink-dataset-fine-grained-recovery-test/src/main/java/org/apache/flink/batch/tests/util/FileBasedOneShotLatch.java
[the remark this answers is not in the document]
zentol: ok. I suppose one _could_ check this via `WatchKey#pollEvents#Context, but the version in the PR is wayyyy simpler. Prone to files being deleted in-between, but this seems unlikely.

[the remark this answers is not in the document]
GJL: > Prone to files being deleted in-between, but this seems unlikely.

True, didn't think about that case. I think it's acceptable to leave it as is because:
- we wait until the job is finished before deleting files
- class is contained in this module
- code is simpler (as you already mentioned)

GJL:
Squashed commits and rebased to latest master. Merging when build is green.

PR #9061, opened 2019-07-10
[FLINK-13078][table-common] Add a logical type parser

twalthr:

This adds a parser for all logical types defined in FLIP-37. This parser is both useful for the SQL Client as well as type annotations for UDFs.

## Brief change log

See commit messages.

## Verifying this change

This change added tests and can be verified as follows: `LogicalTypeParserTest`

  - The serializers: no

## Documentation

### DIFF
--- a/flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/logical/UnresolvedUserDefinedType.java
+++ b/flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/logical/UnresolvedUserDefinedType.java
@@ -1,0 +1,153 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.table.types.logical;
+
+import org.apache.flink.annotation.PublicEvolving;
+import org.apache.flink.table.api.TableException;
+import org.apache.flink.table.utils.EncodingUtils;
+import org.apache.flink.util.Preconditions;
+
+import javax.annotation.Nullable;
+
+import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+/**
+ * Placeholder type of an unresolved user-defined type that is identified by a partially or fully
+ * qualified path ({@code [catalog].[database].[type]}).
+ *
+ * <p>It assumes that a type has been registered in a catalog and just needs to be resolved to a
+ * {@link DistinctType} or {@link StructuredType}.
+ *
+ * <p>Two unresolved types are considered equal if they share the same path in a stable session context.
+ *
+ * @see UserDefinedType
+ */
+@PublicEvolving
+public final class UnresolvedUserDefinedType extends LogicalType {
+
+	private final @Nullable String catalog;
+
+	private final @Nullable String database;
+
+	private final String typeIdentifier;
+
+	public UnresolvedUserDefinedType(
+			boolean isNullable,
+			@Nullable String catalog,
+			@Nullable String database,
+			String typeIdentifier) {
+		super(isNullable, LogicalTypeRoot.UNRESOLVED);
+		this.catalog = catalog;
+		this.database = database;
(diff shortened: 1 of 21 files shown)


flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/logical/utils/LogicalTypeParser.java
```diff
@@ -0,0 +1,916 @@
...
+		private LogicalType parseDoubleType() {
+			if (hasNextToken(Keyword.PRECISION)) {
+				nextToken(Keyword.PRECISION);
+			}
+			return new DoubleType();
+		}
+
+		private LogicalType parseTimeType() {
+			int precision = TimeType.DEFAULT_PRECISION;
+			if (hasNextToken(TokenType.BEGIN_PARAMETER)) {
```
dawidwys: nit: use `parseOptionalPrecision`

flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/logical/utils/LogicalTypeParser.java
```diff
@@ -0,0 +1,916 @@
...
+				nextToken(Keyword.WITHOUT);
+				nextToken(Keyword.TIME);
+				nextToken(Keyword.ZONE);
+			}
+			return new TimeType(precision);
+		}
+
+		private LogicalType parseTimestampType() {
+			int precision = TimestampType.DEFAULT_PRECISION;
+			if (hasNextToken(TokenType.BEGIN_PARAMETER)) {
```
dawidwys: nit: use `parseOptionalPrecision`

flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/logical/utils/LogicalTypeParser.java
```diff
@@ -0,0 +1,916 @@
...
+				case MONTH:
+					return new YearMonthIntervalType(
+						YearMonthResolution.MONTH,
+						yearPrecision);
+				default:
+					throw parsingError("Invalid year-month interval resolution.");
+			}
+		}
+
+		private LogicalType parseDayTimeIntervalType() {
```
dawidwys: How about splitting this method by the 3 cases.

dawidwys: Moreover I think this method could be improved, if we first parse lower/upper resolution and then reuse `DataTypes.Resolution#resolveInterval`. In a pseudo code it could look sth like:

```
public Resolution parseAsResolution(IntervalUnit type) {
    if (type == DAY) {
	dayPrecision = parseOptionalPrecision(dayPrecision);
        return new Resolution(type, dayPrecision);
    } else if (lower == SECOND) {
	fractionalPrecision = parseOptionalPrecision(fractionalPrecision);
        return new Resolution(type, fractionalPrecision);
    }
}

public LogicalType parseDayTimeIntervalType() {
    IntervalUnit upper = tokenAsIntervalUnit();
    Optional<IntervalUnit> lower = tokenAsIntervalUnit();

    DataTypes.Resolution.resolveInterval(parseAsResolution(upper), lower.map(parseAsResolution).orNull)
}
```

flink-table/flink-table-common/src/test/java/org/apache/flink/table/types/LogicalTypeParserTest.java
```diff
@@ -0,0 +1,386 @@
...
+/**
+ * Tests for {@link LogicalTypeParser}.
+ */
+@RunWith(Parameterized.class)
+public class LogicalTypeParserTest {
+
+	@Parameters(name = "{index}: [From: {0}, To: {1}]")
+	public static List<Object[]> testData() {
+		return Arrays.asList(
+			new Object[][]{
```
dawidwys: nit: I think it would make sense to have a `TestSpec` object in this case, right now we have a magic `null` value for most of the test cases. I think it would be easier to read if it was:

```
    TestSpec
         .forString("CHAR")
         .parseAsType(new CharType());

    TestSpec
         .forString("ROW<")
         .throwExceptionWithMessage("...")
```

flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/logical/UnresolvedUserDefinedType.java
```diff
@@ -0,0 +1,153 @@
...
+ * @see UserDefinedType
+ */
+@PublicEvolving
+public final class UnresolvedUserDefinedType extends LogicalType {
+
+	private final @Nullable String catalog;
+
+	private final @Nullable String database;
+
+	private final String typeIdentifier;
```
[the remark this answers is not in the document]
twalthr: In the first version `ObjectIdentifier` allowed also partially defined paths. But we changed it to full identifiers. I would be up for relaxing this constraint to be honest as I see already a couple of `String[]` `List<String>` objects travelling through the stack.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/logical/utils/LogicalTypeParser.java
twalthr: TBH I think it is not worth the effort. Usually, such classes are generated and look much worse. Unless, the standard changes we will also not need to make changes in this method. So maintainability is also not an issue. Exposing `DataTypes.Resolution.resolveInterval` is mixing API with internal utils.

PR #9059, opened 2019-07-10
[FLINK-13141][network] Remove getBufferSize method from BufferPoolFactory

zhijiangW:
## What is the purpose of the change

*`BufferPoolFactory#getBufferSize` is only used for creating subpartitions in `ResultPartitionFactory`. We could pass the network buffer size from `NettyShuffleEnvironmentConfiguration` while constructing the `ResultPartitionFactory`, then the interface method `getBufferSize` could be removed form `BufferPoolFactory`.*

## Brief change log

  - *Pass network buffer size while constructing `ResultPartitionFactory`*
  - *Support update network buffer size in `ResultPartitionBuilder`*
  - *Remove `getBufferSize` from `BufferPoolFactory`*

## Verifying this change

This change is a trivial rework / code cleanup without any test coverage.

## Does this pull request potentially affect one of the following parts:

  - Dependencies (does it add or upgrade a dependency): (yes / **no**)
  - The public API, i.e., is any changed class annotated with `@Public(Evolving)`: (yes / **no**)
  - The serializers: (yes / **no** / don't know)
  - The runtime per-record code paths (performance sensitive): (yes / **no** / don't know)
  - Anything that affects deployment or recovery: JobManager (and its components), Checkpointing, Yarn/Mesos, ZooKeeper: (yes / **no** / don't know)
  - The S3 file system connector: (yes / **no** / don't know)

## Documentation

  - Does this pull request introduce a new feature? (yes / **no**)
  - If yes, how is the feature documented? (**not applicable** / docs / JavaDocs / not documented)

PR #9062, opened 2019-07-10
[FLINK-13100][network] Fix the bug of throwing IOException while FileChannelBoundedData#nextBuffer

zhijiangW:

*In the process of `FileChannelBoundedData#nextBuffer`, the implementation is based on the assumption that there always has an available buffer, otherwise an `IOException` is thrown. But in the case of pending flush operation that means the socket channel is not writable sometimes while netty thread calling `writeAndFlush` method, the first fetched buffer from `FileChannelBoundedData` has not been recycled while fetching the second buffer to trigger next read ahead. Then it would cause this problem to break above assumption.*
    
*In order to fix this problem, we could make the assumption that the read ahead is not always available for `FileChannelBoundedData`. If there are no available buffers to read the next data, we could retrigger the read ahead while recycling buffer via `ResultSubpartitionView#notifyDataAvailable`.*

*This PR relies on #9059 for better constructing the unit tests.*

## Brief change log

  - *Move private `NoOpResultSubpartitionView` class out of `PartitionRequestQueueTest`*
  - *Make `BoundedBlockingSubpartitionType` configurable in `ResultPartitionBuilder` for tests usage*
  - *Pass `BufferAvailabilityListener` in the constructor of `BoundedBlockingSubpartitionReader`*
  - *Implement the logic of `notifyDataAvailable` in `BoundedBlockingSubpartitionReader `*
  - *Fix the logic of `FileBufferReader#nextBuffer`*
  - *Fix the logic of `FileBufferReader#recycleBuffer`*

## Verifying this change

  - *Added `FileChannelBoundedDataTest#testReadNextBuffer` for verifying the logic of `nextBuffer`*
  - *Added `FileChannelBoundedDataTest#testRecycleBufferForNotifyingSubpartitionView` for verifying the notification of `ResultSubpartitionView`*
  - *Added `FileChannelBoundedDataTest#testRecycleBufferForNotifyingBufferAvailabilityListener` for verifying the notification of `BufferAvailabilityListener`*

## Documentation

### DIFF
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartition.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartition.java
@@ -198,9 +198,8 @@ final class BoundedBlockingSubpartition extends ResultSubpartition {
 
 			availability.notifyDataAvailable();
 
-			final BoundedData.Reader dataReader = data.createReader();
 			final BoundedBlockingSubpartitionReader reader = new BoundedBlockingSubpartitionReader(
-					this, dataReader, numDataBuffersWritten);
+					this, data, numDataBuffersWritten, availability);
 			readers.add(reader);
 			return reader;
 		}
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartitionReader.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartitionReader.java
@@ -37,6 +37,9 @@ final class BoundedBlockingSubpartitionReader implements ResultSubpartitionView 
 	/** The result subpartition that we read. */
 	private final BoundedBlockingSubpartition parent;
 
+	/** The listener that is notified when there are available buffers for this subpartition view. */
+	private final BufferAvailabilityListener availabilityListener;
+
 	/** The next buffer (look ahead). Null once the data is depleted or reader is disposed. */
 	@Nullable
 	private Buffer nextBuffer;
@@ -57,16 +60,20 @@ final class BoundedBlockingSubpartitionReader implements ResultSubpartitionView 
 	 */
 	BoundedBlockingSubpartitionReader(
 			BoundedBlockingSubpartition parent,
-			BoundedData.Reader dataReader,
-			int numDataBuffers) throws IOException {
-
-		checkArgument(numDataBuffers >= 0);
+			BoundedData data,
+			int numDataBuffers,
+			BufferAvailabilityListener availabilityListener) throws IOException {
 
 		this.parent = checkNotNull(parent);
-		this.dataReader = checkNotNull(dataReader);
+
+		checkNotNull(data);
+		this.dataReader = data.createReader(this);
+		this.nextBuffer = dataReader.nextBuffer();
+
+		checkArgument(numDataBuffers >= 0);
 		this.dataBufferBacklog = numDataBuffers;
 
-		this.nextBuffer = dataReader.nextBuffer();
+		this.availabilityListener = checkNotNull(availabilityListener);
 	}
 
 	@Nullable
@@ -89,9 +96,31 @@ final class BoundedBlockingSubpartitionReader implements ResultSubpartitionView 
 		return BufferAndBacklog.fromBufferAndLookahead(current, nextBuffer, dataBufferBacklog);
 	}
 
+	/**
+	 * This method is actually only meaningful for the {@link BoundedBlockingSubpartitionType#FILE}.
+	 *
+	 * <p>For the other types the {@link #nextBuffer} can not be ever set to null, so it is no need
+	 * to notify available via this method. But the implementation is also compatible with other
+	 * types even though called by mistake.
+	 */
 	@Override
 	public void notifyDataAvailable() {
-		throw new IllegalStateException("No data should become available on a blocking partition during consumption.");
+		if (nextBuffer == null) {
+			assert dataReader != null;
+
+			try {
+				nextBuffer = dataReader.nextBuffer();
+			} catch (IOException ex) {
+				// this exception wrapper is only for avoiding throwing IOException explicitly
+				// in relevant interface methods
+				throw new IllegalStateException("No data available while reading", ex);
+			}
+
+			// next buffer is null indicates the end of partition
+			if (nextBuffer != null) {
+				availabilityListener.notifyDataAvailable();
+			}
+		}
 	}
 
 	@Override
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/FileChannelBoundedData.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/FileChannelBoundedData.java
@@ -75,11 +75,11 @@ final class FileChannelBoundedData implements BoundedData {
 	}
 
 	@Override
-	public Reader createReader() throws IOException {
+	public Reader createReader(ResultSubpartitionView subpartitionView) throws IOException {
 		checkState(!fileChannel.isOpen());
 
 		final FileChannel fc = FileChannel.open(filePath, StandardOpenOption.READ);
-		return new FileBufferReader(fc, memorySegmentSize);
+		return new FileBufferReader(fc, memorySegmentSize, subpartitionView);
 	}
 
 	@Override
@@ -117,7 +117,12 @@ final class FileChannelBoundedData implements BoundedData {
 
 		private final ArrayDeque<MemorySegment> buffers;
 
-		FileBufferReader(FileChannel fileChannel, int bufferSize) {
+		private final ResultSubpartitionView subpartitionView;
+
+		/** The tag indicates whether we have read the end of this file. */
+		private boolean isFinished;
+
+		FileBufferReader(FileChannel fileChannel, int bufferSize, ResultSubpartitionView subpartitionView) {
 			this.fileChannel = checkNotNull(fileChannel);
 			this.headerBuffer = BufferReaderWriterUtil.allocatedHeaderBuffer();
 			this.buffers = new ArrayDeque<>(NUM_BUFFERS);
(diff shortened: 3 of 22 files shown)


zhijiangW:
@flinkbot attention @pnowojski @StephanEwen 

I would add a new ITCase for the code path of `BoundedBlockingSubpartitionType#file` in a hotfix commit later.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartition.java
```diff
@@ -198,9 +198,8 @@ public ResultSubpartitionView createReadView(BufferAvailabilityListener availabi
 
 			availability.notifyDataAvailable();
```
pnowojski: Can we rephrase the commit message a bit? For example: 
```
The implementation of FileChannelBoundedData#nextBuffer assumes that there is always an available buffer, otherwise an IOException is thrown and it always assumes that pool of two buffers is enough (before using the 3rd buffer, first one was expected to be recycled already). But in the case of pending flush operation (when the
socket channel is not writable while netty thread is calling writeAndFlush method), the first fetched buffer from
FileChannelBoundedData has not been recycled while fetching the second buffer to trigger next read ahead, which breaks the above assumption.

In order to fix this problem, we make read ahead is not always available for FileChannelBoundedData.
If there are no available buffers to read the next data, we retrigger the read ahead while recycling buffer via
ResultSubpartitionView#notifyDataAvailable.
```

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartitionReader.java
```diff
@@ -89,9 +114,30 @@ public BufferAndBacklog getNextBuffer() throws IOException {
 		return BufferAndBacklog.fromBufferAndLookahead(current, nextBuffer, dataBufferBacklog);
 	}
 
+	/**
+	 * This method is actually only meaningful for the {@link BoundedBlockingSubpartitionType#FILE}. For the
+	 * other types the next buffer is always not null while {@link ResultSubpartitionView#getNextBuffer()},
```
pnowojski:
```
For the other types the {@link #nextBuffer} can not be ever set to null, so it is (...)
``` 
also if that's true, I think we would need a checkState whenever we set `nextBuffer` to `null`?

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartitionReader.java
```diff
@@ -89,9 +114,30 @@ public BufferAndBacklog getNextBuffer() throws IOException {
...
+	 * so it is no need to notify available via this method. But the implementation is also compatible with
+	 * other types even though called by mistake.
+	 */
 	@Override
 	public void notifyDataAvailable() {
-		throw new IllegalStateException("No data should become available on a blocking partition during consumption.");
+		if (nextBuffer == null) {
```
pnowojski: add `checkState` for `BoundedBlockingSubpartitionType#FILE`?

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartitionReader.java
```diff
@@ -67,6 +73,25 @@
 		this.dataBufferBacklog = numDataBuffers;
 
 		this.nextBuffer = dataReader.nextBuffer();
+		this.availabilityListener = checkNotNull(availabilityListener);
+	}
+
+	BoundedBlockingSubpartitionReader(
```
pnowojski: please re-use one of the constructors in the the other

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/FileChannelBoundedData.java
```diff
@@ -155,6 +159,10 @@ public void close() throws IOException {
 		@Override
 		public void recycle(MemorySegment memorySegment) {
 			buffers.addLast(memorySegment);
+
+			if (!isFinished) {
+				subpartitionView.notifyDataAvailable();
```
pnowojski: Hmm, something is not right here.
1. there is a cyclic dependency between `FileBufferReader` and `BoundedBlockingSubpartitionReader`. Here we are calling `BoundedBlockingSubpartitionReader` which in turns call us back via `FileBufferReader#nextBuffer`
2. it's strange (when you say it out laud) that we are performing an io read in `recycle()` call stack
3. it's strange that `notifyDataAvailable` call is doing some io operation
4. it's also a bit strange that `recycle()` call is triggering `notifyDataAvailable()`, but this maybe unavoidable

for 2. and 3. names completely do not suggest what the methods are actually doing, and it is quite surprising that a light sounding method is doing some heavy work.

Can not we just notify the view, that we have memory to progress, and we would make io calls and maybe fill out read ahead buffer during next `BoundedBlockingSubpartitionReader#getNextBuffer` call? That would fix some of the problems that I mentioned: it would brake the cycle and there would be no io operations during recycle/data notification calls.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartitionReader.java
zhijiangW: Do you mean checking via `checkState(dataReader instanceof FileChannelBoundedData.FileBufferReader)` when next buffer is null?

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/FileChannelBoundedData.java
zhijiangW: In `recycle()` method, it only calls `notifyDataAvailable`, so we would not see io operation from this layer explicitly. So The key concern is that the implementation of `notifyDataAvailable` would be involved in io operation.

Actually `BoundedBlockingSubpartitionReader#getNextBuffer` and `notifyDataAvailable` are both called by netty thread, so it is no difference of io operation from caller aspect.

From the aspect of method naming semantic, it might seem a bit strange to make `notifyDataAvailable` do io operation. Actually I ever considered your suggested way during implementation, but there are some other problems if we do so:

- In `BoundedBlockingSubpartitionReader`, next buffer is also used for other places.  In the case of `nextBufferIsEvent` method, if we do not read ahead the next buffer during `notifyDataAvailable`, we do not know whether it is event or not.  Then in theory we still could not transport this event buffer if no credits.

- In the case of `BoundedBlockingSubpartitionReader#isAvailable()`, we still need to check whether the next buffer is null after `notifyDataAvailable` which keeps consistent with streaming mode and other types in batch mode. So we have to read ahead the next buffer before we actually notify view available to the netty stack.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartitionReader.java
zhijiangW: Do you mean adding `checkState` in `BoundedBlockingSubpartitionReader#getNextBuffer` while next buffer is null? If so, I think it might be not necessary adding it in `getNextBuffer` which is consistent for all the reader types. We could add the check in following `notifyDataAvailable` as you suggested if necessary, because only this method should be only used for file type.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartitionReader.java
zhijiangW: I actually considered this issue. But because of `data.createReader(this)` we could not re-use another constructor directly via `this`. Or we define another `initialize` method for extracting the common codes of two constructors?

StephanEwen:
Looking at this now...

StephanEwen:
My first thought is that it would be great if we could avoid making the bounded data readers aware of the subpartition views.

The way that this pull request changes the code, we do not have the property any more that we gain an abstraction level when going from bounded data to subpartition view.

StephanEwen:
I will try and see if I can suggest a way to decouple this...

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/BoundedBlockingSubpartitionReader.java
zhijiangW: Only retain one constructor for fixing.

zhijiangW:
@pnowojski @StephanEwen I have rebased the codes and made some modifications:

- Adds the unit tests Stephan provided in his branch.

- Make `BoundedBlockingSubpartitionType` configurable in ITCase/tests.

- Add a new ITCase for verifying this fix. This case would throw `IOException` frequently before fixing, and after fixing it could run correctly.

zhijiangW:
Also addressed the commit message and constructor issue for Piotr's comments.

flink-tests/src/test/java/org/apache/flink/test/runtime/FileBufferReaderITCase.java
```diff
@@ -0,0 +1,183 @@
+/*
```
pnowojski: I think it would be better to squash the ITCase commit with the bug fix. Whenever I add a test coverage for some bug, I try to keep it together, as the test might provide some extra context for whoever will be reading the bug fix commit in the future, like how the bug was triggered before the bug fix, what were the conditions/setup, etc.

flink-tests/src/test/java/org/apache/flink/test/runtime/FileBufferReaderITCase.java
```diff
@@ -0,0 +1,183 @@
...
+
+import java.util.concurrent.CompletableFuture;
+
+import static org.hamcrest.Matchers.is;
+import static org.junit.Assert.assertThat;
+
+/**
+ * Tests the bug reported in FLINK-131O0.
+ *
+ * <p>The implementation of FileChannelBoundedData#nextBuffer assumes that there is always an available buffer,
```
pnowojski: When you refer in the java doc comments to a classes I think it's generally speaking better to refer them via `{@link FileChannelBoundedData#nextBuffer}` java doc's feature. That way when code is renamed (via IDE supported refactor/rename feature), the java doc remains accurate automatically.

Could you add the missing `{@link XYZ}` links in your java doc? Like this, or ` FileChannelBoundedData#nextBuffer`, `writeAndFlush`. Maybe `IOException` is debatable nit, as it's unlikely to be changed.

flink-tests/src/test/java/org/apache/flink/test/runtime/FileBufferReaderITCase.java
zhijiangW: Yes, I agree with squashing this ITCase commit with core commit when merging. Make it separate now is used for verifying this fix work.

flink-tests/src/test/java/org/apache/flink/test/runtime/FileBufferReaderITCase.java
zhijiangW: This javadoc is actually copied directly from commit message, forgot adding {@link}.

zhijiangW:
@pnowojski I have squashed the commit and addressed the javadoc comment.

zhijiangW:
Thanks for merging @pnowojski . I would update the ticket.

PR #9064, opened 2019-07-10
[FLINK-13188][Runtime / State Backends][Test] Fix MockStateBackend#resolveCheckpointStorageLocation return null cause intellij assertion detect test failed

mrzhangboss:

*Fix test([MockTtlStateTest](https://github.com/apache/flink/blob/master/flink-runtime/src/test/java/org/apache/flink/runtime/state/ttl/MockTtlStateTest.java))  failed in intellij with supporting run time assertions opening *

## Brief change log

  - *change annotation type in [MockKeyedStateBackend](https://github.com/apache/flink/blob/master/flink-runtime/src/test/java/org/apache/flink/runtime/state/ttl/mock/MockKeyedStateBackend.java)*
 

## Verifying this change

This change is already covered all existing tests

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-runtime/src/test/java/org/apache/flink/runtime/state/ttl/mock/MockKeyedStateBackend.java
+++ b/flink-runtime/src/test/java/org/apache/flink/runtime/state/ttl/mock/MockKeyedStateBackend.java
@@ -54,6 +54,7 @@ import org.apache.flink.runtime.state.ttl.TtlTimeProvider;
 import org.apache.flink.util.FlinkRuntimeException;
 
 import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 
 import java.util.ArrayList;
 import java.util.HashMap;
@@ -178,7 +179,7 @@ public class MockKeyedStateBackend<K> extends AbstractKeyedStateBackend<K> {
 	public RunnableFuture<SnapshotResult<KeyedStateHandle>> snapshot(
 		long checkpointId,
 		long timestamp,
-		@Nonnull CheckpointStreamFactory streamFactory,
+		@Nullable CheckpointStreamFactory streamFactory,
 		@Nonnull CheckpointOptions checkpointOptions) {
 		return new FutureTask<>(() ->
 			SnapshotResult.of(new MockKeyedStateHandle<>(copy(stateValues, stateSnapshotFilters))));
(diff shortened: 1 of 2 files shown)


flink-runtime/src/test/java/org/apache/flink/runtime/state/ttl/mock/MockKeyedStateBackend.java
```diff
@@ -178,7 +179,7 @@ public void notifyCheckpointComplete(long checkpointId) {
 	public RunnableFuture<SnapshotResult<KeyedStateHandle>> snapshot(
 		long checkpointId,
 		long timestamp,
-		@Nonnull CheckpointStreamFactory streamFactory,
+		@Nullable CheckpointStreamFactory streamFactory,
```
klion26: I don't think this is the right direction to fix this.
Could you please share why `streamFactory` is null here?

mrzhangboss: Thanks for review !
 In [MockStateBackend](https://github.com/apache/flink/blob/master/flink-runtime/src/test/java/org/apache/flink/runtime/state/ttl/mock/MockStateBackend.java) when call `snapshot` function it always create a `CheckpointStorage` object with  a `resolveCheckpointStorageLocation` function.In `resolveCheckpointStorageLocation` it only return null back. Maybe just fix this `resolveCheckpointStorageLocation` function in  [MockStateBackend](https://github.com/apache/flink/blob/master/flink-runtime/src/test/java/org/apache/flink/runtime/state/ttl/mock/MockStateBackend.java)  will better. What do you think.

klion26: I can't reproduce your problem locally, All the tests passed on my laptop.
But I think If this failed on your machine, we may better make the `streamFactory` not null instead of marking it as `@Nullable`. So yes, as you analyzed, we may fix `MockStateBackend#resolveCheckpointStorageLocation`.

mrzhangboss: Thank you for reply. I wonder if you turn on run time assertion in intellij as I mentioned in [Jira](https://issues.apache.org/jira/browse/FLINK-13188). In my computer, if I close this assertion detected button and rebuild project, I also can pass all the tests.

By the ways, I put a new commit to fix `MockStateBackend#resolveCheckpointStorageLocation`,  could you help have a look .

PR #9065, opened 2019-07-10
[FLINK-13102][travis] Optimize some travis stages by skipping mvn verify if possible

1u0:

This PR minimizes total Travis CI time for pull request triggered builds by skipping `mvn verify` steps for non `python` and `misc` jobs in Travis `test` stage.

## Brief change log

  - added script to build `mvn` dependencies tree and list all changed files in a PR build run. This happens at the very first `compile` stage;
  - added script to detect if `mvn verify` can be skipped for given list of maven modules and list of changed files.

## Verifying this change

  - run different PRs (build in Travis CI) with this changes included and manually verifying that build is correct and some tests are not run because changes in a PR are not affecting them.

## Documentation

### DIFF
--- a/tools/travis/differential_build.sh
+++ b/tools/travis/differential_build.sh
@@ -56,3 +56,15 @@ function prepare_dependencies_tree_if_pr_build() {
 
     python3 ./tools/travis/mvn_diff_test.py process-mvn-dependencies mvn-dependencies-list.txt mvn-basedir-list.txt > "$MVN_DEPENDENCIES_TREE_FILEPATH"
 }
+
+function can_skip_mvn_test_run_for_stage() {
+    local stage=$1
+
+    if ! is_pr_build ; then
+        echo "Cannot skip non PR test run"
+        return 1
+    fi
+
+    python3 ./tools/travis/mvn_diff_test.py can-skip-mvn-test "$stage" "$PROFILE" "$MVN_DEPENDENCIES_TREE_FILEPATH" "$DIFF_LIST_FILEPATH"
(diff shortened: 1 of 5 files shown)


1u0:
CC: @zentol 

Some notes:

With this PR, there is a code duplication in `tools/travis/stage.sh` and `tools/travis/stages.py` (they both have mapping which maven modules to build for a given stage). I want to remove it (my personal preference keeping python version), but would like to check first if python is fine for others.

The dependencies detection heuristic relies on that inner module (artifacts) dependencies in Flink are explicit (declared in `pom.xml` files).
The dependencies check can be further improved by taking into account maven test and non-test dependencies. But for now, the implementation in favor of simplicity and to avoid false skippings (it's ok if we run tests, even if changes are not relevant; it's not ok, if we don't run tests, if changes are relevant and detection missed it).

1u0:
@zentol, fyi, unfortunately, this optimization doesn't take effect for the new CI setup (running via `/flink-ci/flink`), because the new setup performs builds as standalone branches rather than as PRs (from Travis CI point of view).

The `TRAVIS_COMMIT_RANGE` environment variable would be empty in that case. And I don't know yet how to figure out  git commits ranges to list changed files.

zentol:
this isn't the only issue in regards to the new setup; the `is_pr_build` function won't work as expected.

I have ideas for fixing both issues, but I'll need some time to figure things out, and I'm not sure if I will have it this or next week with the release coming up.

zentol:
The ongoing move

PR #9068, opened 2019-07-10
[FLINK-13195] Add create table support for SqlClient

danny0405:

Support create table for SqlClient

## Brief change log

  - Add create table command support for SqlClient
  - Always cache the current catalog to the original session, so that when we set a variable, the current catalog table still can be reused

## Verifying this change

See tests in LocalExecutorITCase

  - The serializers: no

## Documentation

### DIFF
--- a/flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/cli/CliClient.java
+++ b/flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/cli/CliClient.java
@@ -281,6 +281,9 @@ public class CliClient {
 			case USE_DATABASE:
 				callUseDatabase(cmdCall);
 				break;
+			case CREATE_TABLE:
+				callCreateTable(cmdCall);
+				break;
 			case DESCRIBE:
 				callDescribe(cmdCall);
 				break;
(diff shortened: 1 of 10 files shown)


xuefuz:
I appreciate Timo's vision of client/server model even in case that we only have local execution at the moment. However, I'm a little curious about the goal of a stateless gateway with state being passed back and forth. Some state is easily fit to this model, while others, such as temporary tables created by the user that's further referenced in subsequent queries. This is just one thing that's currently maintained in a table env. Without a live table env instance maintained for a remote client, it's hard to maintain the state via a session context.

To me, stateless gateway seems making a lot of sense for largely stateless client, like one submitting a streaming job. The scenario for batch can be quite different.

Personally, I am in favor of having some DDL support now and leaving re-architecturing for the next release.

danny0405:
@twalthr How about we cache a `catalogName -> DDLs` mapping in the `SessionContext`, just like we cache the `ViewEntry`, then we recover the tables from these DDLs every time we switch to a new session.

twalthr:
Tables registered in catalogs should have been persisted. So we don't need to memorize them. It is true that `SessionContext` has some cache for views, but if you look into the implementation, a view is just a string that comes directly from the CLI. It is not evaluated before submitting an entire query.

I agree that we should not re-register tables in a catalog. A `CREATE TABLE` statement should be executed immediately and be persistent in the catalog. Have we ever thought about properly distinguishing between temporary tables (for the session) and persistent tables (across sessions)? This would make the implementation in the SQL Client much easier:

```
CREATE TEMPORARY TABLE (...) WITH (..) // buffered in the CLI session and re-registered when a query is executed

CREATE TABLE (...) WITH (...) // executed immediately and applied to the catalog
```


This is one idea of fixing it properly.

xuefuz:
@twalthr Your example of temporary table usage might be simple enough to make client-side caching workable. However, that's usually not the case. For instance, user might do
CREATE TEMPORARY TABLE (...) WITH (..) AS SELECT ... FROM X JOIN Y ON .... WHERE ...
it's not efficient to cache this at client side and execute every time when the temp table is referenced.

The essence of the problem is that a SQL user session has a complicated state that cannot be easily managed by a client. Thus, the idea of stateless gateway, while solving some use cases (like submitting a streaming job), has difficulty to apply for batch cases where session state is large and complicated.

PR #9069, opened 2019-07-10
[FLINK-12297] Fixes StackOverflowException of ClosureCleaner when cle…

xiaoguichao:
@aljoscha


StephanEwen:
This PR needs a proper description before being picked up for reviewing.

xiaoguichao:
@StephanEwen This is an imported bug by the implementation of [issue FLINK-12297](https://issues.apache.org/jira/browse/FLINK-12297).

As described there, the recursive closure cleaner will get  an stackoverflow exception when cleaning a back-referenced object (objectMapper in JsonRowDeserializationSchema, for example)

PR #9071, opened 2019-07-10
FLINK-13044 [BuildSystem / Shaded] Fix for wrong shading of AWS SDK in flink-s3-fs-hadoop

derjust:

Due to the bug MSHADE-156 in Maven's shading plugin [1] the merging of the AWS SDK also causes string literals to be changed in Flink classes:

`PACKAGE_PREFIXES_TO_SHADE` in `org.apache.flink.fs.s3hadoop.S3FileSystemFactory` from `com.amazonaws.` to `org.apache.flink.fs.s3base.shaded.com.amazonaws` - which causes settings like `fs.s3a.aws.credentials.provider: com.amazonaws.auth.DefaultAWSCredentialsProviderChain`
to not be remapped properly as the `startsWith` in `org.apache.flink.fs.s3.common.HadoopConfigLoader.shadeClassConfig()` doesn't match the beginning of the FQCN.

Using `DefaultAWSCredentialsProviderChain` instead of Flink's hand-crafted chain can be required depending on the Flink deployment - i.e. as part of Fargate.

This issue is only visible looking into the compiled class file after the shading happened!

## Brief change log

* Fixing the `PACKAGE_PREFIXES_TO_SHADE` value alone doesn't help as based on `FLINK_SHADING_PREFIX` the FQCN is assembled as `org.apache.flink.fs.s3hadoop.shaded.com.amazonaws.auth.DefaultAWSCredentialsProviderChain`
which can't be found either; the class is located at `org.apache.flink.fs.s3base.shaded.com.amazonaws.auth.DefaultAWSCredentialsProviderChain`

    * This commit follows the workaround shown in [1] to workaround the shading issue.
* Also fixing the wrong package name in the Flink code for the shaded AWS SDK classes.

## Verifying this change

- *Set `fs.s3a.aws.credentials.provider: com.amazonaws.auth.DefaultAWSCredentialsProviderChain` in `flink-conf.yaml`*
- *Start JobManager and see it using the `DefaultAWSCredentialsProviderChain` instead of Flink's handcrafted chain*

  - The serializers: no
  - The S3 file system connector: yes

## Documentation

  - Does this pull request introduce a new feature?   no

[1] https://issues.apache.org/jira/browse/MSHADE-156

### DIFF
--- a/flink-annotations/pom.xml
+++ b/flink-annotations/pom.xml
@@ -25,7 +25,7 @@ under the License.
 	<parent>
 		<groupId>org.apache.flink</groupId>
 		<artifactId>flink-parent</artifactId>
-		<version>1.9-SNAPSHOT</version>
+		<version>1.8.0</version>
 		<relativePath>..</relativePath>
 	</parent>
 
--- a/flink-clients/pom.xml
+++ b/flink-clients/pom.xml
@@ -26,7 +26,7 @@ under the License.
 	<parent>
 		<groupId>org.apache.flink</groupId>
 		<artifactId>flink-parent</artifactId>
-		<version>1.9-SNAPSHOT</version>
+		<version>1.8.0</version>
 		<relativePath>..</relativePath>
 	</parent>
 
--- a/flink-connectors/flink-connector-cassandra/pom.xml
+++ b/flink-connectors/flink-connector-cassandra/pom.xml
@@ -26,7 +26,7 @@ under the License.
 	<parent>
 		<groupId>org.apache.flink</groupId>
 		<artifactId>flink-connectors</artifactId>
-		<version>1.9-SNAPSHOT</version>
+		<version>1.8.0</version>
 		<relativePath>..</relativePath>
 	</parent>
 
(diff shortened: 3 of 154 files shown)


zentol:
OK, let's see if I got this correctly:

There are 2 issues here:
1) The shade-plugin relocates `S3FileSystemFactory#PACKAGE_PREFIXES_TO_SHADE`, as a result of which no value for `fs.s3a.aws.credentials.provider` that starts with `com.amazonaws` is relocated to the shaded namespace when creating the hadoop config in the `S3FileSystemFactory`.
2) The shading pattern configured for `com.amazonaws` is not synced between the shade-plugin and the `S3FileSystemFactory`. Even if 1) were fixed, we would still shade the configured value into the wrong namespace.

Am I correct that we can test this by modifying our a3 end-to-end tests to set `fs.s3a.aws.credentials.provider: com.amazonaws.auth.DefaultAWSCredentialsProviderChain`?

zentol:
If so, I'm wondering if we could solve this without having to rely on the shade-plugin removing the `UNSHADE` portion. Ultimately we just have to obfuscate the string constant so that the shade-plugin doesn't relocate it at all. This could be as simple as
```
PACKAGE_PREFIXES_TO_SHADE = Collections.singleton("com.UNSHADE.".replace("UNSHADE", "amazonaws");)
```

zentol:
Subsumed by #9318.

PR #7757, opened 2019-02-19
[FLINK-11630] Triggers the termination of all running Tasks when shutting down TaskExecutor

kisimple:
## What is the purpose of the change

Currently, the `TaskExecutor` does not properly wait for the termination of `Task`s when terminating. This patch triggers the termination of all running `Task`s when shutting down `TaskExecutor`.

## Brief change log

  - Make `Task#stopExecution` returns a `CompletableFuture`
  - Make `Task#cancelExecution` returns a `CompletableFuture`
  - Call `Task#cancelExecution` when invoking `TaskExecutor#onStop`

## Verifying this change

This change is already covered by existing tests.

## Does this pull request potentially affect one of the following parts:

  - Dependencies (does it add or upgrade a dependency): (no)
  - The public API, i.e., is any changed class annotated with `@Public(Evolving)`: (no)
  - The serializers: (no)
  - The runtime per-record code paths (performance sensitive): (no)
  - Anything that affects deployment or recovery: JobManager (and its components), Checkpointing, Yarn/Mesos, ZooKeeper: (no)
  - The S3 file system connector: (no)

## Documentation

  - Does this pull request introduce a new feature? (no)

PR #9072, opened 2019-07-10
[FLINK-11630] Wait for the termination of all running Tasks when shutting down TaskExecutor

azagrebin:

Add a unit test to #7757.

Currently, the `TaskExecutor` does not properly wait for the termination of Tasks when it shuts down in `TaskExecutor#onStop`. This patch changes `TaskExecutor#onStop` to fail all running tasks and wait  for their termination before stopping all services.

## Brief change log

  - add `TaskCompletionTracker` to track task termination futures in `TaskExecutor`
  - try to fail all running tasks and wait  for their termination before stopping all services
  - add `TaskExecutorTest.testTaskInterruptionAndTerminationOnShutdown`

## Verifying this change

run `TaskExecutorTest.testTaskInterruptionAndTerminationOnShutdown`

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/taskexecutor/TaskExecutor.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/taskexecutor/TaskExecutor.java
@@ -132,6 +132,8 @@ import java.util.Objects;
 import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionException;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.TimeoutException;
 import java.util.function.BiConsumer;
 import java.util.stream.Collectors;
@@ -181,6 +183,8 @@ public class TaskExecutor extends RpcEndpoint implements TaskExecutorGateway {
 
 	/** The kvState registration service in the task manager. */
 	private final KvStateService kvStateService;
+
+	private final TaskCompletionTracker taskCompletionTracker;
 
 	// --------- job manager connections -----------
 
@@ -273,6 +277,7 @@ public class TaskExecutor extends RpcEndpoint implements TaskExecutorGateway {
 		this.currentRegistrationTimeoutId = null;
 
 		this.stackTraceSampleService = new StackTraceSampleService(rpcService.getScheduledExecutor());
+		this.taskCompletionTracker = new TaskCompletionTracker();
 	}
 
 	@Override
@@ -333,31 +338,46 @@ public class TaskExecutor extends RpcEndpoint implements TaskExecutorGateway {
 	public CompletableFuture<Void> onStop() {
 		log.info("Stopping TaskExecutor {}.", getAddress());
 
-		Throwable throwable = null;
+		Throwable jobManagerDisconnectThrowable = null;
 
 		if (resourceManagerConnection != null) {
 			resourceManagerConnection.close();
 		}
 
+		FlinkException cause = new FlinkException("The TaskExecutor is shutting down.");
 		for (JobManagerConnection jobManagerConnection : jobManagerConnections.values()) {
 			try {
-				disassociateFromJobManager(jobManagerConnection, new FlinkException("The TaskExecutor is shutting down."));
+				disassociateFromJobManager(jobManagerConnection, cause);
 			} catch (Throwable t) {
-				throwable = ExceptionUtils.firstOrSuppressed(t, throwable);
+				jobManagerDisconnectThrowable = ExceptionUtils.firstOrSuppressed(t, jobManagerDisconnectThrowable);
 			}
 		}
 
-		try {
-			stopTaskExecutorServices();
-		} catch (Exception e) {
-			throwable = ExceptionUtils.firstOrSuppressed(e, throwable);
+		final Throwable throwableBeforeTasksCompletion = jobManagerDisconnectThrowable;
+
+		return FutureUtils
+			.runAfterwards(
+				taskCompletionTracker.failIncompleteTasksAndGetTerminationFuture(),
+				this::stopTaskExecutorServices)
+  		    .handle((ignored, throwable) -> {
+  		    	handleOnStopException(throwableBeforeTasksCompletion, throwable);
+  		    	return null;
+			});
+	}
+
+	private void handleOnStopException(Throwable throwableBeforeTasksCompletion, Throwable throwableAfterTasksCompletion) {
+		final Throwable throwable;
+
+		if (throwableBeforeTasksCompletion != null) {
+			throwable = ExceptionUtils.firstOrSuppressed(throwableBeforeTasksCompletion, throwableAfterTasksCompletion);
+		} else {
+			throwable = throwableAfterTasksCompletion;
 		}
 
 		if (throwable != null) {
-			return FutureUtils.completedExceptionally(new FlinkException("Error while shutting the TaskExecutor down.", throwable));
+			throw new CompletionException(new FlinkException("Error while shutting the TaskExecutor down.", throwable));
 		} else {
 			log.info("Stopped TaskExecutor {}.", getAddress());
-			return CompletableFuture.completedFuture(null);
 		}
 	}
 
@@ -596,6 +616,7 @@ public class TaskExecutor extends RpcEndpoint implements TaskExecutorGateway {
 
 			if (taskAdded) {
 				task.startTaskThread();
+				taskCompletionTracker.trackTaskCompletion(task);
 
 				setupResultPartitionBookkeeping(tdd, task.getTerminationFuture());
 				return CompletableFuture.completedFuture(Acknowledge.get());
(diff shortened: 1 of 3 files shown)


azagrebin:
@flinkbot attention @tillrohrmann @kisimple

ifndef-SleePy:
Hi @azagrebin ,

I'm following the unstable case https://issues.apache.org/jira/browse/FLINK-11631 which is relevant to this PR. 

I guess you have taken over the old PR https://github.com/apache/flink/pull/7757. I have a concern same with @tillrohrmann , there are two different tracking ways of task lifecycle. One is `unregisterTaskAndNotifyFinalState` and `TaskSlotTable`, the other is `TaskCompletionTracker`.

I think this PR could work well at the moment. I'm just afraid we might encounter some subtle corner cases in the future. Maybe we should unify these two lifecycle tracking ways. Or we could merge this PR first, then think about unification. What do you think?

azagrebin:
Hi @ifndef-SleePy 
If you mean this concern https://github.com/apache/flink/pull/7757#pullrequestreview-240008731 about forgetting to wait for some tasks, it is because the old PR used `TaskSlotTable` to get uncompleted tasks at some point of time. The problem is that `TaskSlotTable` might not always track tasks synchronised with their completion futures as it addresses other concerns and there are can be race conditions. I also left more explanation about it on the old PR https://github.com/apache/flink/pull/7757#issuecomment-496118950. This is the reason why the old PR was changed and this PR uses a separate tracking which explicitly removes the task only if it is completed. The completion future is always at the very end of Task.run and is called always on Task.run exit to not miss any action in Task.run (in this case release of resources is important).

ifndef-SleePy:
Hi @azagrebin 
Thank you for explanation. Nice work to take the old PR over!
Anyway, for this PR, it looks good to me.

flink-runtime/src/main/java/org/apache/flink/runtime/taskexecutor/TaskExecutor.java
```diff
@@ -333,32 +338,38 @@ private void handleStartTaskExecutorServicesException(Exception e) throws Except
...
-
-		if (throwable != null) {
-			return FutureUtils.completedExceptionally(new FlinkException("Error while shutting the TaskExecutor down.", throwable));
-		} else {
-			log.info("Stopped TaskExecutor {}.", getAddress());
-			return CompletableFuture.completedFuture(null);
-		}
+		final Throwable throwableBeforeTasksCompletion = jobManagerDisconnectThrowable;
+		return taskCompletionTracker
+			.failIncompleteTasksAndWaitForAllCompleted()
```
tillrohrmann: let's rename this method into `failIncompleteTasksAndGetTerminationFuture()`

flink-runtime/src/main/java/org/apache/flink/runtime/taskexecutor/TaskExecutor.java
```diff
@@ -333,32 +338,38 @@ private void handleStartTaskExecutorServicesException(Exception e) throws Except
...
-		}
-
-		if (throwable != null) {
-			return FutureUtils.completedExceptionally(new FlinkException("Error while shutting the TaskExecutor down.", throwable));
-		} else {
-			log.info("Stopped TaskExecutor {}.", getAddress());
-			return CompletableFuture.completedFuture(null);
-		}
+		final Throwable throwableBeforeTasksCompletion = jobManagerDisconnectThrowable;
+		return taskCompletionTracker
```
tillrohrmann:
```suggestion
		final CompletableFuture<Void> taskTerminationFuture = taskCompletionTracker.failIncompleteTasksAndGetTerminationFuture();
		return FutureUtils
		    .runAfterwards(
		        taskTerminationFuture,
  		        () -> stopTaskExecutorServices())
  		    .handle(
  		        (ignored, throwable) -> {
  		            if (throwable != null) {
  		                throw new CompletionException(new FlinkException("Error while shutting the TaskExecutor down.", throwable));
  		            } else {
  		                log.info("Stopped TaskExecutor {}.", getAddress());
  		                return null;
  		            }
  		        });
```

tillrohrmann: Please do proper formatting of my suggestion.

flink-runtime/src/main/java/org/apache/flink/runtime/taskexecutor/TaskExecutor.java
```diff
@@ -1826,4 +1838,30 @@ public SlotReport retrievePayload(ResourceID resourceID) {
 			return taskSlotTable.createSlotReport(getResourceID());
 		}
 	}
+
+	private static class TaskCompletionTracker {
+		private final Map<ExecutionAttemptID, Task> incompleteTasks;
+
+		private TaskCompletionTracker() {
+			incompleteTasks = new ConcurrentHashMap<>(8);
```
tillrohrmann: Why do we need a `ConcurrentHashMap`? Where does the concurrency come from?

flink-runtime/src/main/java/org/apache/flink/runtime/taskexecutor/TaskExecutor.java
```diff
@@ -1826,4 +1838,30 @@ public SlotReport retrievePayload(ResourceID resourceID) {
...
+		}
+
+		void trackTaskCompletion(Task task) {
+			incompleteTasks.put(task.getExecutionId(), task);
+			task.getTerminationFuture().thenRun(() -> incompleteTasks.remove(task.getExecutionId()));
```
tillrohrmann: could we also rely on `TaskExecutor#unregisterTaskAndNotifyFinalState` to remove the task from the `TaskCompletionTracker`? That way we would not apply concurrent changes to `incompleteTasks`.

flink-runtime/src/test/java/org/apache/flink/runtime/taskexecutor/TaskExecutorTest.java
```diff
@@ -760,6 +767,79 @@ public void invoke() throws Exception {
...
+			createTaskDeploymentDescriptor(TestInterruptableInvokable.class, allocationId);
+		final TaskExecutor taskExecutor = createTaskExecutor(new TaskManagerServicesBuilder()
+			.setTaskSlotTable(taskSlotTable)
+			.setJobManagerTable(createJobManagerTableWithOneJob(jobMasterId))
+			.build());
+
+		try {
+			taskExecutor.start();
+			taskSlotTableStarted.get();
+			taskSlotTable.allocateSlot(0, jobId, allocationId, timeout);
```
tillrohrmann: This is dangerous, as we are mutating state from the testing thread which is actually owned by the `TaskExecutor's` rpc main thread. Better to call `TaskExecutorGateway#requestSlot` after registering a `ResourceManagerGateway`.

flink-runtime/src/test/java/org/apache/flink/runtime/taskexecutor/TaskExecutorTest.java
```diff
@@ -760,6 +767,79 @@ public void invoke() throws Exception {
...
+				.submitTask(taskDeploymentDescriptor, jobMasterId, timeout);
+			TestInterruptableInvokable.STARTED_FUTURE.get();
+		} finally {
+			RpcUtils.terminateRpcEndpoint(taskExecutor, timeout);
+		}
+		taskExecutor.getTerminationFuture().get();
+
+		Task task = taskSlotTable.getTask(taskDeploymentDescriptor.getExecutionAttemptId());
+		assertThat(task.getTerminationFuture().isDone(), is(true));
+		assertThat(TestInterruptableInvokable.INTERRUPTED_FUTURE.isDone(), is(true));
```
tillrohrmann: Can't we change the test so that we have control over the submitted task's termination future `t`. Given that, we could stop the `TaskExecutor` and check that it has not terminated if the task's termination future `t` is not completed. Then we complete `t` and check that the `TaskExecutor` terminates. Then there would also be no need to interact directly with the `TaskSlotTable` which is an implementation detail of the `TaskExecutor`.

flink-runtime/src/test/java/org/apache/flink/runtime/taskexecutor/TaskExecutorTest.java
```diff
@@ -760,6 +767,79 @@ public void invoke() throws Exception {
...
+			ContextClassLoaderLibraryCacheManager.INSTANCE,
+			new NoOpResultPartitionConsumableNotifier(),
+			(j, i, r) -> CompletableFuture.completedFuture(null));
+
+		final JobManagerTable jobManagerTable = new JobManagerTable();
+		jobManagerTable.put(jobId, jobManagerConnection);
+		return jobManagerTable;
+	}
+
+	private TaskDeploymentDescriptor createTaskDeploymentDescriptor(
```
tillrohrmann: What about introducing a `TaskDeploymentDescriptorBuilder` and replacing the different test instantiations of `TDDs` with that?

flink-runtime/src/main/java/org/apache/flink/runtime/taskexecutor/TaskExecutor.java
ifndef-SleePy: This is also my concern. We already have `TaskExecutor#unregisterTaskAndNotifyFinalState` and `Task#getTerminationFuture`. I think we should unify them somehow.

tillrohrmann: Andrey explained to me that we call `unregisterTaskAndNotifyFinalState` before the clean up of the `Task` has completed. I think this is a problem in the `Task's` lifecycle. Hence, one could fix this as a follow up and then simply use the `TerminationFuture` to signal the final state to the `TaskExecutor`. But this is out of scope for this PR.

ifndef-SleePy: Yes, Andrey has explained to me before.
Never mind, it's just a mention when I saw this comment. I agree that we could follow up on it later.
BTW, we could also avoid concurrent modifications through `thenRunAsync` with `mainThreadExecutor` if we prefer relying on `task.getTerminationFuture()`.

azagrebin:
Thanks for the review @tillrohrmann, I have addressed comments.

PR #9074, opened 2019-07-11
[FLINK-13198] Introduce TimeUtils with parsing method

TsReaper:

We need a parse function to parse strings like "123ms", "321s", "12min" and such, so that users can write configurations like "cache.ttl = 1min" and the user input can be changed into a Java `Duration`.

This pull request introduce `TimeUtils` with `parseDuration` method to meet the above needs.

## Brief change log

  - Introduce `TimeUtils`

## Verifying this change

This change can be covered by new tests (`TimeUtilsTest`).

  - The public API, i.e., is any changed class annotated with @Public(Evolving)`: no
  - The serializers: no

## Documentation

### DIFF
--- a/flink-table/flink-table-common/src/main/java/org/apache/flink/table/descriptors/DescriptorProperties.java
+++ b/flink-table/flink-table-common/src/main/java/org/apache/flink/table/descriptors/DescriptorProperties.java
@@ -29,8 +29,10 @@ import org.apache.flink.table.utils.EncodingUtils;
 import org.apache.flink.table.utils.TypeStringUtils;
 import org.apache.flink.util.InstantiationUtil;
 import org.apache.flink.util.Preconditions;
+import org.apache.flink.util.TimeUtils;
 
 import java.math.BigDecimal;
+import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -548,6 +550,26 @@ public class DescriptorProperties {
 	}
 
 	/**
+	 * Returns a Java {@link Duration} under the given key if it exists.
+	 */
+	public Optional<Duration> getOptionalDuration(String key) {
+		return optionalGet(key).map((value) -> {
+			try {
+				return TimeUtils.parseDuration(value);
+			} catch (Exception e) {
+				throw new ValidationException("Invalid duration value for key '" + key + "'.", e);
+			}
+		});
+	}
+
+	/**
+	 * Returns a java {@link Duration} under the given existing key.
+	 */
+	public Duration getDuration(String key) {
+		return getOptionalDuration(key).orElseThrow(exceptionSupplier(key));
+	}
+
+	/**
 	 * Returns the property keys of fixed indexed properties.
 	 *
 	 * <p>For example:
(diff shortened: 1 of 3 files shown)


flink-core/src/test/java/org/apache/flink/configuration/TimeLengthTest.java
```diff
@@ -0,0 +1,200 @@
...
+		assertEquals(1234, TimeLength.parse("1234").getMilliseconds());
+		assertEquals(1234, TimeLength.parse("1234ms").getMilliseconds());
+		assertEquals(1234, TimeLength.parse("1234 ms").getMilliseconds());
+	}
+
+	@Test
+	public void testParseSeconds() {
+		assertEquals(667766, TimeLength.parse("667766s").getSeconds());
+		assertEquals(667766, TimeLength.parse("667766 s").getSeconds());
+		assertEquals(667766, TimeLength.parse("667766s").getSeconds());
```
wuchong: The same with L94?

flink-table/flink-table-common/src/main/java/org/apache/flink/table/descriptors/DescriptorProperties.java
```diff
@@ -540,13 +541,33 @@ public TableSchema getTableSchema(String key) {
...
 		return getOptionalMemorySize(key).orElseThrow(exceptionSupplier(key));
 	}
 
+	/**
+	 * Returns a Flink {@link TimeLength} under the given existing key.
+	 */
+	public TimeLength getTimeLength(String key) {
+		return getOptionalTimeLength(key).orElseThrow(exceptionSupplier(key));
+	}
+
```
wuchong: Please also add `validateTimeLength` similar to `validateMemorySize`.

flink-core/src/test/java/org/apache/flink/configuration/TimeLengthTest.java
```diff
@@ -0,0 +1,200 @@
...
+	}
+
+	@Test
+	public void testParseMinutes() {
+		assertEquals(7657623, TimeLength.parse("7657623min").getMinutes());
+		assertEquals(7657623, TimeLength.parse("7657623 min").getMinutes());
+		assertEquals(7657623, TimeLength.parse("7657623min").getMinutes());
```
wuchong: The same with L101?

flink-core/src/test/java/org/apache/flink/configuration/TimeLengthTest.java
```diff
@@ -0,0 +1,200 @@
...
+	}
+
+	@Test
+	public void testPraseHours() {
+		assertEquals(987654, TimeLength.parse("987654h").getHours());
+		assertEquals(987654, TimeLength.parse("987654 h").getHours());
+		assertEquals(987654, TimeLength.parse("987654h").getHours());
```
wuchong: The same with L108?

wuchong:
cc @zentol , I think you may also want to have a look at this?

wuchong:
Thanks @TsReaper . 

LGTM.

wuchong:
Thanks @TsReaper for the quick updating.

TsReaper:
Thanks @twalthr for the comment. I think `org.apache.flink.api.common.time.Time` has a different functionality compared to the newly introduced `TimeInterval`, so it's ok to keep both the two classes.

zentol:
I'D like to get to the drawing board for this one; it seems this could be trivially solved using Java 8 durations.

wuchong:
Hi @zentol , do you mean java 8 `Duration.parse`? AFAIK, `Duration.parse` only accept ISO-8601 duration format, i.e. `PnDTnHnMn.nS`. For example: `Duration.parse("2s")` is invalid, and `Duration.parse("PT2S")` is valid, however, this is not what we want.

TsReaper:
After discussing with @wuchong and @zentol , we decide not to introduce a new `TimeInterval` class but to add a `TimeUtils` util class with a parsing method.

I've uploaded the commit. Please take a look @wuchong @zentol .

wuchong:
LGTM. +1

wuchong:
Merged.

PR #7550, opened 2019-01-22
[FLINK-10245] [Streaming Connector] Support UpsertHBaseSink and UpsertHBaseTableSink

Clarkkkkk:
## What is the purpose of the change
- Add datastream sink and upsert table sink for HBase

## Brief change log
- Add HBase connector, details are in the doc.

## Verifying this change
This change added tests and can be verified as follows:
- org.apache.flink.streaming.connectors.hbase.HBaseSinkITCase: starts a mini HBase cluster to verify all sink functions.

## Does this pull request potentially affect one of the following parts:

  - Dependencies (does it add or upgrade a dependency):  no
  - The public API, i.e., is any changed class annotated with `@Public(Evolving)`: no
  - The serializers: no
  - The runtime per-record code paths (performance sensitive): no
  - Anything that affects deployment or recovery: no
  - The S3 file system connector: no

## Documentation

  - Does this pull request introduce a new feature? yes
  - If yes, how is the feature documented? JavaDocs and https://docs.google.com/document/d/1of0cYd73CtKGPt-UL3WVFTTBsVEre-TNRzoAt5u2PdQ/edit#

PR #9075, opened 2019-07-11
 [FLINK-10245][hbase] Add an upsert table sink factory for HBase

wuchong:

This commit adds full support for HBase to be used with Table & SQL API as well as SQL Client.  

## Brief change log

This commits is based on pull request #7550 , It includes: 
 - HBase upsert table sink (for append-only and updating queries) 
 - HBase table factory 
 - HBase table descriptors & validators 
 - Unit tests

## Verifying this change

  - Added integration tests to test HBaseTableSink can write contents to HBase table correctly.
  - Added unit tests to verify created table sink via HBaseTableFactory is correct.

## Documentation

### DIFF
--- a/flink-connectors/flink-hbase/src/main/java/org/apache/flink/addons/hbase/HBaseOptions.java
+++ b/flink-connectors/flink-hbase/src/main/java/org/apache/flink/addons/hbase/HBaseOptions.java
@@ -1,0 +1,135 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.addons.hbase;
+
+import javax.annotation.Nullable;
+
+import java.util.Objects;
+import java.util.Optional;
+
+import static org.apache.flink.util.Preconditions.checkNotNull;
+
+/**
+ * Common Options for HBase.
+ */
+public class HBaseOptions {
+
+	private final String tableName;
+	private final String zkQuorum;
+	@Nullable private final String zkNodeParent;
+
+	private HBaseOptions(String tableName, String zkQuorum, @Nullable String zkNodeParent) {
+		this.tableName = tableName;
+		this.zkQuorum = zkQuorum;
+		this.zkNodeParent = zkNodeParent;
+	}
+
+	String getTableName() {
+		return tableName;
+	}
+
+	String getZkQuorum() {
+		return zkQuorum;
+	}
+
+	Optional<String> getZkNodeParent() {
+		return Optional.ofNullable(zkNodeParent);
+	}
+
+	@Override
+	public String toString() {
+		return "HBaseOptions{" +
+			"tableName='" + tableName + '\'' +
+			", zkQuorum='" + zkQuorum + '\'' +
+			", zkNodeParent='" + zkNodeParent + '\'' +
+			'}';
+	}
+
+	@Override
+	public boolean equals(Object o) {
+		if (this == o) {
+			return true;
+		}
+		if (o == null || getClass() != o.getClass()) {
+			return false;
+		}
+		HBaseOptions that = (HBaseOptions) o;
+		return Objects.equals(tableName, that.tableName) &&
+			Objects.equals(zkQuorum, that.zkQuorum) &&
+			Objects.equals(zkNodeParent, that.zkNodeParent);
+	}
+
+	@Override
+	public int hashCode() {
+		return Objects.hash(tableName, zkQuorum, zkNodeParent);
+	}
+
+	/**
+	 * Creates a builder of {@link HBaseOptions}.
+	 */
+	public static Builder builder() {
+		return new Builder();
+	}
+
+	/**
+	 * Builder for {@link HBaseOptions}.
+	 */
+	public static class Builder {
+
+		private String tableName;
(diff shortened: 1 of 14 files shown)


flink-connectors/flink-hbase/src/main/java/org/apache/flink/addons/hbase/HBaseUpsertSinkFunction.java
```diff
@@ -0,0 +1,257 @@
...
+		LOG.info("start open ...");
+		org.apache.hadoop.conf.Configuration config = prepareRuntimeConfiguration();
+		try {
+			this.helper = new HBaseReadWriteHelper(schema);
+			this.numPendingRequests = new AtomicLong(0);
+
+			if (null == connection) {
+				this.connection = ConnectionFactory.createConnection(config);
+			}
+			// co BufferedMutatorExample-04-Params Create a parameter instance, set the table name and custom listener reference.
```
lincoln-lil: refine comments

flink-connectors/flink-hbase/src/main/java/org/apache/flink/addons/hbase/HBaseUpsertSinkFunction.java
```diff
@@ -0,0 +1,257 @@
...
+
+		if (scheduledFuture != null) {
+			scheduledFuture.cancel(false);
+			if (executor != null) {
+				executor.shutdownNow();
+			}
+		}
+
+		// make sure any errors from callbacks are rethrown
+		checkErrorAndRethrow();
```
lincoln-lil: remove this invocation and do `flush()` before close maybe better?

chancelq:
LGTM

flink-connectors/flink-hbase/src/main/java/org/apache/flink/addons/hbase/HBaseUpsertTableSink.java
```diff
@@ -0,0 +1,130 @@
...
+		checkArgument(hbaseTableSchema.getRowKeyName().isPresent(), "HBaseUpsertTableSink requires rowkey is set.");
+		this.hbaseTableSchema = hbaseTableSchema;
+		this.tableSchema = hbaseTableSchema.convertsToTableSchema();
+		this.hbaseOptions = hbaseOptions;
+		this.writeOptions = writeOptions;
+	}
+
+	@Override
+	public void setKeyFields(String[] keys) {
+		// hbase always upsert on rowkey, ignore query keys.
```
JingsongLi: Check to row keys?

JingsongLi:
LGTM +1, just left minor comments...

flink-connectors/flink-hbase/src/main/java/org/apache/flink/addons/hbase/HBaseUpsertTableSink.java
wuchong: I'm not sure. Because the support for key derivation is not very well, esp. fields concating which is used heavily in hbase sink.

flink-connectors/flink-hbase/src/main/java/org/apache/flink/addons/hbase/HBaseWriteOptions.java
```diff
@@ -0,0 +1,149 @@
...
+
+import static org.apache.flink.util.Preconditions.checkArgument;
+
+/**
+ * Options for HBase writing.
+ */
+public class HBaseWriteOptions implements Serializable {
+
+	private static final long serialVersionUID = 1L;
+	private static final long DEFAULT_FLUSH_MAX_SIZE_BYTES = 2 * 1024 * 1024; // 2mb
```
carp84: Suggest to use the `ConnectionConfiguration.WRITE_BUFFER_SIZE_DEFAULT` of hbase as the default (which is also 2M). Or is there any special reason for us to maintain a Flink-specified default write buffer size?

wuchong: Thanks for the tip.

flink-connectors/flink-hbase/src/main/java/org/apache/flink/addons/hbase/HBaseUpsertTableSink.java
JingsongLi: Can you add some comments in HBaseUpsertTableSink to explain it? It really confused...

flink-connectors/flink-hbase/src/main/java/org/apache/flink/addons/hbase/HBaseTableSchema.java
```diff
@@ -54,7 +54,7 @@
 	 * @param qualifier the qualifier name
 	 * @param clazz     the data type of the qualifier
 	 */
-	void addColumn(String family, String qualifier, Class<?> clazz) {
+	public void addColumn(String family, String qualifier, Class<?> clazz) {
```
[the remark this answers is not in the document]
wuchong: The reason is we want to expose `HBaseTableSchema` to users. Because `HBaseUpsertSinkFunction` accepts `HBaseTableSchema` as a parameter. If we want users to construct a `HBaseUpsertSinkFunction` and used in DataStream,  core methods of `HBaseTableSchema` should also be exposed. 
What do you think?

flink-connectors/flink-hbase/src/main/java/org/apache/flink/addons/hbase/HBaseOptions.java
```diff
@@ -0,0 +1,135 @@
...
+	private final String zkQuorum;
+	@Nullable private final String zkNodeParent;
+
+	private HBaseOptions(String tableName, String zkQuorum, String zkNodeParent) {
+		this.tableName = tableName;
+		this.zkQuorum = zkQuorum;
+		this.zkNodeParent = zkNodeParent;
+	}
+
+	String getTableName() {
```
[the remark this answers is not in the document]
wuchong: They are only used internally, so I make them package-private.

wuchong:
@JingsongLi  @carp84  comments are addressed. Sorry for commits squashed.

flink-connectors/flink-hbase/src/main/java/org/apache/flink/addons/hbase/HBaseTableSchema.java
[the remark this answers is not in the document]
carp84: I see, makes sense.

wuchong:
Thanks all. Wait until travis pass. https://travis-ci.org/wuchong/flink/builds/557314022

PR #9081, opened 2019-07-11
[FLINK-13209] Following FLINK-12951: Remove TableEnvironment#sql and …

danny0405:

This patch move create table ddl support from `#sql` to `#sqlUpdate`, also remove the deprecation of `sqlUpdate` and `sqlQuery`

### DIFF
--- a/flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/TableEnvironment.java
+++ b/flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/TableEnvironment.java
@@ -30,8 +30,6 @@ import org.apache.flink.table.functions.ScalarFunction;
 import org.apache.flink.table.sinks.TableSink;
 import org.apache.flink.table.sources.TableSource;
 
-import javax.annotation.Nullable;
-
 import java.util.Optional;
 
 /**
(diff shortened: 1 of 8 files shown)


flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/TableEnvironment.java
```diff
@@ -430,11 +362,51 @@ static TableEnvironment create(EnvironmentSettings settings) {
...
+	 * For example, the below DDL statement would create a CSV table named `tbl1`
+	 * into the current catalog:
+	 * <blockquote><pre>
+	 *    create table tbl1(
+	 *      a int,
+	 *      b bigint,
+	 *      c varchar
+	 *    ) with (
+	 *      connector = 'csv',
+	 *      csv.path = 'xxx'
```
wuchong: Give a validate with property? 
```
connector.type = 'filesystem',
connector.property-version = '1',
connector.path = '/path/to/file',
format.type = 'csv',
format.property-version = '1',
format.derive-schema = 'true'
```
However, I'm not sure the above properties work.

KurtYoung:
cc @twalthr

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/TableEnvironment.java
twalthr: I agree with Jark. If we show examples, then examples that also work in the current version.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/TableEnvironment.java
```diff
@@ -430,11 +362,51 @@ static TableEnvironment create(EnvironmentSettings settings) {
 	 * }
 	 * </pre>
 	 *
-	 * @deprecated Use {@link #sql(String)}.
+	 * <p>A DDL statement can also execute to create a table:
```
twalthr: Rephrase: "DDL statements can be executed to create tables:"

KurtYoung:
+1 from my side, waiting for travis

PR #9082, opened 2019-07-11
[FLINK-13207][ml] Add the algorithm of Fast Fourier Transformation(FFT)

xuyang1706:

Add 2 common used algorithms of Fast Fourier Transformation(FFT)

1. Cooley-Tukey algorithm, high performance, but only supports length of power-of-2.
2. Chirp-Z algorithm, can perform FFT with any length.

## Brief change log

  - *add Cooley-Tukey algorithm*
  - *add Chirp-Z algorithm*
  - *add unit tests*

## Verifying this change

- run test case pass

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/FFT.java
+++ b/flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/FFT.java
@@ -1,0 +1,199 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.flink.ml.common.utils;
+
+import org.apache.commons.math3.complex.Complex;
+
+/**
+ * Fast Fourier Transformation(FFT).
+ * Provides 2 algorithms:
+ * 1. Cooley-Tukey algorithm, high performance, but only supports length of power-of-2.
+ * 2. Chirp-Z algorithm, can perform FFT with any length.
+ */
+public class FFT {
+	/**
+	 * Helper for root of unity. Returns the group for power "length".
+	 */
+	public static Complex[] getOmega(int length) throws Exception {
+		Complex[] omega = new Complex[length];
+		Complex unit = new Complex(Math.cos(2 * Math.PI / length),
+			Math.sin(2 * Math.PI / length));
+		omega[0] = new Complex(1, 0);
+		omega[1] = unit;
+		for (int index = 2; index < length; index++) {
+			omega[index] = omega[index - 1].multiply(unit);
+		}
+		return omega;
+	}
+
+	/**
+	 * Cooley-Tukey algorithm, can perform fft for any composite base.
+	 * Specifically, it can perform power-of-2 base fft(with some modification to make it an in-place algorithm).
+	 * See:
+	 * "An algorithm for the machine calculation of complex Fourier series", JW Cooley, JW Tukey, 1965
+	 * for a rough reference.
+	 * Detail of radix-2 in-place Cooley-Tukey algorithm can be found in many places, e.g. CLRS textbook.
+	 */
+	public static Complex[] fftRadix2CooleyTukey(Complex[] input, Boolean inverse, Complex[] omega) throws Exception {
+
+		//1. length
+		int length = input.length;
+		int logl = (int) (Math.log(length + 0.01) / Math.log(2));
+
+		//notice: only support power of 2
+		//fftChirpZ support other lengths
+		if ((1 << logl) != length) {
+			throw new RuntimeException("Radix-2 Cooley-Tukey only supports lengths of power-of-2.");
+		}
+
+		//2. copy data
+		Complex[] inputCopy = new Complex[length];
+		for (int index = 0; index < length; index++) {
+			inputCopy[index] = new Complex(input[index].getReal(), input[index].getImaginary());
+		}
+
+		//3. bit reverse
+		int[] reverse = new int[length];
+		for (int index = 0; index < length; index++) {
+			int t = 0;
+			for (int pos = 0; pos < logl; pos++) {
+				if ((index & (1 << pos)) != 0) {
+					t |= (1 << (logl - pos - 1));
+				}
+			}
+			reverse[index] = t;
+		}
+
+		//4. reverse the input
+		for (int index = 0; index < length; index++) {
+			if (index < reverse[index]) {
+				Complex t = inputCopy[index];
+				inputCopy[index] = inputCopy[reverse[index]];
+				inputCopy[reverse[index]] = t;
+			}
+		}
+
+		//5. perform in-place fft
+		if (inverse) {
+			//inverse fft
+			for (int len = 2; len <= length; len *= 2) {
+				int mid = len / 2;
+				for (int step = 0; step < length; step += len) {
+					for (int index = 0; index < mid; index++) {
+						Complex t = omega[length / len * index]
+							.multiply(inputCopy[step + mid + index]);
+						inputCopy[step + mid + index] = inputCopy[step + index].subtract(t);
+						inputCopy[step + index] = inputCopy[step + index].add(t);
+					}
+				}
+			}
+
+			for (int index = 0; index < length; index++) {
+				inputCopy[index] = inputCopy[index].divide(length);
+			}
+		} else {
+			//forward fft
+			for (int len = 2; len <= length; len *= 2) {
+				int mid = len / 2;
+				for (int step = 0; step < length; step += len) {
+					for (int index = 0; index < mid; index++) {
+						Complex t = omega[length / len * index].conjugate()
+							.multiply(inputCopy[step + mid + index]);
+						inputCopy[step + mid + index] = inputCopy[step + index].subtract(t);
+						inputCopy[step + index] = inputCopy[step + index].add(t);
+					}
+				}
+			}
+		}
+
+		return inputCopy;
+
+	}
+
+	/**
+	 * Chirp-Z algorithm, also called Bluestein algorithm,
+	 * can perform fft with any base.
+	 * It use convolution and "chirp-z", and the convolution can be performed by a power-of-2 base fft.
+	 * See:
+	 * "The chirp z-transform algorithm", L Rabiner, RW Schafer, C Rader, 1969
+	 * for details.
+	 **/
+	public static Complex[] fftChirpZ(Complex[] input, Boolean inverse, Complex[] omega, Complex[] omega2)
+		throws Exception {
+
+		//1. length
+		int length = input.length;
+		int logl = (int) (Math.log(length + 0.01) / Math.log(2));
+		if ((1 << logl) == length) {
+			throw new RuntimeException(
+				"Chirp-Z is not efficient for lengths of power-of-2. Use Radix-2 Cooley-Tukey instead.");
+		}
+
+		int nextLength = 1 << (logl + 2);
+
+		//2. copy data & construct chirps
+		Complex[] inputSet = new Complex[nextLength];
+		Complex[] convChirpSet = new Complex[nextLength];
+		if (inverse) {
+			for (int index = 0; index < length; index++) {
+				Complex curChirp = omega2[(index * index) % (2 * length)];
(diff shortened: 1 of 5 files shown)


flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/FFT.java
```diff
@@ -0,0 +1,199 @@
...
+ * Fast Fourier Transformation(FFT).
+ * Provides 2 algorithms:
+ * 1. Cooley-Tukey algorithm, high performance, but only supports length of power-of-2.
+ * 2. Chirp-Z algorithm, can perform FFT with any length.
+ */
+public class FFT {
+	/**
+	 * Helper for root of unity. Returns the group for power "length".
+	 */
+	public static Complex[] getOmega(int length) throws Exception {
```
c4emmmm: I suggest to avoid throwing Exception but use specific exception subclass instead, and add throws in javaDoc to explain what exception may be thrown, or remove it if there's no exception requiring predefined handling.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/FFT.java
```diff
@@ -0,0 +1,199 @@
...
+
+	/**
+	 * Cooley-Tukey algorithm, can perform fft for any composite base.
+	 * Specifically, it can perform power-of-2 base fft(with some modification to make it an in-place algorithm).
+	 * See:
+	 * "An algorithm for the machine calculation of complex Fourier series", JW Cooley, JW Tukey, 1965
+	 * for a rough reference.
+	 * Detail of radix-2 in-place Cooley-Tukey algorithm can be found in many places, e.g. CLRS textbook.
+	 */
+	public static Complex[] fftRadix2CooleyTukey(Complex[] input, Boolean inverse, Complex[] omega) throws Exception {
```
c4emmmm: It seems that you never use "inverse" as "Boolean" which can be null, but only treat it as a simple boolean. It's better to avoid using boxed type unless it's strictly required.

By the way, in flink code style, all input params are "#NotNull" by default. It's recommended to add a "#Nullable" annotation for those who can be null. Please replace # with @, I use # because @ in github has specific usage.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/FFT.java
```diff
@@ -0,0 +1,199 @@
...
+
+		//1. length
+		int length = input.length;
+		int logl = (int) (Math.log(length + 0.01) / Math.log(2));
```
c4emmmm: I'm not really sure if it influences the performance, but maybe we can define 1/Math.log(2) as a constant to avoid duplicate computing?

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/FFT.java
```diff
@@ -0,0 +1,199 @@
...
+/**
+ * Fast Fourier Transformation(FFT).
+ * Provides 2 algorithms:
+ * 1. Cooley-Tukey algorithm, high performance, but only supports length of power-of-2.
+ * 2. Chirp-Z algorithm, can perform FFT with any length.
+ */
+public class FFT {
+	/**
+	 * Helper for root of unity. Returns the group for power "length".
+	 */
```
c4emmmm: I think we should also have params, returns and throws tag to explain the function better in javaDoc, not only this one but also the javaDoc for other functions.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/FFT.java
xuyang1706: Thanks for your advice.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/FFT.java
xuyang1706: Thanks, "Boolean" is changed to "boolean"

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/FFT.java
xuyang1706: Yes，it is better. 	private static final double INVERSE_LOG_2 =  1.0/ Math.log(2);

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/FFT.java
xuyang1706: This is internal functions for the DCT algorithm. DCT will be summited later.

xuyang1706:
> Hi @xuyang1706, thanks for your PR.
> Could you explain the goal of the PR, please? I do not see any logic related to Flink API, also I do not see how your implementations of FFT will be work in Flink.
> I suppose FFT should be related to Flink ML API directly, for example, FFT can implement https://github.com/apache/flink/blob/master/flink-ml-parent/flink-ml-api/src/main/java/org/apache/flink/ml/api/core/Transformer.java

This is internal functions for the DCT algorithm. DCT need the implementation of Transformer,  will be summited later.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/FFT.java
```diff
@@ -0,0 +1,201 @@
...
+ */
+public class FFT {
+
+	private static final double INVERSE_LOG_2 = 1.0 / Math.log(2);
+
+	/**
+	 * Helper for root of unity. Returns the group for power "length".
+	 */
+	public static Complex[] getOmega(int length) {
+		Complex[] omega = new Complex[length];
```
walterddr: Can we use [breeze Complex](https://github.com/scalanlp/breeze/blob/master/math/src/main/scala/breeze/math/Complex.scala) ?

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/FFT.java
```diff
@@ -0,0 +1,201 @@
...
+		for (int index = 2; index < length; index++) {
+			omega[index] = omega[index - 1].multiply(unit);
+		}
+		return omega;
+	}
+
+	/**
+	 * Cooley-Tukey algorithm, can perform fft for any composite base.
+	 * Specifically, it can perform power-of-2 base fft(with some modification to make it an in-place algorithm).
+	 * See:
```
walterddr:
```
/**
 * Cooley-Tukey ...
 * 
 * <p>See reference for more details ...
 * <p><ul>
 *   * <li>"An algorithm for..."
 *   * <li>"CLRS text book"
 *   * </ul>
 * </p>
 */
```
reference should have itemized. and should separate from previous doc.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/FFT.java
```diff
@@ -0,0 +1,201 @@
...
+	public static Complex[] fftRadix2CooleyTukey(Complex[] input, boolean inverse, Complex[] omega) {
+
+		//1. length
+		int length = input.length;
+		int logl = (int) (Math.log(length + 0.01) * INVERSE_LOG_2);
+
+		//notice: only support power of 2
+		//fftChirpZ support other lengths
+		if ((1 << logl) != length) {
+			throw new RuntimeException("Radix-2 Cooley-Tukey only supports lengths of power-of-2.");
```
walterddr: throw `IllegalArgumentException` or `UnsupportedOperatorException` makes more sense to me.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/FFT.java
```diff
@@ -0,0 +1,201 @@
...
+		}
+
+		//2. copy data
+		Complex[] inputCopy = new Complex[length];
```
walterddr: nit: Maybe renamed it to `inPlaceFFT`. inputCopy doesn't sound immediately clear what it does especially with such a long method.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/FFT.java
```diff
@@ -0,0 +1,201 @@
...
+
+		return inputCopy;
+
+	}
+
+	/**
+	 * Chirp-Z algorithm, also called Bluestein algorithm,
+	 * can perform fft with any base.
+	 * It use convolution and "chirp-z", and the convolution can be performed by a power-of-2 base fft.
+	 * See:
```
walterddr: same comment for javadoc

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/FFT.java
qiuxiafei: I remember there used to be a discussion about breeze months ago. The conclusion at that time was NOT to introducing it into flink ml, because it's written in scala and most of the time we need to write some kinda wrapper classes to use them. Correct me if anything wrong.

walterddr: oops... I was working on something else and it must've been the context switching.. I will resolve this.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/FFT.java
xuyang1706: Thanks, changed

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/FFT.java
xuyang1706: Thanks, changed

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/FFT.java
xuyang1706: Thanks, changed

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/FFT.java
xuyang1706: Thanks, changed

xuyang1706:
> Thanks for the contribution @xuyang1706 . I left a few comments with more higher level questions.
> Regarding the placement. I think this should be considered as a transform since FFT is a transform itself. even if `DCT` is another transform that utilizes this algorithm. what do you think?

Thanks for your advices @walterddr . I have refined the code. Yes, there is DCT-Transformer using this FFT class. For the FFT transform real number to complex number, but the common ML algorithms are dealing with real number only, we have no plan to submit FFT-Transformer recently.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/FFT.java
```diff
@@ -0,0 +1,210 @@
...
+
+import org.apache.commons.math3.complex.Complex;
+
+/**
+ * Fast Fourier Transformation(FFT).
+ * Provides 2 algorithms:
+ * 1. Cooley-Tukey algorithm, high performance, but only supports length of power-of-2.
+ * 2. Chirp-Z algorithm, can perform FFT with any length.
+ */
+public class FFT {
```
walterddr: I am fine with the `FFT` class currently living in the `common/utils` package. however, we probably need to consider moving these to more suitable location. since these are heavy-weighted classes with dedicated purpose. 
I think maybe the reason that raised the question @ex00 have - since it does not immediately clear why a "transform" class sits in a common utility package. 

Some thoughts I have based on some digging: splitting based on functionality group might be better? E.g.
1. in feature extraction: embedding, hashing, PCA, etc: Both Spark & Sckit-Learn put them in dedicated package "feature extraction"
2. in algorithms/optimizers: E.g. GaussianMixture, DistanceMeasure: are put in "clustering" package;
#REF: [1](https://github.com/apache/spark/tree/master/mllib/src/main/scala/org/apache/spark/ml/feature), [2](https://github.com/scikit-learn/scikit-learn/tree/master/sklearn/feature_extraction) 

So IMO it makes sense to put them in a package name `feature.extraction` or `feature`. Any idea you have in mind @xuyang1706 @ex00 @qiuxiafei ?

xuyang1706: Thanks, the fft has been moved  to the `operator.common.dataproc`

xuyang1706:
> Thanks @xuyang1706 for the explanation. I would like to withdraw my previous comment: yeah you are right most of the ML algorithms are real-number-based and having a FFT transform interface doesn't make much sense.

Thanks for your kindly advice @walterddr . Yes, the package path `common.utils` is not suitable for FFT, I have changed the path.

PR #9085, opened 2019-07-11
[FLINK-13215][hive] Mirate ParseUtil usages to URLEncoder

zentol:
Fixes a Java 9 compilation problem; the `ParseUtil` class is no longer accessible by default. Migrate to public APIs instead.

@bowenli86 Could you take a look at this one? What's a good way to test this change?


bowenli86:
@zentol Thanks for the fix! 

`FlinkStandaloneHiveRunner` runs on test classes annotated with `@RunWith(FlinkStandaloneHiveRunner.class)` and we use it in `HiveTableSinkTest` now. The change should be good as long as normal CI passes

zentol:
Rebased the PR as a way more hive tests were failing than I would consider possible, considering this change.
I'm a bit sceptical as to whether my change is correct though; the output is not the same as before and I haven't found a 1:1 replacement yet.

zentol:
Have to find another replacement.

PR #9090, opened 2019-07-11
[FLINK-13124] Don't forward exceptions when finishing SourceStreamTask

aljoscha:

Before, exceptions that occurred after cancelling a source (as the
KafkaConsumer did, for example) would make a job fail when attempting a
"stop-with-savepoint". Now we ignore those exceptions.

## Brief change log

- add a `isFinished` flag in `SourceStreamTask`
- check this flag before re-throwing exceptions

## Verifying this change

- Added unit tests in `SourceStreamTaskTest`

  - Checkpointing, yes, because it fixes stop-with-savepoint

## Documentation

### DIFF
--- a/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SourceStreamTask.java
+++ b/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SourceStreamTask.java
@@ -127,6 +127,10 @@ public class SourceStreamTask<OUT, SRC extends SourceFunction<OUT>, OP extends S
 
 	@Override
 	protected void finishTask() throws Exception {
+		// We tell the mailbox to finish, to prevent any exceptions that might occur during
+		// finishing from leading to a FAILED state. This could happen, for example, when cancelling
+		// sources as part of a "stop-with-savepoint".
+		mailboxProcessor.allActionsCompleted();
 		cancelTask();
 	}
 
--- a/flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/SourceStreamTaskTest.java
+++ b/flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/SourceStreamTaskTest.java
@@ -231,6 +231,27 @@ public class SourceStreamTaskTest {
 			testHarness.getOutput());
 	}
 
+	/**
+	 * If finishing a task doesn't swallow exceptions this test would fail with an exception.
+	 */
+	@Test
+	public void finishingIgnoresExceptions() throws Exception {
+		final StreamTaskTestHarness<String> testHarness = new StreamTaskTestHarness<>(
+				SourceStreamTask::new,
+				BasicTypeInfo.STRING_TYPE_INFO);
+
+		testHarness.setupOutputForSingletonOperatorChain();
+		StreamConfig streamConfig = testHarness.getStreamConfig();
+		streamConfig.setStreamOperator(new StreamSource<>(new ExceptionThrowingSource()));
+		streamConfig.setOperatorID(new OperatorID());
+
+		testHarness.invoke();
+		ExceptionThrowingSource.isInRunLoop.get();
+		testHarness.getTask().finishTask();
+
+		testHarness.waitForTaskCompletion();
+	}
+
 	private static class MockSource implements SourceFunction<Tuple2<Long, Integer>>, ListCheckpointed<Serializable> {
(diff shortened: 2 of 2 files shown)


aljoscha:
cc @knaufk

knaufk:
`stop` now works with the fix in my test setup. Thanks @aljoscha.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SourceStreamTask.java
```diff
@@ -49,6 +49,12 @@
 
 	private volatile boolean externallyInducedCheckpoints;
 
+	/**
+	 * Indicates whether this Task was purposefully finished (by finishTask()), in this case we
+	 * want to ignore exceptions thrown after finishing, to ensure shutdown works smoothly.
+	 */
+	private volatile boolean isFinished = false;
+
```
kl0u: I would suggest to move this flag to the `StreamTask`. The reason is that although for now it is used only in the `SourceStreamTask`, know if the job was normally terminated or cancelled is a property relevant to all the `StreamTask`s.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SourceStreamTask.java
```diff
@@ -147,6 +155,7 @@ protected void cancelTask() {
 
 	@Override
```
kl0u: With the above comments in mind, I would suggest to have a method in the `StreamTask` like the following:

```
@VisibleForTesting
 void finish() throws Exception {
		finished = true;
		finishTask(); // this is implemented by subclasses
	}
```
(needed in the `finishingIgnoresExceptions()` test).

And with this, the end of the `performCheckpoint` should become:

```
if (isRunning && syncSavepointLatch.isSet()) {

			final boolean checkpointWasAcked =
					syncSavepointLatch.blockUntilCheckpointIsAcknowledged();

			if (checkpointWasAcked) {
				finishTask();
			}
		}
```

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SourceStreamTask.java
```diff
@@ -118,7 +124,9 @@ protected void performDefaultAction(ActionContext context) throws Exception {
 		}
 
 		sourceThread.join();
-		sourceThread.checkThrowSourceExecutionException();
+		if (!isFinished) {
```
kl0u: Given the above comment, this can be a `protected` getter method for the `finished` flag in the `StreamTask`, available to all `StreamTask`s.

aljoscha:
Master has moved since I opened this PR, the fix still works, with a change, but I want to investigate a bit further because that change on master removes other new behaviour that my PR was testing in one of the test cases.

aljoscha:
PTAL (please take another look). I now changed this to work with latest master. (there was a change in how the mailbox worked.)

aljoscha:
ok, thanks! could you please merge? (just arrived in china 😅 )

kl0u:
Sure! I will merge now.

kl0u:
Merged on master

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SourceStreamTask.java
```diff
@@ -127,6 +127,10 @@ protected void cancelTask() {
 
 	@Override
 	protected void finishTask() throws Exception {
+		// We tell the mailbox to finish, to prevent any exceptions that might occur during
+		// finishing from leading to a FAILED state. This could happen, for example, when cancelling
+		// sources as part of a "stop-with-savepoint".
+		mailboxProcessor.allActionsCompleted();
```
1u0: Sorry for being late in this PR.
Is there a reason why in `master` branch version there is `mailboxProcessor.allActionsCompleted();` in `finishTask()` but in `release-1.9` branch doesn't have it (#9115)?

aljoscha: On the `release-1.9` branch `StreamTask.cancel()` also doesn't use this method, in fact the `MailboxProcessor` doesn't yet exist and isn't used like this. However, this solution leads to a race condition, see also https://github.com/apache/flink/pull/9125.

PR #9103, opened 2019-07-12
[FLINK-13243][tests] Simplify exception matching

zentol:
Reworks how the exception message matching is being done in the `AbstractOperatorRestoreTestBase` to not be subjected to OS-specific line-endings.


flink-tests/src/test/java/org/apache/flink/test/state/operator/restore/AbstractOperatorRestoreTestBase.java
```diff
@@ -65,6 +66,13 @@
 	private static final int NUM_TMS = 1;
 	private static final int NUM_SLOTS_PER_TM = 4;
 	private static final Duration TEST_TIMEOUT = Duration.ofSeconds(10000L);
+	private static final Pattern PATTERN_CANCEL_WITH_SAVEPOINT_TOLERATED_EXCEPTIONS = Pattern
+		.compile(
+			"(savepoint for the job .* failed)" +
```
1u0: Looks like this exception string snippet is not needed anymore.

PR #9105, opened 2019-07-13
[FLINK-13241][Yarn/Mesos] Fix Yarn/MesosResourceManager setting managed memory size into wrong configuration instance.

xintongsong:

This PR fix the bug that Yarn/MesosResourceManager set managed memory size into wrong configuration instance. As a result, such setting does not overwrite the config on TM side, causing RM/TM slot resource profile inconsistency. As a result, pending task manager slots on RM side cannot be completed once created, prevent RM from starting new TMs for subsequence slot requests.

## Brief change log

- Yarn/MesosResourceManager set managed memory into the proper configuration instance.

## Verifying this change

  - Manually verified the change by running a streaming job with two vertices using lazy from source strategy on Yarn. The job requests slots in 2 rounds and we verified that RM does start new TM for the second round slot requests, which it doesn't without this change.

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-mesos/src/main/java/org/apache/flink/mesos/runtime/clusterframework/MesosResourceManager.java
+++ b/flink-mesos/src/main/java/org/apache/flink/mesos/runtime/clusterframework/MesosResourceManager.java
@@ -42,7 +42,6 @@ import org.apache.flink.mesos.util.MesosArtifactServer;
 import org.apache.flink.mesos.util.MesosConfiguration;
 import org.apache.flink.runtime.clusterframework.ApplicationStatus;
 import org.apache.flink.runtime.clusterframework.ContainerSpecification;
-import org.apache.flink.runtime.clusterframework.ContaineredTaskManagerParameters;
 import org.apache.flink.runtime.clusterframework.types.ResourceID;
 import org.apache.flink.runtime.clusterframework.types.ResourceProfile;
 import org.apache.flink.runtime.concurrent.FutureUtils;
@@ -184,8 +183,7 @@ public class MesosResourceManager extends ResourceManager<RegisteredMesosWorkerN
 		this.mesosServices = Preconditions.checkNotNull(mesosServices);
 		this.actorSystem = Preconditions.checkNotNull(mesosServices.getLocalActorSystem());
 
-		// copy the config, because we might change it for the TaskManagers
-		this.flinkConfig = new Configuration(Preconditions.checkNotNull(flinkConfig));
+		this.flinkConfig = Preconditions.checkNotNull(flinkConfig);
 		this.mesosConfig = Preconditions.checkNotNull(mesosConfig);
 
 		this.artifactServer = Preconditions.checkNotNull(mesosServices.getArtifactServer());
@@ -198,9 +196,7 @@ public class MesosResourceManager extends ResourceManager<RegisteredMesosWorkerN
 		this.workersInLaunch = new HashMap<>(8);
 		this.workersBeingReturned = new HashMap<>(8);
 
-		final ContaineredTaskManagerParameters containeredTaskManagerParameters = taskManagerParameters.containeredParameters();
-		this.slotsPerWorker = updateTaskManagerConfigAndCreateWorkerSlotProfiles(
-			this.flinkConfig, containeredTaskManagerParameters.taskManagerTotalMemoryMB(), containeredTaskManagerParameters.numSlots());
+		this.slotsPerWorker = createWorkerSlotProfiles(flinkConfig);
 		setFailUnfulfillableRequest(true);
 	}
 
--- a/flink-mesos/src/main/java/org/apache/flink/mesos/runtime/clusterframework/MesosResourceManager.java
+++ b/flink-mesos/src/main/java/org/apache/flink/mesos/runtime/clusterframework/MesosResourceManager.java
@@ -200,7 +200,7 @@ public class MesosResourceManager extends ResourceManager<RegisteredMesosWorkerN
 
 		final ContaineredTaskManagerParameters containeredTaskManagerParameters = taskManagerParameters.containeredParameters();
 		this.slotsPerWorker = updateTaskManagerConfigAndCreateWorkerSlotProfiles(
-			flinkConfig, containeredTaskManagerParameters.taskManagerTotalMemoryMB(), containeredTaskManagerParameters.numSlots());
+			this.flinkConfig, containeredTaskManagerParameters.taskManagerTotalMemoryMB(), containeredTaskManagerParameters.numSlots());
 		setFailUnfulfillableRequest(true);
 	}
 
--- a/flink-mesos/src/main/java/org/apache/flink/mesos/runtime/clusterframework/MesosResourceManagerFactory.java
+++ b/flink-mesos/src/main/java/org/apache/flink/mesos/runtime/clusterframework/MesosResourceManagerFactory.java
@@ -19,6 +19,7 @@
 package org.apache.flink.mesos.runtime.clusterframework;
 
 import org.apache.flink.configuration.Configuration;
+import org.apache.flink.configuration.TaskManagerOptions;
 import org.apache.flink.mesos.runtime.clusterframework.services.MesosServices;
 import org.apache.flink.mesos.util.MesosConfiguration;
 import org.apache.flink.runtime.clusterframework.ContainerSpecification;
@@ -98,5 +99,12 @@ public class MesosResourceManagerFactory implements ResourceManagerFactory<Regis
 			taskManagerContainerSpec,
 			webInterfaceUrl,
 			jobManagerMetricGroup);
+	}
+
+	@Override
+	public Configuration getResourceManagerConfiguration(Configuration flinkConfig) {
+		final Configuration config = new Configuration(flinkConfig);
+		config.setString(TaskManagerOptions.MANAGED_MEMORY_SIZE, flinkConfig.getInteger(MesosTaskManagerParameters.MESOS_RM_TASKS_MEMORY_MB) + "m");
+		return ResourceManagerFactory.super.getResourceManagerConfiguration(config);
 	}
 }
--- a/flink-mesos/src/main/java/org/apache/flink/mesos/runtime/clusterframework/MesosResourceManagerFactory.java
+++ b/flink-mesos/src/main/java/org/apache/flink/mesos/runtime/clusterframework/MesosResourceManagerFactory.java
@@ -19,7 +19,6 @@
 package org.apache.flink.mesos.runtime.clusterframework;
 
 import org.apache.flink.configuration.Configuration;
-import org.apache.flink.configuration.TaskManagerOptions;
 import org.apache.flink.mesos.runtime.clusterframework.services.MesosServices;
 import org.apache.flink.mesos.util.MesosConfiguration;
 import org.apache.flink.runtime.clusterframework.ContainerSpecification;
@@ -99,12 +98,5 @@ public class MesosResourceManagerFactory implements ResourceManagerFactory<Regis
 			taskManagerContainerSpec,
 			webInterfaceUrl,
 			jobManagerMetricGroup);
-	}
-
-	@Override
-	public Configuration getResourceManagerConfiguration(Configuration flinkConfig) {
-		final Configuration config = new Configuration(flinkConfig);
-		config.setString(TaskManagerOptions.MANAGED_MEMORY_SIZE, flinkConfig.getInteger(MesosTaskManagerParameters.MESOS_RM_TASKS_MEMORY_MB) + "m");
-		return ResourceManagerFactory.super.getResourceManagerConfiguration(config);
 	}
 }
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/resourcemanager/ResourceManager.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/resourcemanager/ResourceManager.java
@@ -27,7 +27,6 @@ import org.apache.flink.configuration.MemorySize;
 import org.apache.flink.configuration.TaskManagerOptions;
 import org.apache.flink.runtime.blob.TransientBlobKey;
 import org.apache.flink.runtime.clusterframework.ApplicationStatus;
-import org.apache.flink.runtime.clusterframework.ContaineredTaskManagerParameters;
 import org.apache.flink.runtime.clusterframework.types.AllocationID;
 import org.apache.flink.runtime.clusterframework.types.ResourceID;
 import org.apache.flink.runtime.clusterframework.types.ResourceIDRetrievable;
@@ -1210,24 +1209,6 @@ public abstract class ResourceManager<WorkerType extends ResourceIDRetrievable>
 
 		final ResourceProfile resourceProfile = TaskManagerServices.computeSlotResourceProfile(numSlots, managedMemoryBytes);
 		return Collections.nCopies(numSlots, resourceProfile);
-	}
-
-	@VisibleForTesting
-	public static Collection<ResourceProfile> updateTaskManagerConfigAndCreateWorkerSlotProfiles(
-			Configuration config, long totalMemoryMB, int numSlots) {
-
-		final long cutoffMB = ContaineredTaskManagerParameters.calculateCutoffMB(config, totalMemoryMB);
-		final long processMemoryBytes = (totalMemoryMB - cutoffMB) << 20; // megabytes to bytes
-		final long managedMemoryBytes = TaskManagerServices.getManagedMemoryFromProcessMemory(config, processMemoryBytes);
-
-		updateFlinkConfForManagedMemory(config, managedMemoryBytes);
-
-		final ResourceProfile resourceProfile = TaskManagerServices.computeSlotResourceProfile(numSlots, managedMemoryBytes);
-		return Collections.nCopies(numSlots, resourceProfile);
-	}
-
-	static void updateFlinkConfForManagedMemory(Configuration conf, long managedMemorySize) {
-		conf.setString(TaskManagerOptions.MANAGED_MEMORY_SIZE, managedMemorySize + "b");
 	}
 }
 
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/resourcemanager/ResourceManager.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/resourcemanager/ResourceManager.java
@@ -23,6 +23,7 @@ import org.apache.flink.api.common.JobID;
 import org.apache.flink.api.common.time.Time;
 import org.apache.flink.api.java.tuple.Tuple2;
 import org.apache.flink.configuration.Configuration;
+import org.apache.flink.configuration.MemorySize;
 import org.apache.flink.configuration.TaskManagerOptions;
 import org.apache.flink.runtime.blob.TransientBlobKey;
 import org.apache.flink.runtime.clusterframework.ApplicationStatus;
@@ -1202,6 +1203,14 @@ public abstract class ResourceManager<WorkerType extends ResourceIDRetrievable>
 	// ------------------------------------------------------------------------
 	//  Helper methods
 	// ------------------------------------------------------------------------
+
+	public static Collection<ResourceProfile> createWorkerSlotProfiles(Configuration config) {
+		final int numSlots = config.getInteger(TaskManagerOptions.NUM_TASK_SLOTS);
+		final long managedMemoryBytes = MemorySize.parse(config.getString(TaskManagerOptions.MANAGED_MEMORY_SIZE)).getBytes();
+
+		final ResourceProfile resourceProfile = TaskManagerServices.computeSlotResourceProfile(numSlots, managedMemoryBytes);
+		return Collections.nCopies(numSlots, resourceProfile);
+	}
 
 	@VisibleForTesting
 	public static Collection<ResourceProfile> updateTaskManagerConfigAndCreateWorkerSlotProfiles(
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/resourcemanager/ResourceManagerFactory.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/resourcemanager/ResourceManagerFactory.java
@@ -19,6 +19,9 @@
 package org.apache.flink.runtime.resourcemanager;
 
 import org.apache.flink.configuration.Configuration;
+import org.apache.flink.configuration.ConfigurationUtils;
+import org.apache.flink.configuration.TaskManagerOptions;
+import org.apache.flink.runtime.clusterframework.ContaineredTaskManagerParameters;
 import org.apache.flink.runtime.clusterframework.types.ResourceID;
 import org.apache.flink.runtime.clusterframework.types.ResourceIDRetrievable;
 import org.apache.flink.runtime.entrypoint.ClusterInformation;
@@ -28,6 +31,7 @@ import org.apache.flink.runtime.metrics.MetricRegistry;
 import org.apache.flink.runtime.metrics.groups.JobManagerMetricGroup;
 import org.apache.flink.runtime.rpc.FatalErrorHandler;
 import org.apache.flink.runtime.rpc.RpcService;
+import org.apache.flink.runtime.taskexecutor.TaskManagerServices;
 
 import javax.annotation.Nullable;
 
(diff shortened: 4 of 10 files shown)


xintongsong:
Hi @tillrohrmann, this is a bug fix for Flink-13241. Please take a look at your convenience.

I think this fix needs to go into both master and release-1.9. Typically what should I do in such cases? Should I open 2 separate PRs based on master/release-1.9 respectively, or just one PR and mention that in description/comment?

xintongsong:
Thanks for the comment, @azagrebin.
However, I'm not sure whether we need another test case to cover this issue, since we already have `YarnResourceManagerTest#testCreateSlotsPerWorker`. This test case was using the wrong instance for TM side calculation, which is also fixed in this PR. It should fail now if RM did not set the configuration instance for starting TM correctly.
Maybe I'm overlooking something that may trigger this bug again without failing the existing test cases?

azagrebin:
@xintongsong true, technically the currently adjusted test would have been failed. Only `MesosResourceManager` change seems still to be uncovered.

My question was more about the test coverage of the original effort that did not catch this bug. Basically do we have IT tests to check that the TM containers are created with the correct profile and/or number of slots?

xintongsong:
@azagrebin 
Ok, I see what you mean. I think you are right. We do have a `YarnConfigurationITCase` that checks TMs are started with correct number of slots and memory sizes. Unfortunately, this test case does not cover the managed memory size. Just pushed another commit adding managed memory checking to this test case.

flink-mesos/src/main/java/org/apache/flink/mesos/runtime/clusterframework/MesosResourceManager.java
```diff
@@ -200,7 +200,7 @@ public MesosResourceManager(
 
 		final ContaineredTaskManagerParameters containeredTaskManagerParameters = taskManagerParameters.containeredParameters();
 		this.slotsPerWorker = updateTaskManagerConfigAndCreateWorkerSlotProfiles(
-			flinkConfig, containeredTaskManagerParameters.taskManagerTotalMemoryMB(), containeredTaskManagerParameters.numSlots());
+			this.flinkConfig, containeredTaskManagerParameters.taskManagerTotalMemoryMB(), containeredTaskManagerParameters.numSlots());
```
azagrebin: Which tests cover this change?
How much effort would it be to add tests to e.g. `MesosResourceManagerTest`?

flink-yarn-tests/src/test/java/org/apache/flink/yarn/YarnConfigurationITCase.java
```diff
@@ -187,6 +191,12 @@ public void testFlinkContainerMemory() throws Exception {
 					assertThat(
 						(double) taskManagerInfo.getHardwareDescription().getSizeOfJvmHeap() / (double) expectedHeadSize,
 						is(closeTo(1.0, 0.15)));
+
+					final int defaultTaskManagerMemoryMB = ConfigurationUtils.getTaskManagerHeapMemory(configuration).getMebiBytes();
```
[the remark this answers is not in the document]
xintongsong: I kind of think it differently. 
I think we already have `testCreateSlotsPerWorker` that checks the behavior of `updateTaskManagerConfigAndCreateWorkerSlotProfiles`. If we just set constant `defaultTaskManagerMemoryMB` and expect a corresponding constant `expectedManagedMemoryMB`, then later if we change the logic in `updateTaskManagerConfigAndCreateWorkerSlotProfiles `  that it calculates managed memory from task manager memory differently, we may have to also change this test case.

flink-yarn-tests/src/test/java/org/apache/flink/yarn/YarnConfigurationITCase.java
```diff
@@ -187,6 +191,12 @@ public void testFlinkContainerMemory() throws Exception {
...
+					YarnResourceManager.updateTaskManagerConfigAndCreateWorkerSlotProfiles(configuration, defaultTaskManagerMemoryMB, slotsPerTaskManager);
+					final int expectedManagedMemoryMB = MemorySize.parse(configuration.getString(TaskManagerOptions.MANAGED_MEMORY_SIZE)).getMebiBytes();
+
+					assertEquals(expectedManagedMemoryMB, taskManagerInfo.getHardwareDescription().getSizeOfManagedMemory() >> 20);
```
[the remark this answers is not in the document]
xintongsong: True, I should change that.

flink-mesos/src/main/java/org/apache/flink/mesos/runtime/clusterframework/MesosResourceManager.java
xintongsong: AFAIK, we don't have any IT case for Mesos, probably because we don't have a Mesos mini cluster. So I guess it's hard to check whether TMs are started with proper configurations just like we do for Yarn.

We might have a `testCreateSlotsPerWorker` in `MesosResourceManagerTest` as well. When I first added this test case for Yarn, I thought it might be redundant to add the same test case for Mesos, because it's the same `createSlotsPerWorker` method of the common base class `ResourceManager` is invoked on both Yarn and Mesos. But now I feel, as you pointed out, that this should be useful to cover the case that `MesosResourceManager` may also update the wrong configuration instance.

flink-yarn-tests/src/test/java/org/apache/flink/yarn/YarnConfigurationITCase.java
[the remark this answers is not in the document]
azagrebin: True, both tests have to be changed, more effort. On the other side, this brings attention to them and forces to check the change against any previous assumptions expressed by tests. In this case, the real computation might change in a way that `updateTaskManagerConfigAndCreateWorkerSlotProfiles` is not used at all etc.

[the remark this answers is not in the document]
xintongsong: I tend to disagree. I don't think it is necessary for this test case to have any assumption on how `updateTaskManagerConfigAndCreateWorkerSlotProfiles` calculates managed memory. The purpose of this test case, to my understanding, is to validate that configurations are well passed to TMs rather than the calculation logics. The latter is already covered by another test case. If it does happen that the real calculation changes in a way that `updateTaskManagerConfigAndCreateWorkerSlotProfiles` is not used any more, then this test case should also fail as long as the calculation on TM side changes consistently, or otherwise the other test case `testCreateSlotsPerWorker` should fail.

flink-mesos/src/test/java/org/apache/flink/mesos/runtime/clusterframework/MesosResourceManagerTest.java
```diff
@@ -829,4 +839,45 @@ public void testClearStateAfterRevokeLeadership() throws Exception {
 			verify(rmServices.schedulerDriver).stop(true);
 		}};
 	}
+
+	/**
+	 * Tests that RM and TM calculate same slot resource profile.
+	 */
+	@Test
+	public void testCreateSlotsPerWorker() throws Exception {
```
azagrebin: Can this code be deduplicated? I guess because the relevant production code in yarn/mesos RM constructor is the same. Maybe, at least some separate test class called from both or some common component to share the same concern between yarn/mesos RMs.

xintongsong: It also came to my mind that whether these two test cases could share some common component. My concern is where should add this common component to, given that these two cases actually belong to two different modules (flink-yarn and flink-mesos). I think it might be even worse to just have one not self-contained component in one module and using it from two other modules. What do you think?

flink-yarn-tests/src/test/java/org/apache/flink/yarn/YarnConfigurationITCase.java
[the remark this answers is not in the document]
StephanEwen: I tend to agree with @xintongsong here. Tests ideally test one thing targeted, and the other aspect is already covered.

StephanEwen:
Thanks @xintongsong and @azagrebin for addressing this. I feel a bit bad for introducing this bug.

If there is an easy way to deduplicate the code between the yarn and mesos tests, it would be nice. But if this is not easily possible or not meaningful here (they don't depend on each other) I would not see this as a blocker for merging.

Otherwise, looks good!

flink-mesos/src/test/java/org/apache/flink/mesos/runtime/clusterframework/MesosResourceManagerTest.java
GJL: @xintongsong I think it is not maintenance friendly to have a verbatim copy of `YarnResourceManagerTest#testCreateSlotsPerWorker()` for Mesos:

1. These tests should be kept in sync but this is not obvious to the developer
1. It looks like that we have to copy these tests once again for the Kubernetes ResourceManager. The development efforts are already in progress:
    * https://issues.apache.org/jira/browse/FLINK-9953
    * https://issues.apache.org/jira/browse/FLINK-9495

I wonder if: 
1. it is necessary to duplicate the entire test to be confident that `updateTaskManagerConfigAndCreateWorkerSlotProfiles()` is called in the constructor of the ResourceManager?
1. duplicating the invocations of `updateTaskManagerConfigAndCreateWorkerSlotProfiles()` in the ResourceManager implementations, which consequently requires us to duplicate tests, is a design flaw?

Maybe someone who was in involved in the design of this feature can chime in (cc: @StephanEwen)

flink-yarn-tests/src/test/java/org/apache/flink/yarn/YarnConfigurationITCase.java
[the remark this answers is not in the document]
GJL: I think it's acceptable to depend on `updateTaskManagerConfigAndCreateWorkerSlotProfiles` here.

flink-mesos/src/test/java/org/apache/flink/mesos/runtime/clusterframework/MesosResourceManagerTest.java
azagrebin: By the common component I meant enabling code reuse as a benefit. Is it not possible to well define it and test once?

flink-yarn-tests/src/test/java/org/apache/flink/yarn/YarnConfigurationITCase.java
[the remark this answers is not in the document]
azagrebin: ok, agreed
it is used more like a helper function to test another thing then

flink-yarn-tests/src/test/java/org/apache/flink/yarn/YarnConfigurationITCase.java
```diff
@@ -187,6 +190,12 @@ public void testFlinkContainerMemory() throws Exception {
 					assertThat(
 						(double) taskManagerInfo.getHardwareDescription().getSizeOfJvmHeap() / (double) expectedHeadSize,
 						is(closeTo(1.0, 0.15)));
+
+					final int defaultTaskManagerMemoryMB = ConfigurationUtils.getTaskManagerHeapMemory(configuration).getMebiBytes();
+					YarnResourceManager.updateTaskManagerConfigAndCreateWorkerSlotProfiles(configuration, defaultTaskManagerMemoryMB, slotsPerTaskManager);
+					final int expectedManagedMemoryMB = MemorySize.parse(configuration.getString(TaskManagerOptions.MANAGED_MEMORY_SIZE)).getMebiBytes();
```
azagrebin: I would suggest to calculate `expectedManagedMemoryMB` in a separate function, ideally w/o implicit modification of original `configuration`. Now it does not matter, just not to spoil something potentially added later.

flink-mesos/src/test/java/org/apache/flink/mesos/runtime/clusterframework/MesosResourceManagerTest.java
xintongsong: Thank you for the suggestion, @GJL. I think this does solve the problem. In that case we should also move `slotsPerWorker` to the common base class ResourceManager, which may not be used by StandaloneResourceManager but still can be shared by Yarn/Mesos/Kubernetes ResourceManagers. I'll make an update.

flink-yarn-tests/src/test/java/org/apache/flink/yarn/YarnConfigurationITCase.java
xintongsong: You are right, we should not modify the original `configuration` here.

xintongsong:
@GJL @azagrebin @StephanEwen 
Just pushed an update moving `updateTaskManagerConfigAndCreateWorkerSlotProfiles()` and `testCreateSlotsPerWorker()` to `ResourceManager` and `ResourceManagerTest` respectively. I did moved some other properties to `ResourceManager` as well (`flinkConfig`, `numberOfTaskSlots`, and `slotsPerWorker`), and I think it makes sense to move them for a common usage for Yarn/Mesos/K8s. What do you think?

flink-runtime/src/main/java/org/apache/flink/runtime/resourcemanager/ResourceManager.java
```diff
@@ -185,6 +199,10 @@ public ResourceManager(
 		this.jmResourceIdRegistrations = new HashMap<>(4);
 		this.taskExecutors = new HashMap<>(8);
 		this.taskExecutorGatewayFutures = new HashMap<>(8);
+
+		this.defaultTaskManagerMemoryMB = ConfigurationUtils.getTaskManagerHeapMemory(flinkConfig).getMebiBytes();
+		this.numberOfTaskSlots = flinkConfig.getInteger(TaskManagerOptions.NUM_TASK_SLOTS);
+		this.slotsPerWorker = updateTaskManagerConfigAndCreateWorkerSlotProfiles(this.flinkConfig, defaultTaskManagerMemoryMB, numberOfTaskSlots);
```
azagrebin: Should this also happen in `StandaloneResourceManager`?
It just looks as if this change was kept out of `StandaloneResourceManager` on purpose or there was another reason not to do it like this before?

azagrebin: ok, I see it is actually not used in `StandaloneResourceManager`.
Yarn and mesos RMs could extend an abstract `ResourceManagerWithSlotsPerWorker` but it will probably complicate things.
It will probably need to change with the dynamic memory slicing anyways.

xintongsong: Yes, the background of this calculating `slotsPerWorker` on RM is that, we need to know the resource profiles of slots for `PendingTaskManagerSlot` before TM is started and registered. For standalone, we don't have any pending task manager slot because RM cannot actively start any TM.

flink-runtime/src/test/java/org/apache/flink/runtime/resourcemanager/ResourceManagerTest.java
```diff
@@ -270,4 +282,49 @@ private TestingResourceManager createAndStartResourceManager(HeartbeatServices h
...
+		Configuration config3 = new Configuration();
+		config3.setString(TaskManagerOptions.MANAGED_MEMORY_SIZE, "300m");
+		config3.setBoolean(TaskManagerOptions.MEMORY_OFF_HEAP, true);
+		testCreateSlotsPerWorker(config3);
+
+		Configuration config4 = new Configuration();
+		config4.setString(NettyShuffleEnvironmentOptions.NETWORK_BUFFERS_MEMORY_MAX, "10m");
+		config4.setString(NettyShuffleEnvironmentOptions.NETWORK_BUFFERS_MEMORY_MIN, "10m");
+		config4.setBoolean(TaskManagerOptions.MEMORY_OFF_HEAP, true);
+		testCreateSlotsPerWorker(config4);
```
tillrohrmann: Please split this test case up into individual tests. A single test for a single thing to test with a meaningful test name.

flink-yarn-tests/src/test/java/org/apache/flink/yarn/YarnConfigurationITCase.java
```diff
@@ -208,4 +215,12 @@ private boolean hasTaskManagerConnectedAndReportedSlots(Collection<TaskManagerIn
 			return taskManagerInfo.getNumberSlots() > 0;
 		}
 	}
+
+	private static int calculateManagedMemorySizeMB(Configuration originalConfiguration, int numSlotsPerTaskManager) {
+		Configuration configuration = new Configuration(originalConfiguration); // copy, because we alter the config
```
tillrohrmann: It is a bit weird that we have to do something like this in a test. This might indicate a design flaw of the `YarnResourceManager#updateTaskManagerConfigAndCreateWorkerSlotProfiles` method.

flink-runtime/src/test/java/org/apache/flink/runtime/resourcemanager/ResourceManagerTest.java
```diff
@@ -270,4 +282,49 @@ private TestingResourceManager createAndStartResourceManager(HeartbeatServices h
...
+	}
+
+	private void testCreateSlotsPerWorker(Configuration config) throws  Exception {
+		resourceManager = createAndStartResourceManager(heartbeatServices, config);
```
tillrohrmann: I feels a bit weird that we are instantiating the `ResourceManager` in order to test a static helper method. I think this is an indicator that something is wrong with the chosen abstractions.

flink-runtime/src/main/java/org/apache/flink/runtime/resourcemanager/ResourceManager.java
```diff
@@ -1203,6 +1221,16 @@ protected int getNumberRequiredTaskManagerSlots() {
...
 
+	@VisibleForTesting
+	Configuration getFlinkConfig() {
+		return flinkConfig;
+	}
+
+	@VisibleForTesting
+	Collection<ResourceProfile> getSlotsPerWorker() {
+		return slotsPerWorker;
+	}
```
tillrohrmann: Getting access to internal state is usually an indicator that our class under test is too powerful and is not well suited for testing.

flink-runtime/src/main/java/org/apache/flink/runtime/resourcemanager/ResourceManager.java
tillrohrmann: Can we move these computations out of the `ResourceManager`? It seems as if we only rely on static helper methods. We could create a `TaskManagerSpecification` object containing the required information. That way, it would also be easier to test the correct computation without having to introduce these `VisibleForTesting` methods and having to create `RpcEndpoints` for only testing static helper methods.

flink-yarn-tests/src/test/java/org/apache/flink/yarn/YarnConfigurationITCase.java
```diff
@@ -208,4 +215,12 @@ private boolean hasTaskManagerConnectedAndReportedSlots(Collection<TaskManagerIn
...
+
+		final int defaultTaskManagerMemoryMB = ConfigurationUtils.getTaskManagerHeapMemory(configuration).getMebiBytes();
+		YarnResourceManager.updateTaskManagerConfigAndCreateWorkerSlotProfiles(configuration, defaultTaskManagerMemoryMB, numSlotsPerTaskManager);
```
tillrohrmann: I would suggest to let such a method return the modified configuration instead of using an output parameter.

xintongsong:
Thanks for the comments, @tillrohrmann. I find them very helpful. Will update the PR addressing your comments ASAP.

xintongsong:
@tillrohrmann 
I've addressed the comments and updated the PR. Also rebased to the latest release-1.9 branch resolving the new conflicts. Below is a brief change log.
- 3b20ed5c8684466d59220d13019b5f481c78a3cd..8369fc5c36adeb4a27b0c83653190de4278dbc90: These are the original commits of fixing the problem for setting wrong configuration instance.
- Then I reverted the rest of the commits that move `updateTaskManagerConfigAndCreateWorkerSlotProfiles()` and `testCreateSlotsPerWorker()` to `ResourceManager` and `ResourceManagerTest` respectively.
- a31e73a347e357c1a6fedcd1bad322302848d680: Introduce `taskManagerConfig` and `slotResourceProfile` and the computing of slot resource profile in `ContaineredTaskManagerParameters`.
- 1395bd7a7ac7a8e9e6801685415337a0d6edec71: Use `ContaineredTaskManagerParameters` to compute and record default configs and slot resource profiles of task managers in Yarn / Mesos.
-  779d3038af44c64cdb3b30fe40e9ddff88116ffd..70d2ed3eebc9874eb70559fc5a72b5366392b0df: Test updates.
The PR has already passed the travis test.  https://travis-ci.org/xintongsong/flink/builds/563897220

tillrohrmann:
@xintongsong please also open a PR against the `master` branch since we need to forward port this fix.

flink-yarn/src/main/java/org/apache/flink/yarn/YarnResourceManager.java
```diff
@@ -184,7 +183,9 @@ public YarnResourceManager(
 		this.defaultCpus = flinkConfig.getInteger(YarnConfigOptions.VCORES, numberOfTaskSlots);
 		this.resource = Resource.newInstance(defaultTaskManagerMemoryMB, defaultCpus);
 
-		this.slotsPerWorker = updateTaskManagerConfigAndCreateWorkerSlotProfiles(flinkConfig, defaultTaskManagerMemoryMB, numberOfTaskSlots);
+		this.defaultTaskManagerParameters = ContaineredTaskManagerParameters.create(flinkConfig, defaultTaskManagerMemoryMB, numberOfTaskSlots);
```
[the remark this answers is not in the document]
xintongsong: Yes, the `createTaskExecutorLaunchContext ` recreates a `ContaineredTaskManagerParameters`, but with the actual container memory allocated by Yarn, which could be slightly larger than requested. Reusing the same `ContaineredTaskManagerParameters` instance prevents task managers from making use of the extra allocated memory, which could be a user aware behavior change compare to the current behavior. I think maybe we should not introduce such behavior change in this version, since we are going to change resource configuration and management anyway in the next version.

xintongsong:
@tillrohrmann 
PR is updated and passed [travis](https://travis-ci.org/xintongsong/flink/builds/564304178).
New PR #9246 opened against `master` and also passed [travis](https://travis-ci.org/xintongsong/flink/builds/564304333).

flink-mesos/src/main/java/org/apache/flink/mesos/runtime/clusterframework/MesosResourceManagerFactory.java
```diff
@@ -99,4 +100,11 @@ public MesosResourceManagerFactory(@Nonnull MesosServices mesosServices, @Nonnul
 			webInterfaceUrl,
 			jobManagerMetricGroup);
 	}
+
+	@Override
+	public Configuration getResourceManagerConfiguration(Configuration flinkConfig) {
+		final Configuration config = new Configuration(flinkConfig);
+		config.setString(TaskManagerOptions.MANAGED_MEMORY_SIZE, flinkConfig.getInteger(MesosTaskManagerParameters.MESOS_RM_TASKS_MEMORY_MB) + "m");
```
tillrohrmann: Isn't `MESOS_RM_TASKS_MEMORY_MB` the memory a Mesos task should get assigned (aka the total container memory size)? Why did we introduce this change in this PR? It seems a bit unrelated.

flink-runtime/src/main/java/org/apache/flink/runtime/resourcemanager/ResourceManagerFactory.java
```diff
@@ -59,4 +63,16 @@ default String generateEndpointIdWithUUID() {
 	default String getEndpointId() {
 		return ResourceManager.RESOURCE_MANAGER_NAME;
 	}
+
+	default Configuration getResourceManagerConfiguration(Configuration flinkConfig) {
```
tillrohrmann: Why do we make this method part of the `ResourceManagerFactory`? It seems to me that we are mixing concerns here (creating a `ResourceManager` vs. properly setting up the ResourceManager's `configuration`).

xintongsong:
close in 86491b6b4ccc8c9ef5fa15e4755e2e5c252c29ad

PR #9106, opened 2019-07-14
[FLINK-13184][yarn] Support launching task executors with multi-thread on YARN.

xintongsong:

This pull request support starting new TaskExecutors with multi-thread in YarnResourceManager, so when starting a large amount of TaskExecutors it won't the block main thread on RM from responding to other RPC messages (TE registration, heartbeats, etc.).

## Brief change log

- b2ed97d0e6dd506602888aacf6146238fbcdfd48: Introduce a config option for the max number of threads on RM used for starting new TaskExecutors. 
- 06c7209af04f49ca69a263f57fc713f3b9c55c87: Introduce a thread pool in YarnResourceManager and start new TaskExecutors with multi-thread. 
- 25fc95f30720209e19bd010cdd517ff5e3c685d8: Update relevant test cases to wait for threads starting new TaskExecutors to finish.

## Verifying this change

  - Updated YarnResourceManagerTest to verify that YarnResourceManager starts new TaskExecutors properly.

  - The serializers: (no)

## Documentation

### DIFF
--- a/docs/_includes/generated/yarn_config_configuration.html
+++ b/docs/_includes/generated/yarn_config_configuration.html
@@ -54,7 +54,7 @@
         </tr>
         <tr>
             <td><h5>yarn.max-container-launcher-number</h5></td>
-            <td style="word-wrap: break-word;">100</td>
+            <td style="word-wrap: break-word;">10</td>
             <td>The max number of threads for starting yarn containers in yarn resource manager.</td>
         </tr>
         <tr>
--- a/docs/_includes/generated/yarn_config_configuration.html
+++ b/docs/_includes/generated/yarn_config_configuration.html
@@ -53,6 +53,11 @@
             <td>Time between heartbeats with the ResourceManager in seconds.</td>
         </tr>
         <tr>
+            <td><h5>yarn.max-container-launcher-number</h5></td>
+            <td style="word-wrap: break-word;">100</td>
+            <td>The max number of threads for starting yarn containers in yarn resource manager.</td>
+        </tr>
+        <tr>
             <td><h5>yarn.maximum-failed-containers</h5></td>
             <td style="word-wrap: break-word;">(none)</td>
             <td>Maximum number of containers the system is going to reallocate in case of a failure.</td>
--- a/flink-yarn/src/main/java/org/apache/flink/yarn/YarnResourceManager.java
+++ b/flink-yarn/src/main/java/org/apache/flink/yarn/YarnResourceManager.java
@@ -140,7 +140,7 @@ public class YarnResourceManager extends ResourceManager<YarnWorkerNode> impleme
 	/**
 	 * Executor for starting new yarn containers.
 	 */
-	private final Executor startContainerExecutor;
+	private final ThreadPoolExecutor startContainerExecutor;
 
 	public YarnResourceManager(
 			RpcService rpcService,
@@ -289,6 +289,14 @@ public class YarnResourceManager extends ResourceManager<YarnWorkerNode> impleme
 		if (nodeManagerClient != null) {
 			try {
 				nodeManagerClient.stop();
+			} catch (Throwable t) {
+				firstException = ExceptionUtils.firstOrSuppressed(t, firstException);
+			}
+		}
+
+		if (startContainerExecutor != null) {
+			try {
+				startContainerExecutor.shutdownNow();
 			} catch (Throwable t) {
 				firstException = ExceptionUtils.firstOrSuppressed(t, firstException);
 			}
--- a/flink-yarn/src/main/java/org/apache/flink/yarn/YarnResourceManager.java
+++ b/flink-yarn/src/main/java/org/apache/flink/yarn/YarnResourceManager.java
@@ -599,4 +599,9 @@ public class YarnResourceManager extends ResourceManager<YarnWorkerNode> impleme
 				.put(ENV_FLINK_NODE_ID, host);
 		return taskExecutorLaunchContext;
 	}
+
+	@VisibleForTesting
+	Executor getStartContainerExecutor() {
+		return startContainerExecutor;
+	}
 }
--- a/flink-yarn/src/main/java/org/apache/flink/yarn/YarnResourceManager.java
+++ b/flink-yarn/src/main/java/org/apache/flink/yarn/YarnResourceManager.java
@@ -46,6 +46,8 @@ import org.apache.flink.util.FlinkException;
 import org.apache.flink.util.Preconditions;
 import org.apache.flink.yarn.configuration.YarnConfigOptions;
 
+import org.apache.flink.shaded.guava18.com.google.common.util.concurrent.ThreadFactoryBuilder;
+
 import org.apache.hadoop.yarn.api.ApplicationConstants;
 import org.apache.hadoop.yarn.api.protocolrecords.RegisterApplicationMasterResponse;
 import org.apache.hadoop.yarn.api.records.Container;
@@ -76,6 +78,10 @@ import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.Executor;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
 
 /**
  * The yarn implementation of the resource manager. Used when the system is started
@@ -131,6 +137,11 @@ public class YarnResourceManager extends ResourceManager<YarnWorkerNode> impleme
 
 	private final Resource resource;
 
+	/**
+	 * Executor for starting new yarn containers.
+	 */
+	private final Executor startContainerExecutor;
+
 	public YarnResourceManager(
 			RpcService rpcService,
 			String resourceManagerEndpointId,
@@ -185,6 +196,10 @@ public class YarnResourceManager extends ResourceManager<YarnWorkerNode> impleme
 		this.resource = Resource.newInstance(defaultTaskManagerMemoryMB, defaultCpus);
 
 		this.slotsPerWorker = createWorkerSlotProfiles(flinkConfig);
+		this.startContainerExecutor = new ThreadPoolExecutor(0, flinkConfig.getInteger(YarnConfigOptions.CONTAINER_LAUNCHER_NUM_MAX),
+			60L, TimeUnit.SECONDS, new LinkedBlockingQueue<>(), new ThreadFactoryBuilder()
+			.setNameFormat("ContainerLauncher #%d")
+			.build());
 	}
 
 	protected AMRMClientAsync<AMRMClient.ContainerRequest> createAndStartResourceManagerClient(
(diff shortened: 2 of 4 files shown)


flink-yarn/src/main/java/org/apache/flink/yarn/YarnResourceManager.java
```diff
@@ -131,6 +137,11 @@
 
 	private final Resource resource;
 
+	/**
+	 * Executor for starting new yarn containers.
+	 */
+	private final Executor startContainerExecutor;
```
guoweiM: This executor might need to be shutdown on 'onStop' is called.

xintongsong: I would rather keep it the current way, for 2 reasons.
1. It does not cause any resource leak. The `ThreadPoolExecutor` is configured that pooled threads will be terminated if there's no tasks for 60s, and the executor itself is shutdown automatically if there is no remaining threads and it is no longer referenced.
2. The general `Executor` interface does not have a `shutdown` method. And I would like to keep the thread pool declared in the general type `Executor`, in case later we want to use a different executor implementation.

flink-yarn/src/main/java/org/apache/flink/yarn/configuration/YarnConfigOptions.java
```diff
@@ -188,6 +188,14 @@
 		.defaultValue("")
 		.withDescription("A comma-separated list of tags to apply to the Flink YARN application.");
 
+	/**
+	 * The max number of threads for starting yarn containers in yarn resource manager.
+	 */
+	public static final ConfigOption<Integer> CONTAINER_LAUNCHER_NUM_MAX =
+		key("yarn.max-container-launcher-number")
+			.defaultValue(100)
```
wangyang0918: Set it to 10 and is enough for launching more than 2000 containers.

xintongsong:
@wangyang0918 @guoweiM 
Thanks for the review. I've updated the PR. Also rebased to the latest `master` for conflict resolving.

tillrohrmann:
Is this the only blocking call we execute from the `YarnResourceManager` or could there be others as well. If yes, then it might make sense to have a plan how to fix the other call-sites as well.

flink-yarn/src/main/java/org/apache/flink/yarn/YarnResourceManager.java
```diff
@@ -131,6 +137,11 @@
 
 	private final Resource resource;
 
+	/**
+	 * Executor for starting new yarn containers.
+	 */
+	private final ThreadPoolExecutor startContainerExecutor;
```
tillrohrmann: Wouldn't an `ExecutorService` be enough here?

flink-yarn/src/main/java/org/apache/flink/yarn/YarnResourceManager.java
```diff
@@ -185,6 +196,12 @@ public YarnResourceManager(
 		this.resource = Resource.newInstance(defaultTaskManagerMemoryMB, defaultCpus);
 
 		this.slotsPerWorker = createWorkerSlotProfiles(flinkConfig);
+
+		final int containerLauncherNum = flinkConfig.getInteger(YarnConfigOptions.CONTAINER_LAUNCHER_NUM_MAX);
+		this.startContainerExecutor = new ThreadPoolExecutor(containerLauncherNum, containerLauncherNum,
```
tillrohrmann: Easier to use `Executors.newFixedThreadPool(size, ExecutorThreadFactory)`

tillrohrmann: I think it would also be better to pass the `ExecutorService` into the `YarnResourceManager` using dependency injection.

flink-yarn/src/main/java/org/apache/flink/yarn/YarnResourceManager.java
tillrohrmann: Let's rename the `Executor` into `blockingCallsExecutor`, `yarnCallsExecutor` or `ioExecutor`

flink-yarn/src/main/java/org/apache/flink/yarn/YarnResourceManager.java
```diff
@@ -574,4 +609,9 @@ private ContainerLaunchContext createTaskExecutorLaunchContext(Resource resource
 				.put(ENV_FLINK_NODE_ID, host);
 		return taskExecutorLaunchContext;
 	}
+
+	@VisibleForTesting
+	Executor getStartContainerExecutor() {
+		return startContainerExecutor;
+	}
```
tillrohrmann: This looks like a code smell. If we pass in the `ExecutorService`, then there should be no need for a `@VisibleForTesting` method.

flink-yarn/src/main/java/org/apache/flink/yarn/configuration/YarnConfigOptions.java
```diff
@@ -188,6 +188,14 @@
 		.defaultValue("")
 		.withDescription("A comma-separated list of tags to apply to the Flink YARN application.");
 
+	/**
+	 * The max number of threads for starting yarn containers in yarn resource manager.
+	 */
+	public static final ConfigOption<Integer> CONTAINER_LAUNCHER_NUM_MAX =
+		key("yarn.max-container-launcher-number")
+			.defaultValue(10)
```
tillrohrmann: I would set the default value to `1`.

flink-yarn/src/main/java/org/apache/flink/yarn/configuration/YarnConfigOptions.java
```diff
@@ -188,6 +188,14 @@
 		.defaultValue("")
 		.withDescription("A comma-separated list of tags to apply to the Flink YARN application.");
 
+	/**
+	 * The max number of threads for starting yarn containers in yarn resource manager.
+	 */
+	public static final ConfigOption<Integer> CONTAINER_LAUNCHER_NUM_MAX =
+		key("yarn.max-container-launcher-number")
```
tillrohrmann: Maybe call this option `yarn.resourcemanager.io-executor.num-threads`

flink-yarn/src/main/java/org/apache/flink/yarn/configuration/YarnConfigOptions.java
```diff
@@ -188,6 +188,14 @@
 		.defaultValue("")
 		.withDescription("A comma-separated list of tags to apply to the Flink YARN application.");
 
+	/**
+	 * The max number of threads for starting yarn containers in yarn resource manager.
+	 */
+	public static final ConfigOption<Integer> CONTAINER_LAUNCHER_NUM_MAX =
```
tillrohrmann: Do we need to expose this option? I'm wondering how many people will actually set it.

flink-yarn/src/main/java/org/apache/flink/yarn/YarnResourceManager.java
```diff
@@ -391,23 +416,33 @@ public void onContainersAllocated(List<Container> containers) {
...
+
+								nodeManagerClient.startContainer(container, taskExecutorLaunchContext);
+							} catch (Throwable t) {
+								log.error("Could not start TaskManager in container {}.", container.getId(), t);
+
+								// release the failed container
+								workerNodeMap.remove(resourceId);
+								resourceManagerClient.releaseAssignedContainer(container.getId());
+								// and ask for a new one
+								requestYarnContainerIfRequired();
```
tillrohrmann: This method must not be called from outside of the `RpcEndpoint's` main thread because it mutates the internal state of the `ResourceManager`.

flink-yarn/src/main/java/org/apache/flink/yarn/YarnResourceManager.java
```diff
@@ -391,23 +416,33 @@ public void onContainersAllocated(List<Container> containers) {
...
+								return;
+							}
+							try {
+								// Context information used to start a TaskExecutor Java process
+								ContainerLaunchContext taskExecutorLaunchContext = createTaskExecutorLaunchContext(
+									container.getResource(),
+									containerIdStr,
+									container.getNodeId().getHost());
+
+								nodeManagerClient.startContainer(container, taskExecutorLaunchContext);
```
tillrohrmann: This method is thread safe?

flink-yarn/src/main/java/org/apache/flink/yarn/YarnResourceManager.java
```diff
@@ -391,23 +416,33 @@ public void onContainersAllocated(List<Container> containers) {
...
+							} catch (Throwable t) {
+								log.error("Could not start TaskManager in container {}.", container.getId(), t);
+
+								// release the failed container
+								workerNodeMap.remove(resourceId);
+								resourceManagerClient.releaseAssignedContainer(container.getId());
```
tillrohrmann: This method is thread safe?

flink-yarn/src/test/java/org/apache/flink/yarn/YarnResourceManagerTest.java
```diff
@@ -310,6 +311,14 @@ void runTest(RunnableWithException testMethod) throws Exception {
 				stopResourceManager();
 			}
 		}
+
+		public void waitForThreadPoolExecution() throws Exception {
+			ThreadPoolExecutor executor = (ThreadPoolExecutor) resourceManager.getStartContainerExecutor();
+			while (!(executor.getQueue().size() == 0 && executor.getActiveCount() == 0)) {
+				log.info("Waiting for all launch container requests been executed.");
+				Thread.sleep(500);
```
tillrohrmann: sleep is too long

flink-yarn/src/main/java/org/apache/flink/yarn/YarnResourceManager.java
xintongsong: Yes, it is. I checked the source codes myself, and also double checked with Tao Yang who is an apache hadoop committer.

flink-yarn/src/main/java/org/apache/flink/yarn/YarnResourceManager.java
xintongsong: Yes, it is. I checked the source codes myself, and also double checked with Tao Yang who is an apache hadoop committer.

flink-yarn/src/main/java/org/apache/flink/yarn/configuration/YarnConfigOptions.java
xintongsong: I think we need to expose option to make the thread number configurable. We run into many cases in our production where we need to set the thread number to 100 for large scale jobs to meet our SLA for deploying a job cluster. On the other hand, having too many threads consumes significant unnecessary resources for small jobs.

xintongsong:
Thanks for the review, @tillrohrmann.
I've left feedbacks on some of your comments. Will update the PR addressing the other comments. I'll ping you when I finish.

flink-yarn/src/main/java/org/apache/flink/yarn/configuration/YarnConfigOptions.java
tillrohrmann: I guess I would need to see some benchmarks to see the positive effect of having 100 threads instead of single worker thread. What I could see is if the Yarn site can work on multiple requests in parallel but even then there must be an upper limit to it. Imagine multiple Flink ResourceManager which bombard the Yarn NodeManager with 100s of container requests. This does not feel right.

xintongsong: I think the benefit of having many threads starting containers mainly comes from that Flink RM can concurrently start containers from different Yarn NMs. Imagine a job with 1000s of containers on a Yarn cluster of 100s of NMs, ideally Flink RM would be able to concurrently start containers from 100 different NMs, and averagely each NM only have about 10 containers to start. The burden for each NM is affordable even with multiple Flink RMs, because it is not a common case that all the Flink RMs start massive of containers at the same time.

I can try provide some benchmarks. This is not a rare case in our production.

tillrohrmann: Hmm ok, I think the problem arises because we are using the `NMClient` instead of the `NMClientAsync`. Hence the `startContainer` waits for the completion. Wouldn't using the `NMClientAsync` be the much better solution instead of introducing a new `ExecutorService` which makes the `NMClient` asynchronous?

flink-yarn/src/main/java/org/apache/flink/yarn/YarnResourceManager.java
walterddr: +1. according to the [Javadoc](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ExecutorService.html): `ExecutorService` can "provides methods to manage termination and methods that can produce a {@link Future} for tracking progress of one or more asynchronous tasks." seems to be a better fit for the purpose.

walterddr: also, I am considering, should we enforce a hard limit on how many threads a user can configure instead of a MAX_INT? currently the `maximumPoolSize` argument is set equivalent to `corePoolSize`.

docs/_includes/generated/yarn_config_configuration.html
```diff
@@ -52,6 +52,11 @@
             <td style="word-wrap: break-word;">5</td>
             <td>Time between heartbeats with the ResourceManager in seconds.</td>
         </tr>
+        <tr>
+            <td><h5>yarn.max-container-launcher-number</h5></td>
+            <td style="word-wrap: break-word;">10</td>
+            <td>The max number of threads for starting yarn containers in yarn resource manager.</td>
```
walterddr: technically, this argument allows user to set the "EXACT" num of thread for strarting, instead of the maximum.

xintongsong:
Thanks for the review, @walterddr.

Till has suggested that instead of maintaining a thread pool in Flink, we may simply leverage Yarn `NMClientAsync` (currently we are using `NMClient`) for starting new task executors. And I tend to agree with him. I'll try do some investigation along this direction (and maybe some POC).

WDYT?

tillrohrmann:
Closing this PR since it needs a major overhaul.

PR #9107, opened 2019-07-15
[FLINK-13253][jdbc] Deadlock may occur in JDBCUpsertOutputFormat

JingsongLi:

In close, it await the flush scheduler terminal, but it hold the lock of JDBCUpsertOutputFormat instance, maybe the async thread is waiting for this lock in the flush method, so there might be a deadlock here.

First, it should not await scheduler terminal, because it has flushed all data to jdbc, what we should do is let async thread quit.
Second, we should add lock outside the closed check in the flusher, in this way, we can ensure async thread secure exiting.

## Verifying this change

Manual test

  - The serializers: no

## Documentation

### DIFF
--- a/flink-connectors/flink-jdbc/src/main/java/org/apache/flink/api/java/io/jdbc/JDBCUpsertOutputFormat.java
+++ b/flink-connectors/flink-jdbc/src/main/java/org/apache/flink/api/java/io/jdbc/JDBCUpsertOutputFormat.java
@@ -116,13 +116,15 @@ public class JDBCUpsertOutputFormat extends AbstractJDBCOutputFormat<Tuple2<Bool
 			this.scheduler = Executors.newScheduledThreadPool(
 					1, new ExecutorThreadFactory("jdbc-upsert-output-format"));
 			this.scheduledFuture = this.scheduler.scheduleWithFixedDelay(() -> {
-				if (closed) {
-					return;
-				}
-				try {
-					flush();
-				} catch (Exception e) {
-					flushException = e;
+				synchronized (JDBCUpsertOutputFormat.this) {
+					if (closed) {
+						return;
+					}
+					try {
+						flush();
+					} catch (Exception e) {
+						flushException = e;
+					}
 				}
 			}, flushIntervalMills, flushIntervalMills, TimeUnit.MILLISECONDS);
 		}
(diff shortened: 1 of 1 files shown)


flink-connectors/flink-jdbc/src/main/java/org/apache/flink/api/java/io/jdbc/JDBCUpsertOutputFormat.java
```diff
@@ -116,13 +116,15 @@ public void open(int taskNumber, int numTasks) throws IOException {
...
 					1, new ExecutorThreadFactory("jdbc-upsert-output-format"));
 			this.scheduledFuture = this.scheduler.scheduleWithFixedDelay(() -> {
-				if (closed) {
-					return;
-				}
-				try {
-					flush();
-				} catch (Exception e) {
-					flushException = e;
+				synchronized (JDBCUpsertOutputFormat.this) {
```
wuchong: Why should we add `synchronized`? `flush()` is already `synchronized`, right?

flink-connectors/flink-jdbc/src/main/java/org/apache/flink/api/java/io/jdbc/JDBCUpsertOutputFormat.java
```diff
@@ -184,15 +186,6 @@ public synchronized void close() throws IOException {
 		if (this.scheduledFuture != null) {
 			scheduledFuture.cancel(false);
 			this.scheduler.shutdown();
```
wuchong: Should we change this to `shutdownNow()`? I think the async flusher can quit now.

flink-connectors/flink-jdbc/src/main/java/org/apache/flink/api/java/io/jdbc/JDBCUpsertOutputFormat.java
JingsongLi: Make sure the flusher will graceful exit after close.

flink-connectors/flink-jdbc/src/main/java/org/apache/flink/api/java/io/jdbc/JDBCUpsertOutputFormat.java
JingsongLi: In this case, I think we should not interrupt thread, we have good way to deal with it now, interrupt may make things complicate.

PR #9115, opened 2019-07-15
[FLINK-13124] Don't forward exceptions when finishing SourceStreamTask

aljoscha:

Before, exceptions that occurred after cancelling a source (as the
KafkaConsumer did, for example) would make a job fail when attempting a
"stop-with-savepoint". Now we ignore those exceptions.

## Brief change log

- add a `isFinished` flag in `SourceStreamTask`
- check this flag before re-throwing exceptions

## Verifying this change

- Added unit tests in `SourceStreamTaskTest`

  - Checkpointing, yes, because it fixes stop-with-savepoint

## Documentation


kl0u:
Merge on 1.9.

PR #9117, opened 2019-07-15
[FLINK-13223][dist] Enable fine-grained recovery by default in distribution

zentol:
Enables fine-grained recovery by default in the flink distribution. We don't enable it by default for all tests since too many tests make implicit assumptions about the failover behavior, that is they rely on every task being restarted on failure.

Change log:
* disables `FORCE_PARTITION_RELEASE_ON_CONSUMPTION` as otherwise fine-grained recovery cannot really work; this change should be safe for tests
* sets the failover strategy to "region" in the flink-conf.yaml contained in the distribution
* disables fine-grained recovery in the StreamingFileSink E2E test since it relies on global failovers to occur

@KurtYoung @tzulitai I would like to include this in 1.9, as otherwise I fear this feature will go largely unnoticed by new users and during testing.

### DIFF
--- a/flink-core/src/main/java/org/apache/flink/configuration/JobManagerOptions.java
+++ b/flink-core/src/main/java/org/apache/flink/configuration/JobManagerOptions.java
@@ -186,7 +186,7 @@ public class JobManagerOptions {
 	@Documentation.ExcludeFromDocumentation("dev use only; likely temporary")
 	public static final ConfigOption<Boolean> FORCE_PARTITION_RELEASE_ON_CONSUMPTION =
 			key("jobmanager.scheduler.partition.force-release-on-consumption")
-			.defaultValue(true);
+			.defaultValue(false);
 
 	// ---------------------------------------------------------------------------------------------
 
--- a/flink-end-to-end-tests/test-scripts/test_streaming_file_sink.sh
+++ b/flink-end-to-end-tests/test-scripts/test_streaming_file_sink.sh
@@ -29,6 +29,8 @@ echo "Executing test with ${OPENSSL_LINKAGE} openSSL linkage (random selection b
 s3_setup hadoop
 set_conf_ssl "mutual" "OPENSSL" "${OPENSSL_LINKAGE}"
 set_config_key "metrics.fetcher.update-interval" "2000"
+# this test relies on global failovers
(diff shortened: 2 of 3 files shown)


flink-core/src/main/java/org/apache/flink/configuration/JobManagerOptions.java
```diff
@@ -186,7 +186,7 @@
 	@Documentation.ExcludeFromDocumentation("dev use only; likely temporary")
 	public static final ConfigOption<Boolean> FORCE_PARTITION_RELEASE_ON_CONSUMPTION =
 			key("jobmanager.scheduler.partition.force-release-on-consumption")
-			.defaultValue(true);
+			.defaultValue(false);
```
tillrohrmann: Quick question: How would Flink behave if failover strategy is set to `full` and this config option being set to `false`? Would we keep blocking result partitions until a pipelined region has been completely executed?

flink-end-to-end-tests/test-scripts/test_streaming_file_sink.sh
```diff
@@ -29,6 +29,8 @@ echo "Executing test with ${OPENSSL_LINKAGE} openSSL linkage (random selection b
 s3_setup hadoop
 set_conf_ssl "mutual" "OPENSSL" "${OPENSSL_LINKAGE}"
 set_config_key "metrics.fetcher.update-interval" "2000"
+# this test relies on global failovers
+set_config_key "jobmanager.execution.failover-strategy", "full"
```
1u0: The comma between key and value is not needed, it would result in a wrong config key.

flink-core/src/main/java/org/apache/flink/configuration/JobManagerOptions.java
zentol: the partitions will be released when the vertices are being reset; i.e. on each full restart all partitions would be released.

zentol: if no failures occur, then yes we will keep partitions until the failover region is consumed.

Long-term we should couple the partition release logic / releasedOnConsumption flag with the failover strategy so we don't keep partitions around for no reason.

PR #9120, opened 2019-07-15
[FLINK-13216][FLINK-13153][table-planner-blink] Fix Max_Retract and Min_Retract may produce incorrect result

wuchong:

Fix unstable tests failed on travis, including `AggregateITCase.testNestedGroupByAgg` and `SplitAggregateITCase.testMinMaxWithRetraction`.

## Brief change log

The reason is we didn't set null to acc.max/min which may have an old value when we need to get a new max/min from an empty MapView. And the old value will be output to downstream instead of a null value. This influences the final result. 

This causes many unstable cases, which all contains max/min with retract, including:
- AggregateITCase.testNestedGroupByAgg
- SplitAggregateITCase.testMinMaxWithRetraction

## Verifying this change

The situation is not reproduced 100%. I tested in my local machine, for example, `AggregateITCase.testNestedGroupByAgg` can reproduce this problem every 100 times. After fix this problem, I didn't reproduce this problem after running 1000 times.

## Documentation

### DIFF
--- a/flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/functions/aggfunctions/MaxWithRetractAggFunction.java
+++ b/flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/functions/aggfunctions/MaxWithRetractAggFunction.java
@@ -135,6 +135,8 @@ public abstract class MaxWithRetractAggFunction<T extends Comparable>
 		// when both of them are expired.
 		if (!hasMax) {
 			acc.mapSize = 0L;
+			// we should also override max value, because it may have an old value.
+			acc.max = null;
 		}
 	}
 
@@ -142,7 +144,7 @@ public abstract class MaxWithRetractAggFunction<T extends Comparable>
 		boolean needUpdateMax = false;
 		for (MaxWithRetractAccumulator<T> a : its) {
 			// set max element
-			if (acc.mapSize == 0 || (a.max != null && acc.max.compareTo(a.max) < 0)) {
+			if (acc.mapSize == 0 || (a.mapSize > 0 && a.max != null && acc.max.compareTo(a.max) < 0)) {
 				acc.max = a.max;
 			}
 			// merge the count for each key
@@ -195,7 +197,7 @@ public abstract class MaxWithRetractAggFunction<T extends Comparable>
 
 	@Override
 	public T getValue(MaxWithRetractAccumulator<T> acc) {
-		if (acc.mapSize != 0) {
+		if (acc.mapSize > 0) {
 			return acc.max;
 		} else {
 			return null;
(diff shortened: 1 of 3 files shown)


godfreyhe:
thanks for this pr @wuchong , could your add some tests to recur the failure deterministically ?

wuchong:
Hi @godfreyhe ,  the failure happens in a very extreme situation which can't be reproduced easily in IT case. However, I can add some unit tests to verify this.

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/functions/aggfunctions/MaxWithRetractAggFunction.java
```diff
@@ -135,6 +135,8 @@ private void updateMax(MaxWithRetractAccumulator<T> acc) throws Exception {
 		// when both of them are expired.
```
JingsongLi: I am confused about this comments, expired?
If it is an empty map, why there is a max?
Should set max to null at the moment of becoming empty?

JingsongLi:
> Hi @godfreyhe , the failure happens in a very extreme situation which can't be reproduced easily in IT case. However, I can add some unit tests to verify this.

I think single parallelism behavior should be deterministic?

wuchong:
Hi @JingsongLi , the single parallelism is not deterministic too with a lower reproduce rate.
Hi @godfreyhe , I just added a unit test to verify this which also illustrate which situation will cause this problem.

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/functions/aggfunctions/MaxWithRetractAggFunction.java
wuchong: Yes. This is the problem. We should set max to null when the map is empty.

The "expired" means the content of MapView might be cleared because of state ttl.

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/functions/aggfunctions/MinWithRetractAggFunction.java
```diff
@@ -195,7 +197,7 @@ public void resetAccumulator(MinWithRetractAccumulator<T> acc) {
 
 	@Override
 	public T getValue(MinWithRetractAccumulator<T> acc) {
-		if (acc.mapSize != 0) {
+		if (acc.mapSize > 0) {
```
wuchong: I just added this for safety. This is not the cause of this problem.

wuchong:
@godfreyhe , updated.

godfreyhe:
@wuchong thanks for the update, LGTM. @flinkbot approve all

wuchong:
Travis passed except a SQL CLI case failed (FLINK-13282)

https://travis-ci.org/wuchong/flink/builds/559247226

PR #9123, opened 2019-07-16
[FLINK-13281] Fix the verify logic of AdaptedRestartPipelinedRegionStrategyNGAbortPendingCheckpointsTest

Myasuka:

[FLINK-13060](https://issues.apache.org/jira/browse/FLINK-13060) ensure FailoverStrategies should respect restart constraints. However, this change did not cover the `AdaptedRestartPipelinedRegionStrategyNGAbortPendingCheckpointsTest` which should also set as `FixedDelayRestartStrategy`. Currently, as the pending checkpoint would still be discarded due to `failUnacknowledgedPendingCheckpoints` within `CheckpointCoordinator`, `AdaptedRestartPipelinedRegionStrategyNGAbortPendingCheckpointsTest` could pass the test.

This is actually not what we want the test did, we should actually verify that when region failover, pending checkpoints would be discarded.

## Brief change log

* Fix the verify logic of `AdaptedRestartPipelinedRegionStrategyNGAbortPendingCheckpointsTest`

## Verifying this change

This change modify previous test `AdaptedRestartPipelinedRegionStrategyNGAbortPendingCheckpointsTest`

  - The serializers: **no**

## Documentation

### DIFF
--- a/flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/AdaptedRestartPipelinedRegionStrategyNGAbortPendingCheckpointsTest.java
+++ b/flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/AdaptedRestartPipelinedRegionStrategyNGAbortPendingCheckpointsTest.java
@@ -21,22 +21,25 @@ package org.apache.flink.runtime.executiongraph;
 
 import org.apache.flink.metrics.groups.UnregisteredMetricsGroup;
 import org.apache.flink.runtime.checkpoint.CheckpointCoordinator;
+import org.apache.flink.runtime.checkpoint.CheckpointMetrics;
 import org.apache.flink.runtime.checkpoint.CheckpointRetentionPolicy;
 import org.apache.flink.runtime.checkpoint.CheckpointStatsTracker;
 import org.apache.flink.runtime.checkpoint.StandaloneCheckpointIDCounter;
 import org.apache.flink.runtime.checkpoint.StandaloneCompletedCheckpointStore;
+import org.apache.flink.runtime.checkpoint.TaskStateSnapshot;
 import org.apache.flink.runtime.concurrent.ComponentMainThreadExecutor;
 import org.apache.flink.runtime.concurrent.ComponentMainThreadExecutorServiceAdapter;
 import org.apache.flink.runtime.concurrent.ManuallyTriggeredScheduledExecutor;
 import org.apache.flink.runtime.execution.ExecutionState;
 import org.apache.flink.runtime.executiongraph.failover.AdaptedRestartPipelinedRegionStrategyNG;
-import org.apache.flink.runtime.executiongraph.restart.InfiniteDelayRestartStrategy;
+import org.apache.flink.runtime.executiongraph.restart.FixedDelayRestartStrategy;
 import org.apache.flink.runtime.executiongraph.utils.SimpleSlotProvider;
 import org.apache.flink.runtime.jobgraph.JobGraph;
 import org.apache.flink.runtime.jobgraph.JobVertex;
 import org.apache.flink.runtime.jobgraph.ScheduleMode;
 import org.apache.flink.runtime.jobgraph.tasks.AbstractInvokable;
 import org.apache.flink.runtime.jobgraph.tasks.CheckpointCoordinatorConfiguration;
+import org.apache.flink.runtime.messages.checkpoint.AcknowledgeCheckpoint;
 import org.apache.flink.runtime.state.memory.MemoryStateBackend;
 import org.apache.flink.runtime.taskmanager.TaskExecutionState;
 import org.apache.flink.util.TestLogger;
@@ -54,6 +57,7 @@ import static org.hamcrest.Matchers.empty;
 import static org.hamcrest.Matchers.equalTo;
 import static org.hamcrest.Matchers.is;
 import static org.junit.Assert.assertThat;
+import static org.mockito.Mockito.mock;
 
 /**
  * Tests for {@link AdaptedRestartPipelinedRegionStrategyNG}.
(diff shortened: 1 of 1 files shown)


Myasuka:
@flinkbot attention @zentol

tillrohrmann:
Why do you think that we no longer correctly test that pending checkpoints are being aborted with FLINK-13060 @Myasuka?

Myasuka:
@tillrohrmann I would try to interpret this more clearly below:
After [FLINK-13060](https://issues.apache.org/jira/browse/FLINK-13060), `AdaptedRestartPipelinedRegionStrategyNGAbortPendingCheckpointsTest` still use previous `InfiniteDelayRestartStrategy` which means `AdaptedRestartPipelinedRegionStrategyNG#resetTasks()` would never be called when some task failed. The root cause why previous `AdaptedRestartPipelinedRegionStrategyNGAbortPendingCheckpointsTest` could still pass is because of the pending checkpoints are aborted by `failUnacknowledgedPendingCheckpointsFor()` method in `CheckpointCoordinator` (which is actually introduced vai [FLINK-7844](https://issues.apache.org/jira/browse/FLINK-7844) ). 

In a nutshell, we expect pending checkpoints are aborted by `AdaptedRestartPipelinedRegionStrategyNG#resetTasks()` not `CheckpointCoordinator#failUnacknowledgedPendingCheckpointsFor()`.

zentol:
Why are you adding a second vertex?

Myasuka:
@zentol If only one vertex, after that vertex sent acknowledge message to checkpoint coordinator and that checkpoint would be treated as completed checkpoint. Actually I just want that checkpoint still stay as pending checkpoint.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/AdaptedRestartPipelinedRegionStrategyNGAbortPendingCheckpointsTest.java
```diff
@@ -76,49 +80,64 @@ public void abortPendingCheckpointsWhenRestartingTasks() throws Exception {
...
 		final int pendingCheckpointsBeforeFailure = checkpointCoordinator.getNumberOfPendingCheckpoints();
+		long checkpointId = checkpointCoordinator.getPendingCheckpoints().keySet().iterator().next();
 
-		failVertex(onlyExecutionVertex);
+		AcknowledgeCheckpoint acknowledgeCheckpoint = new AcknowledgeCheckpoint(
+			firstExecutionVertex.getJobId(),
+			firstExecutionVertex.getCurrentExecutionAttempt().getAttemptId(),
+			checkpointId,
+			mock(CheckpointMetrics.class),
+			mock(TaskStateSnapshot.class));
```
zentol: mocking isn't necessary, we can easily create instances for both of these.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/AdaptedRestartPipelinedRegionStrategyNGAbortPendingCheckpointsTest.java
```diff
@@ -76,49 +80,64 @@ public void abortPendingCheckpointsWhenRestartingTasks() throws Exception {
...
+
+		// not let the first checkpoint to be discarded when firstExecutionVertex has been acknowledged.
+		checkpointCoordinator.receiveAcknowledgeMessage(acknowledgeCheckpoint, "Unknown location");
```
zentol: add assertion / state check that after this we still have a pending checkpoint

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/AdaptedRestartPipelinedRegionStrategyNGAbortPendingCheckpointsTest.java
```diff
@@ -76,49 +80,64 @@ public void abortPendingCheckpointsWhenRestartingTasks() throws Exception {
...
 
-		setTaskRunning(executionGraph, onlyExecutionVertex);
+		setTaskRunning(executionGraph, firstExecutionVertex, vertexIterator.next());
 
 		final CheckpointCoordinator checkpointCoordinator = executionGraph.getCheckpointCoordinator();
 		checkState(checkpointCoordinator != null);
 
 		checkpointCoordinator.triggerCheckpoint(System.currentTimeMillis(),  false);
 		final int pendingCheckpointsBeforeFailure = checkpointCoordinator.getNumberOfPendingCheckpoints();
+		long checkpointId = checkpointCoordinator.getPendingCheckpoints().keySet().iterator().next();
```
zentol: add `checkState(pendingCheckpointsBeforeFailure == 1)`, remove assertion at the end

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/AdaptedRestartPipelinedRegionStrategyNGAbortPendingCheckpointsTest.java
```diff
@@ -76,49 +80,64 @@ public void abortPendingCheckpointsWhenRestartingTasks() throws Exception {
...
 
-		failVertex(onlyExecutionVertex);
+		AcknowledgeCheckpoint acknowledgeCheckpoint = new AcknowledgeCheckpoint(
+			firstExecutionVertex.getJobId(),
```
zentol: retrieve the job ID from the execution graph instead

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/AdaptedRestartPipelinedRegionStrategyNGAbortPendingCheckpointsTest.java
```diff
@@ -76,49 +80,64 @@ public void abortPendingCheckpointsWhenRestartingTasks() throws Exception {
...
+			firstExecutionVertex.getCurrentExecutionAttempt().getAttemptId(),
+			checkpointId,
+			mock(CheckpointMetrics.class),
+			mock(TaskStateSnapshot.class));
+
+		// not let the first checkpoint to be discarded when firstExecutionVertex has been acknowledged.
```
zentol:
```suggestion
		// let the first vertex acknowledge the checkpoint, and fail it afterwards
		// the failover strategy should then cancel all pending checkpoints on restart
```

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/AdaptedRestartPipelinedRegionStrategyNGAbortPendingCheckpointsTest.java
```diff
@@ -76,49 +80,64 @@ public void abortPendingCheckpointsWhenRestartingTasks() throws Exception {
...
+				new TaskExecutionState(executionGraph.getJobID(),
+					executionVertex.getCurrentExecutionAttempt().getAttemptId(),
+					ExecutionState.RUNNING));
+		}
 	}
 
 	private void failVertex(final ExecutionVertex onlyExecutionVertex) {
 		onlyExecutionVertex.getCurrentExecutionAttempt().fail(new Exception("Test Exception"));
 		manualMainThreadExecutor.triggerAll();
+		manualMainThreadExecutor.triggerScheduledTasks();
```
zentol: We could do this later after an assertion that we still have a pending checkpoint; this would give use more confidence that the actual restart by the failover strategy discarded the pending checkpoints

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/AdaptedRestartPipelinedRegionStrategyNGAbortPendingCheckpointsTest.java
```diff
@@ -76,49 +80,64 @@ public void abortPendingCheckpointsWhenRestartingTasks() throws Exception {
...
 		assertThat(pendingCheckpointsBeforeFailure, is(equalTo(1)));
 		assertNoPendingCheckpoints(checkpointCoordinator);
 	}
 
-	private void setTaskRunning(final ExecutionGraph executionGraph, final ExecutionVertex executionVertex) {
-		executionGraph.updateState(
-			new TaskExecutionState(executionGraph.getJobID(),
-				executionVertex.getCurrentExecutionAttempt().getAttemptId(),
-				ExecutionState.RUNNING));
+	private void setTaskRunning(final ExecutionGraph executionGraph, final ExecutionVertex... executionVertices) {
```
zentol: `setTasksRunning`

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/AdaptedRestartPipelinedRegionStrategyNGAbortPendingCheckpointsTest.java
zentol: Where did you resolve this? My goal was that we verified that a pending checkpoint still exists after `ExecutionAttempt#fail`, but before the actual restart by the failover strategy.

Myasuka: fixed in latest commit.

PR #9126, opened 2019-07-16
[FLINK-13407][tests][coordination] Get SlotManager.failUnfulfillableRequest in main thread of StandaloneResourceManager for verification

azagrebin:

In `StandaloneResourceManagerTest` we check `isFailingUnfulfillableRequest` in the test thread, although the underlying field `SlotManager.failUnfulfillableRequest` has no synchronisation and set in the main thread of `StandaloneResourceManager`. This can lead to a visibility problem of `SlotManager.failUnfulfillableRequest`: set in the main thread of `StandaloneResourceManager` but never updated in the test thread.

## Brief change log

The idea for the fix is read `SlotManager.failUnfulfillableRequest` in the main thread of `StandaloneResourceManager`, get future result and check in the `StandaloneResourceManagerTest.assertHappensUntil`.

## Verifying this change

run `StandaloneResourceManagerTest`

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/resourcemanager/StandaloneResourceManager.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/resourcemanager/StandaloneResourceManager.java
@@ -18,6 +18,7 @@
 
 package org.apache.flink.runtime.resourcemanager;
 
+import org.apache.flink.annotation.VisibleForTesting;
 import org.apache.flink.api.common.time.Time;
 import org.apache.flink.runtime.clusterframework.ApplicationStatus;
 import org.apache.flink.runtime.clusterframework.types.ResourceID;
@@ -37,6 +38,7 @@ import javax.annotation.Nullable;
 
 import java.util.Collection;
 import java.util.Collections;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.TimeUnit;
 
 /**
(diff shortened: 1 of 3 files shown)


azagrebin:
cc @tillrohrmann

flink-runtime/src/test/java/org/apache/flink/runtime/resourcemanager/StandaloneResourceManagerTest.java
```diff
@@ -105,10 +100,22 @@ private StandaloneResourceManager createResourceManager(Time startupPeriod) thro
...
 		rmServices.grantLeadership();
 
-		return new Tuple2<>(rm, rmServices.slotManager);
+		return rm;
+	}
+
+	private static boolean isFailingUnfulfillableRequest(
+			StandaloneResourceManager resourceManager) throws InterruptedException {
+		try {
+			return resourceManager.isFailingUnfulfillableRequestAsync().get();
```
tillrohrmann: would be simpler to use `CompletableFuture#join` here.

flink-runtime/src/main/java/org/apache/flink/runtime/resourcemanager/StandaloneResourceManager.java
```diff
@@ -92,6 +94,11 @@ protected void initialize() throws ResourceManagerException {
 		}
 	}
 
+	@VisibleForTesting
+	CompletableFuture<Boolean> isFailingUnfulfillableRequestAsync() {
+		return CompletableFuture.supplyAsync(this::isFailingUnfulfillableRequest, getMainThreadExecutor());
+	}
```
tillrohrmann: I'd prefer to have these kind of methods in a `TestingStandaloneResourceManager` because then they don't clutter the production code.

azagrebin:
Thanks for the review @tillrohrmann I have addressed the comments

I think we can merge this PR separately. I factored out in a separate [JIRA task](https://issues.apache.org/jira/browse/FLINK-13407) as a one step to fix the original [test instability FLINK-13242](https://issues.apache.org/jira/browse/FLINK-13242). Could you assign the task to me?

We can address the latest finding in a separate [JIRA task](https://issues.apache.org/jira/browse/FLINK-13408). Could you assign the task to @xintongsong?

PR #9128, opened 2019-07-16
[FLINK-13256] Ensure periodical checkpoint still valid when region failover abort pending checkpoints

Myasuka:
##  What is the purpose of the change

After we introduce region failover, pending checkpoints would be aborted when task failed. However, previous implementation would come across a bug leading to periodical checkpoint is invalid.
Previously, when checkpoint coordinator trigger a checkpoint and the size of pending checkpoints is larger than `maxConcurrentCheckpointAttempts` (default 1), it would just assign the `currentPeriodicTrigger` to null. However, the pending checkpoints would then possibly be aborted due to region failover, which lead to no pending checkpoints and no `currentPeriodicTrigger`. In the end, the periodical checkpoint mechanism would become not valid anymore.

## Brief change log

  - Ensure when `CheckpointCoordinator` abort pending checkpoints, the `currentPeriodicTrigger` would must not be null.
  - Add test `FailoverStrategyCheckpointCoordinatorTest` to verify this case.

## Verifying this change

  - Added test `FailoverStrategyCheckpointCoordinatorTest` to verify `currentPeriodicTrigger` is not null when `checkpointCoordinator#abortPendingCheckpointsWithTriggerValidation` is called.

  - The serializers: **no**

## Documentation

### DIFF
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/checkpoint/CheckpointCoordinator.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/checkpoint/CheckpointCoordinator.java
@@ -478,29 +478,9 @@ public class CheckpointCoordinator {
 					throw new CheckpointException(CheckpointFailureReason.ALREADY_QUEUED);
 				}
 
-				// if too many checkpoints are currently in progress, we need to mark that a request is queued
-				if (pendingCheckpoints.size() >= maxConcurrentCheckpointAttempts) {
-					triggerRequestQueued = true;
-					if (currentPeriodicTrigger != null) {
-						currentPeriodicTrigger.cancel(false);
-						currentPeriodicTrigger = null;
-					}
-					throw new CheckpointException(CheckpointFailureReason.TOO_MANY_CONCURRENT_CHECKPOINTS);
-				}
+				checkConcurrentCheckpoints();
 
-				// make sure the minimum interval between checkpoints has passed
-				final long durationTillNextMillis = getDurationTillNextMillis();
-
-				if (durationTillNextMillis > 0) {
-					if (currentPeriodicTrigger != null) {
-						currentPeriodicTrigger.cancel(false);
-						currentPeriodicTrigger = null;
-					}
-					// Reassign the new trigger to the currentPeriodicTrigger
-					currentPeriodicTrigger = scheduleTriggerWithDelay(durationTillNextMillis);
-
-					throw new CheckpointException(CheckpointFailureReason.MINIMUM_TIME_BETWEEN_CHECKPOINTS);
-				}
+				checkMinPauseBetweenCheckpoints();
 			}
 		}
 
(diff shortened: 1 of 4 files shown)


Myasuka:
@flinkbot attention @tillrohrmann @StefanRRichter

flink-runtime/src/main/java/org/apache/flink/runtime/checkpoint/CheckpointCoordinator.java
```diff
@@ -489,18 +489,15 @@ public PendingCheckpoint triggerCheckpoint(
 				}
 
 				// make sure the minimum interval between checkpoints has passed
-				final long earliestNext = lastCheckpointCompletionNanos + minPauseBetweenCheckpointsNanos;
-				final long durationTillNextMillis = (earliestNext - System.nanoTime()) / 1_000_000;
+				final long durationTillNextMillis = getDurationTillNextMillis();
```
carp84: This is good abstraction. Actually we could further abstract the whole logic of check against max concurrent checkpoint attempt and minimum checkpoint interval (L623-645 duplicates L482-503)

flink-runtime/src/main/java/org/apache/flink/runtime/checkpoint/CheckpointCoordinator.java
```diff
@@ -1270,11 +1266,30 @@ public void stopCheckpointScheduler() {
...
 	}
 
+	/**
+	 * Aborts all the pending checkpoints and check whether {@code currentPeriodicTrigger} has been assigned as null.
+	 * If so, we would create a new trigger as {@code currentPeriodicTrigger}.
+	 *
+	 * @param exception The exception.
+	 */
+	public void abortPendingCheckpointsWithTriggerValidation(CheckpointException exception) {
+		synchronized (lock) {
```
carp84: I could see we're synchronizing on `lock` to prevent starting multiple checkpoint schedulers not only here but also in the `start/stopCheckpointScheduler` methods. IMHO this interferes with the comment on `lock` which indicates it to only safeguard checkpoint updates, or say manipulation of `pendingCheckpoints`. It's kind of dangerous and using a separate lock for scheduler may be more reasonable, but it's not necessary to be improved in this PR.

Myasuka:
@carp84 Thanks for your review, I would address your comments and stabilize the test case.

StefanRRichter:
@Myasuka thanks for the fix, I think it solves the problem. One thing to think about is maybe to connect the concern of ensuring proper scheduling of the triggers closer to `failPendingCheckpoint()`. My reason is that failing a pending checkpoint can be the root cause of the problem that the trigger is not reestablished. But the symptom has currently to be fixed in multiple places that use this method, e.g. aborting checkpoints but also discarding checkpoints. This way of fixing would open the door for future problems, in case that more functionality is introduced that uses `failPendingCheckpoint()` but forgets about ensuring that triggers continue. Wdyt?

Myasuka:
@StefanRRichter I think your concern should be taken into account. After thinking about this issue, I prefer to place the watch dog in the `failPendingCheckpoint()`. If `periodicScheduling` is still `true` and no more pending checkpoints, we would then create a new trigger with random initial delay. If so, I could replace `abortPendingCheckpointsWithTriggerValidation()` in this PR back to previous `abortPendingCheckpoints()`.

Myasuka:
Unfortunately, be848eb failed due to a known issue [FLINK-13299](https://issues.apache.org/jira/browse/FLINK-13299) which has just been fixed.

StefanRRichter:
Yes, I checked it on your private Travis and strongly assumed the errors are unrelated.

PR #9131, opened 2019-07-16
[FLINK-12858][checkpointing] Stop-with-savepoint, workaround: fail whole job when savepoint is declined by a task

1u0:

This pull request is an attempt to address hanging Flink job when stop-with-savepoint fails due to decline of the savepoint by job's task. In such cases, the job manager would fail the whole execution graph (which may trigger a job restart).

## Brief change log

  - The `LegacyScheduler` is modified to track `CheckpointException`s in `stopWithSavepoint()` that originate from tasks and fails the execution graph for such exceptions.

## Verifying this change

This change was partially tested by manual e2e test:
 * configured Flink cluster with savepoints/checkpoints setup and `task.checkpoint.alignment.max-size: 64` set;
 * a test Flink job with two congested sources (joined in `map-1`) and events that exceed the configured limit (see execution graph below).

In the test job, the `map-1` with high probability rejects checkpoint/savepoints due to `checkpointSizeLimitExceeded`.

job-graph

## Documentation

### DIFF
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/checkpoint/CheckpointFailureReason.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/checkpoint/CheckpointFailureReason.java
@@ -23,60 +23,69 @@ package org.apache.flink.runtime.checkpoint;
  */
 public enum CheckpointFailureReason {
 
-	PERIODIC_SCHEDULER_SHUTDOWN("Periodic checkpoint scheduler is shut down."),
+	PERIODIC_SCHEDULER_SHUTDOWN(true, "Periodic checkpoint scheduler is shut down."),
 
-	ALREADY_QUEUED("Another checkpoint request has already been queued."),
+	ALREADY_QUEUED(true, "Another checkpoint request has already been queued."),
 
-	TOO_MANY_CONCURRENT_CHECKPOINTS("The maximum number of concurrent checkpoints is exceeded"),
+	TOO_MANY_CONCURRENT_CHECKPOINTS(true, "The maximum number of concurrent checkpoints is exceeded"),
 
-	MINIMUM_TIME_BETWEEN_CHECKPOINTS("The minimum time between checkpoints is still pending. " +
+	MINIMUM_TIME_BETWEEN_CHECKPOINTS(true, "The minimum time between checkpoints is still pending. " +
 			"Checkpoint will be triggered after the minimum time."),
 
-	NOT_ALL_REQUIRED_TASKS_RUNNING("Not all required tasks are currently running."),
+	NOT_ALL_REQUIRED_TASKS_RUNNING(true, "Not all required tasks are currently running."),
 
-	EXCEPTION("An Exception occurred while triggering the checkpoint."),
+	EXCEPTION(true, "An Exception occurred while triggering the checkpoint."),
 
-	CHECKPOINT_EXPIRED("Checkpoint expired before completing."),
+	CHECKPOINT_EXPIRED(false, "Checkpoint expired before completing."),
 
-	CHECKPOINT_SUBSUMED("Checkpoint has been subsumed."),
+	CHECKPOINT_SUBSUMED(false, "Checkpoint has been subsumed."),
 
-	CHECKPOINT_DECLINED("Checkpoint was declined."),
+	CHECKPOINT_DECLINED(false, "Checkpoint was declined."),
 
-	CHECKPOINT_DECLINED_TASK_NOT_READY("Checkpoint was declined (tasks not ready)"),
+	CHECKPOINT_DECLINED_TASK_NOT_READY(false, "Checkpoint was declined (tasks not ready)"),
 
-	CHECKPOINT_DECLINED_TASK_NOT_CHECKPOINTING("Task does not support checkpointing"),
+	CHECKPOINT_DECLINED_TASK_NOT_CHECKPOINTING(false, "Task does not support checkpointing"),
 
-	CHECKPOINT_DECLINED_SUBSUMED("Checkpoint was canceled because a barrier from newer checkpoint was received."),
+	CHECKPOINT_DECLINED_SUBSUMED(false, "Checkpoint was canceled because a barrier from newer checkpoint was received."),
 
-	CHECKPOINT_DECLINED_ON_CANCELLATION_BARRIER("Task received cancellation from one of its inputs"),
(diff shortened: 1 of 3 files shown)


1u0:
cc @StefanRRichter, @kl0u, @aljoscha.

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/LegacyScheduler.java
```diff
@@ -649,4 +659,22 @@ private String retrieveTaskManagerLocation(ExecutionAttemptID executionAttemptID
...
+	private static boolean isCheckpointDeclinedException(Throwable throwable) {
+		return ExceptionUtils.findThrowable(throwable, CheckpointException.class)
+			.map(CheckpointException::getCheckpointFailureReason)
+			.map(reason -> {
+				switch (reason) {
+					case CHECKPOINT_DECLINED:
+					case CHECKPOINT_DECLINED_TASK_NOT_READY:
+					case CHECKPOINT_DECLINED_SUBSUMED:
+					case CHECKPOINT_DECLINED_ALIGNMENT_LIMIT_EXCEEDED:
+					case CHECKPOINT_DECLINED_INPUT_END_OF_STREAM:
```
1u0: **NB:** this check is very rough, it may be too pessimistic in a way, that some causes not necessary leave the job in half-locked state (`CHECKPOINT_DECLINED, CHECKPOINT_DECLINED_TASK_NOT_READY`).

 * `CHECKPOINT_DECLINED_ALIGNMENT_LIMIT_EXCEEDED` case is the one I can reproduce;
 * `CHECKPOINT_DECLINED_INPUT_END_OF_STREAM` is also a potential issue, but may be not easy to reproduce (should happen on checkpoints alignment in a join when one branch has passed checkpoint and the second one has just ended);
 * `CHECKPOINT_DECLINED_SUBSUMED` - should not happen, but left just to be more future proof;
 * `TASK_CHECKPOINT_FAILURE` - I'm not sure if this one should also be present here.

Also, open question, what to do if exception is not a `CheckpointException`. We expect that such causes would fail the task that originated the exception, but I'm not sure how it would interfere with region recovery.

1u0: **Update:** the check was inverted to be more pessimistic about savepoint failure reasons.

flink-runtime/src/main/java/org/apache/flink/runtime/scheduler/LegacyScheduler.java
```diff
@@ -649,4 +658,24 @@ private String retrieveTaskManagerLocation(ExecutionAttemptID executionAttemptID
 			.map(TaskManagerLocation::toString)
 			.orElse("Unknown location");
 	}
+
+	private static boolean isPotentiallyTaskLocalException(Throwable throwable) {
```
kl0u: This method is pretty fragile. As soon as someone adds another exception, it may break. So I would suggest to add a flag in the `CheckpointFailureReason`, something like `final boolean preFlight`, that is set in the constructor and indicates if the exception at hand was thrown before the checkpoint barrier was emitted to the job (pre-flight), or not. This way, every new exception will have to set the flag and we can have in this method simply something like `isPreFlight()` which will check at which stage the exception was thrown.

flink-runtime/src/main/java/org/apache/flink/runtime/checkpoint/CheckpointFailureReason.java
```diff
@@ -79,4 +79,38 @@
 	public String message() {
 		return message;
 	}
+
+	/**
+	 * @return true if this value indicates a failure reason happening before a checkpoint is passed to a job's tasks.
+	 */
+	public boolean isPreFlight() {
```
kl0u: @1u0 Although better than before, I think that this method is also fragile as it expects that anybody who adds a new member here, will have to update the method accordingly. As I said in the previous comment, I find the solution of a `final boolean` flag which is set in the constructor more future-proof.

1u0: Imo, it is the same in regard of being future-proof. Currently I'm checking at runtime (by a test) that this values are defined for all enum's values.

If someone adds a new enum value, they would need to think and define if it's `isPreFlight` or not in both approaches.

I personally find the approach with separate method less intrusive and more friendlier when you need to delete such attribute. It's also more friendlier (densely located), when you want to see the attribute's values by enum's value.

Please let me know if this is a blocker from your point of view, I'd change it to constructor/field based approach.

kl0u: I still believe that the flag is a better approach. Imagine if in the future we want to add another attribute. Then we will have to add another method that enumerates all the exceptions and another test that checks the completeness of the method.

flink-runtime/src/main/java/org/apache/flink/runtime/checkpoint/CheckpointFailureReason.java
```diff
@@ -23,60 +23,69 @@
...
+	CHECKPOINT_DECLINED_ON_CANCELLATION_BARRIER(false, "Task received cancellation from one of its inputs"),
 
-	CHECKPOINT_DECLINED_ALIGNMENT_LIMIT_EXCEEDED("The checkpoint alignment phase needed to buffer more than the configured maximum bytes"),
+	CHECKPOINT_DECLINED_ALIGNMENT_LIMIT_EXCEEDED(false, "The checkpoint alignment phase needed to buffer more than the configured maximum bytes"),
 
-	CHECKPOINT_DECLINED_INPUT_END_OF_STREAM("Checkpoint was declined because one input stream is finished"),
+	CHECKPOINT_DECLINED_INPUT_END_OF_STREAM(false, "Checkpoint was declined because one input stream is finished"),
 
-	CHECKPOINT_COORDINATOR_SHUTDOWN("CheckpointCoordinator shutdown."),
+	CHECKPOINT_COORDINATOR_SHUTDOWN(true, "CheckpointCoordinator shutdown."),
```
Myasuka: If `preFlight` indicates to whether checkpoint failed before sending to tasks, why `CheckpointCoordinator shutdown` would be treated as preFlight?

kl0u: You are right @Myasuka ! I address it in the https://github.com/apache/flink/pull/9240 which is a follow-up of this PR.

PR #9135, opened 2019-07-16
[FLINK-13296][table] FunctionCatalog.lookupFunction() should check in memory functions if the target function doesn't exist in catalog

bowenli86:

Currently the logic to lookup a function is check either the catalog or the in memory function. But the correct logic is to 1st check the catalog, and if the function doesn't exist there, check in memory functions. There should be a resolution order.

This is root cause of FLINK-13294

## Brief change log

- change `FunctionCatalog.lookupFunction()` to 1st check the catalog, and if the function doesn't exist there, check in memory functions
- re-enable UT `LocalExecutorITCase.testUseCatalogAndUseDatabase`

## Verifying this change

This change is already covered by existing tests, such as *LocalExecutorITCase.testUseCatalogAndUseDatabase*.

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/FunctionCatalog.java
+++ b/flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/FunctionCatalog.java
@@ -162,53 +162,57 @@ public class FunctionCatalog implements FunctionLookup {
 		FunctionDefinition userCandidate = null;
 
 		Catalog catalog = catalogManager.getCatalog(catalogManager.getCurrentCatalog()).get();
+		ObjectPath functionPath = new ObjectPath(catalogManager.getCurrentDatabase(), functionName);
 
-		if (catalog.getTableFactory().isPresent() &&
+		if (catalog.functionExists(functionPath)) {
+			if (catalog.getTableFactory().isPresent() &&
 				catalog.getTableFactory().get() instanceof FunctionDefinitionFactory) {
-			try {
-				CatalogFunction catalogFunction = catalog.getFunction(
-					new ObjectPath(catalogManager.getCurrentDatabase(), functionName));
+				try {
+					CatalogFunction catalogFunction = catalog.getFunction(functionPath);
 
-				FunctionDefinitionFactory factory = (FunctionDefinitionFactory) catalog.getTableFactory().get();
+					FunctionDefinitionFactory factory = (FunctionDefinitionFactory) catalog.getTableFactory().get();
 
-				userCandidate = factory.createFunctionDefinition(functionName, catalogFunction);
-			} catch (FunctionNotExistException e) {
-				// Ignore
-			}
+					userCandidate = factory.createFunctionDefinition(functionName, catalogFunction);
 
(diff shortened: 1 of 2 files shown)


bowenli86:

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/FunctionCatalog.java
```diff
@@ -162,53 +162,57 @@ public void registerScalarFunction(String name, ScalarFunction function) {
...
-			final Optional<FunctionDefinition> foundDefinition;
-			if (userCandidate != null) {
-				foundDefinition = Optional.of(userCandidate);
+					return Optional.of(
+						new FunctionLookup.Result(
+							ObjectIdentifier.of(catalogManager.getCurrentCatalog(), catalogManager.getCurrentDatabase(), name),
+							userCandidate)
+					);
+				} catch (FunctionNotExistException e) {
+					// Ignore
```
xuefuz: Instead of ignoring the exception, I'm wondering if we should throw a TableException. This is because the function exists previously but now stops existing, which implies an unexpected scenario.

bowenli86:
@xuefuz thanks for your review. I've update the PR, please take another look

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/FunctionCatalog.java
```diff
@@ -159,32 +159,31 @@ public void registerScalarFunction(String name, ScalarFunction function) {
...
+				return Optional.of(
+					new FunctionLookup.Result(
+						ObjectIdentifier.of(catalogManager.getCurrentCatalog(), catalogManager.getCurrentDatabase(), name),
+						userCandidate)
+				);
 			} else {
 				// TODO: should go thru function definition discover service
 			}
+		} catch (FunctionNotExistException e) {
+			// Ignore
```
xuefuz: We can probably update the comment to be more meaningful.

bowenli86:

w.r.t the param, I haven't seen where `FunctionDefinitionFactory.createFunctionDefinition()` will need the database name, but wondering if  I missed anything here. At this point, the CatalogFunction should already be retrieved with an ObjectPath. `FunctionDefinitionFactory.createFunctionDefinition()` doesn't talk to Catalog APIs, and the function name passed as 1st param is for the sole purpose of constructing a FunctionDefinition which requires a name. This is different from `TableFactory.createSource/Sink()`, which needs the database name to access partition information through Catalog APIs.

xuefuz:
I understand that database name is not used in HiveTableFactory.createFunctionDefinition(), but it's just the case presently and one implementation. It could be similarly used later or in other implementations. When that comes, we will have to introduce new API. On the other hand, if we add it now, it doesn't hurt if an implementation doesn't use that part. 

More importantly, it's more consistent with other APIs in the class.

Nevertheless, I'm not strongly insisting. It's your call.

bowenli86:
merging as the travis UT failures are unrelated

PR #9138, opened 2019-07-17
[FLINK-13249][runtime] Fix handling of partition producer responses b…

StefanRRichter:

This PR fixes the problem in FLINK-13249 by ensuring that processing the partition producer response is not blocking any netty thread, but is always executed by the task's executor.

  - The serializers: (no)
  - Anything that affects deployment or recovery: JobManager (and its components), Checkpointing, Yarn/Mesos, ZooKeeper: (kind of)

## Documentation

### DIFF
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/PartitionProducerStateProvider.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/PartitionProducerStateProvider.java
@@ -22,7 +22,7 @@ import org.apache.flink.runtime.execution.ExecutionState;
 import org.apache.flink.runtime.jobgraph.IntermediateDataSetID;
 import org.apache.flink.types.Either;
 
-import java.util.concurrent.CompletableFuture;
+import java.util.function.Consumer;
 
 /**
  * Request execution state of partition producer, the response accepts state check callbacks.
@@ -34,11 +34,12 @@ public interface PartitionProducerStateProvider {
 	 * @param intermediateDataSetId ID of the parent intermediate data set.
 	 * @param resultPartitionId ID of the result partition to check. This
 	 * identifies the producing execution and partition.
-	 * @return a future with response handle.
+	 * @param responseConsumer consumer for the response handle.
 	 */
-	CompletableFuture<? extends ResponseHandle> requestPartitionProducerState(
+	void requestPartitionProducerState(
 		IntermediateDataSetID intermediateDataSetId,
-		ResultPartitionID resultPartitionId);
+		ResultPartitionID resultPartitionId,
+		Consumer<? super ResponseHandle> responseConsumer);
 
 	/**
 	 * Result of state query, accepts state check callbacks.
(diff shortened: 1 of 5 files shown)


flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/consumer/SingleInputGate.java
```diff
@@ -601,8 +601,7 @@ void notifyChannelNonEmpty(InputChannel channel) {
 	void triggerPartitionStateCheck(ResultPartitionID partitionId) {
 		partitionProducerStateProvider.requestPartitionProducerState(
 			consumedResultId,
-			partitionId)
-			.thenAccept(responseHandle -> {
+			partitionId, (responseHandle -> {
```
pnowojski: nit: formatting, `(responseHandle ...)` in new line.

Maybe it would be better to extract lambda function to a named method?

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/PartitionProducerStateProvider.java
```diff
@@ -22,7 +22,7 @@
 import org.apache.flink.runtime.jobgraph.IntermediateDataSetID;
```
pnowojski: Can you expand commit message to include at least the same content as in the PR description?

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/consumer/SingleInputGate.java
zhijiangW: `(responseHandle -> {` in a separate line?

zhijiangW:
I opened this page long time no refresh, and just saw @pnowojski 's review after posting my review. Some common concerns are already proposed by him. :)

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/Task.java
```diff
@@ -1080,18 +1080,22 @@ else if (current == ExecutionState.RUNNING) {
...
-		final CompletableFuture<PartitionProducerStateResponseHandle> result =
-			futurePartitionState.handleAsync(PartitionProducerStateResponseHandle::new, executor);
-		FutureUtils.assertNoException(result);
-		return result;
+
+		futurePartitionState.whenCompleteAsync(
+			(ExecutionState executionState, Throwable throwable) -> {
+				responseConsumer.accept(new PartitionProducerStateResponseHandle(executionState, throwable));
+			},
+			executor);
```
tillrohrmann: nit: one could also write it

```
			futurePartitionState
				.handle(PartitionProducerStateResponseHandle::new)
				.thenAcceptAsync(responseConsumer, executor));
```

Which also ensures that any exception coming out of the consumer will make the process to fail.

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/consumer/SingleInputGate.java
StefanRRichter: Agree with formatting, extraction into a named method is probably not a good idea because the lambda is capturing `partitionId` from the enclosing scope.

StefanRRichter:
Thanks for the reviews @tillrohrmann @pnowojski @zhijiangW ! I address all your comments, and I think the current consensus with Till and Chesnay is to merge the fix without a test now because it comes up every now and then as unrelated problem in release testing. I would then directly start working on a test, but my feeling is that even as an I style test this is not going to be easy. Let's see...

PR #9141, opened 2019-07-17
[FLINK-12249][table] Fix type equivalence check problems for Window Aggregates

hequn8128:

This pull request fixes type equivalence check problems for Window Aggregates

### background

Creating Aggregate node fails in rules: LogicalWindowAggregateRule and ExtendedAggregateExtractProjectRule if the only grouping expression is a window and
we compute aggregation on NON NULLABLE field.

The root cause for that, is how return type inference strategies in calcite work and how we handle window aggregates. Take org.apache.calcite.sql.type.ReturnTypes#AGG_SUM as an example, based on groupCount it adjusts type nullability based on groupCount.

### analysis
There is no problem for non-window aggregate that return type turn to nullable if the call occurs within a "GROUP BY ()" query, e.g. in "select sum(x) as s from empty", s may be null. 
This is also the behavior of common databases, such as mysql. The result of query "select sum(x) as s from empty" is a null output.

However, for window aggregate, there is no way to output null, as we don't know which window the null belongs to, so the result should be an empty result which means the result type is also not null.

### solution
To fix this problem we can replace the current sum/avg SqlAggFunction to a self-defined one which avoids turning the "not null" to "nullable" for window aggregate. We can simply use a Rule to archive this.

## Brief change log

  - Add a self-defined sum and avg SqlAggFunction which will not change return type to nullable even if the call occurs within a "GROUP BY ()" for window aggregates.
  - Add a Rule(`ReplaceWindowAggregateFunctionRule`) to change the sum/avg to the self-defined sum/avg(`SqlWindowAvgAggFunction` and `SqlWindowSumAggFunction`).
  - Add `getSqlAggFunctionString` in `CommonAggregate`. This method makes sure that the new self-defined sum/avg returns same string with sum/agg.
  - Add tests.

## Verifying this change

  - Added plan tests for flink, blink.
  - Added plan tests for batch, streaming.

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/plan/rules/logical/LogicalWindowAggregateRuleBase.scala
+++ b/flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/plan/rules/logical/LogicalWindowAggregateRuleBase.scala
@@ -33,8 +33,10 @@ import org.apache.calcite.plan._
 import org.apache.calcite.plan.hep.HepRelVertex
 import org.apache.calcite.rel.`type`.RelDataType
 import org.apache.calcite.rel.core.Aggregate.Group
+import org.apache.calcite.rel.core.{Aggregate, AggregateCall}
 import org.apache.calcite.rel.logical.{LogicalAggregate, LogicalProject}
 import org.apache.calcite.rex._
+import org.apache.calcite.sql.`type`.SqlTypeUtil
 import org.apache.calcite.util.ImmutableBitSet
 
 import _root_.java.math.BigDecimal
@@ -84,13 +86,20 @@ abstract class LogicalWindowAggregateRuleBase(description: String)
       .project(project.getChildExps.updated(windowExprIdx, inAggGroupExpression))
       .build()
 
+    // Currently, this rule removes the window group by field which may lead to the changes of
+    // AggCall's type which brings fails on type checks.
+    // To solve the problem, we can change the types to the inferred types in the Aggregate and then
+    // cast back in the project after Aggregate.
+    val indexAndTypes = getIndexAndInferredTypesForInvalidAggCalls(agg)
+    val finalCalls = replaceAggCalls(indexAndTypes, agg)
+
     // we don't use the builder here because it uses RelMetadataQuery which affects the plan
     val newAgg = LogicalAggregate.create(
       newProject,
       agg.indicator,
       newGroupSet,
       ImmutableList.of(newGroupSet),
-      agg.getAggCallList)
+      finalCalls)
 
     // create an additional project to conform with types
     val outAggGroupExpression0 = getOutAggregateGroupExpression(rexBuilder, windowExpr)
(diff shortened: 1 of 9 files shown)


hequn8128:
Hi @twalthr @dawidwys , would be great if you can take a look at the fix. Thank you.

sunjincheng121:
Thanks for the PR @hequn8128 
Please rebase the PR, let the CI passed.

@flinkbot attention @dawidwys @twalthr :)

Best, Jincheng

hequn8128:
@sunjincheng121 Have rebased to the latest master. Thanks a lot for the advice.

dawidwys:
Thank you for your work @hequn8128. Unfortunately I don't think this is a valid solution. It does not fix the root cause of the problem (that we deceive the logic of Aggregate with a false information about number of grouping fields. In case of Window aggregates it is always >= 1, because we group at least by window).

This just workarounds the problem for subset of aggregates. It does not fix it for:
* SqlAnyValueAggFunction
* SqlBitOpAggFunction
* SqlFirstLastValueAggFunction
* SqlMinMaxAggFunction
* SqlNthValueAggFunction
and any other function that uses the information of `org.apache.calcite.rel.core.Aggregate.AggCallBinding#getGroupCount` for type inference.

hequn8128:
@dawidwys Thanks for your comments. I think you are right. 

We had a discussion just now. (with @godfreyhe @wuchong @sunjincheng121 ) We think that, to solve the problem fundamentally, the current window aggregate should not extend from `org.apache.calcite.rel.core.Aggregate`, instead it should extend from `SingleRel`. This makes sure the right semantics of window aggregate.

To correct this, the changes may be somehow very large as there are many existing window related rules. Also, this is a problem a long time ago and the probability of occurrence is relatively small. 

Considering the reasons above, I think it would be good if we can fix this issue later after release-1.9? 

What do you think? @dawidwys 

Best, Hequn

dawidwys:
I agree we should not treat this issue as a blocker for 1.9. It's rather uncommon as there are a few prerequisites for it to occur (only group by window, not null field for aggregate). I have one idea how could it be fixed I want to check, but overall we could change the priority to minor.

godfreyhe:
> @dawidwys Thanks for your comments. I think you are right.
> 
> We had a discussion just now. (with @godfreyhe @wuchong @sunjincheng121 ) We think that, to solve the problem fundamentally, the current window aggregate should not extend from `org.apache.calcite.rel.core.Aggregate`, instead it should extend from `SingleRel`. This makes sure the right semantics of window aggregate.
> 
> To correct this, the changes may be somehow very large as there are many existing window related rules. Also, this is a problem a long time ago and the probability of occurrence is relatively small.
> 
> Considering the reasons above, I think it would be good if we can fix this issue later after release-1.9?
> 
> What do you think? @dawidwys @godfreyhe @wuchong @sunjincheng121
> 
> Best, Hequn

yes, it's a big issue to refactor WindowAggregate completely using solution3 mentioned in [FLINK-12249](https://issues.apache.org/jira/browse/FLINK-12249). "Creating special window aggregate call function" is not a proper approach, how about users defined aggregate function? currently, i think we should make WindowAggregate does not extend from Aggregate in this pr, which had been validated in  a blink minor branch. and do the clean refactor after 1.9 ?

hequn8128:
@dawidwys @godfreyhe @wuchong @sunjincheng121 Hi, I have updated the PR with the latest solution: Change the types to the inferred types in the Aggregate and then cast back in the Project after Aggregate. 

I think this makes sure the right semantics and also brings minor changes to the under-released Flink.  The long term solution as we discussed in the jira ticket should be considered later after 1.9. 

Would be great if you can take a look.

Best, Hequn

sunjincheng121:
Thanks for the update @hequn8128!
I agree with @dawidwys @hequn8128  that this issue is not the blocker of 1.9 release.

And I like this workaround patch, in this way we can solve this issue for user in 1.9, and create a new JIRA for  the follow-up discussion.

What do you think?

Best, Jincheng

hequn8128:
@sunjincheng121 Hi, thanks a lot for your suggestion. I have created a new jira([FLINK-13392](https://issues.apache.org/jira/browse/FLINK-13392)) to further improve the WindowAggregate. 

Best, Hequn

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/plan/rules/logical/LogicalWindowAggregateRuleBase.scala
```diff
@@ -111,12 +120,75 @@ abstract class LogicalWindowAggregateRuleBase(description: String)
     // The transformation adds an additional LogicalProject at the top to ensure
     // that the types are equivalent.
     transformed.push(windowAgg)
-      .project(transformed.fields().patch(windowExprIdx, Seq(outAggGroupExpression), 0))
+      .project(transformed
```
godfreyhe: could you split the logic to step by step ? lambda expression is hard to debug

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/plan/rules/logical/LogicalWindowAggregateRuleBase.scala
```diff
@@ -111,12 +120,75 @@ abstract class LogicalWindowAggregateRuleBase(description: String)
...
+        .fields()
+        .patch(windowExprIdx, Seq(outAggGroupExpression), 0)
+        .zipWithIndex.map { rexNodeAndIndex =>
```
godfreyhe: map {
case (rexNode, index) =>
....
}

more easier to read

hequn8128:
@godfreyhe Hi, thanks a lot for your review and suggestions. I have updated the PR according to your comments.

Best, Hequn

PR #9146, opened 2019-07-17
[FLINK-13284] [table-planner-blink] Correct some builtin functions' r…

lincoln-lil:

Correct several builtin temporal functions' return type inference in Blink planner which should cascade a `SqlTypeTransforms.FORCE_NULLABLE`, otherwise it may derive wrong nullable info for constant input arg(s) that will cause a incorrect expression reduction.
 

## Brief change log
  - *Update the type inference for `FlinkSqlOperatorTable`*
  - *Add more test cases*
  - *fix a NPE in SqlDateTimeUtils*

## Verifying this change

  - *Added cases for TemporalTypesTest*

  - The serializers: ( no )
  - Anything that affects deployment or recovery: JobManager (and its components), Checkpointing, Yarn/Mesos, ZooKeeper: ( no)
  - The S3 file system connector: ( no)

## Documentation

  - If yes, how is the feature documented? (not applicable )

### DIFF
--- a/flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/functions/sql/FlinkSqlOperatorTable.java
+++ b/flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/functions/sql/FlinkSqlOperatorTable.java
@@ -570,7 +570,7 @@ public class FlinkSqlOperatorTable extends ReflectiveSqlOperatorTable {
 	public static final SqlFunction DATE_FORMAT = new SqlFunction(
 		"DATE_FORMAT",
 		SqlKind.OTHER_FUNCTION,
-		ReturnTypes.cascade(ReturnTypes.explicit(SqlTypeName.VARCHAR), SqlTypeTransforms.TO_NULLABLE),
+		ReturnTypes.cascade(ReturnTypes.explicit(SqlTypeName.VARCHAR), SqlTypeTransforms.FORCE_NULLABLE),
 		InferTypes.RETURN_TYPE,
 		OperandTypes.or(
 			OperandTypes.sequence("'(TIMESTAMP, FORMAT)'",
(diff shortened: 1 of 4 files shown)


godfreyhe:
Thanks for this PR @lincoln-lil . for stream job, I think "return null" is the general case (maybe some case users also want "throw exception" when processing illegal data). while for batch job, "throw exception" is more general ?!  for the long term, maybe we should add a config to let users choose the mode. for this PR, it looks good to me.

JingsongLi:
> Thanks for this PR @lincoln-lil . for stream job, I think "return null" is the general case (maybe for some special cases, users also want "throw exception" when processing illegal data). while for batch job, "throw exception" is more general ?! for the long term, maybe we should add a config to let users choose the mode. for this PR, it looks good to me.

For batch join, "return null" is more popular too in hadoop sql engines.

lincoln-lil:
@godfreyhe @JingsongLi  Agree with you that we should offer a deterministic semantic for those 'dirty data', I think we can achieve this for two steps:
1. unify all the builtin functions' exception handling behavior for blink planner(since it differs with flink planner), I found two exception functions and will create another issue to fix it.
2. add a global configuration to support something like [MySQL's strict/non-strict sql mode](https://dev.mysql.com/doc/refman/5.6/en/sql-mode.html#sql-mode-strict) for exception handling includes numeric out-of-range and overflow and illegal inputs for sources. We can start a new thread to discuss it, what do you think?

godfreyhe:
> @godfreyhe @JingsongLi Agree with you that we should offer a deterministic semantic for those 'dirty data', I think we can achieve this for two steps:
> 
> 1. unify all the builtin functions' exception handling behavior for blink planner(since it differs with flink planner), I found two exception functions and will create another issue to fix it.
> 2. add a global configuration to support something like [MySQL's strict/non-strict sql mode](https://dev.mysql.com/doc/refman/5.6/en/sql-mode.html#sql-mode-strict) for exception handling includes numeric out-of-range and overflow and illegal inputs for sources. We can start a new thread to discuss it, what do you think?

sounds good, look forward the discussion

PR #9147, opened 2019-07-17
[FLINK-11631][test] Harden TaskExecutorITCase

ifndef-SleePy:

* The corrupt `TaskExecutor` might fail `MiniCluster` when shutting down, we should avoid this scenario
* Harden `TaskExecutorITCase#testJobReExecutionAfterTaskExecutorTermination`
* Harden `TastExecutorITCase#testJobRecoveryWithFailingTaskExecutor`

## Brief change log

* Improve `MiniCluster` by clearing terminated components. Use a map to keep task managers in `MiniCluster` instead of a list since the task managers might be mutable in some test cases
* `MiniCluster` exposes more components to `TestingMiniCluster` for enriching testing and Keep all testing codes in `TestingMiniCluster`
* Add a try-catch for `TastExecutorITCase#testJobRecoveryWithFailingTaskExecutor` to avoid unexpected exception caused by corrupt task executor

## Verifying this change

  - The serializers: no

## Documentation

### DIFF
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/minicluster/MiniCluster.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/minicluster/MiniCluster.java
@@ -93,7 +93,8 @@ import java.net.URI;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
-import java.util.List;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionException;
@@ -125,7 +126,7 @@ public class MiniCluster implements JobExecutorService, AutoCloseableAsync {
 	private final Time rpcTimeout;
 
 	@GuardedBy("lock")
-	private final List<TaskExecutor> taskManagers;
+	private final Map<ResourceID, TaskExecutor> taskManagers;
 
 	private final TerminatingFatalErrorHandlerFactory taskManagerTerminatingFatalErrorHandlerFactory = new TerminatingFatalErrorHandlerFactory();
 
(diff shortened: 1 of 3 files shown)


ifndef-SleePy:
As described in https://issues.apache.org/jira/browse/FLINK-11631, the unstable case is caused by https://issues.apache.org/jira/browse/FLINK-11630 directly.
However I still think we need to improve the `MiniCluster` and `TestingMiniCluster`.

ifndef-SleePy:
The Travis building failed due to some irrelevant cases. I have checked one of the failed case. It would be OK on latest master branch. So I will rebase master branch and try again.

ifndef-SleePy:
The Travis building fails again due to known unstable case, see https://issues.apache.org/jira/browse/FLINK-12164.

ifndef-SleePy:
Will close and reopen to trigger Travis building again

ifndef-SleePy:
Oops, there is a conflict with latest master branch. Will try to solve it later.

flink-runtime/src/main/java/org/apache/flink/runtime/minicluster/MiniCluster.java
```diff
@@ -230,6 +231,18 @@ public HighAvailabilityServices getHighAvailabilityServices() {
...
+	@Nonnull
+	protected Map<ResourceID, TaskExecutor> getTaskManagers() {
+		return taskManagers;
+	}
+
+	@VisibleForTesting
+	@Nonnull
+	protected Object getManipulationLock() {
+		return lock;
+	}
```
tillrohrmann: This we should definitely not do. This leaks the internal implementation details of this class to its sub classes.

ifndef-SleePy:
Hi @tillrohrmann , thanks for reviewing! Just a simple discussion below.

> ... it violates encapsulation of the MiniCluster.

Yes, it's by design. Actually the encapsulation of `MiniCluster` is already broken somehow, IMO. For example, there are several unnecessary `protected` methods exposed to `TestingMiniCluster`. The boundary of encapsulation here seems to be a bit fuzzy to me :(

> Moreover, I think the existing behaviour is good because otherwise we would not have noticed the problem with `Tasks` terminating too late.

I have to admit that it indeed found something although I'm not a fan of the behavior of keeping the corrupt `TaskExecutor` even we terminated it. I always think it might bring us some trouble someday.

Actually I think this PR is not a big deal because I believe https://github.com/apache/flink/pull/9072 could be merged soon. The unstable scenario should not exist anymore after merging that. 

Anyway, thanks again for the suggestion :)
This PR and the JIRA ticket can be closed.

PR #9088, opened 2019-07-11
[FLINK-13012][hive] Handle default partition name of Hive table

lirui-apache:
## What is the purpose of the change

To handle default partition name when reading/writing Hive tables.

## Brief change log

  - Handle default partition name.
  - Added a new test case for default partition name.

## Verifying this change

Comes with new test case.

## Does this pull request potentially affect one of the following parts:

  - Dependencies (does it add or upgrade a dependency): no
  - The public API, i.e., is any changed class annotated with `@Public(Evolving)`: no
  - The serializers: no
  - The runtime per-record code paths (performance sensitive): yes
  - Anything that affects deployment or recovery: JobManager (and its components), Checkpointing, Yarn/Mesos, ZooKeeper: no
  - The S3 file system connector: no

## Documentation

  - Does this pull request introduce a new feature? no
  - If yes, how is the feature documented? NA

PR #9149, opened 2019-07-17
[FLINK-13303][hive] Add e2e test for hive connector

lirui-apache:

To add e2e test for hive connector.

## Brief change log

  - Added a new module under `flink-end-to-end-tests` to run e2e test for hive connector.
  - A simple test case comes with the new module to make sure it works properly.
  - Added the e2e test to pre-commit tests.

## Verifying this change

New test case is added.

  - The serializers: no

## Documentation


lirui-apache:
@xuefuz @bowenli86 @zjuwangg 
I have split the e2e test into its own PR. Please have a look, thanks.

flink-end-to-end-tests/flink-connector-hive-test/pom.xml
```diff
@@ -0,0 +1,231 @@
...
+	<artifactId>flink-connector-hive-test</artifactId>
+	<name>flink-connector-hive-test</name>
+	<packaging>jar</packaging>
+
+	<dependencies>
+
+		<dependency>
+			<groupId>org.apache.flink</groupId>
+			<artifactId>flink-connector-hive_${scala.binary.version}</artifactId>
+			<version>${project.version}</version>
```
bowenli86: should its scope be 'test' since flink-connector-hive-test doesn't have any non-testing code?

flink-end-to-end-tests/flink-connector-hive-test/pom.xml
```diff
@@ -0,0 +1,231 @@
...
+		</dependency>
+
+		<dependency>
+			<groupId>org.apache.flink</groupId>
+			<artifactId>flink-sql-client_${scala.binary.version}</artifactId>
```
bowenli86: ditto

flink-end-to-end-tests/flink-connector-hive-test/pom.xml
```diff
@@ -0,0 +1,231 @@
...
+				<exclusion>
+					<groupId>com.google.guava</groupId>
+					<artifactId>guava</artifactId>
+				</exclusion>
+			</exclusions>
+		</dependency>
+
+		<dependency>
+			<groupId>org.apache.hive.hcatalog</groupId>
+			<artifactId>hive-hcatalog-core</artifactId>
```
bowenli86: isn't this duplicate with the above dependency?

flink-end-to-end-tests/test-scripts/test_connector_hive.sh
```diff
@@ -0,0 +1,22 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+################################################################################
+
+cd "${END_TO_END_DIR}/flink-connector-hive-test"
+
+mvn test -DskipTests=false
```
bowenli86: need a new line at EOF

flink-table/flink-sql-client/src/test/java/org/apache/flink/table/client/gateway/utils/LocalExecutorTestUtil.java
```diff
@@ -0,0 +1,58 @@
...
+
+	public static List<String> retrieveTableResult(
+			Executor executor,
+			SessionContext session,
+			String resultID) throws InterruptedException {
+
+		final List<String> actualResults = new ArrayList<>();
+		while (true) {
+			Thread.sleep(50); // slow the processing down
+			final TypedResult<Integer> result = executor.snapshotResult(session, resultID, 2);
```
bowenli86: why do we need to set `pageSize` as 2?

flink-table/flink-sql-client/src/test/java/org/apache/flink/table/client/gateway/utils/LocalExecutorTestUtil.java
```diff
@@ -0,0 +1,58 @@
...
+import org.apache.flink.types.Row;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.IntStream;
+
+/**
+ * Test util for LocalExecutor.
+ */
+public class LocalExecutorTestUtil {
```
xuefuz: I don't see this is used anywhere.

flink-connectors/flink-connector-hive/pom.xml
```diff
@@ -241,7 +241,6 @@ under the License.
 			<groupId>org.apache.hive</groupId>
 			<artifactId>hive-exec</artifactId>
 			<version>${hive.version}</version>
-			<scope>provided</scope>
```
xuefuz: What is the implication from removing this?

flink-end-to-end-tests/flink-connector-hive-test/pom.xml
lirui-apache: No, this is the test-jar of hive-hcatalog-core

flink-table/flink-sql-client/src/test/java/org/apache/flink/table/client/gateway/utils/LocalExecutorTestUtil.java
lirui-apache: Indeed. Will remove it

flink-table/flink-sql-client/src/test/java/org/apache/flink/table/client/gateway/utils/LocalExecutorTestUtil.java
lirui-apache: It's copied from LocalExecutorITCase. I think any small number will do.

flink-connectors/flink-connector-hive/pom.xml
lirui-apache: It makes hive-exec a transitive dependency. It won't be packed into hive connector because we no longer build a shaded jar.

twalthr: Please move changes that are not related to the actual issue of adding an e2e test to separate commits.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/client/HiveShimV1.java
```diff
@@ -87,6 +87,9 @@ public Function getFunction(IMetaStoreClient client, String dbName, String funct
 			if (e.getCause() instanceof NoSuchObjectException) {
 				throw (NoSuchObjectException) e.getCause();
 			}
+			if (e.getMessage().startsWith(NoSuchObjectException.class.getSimpleName())) {
```
twalthr: Separate commit.

flink-end-to-end-tests/flink-connector-hive-test/pom.xml
```diff
@@ -0,0 +1,331 @@
...
+					<artifactId>hive-serde</artifactId>
+				</exclusion>
+				<exclusion>
+					<groupId>org.apache.hadoop</groupId>
+					<artifactId>hadoop-yarn-server-resourcemanager</artifactId>
+				</exclusion>
+			</exclusions>
+		</dependency>
+
+		<dependency>
```
twalthr: End to end tests should use what a user would use. And users use the command line for the SQL Client or to submit a Flink job. Both in a bash script.

flink-end-to-end-tests/flink-connector-hive-test/src/test/java/org/apache/flink/batch/connectors/hive/LocalExecutorHiveConnectorTest.java
```diff
@@ -0,0 +1,150 @@
...
+	private static SessionContext session;
+	private static HiveMetastoreClientWrapper hmsClient;
+
+	private static final String HIVE_CATALOG_YAML_FILE = "test-sql-client-hive.yaml";
+
+	@HiveSQL(files = {})
+	private static HiveShell hiveShell;
+
+	@ClassRule
+	public static final MiniClusterWithClientResource MINI_CLUSTER_RESOURCE = new MiniClusterWithClientResource(
```
twalthr: Using a mini cluster is not a end-to-end test. Use want a user would use.

flink-end-to-end-tests/flink-connector-hive-test/src/test/java/org/apache/flink/batch/connectors/hive/LocalExecutorHiveConnectorTest.java
```diff
@@ -0,0 +1,150 @@
...
+		session = new SessionContext("test-session", new Environment());
+
+		hmsClient = HiveMetastoreClientFactory.create(hiveShell.getHiveConf(), null);
+	}
+
+	@Test
+	public void testWriteHiveTable() throws Exception {
+
+		// primitive types
+		hiveShell.execute("create database db1");
```
twalthr: End-to-end test should involve more features. I would suggest the following tests:

- List all tables and read some metadata
- read from an existing table
- register a new table in Hive
- use a registered function
- write to an existing table
- write to a new table
- Use all types that Hive provides in tables.

flink-connectors/flink-connector-hive/pom.xml
twalthr: btw I just saw that you marked this PR as `Dependencies (does it add or upgrade a dependency): no` please update this.

flink-end-to-end-tests/flink-connector-hive-test/src/test/java/org/apache/flink/batch/connectors/hive/LocalExecutorHiveConnectorTest.java
twalthr: True, yes then let's have a e2e test that submits a regular jar packaged program via the Flink command line.

twalthr:
@lirui-apache would you be up for implementing the suggested changes? This is a blocker for the 1.9 release and should happen in a timely fashion. Please let us know if somebody else should take over?

lirui-apache:
Hi @twalthr, thanks for your suggestions.
I agree a real e2e test should use what users would use in real deployments and is good to have. But implementing such tests requires extra efforts, e.g. rework the `FlinkStandaloneHiveRunner` to provide a standalone hive metastore server, and figure out how to test against different hive versions. Furthermore, development of such test cases are usually less efficient than ordinary JUnit tests, e.g. it'll be harder to debug when something goes wrong.

The intention of this PR is to improve our test coverage of hive connector. Currently, we only have quite primitive tests like `HiveTableSinkTest` and `HiveTableSourceTest`, which are not sufficient to catch issues like [FLINK-13279](https://issues.apache.org/jira/browse/FLINK-13279). I added this test in `flink-end-to-end-tests` instead of `flink-connector-hive` only because `flink-sql-client` depends on `flink-connector-hive`, and therefore we can't access `LocalExecutor` in `flink-connector-hive`.

So I guess another option is to remove the `flink-connector-hive` dependency in `flink-sql-client`. Make `flink-connector-hive` depend on `flink-sql-client` (in test scope), and move the test here to `flink-connector-hive` as an ordinary JUnit test. This way we can maintain the simplicity while being able to catch most of potential issues. Real e2e tests as you suggested can be implemented later, and perhaps in another release. Does this make sense to you?

twalthr:
Hi @lirui-apache,

yes implementing such tests requires extra efforts but we need to do perform these efforts before releasing. Unit tests or ITCases do not cover proper communication with external systems including serialization/deserialization and depencency/class loading issues. We can postpone testing different Hive versions but should at least test one Hive version in-depth.

The SQL Client has nothing to do with the Hive functionality. We can test the Hive connector without SQL Client and the core Table API only. If this issue does not cover the task of adding an e2e test, then please rename the issue, open a new one, and give someone else the chance to implement a full test.

lirui-apache:
@twalthr sure. This PR was split from #9088. I'll continue my work there and only add unit tests. I'd love to help with the e2e test after I finish functionality tasks on my plate. But please feel free to assign this issue to someone else to unblock the release.

xuefuz:
I agree with @twalthr  that we should add real end-to-end test. I will find resource to do this before 1.9 is released. Thanks.

lirui-apache:
One approach I can think of is to build a uber jar (in order to include deps like connector-hive, hive-exec, hiverunner, etc) of the e2e test cases. And run this uber jar via `flink run` of flink-dist (probably with a standalone cluster). The main class of the uber jar creates the HMS and embedded hive-shell, by calling util methods of `FlinkStandaloneHiveRunner`. And then runs all the test cases. This way maybe we can test with different hive versions, as long as we build the e2e uber jar with proper profiles.
What I'm not sure about is how we can make debugging easier. For example, it's better if we're able to run only subset of test cases. And have separated logs for each test case.

twalthr:
Thanks for your support @xuefuz. If you cannot find someone who can implement a real end-to-end test. please let me know. I haven't looked into the details but I don't think that we need `FlinkStandaloneHiveRunner`. We usually download the dependencies and configure them via `curl` or provided scripts. I'm sure Hive provides a command-line util to trigger HMS commends for the end-to-end test. The examples in `flink-end-to-end-tests/test-scripts` serve as good examples.

xuefuz:
Hi @twalthr , Given that Bowen is on the leave, we are short of bandwidth at the moment. Thus, we have to push this back for a while. If there is bandwidth from your side, please let me know. Otherwise, I will as @zjuwangg to look into this after he completes his current tasks. Thanks.

lirui-apache:
Closing this one since it'll be superseded by FLINK-13437

PR #9158, opened 2019-07-18
[FLINK-10052][coordination] Tolerate temporarily suspended ZooKeeper connections

lamber-ken:
### Desc
The `ZooKeeperLeaderElectionService` uses the LeaderLatch Curator recipe for leader election. The leader latch revokes leadership in case of a suspended ZooKeeper connection. This can be premature in case that the system can reconnect to ZooKeeper before its session expires. The effect of the lost leadership is that all jobs will be canceled and directly restarted after regaining the leadership. When more and more tasks deployed in cluster, the situation will get worse and worse.

### Aims
- Fix the impact of zookeeper network disconnect temporarily on flink long running jobs.
- Improve the stability of the flink longrunning jobs.

### Detail
Click here for more jira detail [FLINK-10052](https://issues.apache.org/jira/browse/FLINK-10052).

<br />

## Brief change log

  - upgrage curator-recipes dependency to 4.2.0
  - use `SessionConnectionStateErrorPolicy` as the error connect policy when build `CuratorFramework`

## Verifying this change

This change is already covered by existing tests, such as *ZooKeeperLeaderElectionITCase*.

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/leaderelection/ZooKeeperLeaderElectionService.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/leaderelection/ZooKeeperLeaderElectionService.java
@@ -410,8 +410,7 @@ public class ZooKeeperLeaderElectionService implements LeaderElectionService, Le
 				LOG.debug("Connected to ZooKeeper quorum. Leader election can start.");
 				break;
 			case SUSPENDED:
-				LOG.warn("Connection to ZooKeeper suspended. The contender " + leaderContender.getDescription()
-					+ " no longer participates in the leader election.");
+				LOG.warn("Connection to ZooKeeper suspended, waiting for reconnection.");
 				break;
 			case RECONNECTED:
 				LOG.info("Connection to ZooKeeper was reconnected. Leader election can be restarted.");
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/leaderretrieval/ZooKeeperLeaderRetrievalService.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/leaderretrieval/ZooKeeperLeaderRetrievalService.java
@@ -191,8 +191,7 @@ public class ZooKeeperLeaderRetrievalService implements LeaderRetrievalService, 
 				LOG.debug("Connected to ZooKeeper quorum. Leader retrieval can start.");
 				break;
 			case SUSPENDED:
-				LOG.warn("Connection to ZooKeeper suspended. Can no longer retrieve the leader from " +
-					"ZooKeeper.");
+				LOG.warn("Connection to ZooKeeper suspended, waiting for reconnection.");
 				break;
 			case RECONNECTED:
 				LOG.info("Connection to ZooKeeper was reconnected. Leader retrieval can be restarted.");
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/util/ZooKeeperUtils.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/util/ZooKeeperUtils.java
@@ -44,6 +44,7 @@ import org.apache.curator.framework.CuratorFrameworkFactory;
 import org.apache.curator.framework.api.ACLProvider;
 import org.apache.curator.framework.imps.DefaultACLProvider;
 import org.apache.curator.framework.recipes.cache.PathChildrenCache;
+import org.apache.curator.framework.state.SessionConnectionStateErrorPolicy;
 import org.apache.curator.retry.ExponentialBackoffRetry;
 import org.apache.zookeeper.ZooDefs;
 import org.apache.zookeeper.data.ACL;
@@ -129,6 +130,7 @@ public class ZooKeeperUtils {
 				.sessionTimeoutMs(sessionTimeout)
 				.connectionTimeoutMs(connectionTimeout)
 				.retryPolicy(new ExponentialBackoffRetry(retryWait, maxRetryAttempts))
+			    .connectionStateErrorPolicy(new SessionConnectionStateErrorPolicy())
 				// Curator prepends a '/' manually and throws an Exception if the
 				// namespace starts with a '/'.
 				.namespace(rootWithNamespace.startsWith("/") ? rootWithNamespace.substring(1) : rootWithNamespace)


flink-shaded-curator/src/main/resources/META-INF/NOTICE
```diff
@@ -6,7 +6,7 @@ The Apache Software Foundation (http://www.apache.org/).
 
 This project bundles the following dependencies under the Apache Software License 2.0 (http://www.apache.org/licenses/LICENSE-2.0.txt)
 
-- com.google.guava:guava:16.0.1
-- org.apache.curator:curator-client:2.12.0
-- org.apache.curator:curator-framework:2.12.0
-- org.apache.curator:curator-recipes:2.12.0
+- com.google.guava:guava:27.0.1-jre
```
TisonKun: Unexpected modification?

flink-runtime/src/main/java/org/apache/flink/runtime/leaderelection/ZooKeeperLeaderElectionService.java
```diff
@@ -397,8 +397,6 @@ protected void handleStateChange(ConnectionState newState) {
 				LOG.debug("Connected to ZooKeeper quorum. Leader election can start.");
 				break;
 			case SUSPENDED:
-				LOG.warn("Connection to ZooKeeper suspended. The contender " + leaderContender.getAddress()
```
TisonKun: I don't think we should dismiss this log in this patch. Change it into

```
LOG.warn("Connection to ZooKeeper suspended. The contender " + leaderContender.getAddress()
                  + " is waiting for reconnection.");
```

might be ok.

flink-runtime/src/main/java/org/apache/flink/runtime/leaderretrieval/ZooKeeperLeaderRetrievalService.java
```diff
@@ -191,8 +191,6 @@ protected void handleStateChange(ConnectionState newState) {
 				LOG.debug("Connected to ZooKeeper quorum. Leader retrieval can start.");
 				break;
 			case SUSPENDED:
-				LOG.warn("Connection to ZooKeeper suspended. Can no longer retrieve the leader from " +
```
TisonKun: The same as https://github.com/apache/flink/pull/9158/files#r304770340

flink-shaded-curator/src/main/resources/META-INF/NOTICE
lamber-ken: > Unexpected modification?

yes, curator-4.2.0 depend on guava-27.0.1, and flink-shaded-curator will relocate `com.google.common` to `org.apache.flink.curator.shaded.com.google.common`, so it has no effect.

flink-runtime/src/main/java/org/apache/flink/runtime/leaderelection/ZooKeeperLeaderElectionService.java
lamber-ken: 👍 Thanks, done

flink-runtime/src/main/java/org/apache/flink/runtime/leaderretrieval/ZooKeeperLeaderRetrievalService.java
lamber-ken: Thanks, done.

flink-shaded-curator/src/main/resources/META-INF/NOTICE
TisonKun: A better way would be excluding guava that curator depends on. I don't think we need this transitive dependency.

lamber-ken: > A better way would be excluding guava that curator depends on. I don't think we need this transitive dependency.

I am not sure that wheather curator-4.2.0 use some special guava's api or not, I am glad to test it offline.

TisonKun:

In fact, we can hardly say that we have reached a consensus on how to solve FLINK-10052. Although upgrading curator's version seems a clean way.

lamber-ken:
Hi, @TisonKun, thanks to your help 👍, if these are subtle issues, I think I can solve it.

lamber-ken:
hi, @rmetzger, how to rebuild the pr? I updated some codes, but ci didn't rebuild.

TisonKun:
@lamber-ken the community migrated ci from asf account to ververica sponsored account[1].

As you can see above @flinkbot would report ci status.

[1] https://issues.apache.org/jira/browse/INFRA-18703

lamber-ken:
> @lamber-ken the community migrated ci from asf account to ververica sponsored account[1].
> 
> As you can see above @flinkbot would report ci status.
> 
> [1] https://issues.apache.org/jira/browse/INFRA-18703

Ok, I see. 👍

TisonKun:
cc @tillrohrmann @StefanRRichter we need a committer to push this thread forward. It would be glad to see this issue also fixed in 1.9.0.

azagrebin:
@TisonKun @lamber-ken @Wosin 
should the curator upgrade be firstly discussed and done in a separate JIRA issue? Is it not affected by [FLINK-13417](https://issues.apache.org/jira/browse/FLINK-13417)? I would suggest to wait for some feedback from @tillrohrmann.
I would suggest we move to `SessionConnectionStateErrorPolicy` after upgrading the curator separately.
@TisonKun could you assign @lamber-ken in JIRA if we agree that this is the way to proceed? Correct if I'm wrong, @Wosin did not have time to work on this, right?

TisonKun:
Hi @azagrebin thanks for your attention.

>Is it not affected by FLINK-13417?

Nope. Curator itself deals with zk 3.4 compatibility issue. There shouldn't be a problem which one upgrade first.

>...after upgrading the curator separately

I'm ok to upgrade Curator in a separated thread. Since this pull request goes on its way there should be no further broker. @lamber-ken is it correct? How much it cost upgrade Curator first and then move to `SessionConnectionStateErrorPolicy`.

lamber-ken:
Hi, @azagrebin thanks for your attention.

I think it's no need to make it separately. We can discuss these in this mail list [FLINK-10052-ML](https://lists.apache.org/thread.html/39edb21031bcc366eb051c0aaa8cd9c28ca59ea5454f134df39d1e64@%3Cdev.flink.apache.org%3E)

tillrohrmann:
I'd prefer to split the version bump and the functional change into two PRs @lamber-ken.

lamber-ken:
> I'd prefer to split the version bump and the functional change into two PRs @lamber-ken.

Hi, trill. If so, I'll create

lamber-ken:
After someone fixs FLINIK-14177, will rebase new patchs. After that, this pr will go through.

zentol:
Subsumed by #11338.

PR #9162, opened 2019-07-18
[FLINK-13321][table-planner-blink] Fix bug in Blink Planner, Join a udf with constant arguments or without argument in TableAPI query does not work now

beyond1920:

In blink planner, Join a udf with constant arguments or without argument in TableAPI query does not work now. The root cause is the `FlinkLogicalTableFunctionScan`.CONVERTER translates a `TableFunctionScan` to a `Correlate`, it will translate the original `RelNode` tree to a `RelNode` with two Cascaded ·Correlate` (could be found in the above thrown message), which could not translate to Physical `RelNode`. The detail information could be get in https://issues.apache.org/jira/browse/FLINK-13321.

`FlinkLogicalTableFunctionScan`.CONVERTER is to deal with a special sql like 'SELECT * FROM LATERAL TABLE(FUNC()) as T(C1)'. The pr aims to fix above bug by updating `FlinkLogicalTableFunctionScan`.CONVERTER, and process the sql like 'SELECT * FROM LATERAL TABLE(FUNC()) as T(C1)' in the other solution.

## Brief change log

  - Update `FlinkLogicalTableFunctionScan`.CONVERTER to avoid rewrite `FlinkLogicalTableFunctionScan` node 
  - Add `BatchExecConstantTableFunctionScanRule` and `StreamExecConstantTableFunctionScanRule` to deal with sql like 'SELECT * FROM LATERAL TABLE(FUNC()) as T(C1)'.

## Verifying this change
ITCase

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/plan/rules/physical/batch/BatchExecConstantTableFunctionScanRule.scala
+++ b/flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/plan/rules/physical/batch/BatchExecConstantTableFunctionScanRule.scala
@@ -1,0 +1,85 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.table.plan.rules.physical.batch
+
+import org.apache.flink.table.plan.nodes.FlinkConventions
+import org.apache.flink.table.plan.nodes.logical.FlinkLogicalTableFunctionScan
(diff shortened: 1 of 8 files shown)


flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/plan/rules/physical/batch/BatchExecConstantTableFunctionScanRule.scala
```diff
@@ -0,0 +1,80 @@
...
+/**
+  * Converts [[FlinkLogicalTableFunctionScan]] with constant RexCall to
+  * {{{
+  *                    [[BatchExecCorrelate]]
+  *                          /       \
+  * empty [[BatchExecValues]]  [[FlinkLogicalTableFunctionScan]]
+  * }}}
+  *
+  * Add the rule to support the following SQL:
+  * SELECT * FROM LATERAL TABLE(FUNC()) as T(C1)
```
godfreyhe: it's better add some comments to explain "how does `SELECT * FROM T, LATERAL TABLE(FUNC()) as T(C1)` work"

flink-table/flink-table-planner-blink/src/test/scala/org/apache/flink/table/runtime/stream/sql/CorrelateITCase.scala
```diff
@@ -84,6 +84,30 @@ class CorrelateITCase extends StreamingTestBase {
...
 
+  @Test
+  def testConstantTableFunc(): Unit = {
+    tEnv.registerFunction("str_split", new StringSplit())
+    val query = "SELECT * FROM LATERAL TABLE(str_split()) as T0(d)"
+    val sink = new TestingAppendSink
+    tEnv.sqlQuery(query).toAppendStream[Row].addSink(sink)
+    env.execute()
+
+    val expected = List("a","b","c")
```
godfreyhe: nit: add blank between`a,b,c`

wuchong:
Waiting for the travis passing.

wuchong:
Travis is passed: https://travis-ci.com/flink-ci/flink/builds/119766191
Merging...

PR #9165, opened 2019-07-18
[FLINK-13055] [runtime] Leverage JM tracked partitions to find out unavailable partitions in …

zhuzhurk:
…time

*In current region failover process, most of the input result partition states are unknown. Even though the failure cause is a PartitionException, only one unhealthy partition can be identified.*

*The may lead to multiple unsuccessful failovers before all the unhealthy but needed partitions are identified and their producers are involved in the failover as well. (unsuccessful failover here means the recovered tasks get failed again soon due to some missing input partitions.)*

*This PR leverages JM side tracked partition states to help the region failover to identify unhealthy(missing) partitions earlier, thus to improve the failover experience.*

## Brief change log

  - *Implemented ExecutionGraphResultPartitionAvailabilityChecker*
  - *Let ExecutionGraph build its ExecutionGraphResultPartitionAvailabilityChecker and use it to generate region failover strategy*
  - *Enabled PartitionTracker to support querying whether a given partition is tracked*

## Verifying this change

  - *Added UT for ExecutionGraphResultPartitionAvailabilityChecker*

## Documentation

### DIFF
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/ExecutionGraphResultPartitionAvailabilityChecker.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/ExecutionGraphResultPartitionAvailabilityChecker.java
@@ -19,7 +19,10 @@ package org.apache.flink.runtime.executiongraph;
 
 import org.apache.flink.runtime.executiongraph.failover.flip1.ResultPartitionAvailabilityChecker;
 import org.apache.flink.runtime.io.network.partition.PartitionTracker;
+import org.apache.flink.runtime.io.network.partition.ResultPartitionID;
 import org.apache.flink.runtime.jobgraph.IntermediateResultPartitionID;
+
+import java.util.function.Function;
 
 import static org.apache.flink.util.Preconditions.checkNotNull;
 
@@ -29,19 +32,22 @@ import static org.apache.flink.util.Preconditions.checkNotNull;
  */
 public class ExecutionGraphResultPartitionAvailabilityChecker implements ResultPartitionAvailabilityChecker {
 
-	/** The execution graph that hosts the result partitions. */
-	private final ExecutionGraph executionGraph;
+	/** The function maps an IntermediateResultPartitionID to a ResultPartitionID. */
+	private final Function<IntermediateResultPartitionID, ResultPartitionID> partitionIDMapper;
 
 	/** The tracker that tracks all available result partitions. */
 	private final PartitionTracker partitionTracker;
 
-	ExecutionGraphResultPartitionAvailabilityChecker(final ExecutionGraph executionGraph) {
-		this.executionGraph = checkNotNull(executionGraph);
-		this.partitionTracker = executionGraph.getPartitionTracker();
+	ExecutionGraphResultPartitionAvailabilityChecker(
+			final Function<IntermediateResultPartitionID, ResultPartitionID> partitionIDMapper,
+			final PartitionTracker partitionTracker) {
+
+		this.partitionIDMapper = checkNotNull(partitionIDMapper);
+		this.partitionTracker = checkNotNull(partitionTracker);
 	}
 
 	@Override
 	public boolean isAvailable(final IntermediateResultPartitionID resultPartitionID) {
-		return partitionTracker.isPartitionTracked(executionGraph.createResultPartitionId(resultPartitionID));
+		return partitionTracker.isPartitionTracked(partitionIDMapper.apply(resultPartitionID));
 	}
 }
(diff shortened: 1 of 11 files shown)


flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/ExecutionGraphResultPartitionAvailabilityChecker.java
```diff
@@ -0,0 +1,47 @@
...
+ */
+public class ExecutionGraphResultPartitionAvailabilityChecker implements ResultPartitionAvailabilityChecker {
+
+	/** The execution graph that hosts the result partitions. */
+	private final ExecutionGraph executionGraph;
+
+	/** The tracker that tracks all available result partitions. */
+	private final PartitionTracker partitionTracker;
+
+	ExecutionGraphResultPartitionAvailabilityChecker(final ExecutionGraph executionGraph) {
```
zentol: pass in a `Function<IntermediateResultPartitionID, ResultPartitionID>` instead; this can then be a simply method reference `ExecutionGraph::createResultPartitionId`.
The `ExecutionGraphResultPartitionAvailabilityChecker` could then be simplified by quite a bit.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/AdaptedRestartPipelinedRegionStrategyNGFailoverTest.java
```diff
@@ -379,10 +385,20 @@ private ExecutionGraph createExecutionGraph(
 			final JobGraph jobGraph,
 			final RestartStrategy restartStrategy) throws Exception {
 
+		final Configuration jmConfig = new Configuration();
+		jmConfig.setBoolean(JobManagerOptions.FORCE_PARTITION_RELEASE_ON_CONSUMPTION, false);
```
zentol: this should no longer be necessary; `false` is now the default.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/AdaptedRestartPipelinedRegionStrategyNGConcurrentFailoverTest.java
```diff
@@ -250,37 +253,38 @@ public void testSkipFailoverIfExecutionStateIsNotRunning() throws Exception {
...
 		v2.connectNewDataSetAsInput(v1, DistributionPattern.ALL_TO_ALL, ResultPartitionType.BLOCKING);
 
-		JobGraph jg = new JobGraph(TEST_JOB_ID, "testjob", v1, v2);
-		graph.attachJobGraph(jg.getVerticesSortedTopologicallyFromSources());
+		final JobGraph jg = new JobGraph(TEST_JOB_ID, "testjob", v1, v2);
+
+		final SimpleSlotProvider slotProvider = new SimpleSlotProvider(TEST_JOB_ID, DEFAULT_PARALLELISM);
+
+		final Configuration jmConfig = new Configuration();
+		jmConfig.setBoolean(JobManagerOptions.FORCE_PARTITION_RELEASE_ON_CONSUMPTION, false);
```
zentol: this should no longer be necessary; `false` is now the default.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/AdaptedRestartPipelinedRegionStrategyNGConcurrentFailoverTest.java
```diff
@@ -250,37 +253,38 @@ public void testSkipFailoverIfExecutionStateIsNotRunning() throws Exception {
...
+			.setSlotProvider(slotProvider)
+			.setJobMasterConfig(jmConfig)
+			.setPartitionTracker(partitionTracker)
+			.build();
 
-		graph.start(componentMainThreadExecutor);
+		eg.start(componentMainThreadExecutor);
 
-		return graph;
+		return eg;
```
zentol: let's stick to the existing naming

zhuzhurk: Ok.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/AdaptedRestartPipelinedRegionStrategyNGFailoverTest.java
zhuzhurk: Yes.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/AdaptedRestartPipelinedRegionStrategyNGConcurrentFailoverTest.java
zhuzhurk: Yes.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/ExecutionGraphResultPartitionAvailabilityChecker.java
zhuzhurk: Good idea.

PR #9169, opened 2019-07-18
[FLINK-12998][docs] Update documentation for file systems loading as plugins

1u0:

This PR updates documentation for Flink supported file systems to use `plugins` directory instead of `lib` to load optional file system components.

## Brief change log

  - the documentation in `ops/filesystems` are updated to tell users to use `plugins` dir instead of `lib`;
  - the documentation of `batch/connectors` is updated to reference to file system documentation instead of listing all supported file systems.

## Verifying this change

## Documentation

### DIFF
--- a/docs/dev/batch/connectors.md
+++ b/docs/dev/batch/connectors.md
@@ -25,18 +25,10 @@ under the License.
 * TOC
 {:toc}
 
-## Reading from file systems
+## Reading from and writing to file systems
 
-Flink has built-in support for the following file systems:
-
-| Filesystem                            | Scheme       | Notes  |
-| ------------------------------------- |--------------| ------ |
-| Hadoop Distributed File System (HDFS) &nbsp; | `hdfs://`    | All HDFS versions are supported |
-| Amazon S3                             | `s3://`      | Support through Hadoop file system implementation (see below) |
-| MapR file system                      | `maprfs://`  | The user has to manually place the required jar files in the `lib/` dir |
-| Alluxio                               | `alluxio://` &nbsp; | Support through Hadoop file system implementation (see below) |
-
-
+The Apache Flink project supports multiple [file systems]({{ site.baseurl }}/ops/filesystems/index.html) that can be used as backing stores
+for input and output connectors. 
 
 ### Using Hadoop file system implementations
 
@@ -93,7 +85,7 @@ Also, the serialization framework of Flink is able to handle classes generated f
 <dependency>
   <groupId>org.apache.flink</groupId>
   <artifactId>flink-avro</artifactId>
-  <version>{{site.version }}</version>
+  <version>{{ site.version }}</version>
 </dependency>
 {% endhighlight %}
 
--- a/docs/dev/batch/connectors.zh.md
+++ b/docs/dev/batch/connectors.zh.md
@@ -25,18 +25,10 @@ under the License.
 * TOC
 {:toc}
 
-## Reading from file systems
+## Reading from and writing to file systems
 
-Flink has built-in support for the following file systems:
-
-| Filesystem                            | Scheme       | Notes  |
-| ------------------------------------- |--------------| ------ |
-| Hadoop Distributed File System (HDFS) &nbsp; | `hdfs://`    | All HDFS versions are supported |
-| Amazon S3                             | `s3://`      | Support through Hadoop file system implementation (see below) |
-| MapR file system                      | `maprfs://`  | The user has to manually place the required jar files in the `lib/` dir |
-| Alluxio                               | `alluxio://` &nbsp; | Support through Hadoop file system implementation (see below) |
-
-
+The Apache Flink project supports multiple [file systems]({{ site.baseurl }}/ops/filesystems/index.html) that can be used as backing stores
+for input and output connectors. 
 
 ### Using Hadoop file system implementations
 
@@ -93,7 +85,7 @@ Also, the serialization framework of Flink is able to handle classes generated f
 <dependency>
   <groupId>org.apache.flink</groupId>
   <artifactId>flink-avro</artifactId>
-  <version>{{site.version }}</version>
+  <version>{{ site.version }}</version>
 </dependency>
 {% endhighlight %}
 
--- a/docs/monitoring/debugging_classloading.md
+++ b/docs/monitoring/debugging_classloading.md
@@ -28,17 +28,20 @@ under the License.
 ## Overview of Classloading in Flink
 
 When running Flink applications, the JVM will load various classes over time.
-These classes can be divided into two domains:
+These classes can be divided into three groups based on their origin:
 
   - The **Java Classpath**: This is Java's common classpath, and it includes the JDK libraries, and all code
-    in Flink's `/lib` folder (the classes of Apache Flink and its core dependencies).
+    in Flink's `/lib` folder (the classes of Apache Flink and some dependencies).
+
+  - The **Flink Plugin Components**: The plugins code in folders under Flink's `/plugins` folder. Flink's plugin mechanism will dynamically load them once during startup.
 
   - The **Dynamic User Code**: These are all classes that are included in the JAR files of dynamically submitted jobs,
     (via REST, CLI, web UI). They are loaded (and unloaded) dynamically per job.
 
-What classes are part of which domain depends on the particular setup in which you run Apache Flink. As a general rule, whenever you start the Flink
-processes first, and submit jobs, the job's classes are loaded dynamically. If the Flink processes are started together with the job/application,
-or the application spawns the Flink components (JobManager, TaskManager, etc.) then all classes are in the Java classpath.
+As a general rule, whenever you start the Flink processes first and submit jobs later, the job's classes are loaded dynamically.
+If the Flink processes are started together with the job/application, or if the application spawns the Flink components (JobManager, TaskManager, etc.), then all job's classes are in the Java classpath.
+
+Code in plugin components is loaded dynamically once by a dedicated class loader per plugin.
 
 In the following are some more details about the different deployment modes:
 
@@ -59,7 +62,7 @@ created for an job/application and will contain the job/application's jar files.
 **Docker / Kubernetes Sessions**
 
 Docker / Kubernetes setups that start first a set of JobManagers / TaskManagers and then submit jobs/applications via REST or the CLI
-behave like standalone sessions: Flink's code is in the Java classpath, the job's code is loaded dynamically.
+behave like standalone sessions: Flink's code is in the Java classpath, plugin components are loaded dynamically at startup and the job's code is loaded dynamically.
 
 
 **YARN**
@@ -67,8 +70,8 @@ behave like standalone sessions: Flink's code is in the Java classpath, the job'
 YARN classloading differs between single job deployments and sessions:
 
   - When submitting a Flink job/application directly to YARN (via `bin/flink run -m yarn-cluster ...`), dedicated TaskManagers and
-    JobManagers are started for that job. Those JVMs have both Flink framework classes and user code classes in the Java classpath.
-    That means that there is *no dynamic classloading* involved in that case.
+    JobManagers are started for that job. Those JVMs have user code classes in the Java classpath.
+    That means that there is *no dynamic classloading* involved in that case for the job.
 
   - When starting a YARN session, the JobManagers and TaskManagers are started with the Flink framework classes in the
     classpath. The classes from all jobs that are submitted against the session are loaded dynamically.
@@ -82,37 +85,37 @@ classes are loaded dynamically when the jobs are submitted.
 
 ## Inverted Class Loading and ClassLoader Resolution Order
 
-In setups where dynamic classloading is involved (sessions), there is a hierarchy of typically two ClassLoaders: 
-(1) Java's *application classloader*, which has all classes in the classpath, and (2) the dynamic *user code classloader*.
-for loading classes from the user-code jar(s). The user-code ClassLoader has the application classloader as its parent.
+In setups where dynamic classloading is involved (plugin components, Flink jobs in session setups), there is a hierarchy of typically two ClassLoaders: 
+(1) Java's *application classloader*, which has all classes in the classpath, and (2) the dynamic *plugin/user code classloader*.
+for loading classes from the plugin or the user-code jar(s). The dynamic ClassLoader has the application classloader as its parent.
 
-By default, Flink inverts classloading order, meaning it looks into the user code classloader first, and only looks into
-the parent (application classloader) if the class is not part of the dynamically loaded user code.
+By default, Flink inverts classloading order, meaning it looks into the dynamic classloader first, and only looks into
+the parent (application classloader) if the class is not part of the dynamically loaded code.
 
-The benefit of inverted classloading is that jobs can use different library versions than Flink's core itself, which is very
+The benefit of inverted classloading is that plugins and jobs can use different library versions than Flink's core itself, which is very
 useful when the different versions of the libraries are not compatible. The mechanism helps to avoid the common dependency conflict
 errors like `IllegalAccessError` or `NoSuchMethodError`. Different parts of the code simply have separate copies of the
-classes (Flink's core or one of its dependencies can use a different copy than the user code).
+classes (Flink's core or one of its dependencies can use a different copy than the user code or plugin code).
 In most cases, this works well and no additional configuration from the user is needed.
 
-However, there are cases when the inverted classloading causes problems (see below, "X cannot be cast to X"). 
-You can revert back to Java's default mode by configuring the ClassLoader resolution order via
-[classloader.resolve-order](../ops/config.html#classloader-resolve-order) in the Flink config to `parent-first`
+However, there are cases when the inverted classloading causes problems (see below, ["X cannot be cast to X"](#x-cannot-be-cast-to-x-exceptions)). 
+For user code classloading, you can revert back to Java's default mode by configuring the ClassLoader resolution order via
+[`classloader.resolve-order`](../ops/config.html#classloader-resolve-order) in the Flink config to `parent-first`
 (from Flink's default `child-first`).
 
 Please note that certain classes are always resolved in a *parent-first* way (through the parent ClassLoader first), because they
-are shared between Flink's core and the user code or the user-code facing APIs. The packages for these classes are configured via 
-[classloader.parent-first-patterns-default](../ops/config.html#classloader-parent-first-patterns-default) and
-[classloader.parent-first-patterns-additional](../ops/config.html#classloader-parent-first-patterns-additional).
+are shared between Flink's core and the plugin/user code or the plugin/user-code facing APIs. The packages for these classes are configured via 
+[`classloader.parent-first-patterns-default`](../ops/config.html#classloader-parent-first-patterns-default) and
+[`classloader.parent-first-patterns-additional`](../ops/config.html#classloader-parent-first-patterns-additional).
 To add new packages to be *parent-first* loaded, please set the `classloader.parent-first-patterns-additional` config option.
 
 
-## Avoiding Dynamic Classloading
+## Avoiding Dynamic Classloading for User Code
 
 All components (JobManger, TaskManager, Client, ApplicationMaster, ...) log their classpath setting on startup.
 They can be found as part of the environment information at the beginning of the log.
 
-When running a setup where the Flink JobManager and TaskManagers are exclusive to one particular job, one can put JAR files
+When running a setup where the Flink JobManager and TaskManagers are exclusive to one particular job, one can put user code JAR files
 directly into the `/lib` folder to make sure they are part of the classpath and not loaded dynamically.
 
 It usually works to put the job's JAR file into the `/lib` directory. The JAR will be part of both the classpath
@@ -125,7 +128,7 @@ used by multiple jobs), it may still be possible to put common libraries to the 
 for those.
 
 
-## Manual Classloading in the Job
+## Manual Classloading in User Code
 
 In some cases, a transformation function, source, or sink needs to manually load classes (dynamically via reflection).
 To do that, it needs the classloader that has access to the job's classes.
@@ -149,9 +152,9 @@ The solution here is to either have a setup without any dynamic classloading, or
 The latter means that the library must not be added to Flink's `/lib` folder, but must be part of the application's fat-jar/uber-jar
 
 
-## Unloading of Dynamically Loaded Classes
+## Unloading of Dynamically Loaded Classes in User Code
 
-All scenarios that involve dynamic class loading (sessions) rely on classes being *unloaded* again.
+All scenarios that involve dynamic user code classloading (sessions) rely on classes being *unloaded* again.
 Class unloading means that the Garbage Collector finds that no objects from a class exist and more, and thus removes the class
 (the code, static variable, metadata, etc).
 
--- a/docs/monitoring/debugging_classloading.zh.md
+++ b/docs/monitoring/debugging_classloading.zh.md
@@ -28,17 +28,20 @@ under the License.
 ## Overview of Classloading in Flink
 
 When running Flink applications, the JVM will load various classes over time.
-These classes can be divided into two domains:
+These classes can be divided into three groups based on their origin:
 
   - The **Java Classpath**: This is Java's common classpath, and it includes the JDK libraries, and all code
-    in Flink's `/lib` folder (the classes of Apache Flink and its core dependencies).
+    in Flink's `/lib` folder (the classes of Apache Flink and some dependencies).
+
+  - The **Flink Plugin Components**: The plugins code in folders under Flink's `/plugins` folder. Flink's plugin mechanism will dynamically load them once during startup.
 
   - The **Dynamic User Code**: These are all classes that are included in the JAR files of dynamically submitted jobs,
     (via REST, CLI, web UI). They are loaded (and unloaded) dynamically per job.
 
-What classes are part of which domain depends on the particular setup in which you run Apache Flink. As a general rule, whenever you start the Flink
-processes first, and submit jobs, the job's classes are loaded dynamically. If the Flink processes are started together with the job/application,
-or the application spawns the Flink components (JobManager, TaskManager, etc.) then all classes are in the Java classpath.
+As a general rule, whenever you start the Flink processes first and submit jobs later, the job's classes are loaded dynamically.
+If the Flink processes are started together with the job/application, or if the application spawns the Flink components (JobManager, TaskManager, etc.), then all job's classes are in the Java classpath.
+
+Code in plugin components is loaded dynamically once by a dedicated class loader per plugin.
 
 In the following are some more details about the different deployment modes:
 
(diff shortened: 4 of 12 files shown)


1u0:
cc @StefanRRichter.

docs/ops/filesystems/index.md
```diff
@@ -35,32 +35,50 @@ File system instances are instantiated once per process and then cached/pooled,
...
+Flink has built-in support for the file system of the local machine, including any NFS or SAN drives mounted into that local file system.
+It can be used by default without additional configuration. Local files are referenced with the *file://* URI scheme.
 
-  - **local**: This file system is used when the scheme is *"file://"*, and it represents the file system of the local machine, including any NFS or SAN drives mounted into that local file system.
+## Pluggable File Systems
 
-  - **S3**: Flink directly provides file systems to talk to Amazon S3 with two alternative implementations, `flink-s3-fs-presto` and `flink-s3-fs-hadoop`. Both implementations are self-contained with no dependency footprint.
-    
-  - **MapR FS**: The MapR file system *"maprfs://"* is automatically available when the MapR libraries are in the classpath.  
+Apache Flink project supports the following file systems as plugins:
```
zentol: do we have any reference _anywhere_ what a plugin is in the context of Flink?

docs/ops/filesystems/index.md
```diff
@@ -35,32 +35,50 @@ File system instances are instantiated once per process and then cached/pooled,
...
+
+  - [**Amazon S3**](./s3.html) object storage is supported by two alternative implementations, `flink-s3-fs-presto` and `flink-s3-fs-hadoop`.
+  Both implementations are self-contained with no dependency footprint.
+
+  - **MapR FS**: The MapR file system *maprfs://* is automatically available when the MapR libraries are in the classpath.
+  <span class="label label-info">Note</span> You'd need to build from sources the `flink-mapr-fs` JAR file.
```
zentol: This should be incorrect; the flink-mapr-fs files are contained in the flink-dist jar. As such building the jar should not be necessary; it isn't doing any packaging.

docs/ops/filesystems/index.md
```diff
@@ -35,32 +35,50 @@ File system instances are instantiated once per process and then cached/pooled,
...
+
+To use a pluggable file systems, copy the corresponding JAR file from the `opt` directory to a directory under `plugins` directory
+of your Flink distribution before starting Flink, e.g.
+
+{% highlight bash %}
+mkdir ./plugins/s3-fs-hadoop
+cp ./opt/flink-s3-fs-hadoop-{{ site.version }}.jar ./plugins/s3-fs-hadoop/
+{% endhighlight %}
+
+<span class="label label-danger">Attention</span> The plugin mechanism for file systems was introduced since Flink version `1.9` to
```
zentol:
```suggestion
<span class="label label-danger">Attention</span> The plugin mechanism for file systems was introduced in Flink version `1.9` to
```

docs/ops/filesystems/index.md
```diff
@@ -35,32 +35,50 @@ File system instances are instantiated once per process and then cached/pooled,
...
+support dedicated Java class loaders per file system and move away from class shading mechanism. You can still use the provided file systems
+(or your own implementations) via the old mechanism by copying corresponding JAR file into `lib` directory.
```
zentol:
```suggestion
(or your own implementations) via the old mechanism by copying the corresponding JAR file into `lib` directory.
```

docs/ops/filesystems/index.md
```diff
@@ -35,32 +35,50 @@ File system instances are instantiated once per process and then cached/pooled,
...
+
+The non plugin-based loading mechanism is deprecated and will be not supported in the future.
```
zentol: Are we actually removing support for having filesystems in lib?

docs/ops/filesystems/index.md
```diff
@@ -83,17 +101,21 @@ fs.hdfs.hadoopconf: /path/to/etc/hadoop
 This registers `/path/to/etc/hadoop` as Hadoop's configuration directory and is where Flink will look for the `core-site.xml` and `hdfs-site.xml` files.
 
 
-## Adding new File System Implementations
+## Adding a new pluggable File System implementation
 
-File systems are represented via the `org.apache.flink.core.fs.FileSystem` class, which captures the ways to access and modify files and objects in that file system. 
-Implementations are discovered by Flink through Java's service abstraction, making it easy to add new file system implementations.
```
zentol: This looks like it still valid and could be kept as is imo.

docs/ops/filesystems/index.md
```diff
@@ -83,17 +101,21 @@ fs.hdfs.hadoopconf: /path/to/etc/hadoop
...
 
 To add a new file system:
 
   - Add the File System implementation, which is a subclass of `org.apache.flink.core.fs.FileSystem`.
   - Add a factory that instantiates that file system and declares the scheme under which the FileSystem is registered. This must be a subclass of `org.apache.flink.core.fs.FileSystemFactory`.
-  - Add a service entry. Create a file `META-INF/services/org.apache.flink.core.fs.FileSystemFactory` which contains the class name of your file system factory class.
+  - Add a service entry. Create a file `META-INF/services/org.apache.flink.core.fs.FileSystemFactory` which contains the class name of your file system factory class
+  (see the [Java Service Loader docs](https://docs.oracle.com/javase/8/docs/api/java/util/ServiceLoader.html) for more details).
+
+During plugins discovery, the file system factory class would be loaded by a dedicated Java class loader to avoid class conflicts with other plugins and Flink components.
```
zentol:
```suggestion
During plugin discovery, the file system factory class would be loaded by a dedicated Java class loader to avoid class conflicts with other plugins and Flink components.
```

docs/ops/filesystems/index.md
```diff
@@ -83,17 +101,21 @@ fs.hdfs.hadoopconf: /path/to/etc/hadoop
...
+The same class loader should be used during file system instantiation and the file system operation calls.
```
zentol:
```suggestion
The same class loader will be used during file system instantiation and the file system operation calls.
```

docs/ops/filesystems/index.md
```diff
@@ -83,17 +101,21 @@ fs.hdfs.hadoopconf: /path/to/etc/hadoop
...
 
-See the [Java Service Loader docs](https://docs.oracle.com/javase/8/docs/api/java/util/ServiceLoader.html) for more details on how service loaders work.
+<span class="label label-warning">Warning</span> In practice, it means you should avoid using `Thread.currentThread().getContextClassLoader()` class loader
```
zentol:
```suggestion
<span class="label label-warning">Warning</span> In practice, it means you must not use the ClassLoader returned by  `Thread.currentThread().getContextClassLoader()`
```

docs/ops/filesystems/index.md
1u0: Not removing it now.
But in my understanding the plan (introducing and using plugins effort) is to move away from shared (lib based) class loader and shading.

docs/ops/filesystems/index.md
1u0: I've removed it because it was not exact:
 * it is not, that `FileSystem` classes are discovered by Java's service loading mechanism, but `FileSystemFactory`;
 * there is still reference left to Java's service loading in the `Add a service entry` step.

docs/ops/filesystems/index.md
1u0: No, we don't have it, not yet. In this document, there is a follow up description that tells that those file systems should be copied under `plugins` dir before Flink startup to enable them.

Although, the referenced ticket focuses on documenting the plugins mechanism, I've focused mostly on updating file systems setup docs. My reasoning is the following:
 * currently only file system components are used by the plugins mechanism. Imo, it's fine for now to not have a dedicated documentation page for plugins in general. Also, there is not much to say to users about plugins mechanism, besides which Java interface it's and that they are loaded from `plugins` dir;
 * it's possible that the plugin mechanism/lifecycle may change in the future, when it's applied to other Flink components.

My proposal would be to wait when there are more use cases of plugins and when it's more refined. At the moment, I find it's scope too small to document as a separate section.

docs/ops/filesystems/index.md
zentol: OK, but "Not being supported" is a bit ambiguous; it could mean that it doesn't work at all, or that the Flink project provides no assistance for helping with issue resulting from placing filesystems in /lib.

There's also the MapR filesystem that cannot even be used as a plugin right now.

We could maybe rephrase it along the lines of discouraging the /lib approach for file systems that support the plugin mechanism.

docs/ops/filesystems/index.md
zentol: Ok, given that this section is for developers of new filesystems I'm okay with the change.

docs/ops/filesystems/index.md
1u0: I don't know what is the plan with MapR.
To totally deprecate `/lib` based FS jars loading, MapR adapter also may probably need to become a separate plugin.

The main motivation to use plugins is to get rid of shading (and to allow those components to have their own versions of dependent libraries). Allowing to use the application class loader to load such components would be error-prone and not user friendly.

> OK, but "Not being supported" is a bit ambiguous

I've changed/re-phrased this paragraph, although the new version may sound not concrete too. 😄

docs/dev/batch/connectors.md
```diff
@@ -25,18 +25,10 @@ under the License.
...
-| Filesystem                            | Scheme       | Notes  |
-| ------------------------------------- |--------------| ------ |
-| Hadoop Distributed File System (HDFS) &nbsp; | `hdfs://`    | All HDFS versions are supported |
-| Amazon S3                             | `s3://`      | Support through Hadoop file system implementation (see below) |
-| MapR file system                      | `maprfs://`  | The user has to manually place the required jar files in the `lib/` dir |
-| Alluxio                               | `alluxio://` &nbsp; | Support through Hadoop file system implementation (see below) |
-
+## Reading from and writing to file systems
 
+Apache Flink project supports multiple [file systems]({{ site.baseurl }}/ops/filesystems/index.html) that can be used as backing stores
```
alpinegizmo: Either "The Apache Flink project ..." or "Apache Flink supports ...".

docs/dev/batch/connectors.zh.md
```diff
@@ -25,18 +25,10 @@ under the License.
...
-| Filesystem                            | Scheme       | Notes  |
-| ------------------------------------- |--------------| ------ |
-| Hadoop Distributed File System (HDFS) &nbsp; | `hdfs://`    | All HDFS versions are supported |
-| Amazon S3                             | `s3://`      | Support through Hadoop file system implementation (see below) |
-| MapR file system                      | `maprfs://`  | The user has to manually place the required jar files in the `lib/` dir |
-| Alluxio                               | `alluxio://` &nbsp; | Support through Hadoop file system implementation (see below) |
-
+## Reading from and writing to file systems
 
+Apache Flink project supports multiple [file systems]({{ site.baseurl }}/ops/filesystems/index.html) that can be used as backing stores
```
alpinegizmo: Same as above. Either "the project" supports it, or "Flink" does.

docs/monitoring/debugging_classloading.md
```diff
@@ -28,17 +28,20 @@ under the License.
...
 
 When running Flink applications, the JVM will load various classes over time.
-These classes can be divided into two domains:
+These classes can be divided into three groups based on their origin:
 
   - The **Java Classpath**: This is Java's common classpath, and it includes the JDK libraries, and all code
-    in Flink's `/lib` folder (the classes of Apache Flink and its core dependencies).
+    in Flink's `/lib` folder (the classes of Apache Flink and some dependencies).
+
+  - The **Flink Plugin Components**: All code in folders under Flink's `/plugins` folder that loaded dynamically once via plugins mechanism during Flink's components start up.
```
alpinegizmo: I think this is trying to say that

Flink's plugin mechanism will dynamically load all the code contained in folders under Flink's `/plugins` folder once during startup.

docs/monitoring/debugging_classloading.md
```diff
@@ -28,17 +28,20 @@ under the License.
...
 
   - The **Dynamic User Code**: These are all classes that are included in the JAR files of dynamically submitted jobs,
     (via REST, CLI, web UI). They are loaded (and unloaded) dynamically per job.
 
-What classes are part of which domain depends on the particular setup in which you run Apache Flink. As a general rule, whenever you start the Flink
-processes first, and submit jobs, the job's classes are loaded dynamically. If the Flink processes are started together with the job/application,
-or the application spawns the Flink components (JobManager, TaskManager, etc.) then all classes are in the Java classpath.
+As a general rule, whenever you start the Flink processes first and submit jobs later, the job's classes are loaded dynamically.
+If the Flink processes are started together with the job/application, or the application spawns the Flink components (JobManager, TaskManager, etc.) then all job's classes are in the Java classpath.
```
alpinegizmo: ... or if the application spawns the Flink components (JobManager, TaskManager, etc.), then ...

docs/monitoring/debugging_classloading.zh.md
```diff
@@ -28,17 +28,20 @@ under the License.
...
 
 When running Flink applications, the JVM will load various classes over time.
-These classes can be divided into two domains:
+These classes can be divided into three groups based on their origin:
 
   - The **Java Classpath**: This is Java's common classpath, and it includes the JDK libraries, and all code
-    in Flink's `/lib` folder (the classes of Apache Flink and its core dependencies).
+    in Flink's `/lib` folder (the classes of Apache Flink and some dependencies).
+
+  - The **Flink Plugin Components**: All code in folders under Flink's `/plugins` folder that loaded dynamically once via plugins mechanism during Flink's components start up.
```
alpinegizmo: same as before

docs/monitoring/debugging_classloading.zh.md
```diff
@@ -28,17 +28,20 @@ under the License.
...
 
   - The **Dynamic User Code**: These are all classes that are included in the JAR files of dynamically submitted jobs,
     (via REST, CLI, web UI). They are loaded (and unloaded) dynamically per job.
 
-What classes are part of which domain depends on the particular setup in which you run Apache Flink. As a general rule, whenever you start the Flink
-processes first, and submit jobs, the job's classes are loaded dynamically. If the Flink processes are started together with the job/application,
-or the application spawns the Flink components (JobManager, TaskManager, etc.) then all classes are in the Java classpath.
+As a general rule, whenever you start the Flink processes first and submit jobs later, the job's classes are loaded dynamically.
+If the Flink processes are started together with the job/application, or the application spawns the Flink components (JobManager, TaskManager, etc.) then all job's classes are in the Java classpath.
```
alpinegizmo: same as before

docs/ops/filesystems/azure.md
```diff
@@ -38,7 +38,7 @@ wasb://<your-container>@$<your-azure-account>.blob.core.windows.net/<object-path
 wasbs://<your-container>@$<your-azure-account>.blob.core.windows.net/<object-path>
 {% endhighlight %}
 
-Below shows how to use Azure Blob Storage with Flink:
+Below shows how to use Azure Blob Storage in a Flink job:
```
alpinegizmo: See below for how to use ...

docs/ops/filesystems/azure.zh.md
```diff
@@ -38,7 +38,7 @@ wasb://<your-container>@$<your-azure-account>.blob.core.windows.net/<object-path
 wasbs://<your-container>@$<your-azure-account>.blob.core.windows.net/<object-path>
 {% endhighlight %}
 
-Below shows how to use Azure Blob Storage with Flink:
+Below shows how to use Azure Blob Storage in a Flink job:
```
alpinegizmo: same as above

docs/ops/filesystems/index.md
```diff
@@ -35,32 +35,52 @@ File system instances are instantiated once per process and then cached/pooled,
...
-    
-  - **MapR FS**: The MapR file system *"maprfs://"* is automatically available when the MapR libraries are in the classpath.  
-  
-  - **OpenStack Swift FS**: Flink directly provides a file system to talk to the OpenStack Swift file system, registered under the scheme *"swift://"*. 
-  The implementation of `flink-swift-fs-hadoop` is based on the [Hadoop Project](https://hadoop.apache.org/) but is self-contained with no dependency footprint.
-  To use it when using Flink as a library, add the respective maven dependency (`org.apache.flink:flink-swift-fs-hadoop:{{ site.version }}`
-  When starting a Flink application from the Flink binaries, copy or move the respective jar file from the `opt` folder to the `lib` folder.
+Apache Flink project supports the following file systems as plugins:
+
+  - [**Amazon S3**](./s3.html) object storage is supported by two alternative implementations, `flink-s3-fs-presto` and `flink-s3-fs-hadoop`.
```
alpinegizmo: ... is supported by two alternative implementations: `flink-s3-fs-presto` and `flink-s3-fs-hadoop`.

docs/ops/filesystems/index.md
```diff
@@ -35,32 +35,52 @@ File system instances are instantiated once per process and then cached/pooled,
...
+of your Flink distribution before starting Flink, e.g.
 
-## HDFS and Hadoop File System support 
+{% highlight bash %}
+mkdir ./plugins/s3-fs-hadoop
+cp ./opt/flink-s3-fs-hadoop-{{ site.version }}.jar ./plugins/s3-fs-hadoop/
+{% endhighlight %}
+
+<span class="label label-danger">Attention</span> The plugin mechanism for file systems was introduced in Flink version `1.9` to
+support dedicated Java class loaders per plugin and move away from class shading mechanism.
```
alpinegizmo: and to move away from the class shading mechanism.

docs/ops/filesystems/index.md
```diff
@@ -35,32 +35,52 @@ File system instances are instantiated once per process and then cached/pooled,
...
+You can still use the provided file systems (or your own implementations) via the old mechanism by copying the corresponding
+JAR file into `lib` directory.
+
+It's encouraged to use the plugin-based loading mechanism for file systems that support it. Loading file systems components from `lib`
```
alpinegizmo: Loading file system components from the `lib` directory ...

docs/ops/filesystems/s3.md
```diff
@@ -59,23 +59,24 @@ For some cases, however, e.g., for using S3 as YARN's resource storage dir, it m
...
-wrappers for URIs with the `s3://` scheme, `flink-s3-fs-hadoop` also registers
-for `s3a://` and `flink-s3-fs-presto` also registers for `s3p://`, so you can
+wrappers for URIs with the *s3://* scheme, `flink-s3-fs-hadoop` also registers
+for *s3a://* and `flink-s3-fs-presto` also registers for *s3p://*, so you can
 use this to use both at the same time.
 For example, the job uses the [StreamingFileSink]({{ site.baseurl}}/dev/connectors/streamfile_sink.html) which only supports Hadoop, but uses Presto for checkpointing.
-In this case, it is advised to use explicitly *"s3a://"* as a scheme for the sink (Hadoop) and *"s3p://"* for checkpointing (Presto).
-    
-To use either `flink-s3-fs-hadoop` or `flink-s3-fs-presto`, copy the respective JAR file from the `opt` directory to the `lib` directory of your Flink distribution before starting Flink, e.g.
+In this case, it is advised to use explicitly *s3a://* as a scheme for the sink (Hadoop) and *s3p://* for checkpointing (Presto).
```
alpinegizmo: advised to explicitly use

docs/monitoring/debugging_classloading.md
1u0: I've split this sentence into two and rephrased in a way that it's more similar to what you are proposing.

docs/ops/filesystems/index.md
```diff
@@ -83,17 +103,21 @@ fs.hdfs.hadoopconf: /path/to/etc/hadoop
...
 
 To add a new file system:
 
   - Add the File System implementation, which is a subclass of `org.apache.flink.core.fs.FileSystem`.
   - Add a factory that instantiates that file system and declares the scheme under which the FileSystem is registered. This must be a subclass of `org.apache.flink.core.fs.FileSystemFactory`.
-  - Add a service entry. Create a file `META-INF/services/org.apache.flink.core.fs.FileSystemFactory` which contains the class name of your file system factory class.
+  - Add a service entry. Create a file `META-INF/services/org.apache.flink.core.fs.FileSystemFactory` which contains the class name of your file system factory class
+  (see the [Java Service Loader docs](https://docs.oracle.com/javase/8/docs/api/java/util/ServiceLoader.html) for more details).
+
+During plugins discovery, the file system factory class would be loaded by a dedicated Java class loader to avoid class conflicts with other plugins and Flink components.
```
zentol:
```suggestion
During plugins discovery, the file system factory class will be loaded by a dedicated Java class loader to avoid class conflicts with other plugins and Flink components.
```

docs/ops/filesystems/s3.zh.md
```diff
@@ -59,23 +59,24 @@ For some cases, however, e.g., for using S3 as YARN's resource storage dir, it m
...
-In this case, it is advised to use explicitly *"s3a://"* as a scheme for the sink (Hadoop) and *"s3p://"* for checkpointing (Presto).
-    
-To use either `flink-s3-fs-hadoop` or `flink-s3-fs-presto`, copy the respective JAR file from the `opt` directory to the `lib` directory of your Flink distribution before starting Flink, e.g.
+In this case, it is advised to explicitly use *s3a://* as a scheme for the sink (Hadoop) and *s3p://* for checkpointing (Presto).
+
+To use `flink-s3-fs-hadoop` or `flink-s3-fs-presto`, copy the respective JAR file from the `opt` directory to the `plugins` directory of your Flink distribution before starting Flink, e.g.
 
 {% highlight bash %}
-cp ./opt/flink-s3-fs-presto-{{ site.version }}.jar ./lib/
+mkdir ./plugins/s3-fs-presto
```
zentol: will the plugin loading fail if this directory is not present?

docs/ops/filesystems/index.md
```diff
@@ -35,32 +35,52 @@ File system instances are instantiated once per process and then cached/pooled,
...
-  - **OpenStack Swift FS**: Flink directly provides a file system to talk to the OpenStack Swift file system, registered under the scheme *"swift://"*. 
-  The implementation of `flink-swift-fs-hadoop` is based on the [Hadoop Project](https://hadoop.apache.org/) but is self-contained with no dependency footprint.
-  To use it when using Flink as a library, add the respective maven dependency (`org.apache.flink:flink-swift-fs-hadoop:{{ site.version }}`
-  When starting a Flink application from the Flink binaries, copy or move the respective jar file from the `opt` folder to the `lib` folder.
+Apache Flink project supports the following file systems as plugins:
+
+  - [**Amazon S3**](./s3.html) object storage is supported by two alternative implementations: `flink-s3-fs-presto` and `flink-s3-fs-hadoop`.
+  Both implementations are self-contained with no dependency footprint.
+
+  - **MapR FS** file system adapter is already supported in the main Flink distribution under the *maprfs://* URI scheme.
```
zentol: Do you still intend to re-structure the page to make the difference clearer for MapR? All of the benefits of plugins don't apply to the MapR filesystem.

docs/ops/filesystems/index.md
```diff
@@ -35,32 +35,52 @@ File system instances are instantiated once per process and then cached/pooled,
...
+cp ./opt/flink-s3-fs-hadoop-{{ site.version }}.jar ./plugins/s3-fs-hadoop/
+{% endhighlight %}
+
+<span class="label label-danger">Attention</span> The plugin mechanism for file systems was introduced in Flink version `1.9` to
+support dedicated Java class loaders per plugin and to move away from the class shading mechanism.
+You can still use the provided file systems (or your own implementations) via the old mechanism by copying the corresponding
+JAR file into `lib` directory.
+
+It's encouraged to use the plugin-based loading mechanism for file systems that support it. Loading file systems components from the `lib`
+directory may be not supported in the future.
```
zentol:
```suggestion
directory may not be supported in future versions.
```

docs/ops/filesystems/s3.zh.md
1u0: The name of the containing folder doesn't matter much. The only convention, that it should be in Flink's `/plugins` dir one folder deep, one folder per plugin, afair.

docs/ops/filesystems/index.md
1u0: I've applied the second part: `... supported in future Flink versions`.
Using `may not` is not suitable here, as it means prohibiting something.

docs/ops/filesystems/index.md
1u0: I've left some note in the document, that the MapR filesystem is exception compared to others.
There is also a discussion happening in the Flink dev mailing list to remove this FS from Flink distribution (jar).

docs/ops/filesystems/index.md
```diff
@@ -83,17 +105,21 @@ fs.hdfs.hadoopconf: /path/to/etc/hadoop
...
-  - Add a service entry. Create a file `META-INF/services/org.apache.flink.core.fs.FileSystemFactory` which contains the class name of your file system factory class.
+  - Add a service entry. Create a file `META-INF/services/org.apache.flink.core.fs.FileSystemFactory` which contains the class name of your file system factory class
+  (see the [Java Service Loader docs](https://docs.oracle.com/javase/8/docs/api/java/util/ServiceLoader.html) for more details).
+
+During plugins discovery, the file system factory class will be loaded by a dedicated Java class loader to avoid class conflicts with other plugins and Flink components.
+The same class loader should be used during file system instantiation and the file system operation calls.
 
-See the [Java Service Loader docs](https://docs.oracle.com/javase/8/docs/api/java/util/ServiceLoader.html) for more details on how service loaders work.
+<span class="label label-warning">Warning</span> In practice, it means you should avoid using `Thread.currentThread().getContextClassLoader()` class loader
+in your implementation.
```
pnowojski: Do we want to elaborate on this? That `getContextClassLoader()` can be used inside `org.apache.flink.core.fs.FileSystemFactory#create(URI)`  call stack (however we are discouraging from using it anyway)?

docs/monitoring/debugging_classloading.md
```diff
@@ -28,17 +28,20 @@ under the License.
...
   - The **Dynamic User Code**: These are all classes that are included in the JAR files of dynamically submitted jobs,
     (via REST, CLI, web UI). They are loaded (and unloaded) dynamically per job.
 
-What classes are part of which domain depends on the particular setup in which you run Apache Flink. As a general rule, whenever you start the Flink
-processes first, and submit jobs, the job's classes are loaded dynamically. If the Flink processes are started together with the job/application,
-or the application spawns the Flink components (JobManager, TaskManager, etc.) then all classes are in the Java classpath.
+As a general rule, whenever you start the Flink processes first and submit jobs later, the job's classes are loaded dynamically.
+If the Flink processes are started together with the job/application, or if the application spawns the Flink components (JobManager, TaskManager, etc.), then all job's classes are in the Java classpath.
+
+Code in plugin components is loaded dynamically once by a dedicated class loader per plugin.
```
pnowojski: nit: Maybe we could make this claim a bit stronger?

> Code in plugin components is loaded dynamically once by a dedicated class loader per plugin, 
> independent and completely separate of other class loaders. In other words, plugin's classes
> are not visible outside of that plugin's context.

docs/ops/filesystems/index.md
1u0: I'd prefer not to document it.
Currently the `create()` call has the class loader set only because there are some existing implementation that rely on it.
So far there is no proper reason why would FSs need it, they can explicitly use the correct class loader.

In the future, it should be possible to remove the additional setup of class loader for `create()` call.

docs/monitoring/debugging_classloading.md
1u0: I'd avoid giving such claim.
Currently, a dedicated class loader per plugin provide common java class loading isolation and not more.

It's assumed that a reader is familiar with java class loading. Imo, extending this sentence as you propose raises more questions from the reader.

PR #9172, opened 2019-07-18
[FLINK-13313][table] create CatalogTableBuilder to support building CatalogTable from descriptors

bowenli86:

This PR adds `CatalogTableBuilder` as a replacement of `ExternalCatalogTableBuilder` to help users convert table source/sink descriptors to CatalogTable. The gap was mainly discovered when I  was writing tests for `HiveCatalog` to make sure it works as expected to persist Flink generic tables

## Brief change log

- added `CatalogTableBuilder` as a replacement of `ExternalCatalogTableBuilder` to help users convert table source/sink descriptors to CatalogTable
- added unit test `HiveCatalogITCase` for two things
  - test `HiveCatalog` work as expected to persist Flink generic tables
  - test `CatalogTableBuilder` work as expect 
  - `HiveCatalogITCase` should be moved to `flink-connector-hive-test` module once it's setup

## Verifying this change

This change added tests and can be verified in `HiveCatalogITCase`

  - The serializers: (no)

## Documentation

  - If yes, how is the feature documented? (docs / JavaDocs)

Docs will be added separately


bowenli86:
cc @twalthr @xuefuz @lirui-apache @zjuwangg

flink-table/flink-table-planner/src/main/scala/org/apache/flink/table/catalog/ExternalCatalogTableBuilder.scala
```diff
@@ -52,7 +52,10 @@ import org.apache.flink.table.descriptors._
   * }}}
   *
   * @param connectorDescriptor Connector descriptor describing the external system
+  *
+  * @deprecated use [[CatalogTableBuilder]]
   */
+@Deprecated
```
xuefuz: Do we need two annos for deprecation?

bowenli86: `@Deprecated` is for java, `@deprecated` is for scala

bowenli86:
re: 2, 3 > good idea. Will add
re: 1 > I'm concerned about renaming since we don't have `GenericCatalogTable` and thus `GenericCatalogTableBuilder` would confuse users. To most Flink users who are used to original way to create Flink source/sink with descriptors, they don't need to understand difference of whether a catalog table is generic or not. They just need an equivalent of `ExternalCatalogTableBuilder`.

xuefuz:
Keeping the name unchanged is fine to me. However, I don't expect that we will use this builder to build Hive CatalogTable. That builder probably will reside in Hive connector package rather than here. Therefore, I don't think we even need to have a member variable "isGeneric" in CatalogTableBuilder. Putting it here might give a misconception. 

This is a minor issue, though. Otherwise, the PR looks good to me.

bowenli86:
> Keeping the name unchanged is fine to me. However, I don't expect that we will use this builder to build Hive CatalogTable. That builder probably will reside in Hive connector package rather than here. Therefore, I don't think we even need to have a member variable "isGeneric" in CatalogTableBuilder. Putting it here might give a misconception.

It can be used with any catalog for any source/sink that have a descriptor, like csv, kafka, elasticsearch, etc. Why would it reside in Hive connector?

xuefuz:
To clarify:
This = the builder created here.
That = the possible Hive table builder in the future.

bowenli86:
> To clarify:
> This = the builder created here.
> That = the possible Hive table builder in the future.

Got you, thanks for the clarification!

> Therefore, I don't think we even need to have a member variable "isGeneric" in CatalogTableBuilder. Putting it here might give a misconception.

I will think about it

PR #9174, opened 2019-07-19
[FLINK-13330][table-planner-blink] Remove unnecessary to reduce testing time in blink

JingsongLi:

Some tests are unnecessary.

## Verifying this change

  - The serializers: no

## Documentation

### DIFF
--- a/flink-table/flink-table-planner-blink/src/test/scala/org/apache/flink/table/runtime/utils/StreamingWithAggTestBase.scala
+++ b/flink-table/flink-table-planner-blink/src/test/scala/org/apache/flink/table/runtime/utils/StreamingWithAggTestBase.scala
@@ -63,9 +63,11 @@ object StreamingWithAggTestBase {
 
   @Parameterized.Parameters(name = "LocalGlobal={0}, {1}, StateBackend={2}")
   def parameters(): util.Collection[Array[java.lang.Object]] = {
+    // To avoid too long testing time, we not test all HEAP_BACKEND.
+    // If HEAP_BACKEND has changed, please open it manually for testing.
     Seq[Array[AnyRef]](
-      Array(LocalGlobalOff, MiniBatchOff, HEAP_BACKEND),
-      Array(LocalGlobalOff, MiniBatchOn, HEAP_BACKEND),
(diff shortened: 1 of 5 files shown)


flink-table/flink-table-planner-blink/src/test/scala/org/apache/flink/table/runtime/utils/StreamingWithAggTestBase.scala
```diff
@@ -63,9 +63,11 @@ object StreamingWithAggTestBase {
 
   @Parameterized.Parameters(name = "LocalGlobal={0}, {1}, StateBackend={2}")
   def parameters(): util.Collection[Array[java.lang.Object]] = {
+    // To avoid too long testing time, we not test all HEAP_BACKEND.
+    // If HEAP_BACKEND has changed, please open it manually for testing.
```
twalthr: Comments like this get lost in the code base. Tests like this should be moved to a special manual tests package.

flink-table/flink-table-runtime-blink/src/test/java/org/apache/flink/table/runtime/sort/BinaryExternalSorterTest.java
```diff
@@ -76,11 +75,14 @@ public BinaryExternalSorterTest(
...
-				new Boolean[]{false, true},
-				new Boolean[]{true, false},
-				new Boolean[]{true, true});
+		// To avoid too long testing time, we only tested the default configuration.
+		// If BinaryExternalSorter has changed, please open it manually for testing.
+//		return Arrays.asList(
+//				new Boolean[]{false, false},
+//				new Boolean[]{false, true},
+//				new Boolean[]{true, false},
+//				new Boolean[]{true, true});
```
wuchong: Manually testing is hard to guarantee the correctness. I think we need to find a way to add this to nightly tests in this pull request.

JingsongLi:
Thanks @twalthr @wuchong , this tests should move to crontab and daily run. Every tests is important and precious. If we remove them, there maybe some bugs in future.

JingsongLi:
I close this now, @wuchong will do some work to extract them to daily run.

PR #9175, opened 2019-07-19
[FLINK-12038] [test] fix YARNITCase random fail

shuai-xu:

This pr fix that the YARNITCase may random fail due to [YARN-2853](https://issues.apache.org/jira/browse/YARN-2853). 

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-yarn-tests/src/test/java/org/apache/flink/yarn/YARNITCase.java
+++ b/flink-yarn-tests/src/test/java/org/apache/flink/yarn/YARNITCase.java
@@ -117,10 +117,6 @@ public class YARNITCase extends YarnTestBase {
 					if (clusterClient != null) {
 						clusterClient.shutdown();
 					}
-
-					if (applicationId != null) {
-						yarnClusterDescriptor.killCluster(applicationId);
-					}
 				}
 			}
 		});
--- a/flink-yarn-tests/src/test/java/org/apache/flink/yarn/YARNITCase.java
+++ b/flink-yarn-tests/src/test/java/org/apache/flink/yarn/YARNITCase.java
@@ -18,6 +18,7 @@
 
 package org.apache.flink.yarn;
 
+import org.apache.flink.api.common.time.Deadline;
 import org.apache.flink.client.deployment.ClusterSpecification;
 import org.apache.flink.client.program.ClusterClient;
 import org.apache.flink.client.program.rest.RestClusterClient;
@@ -32,7 +33,9 @@ import org.apache.flink.yarn.util.YarnTestUtils;
 
 import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.yarn.api.records.ApplicationId;
+import org.apache.hadoop.yarn.api.records.YarnApplicationState;
 import org.apache.hadoop.yarn.client.api.YarnClient;
+import org.junit.Assert;
 import org.junit.BeforeClass;
 import org.junit.Test;
 
@@ -52,6 +55,8 @@ import static org.junit.Assert.assertThat;
 public class YARNITCase extends YarnTestBase {
 
 	private final Duration yarnAppTerminateTimeout = Duration.ofSeconds(10);
+
+	private final int sleepIntervalInMS = 100;
 
 	@BeforeClass
 	public static void setup() {
@@ -117,17 +122,35 @@ public class YARNITCase extends YarnTestBase {
 					assertThat(jobResult, is(notNullValue()));
 					assertThat(jobResult.getSerializedThrowable().isPresent(), is(false));
 
-					waitUntilApplicationFinished(applicationId, yarnAppTerminateTimeout);
+					waitApplicationFinishedElseKillIt(applicationId, yarnAppTerminateTimeout, yarnClusterDescriptor);
 				} finally {
 					if (clusterClient != null) {
 						clusterClient.shutdown();
-					}
-
-					if (applicationId != null) {
-						yarnClusterDescriptor.killCluster(applicationId);
 					}
 				}
 			}
 		});
 	}
+
+	private void waitApplicationFinishedElseKillIt(
+			ApplicationId applicationId,
+			Duration timeout,
+			YarnClusterDescriptor yarnClusterDescriptor) throws Exception {
+		Deadline deadline = Deadline.now().plus(timeout);
+		YarnApplicationState state = yarnClient.getApplicationReport(applicationId).getYarnApplicationState();
+
+		while (state != YarnApplicationState.FINISHED) {
+			if (state == YarnApplicationState.FAILED || state == YarnApplicationState.KILLED) {
+				Assert.fail("Application became FAILED or KILLED while expecting FINISHED");
+			} else {
+				sleep(sleepIntervalInMS);
+			}
+			if (deadline.isOverdue()) {
+				yarnClusterDescriptor.killCluster(applicationId);
+				Assert.fail("Application didn't finish before timeout");
+			}
+			state = yarnClient.getApplicationReport(applicationId).getYarnApplicationState();
+		}
+	}
+
 }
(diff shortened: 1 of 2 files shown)


flink-yarn-tests/src/test/java/org/apache/flink/yarn/YARNITCase.java
```diff
@@ -113,6 +113,8 @@ public void testPerJobMode() throws Exception {
 
 					assertThat(jobResult, is(notNullValue()));
 					assertThat(jobResult.getSerializedThrowable().isPresent(), is(false));
+
+					waitUntilApplicationFinished(applicationId, 10);
```
azagrebin: Could you move `10` to a named private static final constant?

flink-yarn-tests/src/test/java/org/apache/flink/yarn/YarnTestBase.java
```diff
@@ -573,6 +574,20 @@ public static int getRunningContainers() {
 		return count;
 	}
 
+	public void waitUntilApplicationFinished(ApplicationId applicationId, int timeout) throws Exception {
```
azagrebin: I would suggest to rename `timeout` to `timeoutSec` or better make it of `Duration` type instead of `int`.
Then also a private static final constant can be of the `Duration` type in the previous comment.

flink-yarn-tests/src/test/java/org/apache/flink/yarn/YarnTestBase.java
```diff
@@ -573,6 +574,20 @@ public static int getRunningContainers() {
...
+		Deadline deadline = Deadline.now().plus(Duration.ofSeconds(timeout));
+		while (deadline.hasTimeLeft()) {
+			YarnApplicationState state = yarnClient.getApplicationReport(applicationId).getYarnApplicationState();
+			if (state == YarnApplicationState.FINISHED ||
+					state == YarnApplicationState.FAILED ||
+					state == YarnApplicationState.KILLED) {
+				break;
+			} else {
+				sleep(500);
```
azagrebin: also a private static final constant_milli
I think 100 ms is actually enough to wait in most of normal cases.

flink-yarn-tests/src/test/java/org/apache/flink/yarn/YarnTestBase.java
azagrebin: the method can be also package private

shuai-xu: OK, we could make it public when tests in other package need it.

flink-yarn-tests/src/test/java/org/apache/flink/yarn/YARNITCase.java
shuai-xu: Done

flink-yarn-tests/src/test/java/org/apache/flink/yarn/YarnTestBase.java
shuai-xu: OK

flink-yarn-tests/src/test/java/org/apache/flink/yarn/YarnTestBase.java
```diff
@@ -573,6 +576,23 @@ public static int getRunningContainers() {
 		return count;
 	}
 
+	void waitUntilApplicationFinished(ApplicationId applicationId, Duration timeout) throws Exception {
+		Deadline deadline = Deadline.now().plus(timeout);
+		while (true) {
```
azagrebin: I would prefer
```
YarnApplicationState state;
do {
   state = yarnClient.getApplicationReport(applicationId).getYarnApplicationState();
   .....
} while (state != YarnApplicationState.FINISHED); 
```

flink-yarn-tests/src/test/java/org/apache/flink/yarn/YarnTestBase.java
```diff
@@ -573,6 +576,23 @@ public static int getRunningContainers() {
...
+			YarnApplicationState state = yarnClient.getApplicationReport(applicationId).getYarnApplicationState();
+			if (state == YarnApplicationState.FINISHED) {
+				break;
+			} else if (state == YarnApplicationState.FAILED || state == YarnApplicationState.KILLED) {
+				Assert.fail("Application became FAILED or KILLED while expecting FINISHED");
+			} else {
+				sleep(sleepIntervalInMS);
+			}
+			if (deadline.isOverdue()) {
+				Assert.fail("Application didn't finish before timeout");
```
azagrebin: I think we should move the app killing from the finally block of `YARNITCase. waitUntilApplicationFinished` to this timeout case, right before `Assert.fail("Application didn't finish before timeout");`. As we need killing only if the app hangs but we do not need if it finished successfully.

shuai-xu: I put the `waitUntilApplicationFinished` into YarnITCase as it is not so universal for other case now.

flink-yarn-tests/src/test/java/org/apache/flink/yarn/YarnTestBase.java
shuai-xu: This may lead to unnecessary `sleep`, I change it to `while (state != YarnApplicationState.FINISHED) {}`

flink-yarn-tests/src/test/java/org/apache/flink/yarn/YarnTestBase.java
```diff
@@ -163,7 +160,7 @@
 
 	protected static File yarnSiteXML = null;
 
-	private YarnClient yarnClient = null;
+	protected YarnClient yarnClient = null;
```
azagrebin: We already have a getter for the `yarnClient` which we use in the `YARNITCase.testPerJobMode`.
The same `getYarnClient` can be used in `waitApplicationFinishedElseKillIt` then there is no need for this `protected` change.

flink-yarn-tests/src/test/java/org/apache/flink/yarn/YARNITCase.java
```diff
@@ -117,17 +122,35 @@ public void testPerJobMode() throws Exception {
...
+			if (state == YarnApplicationState.FAILED || state == YarnApplicationState.KILLED) {
+				Assert.fail("Application became FAILED or KILLED while expecting FINISHED");
+			} else {
+				sleep(sleepIntervalInMS);
+			}
+			if (deadline.isOverdue()) {
+				yarnClusterDescriptor.killCluster(applicationId);
+				Assert.fail("Application didn't finish before timeout");
+			}
+			state = yarnClient.getApplicationReport(applicationId).getYarnApplicationState();
```
azagrebin: nit: I would put `sleep(sleepIntervalInMS);` before `state = ...` so that it does not get lost accidentally in if..else branching.

flink-yarn-tests/src/test/java/org/apache/flink/yarn/YarnTestBase.java
azagrebin: agreed

flink-yarn-tests/src/test/java/org/apache/flink/yarn/YarnTestBase.java
azagrebin: agreed

flink-yarn-tests/src/test/java/org/apache/flink/yarn/YarnTestBase.java
shuai-xu: Done, I didn't see it.

flink-yarn-tests/src/test/java/org/apache/flink/yarn/YARNITCase.java
shuai-xu: OK

PR #2003, opened 2016-05-18
[FLINK-3909] update Maven Failsafe version

mxm:
Failures during execution of the integration tests with the Maven
Failsafe plugin were silently ignored on Travis with Maven version
3.2.5. The problem is that failures are not passed on correctly from the
'integration-test' phase (where failures are recorded and tolerated) to
the 'verify' phase.

The cause of the error is most likely SUREFIRE-1127. An exception in the
'integration-test' is sometimes not flushed back to disk where it is
evaluated in the 'verify' phase. Bumping the version of the Failsafe
plugin from 2.18.1 to 2.19.1 fixes the issue.

Tests will fail for this PR ;)

PR #9180, opened 2019-07-19
[FLINK-13331][table-planner-blink] Add CachedMiniCluster to share cluster between ITCases

JingsongLi:

Add CachedMiniClusterWithClientResource to share cluster between ITCases.
This should reduce a lot of test time.

## Verifying this change

  - The serializers: no

## Documentation

### DIFF
--- a/flink-table/flink-table-planner-blink/pom.xml
+++ b/flink-table/flink-table-planner-blink/pom.xml
@@ -388,6 +388,27 @@ under the License.
 					<configLocation>${project.basedir}/../../tools/maven/scalastyle-config.xml</configLocation>
 				</configuration>
 			</plugin>
+
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-surefire-plugin</artifactId>
+				<executions>
+					<execution>
+						<id>integration-tests</id>
+						<phase>integration-test</phase>
+						<goals>
+							<goal>test</goal>
+						</goals>
+						<configuration>
+							<includes>
+								<include>**/*ITCase.*</include>
+							</includes>
+							<!-- Accelerate testing -->
+							<reuseForks>true</reuseForks>
+						</configuration>
+					</execution>
+				</executions>
+			</plugin>
 		</plugins>
 	</build>
 </project>
--- a/flink-table/flink-table-planner-blink/src/test/java/org/apache/flink/table/runtime/utils/CachedMiniClusterResource.java
+++ b/flink-table/flink-table-planner-blink/src/test/java/org/apache/flink/table/runtime/utils/CachedMiniClusterResource.java
@@ -1,0 +1,122 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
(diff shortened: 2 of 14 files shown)


JingsongLi:
We need do some work to override false reuseForks from https://github.com/apache/flink/pull/2003

wuchong:
I find a potential problem of this approach. After running blink planner IT cases, the cluster resources 
 are still being hold. It may leaks resources (threads, memory) and affect following tests?

zentol:
This approach will always leak resources as @wuchong pointed out.

I have doubts whether this will improve the test times in a significant way, since we're only skipping the creation of ~60 mini clusters.

Nevertheless please also provide actual numbers when doing optimizations such as this; i.e., how long did the tests take beforehand, how long do they take now, so we can gauge how valuable this change is.

JingsongLi:
@wuchong @zentol Thanks for your attention. 

About leak resources, there is a clean thread to invoke `MiniClusterWithClientResource.after` to clean resources. Compared with the past, it only delayed cleaning up resources by 15 seconds.

>I have doubts whether this will improve the test times in a significant way, since we're only skipping the creation of ~60 mini clusters.

Yeah, we should gauge how valuable this change is.
1.I test on my macbook pro, blink-planner will use 17min before, and use 11min now for `mvn clean verify`. If we just run `mvn verify`, the gap will be more pronounced.
2.But the more accurate test time should be travis triggering, I will update later.

JingsongLi:
Now blink-planner test time is 31-32min: https://travis-ci.org/JingsongLi/flink/jobs/562395265
Without this PR, test time is almost 50min: https://travis-ci.com/flink-ci/flink/jobs/218419037
@zentol @wuchong FYI.

zentol:
Based on the test run in #9306 it appears that the speed-up is mostly caused by the reuse of forks, so we may just go with that without the caching of clusters.

wuchong:
> Based on the test run in #9306 it appears that the speed-up is mostly caused by the reuse of forks, so we may just go with that without the caching of clusters.

Do you know any reason why we disabled reuse of forks before?

wuchong:
The testing time of blink planner is 30min if we enable reuse forks. So I'm +1 with @zentol 's opinion.

zentol:
Not re-using forks is _literally_ the only difference between unit tests and IT cases, and I would presume we did that to ensure they are fully isolated from another.

In other words, instead of messing with the reuseForks setting, we could just label them all as unit tests.

wuchong:
However, IMO, integration-test category is helpful. For example, the `*ITCase` suffix is helpful sometimes because they are easy to distinguish with unit tests. And we are still planning to reduce ITCase time in blink planner in some other ways. 

If isolation is the purpose, I think we can enable it for blink planner tests, because we have an individual stage for blink planner ?

zentol:
Well in any case we would be enabling it for the blink planner; be it by enabling fork reuse of marking them as unit tests.

So yes, we can just set it to true in the blink planner.

wuchong:
Thanks @zentol . 

@JingsongLi could you prepare another pull request with only enabling reuse forks for blink planner?

JingsongLi:
Maybe there are some stories why start cluster and shutdown cluster so quick in travis when enable reuse forks. In my local IDE, it still very slow.
But I am +1 to speed up testing to blink-planner by simple code modify.
Maybe we can just use #9306

wuchong:
#9306 has been closed. I created a PR #9326 to continue the work. 

Closing this PR.

PR #9136, opened 2019-07-16
[FLINK-12768][tests] FlinkKinesisConsumerTest.testSourceSynchronization flakiness

tweise:
PR for investigation only.

PR #9183, opened 2019-07-19
[FLINK-12768][tests] FlinkKinesisConsumerTest.testSourceSynchronization flakiness

tweise:
These are the net changes from https://github.com/apache/flink/pull/9136

I propose we merge these and observe if it removes the flakiness in Travis.

## Brief change log

*(for example:)*

## Verifying this change

*(or)*

*(or)*

*(example:)*

## Documentation

### DIFF
--- a/flink-connectors/flink-connector-kinesis/src/test/java/org/apache/flink/streaming/connectors/kinesis/FlinkKinesisConsumerTest.java
+++ b/flink-connectors/flink-connector-kinesis/src/test/java/org/apache/flink/streaming/connectors/kinesis/FlinkKinesisConsumerTest.java
@@ -23,6 +23,7 @@ import org.apache.flink.api.common.serialization.SimpleStringSchema;
 import org.apache.flink.api.common.state.ListState;
 import org.apache.flink.api.common.state.ListStateDescriptor;
 import org.apache.flink.api.common.state.OperatorStateStore;
+import org.apache.flink.api.common.time.Deadline;
 import org.apache.flink.api.common.typeinfo.TypeInformation;
 import org.apache.flink.api.java.tuple.Tuple2;
 import org.apache.flink.api.java.typeutils.runtime.PojoSerializer;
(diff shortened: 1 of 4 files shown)


flink-connectors/flink-connector-kinesis/src/test/java/org/apache/flink/streaming/connectors/kinesis/FlinkKinesisConsumerTest.java
```diff
@@ -952,27 +965,33 @@ public void emitWatermark(Watermark mark) {
...
-		TestWatermarkTracker.assertSingleWatermark(-4);
+		TestWatermarkTracker.assertGlobalWatermark(-4);
 
 		final long record2 = record1 + (watermarkSyncInterval * 3) + 1;
 		shard1.put(Long.toString(record2));
 
-		// TODO: check for record received instead
-		Thread.sleep(100);
+		// wait for the record to be buffered in the emitter
+		final RecordEmitter<?> emitter = org.powermock.reflect.Whitebox.getInternalState(fetcher, "recordEmitter");
```
tillrohrmann: Using `Whitebox` usually indicates that the class under test is not well suited for testing. Maybe it can be refactored.

tweise:
The records and watermarks are emitted from separate threads (the thread that emits the records will indicate what the next candidate watermark is). The logic in this test relies on the watermark for the last record being emitted when advancing the processing time. Therefore it is important that  record emit and watermark state change occur atomically. Otherwise the context switch can occur after the record emit and the watermark is generated using previous state, which lead to the flakiness. All this only matters for this test. In a job watermarks are emitted periodically and it only matters that the record is emitted prior to the corresponding watermark state change.

flink-connectors/flink-connector-kinesis/src/test/java/org/apache/flink/streaming/connectors/kinesis/FlinkKinesisConsumerTest.java
tweise: Yes, that's more of a philosophical discussion outside of the issue at hand. Whitebox avoids extra code to be added to the class under test purely for the purpose of assertions in the test case. The other side of the argument is usually that tests can break more easily because there isn't a compile time check for such access. I think using Whitebox within the same module is fine because one would always run the tests while making potentially breaking modifications.

PR #9184, opened 2019-07-20
[FLINK-13339][ml] Add an implementation of pipeline's api

xuyang1706:

Add an implementation of pipeline's api

## Brief change log

  - *Add an implement PipelineStage, Estimator, Transformer, Model.*
  - *Add MLSession to hold the execution environment and others session shared variable.*
  - *Add AlgoOperator for the implementation of algorithms.*
  - *Add BatchOperator and StreamOperator based on AlgoOperator*
  - *Add TableSourceBatchOp and TableSourceStreamOp *

## Verifying this change

- run test case pass

  - The serializers: (no)

## Documentation


flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/MLSession.java
```diff
@@ -0,0 +1,187 @@
...
+import org.apache.flink.table.api.TableEnvironment;
+import org.apache.flink.table.api.java.BatchTableEnvironment;
+import org.apache.flink.table.api.java.StreamTableEnvironment;
+import org.apache.flink.table.api.java.internal.StreamTableEnvironmentImpl;
+import org.apache.flink.types.Row;
+
+/**
+ * MLSession hold the execution environment and others session shared variable.
+ *
+ * <P>In machine learning, we find that a single execution context is convenient for the user
```
becketqin: What is the scope of an MLSession? Do we allow multiple MLSessions in the same JVM? If so, how are we going to support that?

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/pipeline/Model.java
```diff
@@ -0,0 +1,56 @@
...
+
+import org.apache.flink.ml.api.misc.param.Params;
+import org.apache.flink.table.api.Table;
+
+/**
+ * Abstract class for a machine learning model.
+ *
+ * @param <M> The class type of the {@link Model} implementation itself
+ */
+public abstract class Model<M extends Model <M>> extends Transformer <M>
```
becketqin: Same here, can we change the class name to `ModelBase`?

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/pipeline/PipelineStage.java
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * <p>Each pipeline stage is with parameters, and requires a public empty constructor for
+ * restoration in Pipeline. It implement the {@link org.apache.flink.ml.api.core.PipelineStage}
+ * and hold a {@link Params} as its member, thus the subclasses could do not care about
+ * {@link WithParams#getParams()}
+ *
+ * @param <S> The class type of the {@link PipelineStage} implementation itself, used by {@link
+ *            org.apache.flink.ml.api.misc.param.WithParams} and Cloneable.
+ */
+public abstract class PipelineStage<S extends PipelineStage <S>>
```
becketqin: Same here, `PipelineStageBase`? This class does not implement the `org.apache.flink.ml.api.core.PipelineStage`, is this intended?

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/pipeline/Transformer.java
```diff
@@ -0,0 +1,100 @@
...
+
+/**
+ * Abstract class for a transformer that transform one data into another.
+ * A transformer is a {@link PipelineStage} that transforms an input {@link Table} to a result
+ * {@link Table}.
+ *
+ * @param <T> The class type of the {@link Transformer} implementation itself, used by {@link
+ *            org.apache.flink.ml.api.misc.param.WithParams}
+ */
+public abstract class Transformer<T extends Transformer <T>>
```
becketqin: Same here, `TransformerBase`?

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/streamoperator/StreamOperator.java
```diff
@@ -0,0 +1,86 @@
...
+import org.apache.flink.ml.streamoperator.source.TableSourceStreamOp;
+import org.apache.flink.table.api.Table;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Base class of streaming algorithm operators.
+ */
+public abstract class StreamOperator<T extends StreamOperator <T>> extends AlgoOperator<T> {
```
becketqin: The name `StreamOperator` is colliding with `org.apache.flink.streaming.api.operators.StreamOperator` which is a widely established class in Flink. Can we change it to something like `StreamAlgoOperator`? We may need to change the `BatchOperator` to `BatchAlgoOperator` as well.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/streamoperator/StreamOperator.java
```diff
@@ -0,0 +1,86 @@
...
+	public static StreamOperator sourceFrom(Table table) {
+		return new TableSourceStreamOp(table);
+	}
+
+	@Override
+	public String toString() {
+		return getOutput().toString();
+	}
+
+	public <S extends StreamOperator> S link(S next) {
```
becketqin: I am curious why `link()` and `linkTo()` methods require a template type while `linkFrom()` does not need that?

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/streamoperator/StreamOperator.java
```diff
@@ -0,0 +1,86 @@
...
+import org.apache.flink.ml.api.misc.param.Params;
+import org.apache.flink.ml.common.AlgoOperator;
+import org.apache.flink.ml.streamoperator.source.TableSourceStreamOp;
+import org.apache.flink.table.api.Table;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Base class of streaming algorithm operators.
```
becketqin: This class seems needs some java doc for the public methods.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/batchoperator/BatchOperator.java
```diff
@@ -0,0 +1,86 @@
...
+		ls.add(in2);
+		ls.add(in3);
+		return linkFrom(ls);
+	}
+
+	public T linkFrom(List <BatchOperator> ins) {
+		if (null != ins && ins.size() == 1) {
+			return linkFrom(ins.get(0));
+		} else {
+			throw new RuntimeException("Not support more than 1 inputs!");
```
ex00: Implementations of `linkFrom(BatchOperator, BatchOperator)`, `linkFrom(BatchOperator, BatchOperator,BatchOperator)`, `linkFrom(List)` looks inconsistent.
In methods for BatchOperators is created list of elements more that 1. but in result is throw exception that count elements should be 1.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/batchoperator/BatchOperator.java
```diff
@@ -0,0 +1,86 @@
...
+	public T linkFrom(BatchOperator in1, BatchOperator in2, BatchOperator in3) {
+		List <BatchOperator> ls = new ArrayList();
+		ls.add(in1);
+		ls.add(in2);
+		ls.add(in3);
+		return linkFrom(ls);
+	}
+
+	public T linkFrom(List <BatchOperator> ins) {
+		if (null != ins && ins.size() == 1) {
```
ex00: what is difference on this case from `linkFrom(BatchOperator)`?

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/batchoperator/BatchOperator.java
```diff
@@ -0,0 +1,80 @@
...
+	public <B extends BatchOperator<?>> B link(B next) {
+		return linkTo(next);
+	}
+
+	public <B extends BatchOperator<?>> B linkTo(B next) {
+		next.linkFrom(this);
+		return next;
+	}
+
+	public abstract T linkFrom(BatchOperator<?>... inputs);
```
walterddr: This is a new abstract API. Please provide JavaDoc

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/batchoperator/BatchOperator.java
```diff
@@ -0,0 +1,80 @@
...
+import org.apache.flink.table.api.Table;
+import org.apache.flink.util.Preconditions;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Base class of batch algorithm operators.
+ */
+public abstract class BatchOperator<T extends BatchOperator<T>> extends AlgoOperator<T> {
```
walterddr: please provide java doc for `link` `linkTo` and `linkFrom`.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/batchoperator/BatchOperator.java
```diff
@@ -0,0 +1,80 @@
...
+	public BatchOperator(Params params) {
+		super(params);
+	}
+
+	public <B extends BatchOperator<?>> B link(B next) {
+		return linkTo(next);
+	}
+
+	public <B extends BatchOperator<?>> B linkTo(B next) {
+		next.linkFrom(this);
```
walterddr: what's the rational of making link in this direction? what if downstream operator calls `linkFrom` twice with the same upstream; or twice with different upstreams?

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/batchoperator/source/TableSourceBatchOp.java
```diff
@@ -0,0 +1,44 @@
...
+		super(null);
+		if (null == table) {
+			throw new RuntimeException();
+		}
+		this.setOutput(table);
+	}
+
+	@Override
+	public TableSourceBatchOp linkFrom(BatchOperator<?>... inputs) {
+		throw new UnsupportedOperationException("Not supported.");
```
walterddr: more concise error message: `"Table source operator should not have any upstream to link from"`

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/pipeline/PipelineStage.java
walterddr: I am not sure this class is a must. I think each type of pipeline stage individually can set its own base class. see: https://stackoverflow.com/a/3599379/11332462

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/batchoperator/source/TableSourceBatchOp.java
xuyang1706: Thanks, this message is more concise.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/batchoperator/BatchOperator.java
xuyang1706: Thanks, added intro and example.
For there is no answer area of the above question, I put the answer here:
PipelineStage only supports single input and single output, it is the basic unit for pipeline. AlgoOperator supports multi-input and multi-output. We’d like to implement the algorithm with  AlgoOperator, and PipelineStage’s fit and transform function can call the AlgoOperator.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/batchoperator/BatchOperator.java
xuyang1706: Thanks, provided.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/pipeline/Estimator.java
```diff
@@ -0,0 +1,99 @@
...
+import org.apache.flink.table.api.internal.TableImpl;
+import org.apache.flink.table.api.java.StreamTableEnvironment;
+
+/**
+ * Abstract class for a estimator that fit a {@link Model}.
+ *
+ * @param <E> The class type of the {@link Estimator} implementation itself
+ * @param <M> class type of the {@link Model} this Estimator produces.
+ */
+public abstract class Estimator<E extends Estimator <E, M>, M extends Model <M>>
```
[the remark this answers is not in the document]
xuyang1706: Thanks, changed to EstimatorBase.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/pipeline/Estimator.java
```diff
@@ -0,0 +1,99 @@
...
+import org.apache.flink.ml.common.MLSession;
+import org.apache.flink.ml.streamoperator.StreamOperator;
+import org.apache.flink.ml.streamoperator.source.TableSourceStreamOp;
+import org.apache.flink.table.api.Table;
+import org.apache.flink.table.api.TableEnvironment;
+import org.apache.flink.table.api.internal.TableImpl;
+import org.apache.flink.table.api.java.StreamTableEnvironment;
+
+/**
+ * Abstract class for a estimator that fit a {@link Model}.
```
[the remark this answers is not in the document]
xuyang1706: Thanks, changed.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/pipeline/Estimator.java
```diff
@@ -0,0 +1,99 @@
...
+		return fit(input);
+	}
+
+	/**
+	 * Train and produce a {@link Model} which fits the records in the given {@link Table}.
+	 *
+	 * @param input the table with records to train the Model.
+	 * @return a model trained to fit on the given Table.
+	 */
+	public M fit(Table input) {
```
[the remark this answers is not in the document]
xuyang1706: It is better to be public. We have refined the JavaDoc.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/pipeline/Estimator.java
```diff
@@ -0,0 +1,99 @@
...
+	/**
+	 * Train and produce a {@link Model} which fits the records in the given {@link Table}.
+	 *
+	 * @param tEnv  the table environment to which the input table is bound.
+	 * @param input the table with records to train the Model.
+	 * @return a model trained to fit on the given Table.
+	 */
+	@Override
+	public M fit(TableEnvironment tEnv, Table input) {
+		MLSession.setTableEnvironment(tEnv, input);
```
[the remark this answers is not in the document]
xuyang1706: Thanks for your advice and offline discussion, we defined MLEnvironmentFactory to replace the global static MLSession.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/AlgoOperator.java
```diff
@@ -0,0 +1,98 @@
...
+import org.apache.flink.table.api.internal.TableImpl;
+import org.apache.flink.table.api.java.StreamTableEnvironment;
+
+import java.io.Serializable;
+
+/**
+ * Base class for algorithm operators.
+ * @param <T> The class type of the {@link AlgoOperator} implementation itself
+ */
+public abstract class AlgoOperator<T extends AlgoOperator <T>> implements WithParams<T>, Serializable {
```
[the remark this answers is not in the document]
xuyang1706: Thanks, we have add more description. 
PipelineStage only supports single input and single output, it is the basic unit for pipeline. AlgoOperator supports multi-input and multi-output. We’d like to implement the algorithm with AlgoOperator, and PipelineStage’s fit and transform function can call the AlgoOperator.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/AlgoOperator.java
```diff
@@ -0,0 +1,98 @@
...
+
+	protected AlgoOperator(Params params) {
+		if (null == params) {
+			this.params = new Params();
+		} else {
+			this.params = params.clone();
+		}
+	}
+
+	public static AlgoOperator sourceFrom(Table table) {
```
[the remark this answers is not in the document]
xuyang1706: thx, added

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/MLSession.java
```diff
@@ -0,0 +1,187 @@
...
+ * by the unique {@link StreamTableEnvironment}
+ */
+public class MLSession {
+	private static ExecutionEnvironment env;
+	private static StreamExecutionEnvironment streamEnv;
+	private static BatchTableEnvironment batchTableEnv;
+	private static StreamTableEnvironment streamTableEnv;
+
+	/**
+	 * Factory to create {@link ExecutionEnvironment}.
```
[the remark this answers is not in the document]
xuyang1706: thx, changed

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/pipeline/Transformer.java
xuyang1706: thx, changed

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/streamoperator/StreamOperator.java
xuyang1706: thx, added

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/pipeline/Model.java
xuyang1706: thx, changed

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/MLSession.java
xuyang1706: we defined MLEnvironmentFactory to replace the global static MLSession and support to set multiple Environment

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/pipeline/PipelineStage.java
xuyang1706: > Same here, `PipelineStageBase`? This class does not implement the `org.apache.flink.ml.api.core.PipelineStage`, is this intended?

> I am not sure this class is a must. I think each type of pipeline stage individually can set its own base class. see: https://stackoverflow.com/a/3599379/11332462

It has been renamed and PipelineStage is the standard concept in pipeline and if it implements the `WithParams` , the subclasses could do not care about `WithParams`

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/batchoperator/BatchOperator.java
xuyang1706: > what's the rational of making link in this direction? what if downstream operator calls `linkFrom` twice with the same upstream; or twice with different upstreams?

thx

1. with this method, the operator can link from source to sink if every operator in the dag is single input and single output and it is straightforward that execute in order from front to back
2. It is not recommended to linkFrom itself or link the same group inputs twice（added to javadoc） and the implement of the operator will define the behavior

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/streamoperator/StreamOperator.java
xuyang1706: Thanks for your advice and offline discussion. In our ML algorithm implementations and applications, we need not use org.apache.flink.streaming.api.operators.StreamOperator, and we'd like to use BatchOperator in ML lib.

xuyang1706:
> @xuyang1706 Thanks for the patch. I left some comments. In general it would avoid class name collision. We may also want to consider if the packaging can change a little bit. For example:
> 
> ```
> org.apache.flink.ml.
>   .algooperator
>       .AlgoOperator.java
>       .stream
>           .StreamAlgoOperator.java
>           .source
>                .TableSourceStreamAlgoOperator.java
>       .batch
>           .BatchAlgoOperator.java
>           .source
>                TableSourceBatchAlgoOperator.java
> ```

@becketqin , thanks for your advice. we defined MLEnvironmentFactory to replace the global static MLSession and support to set multiple Environment, refactored the class names and refined the javaDoc.

xuyang1706:
> Thanks for the contribution @xuyang1706 and sorry for joining the discussion late. Please see some of my comments. In general I think there are lots of added APIs that we should carefully document. Please let me know what you guys think ;-) thanks -Rong

@walterddr , thanks for your comments. We added more description on the concepts and APIs, and refactored the core functions of link, linkTo and linkFrom.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/batchoperator/source/TableSourceBatchOp.java
```diff
@@ -0,0 +1,43 @@
...
+
+import org.apache.flink.ml.batchoperator.BatchOperator;
+import org.apache.flink.table.api.Table;
+
+/**
+ * Transform the Table to SourceBatchOp.
+ */
+public final class TableSourceBatchOp extends BatchOperator<TableSourceBatchOp> {
+
+	public TableSourceBatchOp(Table table) {
```
ex00: this logic could be moved to parent, it same as in `TableSourceStreamOp `

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/AlgoOperator.java
```diff
@@ -0,0 +1,156 @@
...
+			this.params = new Params();
+		} else {
+			this.params = params.clone();
+		}
+	}
+
+	@Override
+	public Params getParams() {
+		if (null == this.params) {
+			this.params = new Params();
```
ex00: duplicate logic from constructor. 
If we cant set `params` as null we don't need check any time it.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/AlgoOperator.java
```diff
@@ -0,0 +1,156 @@
...
+	}
+
+	/**
+	 * create a new AlgoOperator from table.
+	 * @param table the input table
+	 * @return the new AlgoOperator
+	 */
+	public static AlgoOperator<?> sourceFrom(Table table) {
+		if (((TableImpl) table).getTableEnvironment() instanceof StreamTableEnvironment) {
+			return new TableSourceStreamOp(table);
```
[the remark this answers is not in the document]
xuyang1706: Thanks, removed.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/AlgoOperator.java
xuyang1706: Thanks for your advice, removed the check process.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/AlgoOperator.java
```diff
@@ -0,0 +1,156 @@
...
+
+	/**
+	 * The side outputs of operator that be similar to the stream's side outputs.
+	 */
+	private Table[] sideOutputs = null;
+
+	/**
+	 * Construct the operator with empty Params.
+	 */
+	protected AlgoOperator() {
```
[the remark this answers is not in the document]
xuyang1706: The Table is the output of the operator. It should be set at linkFrom and call getOutput to get the output for the next linked operator in most of the scenes. So the AlgoOperator() and the AlgoOperator(Params params) should be exist.
The TableSourceBatchOp and the TableSourceStreamOp is the only two cases that construct the Operator using the table. These operators are the special cases and need not add a new constructor in the base class for these cases.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/batchoperator/source/TableSourceBatchOp.java
xuyang1706: refer to https://github.com/apache/flink/pull/9184#discussion_r326942376

xuyang1706:
> Hi @xuyang1706 thanks for update,
> I've left additional comments, please look they.

@ex00 , thanks for your advice,  I have refactored the sourceFrom() method and removed the unnecessary check process.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/batchoperator/BatchOperator.java
```diff
@@ -0,0 +1,127 @@
...
+import org.apache.flink.ml.common.AlgoOperator;
+import org.apache.flink.table.api.Table;
+import org.apache.flink.util.Preconditions;
+
+/**
+ * Base class of batch algorithm operators.
+ *
+ * <p>This class is extended to support the data transmission between the BatchOperators.
+ */
+public abstract class BatchOperator<T extends BatchOperator<T>> extends AlgoOperator<T> {
```
walterddr: one suggestions I have is to put the operators into a single package
for example
```
|--operator
   |--AlgoOperator.java
   |--stream
      |--StreamOperator.java
      |--source
         |--TableSourceStreamOperator.java
   |--batch
      |--BatchOperator.java
      |--source
         |--TableSourceBatchOperator.java
```
IMO this is more clear than putting `AlgoOperator` in common. and others scatter around the package directory, what do you think?

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/MLEnvironment.java
```diff
@@ -0,0 +1,178 @@
...
+/**
+ * MLEnvironment hold the execution environment.
+ *
+ * @see ExecutionEnvironment
+ * @see StreamExecutionEnvironment
+ * @see BatchTableEnvironment
+ * @see StreamTableEnvironment
+ */
+public class MLEnvironment {
+	private ExecutionEnvironment env;
```
walterddr: any reason MLEnvironment supports both stream and batch at the same time? or is this just to avoid having to split into `StreamMLEnv` and `BatchMlEnv` which both implements some sort of common interface?

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/MLEnvironment.java
```diff
@@ -0,0 +1,178 @@
...
+			) {
+				throw new RuntimeException("There should be only one stream table environment");
+			} else {
+				streamTableEnv = (StreamTableEnvironment) tEnv;
+
+				if (this.streamEnv == null) {
+					this.streamEnv = ((StreamTableEnvironmentImpl) streamTableEnv).execEnv();
+				}
+			}
+		} else {
```
walterddr: I would still put a check here 
```
else if (tEnv instanceof BatchTableEnvironment) {
  //...
} else {
  throw new IllegalArgumentEception(...);
}
```

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/pipeline/EstimatorBase.java
```diff
@@ -0,0 +1,101 @@
...
+import org.apache.flink.table.api.internal.TableImpl;
+import org.apache.flink.table.api.java.StreamTableEnvironment;
+
+/**
+ * The base class for estimator implementations.
+ *
+ * @param <E> The class type of the {@link EstimatorBase} implementation itself
+ * @param <M> class type of the {@link ModelBase} this Estimator produces.
+ */
+public abstract class EstimatorBase<E extends EstimatorBase<E, M>, M extends ModelBase<M>>
```
walterddr: I might've been wrong, but looks like `EstimatorBase` is not tested?

walterddr: same as other Pipeline stages.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/AlgoOperator.java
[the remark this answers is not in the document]
walterddr: I think the main point here is that since the `public AlgoOperator()`  API is not used/tested. we can add it in from later PRs, correct me if I were wrong @ex00 ? 
If that's the case then yes I do agree that any new APIs added which are not tested should be part of latter PRs.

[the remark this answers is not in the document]
ex00: yes @walterddr, you are right. I don't see how using this constructor here.
Also, personally for me, logic in parent class, when using `output` field and it is not marked 'as required', is confusing a little bit.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/MLEnvironment.java
xuyang1706: Yes, it is intended to be this for simplicity.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/AlgoOperator.java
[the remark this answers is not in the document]
xuyang1706: In later PRs, these APIs will be used and tested.

For how using this constructor, I'd like to give an example. 
SplitBatchOp is widely used in ML data pre-preprossing, which splits one dataset input 2 dataset: training set and validation set. It is very convenient for us to write code like this:
new SplitBatchOp().setSplitRatio(0.9)

For output field. The AlgoOperator may have one or more result tables, in most cases, it has only one result. The output is the main operation result table, and the other results are kept in the sideOutputs.
For example, 2 AlgoOperators: AlgoA and AlgoB, AlgoB takes the AlgoA’s results as its inputs, we can write the code like this:
AlgoB.linkFrom(AlgoA)
AlgoA.getOutput() provides the main result of AlgoA, and AlgoA.getSideOutputs() provides the other results of AlgoA. AlgoB will take the AlgoA’s results as its inputs by calling AlgoA.getOutput() and AlgoA.getSideOutputs().

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/pipeline/EstimatorBase.java
xuyang1706: Thanks, we have add some cases, and we will add more cases with the extended class.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/MLEnvironment.java
xuyang1706: Thanks for your advice, added.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/batchoperator/BatchOperator.java
xuyang1706: Thanks @becketqin , @walterddr , @ex00 , I have refactored the package paths according your suggestions.

xuyang1706:
> Thanks for the follow up @xuyang1706 . Overall the changes looks good to me. I left some minor comments. Also I notices that some large trunk of code was missing tests? were we going to support individual tests in the future, or they are part of the `MLSessionTest`?

Thanks @walterddr , I refactored the package paths, added some test cases and refined the JavaDoc with examples.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/AlgoOperator.java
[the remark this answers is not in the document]
ex00: >new SplitBatchOp().setSplitRatio(0.9)

How will it be apply to data source and how will define output result? Could you explain more please?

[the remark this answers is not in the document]
xuyang1706: Thanks, we have the `setOutput` and can set the output in any method of the operator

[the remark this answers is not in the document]
ex00: is it mean what it will be look like `new SplitBatchOp().setSplitRatio(0.9).setOutput(table)` ?
if it is right assumption, that mean that who will write code need to know that need to call `setOutput`. And if someone will doing something from you example, easier to lost this call, and exception will be throw in run time only
```
AlgoA = new SplitBatchOp().setSplitRatio(0.9)
AlgoB.linkFrom(AlgoA)
AlgoA.getSchema()  // NPE: output is null
```
But if we define that `SplitBatchOp(table)` and developer don't thing about additional method calls and will get error on compile time in wrong case.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/operator/AlgoOperator.java
```diff
@@ -0,0 +1,163 @@
...
+import org.apache.flink.api.common.typeinfo.TypeInformation;
+import org.apache.flink.ml.api.misc.param.Params;
+import org.apache.flink.ml.api.misc.param.WithParams;
+import org.apache.flink.ml.params.shared.HasMLEnvironmentId;
+import org.apache.flink.table.api.Table;
+import org.apache.flink.table.api.TableSchema;
+
+import java.io.Serializable;
+
+/**
```
becketqin: The class java doc here is a little confusing, especially given that there is no `linkFrom()` method in `AlgoOperator` class. My understanding is the following:


Base class for the algorithm operators. It hosts the parameters and output tables of an algorithm operator. Each AlgoOperator may have one or more output tables. One of the output table is the primary output table which can be obtained by calling {@link #getOutput}. The other output tables are side output tables that can be obtained by calling {@link #getSideOutput()}.

The input of an AlgoOperator is defined in the subclasses of the AlgoOperator.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/operator/AlgoOperator.java
```diff
@@ -0,0 +1,163 @@
...
+
+	/**
+	 * The side outputs of operator that be similar to the stream's side outputs.
+	 */
+	private Table[] sideOutputs = null;
+
+	/**
+	 * Construct the operator with empty Params.
+	 *
+	 * <p>For how using this constructor, there is an example:
```
becketqin: This constructor is especially useful when users want to set parameters for the algorithm operators. For example, .....

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/operator/AlgoOperator.java
```diff
@@ -0,0 +1,163 @@
...
+ * AlgoA.getOutput()
+ * }
+ * </pre>
+ * The code provides the main result of AlgoA, and AlgoA.getSideOutputs()
+ * provides the other results of AlgoA. AlgoB will take the AlgoA’s results as its
+ * inputs by calling AlgoA.getOutput() and AlgoA.getSideOutputs().
+ *
+ * @param <T> The class type of the {@link AlgoOperator} implementation itself
+ */
+public abstract class AlgoOperator<T extends AlgoOperator <T>>
```
becketqin: Extra space before <T>

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/operator/stream/StreamOperator.java
```diff
@@ -0,0 +1,130 @@
...
+import org.apache.flink.ml.operator.stream.source.TableSourceStreamOp;
+import org.apache.flink.table.api.Table;
+import org.apache.flink.util.Preconditions;
+
+/**
+ * Base class of stream algorithm operators.
+ *
+ * <p>This class is extended to support the data transmission between the StreamOperators.
+ */
+public abstract class StreamOperator<T extends StreamOperator<T>> extends AlgoOperator<T> {
```
becketqin: Can we change the java doc in this class to match those in BatchOperator?

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/operator/stream/StreamOperator.java
```diff
@@ -0,0 +1,130 @@
...
+	 * create a new StreamOperator from table.
+	 *
+	 * @param table the input table
+	 * @return the new StreamOperator
+	 */
+	public static StreamOperator<?> sourceFrom(Table table) {
+		return new TableSourceStreamOp(table);
+	}
+
+	protected void checkOpSize(int size, StreamOperator<?>... inputs) {
```
becketqin: Why these methods are not static here just like they are in BatchOperator? Also, it seems that these methods can be merged with the methods in BatchOperator by changing the parameter type to AlgoOperator.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/pipeline/EstimatorBase.java
```diff
@@ -0,0 +1,101 @@
...
+		return fit(input);
+	}
+
+	/**
+	 * Train and produce a {@link ModelBase} which fits the records in the given {@link Table}.
+	 *
+	 * @param input the table with records to train the Model.
+	 * @return a model trained to fit on the given Table.
+	 */
+	public M fit(Table input) {
```
becketqin: Does this method have to be public?

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/pipeline/PipelineStageBase.java
```diff
@@ -0,0 +1,66 @@
...
+ */
+
+package org.apache.flink.ml.pipeline;
+
+import org.apache.flink.ml.api.misc.param.Params;
+import org.apache.flink.ml.api.misc.param.WithParams;
+import org.apache.flink.ml.params.shared.HasMLEnvironmentId;
+
+/**
+ * The base class for a stage in a pipeline, either an [[Estimator]] or a [[Transformer]].
```
becketqin: It seems the java doc format for the classes are not quite consistent. Sometimes it is `Class`, sometimes <code>Class</code>, and here it is [[Class]]. Can we make them consistent?

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/pipeline/PipelineStageBase.java
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * <p>Each pipeline stage is with parameters, and requires a public empty constructor for
```
becketqin: The PipelineStageBase maintains the parameters for the stage. A default constructor is needed in order to restore a pipeline stage.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/pipeline/TransformerBase.java
```diff
@@ -0,0 +1,100 @@
...
+import org.apache.flink.ml.operator.stream.source.TableSourceStreamOp;
+import org.apache.flink.table.api.Table;
+import org.apache.flink.table.api.TableEnvironment;
+import org.apache.flink.table.api.internal.TableImpl;
+import org.apache.flink.table.api.java.StreamTableEnvironment;
+
+/**
+ * The base class for transformer implementations.
+ *
+ * @param <T> The class type of the {@link TransformerBase} implementation itself, used by {@link
```
becketqin: The class type of the {@link TransformerBase} implementation itself =>

A subclass of {@link TransformerBase}, used by ...

flink-ml-parent/flink-ml-lib/src/test/java/org/apache/flink/ml/common/MLEnvironmentTest.java
```diff
@@ -0,0 +1,87 @@
...
+import org.apache.flink.table.api.java.StreamTableEnvironment;
+
+import org.junit.Test;
+
+import java.util.Arrays;
+
+/**
+ * Test cases for MLEnvironment.
+ */
+public class MLEnvironmentTest {
```
becketqin: It looks that there is no assertion in this test class.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/AlgoOperator.java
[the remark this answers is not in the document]
xuyang1706: Yes, there must be one or more data sources. We have defined some common *SourceBatchOp to get the source data conveniently, and from the sourceBatchOp, we can link the algorithm operations.
Here is a whole example:
```
CsvSourceBatchOp algoA =
	new CsvSourceBatchOp()
		.setFilePath("http://alink-dataset.cn-hangzhou.oss.aliyun-inc.com/csv/iris.csv")
		.setSchemaStr("s_length double, s_width double, p_length double, p_width double, category string");

System.out.println("schema of algoA:");
System.out.println(algoA.getSchema());

SplitBatchOp algoB = new SplitBatchOp().setFraction(0.8);

algoB.linkFrom(algoA);

System.out.println("schema of algoB's main result:");
System.out.println(algoB.getSchema());

System.out.println("schema of algoB's side output result:");
System.out.println(algoB.getSideOutput().getSchema());
```

And the print results:
```
schema of algoA:
root
 |-- s_length: Double
 |-- s_width: Double
 |-- p_length: Double
 |-- p_width: Double
 |-- category: String

schema of algoB's main result:
root
 |-- s_length: Double
 |-- s_width: Double
 |-- p_length: Double
 |-- p_width: Double
 |-- category: String

schema of algoB's side output result:
root
 |-- s_length: Double
 |-- s_width: Double
 |-- p_length: Double
 |-- p_width: Double
 |-- category: String
```

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/batchoperator/BatchOperator.java
```diff
@@ -0,0 +1,136 @@
...
+import org.apache.flink.ml.api.misc.param.Params;
+import org.apache.flink.ml.batchoperator.source.TableSourceBatchOp;
+import org.apache.flink.ml.common.AlgoOperator;
+import org.apache.flink.table.api.Table;
+import org.apache.flink.util.Preconditions;
+
+/**
+ * Base class of batch algorithm operators.
+ *
+ * <p>This class is extended to support the data transmission between the BatchOperators.
```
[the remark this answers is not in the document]
xuyang1706: Thanks, changed

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/batchoperator/BatchOperator.java
```diff
@@ -0,0 +1,136 @@
...
+	 * @param params the initial Params.
+	 */
+	public BatchOperator(Params params) {
+		super(params);
+	}
+
+	/**
+	 * Abbreviation of {@link #linkTo(BatchOperator)}.
+	 */
+	public <B extends BatchOperator<?>> B link(B next) {
```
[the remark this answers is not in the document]
xuyang1706: Thanks, removed the linkTo

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/batchoperator/BatchOperator.java
```diff
@@ -0,0 +1,127 @@
...
+	 *
+	 * <pre>
+	 * {@code
+	 * BatchOperator a = ...;
+	 * BatchOperator b = ...;
+	 * BatchOperator c = a.link(b)
+	 * }
+	 * </pre>
+	 *
+	 * <p>the <code>c</code> in upper code is the linked
```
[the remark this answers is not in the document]
xuyang1706: Thanks, changed

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/batchoperator/BatchOperator.java
```diff
@@ -0,0 +1,127 @@
...
+	 * <code>b</code> which use <code>a</code> as input.
+	 *
+	 * @param next the linked BatchOperator
```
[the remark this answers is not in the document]
xuyang1706: Thanks, changed

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/batchoperator/BatchOperator.java
```diff
@@ -0,0 +1,127 @@
...
+	 * @param params the initial Params.
+	 */
+	public BatchOperator(Params params) {
+		super(params);
+	}
+
+	/**
+	 * Link to another {@link BatchOperator}.
+	 *
+	 * <p>Link the <code>next</code> to BatchOperator using this as its input.
```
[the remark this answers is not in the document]
xuyang1706: Thanks, changed

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/batchoperator/BatchOperator.java
```diff
@@ -0,0 +1,127 @@
...
+	 * {@code
+	 * BatchOperator a = ...;
+	 * BatchOperator b = ...;
+	 * BatchOperator c = ...;
+	 *
+	 * BatchOperator d = c.linkFrom(a, b)
+	 * }
+	 * </pre>
+	 *
+	 * <p>the <code>d</code> in upper code is the linked
```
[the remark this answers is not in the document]
xuyang1706: Thanks, changed

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/batchoperator/BatchOperator.java
```diff
@@ -0,0 +1,127 @@
...
+	 * <code>c</code> which use a and b as its inputs.
+	 *
+	 * <p>note: It is not recommended to linkFrom itself or linkFrom the same group inputs twice.
```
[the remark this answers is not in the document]
xuyang1706: Thanks, the subclass define the behavior, so we add the tips in the comment only.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/batchoperator/BatchOperator.java
```diff
@@ -0,0 +1,127 @@
...
+	 * @return the linked this object
+	 */
+	public abstract T linkFrom(BatchOperator<?>... inputs);
+
+	/**
+	 * create a new BatchOperator from table.
+	 * @param table the input table
+	 * @return the new BatchOperator
+	 */
+	public static BatchOperator<?> sourceFrom(Table table) {
```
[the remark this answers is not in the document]
xuyang1706: Thanks, changed

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/batchoperator/BatchOperator.java
```diff
@@ -0,0 +1,127 @@
...
+		return new TableSourceBatchOp(table);
+	}
+
+	protected void checkOpSize(int size, BatchOperator<?>... inputs) {
+		Preconditions.checkNotNull(inputs, "Operators should not be null.");
+		Preconditions.checkState(inputs.length == size, "The size of operators should be equal to "
+			+ size + ", current: " + inputs.length);
+	}
+
+	protected void checkRequiredOpSize(int size, BatchOperator<?>... inputs) {
```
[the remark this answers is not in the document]
xuyang1706: Thanks, changed

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/MLEnvironment.java
```diff
@@ -0,0 +1,184 @@
...
+import org.apache.flink.table.api.Table;
+import org.apache.flink.table.api.TableEnvironment;
+import org.apache.flink.table.api.java.BatchTableEnvironment;
+import org.apache.flink.table.api.java.StreamTableEnvironment;
+import org.apache.flink.table.api.java.internal.StreamTableEnvironmentImpl;
+import org.apache.flink.types.Row;
+import org.apache.flink.util.Preconditions;
+
+/**
+ * MLEnvironment hold the execution environment.
```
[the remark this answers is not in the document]
xuyang1706: Thanks, changed

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/MLEnvironment.java
```diff
@@ -0,0 +1,184 @@
...
+		return streamTableEnv;
+	}
+
+	/**
+	 * Set the ExecutionEnvironment.
+	 * The ExecutionEnvironment should be set only once.
+	 *
+	 * @param env the ExecutionEnvironment
+	 */
+	public void setExecutionEnvironment(ExecutionEnvironment env) {
```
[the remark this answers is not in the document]
xuyang1706: Thanks, removed

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/MLEnvironment.java
```diff
@@ -0,0 +1,184 @@
...
+		this.env = env;
+	}
+
+	/**
+	 * Set the StreamExecutionEnvironment.
+	 * The StreamExecutionEnvironment should be set only once.
+	 *
+	 * @param env the StreamExecutionEnvironment
+	 */
+	public void setStreamExecutionEnvironment(StreamExecutionEnvironment env) {
```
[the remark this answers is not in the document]
xuyang1706: Thanks, removed

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/MLEnvironment.java
```diff
@@ -0,0 +1,184 @@
...
+	/**
+	 * Set the TableEnvironment.
+	 * The TableEnvironment should be set only once.
+	 *
+	 * <p>We also set the {@link ExecutionEnvironment} or {@link StreamExecutionEnvironment} accordingly.
+	 *
+	 * @param tEnv  the TableEnvironment
+	 * @param table the table for get {@link ExecutionEnvironment} for batch
+	 */
+	public void setTableEnvironment(TableEnvironment tEnv, Table table) {
```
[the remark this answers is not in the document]
xuyang1706: Thanks, removed

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/MLEnvironment.java
```diff
@@ -0,0 +1,184 @@
...
+	/**
+	 * Get the ExecutionEnvironment.
+	 * if the ExecutionEnvironment has not been set, it initial the ExecutionEnvironment
+	 * with default Configuration.
+	 *
+	 * @return the batch {@link ExecutionEnvironment}
+	 * @see MLEnvironment#setExecutionEnvironment(ExecutionEnvironment)
+	 */
+	public ExecutionEnvironment getExecutionEnvironment() {
+		if (null == env) {
```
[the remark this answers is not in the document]
xuyang1706: Thanks, I have been added the Constructor that use the environment as its parameters

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/MLEnvironmentFactory.java
```diff
@@ -0,0 +1,98 @@
...
+
+	/**
+	 * A 'id' is a unique identifier of a MLEnvironment.
+	 */
+	private static Long id = 1L;
+
+	/**
+	 * Map that hold the MLEnvironment and use the MLEnvironmentId as its key.
+	 */
+	private static HashMap<Long, MLEnvironment> map = new HashMap<>();
```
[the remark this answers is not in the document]
xuyang1706: Thanks, added

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/MLEnvironmentFactory.java
```diff
@@ -0,0 +1,98 @@
...
+
+package org.apache.flink.ml.common;
+
+import java.util.HashMap;
+
+/**
+ * Factory to get the MLEnvironment using a MLEnvironmentId.
+ *
+ * <p>The MLEnvironmentFactory manages the multiple MLEnvironments by the MLEnvironmentId.
+ * Two steps that the user use the Factory:
```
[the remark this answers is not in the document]
xuyang1706: Thanks, changed

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/MLEnvironmentFactory.java
```diff
@@ -0,0 +1,98 @@
...
+	public static synchronized MLEnvironment get(Long mlEnvId) {
+		if (!map.containsKey(mlEnvId)) {
+			throw new RuntimeException("There is no Environment in factory. " +
+				"Maybe you could call `getNewMLEnvironmentId` to create a new MLEnvironmentId");
+		}
+
+		return map.get(mlEnvId);
+	}
+
+	static {
```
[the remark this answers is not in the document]
xuyang1706: Thanks, removed

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/MLEnvironmentFactory.java
```diff
@@ -0,0 +1,98 @@
...
+	public static synchronized MLEnvironment getDefault() {
+		return get(DEFAULT_ML_ENVIRONMENT_ID);
+	}
+
+	/**
+	 * Create a unique MLEnvironment id and set a new MLEnvironment in the factory.
+	 *
+	 * @return the MLEnvironment id.
+	 */
+	public static synchronized Long getNewMLEnvironmentId() {
```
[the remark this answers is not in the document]
xuyang1706: Thanks, added

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/operator/AlgoOperator.java
```diff
@@ -0,0 +1,163 @@
...
+	 * Returns the column names of the output table.
+	 */
+	public String[] getColNames() {
+		return getSchema().getFieldNames();
+	}
+
+	/**
+	 * Returns the column types of the output table.
+	 */
+	public TypeInformation <?>[] getColTypes() {
```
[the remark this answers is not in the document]
xuyang1706: Thanks, changed

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/operator/batch/BatchOperator.java
```diff
@@ -0,0 +1,127 @@
...
+	 *
+	 * <pre>
+	 * {@code
+	 * BatchOperator a = ...;
+	 * BatchOperator b = ...;
+	 * BatchOperator c = a.link(b)
+	 * }
+	 * </pre>
+	 *
+	 * <p>the <code>c</code> in upper code is the linked
```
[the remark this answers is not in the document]
xuyang1706: Thanks, changed

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/operator/batch/BatchOperator.java
```diff
@@ -0,0 +1,127 @@
...
+	 * {@code
+	 * BatchOperator a = ...;
+	 * BatchOperator b = ...;
+	 * BatchOperator c = ...;
+	 *
+	 * BatchOperator d = c.linkFrom(a, b)
+	 * }
+	 * </pre>
+	 *
+	 * <p>the <code>d</code> in upper code is the linked
```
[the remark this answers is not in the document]
xuyang1706: Thanks, changed

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/operator/batch/BatchOperator.java
```diff
@@ -0,0 +1,127 @@
...
+	 * @return the linked this object
+	 */
+	public abstract T linkFrom(BatchOperator<?>... inputs);
+
+	/**
+	 * create a new BatchOperator from table.
+	 * @param table the input table
+	 * @return the new BatchOperator
+	 */
+	public static BatchOperator<?> sourceFrom(Table table) {
```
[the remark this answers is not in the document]
xuyang1706: Thanks, changed

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/operator/batch/BatchOperator.java
```diff
@@ -0,0 +1,127 @@
...
+		return new TableSourceBatchOp(table);
+	}
+
+	protected void checkOpSize(int size, BatchOperator<?>... inputs) {
+		Preconditions.checkNotNull(inputs, "Operators should not be null.");
+		Preconditions.checkState(inputs.length == size, "The size of operators should be equal to "
+			+ size + ", current: " + inputs.length);
+	}
+
+	protected void checkRequiredOpSize(int size, BatchOperator<?>... inputs) {
```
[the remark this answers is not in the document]
xuyang1706: Thanks, changed

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/operator/AlgoOperator.java
xuyang1706: Thanks, changed

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/operator/AlgoOperator.java
xuyang1706: Thanks, changed

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/operator/AlgoOperator.java
xuyang1706: Thanks, changed

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/operator/stream/StreamOperator.java
xuyang1706: Thanks, changed

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/pipeline/PipelineStageBase.java
xuyang1706: Thanks, changed

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/pipeline/PipelineStageBase.java
xuyang1706: Thanks, changed

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/pipeline/TransformerBase.java
xuyang1706: Thanks, changed

flink-ml-parent/flink-ml-lib/src/test/java/org/apache/flink/ml/common/MLEnvironmentTest.java
xuyang1706: Thanks, changed

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/pipeline/EstimatorBase.java
xuyang1706: Thanks, this method is more convenient with the MLEnvironmentFactory, so I think it should be public

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/operator/stream/StreamOperator.java
xuyang1706: Thanks, these methods in the BatchOperator is also not static. The BatchOperator and the StreamOperator should not be mixed in the array

xuyang1706:
> @xuyang1706 Thanks for updating the patch. I left a few more comments. Can you take a look?

Thanks, the comments in the code have been refined. The `sourceFrom` has been renamed to fromTable. The constructor with the Environment has been added in the MLEnvironment. And the register

Thanks @becketqin, I have refined the code according your comments.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/AlgoOperator.java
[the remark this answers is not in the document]
ex00: now I see, thanks for the explanation

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/pipeline/TransformerBase.java
```diff
@@ -0,0 +1,98 @@
...
+	public TransformerBase() {
+		super();
+	}
+
+	public TransformerBase(Params params) {
+		super(params);
+	}
+
+	@Override
+	public Table transform(TableEnvironment tEnv, Table input) {
```
becketqin: Should we do a sanity check to make sure the tEnv passed in is the same table environment used by the input table?

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/pipeline/EstimatorBase.java
```diff
@@ -0,0 +1,100 @@
...
+	public EstimatorBase() {
+		super();
+	}
+
+	public EstimatorBase(Params params) {
+		super(params);
+	}
+
+	@Override
+	public M fit(TableEnvironment tEnv, Table input) {
```
becketqin: Should we do a sanity check to make sure the tEnv passed in is the same one used by the input table?

flink-ml-parent/flink-ml-lib/src/test/java/org/apache/flink/ml/common/MLEnvironmentTest.java
```diff
@@ -0,0 +1,64 @@
...
+	public void testConstructWithStreamEnv() {
+		StreamExecutionEnvironment streamExecutionEnvironment = StreamExecutionEnvironment.getExecutionEnvironment();
+		StreamTableEnvironment streamTableEnvironment = StreamTableEnvironment.create(streamExecutionEnvironment);
+
+		MLEnvironment mlEnvironment = new MLEnvironment(null, null, streamExecutionEnvironment, streamTableEnvironment);
+
+		Assert.assertSame(mlEnvironment.getStreamExecutionEnvironment(), streamExecutionEnvironment);
+		Assert.assertSame(mlEnvironment.getStreamTableEnvironment(), streamTableEnvironment);
+	}
+}
```
becketqin: If we only expect the MLEnvironment to be constructed in this two ways, maybe we can just have those two specific constructors?

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/MLEnvironmentFactory.java
```diff
@@ -0,0 +1,125 @@
...
+ */
+public class MLEnvironmentFactory {
+
+	/**
+	 * The default MLEnvironmentId.
+	 */
+	public static final Long DEFAULT_ML_ENVIRONMENT_ID = 0L;
+
+	/**
+	 * A 'id' is a unique identifier of a MLEnvironment.
```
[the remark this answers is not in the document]
xuyang1706: Thanks, changed

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/operator/AlgoOperator.java
```diff
@@ -0,0 +1,153 @@
...
+	 * @param output the output table.
+	 */
+	protected void setOutput(Table output) {
+		this.output = output;
+	}
+
+	/**
+	 * Returns the column names of the output table.
+	 */
+	public String[] getColNames() {
```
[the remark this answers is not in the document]
xuyang1706: Thanks, added

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/operator/stream/StreamOperator.java
```diff
@@ -0,0 +1,134 @@
...
+		Preconditions.checkNotNull(inputs, "Operators should not be null.");
+		Preconditions.checkState(inputs.length == size, "The size of operators should be equal to "
+			+ size + ", current: " + inputs.length);
+	}
+
+	protected void checkMinOpSize(int size, StreamOperator<?>... inputs) {
```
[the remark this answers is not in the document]
xuyang1706: Thanks, changed

flink-ml-parent/flink-ml-lib/src/test/java/org/apache/flink/ml/common/MLEnvironmentTest.java
xuyang1706: Thanks, changed

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/MLEnvironment.java
```diff
@@ -0,0 +1,130 @@
...
+	/**
+	 * Get the StreamExecutionEnvironment.
+	 * if the StreamExecutionEnvironment has not been set, it initial the StreamExecutionEnvironment
+	 * with default Configuration.
+	 *
+	 * @return the {@link StreamExecutionEnvironment}
+	 */
+	public StreamExecutionEnvironment getStreamExecutionEnvironment() {
+		if (null == streamEnv) {
+			streamEnv = StreamExecutionEnvironment.getExecutionEnvironment();
```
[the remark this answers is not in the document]
xuyang1706: Thanks, as we discussed offline, for depending an unfixed bug, we will use current implementation as a temp workaround.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/MLEnvironmentFactory.java
```diff
@@ -0,0 +1,125 @@
...
+		return get(DEFAULT_ML_ENVIRONMENT_ID);
+	}
+
+	/**
+	 * Set the default MLEnvironment.
+	 * The default MLEnvironment should be set only once.
+	 *
+	 * @param env the MLEnvironment
+	 */
+	public static synchronized void setDefault(MLEnvironment env) {
```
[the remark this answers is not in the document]
xuyang1706: Thanks, I have checked the existence of default MLEnv, user could only set one default MLEnv.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/pipeline/TransformerBase.java
xuyang1706: I agree with replace `void fit(TableEnvironment, Table)` method with the `MLEnvironmentFactory`, and I suggest we could discuss and change API in another PR.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/pipeline/EstimatorBase.java
xuyang1706: Ditto above.

xuyang1706:
> @xuyang1706 Thanks for updating the patch. I left a few more comments. Most of them are minor. One major thing is I am not sure whether we should completely remove `fit(TableEnvironment, Table)` and `transform(TableEnvironment, Table)` API, because they are completely bypassed in the current implementation given the introduction of `MLEnvironmentFactory`.

Thanks, @becketqin . I agree with replace `void fit(TableEnvironment, Table)` method with the `MLEnvironmentFactory`, and I suggest we could discuss and change the API in another PR.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/MLEnvironmentFactory.java
```diff
@@ -0,0 +1,125 @@
...
+	 *
+	 * @param mlEnvId the MLEnvironmentId
+	 * @return the MLEnvironment
+	 */
+	public static synchronized MLEnvironment get(Long mlEnvId) {
+		if (!map.containsKey(mlEnvId)) {
+			if (mlEnvId.equals(DEFAULT_ML_ENVIRONMENT_ID)) {
+				setDefault(new MLEnvironment());
+			} else {
+				throw new IllegalArgumentException("There is no Environment in factory. " +
```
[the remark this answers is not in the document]
xuyang1706: Thanks for your advice, we refactored code to create the default ML Session in the static block and changed the exception message.

becketqin:
@xuyang1706 Thanks for updating the patch. the patch LGTM overall. I created a PR with some minor improvements against your branch. Can you check if that makes sense? If so, feel free to merge it to your branch.

@walterddr Do you also want to take another look? I am thinking of merging the patch either on Sunday or Monday. Thanks.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/MLEnvironmentFactory.java
```diff
@@ -0,0 +1,111 @@
...
+		return nextId++;
+	}
+
+	/**
+	 * Remove the MLEnvironment using the MLEnvironmentId.
+	 *
+	 * @param mlEnvId the id.
+	 * @return the removed MLEnvironment
+	 */
+	public static synchronized MLEnvironment remove(Long mlEnvId) {
```
walterddr: when would this API be called? seems like it is not necessary?

xuyang1706: MLEnviromentFactory can create multi MLEnviroment, if some of them only used in a time slot, user can use this API to release the resource. In most cases, user just use default MLEnviroment. This API is rarely used.

xuyang1706:
> Thanks for the contribution @xuyang1706 and the reviews @becketqin . Overall it looks good to me. I only left a minor comment. please kindly take a look.
> +1 to merge to unblock future developments.

Thanks @walterddr. Yes, this API is rarely used, but in some cases, if user want to release no-used MLEnviroment in the process, this is only API could be called. Thus, I prefer to keep it.

xuyang1706:
> @xuyang1706 Thanks for updating the patch. the patch LGTM overall. I created a PR with some minor improvements against your branch. Can you check if that makes sense? If so, feel free to merge it to your branch.
> 
> @walterddr Do you also want to take another look? I am thinking of merging the patch either on Sunday or Monday. Thanks.

Thanks for your help @becketqin. I have merged your "minor improvement PR" to my branch.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/MLEnvironmentFactory.java
becketqin: There is one minor bug here. The default MLEnvironment should never be removed. I'll fix this when checking in the code. Thanks.

becketqin:
Merged to master.

PR #9024, opened 2019-07-09
[FLINK-13119] add blink table config to documentation

XuPingyong:
## What is the purpose of the change

1. Add table config annotation to indicate it works for batch or streaming jobs.
2. Add table config to documentation.

The preview screenshot of configuration page looks like:

image

## Verifying this change

*(Please pick either of the following options)*

no

This change added tests and can be verified as follows:

## Does this pull request potentially affect one of the following parts:

  - Dependencies (does it add or upgrade a dependency): (no)
  - The public API, i.e., is any changed class annotated with `@Public(Evolving)`: ( no)
  - The serializers: (no)
  - The runtime per-record code paths (performance sensitive): (no)
  - Anything that affects deployment or recovery: JobManager (and its components), Checkpointing, Yarn/Mesos, ZooKeeper: (no)
  - The S3 file system connector: ( no)

## Documentation

  - Does this pull request introduce a new feature? ( no)

PR #9185, opened 2019-07-20
[FLINK-13266] [table] Relocate blink planner classes to avoid class clashes

godfreyhe:

*Relocate blink planner classes to avoid class clashes, see the relocation part in https://docs.google.com/document/d/15Z1Khy23DwDBp956yBzkMYkGdoQAU_QgBoJmFWSffow *

## Brief change log

  - *move OptimizerConfigOptions & ExecutionConfigOptions to table-api-java module*
  - *move ExpressionParserException & UnresolvedException to table-common module*
  - *remove definedTimeAttributes file in blink planner*
  - *move DataView related classes to table-common module, and remove duplicate classes*
  - *remove Order class from table-runtime-blink*
  - *port descriptors to table-common*
  - *Improve comment for FilterableTableTableSource*
  - *Relocate blink runtime classes to avoid class clashes*

## Verifying this change

no new tests and the existing tests should pass

## Documentation


godfreyhe:
The last commit only involves relocation in blink planner, does not change any code except "import". while the change is not "rename", but is "delete & add" in review page. It's ok in IDEA's "Annotate". I will try to solve this.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/ExecutionConfigOptions.java
```diff
@@ -125,7 +125,7 @@
 	// ------------------------------------------------------------------------
 
 	/**
-	 * See {@link HeapWindowsGrouping}.
+	 * See {@link org.apache.flink.table.runtime.window.grouping.HeapWindowsGrouping}.
```
twalthr: Change this to `@code` otherwise the JavaDoc is invalid.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/ExecutionConfigOptions.java
```diff
@@ -19,12 +19,12 @@
 package org.apache.flink.table.api;
```
twalthr: Move this to `org.apache.flink.table.api.config`?

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/ExecutionConfigOptions.java
```diff
@@ -19,12 +19,12 @@
...
 
 import org.apache.flink.configuration.ConfigOption;
-import org.apache.flink.table.runtime.window.grouping.HeapWindowsGrouping;
 
 import static org.apache.flink.configuration.ConfigOptions.key;
 
 /**
  * This class holds configuration constants used by Flink's table module.
+ * This is only used for BLINK planner.
```
twalthr: nit: `\n<p>This is only used for the Blink planner.`

twalthr:
Some general feedback regarding `ExecutionConfigOptions`. We are now moving it to the API module. We should actually discuss how the properties are named and make sure that those options follow Flink naming conventions. Right now, I see `sql.exec.sort.file-handles.num.max` which should be renamed to `sql.exec.sort.max-num-file-handles` instead of `sql.exec.sort.file-handles.num.max`. Or `sql.exec.window-agg.buffer-size-limit` instead of `sql.exec.window-agg.buffer-size.limit`. When do we plan to do that? @wuchong @godfreyhe

flink-table/flink-table-common/src/main/java/org/apache/flink/table/api/dataview/ListView.java
```diff
@@ -106,7 +106,7 @@ public ListView() {
 	 * Returns an iterable of the list view.
 	 *
 	 * @throws Exception Thrown if the system cannot get data.
-	 * @return The iterable of the list or { @code null} if the list is empty.
+	 * @return The iterable of the list.
```
twalthr: This changes the API. We should return null instead, as documented. Remove the `toString` method in this class. And fix `{ @code null}` to `{@code null}`. Same applies for the MapView.

twalthr: Futhermore, make sure that all views are marked as `@PublicEvolving`. The other classes can be marked as `@Internal`.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/descriptors/ClassInstance.java
```diff
@@ -0,0 +1,230 @@
...
+import java.util.Map;
+
+import static org.apache.flink.types.Either.Left;
+import static org.apache.flink.types.Either.Right;
+
+/**
+ * Descriptor for a class instance. A class instance is a Java/Scala object created from a class
+ * with a public constructor (with or without parameters).
+ */
+public class ClassInstance extends HierarchyDescriptor {
```
twalthr: Annotate all descriptors as `@PublicEvolving`.

KurtYoung:
> Some general feedback regarding `ExecutionConfigOptions`. We are now moving it to the API module. We should actually discuss how the properties are named and make sure that those options follow Flink naming conventions. Right now, I see `sql.exec.sort.file-handles.num.max` which should be renamed to `sql.exec.sort.max-num-file-handles` instead of `sql.exec.sort.file-handles.num.max`. Or `sql.exec.window-agg.buffer-size-limit` instead of `sql.exec.window-agg.buffer-size.limit`. When do we plan to do that? @wuchong @godfreyhe

Thanks for the suggestion. I think now is the good time to adjust the naming since the config name should also be treated as public API. Once we released this, it's very difficult to modify them. BTW, do you have any example which can reflect the flink naming conventions that we can refer to?

KurtYoung:
> From what I see, [2266617](https://github.com/apache/flink/commit/2266617d608ef5f22a75a6d6c6dc809f6f9df1f9) looks good to me for solving the class clashes. But maybe people that are more familiar with the module have further package improvement suggestions? @wuchong @KurtYoung Such a big refactoring does not happen often. Now is the time to improve the package structure a bit. E.g. we could find a better name than `org.apache.flink.table.planner.calcite`? But otherwise +1 from my side.

Yeah, I agree this is also a good time to improve the package structure. We will have a discuss about it and give a proposal.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/ExecutionConfigOptions.java
wuchong: +1

wuchong:
> Some general feedback regarding `ExecutionConfigOptions`. We are now moving it to the API module. We should actually discuss how the properties are named and make sure that those options follow Flink naming conventions. Right now, I see `sql.exec.sort.file-handles.num.max` which should be renamed to `sql.exec.sort.max-num-file-handles` instead of `sql.exec.sort.file-handles.num.max`. Or `sql.exec.window-agg.buffer-size-limit` instead of `sql.exec.window-agg.buffer-size.limit`. When do we plan to do that? @wuchong @godfreyhe

Yes, I think we should do it before release. Actually, we have polished all the property names before. However, there may still be some properties need improvement.

godfreyhe:
> From what I see, [2266617](https://github.com/apache/flink/commit/2266617d608ef5f22a75a6d6c6dc809f6f9df1f9) looks good to me for solving the class clashes. But maybe people that are more familiar with the module have further package improvement suggestions? @wuchong @KurtYoung Such a big refactoring does not happen often. Now is the time to improve the package structure a bit. E.g. we could find a better name than `org.apache.flink.table.planner.calcite`? But otherwise +1 from my side.

thanks for your suggestion @twalthr ,  i will split [2266617](https://github.com/apache/flink/commit/2266617d608ef5f22a75a6d6c6dc809f6f9df1f9) into several smaller commits, and make sure each commits only involves "rename", no "delete/add". 

I and @wuchong have discussed `org.apache.flink.table.planner.calcite` offline:
We deny `org.apache.flink.table.planner.optimizer`, because `org.apache.flink.table.plan` has a sub-package named `org.apache.flink.table.plann.optimize` which defines real `Optimizer`, and optimizer programs. I want to implement different strategy optimizer in `optimize` package in future. 
we think `org.apache.flink.table.planner.plan` is better than `org.apache.flink.table.planner.calcite`, because `plan` more accurately describes its function.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/api/dataview/ListView.java
godfreyhe: i think we should deprecate this method, and avoid `return null` in public method ?

wuchong:
Regarding to the config options, I got some other feedbacks from Timo in #9024. I think we should start a new thread to discuss about this. I will create a new pull request for the config options, so that we can review entry by entry.

Regarding to the naming convention, I think maybe the most important is "hierarchical". For example, 
`sql.exec.window-agg.buffer-size.limit`, `buffer-size` is not a parent level because it doesn't have children options, that's why it should be `buffer-size-limit`. Correct me if I misunderstand according to the [Flink Convention](https://docs.google.com/document/d/1JGRVK6udBFa8xbyv7ipPQiLEhxebA64f8KKcaAH0UPY/edit#heading=h.qmv83r3emtom) Timo pointed out.

wuchong:
I just created a pull request #9203 for config options.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/api/dataview/ListView.java
twalthr: I'm also not a big supporter of returning `null` but I think the semantics are equal to Flink's `ListState.get()`:

```
The operator state value corresponding to the current input or {@code null}
	 * if the state is empty.
```

wuchong: Regarding to the `ListState.get()`, I think it maybe an outdated comment. Because all the States returned to the users will be wrapped in a `UserFacingXXX` (e.g. `UserFacingListState`). The `UserFacingXXX` will return an empty list instead of `null`.

I would prefer not return `null` and I don't think this will break user's behavior, because the null check still work.

godfreyhe:
> Update:
> @godfreyhe what is the purpose of `NullSerializer` and why has `ListViewTypeInfo` a `(TypeInformation<T> elementType, boolean nullSerializer)` constructor? The Flink code did not contain such special cases.

this is introduced in Blink planner, and `nullSerializer` will be always false in Flink planner.  
in blink, `MapView` or `ListView` in `Accumulator` does not need to be serialized, like [MaxWithRetractAggFunction](https://github.com/apache/flink/blob/master/flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/functions/aggfunctions/MaxWithRetractAggFunction.java)

JingsongLi:
> > Update:
> > @godfreyhe what is the purpose of `NullSerializer` and why has `ListViewTypeInfo` a `(TypeInformation<T> elementType, boolean nullSerializer)` constructor? The Flink code did not contain such special cases.
> 
> this is introduced in Blink planner, and `nullSerializer` will be always false in Flink planner.
> in blink, `MapView` or `ListView` in `Accumulator` does not need to be serialized, like [MaxWithRetractAggFunction](https://github.com/apache/flink/blob/master/flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/functions/aggfunctions/MaxWithRetractAggFunction.java)

MapView or ListView in Accumulator can not to be serialized if it is `isStateBackedDataViews`.

In flink-planner: If it is `isStateBackedDataViews`, the flink-planner will remove view field for PojoTypeInfo, the code is in: `UserDefinedFunctionUtils.removeStateViewFieldsFromAccTypeInfo`. (data view fields which are backed by state backend are not serialized.)

In blink-planner: The above approach does not work in blink-planner, because we will convert POJO to BaseRow in blink internal(Use `PojoConverter`), if we discard this field, the BaseRow arity will wrong and field index will confused. And the real state backed data views can not set to the BaseRow. The code is in: `ImperativeAggCodeGen.generateAccumulatorAccess`.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/api/dataview/ListView.java
twalthr: Yes good point. I will change the interfaces while merging.

twalthr:
Thanks for the explanation @godfreyhe and @JingsongLi. I will merge 92270b10c00f858009e4e2f3b9c4d22315312ec9 with @wuchong's comments.

wuchong:
Commit from 8e1d34d1428e185b50824b18d58e8090ed14e96f to 31471ad687d7e36778a14465c9614d9e34d32b72 for solving class clashes looks good to me. It follows the approach we discussed in the doc. I only have 2 suggestions:

- `org.apache.flink.table.runtime.operators.watermarkassigner` -> `org.apache.flink.table.runtime.operators.wmassigners`
- `org.apache.flink.table.runtime.api.window` -> `org.apache.flink.table.runtime.operators.window`

Regarding to the package name `org.apache.flink.table.planner.calcite`, I agree with @godfreyhe to use `org.apache.flink.table.planner.plan`.

godfreyhe:
@twalthr I have updated the PR. 
It seems that the statistics of changed line number in github is not correct.

twalthr:
Thanks for the update @godfreyhe. The annotations are still not present in 4227b18cecefab6d414548bc2ea73204308c2fe3 but I will add them while merging. I will merge 4227b18cecefab6d414548bc2ea73204308c2fe3 now.

Can @wuchong or @KurtYoung merge the big commit when they think it is ready? Thanks.

wuchong:
Thanks for the update @godfreyhe. The rest commits looks good to me. I will merge them when travis passed. 

@twalthr , I will merge eac67e0a39a44c8a92dc0c4f6e2867e9f6fd90df (move options) too to unblock many issues. We can continue to review config names in #9203.

wuchong:
eac67e0a39a44c8a92dc0c4f6e2867e9f6fd90df to 10b5f418f3ad734d85c3476478fcae1128903707 are merged.

PR #9187, opened 2019-07-20
[FLINK-12595][kinesis] Interrupt thread at right time to avoid deadlock

rehevkor5:
- Inside testOriginalExceptionIsPreservedWhenInterruptedDuringShutdown,
consumerThread.interrupt() was getting absorbed inside
KinesisDataFetcher's while(running) loop, therefore
TestableKinesisDataFetcherForShardConsumerException's awaitTermination()
wasn't getting interrupted by it. This led to deadlock, with
KinesisDataFetcher waiting on the test code to send the interrupt, and
the test code waiting for KinesisDataFetcher to throw the expected
exception.
- Now, the test code waits until KinesisDataFetcher is inside
awaitTermination() before producing the interrupt, so it can be sure
that the interrupt it produces will be received/handled inside
awaitTermination().

To fix the occasional deadlock due to multithreading race condition in unit test for the Kinesis connector.

## Brief change log

  - Prevent deadlock due to race condition in KinesisDataFetcherTest

## Verifying this change

This change is already covered by existing tests, specifically KinesisDataFetcherTest.

  - The serializers: no

## Documentation


tweise:
@flinkbot approve all

PR #9193, opened 2019-07-22
[FLINK-13318][table-planner-blink] Fix Blink planner tests failing on Scala 2.12 by setting SerialVersionUID of UDF explicitly

beyond1920:

Fix Blink planner tests failing on Scala 2.12. Some query plan in Scala 2.12 is different with that in Scala 2.11 if it contains udf without SerialVersionUID explicitly.

## Brief change log

Add SerialVersionUID for all `UserDefinedFunction`s in blink planner which is Scala language.

## Verifying this change
Existed UT, IT

  - The serializers: no
  - The S3 file system connector:  no

## Documentation


beyond1920:
@wuchong Updated title. Thanks

twalthr:
@beyond1920 why are only Blink tests affected by this? We also have Scala functions without version ID. Actually I doubt if we should force users to set this. We are not really using it because the serialization is only for shipping to the cluster.

wuchong:
Hi @twalthr , this is because 

1. we print the udf's md5 into the node explanation.
2. blink planner uses xml to verify plan, the xml content is immutable between different Scala versions. 
3. the serialization behavior is different between Scala 2.11 and 2.12 when version id is not defined. 

That's why we add version id to all existing scala udfs. But this doesn't mean we force users to set this. This is just a workaround for the plan test for different Scala versions. Users can still use a Scala udf without version id in blink planner and flink planner. 

> why are only Blink tests affected by this?

Flink uses `"${func.functionIdentifier}"` to generate plan content, so that it can adapt to different Scala versions automatically.

beyond1920:
+1 to wuchong's opinion.
In fact, we don't need to force users to set serializationId. 
We only need to add SerialVersionUID to UDF in plan tests, in order to generate a stable xml plan.

twalthr:
Thanks for the explanation @wuchong. +1 to the changes from my side.

PR #9195, opened 2019-07-22
[FLINK-13289][table-planner-blink] Blink-planner should setKeyFields to upsert table sink

JingsongLi:

Support upsert table sink in blink-planner

## Verifying this change

TableSinkITCase

  - The serializers: no

## Documentation

### DIFF
--- a/flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/plan/nodes/physical/stream/StreamExecSink.scala
+++ b/flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/plan/nodes/physical/stream/StreamExecSink.scala
@@ -109,7 +109,8 @@ class StreamExecSink[T](
               case Some(keys) => upsertSink.setKeyFields(keys)
               case None if isAppendOnlyTable => upsertSink.setKeyFields(null)
               case None if !isAppendOnlyTable => throw new TableException(
-                "UpsertStreamTableSink requires that Table has a full primary keys if it is updated.")
+                "UpsertStreamTableSink requires that Table has" +
+                    " a full primary keys if it is updated.")
             }
 
             translateToTransformation(withChangeFlag = true, planner)
--- a/flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/plan/nodes/physical/stream/StreamExecSink.scala
+++ b/flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/plan/nodes/physical/stream/StreamExecSink.scala
@@ -97,6 +97,21 @@ class StreamExecSink[T](
             // check for append only table
             val isAppendOnlyTable = UpdatingPlanChecker.isAppendOnly(this)
             upsertSink.setIsAppendOnly(isAppendOnlyTable)
+
+            // extract unique key fields
+            // Now we pick shortest one to sink
+            // TODO UpsertStreamTableSink setKeyFields interface should be Array[Array[String]]
+            val tableKeys: Option[Array[String]] =
+              UpdatingPlanChecker.getUniqueKeys(getInput, planner).sortBy(_.length).headOption
+
+            // check that we have keys if the table has changes (is not append-only)
+            tableKeys match {
+              case Some(keys) => upsertSink.setKeyFields(keys)
+              case None if isAppendOnlyTable => upsertSink.setKeyFields(null)
+              case None if !isAppendOnlyTable => throw new TableException(
+                "UpsertStreamTableSink requires that Table has a full primary keys if it is updated.")
+            }
+
             translateToTransformation(withChangeFlag = true, planner)
 
           case _: AppendStreamTableSink[T] =>
--- a/flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/plan/nodes/physical/stream/StreamExecSink.scala
+++ b/flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/plan/nodes/physical/stream/StreamExecSink.scala
@@ -101,8 +101,12 @@ class StreamExecSink[T](
             // extract unique key fields
             // Now we pick shortest one to sink
             // TODO UpsertStreamTableSink setKeyFields interface should be Array[Array[String]]
-            val tableKeys: Option[Array[String]] =
-              UpdatingPlanChecker.getUniqueKeys(getInput, planner).sortBy(_.length).headOption
+            val tableKeys = {
+              UpdatingPlanChecker.getUniqueKeyFields(getInput, planner) match {
+                case Some(keys) => keys.sortBy(_.length).headOption
+                case None => None
+              }
+            }
 
             // check that we have keys if the table has changes (is not append-only)
             tableKeys match {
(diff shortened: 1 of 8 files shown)


flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/plan/util/UpdatingPlanChecker.scala
```diff
@@ -39,14 +40,21 @@ object UpdatingPlanChecker {
...
-  def getUniqueKeyGroups(plan: RelNode): Option[Seq[(String, String)]] = {
-    val keyExtractor = new UniqueKeyExtractor
-    keyExtractor.visit(plan)
+  def getUniqueKeys(relNode: RelNode, planner: StreamPlanner): Array[Array[String]] = {
+    val rowType = relNode.getRowType
+    val fmq = FlinkRelMetadataQuery.reuseOrCreate(planner.getRelBuilder.getCluster.getMetadataQuery)
+    val uniqueKeys = fmq.getUniqueKeys(relNode)
+    if (uniqueKeys != null) {
+      uniqueKeys.filter(_.nonEmpty).toList.map { uniqueKey =>
+        val keys = new util.HashSet[String]()
```
TsReaper: Why a `HashSet` is needed? It seems that the indices returned by the `ImmutableBitSet` is strictly ascending. Why not just `uniqueKey.asList().toArray`?

JingsongLi: I'll remove `HashSet`.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/plan/utils/UpdatingPlanChecker.scala
```diff
@@ -39,14 +38,16 @@ object UpdatingPlanChecker {
...
-  def getUniqueKeyGroups(plan: RelNode): Option[Seq[(String, String)]] = {
-    val keyExtractor = new UniqueKeyExtractor
-    keyExtractor.visit(plan)
+  def getUniqueKeys(relNode: RelNode, planner: StreamPlanner): Array[Array[String]] = {
+    val rowType = relNode.getRowType
+    val fmq = FlinkRelMetadataQuery.reuseOrCreate(planner.getRelBuilder.getCluster.getMetadataQuery)
+    val uniqueKeys = fmq.getUniqueKeys(relNode)
+    if (uniqueKeys != null) {
+      uniqueKeys.filter(_.nonEmpty).toList
+          .map(_.asList().map(rowType.getFieldNames.get(_)).toArray).toArray
```
wuchong: Can be simplified to `uniqueKeys.filter(_.nonEmpty).map(_.toArray.map(fieldNames.get)).toArray`

flink-table/flink-table-planner-blink/src/test/scala/org/apache/flink/table/runtime/stream/table/TableSinkITCase.scala
```diff
@@ -0,0 +1,701 @@
...
+      fieldTypes: Array[TypeInformation[_]]): TableSink[JTuple2[JBool, Row]] = {
+    val copy = new TestRetractSink
+    copy.fNames = fieldNames
+    copy.fTypes = fieldTypes
+    copy
+  }
+
+}
+
+private[flink] class TestUpsertSink(
```
wuchong: I would like to avoid introducing another set of testing sink (incl. upsert sink, retract sink, append sink). We also have `TestingRetractTableSink`, `TestingUpsertTableSink`, `TestingAppendTableSink`. Can we leverage them in this class? Maybe we need some changes for `TestingUpsertTableSink.setKeyFields/setIsAppendOnly`.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/plan/nodes/physical/stream/StreamExecSink.scala
```diff
@@ -97,6 +97,22 @@ class StreamExecSink[T](
             // check for append only table
             val isAppendOnlyTable = UpdatingPlanChecker.isAppendOnly(this)
             upsertSink.setIsAppendOnly(isAppendOnlyTable)
+
+            // extract unique key fields
+            // Now we pick shortest one to sink
+            // TODO UpsertStreamTableSink setKeyFields interface should be Array[Array[String]]
+            val tableKeys: Option[Array[String]] =
+              UpdatingPlanChecker.getUniqueKeys(getInput, planner).sortBy(_.length).headOption
```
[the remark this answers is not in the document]
JingsongLi: I think we can return `Option[Array[Array[String]]]`, let `sortBy and headOption` still remain in caller, keep the TODO with `UpsertStreamTableSink.setKeyFields`

flink-table/flink-table-planner-blink/src/test/scala/org/apache/flink/table/runtime/stream/table/TableSinkITCase.scala
JingsongLi: Yeah, we can make `TestingUpsertTableSink` support `expectedKeys` and `expectedIsAppendOnly`

JingsongLi:
@wuchong Please take a look again.

flink-table/flink-table-planner-blink/src/test/scala/org/apache/flink/table/planner/runtime/stream/table/TableSinkITCase.scala
```diff
@@ -0,0 +1,524 @@
...
+    val env = StreamExecutionEnvironment.getExecutionEnvironment
+    env.getConfig.enableObjectReuse()
+    env.setStreamTimeCharacteristic(TimeCharacteristic.EventTime)
+    val tEnv = StreamTableEnvironment.create(env)
+
+    val t = env.fromCollection(tupleData3)
+      .assignAscendingTimestamps(_._1.toLong)
+      .toTable(tEnv, 'id, 'num, 'text, 'rowtime.rowtime)
+
+    val sink = new TestingUpsertTableSink(Array(0, 1, 2), TimeZone.getDefault)
```
wuchong: Key array can be "0, 1" to align with "wend", "num"

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/plan/nodes/physical/stream/StreamExecSink.scala
```diff
@@ -97,6 +97,26 @@ class StreamExecSink[T](
             // check for append only table
             val isAppendOnlyTable = UpdatingPlanChecker.isAppendOnly(this)
             upsertSink.setIsAppendOnly(isAppendOnlyTable)
+
+            // extract unique key fields
+            // Now we pick shortest one to sink
+            // TODO UpsertStreamTableSink setKeyFields interface should be Array[Array[String]]
+            val tableKeys = {
+              UpdatingPlanChecker.getUniqueKeyFields(getInput, planner) match {
+                case Some(keys) => keys.sortBy(_.length).headOption
```
wuchong: Should we return the longest instead of the shortest one? So that it can be the same with flink planner when window start and window end are both selected.

PR #9197, opened 2019-07-22
[FLINK-13075][table-planner-blink] Project pushdown rule shouldn't require the TableSource return a modified schema in blink planner

TsReaper:

As the javadoc of `org.apache.flink.table.sources.ProjectableTableSource#projectFields` said, the table schema of the `TableSource` copy must not be modified by this method. However, `PushProjectIntoTableSourceScanRule` in blink planner requires the returning a table source with the modified table schema. This conflict with the javadoc and result in the `TableSource` works for flink planner can't work for blink planner.

This PR fixes `PushProjectIntoTableSourceScanRule` and related class to make it work for blink planner.

## Brief change log

 - Fix `PushProjectIntoTableSourceScanRule` so that the schema of the newly created table source is the same with the original one.

## Verifying this change

This change is already covered by existing tests, such as `TableSourceItCase`.

  - The serializers: no

## Documentation

### DIFF
--- a/flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/plan/schema/TableSourceTable.scala
+++ b/flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/plan/schema/TableSourceTable.scala
@@ -34,8 +34,13 @@ import org.apache.flink.table.sources.{TableSource, TableSourceUtil}
 class TableSourceTable[T](
     val tableSource: TableSource[T],
     val isStreamingMode: Boolean,
-    val statistic: FlinkStatistic)
+    val statistic: FlinkStatistic,
+    val selectedFields: Option[Array[Int]])
   extends FlinkTable {
+
+  def this(tableSource: TableSource[T], isStreamingMode: Boolean, statistic: FlinkStatistic) {
+    this(tableSource, isStreamingMode, statistic, None)
+  }
 
   // TODO implements this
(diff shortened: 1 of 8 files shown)


TsReaper:
Travis passed: https://travis-ci.com/TsReaper/flink/builds/120101168

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/plan/schema/TableSourceTable.scala
```diff
@@ -34,16 +34,21 @@ import org.apache.flink.table.sources.{TableSource, TableSourceUtil}
 class TableSourceTable[T](
     val tableSource: TableSource[T],
     val isStreamingMode: Boolean,
-    val statistic: FlinkStatistic)
+    val statistic: FlinkStatistic,
+    val selectedFields: Option[Array[Int]])
```
wuchong: Can be simplified to `val selectedFields: Option[Array[Int]] = None`. Then we don't need to introduce a new constructor.

JingsongLi: There are Java invokers, so introduce this constructor.

wuchong: Thanks for the tip.

PR #9203, opened 2019-07-23
[FLINK-13375][table-api] Improve config names in ExecutionConfigOptions and OptimizerConfigOptions

wuchong:

This pull request moves `ExecutionConfigOptions` and `OptimizerConfigOptions` into API module. We actually only copy them to API module for reviewing. Will refactor blink planner to use the API options in a following PR.

## Brief change log

- moves `ExecutionConfigOptions` and `OptimizerConfigOptions` into API module. 

## Verifying this change

## Documentation

### DIFF
--- a/flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/config/ExecutionConfigOptions.java
+++ b/flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/config/ExecutionConfigOptions.java
@@ -33,93 +33,88 @@ public class ExecutionConfigOptions {
 	//  Source Options
 	// ------------------------------------------------------------------------
 	public static final ConfigOption<String> TABLE_EXEC_SOURCE_IDLE_TIMEOUT =
-			key("sql.exec.source.idle.timeout")
-					.defaultValue("-1 ms")
-					.withDescription("When a source do not receive any elements for the timeout time, " +
-							"it will be marked as temporarily idle. This allows downstream " +
-							"tasks to advance their watermarks without the need to wait for " +
-							"watermarks from this source while it is idle.");
+		key("table.exec.source.idle-timeout")
+			.defaultValue("-1 ms")
+			.withDescription("When a source do not receive any elements for the timeout time, " +
+				"it will be marked as temporarily idle. This allows downstream " +
+				"tasks to advance their watermarks without the need to wait for " +
+				"watermarks from this source while it is idle.");
 
 	// ------------------------------------------------------------------------
 	//  Sort Options
 	// ------------------------------------------------------------------------
 	public static final ConfigOption<Integer> TABLE_EXEC_SORT_DEFAULT_LIMIT =
-			key("sql.exec.sort.default.limit")
-					.defaultValue(200)
-					.withDescription("Default limit when user don't set a limit after order by. ");
+		key("table.exec.sort.default-limit")
+			.defaultValue(200)
+			.withDescription("Default limit when user don't set a limit after order by.");
 
-	public static final ConfigOption<Integer> TABLE_EXEC_SORT_FILE_HANDLES_MAX_NUM =
-			key("sql.exec.sort.max-num-file-handles")
-					.defaultValue(128)
-					.withDescription("The maximal fan-in for external merge sort. It limits the number of file handles per operator. " +
-							"If it is too small, may cause intermediate merging. But if it is too large, " +
-							"it will cause too many files opened at the same time, consume memory and lead to random reading.");
+	public static final ConfigOption<Integer> TABLE_EXEC_SORT_MAX_NUM_FILE_HANDLES =
+		key("table.exec.sort.max-num-file-handles")
+			.defaultValue(128)
+			.withDescription("The maximal fan-in for external merge sort. It limits the number of file handles per operator. " +
+				"If it is too small, may cause intermediate merging. But if it is too large, " +
+				"it will cause too many files opened at the same time, consume memory and lead to random reading.");
 
 	public static final ConfigOption<Boolean> TABLE_EXEC_SORT_ASYNC_MERGE_ENABLED =
-			key("sql.exec.sort.async-merge.enabled")
-					.defaultValue(true)
-					.withDescription("Whether to asynchronously merge sorted spill files.");
+		key("table.exec.sort.async-merge-enabled")
+			.defaultValue(true)
+			.withDescription("Whether to asynchronously merge sorted spill files.");
 
 	// ------------------------------------------------------------------------
 	//  Spill Options
 	// ------------------------------------------------------------------------
 	public static final ConfigOption<Boolean> TABLE_EXEC_SPILL_COMPRESSION_ENABLED =
-			key("sql.exec.spill.compression.enabled")
-					.defaultValue(true)
-					.withDescription("Whether to compress spilled data. " +
-							"(Now include sort and hash agg and hash join)");
+		key("table.exec.spill-compression.enabled")
+			.defaultValue(true)
+			.withDescription("Whether to compress spilled data. " +
+				"Currently we only support compress spilled data for sort and hash-agg and hash-join operators.");
 
-	public static final ConfigOption<String> TABLE_EXEC_SPILL_COMPRESSION_CODEC =
-			key("sql.exec.spill.compression.codec")
-					.defaultValue("lz4")
-					.withDescription("Use that compression codec to compress spilled file. " +
-							"Now we only support lz4.");
-
-	public static final ConfigOption<Integer> TABLE_EXEC_SPILL_COMPRESSION_BLOCK_SIZE =
-			key("sql.exec.spill.compression.block-size")
-					.defaultValue(64 * 1024)
-					.withDescription("The buffer is to compress. The larger the buffer," +
-							" the better the compression ratio, but the more memory consumption.");
+	public static final ConfigOption<String> TABLE_EXEC_SPILL_COMPRESSION_BLOCK_SIZE =
+		key("table.exec.spill-compression.block-size")
+			.defaultValue("64 kb")
+			.withDescription("The memory size used to do compress when spilling data. " +
+				"The larger the memory, the higher the compression ratio, " +
+				"but more memory resource will be consumed by the job.");
 
 	// ------------------------------------------------------------------------
 	//  Resource Options
 	// ------------------------------------------------------------------------
 
 	public static final ConfigOption<Integer> TABLE_EXEC_RESOURCE_DEFAULT_PARALLELISM =
-			key("sql.resource.default.parallelism")
-					.defaultValue(-1)
-					.withDescription("Default parallelism of job operators. If it is <= 0, use parallelism of StreamExecutionEnvironment(" +
-							"its default value is the num of cpu cores in the client host).");
+		key("table.exec.resource.default-parallelism")
+			.defaultValue(-1)
+			.withDescription("Default parallelism of job operators. If it is <= 0, use parallelism of StreamExecutionEnvironment(" +
+				"its default value is the num of cpu cores in the client host).");
 
 	public static final ConfigOption<Integer> TABLE_EXEC_RESOURCE_SOURCE_PARALLELISM =
-			key("sql.resource.source.parallelism")
-					.defaultValue(-1)
-					.withDescription("Sets source parallelism, if it is <= 0, use " + TABLE_EXEC_RESOURCE_DEFAULT_PARALLELISM.key() + " to set source parallelism.");
+		key("table.exec.resource.source.parallelism")
+			.defaultValue(-1)
+			.withDescription("Sets source parallelism, if it is <= 0, use " + TABLE_EXEC_RESOURCE_DEFAULT_PARALLELISM.key() + " to set source parallelism.");
 
 	public static final ConfigOption<Integer> TABLE_EXEC_RESOURCE_SINK_PARALLELISM =
-			key("sql.resource.sink.parallelism")
-					.defaultValue(-1)
-					.withDescription("Sets sink parallelism, if it is <= 0, use " + TABLE_EXEC_RESOURCE_DEFAULT_PARALLELISM.key() + " to set sink parallelism.");
+		key("table.exec.resource.sink.parallelism")
+			.defaultValue(-1)
+			.withDescription("Sets sink parallelism, if it is <= 0, use " + TABLE_EXEC_RESOURCE_DEFAULT_PARALLELISM.key() + " to set sink parallelism.");
 
-	public static final ConfigOption<Integer> TABLE_EXEC_RESOURCE_EXTERNAL_BUFFER_MEMORY =
-			key("sql.resource.external-buffer.memory.mb")
-					.defaultValue(10)
-					.withDescription("Sets the externalBuffer memory size that is used in sortMergeJoin and overWindow.");
+	public static final ConfigOption<String> TABLE_EXEC_RESOURCE_EXTERNAL_BUFFER_MEMORY =
+		key("table.exec.resource.external-buffer-memory")
+			.defaultValue("10 mb")
+			.withDescription("Sets the external buffer memory size that is used in sort merge join and nested join and over window.");
 
-	public static final ConfigOption<Integer> TABLE_EXEC_RESOURCE_HASH_AGG_MEMORY =
-			key("sql.resource.hash-agg.table.memory.mb")
-					.defaultValue(128)
-					.withDescription("Sets the table memory size of hashAgg operator.");
+	public static final ConfigOption<String> TABLE_EXEC_RESOURCE_HASH_AGG_MEMORY =
+		key("table.exec.resource.hash-agg.memory")
+			.defaultValue("128 mb")
+			.withDescription("Sets the managed memory size of hash aggregate operator.");
 
-	public static final ConfigOption<Integer> TABLE_EXEC_RESOURCE_HASH_JOIN_MEMORY =
-			key("sql.resource.hash-join.table.memory.mb")
-					.defaultValue(128)
-					.withDescription("Sets the HashTable reserved memory for hashJoin operator. It defines the lower limit.");
+	public static final ConfigOption<String> TABLE_EXEC_RESOURCE_HASH_JOIN_MEMORY =
+		key("table.exec.resource.hash-join.memory")
+			.defaultValue("128 mb")
+			.withDescription("Sets the managed memory for hash join operator. It defines the lower limit.");
 
-	public static final ConfigOption<Integer> TABLE_EXEC_RESOURCE_SORT_MEMORY =
-			key("sql.resource.sort.buffer.memory.mb")
-					.defaultValue(128)
-					.withDescription("Sets the buffer memory size for sort.");
+	public static final ConfigOption<String> TABLE_EXEC_RESOURCE_SORT_MEMORY =
+		key("table.exec.resource.sort.memory")
+			.defaultValue("128 mb")
+			.withDescription("Sets the managed buffer memory size for sort operator.");
 
 	// ------------------------------------------------------------------------
 	//  Agg Options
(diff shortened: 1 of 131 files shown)


wuchong:
cc @twalthr @aljoscha @dawidwys . We can review and discuss config option names in this pull request.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/config/ExecutionConfigOptions.java
```diff
@@ -0,0 +1,202 @@
...
+			key("sql.exec.spill.compression.block-size")
+					.defaultValue(64 * 1024)
+					.withDescription("The buffer is to compress. The larger the buffer," +
+							" the better the compression ratio, but the more memory consumption.");
+
+	// ------------------------------------------------------------------------
+	//  Resource Options
+	// ------------------------------------------------------------------------
+
+	public static final ConfigOption<Integer> SQL_RESOURCE_DEFAULT_PARALLELISM =
```
twalthr: All options in this class should be prefixed with `table.exec.`. If not, move them into a separate class?

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/config/ExecutionConfigOptions.java
```diff
@@ -0,0 +1,202 @@
...
+					.defaultValue(-1)
+					.withDescription("Sets sink parallelism, if it is <= 0, use " + SQL_RESOURCE_DEFAULT_PARALLELISM.key() + " to set sink parallelism.");
+
+	public static final ConfigOption<Integer> SQL_RESOURCE_EXTERNAL_BUFFER_MEM =
+			key("sql.resource.external-buffer.memory.mb")
+					.defaultValue(10)
+					.withDescription("Sets the externalBuffer memory size that is used in sortMergeJoin and overWindow.");
+
+	public static final ConfigOption<Integer> SQL_RESOURCE_HASH_AGG_TABLE_MEM =
+			key("sql.resource.hash-agg.table.memory.mb")
```
twalthr: The `.mb` syntax is deprecated. See `org.apache.flink.configuration.TaskManagerOptions#TASK_MANAGER_HEAP_MEMORY` vs `org.apache.flink.configuration.TaskManagerOptions#TASK_MANAGER_HEAP_MEMORY_MB`

wuchong:
Hi @twalthr , I agree with the hierarchical naming. 

Regarding to why choosing the `sql.` prefix, there are 3 reasons:
1. The `sql.` prefix is more widely used in Hadoop ecosystem, for example, `spark.sql.xxx` instead of `spark.dataframe.xxx`. We also want to promote SQL as the first citizen in Table API / SQL.
2. The `table.` prefix  may confuse users what "table" does it mean? The table module or `Table` in Table API or table in database? 
3. We have some options contains `table` in the key, for example: `sql.resource.hash-agg.table.memory.mb`, `sql.optimizer.reuse.table-source.enabled`. If we use `table.` prefix, it may conflict with the "table" in the option name. 

What do you think? @twalthr  @godfreyhe

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/config/ExecutionConfigOptions.java
wuchong: +1 to `.exec.resource`

twalthr:
In my opinion we should distinguish between the SQL language and the table ecosystem.

1. In most presentations, documentation, and in discussions we refer to the "table ecosystem" or "Table API" than "SQL API".
2. This is also represented in the Maven module names. `flink-sql-client` and `flink-sql-parser` are around SQL. `flink-table-XXX` are broader and cover working with the concept of "tables" and relational algebra (independent of SQL).
3. The Table API will receive a lot of new features such as machine learning via `flatMap, flatAggregate` etc. in the future. This has nothing to do with SQL and would confuse if runtime options are named like this.

Any other opinions @dawidwys @aljoscha?

dawidwys:
Generally speaking I also prefer the `table` prefix. I agree with @twalthr that in Flink's naming conventions the module that works with the relational algebra was usually referred as a "table ecosystem" or "Table API", of which the SQL language is part.

Some additional comments regarding the @wuchong's points
Ad.1 I am not aware of any decisions that we want to make the SQL string representation as the go-to interface. I think what we want is for the relational Table API (both sql as well as the java/scala counterparts) to be the first class citizen among different APIs like DataStream/DataSet.
Ad.2 To be honest I could revert the point with just substituting `table.` with `sql.` ;) Does it mean the sql in `.sqlUpdate`? Does it mean the `Table API` or just the sql subpart?
Ad.3 I don't think this is a problem. Moreover I think some of the mentioned options actually abuse the term `table`. E.g. in `sql.resource.hash-agg.table.memory.mb` I think it actually refers to managed memory of the hash aggregate operator. What does the `table` there means? 
On the other hand in the option `sql.optimizer.reuse.table-source.enabled` there is no table, only `table-source`. Theoretically we could even think about dropping the `table-` from `table-source` in that option.

wuchong:
Thanks for your feedbacks @twalthr @dawidwys , I think advanced feature of Table API `flatMap`, `flatAggregate` is a good reason to go with `table.` prefix. 

So I'm fine with the `table.` prefix. Do you have some thoughts on this @godfreyhe ?

godfreyhe:
`sql` is more popular than `table` for blink users before, while  for community users `table` is more popular. so I think `table` is better than `sql`.

wuchong:
Hi @twalthr , @dawidwys , I have updated the pull request.

I also update some option names according to the hierarchical naming convention and examples Timo pointed out. 

But I didn't adopt `.enabled` suffix changes, for example `table.exec.sort.async-merge.enabled` -> `table.exec.sort.async-merge-enabled`. I still keep it `table.exec.sort.async-merge.enabled`. Because `async-merge` is an individual word and may have sub-entries in the future, e.g. threads-num. The same to the other `enabled` options: `table.optimizer.reuse.sub-plan.enabled`, `table.optimizer.reuse.source.enabled`, `table.optimizer.predicate-pushdown.enabled`, `table.optimizer.join-reorder.enabled`. IMO, it will be more readable if separate `enabled` with the word before.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/config/ExecutionConfigOptions.java
```diff
@@ -83,97 +83,97 @@
...
+					.withDescription("Sets the managed memory size of hash aggregate operator.");
 
-	public static final ConfigOption<Integer> SQL_RESOURCE_HASH_JOIN_TABLE_MEM =
-			key("sql.resource.hash-join.table.memory.mb")
-					.defaultValue(128)
-					.withDescription("Sets the HashTable reserved memory for hashJoin operator. It defines the lower limit.");
+	public static final ConfigOption<String> TABLE_EXEC_RESOURCE_HASH_JOIN_TABLE_MEM =
+			key("table.exec.resource.hash-join.memory")
+					.defaultValue("128 mb")
+					.withDescription("Sets the managed memory for hash join operator. It defines the lower limit.");
```
twalthr: nit: could you also update the typos in other descriptions such as `externalBuffer` -> `external buffer`, `sortMergeJoin` -> `sort-merge join` etc.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/config/ExecutionConfigOptions.java
```diff
@@ -83,97 +83,97 @@
...
 					.defaultValue(100 * 1000)
 					.withDescription("Sets the window elements buffer size limit used in group window agg operator.");
 
 	// ------------------------------------------------------------------------
 	//  Async Lookup Options
 	// ------------------------------------------------------------------------
-	public static final ConfigOption<Integer> SQL_EXEC_LOOKUP_ASYNC_BUFFER_CAPACITY =
-			key("sql.exec.lookup.async.buffer-capacity")
+	public static final ConfigOption<Integer> TABLE_EXEC_LOOKUP_ASYNC_BUFFER_CAPACITY =
+			key("table.exec.lookup.async.buffer-capacity")
```
twalthr: `table.exec.lookup.async-buffer-capacity` or just `table.exec.lookup.buffer-capacity`?

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/config/ExecutionConfigOptions.java
```diff
@@ -83,97 +83,97 @@
...
 					.defaultValue(100)
 					.withDescription("The max number of async i/o operation that the async lookup join can trigger.");
 
-	public static final ConfigOption<String> SQL_EXEC_LOOKUP_ASYNC_TIMEOUT =
-			key("sql.exec.lookup.async.timeout")
+	public static final ConfigOption<String> TABLE_EXEC_LOOKUP_ASYNC_TIMEOUT =
+			key("table.exec.lookup.async.timeout")
```
twalthr: `table.exec.lookup.async-timeout` or just `table.exec.lookup.timeout`?

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/config/ExecutionConfigOptions.java
```diff
@@ -83,97 +83,97 @@
...
+							"NOTE: MiniBatch only works for non-windowed aggregations currently. If " + TABLE_EXEC_MINIBATCH_ENABLED.key() +
 							" is set true, its value must be positive.");
 
 	// ------------------------------------------------------------------------
 	//  State Options
 	// ------------------------------------------------------------------------
-	public static final ConfigOption<String> SQL_EXEC_STATE_TTL =
-			key("sql.exec.state.ttl")
+	public static final ConfigOption<String> TABLE_EXEC_STATE_TTL =
+			key("table.exec.state.ttl")
```
twalthr: `table.exec.state-ttl`?

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/config/OptimizerConfigOptions.java
```diff
@@ -30,8 +30,8 @@
 	// ------------------------------------------------------------------------
 	//  Optimizer Options
 	// ------------------------------------------------------------------------
-	public static final ConfigOption<String> SQL_OPTIMIZER_AGG_PHASE_STRATEGY =
-			key("sql.optimizer.agg-phase.strategy")
+	public static final ConfigOption<String> TABLE_OPTIMIZER_AGG_PHASE_STRATEGY =
+			key("table.optimizer.agg-phase.strategy")
```
twalthr: `table.optimizer.agg-phase-strategy`?

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/config/OptimizerConfigOptions.java
```diff
@@ -56,33 +56,33 @@
...
+			key("table.optimizer.distinct-agg.split.bucket-num")
 					.defaultValue(1024)
 					.withDescription("Configure the number of buckets when splitting distinct aggregation. " +
 							"The number is used in the first level aggregation to calculate a bucket key " +
 							"'hash_code(distinct_key) % BUCKET_NUM' which is used as an additional group key after splitting.");
 
-	public static final ConfigOption<Boolean> SQL_OPTIMIZER_REUSE_SUB_PLAN_ENABLED =
-			key("sql.optimizer.reuse.sub-plan.enabled")
+	public static final ConfigOption<Boolean> TABLE_OPTIMIZER_REUSE_SUB_PLAN_ENABLED =
+			key("table.optimizer.reuse.sub-plan.enabled")
```
twalthr: `table.optimizer.reuse-sub-plan-enabled`?

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/config/OptimizerConfigOptions.java
```diff
@@ -56,33 +56,33 @@
...
 					.defaultValue(true)
 					.withDescription("When it is true, optimizer will try to find out duplicated " +
 							"sub-plan and reuse them.");
 
-	public static final ConfigOption<Boolean> SQL_OPTIMIZER_REUSE_TABLE_SOURCE_ENABLED =
-			key("sql.optimizer.reuse.table-source.enabled")
+	public static final ConfigOption<Boolean> TABLE_OPTIMIZER_REUSE_SOURCE_ENABLED =
+			key("table.optimizer.reuse.source.enabled")
```
twalthr: `table.optimizer.reuse-source-enabled`?

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/config/OptimizerConfigOptions.java
```diff
@@ -56,33 +56,33 @@
...
 					.defaultValue(true)
-					.withDescription("When it is true, optimizer will try to find out duplicated table-source and " +
-							"reuse them. This works only when " + SQL_OPTIMIZER_REUSE_SUB_PLAN_ENABLED.key() + " is true.");
+					.withDescription("When it is true, optimizer will try to find out duplicated table source and " +
+							"reuse them. This works only when " + TABLE_OPTIMIZER_REUSE_SUB_PLAN_ENABLED.key() + " is true.");
 
-	public static final ConfigOption<Boolean> SQL_OPTIMIZER_PREDICATE_PUSHDOWN_ENABLED =
-			key("sql.optimizer.predicate-pushdown.enabled")
+	public static final ConfigOption<Boolean> TABLE_OPTIMIZER_PREDICATE_PUSHDOWN_ENABLED =
+			key("table.optimizer.predicate-pushdown.enabled")
```
twalthr: `table.optimizer.predicate-pushdown-enabled`?

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/config/OptimizerConfigOptions.java
```diff
@@ -56,33 +56,33 @@
...
 					.defaultValue(true)
 					.withDescription("If it is true, enable predicate pushdown to the FilterableTableSource. " +
 			"Default value is true.");
 
-	public static final ConfigOption<Boolean> SQL_OPTIMIZER_JOIN_REORDER_ENABLED =
-			key("sql.optimizer.join-reorder.enabled")
+	public static final ConfigOption<Boolean> TABLE_OPTIMIZER_JOIN_REORDER_ENABLED =
+			key("table.optimizer.join-reorder.enabled")
```
twalthr: `table.optimizer.join-reorder-enabled`?

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/config/ExecutionConfigOptions.java
```diff
@@ -83,97 +83,97 @@
...
+	public static final ConfigOption<Integer> TABLE_EXEC_RESOURCE_DEFAULT_PARALLELISM =
+			key("table.exec.resource.default-parallelism")
 					.defaultValue(-1)
 					.withDescription("Default parallelism of job operators. If it is <= 0, use parallelism of StreamExecutionEnvironment(" +
 							"its default value is the num of cpu cores in the client host).");
 
-	public static final ConfigOption<Integer> SQL_RESOURCE_SOURCE_PARALLELISM =
-			key("sql.resource.source.parallelism")
+	public static final ConfigOption<Integer> TABLE_EXEC_RESOURCE_SOURCE_PARALLELISM =
+			key("table.exec.resource.source.parallelism")
```
twalthr: `source-parallelism`, `sink-parallelism`, `external-buffer-memory`, `hash-agg-memory` etc.? A hierarchy with one child is not necessary.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/config/OptimizerConfigOptions.java
```diff
@@ -56,33 +56,33 @@
...
 					.defaultValue(false)
 					.withDescription("Enables join reorder in optimizer cbo. Default is disabled.");
```
twalthr: what is "cbo"?

wuchong:
Thanks @twalthr for the detailed review. 

Regarding to the `.enabled` options, I left some thoughts above. I'm afraid we will have more sub-entries for those `.enabled` options. Take join-reorder as an example, spark has `spark.sql.cbo.joinReorder.enabled`, `spark.sql.cbo.joinReorder.dp.threshold`, `spark.sql.cbo.joinReorder.card.weight`. That's why I want to keep `.enabled`. And I think a hierarchy with one child can be valid as long as it is a reasonable word and may have sub-entries in the future.

What do you think/

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/config/ExecutionConfigOptions.java
wuchong: Currently, we have two implementation for lookup a dimension table, one is synchronically, the other is asynchronously. So I think `async` is needed here. I'm not sure whether we will have any options for synchronically lookup. So I would still +1 to `table.exec.async.xxx`.

twalthr:
I think if we add a hierarchy in advance then it should be aligned by operator and not by functionality. 

For example, instead of `table.optimizer.join-reorder.enabled` how about `table.optimizer.join.reorder-enabled`? Such that all join related config is grouped. Or instead of `table.optimizer.predicate-pushdown-enabled` how about `table.optimizer.source.predicate-pushdown-enabled`? 

We should definitely avoid having more then 4 levels because it becomes hart to read.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/config/ExecutionConfigOptions.java
wuchong: Or how about `table.exec.async-lookup.buffer-capacity` ?

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/config/OptimizerConfigOptions.java
godfreyhe: cost based optimizer (e.g. VolcanoPlanner in Calcite)
another concept is "rbo", means rule based optimizer (e.g. HepPlanner in Calcite)

wuchong: @godfreyhe , is it only enabled in CBO? I mean, users shouldn't care about what optimizer is used in Flink.

wuchong:
FLINK-13266 is just merged. We will keep the conflicts to make the reviewing easier. Once we reach a consensus on each config names. I will rebase master and adapt blink planner to the new options.

wuchong:
Hi @twalthr , I agree we should keep config names short. However, IMO, 4 levels is too strict, because in fact, we uses only have 2 levels to describe a config. There are some configs are difficult to squash to 4 levels. For example: `table.optimizer.distinct-agg.split.bucket-num` and `table.optimizer.distinct-agg.split.enabled`. I think maybe we can relax the restrictions to 5 levels?

Regarding to the aligned by operator vs functionality, I would like to align by functionality. Because some options are not relative to a specific operator, e.g `table.optimizer.reuse.sub-plan.enabled`. And some functionality words are well-known, e.g. `hash-agg`, `join-reorder`.

Regarding to the `.enabled`, my concern is 
1. it would be better to align all `enabled` options in the same style
2. using `.enabled` is more flexible for introducing new options under the same parent.

> we don't need a hierarchy because they just consist of a single child.

I think it's fine to have a hierarchy even if it consist of a single child. Because we can't enumerate all the configs at one blow. We should leave some space for the future. And the [Flink convention](https://docs.google.com/document/d/1JGRVK6udBFa8xbyv7ipPQiLEhxebA64f8KKcaAH0UPY/edit#) doesn't restrict this, right?

What do you think? @twalthr @godfreyhe

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/config/ExecutionConfigOptions.java
twalthr: Sounds good.

godfreyhe:
I think readability and extendibility of a config name is more important than strict level. From this point of view, I also tend to align by functionality instead of operator. e.g. `lookup` has `async` and `sync` mode, I think `table.exec.async-lookup.XXX` and `table.exec.sync-lookup.XXX` is more clear than `table.exec.lookup.async-XXX` and `table.exec.lookup.sync-XXX`.

about `enabled`,  I think the readability is almost the same about `table.exec.sort.async-merge.enabled` and `table.exec.sort.async-merge-enabled`, however the extendibility of the first one is better than second one.

twalthr:
5 Levels for special cases like `table.optimizer.distinct-agg.split.bucket-num` is fine with me.

However, I think if we introduce new options under a common key that needs a hierarchy, we could still deprecate the old flat key and introduce the hierarchy. Introducing hierarchy "just in case" should be avoided.

If options are not relative to a specific operator, they should become top-level properties: `table.optimizer.reuse-sub-plan-enabled` instead of `table.optimizer.reuse.sub-plan.enabled`.

I know this topic is typical bikeshedding so I don't want to block too much here. But it would be great if you could go through the list one more time and check whether a hierarchy makes sense. Thanks.

wuchong:
Thanks for the feedbacks. I have update the config names again. 
Welcome to have another look at it @twalthr @godfreyhe @dawidwys .

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/config/ExecutionConfigOptions.java
```diff
@@ -0,0 +1,202 @@
...
+	// ------------------------------------------------------------------------
+	//  MiniBatch Options
+	// ------------------------------------------------------------------------
+
+	public static final ConfigOption<Boolean> TABLE_EXEC_MINIBATCH_ENABLED =
+			key("table.exec.mini-batch.enabled")
+					.defaultValue(false)
+					.withDescription("Specifies whether to enable MiniBatch optimization. " +
+							"MiniBatch is an optimization to buffer input records to reduce state access. " +
+							"This is disabled by default. To enable this, users should set this config to true.");
```
godfreyhe: add description: "NOTE: If this is true, TABLE_EXEC_MINIBATCH_ALLOW_LATENCY.key() and TABLE_EXEC_MINIBATCH_SIZE.key() must be greater than zero."

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/config/ExecutionConfigOptions.java
```diff
@@ -0,0 +1,202 @@
...
+					.withDescription("Specifies a minimum time interval for how long idle state " +
+							"(i.e. state which was not updated), will be retained. State will never be " +
+							"cleared until it was idle for less than the minimum time, and will be cleared " +
+							"at some time after it was idle. Default is never clean-up the state.\n" +
+							"NOTE: Cleaning up state requires additional overhead for bookkeeping.");
+
+	// ------------------------------------------------------------------------
+	//  Other Exec Options
+	// ------------------------------------------------------------------------
+	public static final ConfigOption<String> TABLE_EXEC_DISABLED_OPERATORS =
```
godfreyhe: list current supported operator names ?

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/config/OptimizerConfigOptions.java
```diff
@@ -0,0 +1,88 @@
...
+			key("table.optimizer.join.broadcast-threshold")
+					.defaultValue(1024 * 1024L)
+					.withDescription("Configures the maximum size in bytes for a table that will be broadcast to all worker " +
+							"nodes when performing a join. By setting this value to -1 to disable broadcasting.");
+
+	public static final ConfigOption<Boolean> TABLE_OPTIMIZER_DISTINCT_AGG_SPLIT_ENABLED =
+			key("table.optimizer.distinct-agg.split.enabled")
+					.defaultValue(false)
+					.withDescription("Tells the optimizer whether to split distinct aggregation " +
+							"(e.g. COUNT(DISTINCT col), SUM(DISTINCT col)) into two level. " +
```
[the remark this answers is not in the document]
wuchong: This is on purpose. The two level is different two phase in `table.optimizer.agg-phase-strategy`. They are orthogonal.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/config/OptimizerConfigOptions.java
```diff
@@ -0,0 +1,88 @@
...
+							"The first aggregation is shuffled by an additional key which is calculated using " +
+							"the hashcode of distinct_key and number of buckets. This optimization is very useful " +
+							"when there is data skew in distinct aggregation and gives the ability to scale-up the job. " +
+							"Default is false.");
+
+	public static final ConfigOption<Integer> TABLE_OPTIMIZER_DISTINCT_AGG_SPLIT_BUCKET_NUM =
+			key("table.optimizer.distinct-agg.split.bucket-num")
+					.defaultValue(1024)
+					.withDescription("Configure the number of buckets when splitting distinct aggregation. " +
+							"The number is used in the first level aggregation to calculate a bucket key " +
```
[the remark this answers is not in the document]
wuchong: The same as above.

wuchong:
Thanks for the reviewing @godfreyhe . I have updated the PR.
@twalthr do you want to go through it again?

wuchong:
Thanks @twalthr . I will adapt blink planner to the new config names then.

wuchong:
Hi @JingsongLi @godfreyhe , I have updated the pull request to adapt blink planner to the new config names. In order for easy reviewing, I split it into 2 commits:
1. the first one is just renaming config field names. It touches many files but all is renaming.
2. the second one is renaming config key names, improving description, changing `Integer` option -> `String` option (memory size).

Feel free to leave comments. :)

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/config/ExecutionConfigOptions.java
```diff
@@ -128,80 +123,71 @@
...
-							"cleared until it was idle for less than the minimum time, and will be cleared " +
-							"at some time after it was idle. Default is never clean-up the state.\n" +
-							"NOTE: Cleaning up state requires additional overhead for bookkeeping.");
+	public static final ConfigOption<Boolean> TABLE_EXEC_MINIBATCH_ENABLED =
+		key("table.exec.mini-batch.enabled")
+			.defaultValue(false)
+			.withDescription("Specifies whether to enable MiniBatch optimization. " +
+				"MiniBatch is an optimization to buffer input records to reduce state access. " +
+				"This is disabled by default. To enable this, users should set this config to true. " +
+				"NOTE: If mini-batch is enabled, 'table.exec.mini-batch.allow-latency' and " +
```
godfreyhe: use `TABLE_EXEC_MINIBATCH_ALLOW_LATENCY.key()` instead of `table.exec.mini-batch.allow-latency` ?  and `table.exec.mini-batch.size`

wuchong: We can't do that. Java will complain "Illegal forward reference" for that.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/plan/rules/physical/batch/BatchExecSortRule.scala
```diff
@@ -81,8 +81,8 @@ object BatchExecSortRule {
 
   // It is a experimental config, will may be removed later.
   @Experimental
-  val SQL_EXEC_SORT_RANGE_ENABLED: ConfigOption[JBoolean] =
-  key("sql.exec.sort.range.enabled")
+  val TABLE_EXEC_SORT_RANGE_ENABLED: ConfigOption[JBoolean] =
+  key("table.exec.sort.range.enabled")
```
godfreyhe: -> `table.exec.range-sort.enabled` ?

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/plan/nodes/physical/stream/StreamExecSort.scala
```diff
@@ -146,8 +146,8 @@ object StreamExecSort {
 
   // It is a experimental config, will may be removed later.
   @Experimental
-  val SQL_EXEC_SORT_NON_TEMPORAL_ENABLED: ConfigOption[JBoolean] =
-  key("sql.exec.sort.non-temporal.enabled")
+  val TABLE_EXEC_SORT_NON_TEMPORAL_ENABLED: ConfigOption[JBoolean] =
+  key("table.exec.sort.non-temporal-enabled")
```
godfreyhe: -> `table.exec.non-temporal-sort.enabled` ?

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/plan/nodes/physical/stream/StreamExecRank.scala
```diff
@@ -246,8 +246,8 @@ object StreamExecRank {
 
   // It is a experimental config, will may be removed later.
   @Experimental
-  val SQL_EXEC_TOPN_CACHE_SIZE: ConfigOption[JLong] =
-  key("sql.exec.topn.cache.size")
+  val TABLE_EXEC_TOPN_CACHE_SIZE: ConfigOption[JLong] =
+  key("table.exec.topn.cache.size")
```
godfreyhe: -> `table.exec.topn.cache-size` ?

wuchong:
Thanks for the reviewing @godfreyhe , I will modify them when merging because they are tiny changes.

PR #9207, opened 2019-07-23
[FLINK-9526][e2e] Fix unstable BucketingSink end-to-end test

ifndef-SleePy:

* Fix unstable end-to-end test for `BucketingSink`
* The root cause of this unstable case is an unexpected debug log involved. This case sets the log level to `DEBUG` for checking the truncating log. However it sets the global log level. There might be a debug log "Responding with error: ..." in `PartitionRequestServerHandler` if a `PartitionNotFound` occurred. This error would fail the error checking due to there is a word "error" in this log.

## Brief change log

* Shrink the `DEBUG` log level to `BucketingSink` only, not global anymore

## Verifying this change

* This is a test case fixing. And I have tested locally.

  - The serializers: no

## Documentation

### DIFF
--- a/flink-end-to-end-tests/test-scripts/test_streaming_bucketing.sh
+++ b/flink-end-to-end-tests/test-scripts/test_streaming_bucketing.sh
@@ -22,7 +22,9 @@ source "$(dirname "$0")"/common.sh
 TEST_PROGRAM_JAR=${END_TO_END_DIR}/flink-bucketing-sink-test/target/BucketingSinkTestProgram.jar
 
 # enable DEBUG logging level to retrieve truncate length later
-sed -i -e 's/#log4j.logger.org.apache.flink=INFO/log4j.logger.org.apache.flink=DEBUG/g' $FLINK_DIR/conf/log4j.properties
+# make sure it would always start a new line
+echo "" >> $FLINK_DIR/conf/log4j.properties
+echo "log4j.logger.org.apache.flink.streaming.connectors.fs.bucketing.BucketingSink=DEBUG" >> $FLINK_DIR/conf/log4j.properties
 
 set_conf_ssl
 start_cluster


ifndef-SleePy:
It seems that the Travis building hanging for a while. Would rebase latest master branch and trigger Travis building again.

flink-end-to-end-tests/test-scripts/test_streaming_bucketing.sh
```diff
@@ -22,7 +22,8 @@ source "$(dirname "$0")"/common.sh
 TEST_PROGRAM_JAR=${END_TO_END_DIR}/flink-bucketing-sink-test/target/BucketingSinkTestProgram.jar
 
 # enable DEBUG logging level to retrieve truncate length later
-sed -i -e 's/#log4j.logger.org.apache.flink=INFO/log4j.logger.org.apache.flink=DEBUG/g' $FLINK_DIR/conf/log4j.properties
+echo "" >> $FLINK_DIR/conf/log4j.properties
+echo "log4j.logger.org.apache.flink.streaming.connectors.fs.bucketing.BucketingSink=DEBUG, file" >> $FLINK_DIR/conf/log4j.properties
```
kl0u: The only thing needed is the `echo "log4j.logger.org.apache.flink.streaming.connectors.fs.bucketing.BucketingSink=DEBUG`.

The appender is already inherited by the root logger, so the `, file` is redundant, and the `echo "" >> $FLINK_DIR/conf/log4j.properties` does not add any additional information, so it can be removed.

ifndef-SleePy: 1. Good catch of `, file`
2. `echo "" >> $FLINK_DIR/conf/log4j.properties` is to make sure `log4j.logger.org.apache.flink.streaming.connectors.fs.bucketing.BucketingSink=DEBUG` would be always in a new line. If case of there is no `\n` at the end of log4j.properties. Although there is a `\n` now, but I don't want to depend on it. So I prefer to keep this empty line, what do you think?

ifndef-SleePy:
Hi @kl0u , thank you for feedback.
I have removed the redundant appender.

flink-end-to-end-tests/test-scripts/test_streaming_bucketing.sh
kl0u: @ifndef-SleePy I do not have a strong opinion on keeping it or not. Either way it does not add or remove any important information so we can as well keep it.

PR #9208, opened 2019-07-23
[FLINK-13378][table-planner-blink] Fix bug: Blink-planner not support SingleValueAggFunction

JingsongLi:

There some problem in SingleValueAggFunction:
1.The comment in the method is "value = count == 0 ? exception : operand(0)", but actually it need to be "value = count > 0 ? exception : operand(0)" according to the code and logic.
2.The throwException expression should call(THROW_EXCEPTION, literal(msg), typeLiteral(type)).

And there are some bugs to support it in blink-planner:
1.RexNodeConverter lack converter for THROW_EXCEPTION
2.PlannerExpressions lack expr to support type inference for THROW_EXCEPTION

## Verifying this change

ScalarQueryITCase

  - The serializers: no

## Documentation

### DIFF
--- a/flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/expressions/ExpressionBuilder.java
+++ b/flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/expressions/ExpressionBuilder.java
@@ -140,6 +140,6 @@ public class ExpressionBuilder {
 	}
 
 	public static Expression throwException(String msg, DataType type) {
-		return call(THROW_EXCEPTION, typeLiteral(type));
+		return call(THROW_EXCEPTION, literal(msg), typeLiteral(type));
 	}
 }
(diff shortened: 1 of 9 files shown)


flink-table/flink-table-planner-blink/src/test/scala/org/apache/flink/table/runtime/batch/sql/join/ScalarQueryITCase.scala
```diff
@@ -46,6 +49,18 @@ class ScalarQueryITCase extends BatchTestBase {
...
-}
+  @Before
+  override def before(): Unit = {
+    super.before()
+    registerCollection("l", l, INT_DOUBLE, "a, b")
+    registerCollection("r", r, INT_DOUBLE, "c, d")
+  }
 
+  @Test
+  def testScalarQuery(): Unit = {
```
TsReaper: Also add a validation test for a multiple row agg?

JingsongLi: What is multiple row agg?
You mean use SingleValueAgg directly in sql? If that is what you mean, I prefer not add. Because SingleValueAgg is just add for scalar sub query, we don't recommend user to use in sql.

TsReaper: No. I mean something like `SELECT * FROM l WHERE a = (SELECT c FROM r WHERE c > 3)`, so that the inner select produces more than 1 record. We should expect an validation failure on this case.

JingsongLi: Got it, I will rebase and add it.

PR #9211, opened 2019-07-24
[FLINK-13393][FLINK-13391][table-planner-blink] Fix source conversion and source return type

JingsongLi:

1.Blink-planner should use conversion class of DataType
2.Blink-planner should not invoke deprecated getReturnType of TableSource

## Verifying this change

TableSourceITCase

  - The serializers: no

## Documentation

### DIFF
--- a/flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/plan/nodes/physical/PhysicalTableSourceScan.scala
+++ b/flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/plan/nodes/physical/PhysicalTableSourceScan.scala
@@ -18,9 +18,15 @@
 
 package org.apache.flink.table.planner.plan.nodes.physical
 
+import org.apache.flink.api.common.io.InputFormat
+import org.apache.flink.api.common.typeinfo.TypeInformation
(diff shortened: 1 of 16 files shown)


flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/plan/nodes/physical/PhysicalTableSourceScan.scala
```diff
@@ -52,4 +61,19 @@ abstract class PhysicalTableSourceScan(
     super.explainTerms(pw).item("fields", getRowType.getFieldNames.asScala.mkString(", "))
   }
 
+  def getSourceTransformation(
+      streamEnv: StreamExecutionEnvironment): Transformation[_] = {
+    if (sourceTransform == null) {
+      sourceTransform = tableSource match {
+        case source: InputFormatTableSource[_] =>
```
wuchong: I think the problem is `InputFormatTableSource#getDataStream` shouldn't call `getReturnType`, the following code should be moved to there.

wuchong:
However, I can do the change and verify in my repo travis and merge this.

PR #9212, opened 2019-07-24
[FLINK-13338][table-api] Sql conformance is hard to config in TableConfig

danny0405:

This patch adds an interface `TableConfig#setSqlDialect(SqlDialect)` to make the sql dialect configuration more user friendly.

## Brief change log

  - Add new class `SqlDialect` to enumerate the sql dialects Flink supports now
  - Add configuration interface for sql dialect in `TableConfig`

## Verifying this change

See tests in SqlToOperationConverterTest, PartitionableSinkITCase.

  - The serializers: no

## Documentation

### DIFF
--- a/flink-python/pyflink/common/sql_dialect.py
+++ b/flink-python/pyflink/common/sql_dialect.py
@@ -24,11 +24,11 @@ class SqlDialect(object):
     """
     Enumeration of valid SQL compatibility modes.
 
-    For most of the cases, the built-in compatibility mode will suffice. For some features,
+    In most of the cases, the built-in compatibility mode will suffice. For some features,
     i.e. the "create partitionable table" grammar is only supported in Hive dialect, you may need
     to switch to the Hive dialect if that is your purpose.
 
-    There may introduce other sql dialects in the future.
+    We may introduce other sql dialects in the future.
 
     :data:`DEFAULT`:
 
@@ -36,7 +36,9 @@ class SqlDialect(object):
 
     :data:`HIVE`:
 
-    SQL dialect that allows some Apache HIVE specific grammar.
+    SQL dialect that allows some Apache Hive specific grammar.
+    Note: We might never support all of the Hive grammar.
+    See the documentation for supported features.
     """
 
     DEFAULT = 0
--- a/flink-python/pyflink/common/sql_dialect.py
+++ b/flink-python/pyflink/common/sql_dialect.py
@@ -1,0 +1,66 @@
+################################################################################
+#  Licensed to the Apache Software Foundation (ASF) under one
+#  or more contributor license agreements.  See the NOTICE file
+#  distributed with this work for additional information
+#  regarding copyright ownership.  The ASF licenses this file
+#  to you under the Apache License, Version 2.0 (the
+#  "License"); you may not use this file except in compliance
+#  with the License.  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+# limitations under the License.
+################################################################################
+from pyflink.java_gateway import get_gateway
+
+__all__ = ['SqlDialect']
+
+
+class SqlDialect(object):
+    """
+    Enumeration of valid SQL compatibility modes.
+
+    For most of the cases, the built-in compatibility mode will suffice. For some features,
+    i.e. the "create partitionable table" grammar is only supported in Hive dialect, you may need
+    to switch to the Hive dialect if that is your purpose.
+
+    There may introduce other sql dialects in the future.
+
+    :data:`DEFAULT`:
+
+    Flink's default SQL behavior.
+
+    :data:`HIVE`:
+
+    SQL dialect that allows some Apache HIVE specific grammar.
+    """
+
+    DEFAULT = 0
+    HIVE = 1
+
+    @staticmethod
+    def _from_j_sql_dialect(j_sql_dialect):
+        gateway = get_gateway()
+        JSqlDialect = gateway.jvm.org.apache.flink.table.api.SqlDialect
+        if j_sql_dialect == JSqlDialect.DEFAULT:
+            return SqlDialect.DEFAULT
+        elif j_sql_dialect == JSqlDialect.HIVE:
+            return SqlDialect.HIVE
+        else:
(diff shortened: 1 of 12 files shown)


wuchong:
cc @twalthr

flink-table/flink-table-planner-blink/src/test/java/org/apache/flink/table/sqlexec/SqlToOperationConverterTest.java
```diff
@@ -0,0 +1,155 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.assertEquals;
+
+/** Test cases for SqlExecutableStatement. **/
+public class SqlToOperationConverterTest {
+	private static final EnvironmentSettings settings = EnvironmentSettings.newInstance()
```
twalthr: We should reduce the dependencies of tests to a minimum. Why does this test needs access to a `TableEnvironment`? Could you think of a test that really just tests the `SqlToOperationConverter`?

flink-table/flink-table-planner/src/main/java/org/apache/flink/table/planner/PlanningConfigurationBuilder.java
```diff
@@ -155,11 +156,23 @@ public Context getContext() {
...
 	}
 
+	private FlinkSqlConformance getSqlConformance() {
+		SqlDialect sqlDialect = tableConfig.getSqlDialect();
+		switch (sqlDialect) {
+		case HIVE:
+			return FlinkSqlConformance.HIVE;
+		case DEFAULT:
+			return FlinkSqlConformance.DEFAULT;
+		default:
```
twalthr: Do not add a default. Because otherwise people might forget to update this switch once a new dialect is added.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/SqlDialect.java
```diff
@@ -0,0 +1,42 @@
...
+import org.apache.flink.annotation.PublicEvolving;
+
+/**
+ * Enumeration of valid SQL compatibility modes.
+ *
+ * <p>For most of the cases, the built-in compatibility mode will suffice. For some features,
+ * i.e. the "create partitionable table" grammar is only supported in Hive dialect, you may need
+ * to switch to the Hive dialect if that is your purpose.
+ *
+ * <p>There may introduce other sql dialects in the future.
```
twalthr: "We may introduce other SQL dialects in the future."

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/SqlDialect.java
```diff
@@ -0,0 +1,42 @@
...
+package org.apache.flink.table.api;
+
+import org.apache.flink.annotation.PublicEvolving;
+
+/**
+ * Enumeration of valid SQL compatibility modes.
+ *
+ * <p>For most of the cases, the built-in compatibility mode will suffice. For some features,
+ * i.e. the "create partitionable table" grammar is only supported in Hive dialect, you may need
+ * to switch to the Hive dialect if that is your purpose.
```
twalthr: "if that is your purpose." -> "if required."

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/SqlDialect.java
```diff
@@ -0,0 +1,42 @@
...
+ * <p>There may introduce other sql dialects in the future.
+ */
+@PublicEvolving
+public enum SqlDialect {
+	/**
+	 * Flink's default SQL behavior.
+	 */
+	DEFAULT,
+	/**
+	 * SQL dialect that allows some Apache HIVE specific grammar.
```
twalthr: "HIVE" -> "Hive"
Add "Note: We might never support all of the Hive grammar. See the documentation for supported features."

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/SqlDialect.java
```diff
@@ -0,0 +1,42 @@
...
+ */
+
+package org.apache.flink.table.api;
+
+import org.apache.flink.annotation.PublicEvolving;
+
+/**
+ * Enumeration of valid SQL compatibility modes.
+ *
+ * <p>For most of the cases, the built-in compatibility mode will suffice. For some features,
```
twalthr: "For most of the cases" -> "In most of the cases"

flink-table/flink-table-planner-blink/src/test/java/org/apache/flink/table/sqlexec/SqlToOperationConverterTest.java
danny0405: Because the `SqlToOperationConverter ` needs the `FlinkPlannerImpl` which usually comes from a TableEnvironment, also i want to test that the `TableConfig#setSqlDialect` did work well. There is no much overhead here.

twalthr: There is the overhead of discovering a planner and executor. The test could also be based on the `Planner` only with a mocking catalog. We should always reduce the dependencies of tests to a minimum because otherwise we need to change many tests if we perform changes to the API class. This test must not use the API but can go at least one level deeper.

danny0405: Agreed and updated, thanks.

danny0405:
Hi @twalthr , i have updated for your review address, please review again if you have time, thanks ~

flink-python/pyflink/common/sql_dialect.py
```diff
@@ -24,19 +24,21 @@ class SqlDialect(object):
     """
     Enumeration of valid SQL compatibility modes.
 
-    For most of the cases, the built-in compatibility mode will suffice. For some features,
+    In most of the cases, the built-in compatibility mode will suffice. For some features,
     i.e. the "create partitionable table" grammar is only supported in Hive dialect, you may need
```
twalthr: I just checked the Hive documentation. Such a statement does not exist in Hive, is that correct?

danny0405: For "create partitionable table" i actually mean create a partitionable table, it is not really a sql grammar, maybe i should remove the quotes.

twalthr: Thanks for the clarification @danny0405. I thought you mean some grammar. I will merge this PR now.

twalthr:
Merging...

PR #9224, opened 2019-07-25
[FLINK-13226] [connectors / kafka] Fix race condition between transaction commit and produc…

becketqin:
…er closure.

This patch fixes a race condition between the checkpointing thread and main thread. The sequence causing the deadlock is the following:
1. In `FlinkKafkaProducer`, the main thread encounters a problem and closes all the producer to start failover.
2. The previous checkpoint has completed, so the checkpointing thread grabs the checkpoint lock and tries to commit the transaction on the producer that has been closed in step 1. This commit will never succeed due to [KAFKA-6635](https://issues.apache.org/jira/browse/KAFKA-6635). So the checkpoint thread blocks forever.
3. In `StreamTask`, the main thread will eventually try to release all the record writer. To do that, it attempts to grab the checkpoint lock which is hold by checkpoint thread in step 2 and will never be released. So the main thread also blocks forever.

KAFKA-6635 has been fixed in Kafka 2.3.0. But Flink 1.9 does not rely on that yet, and we also support Kafka 0.11. So we are just going to fix on the Flink side first. The solution is to make sure that in `FlinkKafkaProducer` any operation relying on the underlying sender thread to finish throws an exception if the producer is closed.

This patch also fixes a minor issue of duplicated static inner class name in `KafkaConsumerTestBase`.

## Brief change log
- Make `FlinkKafkaProducer` and `FlinkKafkaInternalProducer` thread safe.
- Fix static inner class name collision in `KafkaConsumerTestBase`.

## Verifying this change

- Added test to ensure exception will be thrown if a blocking method is called on `FlinkKafkaProducer` and `FlinkKafkaInternalProducer` after the producer is closed.

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-connectors/flink-connector-kafka-0.11/src/test/java/org/apache/flink/streaming/connectors/kafka/FlinkKafkaProducerITCase.java
+++ b/flink-connectors/flink-connector-kafka-0.11/src/test/java/org/apache/flink/streaming/connectors/kafka/FlinkKafkaProducerITCase.java
@@ -101,6 +101,66 @@ public class FlinkKafkaProducerITCase extends KafkaTestBase {
 		deleteTestTopic(topicName);
 	}
 
+	@Test(timeout = 30000L, expected = IllegalStateException.class)
+	public void testPartitionsForAfterClosed() {
+		FlinkKafkaProducer<String, String> kafkaProducer = new FlinkKafkaProducer<>(extraProperties);
+		kafkaProducer.close();
+		kafkaProducer.partitionsFor("Topic");
+	}
+
+	@Test(timeout = 30000L, expected = IllegalStateException.class)
+	public void testInitTransactionsAfterClosed() {
+		FlinkKafkaProducer<String, String> kafkaProducer = new FlinkKafkaProducer<>(extraProperties);
+		kafkaProducer.close();
+		kafkaProducer.initTransactions();
+	}
+
+	@Test(timeout = 30000L, expected = IllegalStateException.class)
+	public void testBeginTransactionAfterClosed() {
+		FlinkKafkaProducer<String, String> kafkaProducer = new FlinkKafkaProducer<>(extraProperties);
+		kafkaProducer.initTransactions();
+		kafkaProducer.close();
+		kafkaProducer.beginTransaction();
+	}
+
+	@Test(timeout = 30000L, expected = IllegalStateException.class)
+	public void testCommitTransactionAfterClosed() {
+		String topicName = "testCommitTransactionAfterClosed";
+		FlinkKafkaProducer<String, String> kafkaProducer = getClosedProducer(topicName);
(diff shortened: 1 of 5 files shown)


flink-connectors/flink-connector-kafka-0.11/src/test/java/org/apache/flink/streaming/connectors/kafka/FlinkKafkaProducerITCase.java
```diff
@@ -101,6 +101,62 @@ public void testResumeTransaction() throws IOException {
...
+		kafkaProducer.close();
+		assertThrows(kafkaProducer::initTransactions, IllegalStateException.class);
+	}
+
+	@Test(timeout = 30000L)
+	public void testBeginTransactionAfterClosed() throws Exception {
+		FlinkKafkaProducer<String, String> kafkaProducer = new FlinkKafkaProducer<>(extraProperties);
+		kafkaProducer.initTransactions();
+		kafkaProducer.close();
+		assertThrows(kafkaProducer::beginTransaction, IllegalStateException.class);
```
pnowojski: nit: maybe use `org.junit.rules.ExpectedException`?

flink-connectors/flink-connector-kafka-base/src/test/java/org/apache/flink/streaming/connectors/kafka/KafkaConsumerTestBase.java
```diff
@@ -2264,7 +2264,7 @@ public String getTargetTopic(Tuple3<Integer, Integer, String> element) {
 		}
 	}
 
-	private static class TestDeSerializer extends TestDeserializer
+	private static class TestDeSerializer extends AbstractTestDeserializer
```
pnowojski: nit: rename `TestDeSerializer ` -> `TestDeserializer`?

becketqin:
@pnowojski Thanks for the review. I have updated the patch. I'll push the patch to release-1.9 branch and create another PR to simply bump Kafka dependency in the universal Kafka connector to 2.3.0.

pnowojski:
@becketqin , could you check if after this patch we can re-enable the tests in `org.apache.flink.streaming.connectors.kafka.KafkaProducerExactlyOnceITCase`? (and maybe the equivalent of 0.11 connector)?

davidradl:
@pnowojski  I see https://github.com/apache/flink/pull/22797 , this pr is no longer relevant to this repository, I suggest closing this pr and porting the change to the [new Flink connector kafka repo]( https://github.com/apache/flink-connector-kafka) if it is still relevant.

MartijnVisser:
@becketqin Same story as the other 2 PRs :) if this code change is still relevant, please open the PR in https://github.com/apache/flink-connector-kafka

PR #9229, opened 2019-07-25
 [FLINK-13279][table-sql-client] Fully qualify sink name in sql-client 

dawidwys:

This PR fixes the problem that sink registered automatically via sql-client is not accessible when the current catalog or database is changed to something different than the built-in ones. 

## Brief change log

See commit messages

## Verifying this change

* testStreamQueryExecutionSink (test extended)

  - The serializers: (yes / b / don't know)

## Documentation


dawidwys:
Could you have a look at this PR @twalthr?

xuefuz:
Thanks to @dawidwys for the fix, which seems good to me. Here I'd like to reemphasize an observation for future consideration.

In our original design of CatalogManager, we wanted it to be only aware of multiple catalogs with one being the current. CatalogManager doesn't need to treat any one of them specially. Each catalog may has its own capability and restriction, but upper layers doesn't need to care about. This design would avoid such pervasive, special treatment of the built-in catalog as we see in this PR.

While it's too late to revisit this, I'd suggest we open up the discussion again post 1.9.

Thanks!

flink-table/flink-table-planner/src/main/scala/org/apache/flink/table/api/internal/TableEnvImpl.scala
```diff
@@ -60,9 +60,9 @@ abstract class TableEnvImpl(
     private val catalogManager: CatalogManager)
   extends TableEnvironment {
 
-  // The current catalog and database are definitely builtin.
-  protected val builtinCatalogName: String = catalogManager.getCurrentCatalog
-  protected val builtinDatabaseName: String = catalogManager.getCurrentDatabase
+  protected val builtInCatalogName: String = catalogManager.getBuiltInCatalogName
```
twalthr: nit: remove those variables and access catalog manager directly

flink-table/flink-sql-client/src/test/java/org/apache/flink/table/client/gateway/utils/SimpleCatalogFactory.java
```diff
@@ -0,0 +1,118 @@
...
+
+/**
+ * Catalog factory for an in-memory catalog that contains a single non-empty table.
+ * The contents of the table are equal to {@link SimpleCatalogFactory#TABLE_CONTENTS}.
+ */
+public class SimpleCatalogFactory implements CatalogFactory {
+
+	public static final String CATALOG_TYPE_VALUE = "simple-catalog";
+
+	public static final String TEST_TABLE_KEY = "test-table";
```
twalthr: nit: `TEST_TABLE_NAME`

flink-table/flink-sql-client/src/test/java/org/apache/flink/table/client/gateway/local/LocalExecutorITCase.java
```diff
@@ -151,6 +155,44 @@ public void testValidateSession() throws Exception {
 		assertEquals(expectedTables, actualTables);
 	}
 
+	@Test(timeout = 30_000L)
+	public void testDefaultCatalogDifferentFromBuiltin() throws Exception {
+		Environment defaultEnvironment = new Environment();
+		Map<String, Object> catalogs = new HashMap<>();
+		catalogs.put(CatalogEntry.CATALOG_NAME, "simple-catalog");
```
twalthr: add this to the yaml file instead?

flink-table/flink-sql-client/src/test/java/org/apache/flink/table/client/gateway/local/LocalExecutorITCase.java
```diff
@@ -151,6 +155,44 @@ public void testValidateSession() throws Exception {
 		assertEquals(expectedTables, actualTables);
 	}
 
+	@Test(timeout = 30_000L)
+	public void testDefaultCatalogDifferentFromBuiltin() throws Exception {
```
twalthr: Most IT cases here are testing runtime behavior, I'm not sure if we need to run a full IT case. But I'm fine if there is no better solution yet.

dawidwys: I agree. I also don't like the test. I couldn't find a well isolated way to test this behavior though.

PR #9230, opened 2019-07-25
[FLINK-13430][build] Configure sending travis build notifications to builds@flink.apache.org

wuchong:

Send travis build notifications the builds@flink.apache.org mailing list. 

## Brief change log

Integrate builds mailing list to the travis yaml according to the [travis doc](https://docs.travis-ci.com/user/notifications/#configuring-email-notifications).

## Verifying this change

This change is trivial without any test coverage.

## Documentation

### DIFF
--- a/flink-core/src/main/java/org/apache/flink/api/common/accumulators/Accumulator.java
+++ b/flink-core/src/main/java/org/apache/flink/api/common/accumulators/Accumulator.java
@@ -1,20 +1,4 @@
-/*
- * Licensed to the Apache Software Foundation (ASF) under one
- * or more contributor license agreements.  See the NOTICE file
- * distributed with this work for additional information
- * regarding copyright ownership.  The ASF licenses this file
- * to you under the Apache License, Version 2.0 (the
- * "License"); you may not use this file except in compliance
- * with the License.  You may obtain a copy of the License at
- *
- *     http://www.apache.org/licenses/LICENSE-2.0
- *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
- */
+
 
 package org.apache.flink.api.common.accumulators;
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -61,6 +61,11 @@ notifications:
       - secure: ikPQn5JTpkyzxVyOPm/jIl3FPm6hY8xAdG4pSwxGWjBqF+NmmNTp9YZsJ6fD8xPql6T5n1hNDbZSC14jVUw/vvXGvibDXLN+06f25ZQl+4LJBXaiR7gTG6y3nO8G90Vw7XpvCme6n5Md9tvjygb17a4FEgRJFfwzWnnyPA1yvK0=
     on_success: never
     on_pull_requests: false
+  webhooks:
+    urls:
+      secure: gOR0ayYgTHvx2QsFxxH07KzI8fgrDk2vqp/6q2eN/RuWHJa6kHecXz/9F0JOXNI1siXrSJ1FwCKOxiMmTbnImJP4015L9VGMS++us6AYtnFp6SIRorRqA1QS5tHNrpBIM38ANxQvuPhFj8S6216BsonQOqBu9sp0jQCOZGKwBLg=
+    on_success: change
+    on_pull_requests: false
 
 stages:
   - name: compile


wuchong:
I setup a CRON job on my testing branch to test the email setting: https://travis-ci.org/wuchong/flink/builds/563609665

zentol:
how can we prevent this from sending notifications for contributor repositories?

wuchong:
That's a good question @zentol . How do we handle this for slack channel?

beyond1920:
@zentol @wuchong Sending notifications for contributor repositories as well seems not be useless as it looks like. 
I could receives build notifications from my own Flink repository and public Flink repository by 
subscribe builds mail-list and custom my own email forwarding rules.
Otherwise, I may need custom travis config in my own repository.

zentol:
@wuchong the slack channel credentials are bound to the repository, and are simply invalid in all other repositories.

Having all contributor builds in builds@flink.apache.org just turns it into a massive spam fest, with the majority of builds not being relevant to the majority of subscribers.

zentol:
Could you try this instead and check whether there still are notifications for contributor forks?
```
  email:
    recipients:
        secure: Tnf61eG2XC4uH9VOS2FnmDZVKMQqoXxQh1tPBuj7wcXCxSBtRKZtwrLHeQYDaH+cdDvy4rgzInKxq0imUyVnZ3vl3eTJyCtUW33RanUN8g4zZANy1nj9NbyhE7U8MCkIOMEuIkqGrI1XfFU0k+OMKaeLlUWtNUkBxeZVueL0iP0=
    on_success: change
    on_pull_requests: false
```

wuchong:
Thanks @zentol , I will try that.

wuchong:
> Could you try this instead and check whether there still are notifications for contributor forks?
> (and check whether it still works for apache/flink by pushing this PR into a temporary branch)
> 
> ```
>   email:
>     recipients:
>         secure: Tnf61eG2XC4uH9VOS2FnmDZVKMQqoXxQh1tPBuj7wcXCxSBtRKZtwrLHeQYDaH+cdDvy4rgzInKxq0imUyVnZ3vl3eTJyCtUW33RanUN8g4zZANy1nj9NbyhE7U8MCkIOMEuIkqGrI1XfFU0k+OMKaeLlUWtNUkBxeZVueL0iP0=
>     on_success: change
>     on_pull_requests: false
> ```

Good news! The secure works. The fail notification from my branch wuchong/flink:builds is sent to my own mail address.  And the notification from apache/flink:builds-test is sent to builds@flink.apache.org.

zentol:
Well that sounds perfect then, +1.

wuchong:
> Well that sounds perfect then, +1.

However, we still need to figure out how to accept builds emails. Currently, all notification from builds@travis-ci.org are rejected.

zentol:
@wuchong see https://apache.org/dev/committers.html#mail-moderate

wuchong:
Yes @zentol . I have read it and tried to reply **-accept** and **-allow** address. But it still doesn't work.

wuchong:
I checked the builds@travis-ci.org has already been in the subscriber list and allow list.

zentol:
@wuchong Could you reach out to INFRA on what the problem might be? As far as I can tell we should've setup everything correctly.

wuchong:
Thanks for the advice @zentol , I have created an INFRA issue INFRA-18843.

wuchong:
Hi @zentol , the INFRA guys have figured out the problem and the builds@travis-ci.org works now. I removed the fast-fail commit. 

I think it's ready to be merged now. 

Cheers,
Jark

zentol:
+1 from my side.

wuchong:
I would like to hold this PR a bit moment @zentol . Because I find it will lose the notification to my own email address for the branches in my own repo. 

After reading the Travis documentation, I think we will also lose the build notification to the committer who pushed the commit, which I think is important.
image

zentol:
I had hoped this wouldn't apply to contributor builds since the mail was encrypted...

oh well, back to the drawing board instead. I guess we'll have to setup some kind of proxying over slack or similar for this to work then.

wuchong:
Hi @zentol , I have setup a webhook to receive travis' notifications and generate a email and send to mailing list. 

All the things looks good (send to my own email address), I will test on the mailing list then.

flink-core/src/main/java/org/apache/flink/api/common/accumulators/Accumulator.java
```diff
@@ -1,20 +1,4 @@
-/*
```
TisonKun: Accidentally removal I think. Revert to pass the checker :-)

wuchong: Thanks. I will revert this before merge. This is used to fail the travis fast.

TisonKun: Aha, I think we will back to the thread reduce building time :-)

zentol:
@wuchong awesome, looking forward to seeing this in action. Once the mail passes through moderation and retains the formatting when posted on builds@flink.apache.org, feel free to merge this.

Can you expand on how you setup the webhook? Just to ensure we can set this up elsewhere if anything happens to your setup.

wuchong:
@zentol , Sure, here is the source code: https://github.com/wuchong/flink-notification-bot
I deployed it in my VM.

wuchong:
Once it is done. I can contribute it to, e.g., https://github.com/flinkbot

wuchong:
Progress Update: 

The mailing list doesn't accept "text/html" context type. So I send a request to the INFRA to enable html the builds.  [NFRA-18918](https://issues.apache.org/jira/browse/INFRA-18918)

wuchong:
The HTML is enabled. And I tested several emails and looks good now. I will merge this then.

PR #9235, opened 2019-07-26
[FLINK-13228][tests][filesystems] Harden HadoopRecoverableWriterTest

carp84:
Currently test cases will fail when trying to close the output stream if all data written but `ClosedByInterruptException` occurs at the ending phase. This PR proposes to close the stream silently if the test case logic passed properly.

## Brief change log
Close the output stream silently in `HadoopRecoverableWriterTest` if the test case logic passed properly.

## Verifying this change
This change is on the test cases and verified by sanity test through modifying hadoop source code to simulate the rarely happening situation that triggers the issue.

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-core/src/test/java/org/apache/flink/core/fs/AbstractRecoverableWriterTest.java
+++ b/flink-core/src/test/java/org/apache/flink/core/fs/AbstractRecoverableWriterTest.java
@@ -111,7 +111,9 @@ public abstract class AbstractRecoverableWriterTest extends TestLogger {
 
 		final Path path = new Path(testDir, "part-0");
 
-		try (final RecoverableFsDataOutputStream stream = writer.open(path)) {
+		RecoverableFsDataOutputStream stream = null;
+		try {
+			stream = writer.open(path);
 			stream.write(testData1.getBytes(StandardCharsets.UTF_8));
 			stream.closeForCommit().commit();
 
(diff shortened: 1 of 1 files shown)


carp84:
This should be able to resolve FLINK-12071 also.

@flinkbot attention @tillrohrmann @StephanEwen

flink-core/src/test/java/org/apache/flink/core/fs/AbstractRecoverableWriterTest.java
```diff
@@ -306,13 +330,17 @@ public void testResumeAfterCommit() throws Exception {
 		final Path path = new Path(testDir, "part-0");
 
 		RecoverableWriter.ResumeRecoverable recoverable;
-		try (final RecoverableFsDataOutputStream stream = writer.open(path)) {
+		RecoverableFsDataOutputStream stream = null;
+		try{
```
zentol:
```suggestion
		try {
```

carp84:
Thanks for helping fix the checkstyle issue @zentol !

StephanEwen:
FYI: Y(o)u can use `IOUtils.closeQuietly(...)` for this as well.

carp84:
Thanks for the reminder @StephanEwen , I thought it might be useful to log the exception out at debug level but yes we could also manually change it when need to debug, will change to use `IOUtils.closeQuietly`

StephanEwen:
Looks good, thanks.
Merging this...

PR #9237, opened 2019-07-26
[FLINK-13431][hive] NameNode HA configuration was not loaded when running HiveConnector on Yarn

hongtao12310:
hdfs-site.xml was not loaded by HiveConf when running the Hive Connector on YARN Cluster

to fix the issue that hdfs-site.xml was not loaded by HiveConf when running Hive Jobs on YARN cluster

## Brief change log

  - create HiveConf from Hadoop configuration to make sure hdfs-site.xml was loaded

## Verifying this change

  - no new tests introduced

  - The serializers: (no)

## Documentation

  - If yes, how is the feature documented? (NA)


xuefuz:
Thanks for the contribution! The are a couple of high-level suggestions:

1. Please update the PR description detailing the changes introduced by this PR.
2. The PR seemingly breaks the build.
3. Consider adding some tests to cover the changes.

hongtao12310:
miss the description of the PR and failed the CI build. will close the pr and launch a new one

lirui-apache:
Thanks @hongtao12310 for working on this. Maybe we can just call `new HiveConf(HadoopUtils.getHadoopConfiguration, HiveConf.class)` to create the HiveConf? For testing, I think we can add some test case similar to `HadoopConfigLoadingTest`.
And btw, you can update the PR to add description, fix travis build, etc. No need to close and create a new one.

hongtao12310:
Thanks @lirui-apache for your comment

I will reopen the PR and add some description

==> Maybe we can just call new HiveConf(HadoopUtils.getHadoopConfiguration, HiveConf.class) to create the HiveConf? 
yes, I think it is reasonable

===> I think we can add some test case similar to HadoopConfigLoadingTest
do you mean we should check hiveConf properties load from Hadoop configuration? and these test cases should be going into HiveCatalogITCase.java ?

lirui-apache:
> do you mean we should check hiveConf properties load from Hadoop configuration? and these test cases should be going into HiveCatalogITCase.java ?

@hongtao12310 I just realized we don't have e2e test to cover this code path. Currently hive conf in tests is created by `HiveTestUtils`.
@xuefuz do you think we can postpone the test for now?

xuefuz:
For test, is it possible to add the following in HiveCatalogITCase.java? We can do sth like this:
1. Create a HiveCatalog.
2. Get the HiveConf object that backs HiveCatalog (Add package getHiveConf() in HiveCatalog if necessary.
3. Verify if HDFS configs are in HiveConf.

Otherwise, we can skip for now if it's too complicated. 

Thanks.

hongtao12310:
> For test, is it possible to add the following in HiveCatalogITCase.java? We can do sth like this:
> 
> 1. Create a HiveCatalog.
> 2. Get the HiveConf object that backs HiveCatalog (Add package getHiveConf() in HiveCatalog if necessary.
> 3. Verify if HDFS configs are in HiveConf.
> 
> Otherwise, we can skip for now if it's too complicated.
> 
> Thanks.

@xuefuz according to lirui's comment I have put the load HDFS configuration logic into HiveConf. so we don't need to create HiveCatalog even if we want to add the tests

lirui-apache:
@hongtao12310 I think we can have some test in `HiveCatalogFactoryTest`. You can place an hdfs-site.xml in our test resources dir and create a HiveCatalog with HiveCatalogFactory. Then verify properties in the hdfs-site.xml are loaded. As @xuefuz mentioned, we need to expose `hiveConf` in `HiveCatalog` for the test.

zjuwangg:
Thanks for your efforts @hongtao12310

hongtao12310:
@lirui-apache @xuefuz Now the changes I made was in the createHiveConf function. and the function is a static function. we can simply create the function use the HiveCatalog class. so i think the test should related to HiveCatalog tests not HiveCatalogFactory. but so far sounds like we didn't have any test cases for HiveCatalog

lirui-apache:
@hongtao12310 Creating hive conf is part of creating HiveCatalog instance. Currently we don't have a test class to test creating HiveCatalog instances. `HiveCatalogFactoryTest ` is the closet to what we need, and probably the only test that triggers `HiveCatalog.createHiveConf`.

hongtao12310:
cc @lirui-apache @xuefuz added test case, please take a look at

flink-connectors/flink-connector-hive/src/test/java/org/apache/flink/table/catalog/hive/factories/HiveCatalogFactoryTest.java
```diff
@@ -54,9 +65,49 @@ public void test() {
...
+		CommonTestUtils.setEnv(newEnv);
+
+		// create HiveCatalog use the Hadoop Configuration
+		final CatalogDescriptor catalogDescriptor = new HiveCatalogDescriptor();
+		final Map<String, String> properties = catalogDescriptor.toProperties();
+		final HiveCatalog hiveCatalog = (HiveCatalog) TableFactoryService.find(CatalogFactory.class, properties)
+			.createCatalog(catalogName, properties);
+		final HiveConf hiveConf = hiveCatalog.getHiveConf();
+		// set the Env back
+		CommonTestUtils.setEnv(originalEnv);
```
lirui-apache: I'd suggest move this to a finally block, so that if this test case fails, it doesn't pollute env for other tests.

hongtao12310: good catch !

hongtao12310: @lirui-apache update the test case. please take a look at it

PR #9238, opened 2019-07-26
[FLINK-13387][WebUI] Fix log download for old UI

zentol:
Corrects the URLs used by the WebUI when downloading the logs. When using the WebUI the URL contains a `old-version` prefix, which leaks into the relative path defined in the `.html` files. The JS files are unaffected as they ignore the prefix and have their own way of assembling URLs.

This PR simply modifies the `href` to go up one layer, which effectively removes the prefix from the call.


yanghua:
Hi @zentol When clicking the download button, the behavior is open a new web page and show the content of the whole log file. There is a simple way to avoid this behavior, we could add a `download` attribute for `<a></a>` tag which is [a standard of HTML](https://www.w3schools.com/tags/att_a_download.asp). It seems the new web UI also uses this attribute.

zentol:
I don't really want to change the behavior of the UI at this point.

PR #9242, opened 2019-07-26
[FLINK-13408][runtime] Schedule StandaloneResourceManager.setFailUnfulfillableRequest whenever the leadership is acquired

xintongsong:

We introduced `StandaloneResourceManager#setFailUnfulfillableRequest` to give some time to task executors to register the available slots before the slot requests can be checked whether they can be fulfilled or not. `setFailUnfulfillableRequest` is scheduled now only once when the RM is initialized but the task executors will register themselves every time this RM gets the leadership. Hence, `setFailUnfulfillableRequest` should be scheduled after each leader election.

## Brief change log

- 5ff534941925c49556462d188b9457334b4f1026: Override `StandaloneResourceManager#setFailUnfulfillableRequest` to schedule invoking `ResourceManager#setFailUnfulfillableRequest` respecting the start up period.
- e41db5e0470c5de4bf998c4fb4b6ca719fd6acb3: Resource managers set fail unfulfillable requests when leadership is granted, and reset it when leadership is revoked.
- bc4bc3a00e1692381368af36b87d233677e8c4ac: Add `StandaloneResourceManagerTest#testStartUpPeriodAfterLeadershipSwitch` validates that StandaloneResourceManager applies a startup period whenever the leadership is acquired.

## Verifying this change

- Add `StandaloneResourceManagerTest#testStartUpPeriodAfterLeadershipSwitch` validates that StandaloneResourceManager applies a startup period whenever the leadership is acquired.

  - The serializers: (no)

## Documentation


flink-mesos/src/main/java/org/apache/flink/mesos/runtime/clusterframework/MesosResourceManager.java
```diff
@@ -201,7 +201,6 @@ public MesosResourceManager(
 		final ContaineredTaskManagerParameters containeredTaskManagerParameters = taskManagerParameters.containeredParameters();
 		this.slotsPerWorker = updateTaskManagerConfigAndCreateWorkerSlotProfiles(
 			flinkConfig, containeredTaskManagerParameters.taskManagerTotalMemoryMB(), containeredTaskManagerParameters.numSlots());
-		setFailUnfulfillableRequest(true);
```
tillrohrmann: This is a good finding. Don't we have tests for Mesos which test this behavior? Given the previous code, unfulfillable requests should have failed right away, right?

flink-runtime/src/main/java/org/apache/flink/runtime/resourcemanager/StandaloneResourceManager.java
```diff
@@ -112,4 +103,21 @@ protected ResourceID workerStarted(ResourceID resourceID) {
...
+			if (startupPeriodMillis > 0) {
+				getRpcService().getScheduledExecutor().schedule(
+					() -> getMainThreadExecutor().execute(
+						() -> super.setFailUnfulfillableRequest(failUnfulfillableRequest)),
+					startupPeriodMillis,
+					TimeUnit.MILLISECONDS
+				);
+			}
+		} else {
+			super.setFailUnfulfillableRequest(failUnfulfillableRequest);
```
tillrohrmann: Let's replace `failUnfillableRequest` with `false` here.

flink-runtime/src/main/java/org/apache/flink/runtime/resourcemanager/StandaloneResourceManager.java
```diff
@@ -112,4 +103,21 @@ protected ResourceID workerStarted(ResourceID resourceID) {
 		return resourceID;
 	}
 
+	@Override
+	protected void setFailUnfulfillableRequest(boolean failUnfulfillableRequest) {
```
tillrohrmann: Let's not override the method but introduce another method which says that you start/stop the timeout. For the non standalone implementations the method should do nothing.

flink-runtime/src/main/java/org/apache/flink/runtime/resourcemanager/StandaloneResourceManager.java
```diff
@@ -112,4 +103,21 @@ protected ResourceID workerStarted(ResourceID resourceID) {
...
+		if (failUnfulfillableRequest) {
+			final long startupPeriodMillis = startupPeriodTime.toMilliseconds();
+
+			if (startupPeriodMillis > 0) {
+				getRpcService().getScheduledExecutor().schedule(
```
tillrohrmann: Let's not use this construct. I think it would be better to simply use `RpcEndpoint#scheduleRunAsync`.

flink-mesos/src/main/java/org/apache/flink/mesos/runtime/clusterframework/MesosResourceManager.java
xintongsong: Yes, unfulfillable requests should have failed right away for Yarn and Mesos. 

We have `SlotManagerFailUnfulfillableTest` that tests this behavior at the scope of slot manager, but no tests at the scope of resource manager for Yarn and Mesos. 

I've update the PR changing the default value of `SlotManager#failUnfulfillableRequest` to true and set it to false on start of `StandaloneResourceManager`. So the Yarn / Mesos resource manager do not need to do anything special regarding failing unfulfillable requests. I don't think we need more test cases other than `SlotManagerFailUnfulfillableTest` to test this behavior for Yarn / Mesos.

xintongsong:
@tillrohrmann 
The PR

xintongsong:
@tillrohrmann 
PR is updated and passed [travis](https://travis-ci.org/xintongsong/flink/builds/564576681).

flink-runtime/src/main/java/org/apache/flink/runtime/resourcemanager/StandaloneResourceManager.java
```diff
@@ -112,4 +104,22 @@ protected ResourceID workerStarted(ResourceID resourceID) {
 		return resourceID;
 	}
 
+	@Override
+	protected void startStartupPeriod() {
+		final long startupPeriodMillis = startupPeriodTime.toMilliseconds();
+
+		if (startupPeriodMillis > 0) {
+			scheduleRunAsync(
+				() -> getMainThreadExecutor().execute(
```
tillrohrmann: why are we using the main thread executor here?

flink-runtime/src/main/java/org/apache/flink/runtime/resourcemanager/ResourceManager.java
```diff
@@ -1070,6 +1074,14 @@ protected void setFailUnfulfillableRequest(boolean failUnfulfillableRequest) {
 		slotManager.setFailUnfulfillableRequest(failUnfulfillableRequest);
 	}
 
+	protected void startStartupPeriod() {
+		// noop
+	}
+
+	protected void resetStartupPeriod() {
```
tillrohrmann: I think we don't need this method. Instead we could let the `startStartupPeriod` always call `setFailUnfillableRequests(false)`.

flink-runtime/src/test/java/org/apache/flink/runtime/resourcemanager/StandaloneResourceManagerTest.java
```diff
@@ -83,6 +83,24 @@ public void testNoStartupPeriod() throws Exception {
...
+	public void testStartUpPeriodAfterLeadershipSwitch() throws Exception {
+		final TestingStandaloneResourceManager rm = createResourceManager(Time.milliseconds(1L));
+
+		final Deadline deadline1 = Deadline.fromNow(Duration.ofSeconds(10L));
+		assertHappensUntil(() -> rm.isFailingUnfulfillableRequestAsync().join(), deadline1);
+
+		rm.rmServices.revokeLeadership();
+
+		final Deadline deadline2 = Deadline.fromNow(Duration.ofSeconds(10L));
+		assertHappensUntil(() -> !rm.rmServices.slotManager.isFailingUnfulfillableRequest(), deadline2);
```
tillrohrmann: This pattern is problematic. We are accessing state of the `SlotManager` which is owned by the `ResourceManager` and which should only be accessed from the `RpcEndpoint's` main thread. Otherwise we might run into visibility issues. I would suggest to rather test the behaviour of the RM via its public API instead of inspecting its internal state.

flink-runtime/src/test/java/org/apache/flink/runtime/resourcemanager/StandaloneResourceManagerTest.java
```diff
@@ -83,6 +83,21 @@ public void testNoStartupPeriod() throws Exception {
...
+
+		final Deadline deadline1 = Deadline.fromNow(Duration.ofSeconds(10L));
+		assertHappensUntil(() -> rm.isFailingUnfulfillableRequestAsync().join(), deadline1);
+
+		rm.rmServices.revokeLeadership();
+
+		rm.rmServices.grantLeadership();
+
+		final Deadline deadline3 = Deadline.fromNow(Duration.ofSeconds(10L));
+		assertHappensUntil(() -> rm.isFailingUnfulfillableRequestAsync().join(), deadline3);
```
tillrohrmann: I think the test does not really ensure that we start a new startup period after granting the second leadership. It could also be the case that the timeout won't be reset. I think it would be better to pass in a TestingSlotManager on which we verify the behavior (e.g. that certain methods are being called)

xintongsong:
Thanks for the review, @tillrohrmann. I learned a lot from it.

PR #9243, opened 2019-07-26
[FLINK-13335][sql-parser] Align the SQL CREATE TABLE DDL with FLIP-37

danny0405:

Support parse all the data types defined in FLINK-37
- STRING
- BYTES
- TIME WITHOUT TIME ZONE
- TIME(3) WITHOUT TIME ZONE
- TIMESTAMP WITHOUT TIME ZONE
- TIMESTAMP(3) WITHOUT TIME ZONE
- TIMESTAMP WITH TIME ZONE (throws error)
- TIMESTAMP(3) WITH TIME ZONE (throws error)
- TIMESTAMP WITH LOCAL TIME ZONE
- TIMESTAMP(3) WITH LOCAL TIME ZONE
- INT NOT NULL ARRAY
- ROW(f0 INT NOT NULL, f1 BOOLEAN)
- ROW<f0 INT NOT NULL 'This is a comment.', f1 BOOLEAN 'This as well.'>
- ANY('unknown.class', '') (throws error)
- INT MULTISET
- MULTISET<INT NOT NULL>
- INT NOT NULL ARRAY
- INT NOT NULL MULTISET
- ROW()

## Brief change log

  - Add new data types in FLINK-37

## Verifying this change

See tests in FlinkDDLDataTypeTest.

  - The serializers: no

## Documentation

  - If yes, how is the feature documented? no


flink-table/flink-sql-parser/src/test/java/org/apache/flink/sql/parser/FlinkDDLDataTypeTest.java
```diff
@@ -0,0 +1,441 @@
...
+import java.util.Collections;
+import java.util.List;
+
+import static org.junit.Assert.assertEquals;
+
+/**
+ * Tests for all the sup[ported flink DDL data types.
+ */
+@RunWith(Parameterized.class)
+public class FlinkDDLDataTypeTest {
```
twalthr: Instead of prefixing it with Flink. You could just call this test `SqlDataTypeParserTest`?

flink-table/flink-sql-parser/src/test/java/org/apache/flink/sql/parser/FlinkDDLDataTypeTest.java
```diff
@@ -0,0 +1,441 @@
...
+		"  f0 %s\n" +
+		") with (\n" +
+		"  k1 = 'v1'\n" +
+		")";
+
+	@Parameterized.Parameters(name = "{index}: {0}")
+	public static List<TestItem> testData() {
+		return Arrays.asList(
+			createTestItem("CHAR", nullable(FIXTURE.char1Type), "CHAR"),
+			createTestItem("CHAR NOT NULL", FIXTURE.char1Type, "CHAR NOT NULL"),
```
twalthr: This test does not distinguish between nullable and non-nullable type. So it just swallows the information?

flink-table/flink-sql-parser/src/test/java/org/apache/flink/sql/parser/FlinkDDLDataTypeTest.java
```diff
@@ -0,0 +1,441 @@
...
+			createTestItem("CHAR   NOT \t\nNULL", FIXTURE.char1Type, "CHAR NOT NULL"),
+			createTestItem("char not null", FIXTURE.char1Type, "CHAR NOT NULL"),
+			createTestItem("CHAR NULL", nullable(FIXTURE.char1Type), "CHAR"),
+			createTestItem("CHAR(33)", nullable(FIXTURE.char33Type), "CHAR(33)"),
+			createTestItem("VARCHAR", nullable(FIXTURE.varcharType), "VARCHAR"),
+			createTestItem("VARCHAR(33)", nullable(FIXTURE.varchar33Type), "VARCHAR(33)"),
+			createTestItem("STRING",
+				nullable(FIXTURE.createSqlType(SqlTypeName.VARCHAR, Integer.MAX_VALUE)), "STRING"),
```
twalthr: Add this to `FIXTURE` as well?

flink-table/flink-sql-parser/src/test/java/org/apache/flink/sql/parser/Fixture.java
```diff
@@ -0,0 +1,115 @@
...
+import org.apache.calcite.rel.type.RelDataType;
+import org.apache.calcite.rel.type.RelDataTypeFactory;
+import org.apache.calcite.sql.type.SqlTypeName;
+
+import java.util.List;
+
+/**
+ * Type used during tests.
+ */
+public class Fixture {
```
twalthr: The name of this class is not very useful. How about `RelDataTypeTestData`?

flink-table/flink-sql-parser/src/test/java/org/apache/flink/sql/parser/FlinkDDLDataTypeTest.java
```diff
@@ -0,0 +1,441 @@
...
+				nullable(FIXTURE.timestamp3Type), "TIMESTAMP(3)"),
+			createTestItem("TIMESTAMP WITH LOCAL TIME ZONE",
+				nullable(FIXTURE.timestampWithLocalTimeZoneType),
+				"TIMESTAMP WITH LOCAL TIME ZONE"),
+			createTestItem("TIMESTAMP(3) WITH LOCAL TIME ZONE",
+				nullable(FIXTURE.timestamp3WithLocalTimeZoneType),
+				"TIMESTAMP(3) WITH LOCAL TIME ZONE"),
+			createTestItem("ARRAY<TIMESTAMP(3) WITH LOCAL TIME ZONE>",
+				nullable(FIXTURE.createArrayType(nullable(FIXTURE.timestamp3WithLocalTimeZoneType))),
+				"ARRAY< TIMESTAMP(3) WITH LOCAL TIME ZONE >"),
```
twalthr: Tests for intervals are missing.

flink-table/flink-sql-parser/src/test/java/org/apache/flink/sql/parser/FlinkDDLDataTypeTest.java
```diff
@@ -0,0 +1,441 @@
...
+			createTestItem("STRING",
+				nullable(FIXTURE.createSqlType(SqlTypeName.VARCHAR, Integer.MAX_VALUE)), "STRING"),
+			createTestItem("BOOLEAN", nullable(FIXTURE.booleanType), "BOOLEAN"),
+			createTestItem("BINARY", nullable(FIXTURE.binaryType), "BINARY"),
+			createTestItem("BINARY(33)", nullable(FIXTURE.binary33Type), "BINARY(33)"),
+			createTestItem("VARBINARY", nullable(FIXTURE.varbinaryType), "VARBINARY"),
+			createTestItem("VARBINARY(33)", nullable(FIXTURE.varbinary33Type),
+				"VARBINARY(33)"),
+			createTestItem("BYTES",
+				nullable(FIXTURE.createSqlType(SqlTypeName.VARBINARY, Integer.MAX_VALUE)),
```
twalthr: Add this to `FIXTURE` as well?

flink-table/flink-sql-parser/src/test/java/org/apache/flink/sql/parser/FlinkSqlParserImplTest.java
```diff
@@ -362,19 +362,19 @@ public void testCreateTableWithWatermarkStrategyFromSource() {
 	}
 
 	@Test
-	public void testCreateTableWithComplexType() {
+	public void testCreateTableWithDecimalType() {
```
twalthr: Shall we drop this test. Just testing decimal seems not useful given we have a `testCreateTableWithComplexType` already?

flink-table/flink-sql-parser/src/test/java/org/apache/flink/sql/parser/FlinkSqlParserImplTest.java
```diff
@@ -383,19 +383,21 @@ public void testCreateTableWithComplexType() {
...
 			"  x = 'y', \n" +
 			"  asd = 'data'\n" +
 			")\n", "CREATE TABLE `TBL1` (\n" +
-			"  `A`  DECIMAL,\n" +
-			"  `B`  DECIMAL(10, 0),\n" +
-			"  `C`  DECIMAL(38, 38),\n" +
+			"  `A`  ARRAY< BIGINT >,\n" +
+			"  `B`  MAP< INTEGER, VARCHAR >,\n" +
+			"  `C`  ROW< `CC0` INTEGER, `CC1` FLOAT, `CC2` VARCHAR >,\n" +
+			"  `D`  MULTISET< VARCHAR >,\n" +
```
twalthr: `STRING` instead of `VARCHAR`? We should encourage this type everywhere (e.g. update also the DDL doc in table environments), because `VARCHAR` is equal to `VARCHAR(1)` in SQL standard.

flink-table/flink-table-planner/src/test/scala/org/apache/flink/table/runtime/stream/TimeAttributesITCase.scala
```diff
@@ -512,7 +512,7 @@ class TimeAttributesITCase extends AbstractTestBase {
       .assignTimestampsAndWatermarks(new TimestampWithEqualWatermark())
     val table = stream.toTable(tEnv, 'rowtime.rowtime, 'int, 'double, 'float, 'bigdec, 'string)
     tEnv.registerTable("T1", table)
-    val querySql = "select rowtime as ts, string as msg from T1"
+    val querySql = "select rowtime as ts, `string` as msg from T1"
```
twalthr: Update the list of keywords in `dev/table/sql.html#reserved-keywords`.

twalthr:
One additional comment: I could not find test cases for the unsupported `ANY` type.

flink-table/flink-sql-parser/src/main/java/org/apache/flink/sql/parser/type/SqlRowType.java
```diff
@@ -51,6 +55,10 @@ public SqlRowType(SqlParserPos pos,
 		return fieldTypes;
 	}
 
+	public List<SqlCharStringLiteral> getComments() {
```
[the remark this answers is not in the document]
danny0405: Yes, because Calcite's record type does not support comments, but i have support it for SqlNode unparse.

flink-table/flink-sql-parser/src/main/java/org/apache/flink/sql/parser/type/SqlZonedTimeType.java
```diff
@@ -0,0 +1,73 @@
...
+import org.apache.calcite.sql.SqlIdentifier;
+import org.apache.calcite.sql.SqlWriter;
+import org.apache.calcite.sql.parser.SqlParserPos;
+import org.apache.calcite.sql.type.SqlTypeName;
+
+/**
+ * Parse type "TIME WITHOUT TIME ZONE", "TIME(3) WITHOUT TIME ZONE", "TIME WITH LOCAL TIME ZONE",
+ * or "TIME(3) WITH LOCAL TIME ZONE".
+ */
+public class SqlZonedTimeType extends SqlIdentifier implements ExtendedSqlType {
```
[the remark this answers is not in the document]
danny0405: Renamed to `SqlTimeType`.

flink-table/flink-sql-parser/src/test/java/org/apache/flink/sql/parser/Fixture.java
```diff
@@ -0,0 +1,115 @@
...
+	final RelDataType time3Type;
+	final RelDataType timestampType;
+	final RelDataType timestamp3Type;
+	final RelDataType timestampWithLocalTimeZoneType;
+	final RelDataType timestamp3WithLocalTimeZoneType;
+	final RelDataType nullType;
+
+	Fixture(RelDataTypeFactory typeFactory) {
+		this.typeFactory = typeFactory;
+		this.char1Type = typeFactory.createSqlType(SqlTypeName.CHAR);
```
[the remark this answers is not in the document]
danny0405: I would rather keep it.

flink-table/flink-sql-parser/src/test/java/org/apache/flink/sql/parser/Fixture.java
danny0405: Ignored

flink-table/flink-sql-parser/src/test/java/org/apache/flink/sql/parser/FlinkSqlParserImplTest.java
danny0405: Removed.

flink-table/flink-sql-parser/src/test/java/org/apache/flink/sql/parser/FlinkSqlParserImplTest.java
danny0405: Actually for Calcite its default precision is 65536 which most of the sql vendors implement, i think `VARCHAR` type should be the first place although we introduce an alias `STRING` type, it is more widely used.

flink-table/flink-sql-parser/src/test/java/org/apache/flink/sql/parser/FlinkDDLDataTypeTest.java
danny0405: There is one, because `INTERVAL` is not supported as DDL type, one test is enough.

flink-table/flink-sql-parser/src/test/java/org/apache/flink/sql/parser/FlinkDDLDataTypeTest.java
danny0405: I put it here intentionally, so we can see that this type has the max precision.

flink-table/flink-sql-parser/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -367,53 +375,299 @@ SqlDrop SqlDropView(Span s, boolean replace) :
...
+        LOOKAHEAD(2)
+        sqlTypeName = SqlTypeName(s) {
+            typeName = new SqlIdentifier(sqlTypeName.name(), s.end(this));
+        }
+    |
+        LOOKAHEAD(2)
+        typeName = FlinkCollectionsTypeName()
+    |
+        typeName = CompoundIdentifier() {
+            throw new ParseException("UDT in DDL is not supported yet.");
```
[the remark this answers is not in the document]
twalthr: But this means that the parser is still not compatible with FLIP-37. So it is still not possible to execute `CAST(12 AS STRING)`?

flink-table/flink-sql-parser/src/test/java/org/apache/flink/sql/parser/FlinkSqlParserImplTest.java
twalthr: According to the SQL standard section 6.1 clause (5):
"If <length> is omitted, then a <length> of 1 (one) is implicit." So we should encourage users to use `STRING` instead to prevent unwanted side effects.

twalthr: Also SQL Server has such behavior:
https://docs.microsoft.com/de-de/sql/t-sql/data-types/char-and-varchar-transact-sql?view=sql-server-2017
"When n isn't specified in a data definition or variable declaration statement, the default length is 1."

flink-table/flink-sql-parser/src/main/codegen/includes/parserImpls.ftl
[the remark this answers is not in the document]
danny0405: Yes. Let's postpone full type support for cast for 1.10, when i finished the data type parse refactoring in Calcite[1], which would be into release 1.21.

BTW, the full implicit type cast[2] support would be into release 1.21 too.

[1] https://issues.apache.org/jira/browse/CALCITE-3213
[2] https://issues.apache.org/jira/browse/CALCITE-2302

flink-table/flink-sql-parser/src/test/java/org/apache/flink/sql/parser/FlinkSqlParserImplTest.java
danny0405: But most of the users have used the `VARCHAR(65536)` for the long time, it would be a breaking change if we change it to default `VARCHAR(1)`, it is a SQL standard, but we do need follow everything.

twalthr: That's why I would suggest to introduce `STRING` as early as possible such that people start using it. We don't need to drop the old behavior immediately but can do it maybe for Flink 2.0. This should have been discussed in FLIP-37. Now it is documented and assumed at a couple of places.

flink-table/flink-sql-parser/src/main/codegen/includes/parserImpls.ftl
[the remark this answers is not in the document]
twalthr: Agreed, but please open an issue also in Flink to track the support. It is very confusing for users if data types are only available in certain SQL locations. Users don't understand the difference between DDL and DQL. Esp. with the introduction of computed columns it gets more confusing because special cases like `STRING` would be available in DDL but not in the `CAST` of the computed column.

flink-table/flink-sql-parser/src/test/java/org/apache/flink/sql/parser/Fixture.java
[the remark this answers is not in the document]
twalthr: Why? You also don't do it for `binaryType`?

flink-table/flink-sql-parser/src/test/java/org/apache/flink/sql/parser/FlinkSqlParserImplTest.java
wuchong: Jump in the discussion. Mysql forces to specify a length when using VARCHAR. So in order to avoid misuse `VARCHAR`, we can drop the support of `VARCHAR` (without length), and force users to specify a length. I think this doesn't break ANSI-SQL?

danny0405:
@twalthr Thanks, I would fire the issues soon ~

flink-table/flink-sql-parser/src/test/java/org/apache/flink/sql/parser/FlinkSqlParserImplTest.java
twalthr: Yes, that would also be an option. It is not really useful anyway and only exists to cover the corner cases of the standard.

flink-table/flink-sql-parser/src/main/java/org/apache/flink/sql/parser/type/SqlRowType.java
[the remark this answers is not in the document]
twalthr: @danny0405 We need to document all of that. I think we need to mark the DDL as beta for now. There are too many limitations. I hope we can it into a better shape in 1.10.

PR #9247, opened 2019-07-27
[FLINK-13386][web]: Fix frictions in the new default Web Frontend

vthinkxie:
Fix frictions in the new default Web Frontend

https://issues.apache.org/jira/browse/FLINK-13386

Fix below bugs

- [x] the jobmanager configuration is not sorted
- [x] different sorting of the operators (the old UI showed the sources first)
- [x] the drop-down list for choosing operator/tasks metrics is not sorted, which makes it super hard to screen through available metrics

There are also some views missing in the new UI that I personally found useful in the old UI:

- [x] can't see watermarks for all operators at once
- [x] no numeric metrics (only graphs)

## Documentation

### DIFF
--- a/flink-runtime-web/web-dashboard/package-lock.json
+++ b/flink-runtime-web/web-dashboard/package-lock.json
@@ -1572,6 +1572,7 @@
       "resolved": "https://registry.npmjs.org/are-we-there-yet/-/are-we-there-yet-1.1.5.tgz",
       "integrity": "sha512-5hYdAkZlcG8tOLujVDTgCT+uPX0VnpAH28gWsLfzpXYm7wP6mp5Q/gYyR7YQ0cKVJcXJnl3j2kpBan13PtQf6w==",
       "dev": true,
+      "optional": true,
       "requires": {
         "delegates": "^1.0.0",
         "readable-stream": "^2.0.6"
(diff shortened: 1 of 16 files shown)


dawidwys:
Hi @vthinkxie 
Thank you for the PR and really sorry it took me so long to check it.

All points beside `can't see watermarks for all operators at once` works as expected.
What I meant is I missed a view like this one in old frontend:
watermarks
This is not a pressing issue though. We might just see if users complain about this view.

I can't really check the PR contents as I am not very familiar with the frontend technology. Will try to ask a friend for that.

vthinkxie:
Hi @dawidwys 
I got your point, I will fix it later, thanks for your check!

vthinkxie:
@dawidwys 
I have added watermarks in the graph and the subtask watermark in the tab 


could you check this, and when there are too many vertexes, the web ui will request too many requests to the REST API, I have created another jira here https://issues.apache.org/jira/browse/FLINK-14147, I think we can merge this PR first, since it may need long time to resolve the jira.

sorahn:
Codewise this looks good to me.

@dawidwys could you verify the functionality then we can call this good?

dawidwys:
I verified the ui and it looks good... merging.

PR #9248, opened 2019-07-27
[FLINK-13446][table-runtime-blink] Fix assign logic for row count sliding window

hequn8128:

For blink planner, the Row count sliding window outputs incorrectly. The window assigner assigns less window than what expected. This means the window outputs fewer data. 

This pull request fixes the window assigner and trigger logic, i.e., correct the window start and end calculation. 

## Brief change log

  - Add trigger size property to `CountWindow`, because in some cases, the count window fires with a trigger size less than the window size.
  - Correct the window start and end calculation in `CountSlidingWindowAssigner`. 
  - Correct existing test cases.

## Verifying this change

This change is already covered by existing tests, such as GroupWindowITCase and WindowOperatorTest.

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/runtime/operators/window/assigners/CountSlidingWindowAssigner.java
+++ b/flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/runtime/operators/window/assigners/CountSlidingWindowAssigner.java
@@ -67,17 +67,19 @@ public class CountSlidingWindowAssigner extends WindowAssigner<CountWindow> {
 		Long countValue = count.value();
 		long currentCount = countValue == null ? 0L : countValue;
 		count.update(currentCount + 1);
-		long lastId = currentCount / windowSlide;
-		long lastStart = lastId * windowSlide;
-		long lastEnd = lastStart + windowSize - 1;
+		long windowId = currentCount / windowSlide;
+		long windowEnd = windowId * windowSlide + windowSlide - 1;
+		long windowStart = windowEnd - windowSize + 1;
 		List<CountWindow> windows = new ArrayList<>();
-		while (lastId >= 0 && lastStart <= currentCount && currentCount <= lastEnd) {
-			if (lastStart <= currentCount && currentCount <= lastEnd) {
-				windows.add(new CountWindow(lastId));
+		while ((windowStart <= currentCount) && (currentCount <= windowEnd)) {
+			if (0 > windowStart) {
+				windows.add(new CountWindow(windowId, windowSize + windowStart));
+			} else {
+				windows.add(new CountWindow(windowId, windowSize));
 			}
-			lastId--;
-			lastStart -= windowSlide;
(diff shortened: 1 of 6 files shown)


flink-table/flink-table-runtime-blink/src/test/java/org/apache/flink/table/runtime/operators/window/WindowOperatorTest.java
```diff
@@ -1045,7 +1045,7 @@ public void testSlidingCountWindow() throws Exception {
 
 		testHarness.processWatermark(new Watermark(12000));
 		testHarness.setProcessingTime(12000L);
-		expectedOutput.add(record("key2", 15L, 5L, 0L));
+		expectedOutput.add(record("key2", 6L, 3L, 0L));
```
wuchong: Do we miss "15L, 5L" here?

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/runtime/operators/window/assigners/CountSlidingWindowAssigner.java
```diff
@@ -67,17 +67,19 @@ public void open(InternalWindowProcessFunction.Context<?, CountWindow> ctx) thro
...
+		long windowStart = windowEnd - windowSize + 1;
 		List<CountWindow> windows = new ArrayList<>();
-		while (lastId >= 0 && lastStart <= currentCount && currentCount <= lastEnd) {
-			if (lastStart <= currentCount && currentCount <= lastEnd) {
-				windows.add(new CountWindow(lastId));
+		while ((windowStart <= currentCount) && (currentCount <= windowEnd)) {
+			if (0 > windowStart) {
+				windows.add(new CountWindow(windowId, windowSize + windowStart));
+			} else {
+				windows.add(new CountWindow(windowId, windowSize));
```
wuchong: What about to encode CountWindow as "countStart, countEnd" like TimeWindow. Then the logic of assigning windows will be similar to `SlidingWindowAssigner` which is much simpler IMO. What do you think?

flink-table/flink-table-runtime-blink/src/test/java/org/apache/flink/table/runtime/operators/window/WindowOperatorTest.java
hequn8128: I think we have not missed outputs here. 
For the slide window(size=5, slide=3), the first window is [-2,2], the second is [1, 5], so here, we only have one output for key2.

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/runtime/operators/window/assigners/CountSlidingWindowAssigner.java
hequn8128: Good idea.

PR #9249, opened 2019-07-27
[FLINK-13447][table-api] Change default planner to legacy planner instead of any one

wuchong:

We want to change the default behavior of the `EnvironmentSettings` to use old planner instead of any planner. This will enable us to have both planner in the classpath by default. This will also enable users/connectors to have both planner in dependency and without using EnvironmentSettings explicitly.

## Brief change log

Update `EnvironmentSettings` to use old planner instead of any planner by default.

## Verifying this change

## Documentation

### DIFF
--- a/flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/EnvironmentSettings.java
+++ b/flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/EnvironmentSettings.java
@@ -161,8 +161,7 @@ public class EnvironmentSettings {
 		private boolean isStreamingMode = true;
 
 		/**
-		 * Sets the old Flink planner as the required module. By default, {@link #useAnyPlanner()} is
-		 * enabled.
+		 * Sets the old Flink planner as the required module.
 		 *
 		 * <p>This is the default behavior.
 		 */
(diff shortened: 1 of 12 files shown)


flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/EnvironmentSettings.java
```diff
@@ -149,19 +149,26 @@ public boolean isStreamingMode() {
...
+		private static final String BLINK_EXECUTOR_FACTORY = "org.apache.flink.table.planner.delegation.BlinkExecutorFactory";
+
+		private String plannerClass = OLD_PLANNER_FACTORY;
+		private String executorClass = OLD_EXECUTOR_FACTORY;
 		private String builtInCatalogName = "default_catalog";
 		private String builtInDatabaseName = "default_database";
 		private boolean isStreamingMode = true;
 
 		/**
 		 * Sets the old Flink planner as the required module. By default, {@link #useAnyPlanner()} is
```
dawidwys: Please update the javadoc: `By default, {@link #useAnyPlanner()} is` -> `By default, {@link #useOldPlanner()} is`. Also in the remaining two methods.

wuchong: Good catch!

wuchong:
Travis passed: https://travis-ci.org/wuchong/flink/builds/564895873

Merging

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/EnvironmentSettings.java
twalthr: The Javadoc is still not entirely updated.

PR #9251, opened 2019-07-28
[FLINK-13451][tests] Remove use of Unsafe.defineClass() from CommonTestUtils

StephanEwen:

The method `Unsafe.defineClass()` is removed in Java 11. To support Java 11, we rework the method
`CommonTestUtils.createClassNotInClassPath()` to use a different mechanism.

This commit now writes the class byte code out to a temporary file and create a new URLClassLoader that loads the class from that file. 

The solution is not a complete drop-in replacement, because it cannot add the class to an existing class loader, but can only create a new pair of (classloader & new-class-in-that-classloader).
Because of that, the commit also adjusts the existing tests to work with that new mechanism.

## Brief change log

  - Change `CommonTestUtils`:  `createObjectForClassNotInClassPath()` ==> `createObjectFromNewClassLoader()`
  - Adjust existing tests

## Verifying this change

There are two unit tests that validate the method of the new test utility.
Aside from that, this change is a refactoring of existing tests and thus needs no additional tests.

  - The serializers: **no**

## Documentation


StephanEwen:
Thanks, merging this...

PR #9259, opened 2019-07-29
[MINOR][docs] Doc fix for cluster_setup.md

zjffdu:

This is a trivial PR to update the doc `cluster_setup.md`

## Brief change log

IIUC, `parallelism.default` is job level configuration instead of cluster configuration, so I remove it from `cluster_setup.md`


zentol:
If a config option applies to all jobs, then it is not a job-level option.

PR #9262, opened 2019-07-29
[FLINK-13442][network] Remove unnecessary notifySubpartitionConsumed method from view reader

zhijiangW:

*Currently the methods of `NetworkSequenceViewReader#notifySubpartitionConsumed` and `NetworkSequenceViewReader#releaseAllResources` would be called meanwhile in netty stack during releasing resources.*

*As confirmed in FLINK-13245, in order to make this release logic simple and clean, we could remove the redundant `notifySubpartitionConsumed` from `NetworkSequenceViewReader` side, and also remove it from `ResultSubpartitionView` side. In the implementation of `ResultSubpartitionView#releaseAllResources` it would further notify the parent subpartition of consumed state via `ResultSubpartition#notifySubpartitionConsumed` which further feedback to parent `ResultPartition` layer via `onConsumedSubpartition`. Finally `ResultPartition` could decide whether to release itself or not.*

*E.g. for the case of `ReleaseOnConsumptionResultPartition` which is mainly used for pipelined partition, it would release partition after reference counter decreased to
0. For the case of `ResultPartition` which would be generated for blocking partition by default, it would never be released after notifying consumed. And the JM/scheduler
would decide when to release partition properly. In addition, `InputChannel#notifySubpartitionConsumed` could also be removed as a result of above.*

## Brief change log

  - *Remove `notifySubpartitionConsumed` from `NetworkSequenceViewReader`*
  - *Remove `notifySubpartitionConsumed` from `ResultSubpartitionView`*
  - *Remove `notifySubpartitionConsumed` from `InputChannel`*

## Verifying this change

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NetworkSequenceViewReader.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/NetworkSequenceViewReader.java
@@ -61,8 +61,6 @@ public interface NetworkSequenceViewReader {
 	 */
 	void setRegisteredAsAvailable(boolean isRegisteredAvailable);
 
-	void notifySubpartitionConsumed() throws IOException;
-
 	boolean isReleased();
 
 	void releaseAllResources() throws IOException;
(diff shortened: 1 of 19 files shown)


zhijiangW:
During refactoring to remove the `notifySubpartitionConsumed` methods, I found another possible bug in previous implementation.  So I submitted these two things in separate commit.

@flinkbot attention @StephanEwen @pnowojski

zhijiangW:
I removed the last commit for fixing the problem of `notifySubpartitionConsumed` which does not consider multiple readers in blocking subpartition. That issue would be traced separately in FLINK-13493 and this PR is not necessary for current release-1.9 now.

zhijiangW:
Thanks for review @pnowojski ! I would handle it later.

PR #9264, opened 2019-07-29
[FLINK-13192][hive] Add tests for different Hive table formats

lirui-apache:

To add test for different table storage formats and fix issue with `HiveTableOutputFormat`.

## Brief change log

  - Make sure to use a common base class of the SerDe to support Hive 2.3.4 and 1.2.1.
  - Include some hadoop dependencies in test since hive runner needs to run MR job
  - Add tests for different table formats.

## Verifying this change

Added new test cases.

  - The serializers: no

## Documentation

### DIFF
--- a/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/connectors/hive/HiveTableOutputFormat.java
+++ b/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/connectors/hive/HiveTableOutputFormat.java
@@ -58,9 +58,10 @@ import org.apache.hadoop.hive.metastore.api.Table;
 import org.apache.hadoop.hive.ql.exec.FileSinkOperator;
 import org.apache.hadoop.hive.ql.io.HiveFileFormatUtils;
 import org.apache.hadoop.hive.ql.metadata.HiveException;
-import org.apache.hadoop.hive.serde2.AbstractSerDe;
+import org.apache.hadoop.hive.serde2.Deserializer;
 import org.apache.hadoop.hive.serde2.SerDeException;
 import org.apache.hadoop.hive.serde2.SerDeUtils;
+import org.apache.hadoop.hive.serde2.Serializer;
 import org.apache.hadoop.hive.serde2.objectinspector.ObjectInspector;
 import org.apache.hadoop.hive.serde2.objectinspector.ObjectInspectorFactory;
 import org.apache.hadoop.hive.serde2.objectinspector.StructObjectInspector;
@@ -122,7 +123,8 @@ public class HiveTableOutputFormat extends HadoopOutputFormatCommonBase<Row> imp
 	// number of non-partitioning columns
 	private transient int numNonPartitionColumns;
 
-	private transient AbstractSerDe serializer;
+	// SerDe in Hive-1.2.1 and Hive-2.3.4 can be of different classes, make sure to use a common base class
+	private transient Serializer serializer;
 	//StructObjectInspector represents the hive row structure.
 	private transient StructObjectInspector rowObjectInspector;
 	private transient Class<? extends Writable> outputClass;
@@ -257,10 +259,12 @@ public class HiveTableOutputFormat extends HadoopOutputFormatCommonBase<Row> imp
 	public void open(int taskNumber, int numTasks) throws IOException {
 		try {
 			StorageDescriptor sd = hiveTablePartition.getStorageDescriptor();
-			serializer = (AbstractSerDe) Class.forName(sd.getSerdeInfo().getSerializationLib()).newInstance();
+			serializer = (Serializer) Class.forName(sd.getSerdeInfo().getSerializationLib()).newInstance();
+			Preconditions.checkArgument(serializer instanceof Deserializer,
+					"Expect to get a SerDe, but actually got " + serializer.getClass().getName());
 			ReflectionUtils.setConf(serializer, jobConf);
 			// TODO: support partition properties, for now assume they're same as table properties
-			SerDeUtils.initializeSerDe(serializer, jobConf, tableProperties, null);
+			SerDeUtils.initializeSerDe((Deserializer) serializer, jobConf, tableProperties, null);
 			outputClass = serializer.getSerializedClass();
 		} catch (IllegalAccessException | SerDeException | InstantiationException | ClassNotFoundException e) {
 			throw new FlinkRuntimeException("Error initializing Hive serializer", e);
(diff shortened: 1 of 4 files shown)


lirui-apache:
cc @KurtYoung @xuefuz @zjuwangg

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/batch/connectors/hive/HiveTableOutputFormat.java
```diff
@@ -256,10 +258,12 @@ public void configure(Configuration parameters) {
 	public void open(int taskNumber, int numTasks) throws IOException {
 		try {
 			StorageDescriptor sd = hiveTablePartition.getStorageDescriptor();
-			serializer = (AbstractSerDe) Class.forName(sd.getSerdeInfo().getSerializationLib()).newInstance();
+			serializer = (Serializer) Class.forName(sd.getSerdeInfo().getSerializationLib()).newInstance();
+			Preconditions.checkArgument(serializer instanceof Deserializer,
```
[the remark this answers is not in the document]
lirui-apache: The problem is `SerDeUtils.initializeSerDe` requires a `Deserializer`. So we have to do the cast if we want to reuse this util method. Since most, if not all, SerDe lib implement both Serializer and Deserializer, I suppose this cast is OK?

flink-connectors/flink-connector-hive/src/test/java/org/apache/flink/batch/connectors/hive/TableEnvHiveConnectorTest.java
```diff
@@ -85,4 +87,40 @@ public void testDefaultPartitionName() throws Exception {
 
 		hiveShell.execute("drop database db1 cascade");
 	}
+
+	@Test
+	public void testDifferentFormats() throws Exception {
+		String[] formats = new String[]{"orc", "parquet", "sequencefile"};
```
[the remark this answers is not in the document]
lirui-apache: I'll add test for csv. Since all other test cases are using text tables, I don't think we need to cover it here.

lirui-apache:
Updated to add test for CSV table. Also found that Hive table schema can be get from either metastore or SerDe. For CSV tables, we should get schema from SerDe, but currently HiveCatalog doesn't support it. Hence the changes to HiveCatalog.
@xuefuz please have another look. Thanks.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/batch/connectors/hive/HiveTableOutputFormat.java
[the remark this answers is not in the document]
xuefuz: Casting is fine, but can we name the variable differently?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalog.java
```diff
@@ -494,8 +495,22 @@ private static CatalogBaseTable instantiateCatalogTable(Table hiveTable) {
 		String comment = properties.remove(HiveCatalogConfig.COMMENT);
 
 		// Table schema
+		List<FieldSchema> fields;
+		if (org.apache.hadoop.hive.ql.metadata.Table.hasMetastoreBasedSchema(hiveConf,
+				hiveTable.getSd().getSerdeInfo().getSerializationLib())) {
```
xuefuz: It seems that Table.getCols() is already doing what it's done here. Could you please check?

lirui-apache: I think you mean `org.apache.hadoop.hive.ql.metadata.Table.getCols()`. But we only have an instance of `org.apache.hadoop.hive.metastore.api.Table` here. Do you think we should create a `org.apache.hadoop.hive.ql.metadata.Table` and call `getCols()`?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/batch/connectors/hive/HiveTableOutputFormat.java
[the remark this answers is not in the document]
lirui-apache: Any suggestions about the name? Like `serDe`?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/HiveCatalog.java
lirui-apache: Unfortunately we cannot use `Table.getCols()` because this method gets HiveConf from SessionState, which we don't possess on Flink side.

lirui-apache:
Latest travis build succeeded. @xuefuz do you have any further comments?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/batch/connectors/hive/HiveTableOutputFormat.java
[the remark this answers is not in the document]
xuefuz: yeah! :)

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/connectors/hive/HiveTableOutputFormat.java
```diff
@@ -124,7 +124,7 @@
 	private transient int numNonPartitionColumns;
 
 	// SerDe in Hive-1.2.1 and Hive-2.3.4 can be of different classes, make sure to use a common base class
-	private transient Serializer serializer;
+	private transient Serializer recordSerDe;
```
xuefuz: Maybe the type here should be just "Object".

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/connectors/hive/HiveTableOutputFormat.java
```diff
@@ -259,13 +259,13 @@ public void configure(Configuration parameters) {
 	public void open(int taskNumber, int numTasks) throws IOException {
 		try {
 			StorageDescriptor sd = hiveTablePartition.getStorageDescriptor();
-			serializer = (Serializer) Class.forName(sd.getSerdeInfo().getSerializationLib()).newInstance();
-			Preconditions.checkArgument(serializer instanceof Deserializer,
-					"Expect to get a SerDe, but actually got " + serializer.getClass().getName());
-			ReflectionUtils.setConf(serializer, jobConf);
+			recordSerDe = (Serializer) Class.forName(sd.getSerdeInfo().getSerializationLib()).newInstance();
```
xuefuz: Can we assert the type to be "Serializer" before casting it, similar to below for "Deserializer"?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/connectors/hive/HiveTableOutputFormat.java
lirui-apache: It has to be a serializer because we need it to serialize records. Besides, using Object means we have to use reflection to call the `serialize` method. And if we do this for each record, it might hurt performance.

lirui-apache:
@xuefuz thanks for the review. PR updated to address your comments.

lirui-apache:
@KurtYoung could you help review and merge this PR? Thanks.

KurtYoung:
sure, I will take a look soon.

PR #9265, opened 2019-07-29
[FLINK-13477][core] Add `memory-overhead-ratio` conf for containers

BenoitHanotte:

Currently, the `-XX:MaxDirectMemorySize` parameter is set as:
`MaxDirectMemorySize = containerMemoryMB - heapSizeMB`

However as explained at
https://docs.oracle.com/javase/8/docs/technotes/tools/unix/java.html,
`MaxDirectMemorySize` only sets the maximum amount of memory that can be
used for direct buffers, thus the amount of off-heap memory used can be
greater than that value, leading to the sum of heap + off-heap memory used being
higher than the allocated container memory and the container being killed by Yarn
or Mesos for that reason.

In addition, users might want to allocate off-heap memory through native
code, in which case they will want to keep some of the container memory
free and unallocated by Flink.

This PR introduces the `containerized.memory-overhead-ratio` config to allow keeping some of the container memory free. With this change, the heap and direct memory sizes are computed as fractions of the usable memory, with `usableMemoryMB = containerMemoryMB * (1 - overheadRatio)` instead of on the full container memory. The overhead ratio default value is 0 to keep the current behavior.

## Brief change log

  - *Add `containerized.memory-overhead-ratio` config to allow keeping some of the container memory free (default is 0 to keep the current behavior)*

## Verifying this change

  - *Added `testUsableMemory` unit test to ensure memory-overhead-ratio is honored in `ContainerTaskManagerParametersTest.java`*

  - The serializers: no

## Documentation

### DIFF
--- a/docs/_includes/generated/resource_manager_configuration.html
+++ b/docs/_includes/generated/resource_manager_configuration.html
@@ -18,6 +18,11 @@
             <td>Percentage of heap space to remove from containers (YARN / Mesos), to compensate for other JVM memory usage.</td>
         </tr>
         <tr>
+            <td><h5>containerized.memory-overhead-ratio</h5></td>
+            <td style="word-wrap: break-word;">0.0</td>
+            <td>Percentage of the container memory (YARN / Mesos) to keep unallocated (not used for JVM heap memory or JVM off-heap direct memory).</td>
+        </tr>
+        <tr>
             <td><h5>local.number-resourcemanager</h5></td>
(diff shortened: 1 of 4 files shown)


tillrohrmann:
Thanks for creating this PR @BenoitHanotte. Sorry for the late response. In the meantime, the community has completed FLIP-49 and, hence, this change is no longer required. Closing this PR now.

PR #9268, opened 2019-07-30
[FLINK-13452] Ensure to fail global when exception happens during reseting tasks of regions

Myasuka:

After [FLINK-13060](https://issues.apache.org/jira/browse/FLINK-13060), we would run `createResetAndRescheduleTasksCallback` within another runnable `resetAndRescheduleTasks`. Unfortunately, any exception happened in `createResetAndRescheduleTasksCallback` would cause the thread terminated silently but record the exception in `outcome` of `FutureTask`. We should change the code back to previously that would `failGlobal` within the `createResetAndRescheduleTasksCallback` runnable.

## Brief change log

  - [59b1a6d5](https://github.com/apache/flink/pull/9268/commits/59b1a6d50b025925afd14a09b2b95a507889800a) :
    - Let runnable `createResetAndRescheduleTasksCallback` fail global if come across any exception.
    - Refine `RegionFailoverITCase` to mock the exception that checkpoint store would failed when recover from checkpoint for the 1st time.

  - [b732229](https://github.com/apache/flink/pull/9268/commits/b732229f0644a896b51245096fc0c7b7e19f0b02) :  Refactor interface of `RestartStrategy#restart` and add UT to verify `failGlobalIfErrorOnResetTasks`

 - [073d815](https://github.com/apache/flink/pull/9268/commits/073d8158a43ed8ac9ed44a3e563c5e8aca8c574a) 
    - Refactor the return value of RestartStrategy#restart to CompletableFuture<Void>
    - Refatcor return valud of method ExecutionGraph#allVerticesInTerminalState to align with new RestartStrategy#restart
    - Simplify AdaptedRestartPipelinedRegionStrategyNGFailoverTest with FailureMultiTimesRestartStrategy to throw exception directly
    - Refactor RegionFailoverITCase with FailureMultiTimesRestartStrategy
    - Add unit tests for FutureUtils#scheduleAsync

## Verifying this change

  - Refine `RegionFailoverITCase` to mock the exception that checkpoint store would failed when recover from checkpoint for the 1st time.

  - The serializers: **no**

## Documentation

### DIFF
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/AdaptedRestartPipelinedRegionStrategyNG.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/AdaptedRestartPipelinedRegionStrategyNG.java
@@ -54,6 +54,7 @@ import java.util.concurrent.CancellationException;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionException;
 import java.util.function.BiFunction;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 
 import static org.apache.flink.util.Preconditions.checkNotNull;
@@ -108,16 +109,18 @@ public class AdaptedRestartPipelinedRegionStrategyNG extends FailoverStrategy {
 
 		FutureUtils.assertNoException(
 			cancelTasks(verticesToRestart)
-				.thenRunAsync(resetAndRescheduleTasks(globalModVersion, vertexVersions), executionGraph.getJobMasterMainThreadExecutor())
+				.thenComposeAsync(resetAndRescheduleTasks(globalModVersion, vertexVersions), executionGraph.getJobMasterMainThreadExecutor())
 				.handle(failGlobalOnError()));
 	}
 
-	private Runnable resetAndRescheduleTasks(final long globalModVersion, final Set<ExecutionVertexVersion> vertexVersions) {
-		final RestartStrategy restartStrategy = executionGraph.getRestartStrategy();
-		return () -> restartStrategy.restart(
-			createResetAndRescheduleTasksCallback(globalModVersion, vertexVersions),
-			executionGraph.getJobMasterMainThreadExecutor()
-		);
+	private Function<Object, CompletableFuture<Void>> resetAndRescheduleTasks(final long globalModVersion, final Set<ExecutionVertexVersion> vertexVersions) {
+		return (ignored) -> {
+			final RestartStrategy restartStrategy = executionGraph.getRestartStrategy();
+			return restartStrategy.restart(
+				createResetAndRescheduleTasksCallback(globalModVersion, vertexVersions),
+				executionGraph.getJobMasterMainThreadExecutor()
+			);
+		};
 	}
 
 	private RestartCallback createResetAndRescheduleTasksCallback(final long globalModVersion, final Set<ExecutionVertexVersion> vertexVersions) {
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/restart/FixedDelayRestartStrategy.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/restart/FixedDelayRestartStrategy.java
@@ -18,13 +18,15 @@
 
 package org.apache.flink.runtime.executiongraph.restart;
 
+import org.apache.flink.api.common.time.Time;
 import org.apache.flink.configuration.ConfigConstants;
 import org.apache.flink.configuration.Configuration;
+import org.apache.flink.runtime.concurrent.FutureUtils;
 import org.apache.flink.runtime.concurrent.ScheduledExecutor;
 import org.apache.flink.runtime.executiongraph.ExecutionGraph;
 import org.apache.flink.util.Preconditions;
 
-import java.util.concurrent.TimeUnit;
+import java.util.concurrent.CompletableFuture;
 
 import scala.concurrent.duration.Duration;
 
@@ -60,9 +62,9 @@ public class FixedDelayRestartStrategy implements RestartStrategy {
 	}
 
 	@Override
-	public void restart(final RestartCallback restarter, ScheduledExecutor executor) {
+	public CompletableFuture<Void> restart(final RestartCallback restarter, ScheduledExecutor executor) {
 		currentRestartAttempt++;
-		executor.schedule(restarter::triggerFullRecovery, delayBetweenRestartAttempts, TimeUnit.MILLISECONDS);
+		return FutureUtils.scheduleWithDelay(restarter::triggerFullRecovery, Time.milliseconds(delayBetweenRestartAttempts), executor);
 	}
 
 	/**
--- a/flink-runtime/src/test/java/org/apache/flink/runtime/concurrent/FutureUtilsTest.java
+++ b/flink-runtime/src/test/java/org/apache/flink/runtime/concurrent/FutureUtilsTest.java
@@ -236,6 +236,65 @@ public class FutureUtilsTest extends TestLogger {
 	}
 
 	/**
+	 * Tests that the operation could be scheduled with expected delay.
+	 */
+	@Test
+	public void testScheduleWithDelay() throws Exception {
+		final ManuallyTriggeredScheduledExecutor scheduledExecutor = new ManuallyTriggeredScheduledExecutor();
+
+		final int expectedResult = 42;
+		CompletableFuture<Integer> completableFuture = FutureUtils.scheduleWithDelay(
+			() -> expectedResult,
+			Time.milliseconds(0),
+			scheduledExecutor);
+
+		scheduledExecutor.triggerScheduledTasks();
+		final int actualResult = completableFuture.get();
+
+		assertEquals(expectedResult, actualResult);
+	}
+
+	/**
+	 * Tests that a scheduled task is canceled if the scheduled future is being cancelled.
+	 */
+	@Test
+	public void testScheduleWithDelayCancellation() {
+		final ManuallyTriggeredScheduledExecutor scheduledExecutor = new ManuallyTriggeredScheduledExecutor();
+
+		final Runnable noOpRunnable = () -> {};
+		CompletableFuture<Void> completableFuture = FutureUtils.scheduleWithDelay(
+			noOpRunnable,
+			TestingUtils.infiniteTime(),
+			scheduledExecutor);
+
+		final ScheduledFuture<?> scheduledFuture = scheduledExecutor
+			.getScheduledTasks()
+			.iterator()
+			.next();
+
+		completableFuture.cancel(false);
+
+		assertTrue(completableFuture.isCancelled());
+		assertTrue(scheduledFuture.isCancelled());
+	}
+
+	/**
+	 * Tests that the operation is never scheduled if the delay is virtually infinite.
+	 */
+	@Test
+	public void testScheduleWithInfiniteDelayNeverSchedulesOperation() {
+		final Runnable noOpRunnable = () -> {};
+		final CompletableFuture<Void> completableFuture = FutureUtils.scheduleWithDelay(
+			noOpRunnable,
+			TestingUtils.infiniteTime(),
+			TestingUtils.defaultScheduledExecutor());
+
+		assertFalse(completableFuture.isDone());
+
+		completableFuture.cancel(false);
+	}
+
+	/**
 	 * Tests that a future is timed out after the specified timeout.
 	 */
 	@Test
--- a/flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/AdaptedRestartPipelinedRegionStrategyNGFailoverTest.java
+++ b/flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/AdaptedRestartPipelinedRegionStrategyNGFailoverTest.java
@@ -19,20 +19,17 @@
 package org.apache.flink.runtime.executiongraph;
 
 import org.apache.flink.api.common.JobID;
-import org.apache.flink.api.common.time.Time;
-import org.apache.flink.runtime.clusterframework.types.SlotProfile;
 import org.apache.flink.runtime.concurrent.ComponentMainThreadExecutor;
 import org.apache.flink.runtime.concurrent.ComponentMainThreadExecutorServiceAdapter;
 import org.apache.flink.runtime.concurrent.FutureUtils;
 import org.apache.flink.runtime.concurrent.ManuallyTriggeredScheduledExecutor;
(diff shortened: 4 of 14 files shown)


Myasuka:
[CI failed](https://travis-ci.com/flink-ci/flink/builds/121189644) due to [FLINK-13488](https://issues.apache.org/jira/browse/FLINK-13488) and travis file cache not found.

Myasuka:
This fix still has problem when `failGlobal` meet exception, close this PR first.

Myasuka:
Reopen this wrt the comments from @zhuzhurk in FLINK-13452.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/AdaptedRestartPipelinedRegionStrategyNG.java
```diff
@@ -108,8 +106,7 @@ protected void restartTasks(final Set<ExecutionVertexID> verticesToRestart) {
 
 		FutureUtils.assertNoException(
 			cancelTasks(verticesToRestart)
-				.thenRunAsync(resetAndRescheduleTasks(globalModVersion, vertexVersions), executionGraph.getJobMasterMainThreadExecutor())
-				.handle(failGlobalOnError()));
+				.thenRunAsync(resetAndRescheduleTasks(globalModVersion, vertexVersions), executionGraph.getJobMasterMainThreadExecutor()));
```
zhuzhurk: I think failure handling for the first part is still needed.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/AdaptedRestartPipelinedRegionStrategyNG.java
```diff
@@ -148,19 +145,10 @@ private RestartCallback createResetAndRescheduleTasksCallback(final long globalM
 			} catch (GlobalModVersionMismatch e) {
 				throw new IllegalStateException(
```
zhuzhurk: All exceptions should be handled.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/AdaptedRestartPipelinedRegionStrategyNG.java
```diff
@@ -148,19 +145,10 @@ private RestartCallback createResetAndRescheduleTasksCallback(final long globalM
...
 					"Bug: ExecutionGraph was concurrently modified outside of main thread", e);
-			} catch (Exception e) {
-				throw new CompletionException(e);
-			}
-		};
-	}
-
-	private BiFunction<Object, Throwable, Object> failGlobalOnError() {
```
zhuzhurk: This method is still needed for failure handling of the first part of failover process.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/AdaptedRestartPipelinedRegionStrategyNG.java
```diff
@@ -148,19 +145,10 @@ private RestartCallback createResetAndRescheduleTasksCallback(final long globalM
...
-		return (Object ignored, Throwable t) -> {
-			if (t != null) {
+			} catch (Throwable t) {
 				LOG.info("Unexpected error happens in region failover. Fail globally.", t);
 				failGlobal(t);
```
zhuzhurk: We can use `FatalExitExceptionHandler` to handle  exceptions thrown from `failGlobal` directly.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/AdaptedRestartPipelinedRegionStrategyNG.java
```diff
@@ -108,8 +106,7 @@ protected void restartTasks(final Set<ExecutionVertexID> verticesToRestart) {
 
 		FutureUtils.assertNoException(
 			cancelTasks(verticesToRestart)
-				.thenRunAsync(resetAndRescheduleTasks(globalModVersion, vertexVersions), executionGraph.getJobMasterMainThreadExecutor())
-				.handle(failGlobalOnError()));
```
tillrohrmann: I think we should not change this logic since `cancelTasks` can return an exceptionally completed future which would now cause the process to terminate.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/AdaptedRestartPipelinedRegionStrategyNG.java
tillrohrmann: I think it is better to let the exception handling happen where it has been. Usually it is a good idea to do the exception handling at the highest possible level.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/failover/AdaptedRestartPipelinedRegionStrategyNG.java
Myasuka: Agreed, I am refactoring this PR by changing the result of `RestartStrategy#restart` to a `CompletableFuture<?>`

flink-tests/src/test/java/org/apache/flink/test/checkpointing/RegionFailoverITCase.java
```diff
@@ -282,7 +301,11 @@ public void initializeState(FunctionInitializationContext context) throws Except
 
 				unionListState = context.getOperatorStateStore().getUnionListState(unionStateDescriptor);
 				Set<Integer> actualIndices = StreamSupport.stream(unionListState.get().spliterator(), false).collect(Collectors.toSet());
-				Assert.assertTrue(CollectionUtils.isEqualCollection(EXPECTED_INDICES, actualIndices));
+				if (getRuntimeContext().getTaskName().contains(SINGLE_REGION_SOURCE_NAME)) {
```
GJL: Is this change really necessary to resolve FLINK-13452, or can it be in a separate ticket?

flink-tests/src/test/java/org/apache/flink/test/checkpointing/RegionFailoverITCase.java
```diff
@@ -97,8 +111,10 @@
 
 	@Before
 	public void setup() throws Exception {
+		HighAvailabilityServicesUtilsTest.TestHAFactory.haServices = new FailHaServices(new TestingCheckpointRecoveryFactory(new FailRecoverCompletedCheckpointStore(1, 1), new StandaloneCheckpointIDCounter()), TestingUtils.defaultExecutor());
 		Configuration configuration = new Configuration();
 		configuration.setString(JobManagerOptions.EXECUTION_FAILOVER_STRATEGY, "region");
+		configuration.setString(HighAvailabilityOptions.HA_MODE, HighAvailabilityServicesUtilsTest.TestHAFactory.class.getName());
```
GJL: Interesting feature, didn't know about it before. However, I have two questions regarding this integration test:
1.  As before, isn't it enough to use a [RestartStrategy](https://github.com/apache/flink/blob/1dd64058985afd44cae6a5119c4843b2dd85873d/flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/restart/RestartStrategyFactory.java#L136) that returns an exceptionally completed future?
1. We already test that `failGlobal` is invoked from the unit tests. Does the integration test add coverage?

GJL:
Thank you for your contribution to Apache Flink @Myasuka. I think the PR is going in the right direction.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/AdaptedRestartPipelinedRegionStrategyNGFailoverTest.java
```diff
@@ -376,12 +410,13 @@ private JobGraph createBatchJobGraph() {
...
 	private ExecutionGraph createExecutionGraph(final JobGraph jobGraph) throws Exception {
-		return createExecutionGraph(jobGraph, new FixedDelayRestartStrategy(10, 0));
+		return createExecutionGraph(jobGraph, new FixedDelayRestartStrategy(10, 0), false);
 	}
 
 	private ExecutionGraph createExecutionGraph(
 			final JobGraph jobGraph,
-			final RestartStrategy restartStrategy) throws Exception {
+			final RestartStrategy restartStrategy,
+			final boolean failOnRecoverCheckpoint) throws Exception {
```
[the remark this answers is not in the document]
Myasuka: Agreed, combine `testFailGlobalIfErrorOnResetTasks` and previous `testFailGlobalIfErrorOnRescheduleTasks` into one unit test as `testFailGlobalIfErrorOnRestartTasks` with given RestartStrategy which would fail when restarting.

flink-tests/src/test/java/org/apache/flink/test/checkpointing/RegionFailoverITCase.java
Myasuka: Previously, the `RegionFailoverITCase` did not catch bug of [FLINK-13452](https://issues.apache.org/jira/browse/FLINK-13452) due to it did not involve global failover but only region failover. I prefer to add global failover in this integration test to verify the job could be restarted and result still correct.

flink-tests/src/test/java/org/apache/flink/test/checkpointing/RegionFailoverITCase.java
Myasuka: If we want to let this integration test to include global failover scenario, this is necessary.

Myasuka:
@GJL new commit content is updated in description.

There is another question which might be out of the scope of this PR but confused me. If I set the failure times of `FailureMultiTimesRestartStrategy` as 2, which means the restartStrategy would failed twice when calling restart. During `AdaptedRestartPipelinedRegionStrategyNG#restartTasks`, job would first meet exception when `restartStrategy.restart` and then call `failGlobalOnError`. However, in the call stack of `ExecutionGraph#failGlobal` --> `ExecutionGraph#allVerticesInTerminalState` --> `ExecutionGraph#tryRestartOrFail` --> `restartStrategy.restart`, we would meet the 2nd exception and caught by `FatalExitExceptionHandler`, resulting in the process exited. 

In other words, if we meet unchecked exception during `ExecutionGraph#failGlobal`, we would just let the whole process exit instead of trying to fail global again. Is this behavior expected?

GJL:
Thanks for the update. I'll have a look.

flink-runtime/src/test/java/org/apache/flink/runtime/concurrent/FutureUtilsTest.java
```diff
@@ -235,6 +237,57 @@ public void testRetryWithDelayCancellation() {
...
+		final long start = System.currentTimeMillis();
+		final AtomicInteger updateValue = new AtomicInteger(0);
+
+		CompletableFuture<Integer> scheduleAsyncFuture = FutureUtils.scheduleAsync(
+			() -> updateValue.addAndGet(ThreadLocalRandom.current().nextInt(100)),
+			delay,
+			TestingUtils.defaultScheduledExecutor());
+
+		int result = scheduleAsyncFuture.get();
+		long completionTime = System.currentTimeMillis() - start;
```
GJL: There some issues that I see here:
1. `System.currentTimeMillis()` is [impacted by system clock changes](https://stackoverflow.com/a/15620527) – this may cuase test failures in conjunction with NTP. Therefore `System.nanoTime()` should be preferred.  
1. Test will run at least 500ms which is on the slow side for unit tests. I think it's enough to use `ManuallyTriggeredScheduledExecutor` and see if the correct result is produced.
1. I don't think using `AtomicInteger` adds test coverage. Imo it's enough to assert on the value of `result`.

flink-runtime/src/test/java/org/apache/flink/runtime/concurrent/FutureUtilsTest.java
```diff
@@ -235,6 +237,57 @@ public void testRetryWithDelayCancellation() {
...
+
+		assertFalse(scheduleAsyncFuture.isDone());
+
+		final Collection<ScheduledFuture<?>> scheduledTasks = scheduledExecutor.getScheduledTasks();
+
+		assertFalse(scheduledTasks.isEmpty());
+
+		final ScheduledFuture<?> scheduledFuture = scheduledTasks.iterator().next();
+
+		assertFalse(scheduledFuture.isDone());
```
GJL: I don't think this assertion is needed because you already test if the future can be cancelled:

```
assertTrue(scheduledFuture.isCancelled());
```

This assertion will fail if the future is finished normally.

flink-runtime/src/test/java/org/apache/flink/runtime/concurrent/FutureUtilsTest.java
```diff
@@ -235,6 +237,57 @@ public void testRetryWithDelayCancellation() {
...
+		final ManuallyTriggeredScheduledExecutor scheduledExecutor = new ManuallyTriggeredScheduledExecutor();
+
+		CompletableFuture<?> scheduleAsyncFuture = FutureUtils.scheduleAsync(
+			() -> {
+				throw new FlinkRuntimeException("Test Exception");
+			},
+			TestingUtils.infiniteTime(),
+			scheduledExecutor);
+
+		assertFalse(scheduleAsyncFuture.isDone());
```
GJL: I think we can move this to the bottom so that all assertions are in one place.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/restart/FixedDelayRestartStrategy.java
```diff
@@ -60,9 +63,13 @@ public boolean canRestart() {
 	}
 
 	@Override
-	public void restart(final RestartCallback restarter, ScheduledExecutor executor) {
+	public CompletableFuture<Void> restart(final RestartCallback restarter, ScheduledExecutor executor) {
 		currentRestartAttempt++;
-		executor.schedule(restarter::triggerFullRecovery, delayBetweenRestartAttempts, TimeUnit.MILLISECONDS);
+		Supplier<Void> restartSupplier = () -> {
```
GJL: This conversion to `Supplier` is duplicated multiple times. See my other comment in `FutureUtils`.

Myasuka:
Sorry for late reply.
@GJL , No problem, please go ahead to take over this PR to finish the rest part.

GJL:
Please review @tillrohrmann

GJL:
Merging as soon as build is green.

GJL:
Merged manually.

PR #9269, opened 2019-07-30
[FLINK-9900][tests] Fix unstable ZooKeeperHighAvailabilityITCase

ifndef-SleePy:

* Fix unstable `ZooKeeperHighAvailabilityITCase`.`testRestoreBehaviourWithFaultyStateHandles`

* The case is designed as below
  - This case assume that the first 5 checkpoints (1-5) would success
  - Then the job blocks on the snapshot of checkpoint 6
  - At this time, the checkpoint files are moved on purpose
  - The checkpoint 6 would fail due to an expected snapshot failure
  - Then the job would be fail due to this failure checkpoint
  - And the job could not recover from checkpoint 5 because there is no checkpoint file
  - After moving these checkpoint files back, the job could recover and continue working.

* But there is a race condition of failing the job and triggering another checkpoint
* There might be an unexpected successful checkpoint 7 if the job canceling is not fast enough
* This job could recover from checkpoint 7 without waiting these checkpoint files moved back

## Brief change log

* The basic idea of fixing is that preventing the unexpected checkpoint 7
* Add a latch to block snapshot until the HA storage is recovered

## Verifying this change

* This unstable scenario can be reproduced as below
  - There is a race condition of failing the job and triggering another checkpoint
  - Making the job failing more slowly would reproduce the scenario
  - Modify the `FailJobCallback` of `CheckpointFailureManager` in `ExecutionGraph`.`enableCheckpointing`, change the `execute` to `schedule` with a delay
  - There would be an unexpected successful checkpoint 7
  - This case would hang forever because it never fail 5 times because it could recover from checkpoint 7

  - The serializers: no

## Documentation

### DIFF
--- a/flink-tests/src/test/java/org/apache/flink/test/checkpointing/ZooKeeperHighAvailabilityITCase.java
+++ b/flink-tests/src/test/java/org/apache/flink/test/checkpointing/ZooKeeperHighAvailabilityITCase.java
@@ -106,8 +106,9 @@ public class ZooKeeperHighAvailabilityITCase extends TestLogger {
 
 	private static MiniClusterWithClientResource miniClusterResource;
 
-	private static OneShotLatch waitForCheckpointLatch = new OneShotLatch();
-	private static OneShotLatch failInCheckpointLatch = new OneShotLatch();
+	private static OneShotLatch waitForCheckpointLatch;
+	private static OneShotLatch failInCheckpointLatch;
+	private static OneShotLatch blockSnapshotLatch;
 
 	@BeforeClass
 	public static void setup() throws Exception {
@@ -179,6 +180,7 @@ public class ZooKeeperHighAvailabilityITCase extends TestLogger {
 
 		waitForCheckpointLatch = new OneShotLatch();
 		failInCheckpointLatch = new OneShotLatch();
+		blockSnapshotLatch = new OneShotLatch();
 
 		ClusterClient<?> clusterClient = miniClusterResource.getClusterClient();
 		final Deadline deadline = Deadline.now().plus(TEST_TIMEOUT);
@@ -255,6 +257,7 @@ public class ZooKeeperHighAvailabilityITCase extends TestLogger {
 				return FileVisitResult.CONTINUE;
 			}
 		});
+		blockSnapshotLatch.trigger();
 
 		// now the job should be able to go to RUNNING again and then eventually to FINISHED,
 		// which it only does if it could successfully restore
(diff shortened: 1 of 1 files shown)


flink-tests/src/test/java/org/apache/flink/test/checkpointing/ZooKeeperHighAvailabilityITCase.java
```diff
@@ -366,6 +369,11 @@ public void snapshotState(FunctionSnapshotContext context) throws Exception {
 				failInCheckpointLatch.await();
 				if (!failedAlready.getAndSet(true)) {
 					throw new RuntimeException("Failing on purpose.");
+				} else {
```
tillrohrmann: Instead of this else branch, couldn't we always fail unless `afterMessWithZooKeeper` is `true`?

tillrohrmann:
Thanks for opening this fix @ifndef-SleePy. I actually think that you uncovered a true inconsistency which has been introduced with FLINK-12364. I will create another issue tracking the underlying problem.

flink-tests/src/test/java/org/apache/flink/test/checkpointing/ZooKeeperHighAvailabilityITCase.java
```diff
@@ -366,6 +369,11 @@ public void snapshotState(FunctionSnapshotContext context) throws Exception {
...
+					// make sure there would be no more successful checkpoint before job failing
+					// otherwise there might be a successful checkpoint 7 which is unexpected
+					// we expect checkpoint 5 is the last successful one before ha storage recovering
+					blockSnapshotLatch.await();
```
tillrohrmann: This might block the `Task` thread indefinitely until it is interrupted.

ifndef-SleePy:
Hi @tillrohrmann , thanks for reviewing.

Actually I implemented the solution at first as same as your suggestion. However it fails sometime when I test the solution by slowing the job failing a lot. There are some unexpected restorations. The assertion of `illegalRestores` fails because it is greater than zero. I believe the reason is also the race condition of `CheckpointFailureManager`.

It seems to need more changes than I expected, so I abandoned the implementation. I could give a deeper analysis later about this.

tillrohrmann:
I think it is definitely good enough as a bandaid because we see the test failing quite often. Maybe we merge it as is and remove it once the underlying problem has been properly fixed. I agree that this test is also quite complicated with a lot of conditions which makes it quite hard to change.

ifndef-SleePy:
Hi @tillrohrmann , I think we could merge this PR. Actually this `blockSnapshotLatch` only works when the race condition happens. It should only block the checkpoint 7. Because normally this test job could not recover because lack of checkpoint file. So there should be no task deployed since checkpoint 6.

WRT another race condition of the alternative, I will check it whether it's a bug or not. We could talk about it in another thread if needed.

ifndef-SleePy:
@tillrohrmann , thanks for reviewing. Travis has given a green light.

ifndef-SleePy:
This PR has been merged, https://github.com/apache/flink/commit/51582b6d837aeb646686a1497514f51cf5fa0702.
Will close it.

PR #9271, opened 2019-07-30
[FLINK-13384][1.9][runtime] Fix back pressure sampling for SourceStreamTask

GJL:

*This fixes back pressure sampling for sources.*

cc: @pnowojski 

## Brief change log

  - *Fix back pressure sampling for SourceStreamTask*

## Verifying this change

  - *Added unit tests to `TaskTest`*
  - *Manually verified the change back pressure sampling for sources*

## Documentation


flink-runtime/src/main/java/org/apache/flink/runtime/jobgraph/tasks/AbstractInvokable.java
```diff
@@ -123,6 +126,17 @@ public boolean shouldInterruptOnCancel() {
...
 	}
 
+	/**
+	 * If the invokable implementation executes user code in a thread other than,
+	 * {@link Task#getExecutingThread()}, this method returns the stack trace of that executing
+	 * thread.
+	 *
+	 * @see Task#getStackTraceOfExecutingThread()
+	 */
+	public Optional<StackTraceElement[]> getStackTraceOfExecutingThread() {
```
pnowojski: nit: maybe change this to `Optional<Thread> getExecutingThread()`? It would feel somehow a bit cleaner for me

flink-runtime/src/test/java/org/apache/flink/runtime/taskmanager/TaskTest.java
```diff
@@ -922,6 +926,40 @@ public void testTerminationFutureCompletesOnErrorInInvoke() throws Exception {
 		assertEquals(ExecutionState.FAILED, task.getTerminationFuture().getNow(null));
 	}
 
+	@Test
```
pnowojski: Either instead of those unit tests or additionally to them, could we add an ITCase which would validate that some simple `DataStream` application with a source and not chained task correctly reports a backpressure?

GJL: Done. I think we can delete `BackPressureStatsTrackerImplITCase`. WDYT?

flink-runtime/src/main/java/org/apache/flink/runtime/jobgraph/tasks/AbstractInvokable.java
GJL: Ok, `Optional<StackTraceElement[]>` feels awkward.

GJL: Done.

flink-tests/src/test/java/org/apache/flink/test/streaming/runtime/BackPressureITCase.java
```diff
@@ -0,0 +1,165 @@
...
+		final int memorySegmentSizeKb = 32;
+		final String networkBuffersMemory = (memorySegmentSizeKb * NUM_TASKS) + "kb";
+
+		configuration.setString(TaskManagerOptions.MEMORY_SEGMENT_SIZE, memorySegmentSizeKb + "kb");
+		configuration.setString(NettyShuffleEnvironmentOptions.NETWORK_BUFFERS_MEMORY_MIN, networkBuffersMemory);
+		configuration.setString(NettyShuffleEnvironmentOptions.NETWORK_BUFFERS_MEMORY_MAX, networkBuffersMemory);
+		return configuration;
+	}
+
+	@Test
```
pnowojski: Just a control question: was this test failing before your fix?

flink-runtime/src/test/java/org/apache/flink/runtime/taskmanager/TaskTest.java
pnowojski: I'm not entirely sure. Technically speaking with new test we are only checking for the back pressure presence in `DataStream` api. However old test `BackPressureStatsTrackerImplITCase` was not doing much, since it was asserting that custom task manually invokes `requestBufferBuilderBlocking`, which I would say is superseded and covered even better by the new test (since it's at least using production `StreamTask` class).

So I think I would drop the old test, but it's a pitty that we do not have a coverage for `DataSet` api.

flink-tests/src/test/java/org/apache/flink/test/streaming/runtime/BackPressureITCase.java
GJL: yes it fails without my fix

flink-runtime/src/test/java/org/apache/flink/runtime/taskmanager/TaskTest.java
GJL: Let's delete it.

GJL: > Either instead of those unit tests or additionally to them [...]

I decided to drop the unit tests here except for `testReturnsEmptyStackTraceIfTaskNotStarted` because
- there was a race condition in `testReturnsStackTraceOfTaskExecutingThread` and `testReturnsStackTraceOfAbstractInvokableExecutingThread`
- they are covered by `BackPressureITCase`

GJL: `BackPressureStatsTrackerImplITCase` is deleted

GJL:
Kafka 0.10 E2E test fails due to:

```
java.lang.Exception: org.apache.kafka.common.config.ConfigException: Invalid value org.apache.kafka.common.serialization.ByteArrayDeserializer for conf
iguration value.deserializer: Class org.apache.kafka.common.serialization.ByteArrayDeserializer could not be found.
        at org.apache.flink.streaming.runtime.tasks.SourceStreamTask$LegacySourceFunctionThread.checkThrowSourceExecutionException(SourceStreamTask.jav
a:212)
        at org.apache.flink.streaming.runtime.tasks.SourceStreamTask.performDefaultAction(SourceStreamTask.java:132)
        at org.apache.flink.streaming.runtime.tasks.StreamTask.run(StreamTask.java:298)
        at org.apache.flink.streaming.runtime.tasks.StreamTask.invoke(StreamTask.java:403)
        at org.apache.flink.runtime.taskmanager.Task.doRun(Task.java:700)
        at org.apache.flink.runtime.taskmanager.Task.run(Task.java:530)
Caused by: org.apache.kafka.common.config.ConfigException: Invalid value org.apache.kafka.common.serialization.ByteArrayDeserializer for configuration
value.deserializer: Class org.apache.kafka.common.serialization.ByteArrayDeserializer could not be found.
        at org.apache.kafka.common.config.ConfigDef.parseType(ConfigDef.java:690)
        at org.apache.kafka.common.config.ConfigDef.parse(ConfigDef.java:433)
        at org.apache.kafka.common.config.AbstractConfig.<init>(AbstractConfig.java:56)
        at org.apache.kafka.common.config.AbstractConfig.<init>(AbstractConfig.java:63)
        at org.apache.kafka.clients.consumer.ConsumerConfig.<init>(ConsumerConfig.java:426)
        at org.apache.kafka.clients.consumer.KafkaConsumer.<init>(KafkaConsumer.java:597)
        at org.apache.kafka.clients.consumer.KafkaConsumer.<init>(KafkaConsumer.java:579)
        at org.apache.flink.streaming.connectors.kafka.internal.KafkaConsumerThread.getConsumer(KafkaConsumerThread.java:502)
        at org.apache.flink.streaming.connectors.kafka.internal.KafkaConsumerThread.run(KafkaConsumerThread.java:181)
```

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/Task.java
```diff
@@ -593,6 +593,10 @@ else if (current == ExecutionState.CANCELING) {
 			LOG.info("Loading JAR files for task {}.", this);
 
 			userCodeClassLoader = createUserCodeClassloader();
+
+			// make sure the user code classloader is accessible thread-locally
+			executingThread.setContextClassLoader(userCodeClassLoader);
```
pnowojski: Maybe moving this just before
```
invokable = loadAndInstantiateInvokable(userCodeClassLoader, nameOfInvokableClass, env);
```
would be more clean and safe? And maybe add some comment like
```
// We are setting the correct context class loader for the whole interaction with the `invokable`
```
? 

The contract that we would introduce here, is that user code should be executed with the `userCodeClassLoader`. Before constructing the invokable, we are not touching any user code, so there should be no point (and could potentially do some harm) in setting context class loader so early.

Arguably this is a separate bug fix that just wasn't discovered before.

GJL: > Arguably this is a separate bug fix that just wasn't discovered before.

I can keep the change as a separate commit that I add before the other commits.

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/Task.java
```diff
@@ -674,6 +674,11 @@ else if (current == ExecutionState.CANCELING) {
 				metrics,
 				this);
 
+			// Make sure the user code classloader is accessible thread-locally.
+			// We are setting the correct context class loader before instantiating the invokable
+			// so that it is available to the invokable during its entire lifetime.
+			executingThread.setContextClassLoader(userCodeClassLoader);
```
GJL: I will move this commit to the beginning before pushing to master/release branch.

pnowojski:
Looks even better to me :)

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/Task.java
1u0: Imo, the issue with class loader is happening to the change that introduced in this PR.

This happens, because you are eagerly creating `LegacySourceFunctionThread` and it inherits 
default class loader. So to fix your issue, you are either need to set class loader in `LegacySourceFunctionThread.run()` or on the thread that instantiates the `invokable`.

Instead of setting the `userCodeClassLoader` eagerly, I propose to revert any changes related to it. You can instantiate the `LegacySourceFunctionThread` field reference in the `SourceStreamTask` later: 
 * as it was instantiated before in `SourceStreamTask::performDefaultAction()`;
 * or move the `LegacySourceFunctionThread` creation to `SourceStreamTask::init()`.

GJL:
@1u0 Why is your solution better? Now the `LegacySourceFunctionThread` field can be `final`, and we can always return a non empty `Optional<Thread>`. Not being able to access the user class loader in the constructor but in `invoke` is also strange.

1u0:
> @1u0 Why is your solution better? Now the `LegacySourceFunctionThread` field can be `final`, and we can always return a non empty `Optional<Thread>` from `SourceStreamTask#getExecutingThread()`. Not being able to access the user class loader in the constructor but in `invoke` is also strange.

I assumed that instantiation of `invokable` (StreamTask) has different, more loose contract (not requiring `userCodeClassLoader` set). It's also a little bit misleading that `Task.doRun()` calls `executingThread.setContextClassLoader(userCodeClassLoader)` where in fact the `executingThread == Thread.currentThread()`.

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/Task.java
1u0: **Update:** it looks like that in this context (`Task::doRun()`) `executingThread` is supposed to be the same as `Thread.currentThread()`.
What about changing `executingThread.setContextClassLoader(userCodeClassLoader)` to `Thread.currentThread().setContextClassLoader(userCodeClassLoader)` just before
```
invokable = loadAndInstantiateInvokable(...);
```
?
This also may help some tests, that call `Task.run()` directly.

GJL:
Imo using 

```
Thread.currentThread().setContextClassLoader(userCodeClassLoader)
```

doesn't make the code more maintainable. Moreover, tests that currently call `Task#run()` directly can only break if we introduce this change (also note that we do not unset the context class loader). FLINK-13384 is a bug ticket, and I'd refrain from changing more behavior unless absolutely necessary. If a code restructuring/refactoring is needed, we can do it in a separate ticket.

GJL:
Rebased. Merging as soon as build is green.

GJL:
Merged

PR #9272, opened 2019-07-30
[FLINK-13435] Remove ShuffleDescriptor.ReleaseType and make release semantics fixed per partition type

azagrebin:

In a long term we do not need auto-release semantics for blocking (persistent) partition. We expect them always to be released externally by JM and assume they can be consumed multiple times.

The pipelined partitions have always only one consumer and one consumption attempt. Afterwards they can be always released automatically.

`ShuffleDescriptor#ReleaseType` was introduced to make release semantics more flexible but it is not needed in a long term.

`FORCE_PARTITION_RELEASE_ON_CONSUMPTION` was introduced as a safety net to be able to fallback to 1.8 behaviour without the partition tracker and JM taking care about blocking partition release. We can make this option specific for `NettyShuffleEnvironment` which was the only existing shuffle service before. If it is activated then the blocking partition is also auto-released on a consumption attempt as it was before. The fine-grained recovery will just not find the partition after the job restart in this case and will restart the producer.

## Brief change log

  - Remove `ShuffleDescriptor#ReleaseType` and assume predefined release semantics for certain partition type
  - Adjust test and other involved code
  - Move `FORCE_PARTITION_RELEASE_ON_CONSUMPTION` from `JobManagerOptions` to `NettyShuffleEnvironmentOptions`
  - codestyle fixes

## Verifying this change

existing tests

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/ResultPartitionFactory.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/ResultPartitionFactory.java
@@ -33,8 +33,6 @@ import org.apache.flink.util.function.FunctionWithException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import javax.annotation.Nonnull;
-
 import java.io.File;
 import java.io.IOException;
 import java.util.Optional;
@@ -46,13 +44,10 @@ public class ResultPartitionFactory {
 
 	private static final Logger LOG = LoggerFactory.getLogger(ResultPartitionFactory.class);
 
-	@Nonnull
 	private final ResultPartitionManager partitionManager;
 
-	@Nonnull
 	private final FileChannelManager channelManager;
 
-	@Nonnull
 	private final BufferPoolFactory bufferPoolFactory;
 
 	private final BoundedBlockingSubpartitionType blockingSubpartitionType;
@@ -66,9 +61,9 @@ public class ResultPartitionFactory {
 	private final boolean forcePartitionReleaseOnConsumption;
 
 	public ResultPartitionFactory(
-		@Nonnull ResultPartitionManager partitionManager,
-		@Nonnull FileChannelManager channelManager,
-		@Nonnull BufferPoolFactory bufferPoolFactory,
+		ResultPartitionManager partitionManager,
+		FileChannelManager channelManager,
+		BufferPoolFactory bufferPoolFactory,
 		BoundedBlockingSubpartitionType blockingSubpartitionType,
 		int networkBuffersPerChannel,
 		int floatingNetworkBuffersPerGate,
@@ -86,9 +81,8 @@ public class ResultPartitionFactory {
 	}
 
 	public ResultPartition create(
-		@Nonnull String taskNameWithSubtaskAndId,
-		@Nonnull ResultPartitionDeploymentDescriptor desc) {
-
+			String taskNameWithSubtaskAndId,
+			ResultPartitionDeploymentDescriptor desc) {
 		return create(
 			taskNameWithSubtaskAndId,
 			desc.getShuffleDescriptor().getResultPartitionID(),
@@ -100,13 +94,12 @@ public class ResultPartitionFactory {
 
 	@VisibleForTesting
 	public ResultPartition create(
-		@Nonnull String taskNameWithSubtaskAndId,
-		@Nonnull ResultPartitionID id,
-		@Nonnull ResultPartitionType type,
-		int numberOfSubpartitions,
-		int maxParallelism,
-		FunctionWithException<BufferPoolOwner, BufferPool, IOException> bufferPoolFactory) {
-
+			String taskNameWithSubtaskAndId,
+			ResultPartitionID id,
+			ResultPartitionType type,
+			int numberOfSubpartitions,
+			int maxParallelism,
+			FunctionWithException<BufferPoolOwner, BufferPool, IOException> bufferPoolFactory) {
 		ResultSubpartition[] subpartitions = new ResultSubpartition[numberOfSubpartitions];
 
 		ResultPartition partition = forcePartitionReleaseOnConsumption || !type.isBlocking()
@@ -139,10 +132,10 @@ public class ResultPartitionFactory {
 			ResultPartitionType type,
 			BoundedBlockingSubpartitionType blockingSubpartitionType,
 			ResultSubpartition[] subpartitions) {
-
 		// Create the subpartitions.
 		switch (type) {
 			case BLOCKING:
+			case BLOCKING_PERSISTENT:
 				initializeBoundedBlockingPartitions(subpartitions, partition, blockingSubpartitionType, networkBufferSize, channelManager);
 				break;
 
@@ -160,15 +153,14 @@ public class ResultPartitionFactory {
 	}
 
 	private static void initializeBoundedBlockingPartitions(
-		ResultSubpartition[] subpartitions,
-		ResultPartition parent,
-		BoundedBlockingSubpartitionType blockingSubpartitionType,
-		int networkBufferSize,
-		FileChannelManager channelManager) {
-
+			ResultSubpartition[] subpartitions,
+			ResultPartition parent,
+			BoundedBlockingSubpartitionType blockingSubpartitionType,
+			int networkBufferSize,
+			FileChannelManager channelManager) {
 		int i = 0;
 		try {
-			for (; i < subpartitions.length; i++) {
+			for (i = 0; i < subpartitions.length; i++) {
 				final File spillFile = channelManager.createChannel().getPathFile();
 				subpartitions[i] = blockingSubpartitionType.create(i, parent, spillFile, networkBufferSize);
 			}
@@ -194,8 +186,8 @@ public class ResultPartitionFactory {
 
 	@VisibleForTesting
 	FunctionWithException<BufferPoolOwner, BufferPool, IOException> createBufferPoolFactory(
-		int numberOfSubpartitions, ResultPartitionType type) {
-
+			int numberOfSubpartitions,
+			ResultPartitionType type) {
 		return p -> {
 			int maxNumberOfMemorySegments = type.isBounded() ?
 				numberOfSubpartitions * networkBuffersPerChannel + floatingNetworkBuffersPerGate : Integer.MAX_VALUE;
@@ -213,6 +205,7 @@ public class ResultPartitionFactory {
 				return BoundedBlockingSubpartitionType.FILE_MMAP;
 			case _32_BIT:
 				return BoundedBlockingSubpartitionType.FILE;
+			case UNKNOWN:
 			default:
 				LOG.warn("Cannot determine memory architecture. Using pure file-based shuffle.");
 				return BoundedBlockingSubpartitionType.FILE;
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/ResultPartitionFactory.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/ResultPartitionFactory.java
@@ -63,6 +63,8 @@ public class ResultPartitionFactory {
 
 	private final int networkBufferSize;
 
+	private final boolean forcePartitionReleaseOnConsumption;
+
 	public ResultPartitionFactory(
 		@Nonnull ResultPartitionManager partitionManager,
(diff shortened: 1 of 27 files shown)


flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/NettyShuffleEnvironmentConfiguration.java
```diff
@@ -29,8 +29,8 @@
 import org.apache.flink.runtime.io.network.netty.NettyConfig;
 import org.apache.flink.runtime.io.network.partition.BoundedBlockingSubpartitionType;
 import org.apache.flink.runtime.util.ConfigurationParserUtils;
-
```
zhijiangW: This change should be a separate hotfix

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/NettyShuffleEnvironmentConfiguration.java
```diff
@@ -73,6 +73,8 @@
 
 	private final BoundedBlockingSubpartitionType blockingSubpartitionType;
 
+	private final boolean forcePartitionReleaseOnConsumption;
```
zhijiangW: Not sure to also cover this new field in following `hashCode`, `toString` and `equals` methods.

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleEnvironment.java
```diff
@@ -61,17 +60,14 @@
...
- *     {@link ShuffleMaster#releasePartitionExternally(ShuffleDescriptor)} and {@link ShuffleEnvironment#releasePartitionsLocally(Collection)},
- *     if it occupies any producer local resources ({@link ShuffleDescriptor#storesLocalResourcesOn()}),
- *     are called outside of the producer thread, e.g. to manage the lifecycle of BLOCKING result partitions
- *     which can outlive their producers. The partition has to support the {@link ReleaseType#MANUAL} in
- *     {@link ShuffleDescriptor#getSupportedReleaseTypes()}.</li>
+ *     <li>for bounded pipelined partitions if the production is done ({@link ResultPartitionWriter#finish()} and
+ *     {@link ResultPartitionWriter#close()} are called) and there was a detected consumption attempt.
+ *     Only one consumption attempt is ever expected for the pipelined partition at the moment so
+ *     it cqn be released afterwards.
+ *     <li>{@link ShuffleMaster#releasePartitionExternally(ShuffleDescriptor)} and {@link ShuffleEnvironment#releasePartitionsLocally(Collection)},
```
zhijiangW: There might some confusing here. `ShuffleMaster#releasePartitionExternally` is mainly used for releasing external resources outside of local producer. So it is not very inconsistent with `if it occupies any producer local resources`.

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleEnvironment.java
```diff
@@ -61,17 +60,14 @@
...
+ *     if it occupies any producer local resources ({@link ShuffleDescriptor#storesLocalResourcesOn()}), are called
+ *     outside of the producer thread, e.g. to manage the lifecycle of BLOCKING or pipelined bounded result partitions
```
zhijiangW: also lowercase for `BLOCKING` consistent with pipelined

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleEnvironment.java
```diff
@@ -61,17 +60,14 @@
...
- *     <li>if {@link ResultPartitionDeploymentDescriptor#isReleasedOnConsumption()} is {@code false} and
- *     {@link ShuffleMaster#releasePartitionExternally(ShuffleDescriptor)} and {@link ShuffleEnvironment#releasePartitionsLocally(Collection)},
- *     if it occupies any producer local resources ({@link ShuffleDescriptor#storesLocalResourcesOn()}),
- *     are called outside of the producer thread, e.g. to manage the lifecycle of BLOCKING result partitions
- *     which can outlive their producers. The partition has to support the {@link ReleaseType#MANUAL} in
- *     {@link ShuffleDescriptor#getSupportedReleaseTypes()}.</li>
+ *     <li>for bounded pipelined partitions if the production is done ({@link ResultPartitionWriter#finish()} and
+ *     {@link ResultPartitionWriter#close()} are called) and there was a detected consumption attempt.
+ *     Only one consumption attempt is ever expected for the pipelined partition at the moment so
+ *     it cqn be released afterwards.
```
zhijiangW: typo cqn

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleEnvironment.java
```diff
@@ -61,17 +60,14 @@
...
- *     The actual release can take some time depending on implementation details,
- *     e.g. if the `end of consumption' confirmation from the consumer is being awaited implicitly.
- *     The partition has to support the {@link ReleaseType#AUTO} in {@link ShuffleDescriptor#getSupportedReleaseTypes()}.</li>
- *     <li>if {@link ResultPartitionDeploymentDescriptor#isReleasedOnConsumption()} is {@code false} and
- *     {@link ShuffleMaster#releasePartitionExternally(ShuffleDescriptor)} and {@link ShuffleEnvironment#releasePartitionsLocally(Collection)},
- *     if it occupies any producer local resources ({@link ShuffleDescriptor#storesLocalResourcesOn()}),
- *     are called outside of the producer thread, e.g. to manage the lifecycle of BLOCKING result partitions
- *     which can outlive their producers. The partition has to support the {@link ReleaseType#MANUAL} in
- *     {@link ShuffleDescriptor#getSupportedReleaseTypes()}.</li>
+ *     <li>for bounded pipelined partitions if the production is done ({@link ResultPartitionWriter#finish()} and
```
zhijiangW: I am a bit confused here. From the following comments of `Only one consumption attempt is ever expected for the pipelined partition`, I guess the `bounded pipelined partitions` is only meaning current pipelined partitions. Then for pipelined partition we do not need the production done to trigger the consumption attempt.

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/PartitionTestUtils.java
```diff
@@ -60,9 +57,9 @@ public static ResultPartition createPartition(ResultPartitionType type, FileChan
 	}
 
 	public static ResultPartition createPartition(
-			NettyShuffleEnvironment environment,
-			ResultPartitionType partitionType,
-			int numChannels) {
```
zhijiangW: irrelevant changes.

flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/PartitionTestUtils.java
```diff
@@ -72,10 +69,10 @@ public static ResultPartition createPartition(
 	}
 
 	public static ResultPartition createPartition(
-			NettyShuffleEnvironment environment,
-			FileChannelManager channelManager,
-			ResultPartitionType partitionType,
-			int numChannels) {
```
zhijiangW: ditto: irrelevant changes

flink-runtime/src/main/java/org/apache/flink/runtime/io/network/partition/ResultPartitionFactory.java
```diff
@@ -139,10 +132,10 @@ private void createSubpartitions(
 			ResultPartitionType type,
 			BoundedBlockingSubpartitionType blockingSubpartitionType,
 			ResultSubpartition[] subpartitions) {
-
 		// Create the subpartitions.
 		switch (type) {
 			case BLOCKING:
+			case BLOCKING_PERSISTENT:
```
zhijiangW: I think we might need a uint test for covering this change.

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleDescriptor.java
```diff
@@ -41,14 +41,14 @@
 	 * that the producer of the partition (consumer input channel) has not been scheduled
 	 * and its location and other relevant data is yet to be defined.
 	 * To proceed with the consumer deployment, currently unknown input channels have to be
-	 * marked with placeholders which are special implementation of {@link ShuffleDescriptor}:
+	 * marked with placeholders which are special implementation of shuffle descriptor:
```
zhijiangW: implementation -> implementations?

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleDescriptor.java
```diff
@@ -41,14 +41,14 @@
...
 	 * {@link UnknownShuffleDescriptor}.
 	 *
 	 * <p>Note: this method is not supposed to be overridden in concrete shuffle implementation.
 	 * The only class where it returns {@code true} is {@link UnknownShuffleDescriptor}.
 	 *
 	 * @return whether the partition producer has been ever deployed and
-	 * the corresponding {@link ShuffleDescriptor} is obtained from the {@link ShuffleMaster} implementation.
+	 * the corresponding shuffle descriptor is obtained from the {@link ShuffleMaster} implementation.
```
zhijiangW: Are there any concerns for this change?

zhijiangW:
@azagrebin I have finished the whole review and left some nit comments besides the above concerns.

azagrebin:
Thanks for the review @zhijiangW ! I will go through the smaller comments. For general concerns:

- I think this PR should fit to the overall change we plan with #9250 and FLINK-13476. In #9250 we basically stop using explicitly any notion of 'release on consumption' and assume that blocking partitions always have to be released by JM/TM. FLINK-13476 should also fix it for pipelined bounded until we couple Task finished state with the consumption confirmation attempt, basically with the partition state. Not sure, there are other ways for the direct release internally in netty shuffle service.

- FLINK-13478 and FLINK-13493 seem to be independent issues or do I miss something?

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleEnvironment.java
azagrebin: I will rephrase it a bit, I meant that basically only bounded pipelined can be ever done.

flink-runtime/src/main/java/org/apache/flink/runtime/shuffle/ShuffleDescriptor.java
azagrebin: there are some check style warnings in Idea IDE to not use self links

azagrebin:
Thanks for the review @zhijiangW ! I have addressed the comments

zhijiangW:
Thanks for the updates @azagrebin !

Almost LGTM on my side, only confirm one thing before merging:
I thought `FORCE_PARTITION_RELEASE_ON_CONSUMPTION` only suitable for pipelined partitions. If we make this option true for blocking partitions, then the partition release should be still controlled by JM always. If so, do we need to avoid generate `ReleaseOnConsumptionResultPartition` for blocking partitions in `ResultPartitionFactory`? In current implementation, the `ReleaseOnConsumptionResultPartition` would be constructed if `forcePartitionReleaseOnConsumption || !type.isBlocking()`.

I wonder the blocking partition release controlled by `ReleaseOnConsumptionResultPartition` might cause some problem because of FLINK-13493.

azagrebin:
Thanks for the review @zhijiangW !

I mentioned the explanation in FLINK-13493. `forcePartitionReleaseOnConsumption` is just a safety net to fallback to the old 1.8 behaviour and should not be used for 1.9 in general. JM does not use this option and always tries to take care about release (as you mentioned it should be always the case) so `forcePartitionReleaseOnConsumption` will just enforce/speed up release in certain cases and degrade fine-grained recovery. In future, we can remove `forcePartitionReleaseOnConsumption` at all and rely only on JM.

zhijiangW:
Thanks for the explanation @azagrebin !

I understand it, wish the temporary `forcePartitionReleaseOnConsumption` removed soon to avoid unnecessary concern here. :)
Will merge after travis green.

zhijiangW:
Unrelated python component failure in travis, merging.

PR #9274, opened 2019-07-30
[FLINK-13495][table-planner-blink] blink-planner should support varchar/char/decimal precision to connector

JingsongLi:

Now there is an exception when use DataTypes.DECIMAL(5, 2) to table source when use blink-planner.
Some conversions between DataType and TypeInfo loose precision information.

## Verifying this change

TableSourceITCase

  - The serializers: no

## Documentation

### DIFF
--- a/flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/plan/nodes/physical/PhysicalTableSourceScan.scala
+++ b/flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/plan/nodes/physical/PhysicalTableSourceScan.scala
@@ -18,11 +18,15 @@
 
 package org.apache.flink.table.planner.plan.nodes.physical
 
+import org.apache.flink.api.common.io.InputFormat
+import org.apache.flink.api.common.typeinfo.TypeInformation
 import org.apache.flink.api.dag.Transformation
+import org.apache.flink.core.io.InputSplit
 import org.apache.flink.streaming.api.environment.StreamExecutionEnvironment
 import org.apache.flink.table.planner.calcite.FlinkTypeFactory
 import org.apache.flink.table.planner.plan.schema.{FlinkRelOptTable, TableSourceTable}
-import org.apache.flink.table.sources.{StreamTableSource, TableSource}
+import org.apache.flink.table.runtime.types.TypeInfoDataTypeConverter.fromDataTypeToTypeInfo
+import org.apache.flink.table.sources.{InputFormatTableSource, StreamTableSource, TableSource}
 
 import org.apache.calcite.plan.{RelOptCluster, RelTraitSet}
 import org.apache.calcite.rel.RelWriter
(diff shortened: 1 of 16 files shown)


lirui-apache:
@JingsongLi As we discussed offline, `SinkCodeGeneratorcheckRowConverterValid` also needs to be updated, otherwise only system default `decimal(38, 18)` is supported.

JingsongLi:
> @JingsongLi As we discussed offline, `SinkCodeGeneratorcheckRowConverterValid` also needs to be updated, otherwise only system default `decimal(38, 18)` is supported.

@lirui-apache I updated PR, I add all new-api it cases to ensure everything running well. You can take a look to ensure hive connector running well.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/plan/nodes/physical/PhysicalTableSourceScan.scala
```diff
@@ -60,8 +64,14 @@ abstract class PhysicalTableSourceScan(
   def getSourceTransformation(
       streamEnv: StreamExecutionEnvironment): Transformation[_] = {
     if (sourceTransform == null) {
-      sourceTransform = tableSource.asInstanceOf[StreamTableSource[_]].
-        getDataStream(streamEnv).getTransformation
+      sourceTransform = tableSource match {
+        case format: InputFormatTableSource[_] =>
+          streamEnv.createInput(
+            format.getInputFormat.asInstanceOf[InputFormat[Any, _ <: InputSplit]],
+            fromDataTypeToTypeInfo(format.getProducedDataType).asInstanceOf[TypeInformation[Any]]
```
godfreyhe: please add some comments about why not call `InputFormatTableSource#getDataStream` directly

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/runtime/types/PlannerTypeUtils.java
```diff
@@ -110,4 +110,45 @@ public static boolean isInteroperable(LogicalType t1, LogicalType t2) {
 				return t1.copy(true).equals(t2.copy(true));
 		}
 	}
+
+	/**
+	 * Now in the conversion to the TypeInformation from DataType, type may loose some information
+	 * about nullable and precision. So we add this method to do a soft check.
+	 * The difference of {@link #isInteroperable} is ignore decimal precision.
+	 */
+	public static boolean isAssignable(LogicalType t1, LogicalType t2) {
```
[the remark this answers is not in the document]
JingsongLi: It is so short and the logical is simple, and it is invoked by only one place. But it is reasonable to add test.

JingsongLi:
@godfreyhe addressed your comments, please take a look again.
@wuchong can you take a look too?

flink-table/flink-table-planner-blink/src/test/scala/org/apache/flink/table/planner/utils/testTableSources.scala
```diff
@@ -620,3 +622,39 @@ class TestInputFormatTableSource[T](
...
+class TestStreamTableSource(
+    tableSchema: TableSchema,
+    values: Seq[Row]) extends StreamTableSource[Row] {
+
+  override def getDataStream(execEnv: StreamExecutionEnvironment): DataStream[Row] = {
+    execEnv.fromCollection(values, tableSchema.toRowType)
+  }
+
+  override def getReturnType: TypeInformation[Row] =
+    throw new RuntimeException("Should not invoke this deprecated method.")
```
wuchong: I would like to implement this method with returning `tableSchema.toRowType` instead of implementing `getProducedDataType`. Because currently, the behavior of `getProducedDataType` and return type of `getDataStream` is not align. For example, if users pass in a DECIMAL(5, 2), it might fail.

JingsongLi: ok

JingsongLi:
@wuchong Thanks your advice, comment addressed now, What do you think?

wuchong:
Merging...

wuchong:
Commits have been merged via:

49202e2cf849dce7df3856ec9f40b1b5596a7ae9
ebf18db29b9b8480459e20b307e07980657e75e5
9dddae827e3f117b3527254701119c10909a3ac5
e2a8b95e4703e77a963cee5641bffe2e5b8a6f49
5b90b16b1d2d4dc48a0edfd12844ab1ada871586

PR #9275, opened 2019-07-30
[FLINK-13290][table-planner-blink][hbase] SinkCodeGenerator should not compare row type field names and enable blink planner for hbase IT case

wuchong:

Fix `SinkCodeGenerator` shouldn't compare type field names. Also we added blink planner for hbase IT cases.

## Brief change log

- `SinkCodeGenerator` still use type information to compare
- make hbase IT case to run with blink planner and flink planner both.

## Verifying this change

- We added some IT case in blink planner to verify `SinkCodeGenerator`

## Documentation


wuchong:
cc @carp84  could you help review the hbase part?
cc @JingsongLi @godfreyhe welcome to review when you are free.

flink-connectors/flink-hbase/src/test/java/org/apache/flink/addons/hbase/HBaseLookupSourceITCase.java
```diff
@@ -163,6 +165,7 @@ private static Put putRow(int rowKey, int f1c1, String f2c1, long f2c2, double f
 		return descriptorProperties.asMap();
 	}
 
+	@Ignore
```
JingsongLi: Why ignore?

flink-connectors/flink-hbase/src/test/java/org/apache/flink/addons/hbase/HBaseSinkITCase.java
```diff
@@ -52,9 +55,13 @@
 /**
  * IT case Test for {@link HBaseUpsertTableSink}.
  */
+@RunWith(Parameterized.class)
 public class HBaseSinkITCase extends HBaseTestingClusterAutostarter {
 	private static final long serialVersionUID = 1L;
 
+	private static final String BLINK_PLANNER = "Blink";
```
JingsongLi: use enum instead?

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/codegen/SinkCodeGenerator.scala
```diff
@@ -108,8 +108,7 @@ object SinkCodeGenerator {
     val inputTerm = CodeGenUtils.DEFAULT_INPUT1_TERM
     var afterIndexModify = inputTerm
     val fieldIndexProcessCode =
-      if (getCompositeTypes(convertOutputType).map(fromTypeInfoToLogicalType) sameElements
-          inputTypeInfo.getFieldTypes.map(fromTypeInfoToLogicalType)) {
+      if (getCompositeTypes(convertOutputType) sameElements inputTypeInfo.getFieldTypes) {
```
JingsongLi: it not work now...
`inputTypeInfo` is a BaseRowTypeInfo with internal type informations.
And `convertOutputType` is the type information from sink and users.
Consider Types.STRING and BinaryStringTypeInfo, Types.DECIMAL and DecimalTypeInfo.

JingsongLi: And I think maybe you can try `LogicalType.asSerializableString`

flink-connectors/flink-hbase/src/test/java/org/apache/flink/addons/hbase/HBaseLookupSourceITCase.java
```diff
@@ -200,4 +203,44 @@ public void testHBaseLookupFunction() throws Exception {
 
 		StreamITCase.compareWithList(expected);
 	}
+
+	@Test
+	public void testHBaseLookupTableSource() throws Exception {
+		EnvironmentSettings settings = EnvironmentSettings.newInstance()
+			// only blink planner support lookup table source
+			.useBlinkPlanner().inStreamingMode().build();
```
carp84: We should use parameterized test to both cover blink-planner and flink-planner just like the way in `HBaseSinkITCase`, and remove the `testHBaseLookupFunction` test.

wuchong:
Hi @carp84 , @JingsongLi , I have updated the PR.

I split it into 3 commits:

- 0a59bfc570f793a54b2b3d14063858934a67a1fc: Add method to equals LogicalType without field names
- 66492281e2ad34ac737ae7f87945ab60276ea0ed: SinkCodeGenerator should not compare row type field names
- cef04c07f422c8d0190780e6dc3ab5a02dc798a2: Enable blink planner for integration tests of HBase


cc @twalthr , it would be great if you can help to review 0a59bfc570f793a54b2b3d14063858934a67a1fc.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/logical/utils/LogicalTypeChecks.java
```diff
@@ -145,6 +156,14 @@ public static boolean isSingleFieldInterval(LogicalType logicalType) {
 		return logicalType.accept(SINGLE_FIELD_INTERVAL_EXTRACTOR);
 	}
 
+	public static boolean areTypeEqualsWithoutNames(LogicalType thisType, LogicalType thatType) {
+		if (thisType == null || thatType == null) {
```
twalthr: It's better to add a precondition here. Types should not be null. If they are null, it is a bug that should throw an exception.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/logical/utils/LogicalTypeChecks.java
```diff
@@ -332,4 +351,162 @@ public Boolean visit(DayTimeIntervalType dayTimeIntervalType) {
 			}
 		}
 	}
+
+	private static class TypeEquivalenceIgnoreNamesExtractor extends Extractor<Boolean> {
```
twalthr: Should directly extend from `LogicalTypeDefaultVisitor` because we override the default method anyway.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/logical/utils/LogicalTypeChecks.java
```diff
@@ -332,4 +351,162 @@ public Boolean visit(DayTimeIntervalType dayTimeIntervalType) {
...
+
+		private TypeEquivalenceIgnoreNamesExtractor(LogicalType thisType) {
+			this.thisType = thisType;
+		}
+
+		/**
+		 * Returns true if tow LogicalType can equal. This is the same with {@link LogicalType#equals(Object)}.
+		 */
+		private boolean canEqual(LogicalType thatType) {
+			if (thisType == null || thatType == null) {
```
twalthr: same comment as above regarding `null` checking

flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/logical/utils/LogicalTypeChecks.java
```diff
@@ -332,4 +351,162 @@ public Boolean visit(DayTimeIntervalType dayTimeIntervalType) {
...
+			return areTypeEqualsWithoutNames(thisKeyType, thatKeyType) &&
+				areTypeEqualsWithoutNames(thisValueType, thatValueType);
+		}
+
+		@Override
+		public Boolean visit(DistinctType thatType) {
+			if (!canEqual(thatType)) {
+				return false;
+			}
+			boolean thisIsFinal = ((DistinctType) thisType).isFinal();
```
twalthr: Distinct types are always final. We don't need to check that.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/logical/utils/LogicalTypeChecks.java
```diff
@@ -145,6 +156,14 @@ public static boolean isSingleFieldInterval(LogicalType logicalType) {
 		return logicalType.accept(SINGLE_FIELD_INTERVAL_EXTRACTOR);
 	}
 
+	public static boolean areTypeEqualsWithoutNames(LogicalType thisType, LogicalType thatType) {
```
twalthr: nit: rename to `hasEqualTypesWithoutNames`

flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/logical/utils/LogicalTypeChecks.java
```diff
@@ -332,4 +351,162 @@ public Boolean visit(DayTimeIntervalType dayTimeIntervalType) {
...
+			List<RowType.RowField> thatFields = thatType.getFields();
+			if (thisFields.size() != thatFields.size()) {
+				return false;
+			}
+			for (int i = 0; i < thisFields.size(); i++) {
+				RowType.RowField thisField = thisFields.get(i);
+				RowType.RowField thatField = thatFields.get(i);
+				// ignore field names
+				if (!thisField.getType().equals(thatField.getType())
+					|| !Objects.equals(thisField.getDescription(), thatField.getDescription())) {
```
twalthr: We do we check the description? Isn't it as meaningless as the field names?

flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/logical/utils/LogicalTypeChecks.java
```diff
@@ -332,4 +351,162 @@ public Boolean visit(DayTimeIntervalType dayTimeIntervalType) {
...
+			}
+			boolean thisIsFinal = ((DistinctType) thisType).isFinal();
+			ObjectIdentifier thisObjectIdentifier = ((DistinctType) thisType).getObjectIdentifier();
+			Optional<String> thisDescription = ((DistinctType) thisType).getDescription();
+			LogicalType thisSourceType = ((DistinctType) thisType).getSourceType();
+
+			return thisIsFinal == thatType.isFinal() &&
+				thisObjectIdentifier.equals(thatType.getObjectIdentifier()) &&
+				thisDescription.equals(thatType.getDescription()) &&
+				areTypeEqualsWithoutNames(thisSourceType, thatType.getSourceType());
```
twalthr: I think we can just compare the source type. Everything else is logical meta data, no?

flink-table/flink-table-common/src/main/java/org/apache/flink/table/types/logical/utils/LogicalTypeChecks.java
wuchong: What's the reason to use `hasEqualTypesWithoutNames`? It looks like whether it contains some equal types (might children types are equal).

wuchong:
Hi @twalthr , thanks for the advice. The implementation of type equals has been much simpler.

Please have a look again.

flink-connectors/flink-hbase/src/test/java/org/apache/flink/addons/hbase/util/HBaseTestBase.java
```diff
@@ -0,0 +1,157 @@
...
+		TableName tableName = TableName.valueOf(TEST_TABLE_1);
+		// column families
+		byte[][] families = new byte[][]{
+			Bytes.toBytes(FAMILY1),
+			Bytes.toBytes(FAMILY2),
+			Bytes.toBytes(FAMILY3)
+		};
+		// split keys
+		byte[][] splitKeys = new byte[][]{ Bytes.toBytes(4) };
+		createTable(tableName, families, splitKeys);
```
carp84: Minor: we could extract L101-L111 out to remove some duplicated codes.

flink-connectors/flink-hbase/src/test/java/org/apache/flink/addons/hbase/util/HBaseTestingClusterAutostarter.java
```diff
@@ -74,7 +75,7 @@
 //
 // https://github.com/apache/hbase/blob/master/hbase-server/src/test/java/org/apache/hadoop/hbase/filter/FilterTestingCluster.java
 //
-public class HBaseTestingClusterAutostarter extends TestLogger implements Serializable {
+public abstract class HBaseTestingClusterAutostarter extends AbstractTestBase {
```
carp84: Minor: HBaseTestingClusterAutostarter -> HBaseTestingClusterAutoStarter

flink-connectors/flink-hbase/src/test/java/org/apache/flink/addons/hbase/HBaseConnectorITCase.java
```diff
@@ -20,281 +18,464 @@
...
+				"Hello-7,700\n" +
+				"null,800\n";
+
+		TestBaseUtils.compareResultAsText(results, expected);
+	}
+
+	@Test
+	public void testTableInputFormat() throws Exception {
+		if (BLINK_PLANNER.equals(planner)) {
+			// this case is for testing TableInputFormat which is not works for flink-table
```
carp84: This comment is a little bit ambiguous, from the code logic it seems we should say "this case is for testing TableInputFormat which only applies to flink-table planner" or so?

flink-connectors/flink-hbase/src/test/java/org/apache/flink/addons/hbase/util/HBaseTestBase.java
```diff
@@ -0,0 +1,157 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.addons.hbase.util;
+
+import org.apache.flink.addons.hbase.HBaseConnectorITCase2;
```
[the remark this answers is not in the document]
carp84: Oops, it seems this issue is already resolved by the later force push, and now the code here looks good.

flink-connectors/flink-hbase/src/test/java/org/apache/flink/addons/hbase/HBaseConnectorITCase.java
wuchong: Not exactly. I mean `TableInputFormat` is a runtime implementation, not for flink table, not relative to what the planner is. You can see the test is only verified on `DataSet`, not on `TableEnvironment`.

wuchong:
As discussed with @twalthr , we would like to go with `areTypeCompatible` for the method. And add a detailed javadoc on the method.

PR updated and squashed.

flink-connectors/flink-hbase/src/test/java/org/apache/flink/addons/hbase/HBaseConnectorITCase.java
carp84: After some offline discussion, the case is irrelative to planner actually, but in case that later change somehow making the test case relative to planner implementation (although low possibility), let's just don't skip the test for any planner.

PR #9277, opened 2019-07-30
[FLINK-13494][table-planner-blink] Only use env parallelism for sql job

XuPingyong:

Only use env parallelism for sql job, not to calculate parallelism of table source and sink.

## Brief change log

Remove parallelism calculator in blink planner.

## Verifying this change

UT

  - The serializers: ( no)
  - Anything that affects deployment or recovery: JobManager (and its components), Checkpointing, Yarn/Mesos, ZooKeeper: (no )

## Documentation


flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/plan/nodes/physical/stream/StreamExecExchange.scala
```diff
@@ -104,6 +106,7 @@ class StreamExecExchange(
           inputTransform,
           partitioner.asInstanceOf[StreamPartitioner[BaseRow]])
         transformation.setOutputType(outputTypeInfo)
+        transformation.setParallelism(ExecutionConfig.PARALLELISM_DEFAULT)
```
wuchong: Could we use `planner.getExecEnv.getParallelism` instead of `ExecutionConfig.PARALLELISM_DEFAULT`? 
Set a -1 on exchange is too subtle to understand it is a `env.Parallelism`.

And it would be better to add a comment on this line, for example, "this enables all the downstream physical nodes to use `inputTransform.getParallelism` to get a good initial parallelism."

flink-table/flink-table-planner-blink/src/test/scala/org/apache/flink/table/api/batch/ExplainTest.scala
```diff
@@ -40,6 +40,12 @@ class ExplainTest(extended: Boolean) extends TableTestBase {
   val LONG = new BigIntType()
   val INT = new IntType()
 
+  @Before
+  def before(): Unit = {
+    util.tableEnv.getConfig.getConfiguration.setInteger(
+      ExecutionConfigOptions.TABLE_EXEC_RESOURCE_DEFAULT_PARALLELISM, 4)
```
wuchong: What's this used for?

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/delegation/PlannerBase.scala
```diff
@@ -148,6 +149,15 @@ abstract class PlannerBase(
     translateToPlan(execNodes)
   }
 
+  protected def overrideEnvParallelism(): Unit = {
```
wuchong: What about moving the implementation to `mergeParameters()`, so that `StreamPlanner` and `BatchPlanner` don't need to call this method again.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/plan/nodes/physical/stream/StreamExecDataStreamScan.scala
```diff
@@ -110,7 +110,6 @@ class StreamExecDataStreamScan(
     val config = planner.getTableConfig
```
wuchong: Remove `getSourceTransformation()` method in this file. It is never used.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/config/ExecutionConfigOptions.java
```diff
@@ -83,18 +83,8 @@
...
-		key("table.exec.resource.source.parallelism")
-			.defaultValue(-1)
-			.withDescription("Sets source parallelism, if it is <= 0, use " + TABLE_EXEC_RESOURCE_DEFAULT_PARALLELISM.key() + " to set source parallelism.");
-
-	public static final ConfigOption<Integer> TABLE_EXEC_RESOURCE_SINK_PARALLELISM =
-		key("table.exec.resource.sink.parallelism")
-			.defaultValue(-1)
-			.withDescription("Sets sink parallelism, if it is <= 0, use " + TABLE_EXEC_RESOURCE_DEFAULT_PARALLELISM.key() + " to set sink parallelism.");
+			.withDescription("Default parallelism to override parallelism of " +
+					"StreamExecutionEnvironment. If it is <= 0, not to override.");
```
wuchong: Sets default parallelism of x for all operators (such as aggregate, join, filter)  to run with x parallel instances. This config has a higher priority than parallelism of StreamExecutionEnvironment (actually, this config overrides the parallelism of StreamExecutionEnvironment). A value of -1 indicates that no default parallelism is set, then it will fallback to use the parallelism of StreamExecutionEnvironment.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/plan/nodes/physical/stream/StreamExecExchange.scala
XuPingyong: I think it is better to use ExecutionConfig.PARALLELISM_DEFAULT instead of -1 or env.Parallelism. 
ExecutionConfig.PARALLELISM_DEFAULT means table do not set specific parallelism for  transformations and its parallelism is to be decided by runtime.

flink-table/flink-table-planner-blink/src/test/scala/org/apache/flink/table/api/batch/ExplainTest.scala
XuPingyong: StreamPartitioner in StreamGraph produced by explaining is influenced by parallelism. So env parallelism should be config.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/delegation/PlannerBase.scala
XuPingyong: `mergeParameters()` is not called by explain. So `StreamPlanner` and `BatchPlanner` still need to call this method.
And I think `overrideEnvParallelism` is near `translateToPlan` that is better.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/plan/nodes/physical/stream/StreamExecDataStreamScan.scala
XuPingyong: ok

XuPingyong:
Thanks @wuchong , I have fixed according to your comments.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/delegation/PlannerBase.scala
```diff
@@ -148,6 +149,15 @@ abstract class PlannerBase(
...
+    // Use config parallelism to override env parallelism.
+    val defaultParallelism = getTableConfig.getConfiguration.getInteger(
+      ExecutionConfigOptions.TABLE_EXEC_RESOURCE_DEFAULT_PARALLELISM)
```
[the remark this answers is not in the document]
XuPingyong: Users are not able to touch streamEnv, so this config is need.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/delegation/BatchPlanner.scala
```diff
@@ -61,13 +60,12 @@ class BatchPlanner(
...
     // breakup deadlock
-    val postNodeDag = new DeadlockBreakupProcessor().process(execNodePlan, context)
-    // set parallelism
-    new ParallelismProcessor().process(postNodeDag, context)
+    new DeadlockBreakupProcessor().process(execNodePlan, context)
   }
 
   override protected def translateToPlan(
       execNodes: util.List[ExecNode[_, _]]): util.List[Transformation[_]] = {
+    overrideEnvParallelism()
```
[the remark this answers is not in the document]
XuPingyong: This action only affects `translateToPlan`, and if it is put in `translate`, `explain` still need it.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/delegation/PlannerBase.scala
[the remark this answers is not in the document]
JingsongLi: It sounds like a temporary thing. If we abstract configuration later, we can unify it with the parallelism configuration in ExecutionConfig.
But I agree it is important to user to config it.

JingsongLi:
LGTM

wuchong:
LGTM

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/plan/nodes/physical/stream/StreamExecSink.scala
```diff
@@ -145,12 +145,12 @@ class StreamExecSink[T](
             "implemented and return the sink transformation DataStreamSink. " +
             s"However, ${sink.getClass.getCanonicalName} doesn't implement this method.")
         }
-        if (!UpdatingPlanChecker.isAppendOnly(this) &&
-            dsSink.getTransformation.getParallelism != transformation.getParallelism) {
-          throw new TableException(s"The configured sink parallelism should be equal to the" +
+        if (!UpdatingPlanChecker.isAppendOnly(this) && dsSink.getTransformation.getParallelism != 1
+            && dsSink.getTransformation.getParallelism != transformation.getParallelism) {
```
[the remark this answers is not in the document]
XuPingyong: There are some problems about this check, remove it first.

wuchong:
Merging...

PR #9279, opened 2019-07-30
[FLINK-13423][hive] Unable to find function in hive 1

lirui-apache:

To fix exception handling when getting function from hive-1.2.1.

## Brief change log

  - Hive-1.2.1 did a poor job regarding exception handling of `getFunction` API. So we have to parse the exception to decide whether to propagate or not.
  - Added test case to cover it. The test case runs for both 2.3.4 and 1.2.1, although the issue only exists in 1.2.1.

## Verifying this change

New test case.

  - The serializers: no

## Documentation

### DIFF
--- a/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/client/HiveShimV1.java
+++ b/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/client/HiveShimV1.java
@@ -84,6 +84,7 @@ public class HiveShimV1 implements HiveShim {
 			// hive-1.x doesn't throw NoSuchObjectException if function doesn't exist, instead it throws a MetaException
 			return client.getFunction(dbName, functionName);
 		} catch (MetaException e) {
+			// need to check the cause and message of this MetaException to decide whether it should actually be a NoSuchObjectException
 			if (e.getCause() instanceof NoSuchObjectException) {
 				throw (NoSuchObjectException) e.getCause();
 			}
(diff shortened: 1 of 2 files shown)


lirui-apache:
cc @xuefuz @zjffdu @zjuwangg @wuchong

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/client/HiveShimV1.java
```diff
@@ -87,6 +87,9 @@ public Function getFunction(IMetaStoreClient client, String dbName, String funct
 			if (e.getCause() instanceof NoSuchObjectException) {
 				throw (NoSuchObjectException) e.getCause();
 			}
+			if (e.getMessage().startsWith(NoSuchObjectException.class.getSimpleName())) {
```
xuefuz: Nit: it would be nice if we can document why we need this additional check.

lirui-apache:
@xuefuz PR updated to address your comment

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/client/HiveShimV1.java
```diff
@@ -84,6 +84,7 @@ public Function getFunction(IMetaStoreClient client, String dbName, String funct
 			// hive-1.x doesn't throw NoSuchObjectException if function doesn't exist, instead it throws a MetaException
 			return client.getFunction(dbName, functionName);
 		} catch (MetaException e) {
+			// need to check the cause and message of this MetaException to decide whether it should actually be a NoSuchObjectException
```
xuefuz: Is this specific to HIve 1.2.1? If so, let's be clear, since we didn't have this check previously.

lirui-apache: Yeah it's specific to 1.2.1. Actually there's a comment about that a couple of lines above:
```
// hive-1.x doesn't throw NoSuchObjectException if function doesn't exist, instead it throws a MetaException
```

wuchong:
Hi @xuefuz , do you have other comments?

xuefuz:
LGTM as well. +1

lirui-apache:
I don't see how this could possibly cause python test failures. So suppose that's not related?

wuchong:
The python test failed for several pull requests. And this PR shouldn't affect python test.
I will merge this.

PR #9285, opened 2019-07-31
[FLINK-13433][table-planner-blink]  Do not fetch data from LookupableTableSource if the JoinKey in left side of LookupJoin contains null value.

beyond1920:

For LookupJoin, if joinKey in left side of a LeftOuterJoin/InnerJoin contains null values, there is no need to fetch data from `LookupableTableSource`. 
However, we don't shortcut the fetch function under the case at present, the correctness of results depends on the `TableFunction` implementation of each `LookupableTableSource`.

## Brief change log

  - correct the behavior of `LookupJoin` where lookupKeys contains null.
  - Add more testcases.

## Verifying this change
testcases

  - The serializers: no

## Documentation

### DIFF
--- a/flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/codegen/LookupJoinCodeGenerator.scala
+++ b/flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/codegen/LookupJoinCodeGenerator.scala
@@ -65,7 +65,7 @@ object LookupJoinCodeGenerator {
     : GeneratedFunction[FlatMapFunction[BaseRow, BaseRow]] = {
 
     val ctx = CodeGeneratorContext(config)
-    val (prepareCode, parameters) = prepareParameters(
+    val (prepareCode, parameters, nullInParameters) = prepareParameters(
       ctx,
       typeFactory,
       inputType,
@@ -87,11 +87,17 @@ object LookupJoinCodeGenerator {
         s"$lookupFunctionTerm.setCollector($DEFAULT_COLLECTOR_TERM);"
     }
 
+    // TODO: filter all records when there is any nulls on the join key, because
+    //  "IS NOT DISTINCT FROM" is not supported yet.
     val body =
       s"""
          |$prepareCode
          |$setCollectorCode
-         |$lookupFunctionTerm.eval($parameters);
+         |if ($nullInParameters) {
+         |  return;
+         |} else {
+         |  $lookupFunctionTerm.eval($parameters);
+         | }
       """.stripMargin
 
     FunctionCodeGenerator.generateFunction(
@@ -118,7 +124,7 @@ object LookupJoinCodeGenerator {
     : GeneratedFunction[AsyncFunction[BaseRow, AnyRef]] = {
 
     val ctx = CodeGeneratorContext(config)
-    val (prepareCode, parameters) = prepareParameters(
+    val (prepareCode, parameters, nullInParameters) = prepareParameters(
       ctx,
       typeFactory,
       inputType,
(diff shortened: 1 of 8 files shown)


beyond1920:
cc @wuchong @lincoln-lil

JingsongLi:
I am interested to sql: `SELECT T.id, T.len, T.content, D.name FROM T JOIN userTable for system_time as of T.proctime AS D ON T.id = D.id OR (T.id is null and D.id is null)`.
Does it work?

beyond1920:
@JingsongLi At first glance, I guess the query is not supported. 
However, after run it actually, The query could be executed, but results in error outputs……
Good catch, I will looks into the problem.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/codegen/LookupJoinCodeGenerator.scala
```diff
@@ -133,8 +137,13 @@ object LookupJoinCodeGenerator {
     val body =
       s"""
          |$prepareCode
-         |$DELEGATE delegates = new $DELEGATE($DEFAULT_COLLECTOR_TERM);
-         |$lookupFunctionTerm.eval(delegates.getCompletableFuture(), $parameters);
+         |if ($parametersCheckNullCode) {
+         |  $DEFAULT_COLLECTOR_TERM.complete(new java.util.ArrayList());
```
lincoln-lil: use `Collections.emptyList()` is better

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/plan/rules/physical/common/CommonLookupJoinRule.scala
```diff
@@ -86,6 +91,21 @@ trait CommonLookupJoinRule {
...
+
+    val filterNulls: Array[Boolean] = {
+      val filterNulls = new util.ArrayList[java.lang.Boolean]
+      JoinUtil.createJoinInfo(join.getLeft, join.getRight, join.getCondition, filterNulls)
+      filterNulls.map(_.booleanValue()).toArray
+    }
+
+    if (filterNulls.contains(false)) {
+      throw new TableException(
+        "IS NOT DISTINCT FROM (or an expanded version) is not supported in LookupJoin.")
```
lincoln-lil: Give the detail condition maybe more clearly to users?

flink-table/flink-table-planner-blink/src/test/scala/org/apache/flink/table/planner/plan/batch/sql/join/LookupJoinTest.scala
```diff
@@ -85,6 +85,26 @@ class LookupJoinTest extends TableTestBase {
     )
   }
 
+  @Test
+  def testNotDistinctFromInJoinCondition(): Unit = {
+
+    // does not support join condition contains `IS NOT  DISTINCT`
```
lincoln-lil: `NOT  DISTINCT`  rm one space.

beyond1920:
@wuchong 
> could you also update InMemoryAsyncLookupFunction#eval and InMemoryLookupFunction#eval to throw exception when one of the argument is null？
Hi, JarkWu, InMemoryAsyncLookupFunction and InMemoryLookupFunction lookup records with null value on the lookup key field when argument contains null, which complies with the contract mentioned in https://github.com/apache/flink/pull/9335/files, I think the behavior is also reasonable.

wuchong:
I agree with you @beyond1920 , however, I want to have some test to cover we didn't push any nulls into `LookupFunction`. That's why I want to add such check in `InMemoryLookupFunction`. We can remove the check when we support push `is null` into LookupFunction?

beyond1920:
@wuchong , Ok

wuchong:
Thanks @beyond1920 , looks good to me now.

KurtYoung:
BTW, could you also open a jira for empty key for hbase lookup function? hbase forbidden empty row key.

beyond1920:
@KurtYoung The hbase lookup function is updated in https://github.com/apache/flink/pull/9335.

PR #9287, opened 2019-07-31
[FLINK-13498][kafka] abort transactions in parallel

NicoK:

This makes `FlinkKafkaProducer` abort transactions, e.g. during a first startup, in parallel making use of lingering CPU resources (using at most `kafkaProducersPoolSize` producers at once each, just like during runtime).

Especially during that first startup (and thus also in tests), a lot of producers `(5*poolSize)` are being created at each sink instance to abort potentially existing previous transactions (in most cases, they don't exist).

With this change, Kafka test times (on my PC) change like this:
```
FlinkKafkaProducerMigrationTest        :  36.689s -> 29.438s
FlinkKafkaProducerMigrationOperatorTest:  33.288s -> 26.232s
------
FlinkKafkaProducerITCase               : 289.475s -> 228.469s
KafkaITCase                            : 139.7s   -> 127.595s
FlinkKafkaInternalProducerITCase       :  34.436s ->  35.391s
KafkaProducerAtLeastOnceITCase         : 391.627s -> 391.473s
KafkaProducerExactlyOnceITCase         :  84.366s ->  72.608s
```

## Brief change log

- change `FlinkKafkaProducer`:
  - abort transactions using a parallel stream
  - use a new thread pool with `size = kafkaProducersPoolSize`
  - do not use `#initTransactionalProducer()` but rather only instantiate the parts needed for aborting transactions, i.e. a `FlinkKafkaInternalProducer`
- do the same for `FlinkKafkaProducer011`

## Verifying this change

This change is already covered by existing tests, such as Kafka unit tests and ITCases.

  - The serializers: **no**

## Documentation

### DIFF
--- a/flink-connectors/flink-connector-kafka-0.11/src/main/java/org/apache/flink/streaming/connectors/kafka/FlinkKafkaProducer011.java
+++ b/flink-connectors/flink-connector-kafka-0.11/src/main/java/org/apache/flink/streaming/connectors/kafka/FlinkKafkaProducer011.java
@@ -912,13 +912,20 @@ public class FlinkKafkaProducer011<IN>
 	// ----------------------------------- Utilities --------------------------
 
 	private void abortTransactions(Set<String> transactionalIds) {
-		for (String transactionalId : transactionalIds) {
+		transactionalIds.parallelStream().forEach(transactionalId -> {
+			// don't mess with the original configuration or any other properties of the
+			// original object
+			// -> create an internal kafka producer on our own and do not rely on
+			//    initTransactionalProducer().
+			final Properties myConfig = new Properties();
+			myConfig.putAll(producerConfig);
+			initTransactionalProducerConfig(myConfig, transactionalId);
 			try (FlinkKafkaProducer<byte[], byte[]> kafkaProducer =
-					initTransactionalProducer(transactionalId, false)) {
-				// it suffice to call initTransactions - this will abort any lingering transactions
+					new FlinkKafkaProducer<>(myConfig)) {
+				// it suffices to call initTransactions - this will abort any lingering transactions
 				kafkaProducer.initTransactions();
 			}
-		}
+		});
 	}
 
 	int getTransactionCoordinatorId() {
@@ -952,12 +959,16 @@ public class FlinkKafkaProducer011<IN>
 	}
 
 	private FlinkKafkaProducer<byte[], byte[]> initTransactionalProducer(String transactionalId, boolean registerMetrics) {
-		producerConfig.put("transactional.id", transactionalId);
+		initTransactionalProducerConfig(producerConfig, transactionalId);
 		return initProducer(registerMetrics);
 	}
 
+	private static void initTransactionalProducerConfig(Properties producerConfig, String transactionalId) {
+		producerConfig.put(ProducerConfig.TRANSACTIONAL_ID_CONFIG, transactionalId);
+	}
+
 	private FlinkKafkaProducer<byte[], byte[]> initNonTransactionalProducer(boolean registerMetrics) {
-		producerConfig.remove("transactional.id");
+		producerConfig.remove(ProducerConfig.TRANSACTIONAL_ID_CONFIG);
 		return initProducer(registerMetrics);
 	}
 
--- a/flink-connectors/flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/FlinkKafkaProducer.java
+++ b/flink-connectors/flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/FlinkKafkaProducer.java
@@ -1090,14 +1090,21 @@ public class FlinkKafkaProducer<IN>
 
 	// ----------------------------------- Utilities --------------------------
 
-	private void abortTransactions(Set<String> transactionalIds) {
-		for (String transactionalId : transactionalIds) {
+	private void abortTransactions(final Set<String> transactionalIds) {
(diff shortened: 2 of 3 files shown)


flink-connectors/flink-connector-kafka-0.11/src/main/java/org/apache/flink/streaming/connectors/kafka/FlinkKafkaProducer011.java
```diff
@@ -906,11 +907,33 @@ private void resetAvailableTransactionalIdsPool(Collection<String> transactional
...
-			try (FlinkKafkaProducer<byte[], byte[]> kafkaProducer =
-					initTransactionalProducer(transactionalId, false)) {
-				// it suffice to call initTransactions - this will abort any lingering transactions
-				kafkaProducer.initTransactions();
+		// shortcut for non-exactly-once producers
+		if (transactionalIds.isEmpty()) {
+			return;
+		}
+
+		ForkJoinPool forkJoinPool = null;
```
pnowojski: I think you can change this line to
```
ForkJoinPool forkJoinPool = new ForkJoinPool(kafkaProducersPoolSize);
```
thus avoiding null handling.

flink-connectors/flink-connector-kafka-0.11/src/main/java/org/apache/flink/streaming/connectors/kafka/FlinkKafkaProducer011.java
```diff
@@ -906,11 +907,33 @@ private void resetAvailableTransactionalIdsPool(Collection<String> transactional
...
+		try {
+			// limit the number of connections to the number that is used during runtime
+			forkJoinPool = new ForkJoinPool(kafkaProducersPoolSize);
```
pnowojski: Hmm, when we were discussing offline I understood that this will be global pool with static size. As it is now, we can be creating some substantial number (hundreds) of threads (albeit short lived). I'm not sure about this...

Crazy idea... what about making this pool `static` and with killing/shutting down idling threads?

flink-connectors/flink-connector-kafka-0.11/src/main/java/org/apache/flink/streaming/connectors/kafka/FlinkKafkaProducer011.java
```diff
@@ -906,11 +907,33 @@ private void resetAvailableTransactionalIdsPool(Collection<String> transactional
...
+			forkJoinPool.submit(() ->
+				transactionalIds.parallelStream().forEach(transactionalId -> {
+					// don't mess with the original configuration or any other properties of the
+					// original object
+					// -> create an internal kafka producer on our own and do not rely on
+					//    initTransactionalProducer().
+					final Properties myConfig = new Properties();
+					myConfig.putAll(producerConfig);
+					myConfig.put("transactional.id", transactionalId);
```
pnowojski: can you deduplicate the config setting code with `initTransactionalProducer`? Yes, now it's only one `put("transactional.id", transactionalId);` call, but this can change at some point of time and such code duplication can make this fragile.

flink-connectors/flink-connector-kafka-0.11/src/main/java/org/apache/flink/streaming/connectors/kafka/FlinkKafkaProducer011.java
```diff
@@ -906,11 +907,33 @@ private void resetAvailableTransactionalIdsPool(Collection<String> transactional
...
-				// it suffice to call initTransactions - this will abort any lingering transactions
-				kafkaProducer.initTransactions();
+		// shortcut for non-exactly-once producers
+		if (transactionalIds.isEmpty()) {
+			return;
+		}
+
+		ForkJoinPool forkJoinPool = null;
+		try {
+			// limit the number of connections to the number that is used during runtime
```
pnowojski: One more thought - what if kafka/zookeeper are limiting number of concurrent connections/operations? Technically speaking yes, previously we COULD have the same number of active `KafkaProducer`s during runtime but:
1. **usually**, only 2 of them were being used - here it's always `kafkaProducersPoolSize`. This argument maybe we can ignore, since we can argue that the user's setup should always support `kafkaProducersPoolSize` number of active `KafkaProducer`s.
2. however, previously `KafkaProducers` they were not being used concurrently. If we use them concurrently now, we can briefly flood kafka brokers. Can this be an issue? 

Can you @NicoK maybe clear this out with @becketqin? He has better understanding of how Kafka internally works.

becketqin: I would not concern too much about the load on the Kafka brokers. They are designed to accept thousands of requests per second. And committing / aborting the transactions in Kafka is pretty much just like send an ordinary message. So it is probably OK.

becketqin: I would not concern too much about the load on the Kafka brokers. They are designed to accept thousands of requests per second. And committing / aborting the transactions in Kafka is pretty much just like send an ordinary message. So it is probably OK.

NicoK:
Thanks @pnowojski for the review. Following a discussion with @becketqin, I reverted the parts that used our own `ForkJoinPool` and changed the code to use Java's default/common pool. That greatly simplifies the code but I was initially also worried and hence created that version.

PR #9288, opened 2019-07-31
[FLINK-13421] [runtime] Do not allocate slots in a MultiTaskSlot when it’s releasing children

zhuzhurk:

*ConcurrentModificationException happened in the children release loop in MultiTaskSlot#release.*

*It happened because new slot allocations happened, adding a new child to the MultiTaskSlot, thus modified the children field.*

*New slot allocations happened because a task failover is caused by a child slot's payload releasing. The failover canceled tasks and returned allocated slots to SlotPool. The returned allocated slots were assigned to some tasks that were not canceled yet. The slot assignment satisfies the input location preference constraints of some downstream tasks, and triggered the slot allocation of these tasks.*

*To fix this, this PR changes MultiTaskSlot to be not available for slot allocation when it is in releasingChildren state. The change is in SlotSharingManager#listResolvedRootSlotInfo.(SlotSharingManager#getUnresolvedRootSlot is not necessary for a change as unresolved multi-task slot is simple and does not lead to a slot allocation.) *

## Brief change log

*(for example:)*
  - *in SlotSharingManager#listResolvedRootSlotInfo*

## Verifying this change

  - *Added uint test SlotSharingManagerTest#testResolvedSlotInReleasingIsNotAvailable*

## Documentation

### DIFF
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/jobmaster/slotpool/SlotSharingManager.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/jobmaster/slotpool/SlotSharingManager.java
@@ -51,6 +51,7 @@ import java.util.Objects;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionException;
 import java.util.function.Function;
+import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
 /**
@@ -185,13 +186,17 @@ public class SlotSharingManager {
 			.values()
 			.stream()
 				.flatMap((Map<AllocationID, MultiTaskSlot> map) -> map.values().stream())
-				.filter((MultiTaskSlot multiTaskSlot) -> !multiTaskSlot.contains(groupId))
+				.filter(validMultiTaskSlotAndDoesNotContain(groupId))
 				.map((MultiTaskSlot multiTaskSlot) -> {
 					SlotInfo slotInfo = multiTaskSlot.getSlotContextFuture().join();
 					return new SlotSelectionStrategy.SlotInfoAndResources(
 							slotInfo,
 							slotInfo.getResourceProfile().subtract(multiTaskSlot.getReservedResources()));
 				}).collect(Collectors.toList());
+	}
+
+	private Predicate<MultiTaskSlot> validMultiTaskSlotAndDoesNotContain(@Nullable AbstractID groupId) {
+		return (MultiTaskSlot multiTaskSlot) -> !multiTaskSlot.contains(groupId) && !multiTaskSlot.isReleasing();
 	}
 
 	@Nullable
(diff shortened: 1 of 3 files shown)


tillrohrmann:
Thanks for opening this PR @zhuzhurk. I have to admit that I cannot completely follow how the `ConcurrentModificationException` occurs. Could you maybe show me a stack trace of the problematic behaviour? How does it happen that we assign a new child to a `MultiTaskSlot` while it is being released?

At the moment I don't see how the behaviour you've covered with the test could occur.

Do you have an example which reproduces the reported problem? With such an example I could better understand the problem or one could generate logs on DEBUG log level.

tillrohrmann:
Moreover, it would be helpful to know which revision you used for generating the log snippet which is attached to the JIRA issue.

zhuzhurk:
> Thanks for opening this PR @zhuzhurk. I have to admit that I cannot completely follow how the `ConcurrentModificationException` occurs. Could you maybe show me a stack trace of the problematic behaviour? How does it happen that we assign a new child to a `MultiTaskSlot` while it is being released?
> 
> At the moment I don't see how the behaviour you've covered with the test could occur.
> 
> Do you have an example which reproduces the reported problem? With such an example I could better understand the problem or one could generate logs on DEBUG log level.

Hi, @tillrohrmann, I added a case to repro this issue at https://github.com/zhuzhurk/flink/commit/7ef30934841854ed7711a5a0eabf89ecefb0ec2d.

For the log, unfortunately we ran that round of stability test is running with some local path for FLINK-13241, and cannot find the revision number. But I think it's not a new issue, which may happen since 1.7 or 1.8.

tillrohrmann:
Thanks a lot @zhuzhurk. I'll check out the test case and try to reproduce the problem locally.

tillrohrmann:
Thanks a lot @zhuzhurk. I think with your test case I was able to understand the problem. Here is the description for future reference:

The underlying problem is that a releasing `MultiTaskSlot` might be reused by other slot requests as @zhuzhurk stated. This can happen in the following case: Assume we have the following topology with slot sharing enabled, eager deployment and two allocated slots `s1, s2`.

```
src1 \
src2 - sink1 
src3 /
```

Flink would assign `s1 -> src1` and `s2 -> src2`. Since `sink1` has three predecessors it will wait for `src3` . to get a slot assigned before it can get scheduled.

Now we fail `s1`. This will trigger a global failover which cancels `src2`. The cancellation of `src2` will return `s2` to the `SlotPool` where it is used to fulfill the slot request from `src3`. With the assignment of `s2 -> src3`, `sink1` learns about the locations of all of its inputs and will be scheduled. Since slot sharing is enabled, `s1`, which is currently being released, will be assigned to `sink1`. Due to this assignment we modify the `MultiTaskSlot#children` map which causes the `ConcurrentModificationException`.

tillrohrmann:
I've updated the PR @zhuzhurk. Please take a look and say if you are ok with it.

PR #9297, opened 2019-07-31
[FLINK-13504][table] Fixed shading issues in table modules.

dawidwys:

Moved the common calcite specific dependencies configuration to
flink-table in order to have them unified across flink-table-planner,
flink-table-planner-blink, flink-table-runtime-blink & flink-sql-parser.

Shading configuration is also done in the table parent pom so that all
modules have common relocation policies.

Enabled shading for flink-sql-parser.

## Verifying this change

* For verification of sql-parser module e.g. run the query attached to jira issue.
* Verify contents of flink-table-planner, flink-table-planner-blink, flink-table-runtime-blink, flink-sql-parser, flink-table-uber, flink-table-uber-blink do not contain unwanted classes.

  - The S3 file system connector: (yes / **no /** don't know)

## Documentation

### DIFF
--- a/flink-table/flink-sql-parser/pom.xml
+++ b/flink-table/flink-sql-parser/pom.xml
@@ -25,6 +25,7 @@ under the License.
 		<artifactId>flink-table</artifactId>
 		<groupId>org.apache.flink</groupId>
 		<version>1.10-SNAPSHOT</version>
+		<relativePath>..</relativePath>
 	</parent>
 
 	<artifactId>flink-sql-parser</artifactId>
@@ -35,7 +36,6 @@ under the License.
 	<properties>
 		<!-- override parent pom -->
 		<test.excludedGroups/>
-		<calcite.version>1.20.0</calcite.version>
 	</properties>
 
 	<dependencies>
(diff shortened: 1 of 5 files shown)


dawidwys:
@twalthr @zentol I would appreciate if you could have a look.

zentol:
This _looks_ correct to me, but when I run it locally the shade-plugin fails with an IllegalArgumentException when building the blink uber jar. Currently checking what's going on,

zentol:
For clarification: all non-uber modules appear to be working fine.

zentol:
nvm, my issue is due to using different JDKs. +1 if Travis passes.

dawidwys:
Unfortunately it does not :( There are some convergence problems in flink-json (sic!) Checking them out.

1u0:
I have run a simple check and the issue described in the parent ticket is gone (maven version `3.2.5`):
the `SqlCreateTable` uses proper shaded `SqlIdentifier.names` field.

dawidwys:
Unfortunately this approach does not work.

The problem is that when we relocate `flink-sql-parser` it uses the relocated guava classes. `flink-table-planner*` have direct dependency on calcite which pulls the unshaded versions of the same classes in tests (as tests are executed before shading). Therefore we cannot move the shading configuration to the `flink-table` module.

I will close this PR in favor of a new approach (#9313 )

PR #9301, opened 2019-07-31
[FLINK-13469][state] Ensure resource used by StateMapSnapshot will be released if snapshot fails

banmoy:

Currently, resource used by `StateMapSnapshot` may not be released by `StateTableSnapshot` if snapshot fails. For example, If exception happens in `StateMapSnapshot#writeState`, resources used by current and left `StateMapSnapshot` can not be released. This PR will fix this problem.

## Brief change log
- make sure all resources are released in `StateTableSnapshot#release()`
    - `CopyOnWriteStateTableSnapshot` will check whether the resource is released for each `CopyOnWriteStateMapSnapshot` in `CopyOnWriteStateTableSnapshot#release()`, and release them if not
    -  `NestedMapsStateTableSnapshot` need do nothing in `NestedMapsStateTableSnapshot#release()`.  On the one hand, nested snapshot is synchronous, and doesn't hold all state map snapshots which will be used in async part like copy-on-write snapshot. State map snapshot for a key-group will not be created until it's turn to write state of this key group, so there is no resource  to need to release except for current state map snapshot. On the other hand, nested state map snapshot actually hold no resource. So `NestedMapsStateTableSnapshot` need do nothing for release.
- remove default implementation `AbstractStateTableSnapshot#release()`. The reasons are as follows:
    -  the way to release for the copy-on-write and the nested is very different, so there is no much meaning to provide default implementation
    - force subclass to think about the proper way to release, and make it more safety

## Verifying this change

  - Added test `CopyOnWriteStateTableTest#testReleaseForSuccessfulSnapshot` and `CopyOnWriteStateTableTest#testReleaseForFailedSnapshot` to validates that resources can be released no matter snapshot is successful or failed

  - The serializers: no

## Documentation

### DIFF
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/state/heap/AbstractStateTableSnapshot.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/state/heap/AbstractStateTableSnapshot.java
@@ -88,13 +88,6 @@ abstract class AbstractStateTableSnapshot<K, N, S>
 	 */
 	protected abstract StateMapSnapshot<K, N, S, ? extends StateMap<K, N, S>> getStateMapSnapshotForKeyGroup(int keyGroup);
 
-	/**
-	 * Optional hook to release resources for this snapshot at the end of its lifecycle.
-	 */
-	@Override
-	public void release() {
-	}
-
 	@Nonnull
 	@Override
 	public StateMetaInfoSnapshot getMetaInfoSnapshot() {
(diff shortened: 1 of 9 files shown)


banmoy:
The failed test is `KafkaProducerExactlyOnceITCase.testExactlyOnceCustomOperator` which does not use keyed state, so it seems irrelevant to this PR. And [FLINK-13527](https://issues.apache.org/jira/browse/FLINK-13527) has been created to track this issue. Retrigger the CI again.

banmoy:
@flinkbot run travis

PR #9307, opened 2019-07-31
[FLINK-13512][kinesis][build] Add jaxb-api dependency on Java 11

zentol:
KPL relies on `javax.xml.bind.DatatypeConverter` (contained in jaxb-api), but relies on it being on the classpath by default, which no longer holds on Java 11.
Hence we add an explicit dependency when compiling on Java 11, and also bundle it in the jar.


zentol:
The `JobManagerWatermarkTrackerTest` fails due to asm 6, which doesn't handle Java 11 properly in all cases. This failure will be resolved once we bump asm as part of FLINK-13467.

PR #9312, opened 2019-08-01
[FLINK-13436][end-to-end-tests] Add TPC-H queries as E2E tests

TsReaper:

This PR adds TPC-H queries as end to end tests to flink.

## Brief change log

 - Add TPC-H queries as E2E tests

## Verifying this change

This change added tests and can be verified as follows: run the newly created `test_tpch.sh` test in end to end test module.

  - The serializers: no

## Documentation

  - If yes, how is the feature documented? no

### DIFF
--- a/flink-end-to-end-tests/flink-tpch-test/src/main/java/org/apache/flink/table/tpch/TpchResultComparator.java
+++ b/flink-end-to-end-tests/flink-tpch-test/src/main/java/org/apache/flink/table/tpch/TpchResultComparator.java
@@ -1,0 +1,123 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.table.tpch;
+
+import java.io.BufferedReader;
+import java.io.FileReader;
+import java.io.IOException;
+
+/**
+ * Result comparator for TPC-H test, according to the TPC-H standard specification v2.18.0.
+ */
+public class TpchResultComparator {
+
+	public static void main(String[] args) throws IOException {
+		if (args.length != 2) {
+			System.out.println(
+				"Exactly 2 paths must be provided, the expected result path and the actual result path");
+			System.exit(1);
+		}
+
+		String expectedPath = args[0];
+		String actualPath = args[1];
+
+		try (
+			BufferedReader expectedReader = new BufferedReader(new FileReader(expectedPath));
+			BufferedReader actualReader = new BufferedReader(new FileReader(actualPath))
+		) {
+			int expectedLineNum = 0;
+			int actualLineNum = 0;
+
+			String expectedLine, actualLine;
+			while (
+				(expectedLine = expectedReader.readLine()) != null &&
+					(actualLine = actualReader.readLine()) != null
+			) {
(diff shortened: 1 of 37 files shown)


flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/config/ExecutionConfigOptions.java
```diff
@@ -45,8 +45,8 @@
 	// ------------------------------------------------------------------------
 	public static final ConfigOption<Integer> TABLE_EXEC_SORT_DEFAULT_LIMIT =
 		key("table.exec.sort.default-limit")
-			.defaultValue(200)
-			.withDescription("Default limit when user don't set a limit after order by.");
+			.defaultValue(-1)
```
twalthr: +1 for this change. I was also confused about a default limit which might confuse users.

flink-end-to-end-tests/flink-tpch-test/src/main/java/org/apache/flink/table/tpch/TpchResultComparator.java
```diff
@@ -0,0 +1,123 @@
...
+							double a = Double.valueOf(actual[i]);
+							if (e < 0 && a > 0 || e > 0 && a < 0) {
+								failed = true;
+							} else {
+								if (e < 0) {
+									e = -e;
+									a = -a;
+								}
+								double t = round(a, 2);
+								// defined in TPCH standard specification v2.18.0 section 2.1.3.5
```
twalthr: nit: could you string replace `TPCH` for `TPC-H` in this PR for all comments and the name of the e2e test.

flink-end-to-end-tests/run-nightly-tests.sh
```diff
@@ -152,6 +152,8 @@ run_test "SQL Client end-to-end test for Kafka 0.10" "$END_TO_END_DIR/test-scrip
 run_test "SQL Client end-to-end test for Kafka 0.11" "$END_TO_END_DIR/test-scripts/test_sql_client_kafka011.sh"
 run_test "SQL Client end-to-end test for modern Kafka" "$END_TO_END_DIR/test-scripts/test_sql_client_kafka.sh"
 
+run_test "TPCH end-to-end test (Blink planner)" "$END_TO_END_DIR/test-scripts/test_tpch.sh" "skip_check_exceptions"
```
twalthr: Are there exceptions that need to be skipped?

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/config/ExecutionConfigOptions.java
JingsongLi: +1 for this change. The previous default value is just for stability.

JingsongLi:
> Very nice PR @TsReaper. I just had minor comments. One question that I have is how do we ensure that the result verification is correct. Have you also manually tried to execute some queries in a popular system such as Oracle or Hive and compared it against Flink's results. I would just like to be on the safe side.

The library has its result: https://github.com/airlift/tpch/tree/master/src/main/resources/io/airlift/tpch/queries
Whatever the scale factor, the output is the same.
Now, we just compare with its result. I think its result is come from presto. And I agree with execute some queries in a popular system.

TPCH document also define some sample output for validation, (Like 2.4.1.4 in http://www.tpc.org/tpc_documents_current_versions/pdf/tpc-h_v2.18.0.pdf). But maybe it requires
 1 scale factor at least.

flink-end-to-end-tests/test-scripts/test-data/tpch/query/q1.sql
```diff
@@ -0,0 +1,22 @@
+-- database: presto; groups: tpch; tables: lineitem
```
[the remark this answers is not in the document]
twalthr: What has been modified?

TsReaper:
> Very nice PR @TsReaper. I just had minor comments. One question that I have is how do we ensure that the result verification is correct. Have you also manually tried to execute some queries in a popular system such as Oracle or Hive and compared it against Flink's results. I would just like to be on the safe side.

I've run the queries and the data on mysql and the results are compatible.

TsReaper:
Oops. After modifying the default value of limit some plan validation tests failed. Let me fix it.

KurtYoung:
@twalthr I and @wuchong have discussed the default limit today, and we both agree `-1` will be more appropriate default value.

TsReaper:
Travis passed: https://travis-ci.com/TsReaper/flink/builds/121605193

PR #9313, opened 2019-08-01
 [FLINK-13504][table] Fixed shading issues in table modules. 

dawidwys:

This PR properly includes the flink-sql-parser into flink-planner* modules shaded
jars. It also removes the flink-sql-parser module from flink-table-uber* jars as
it is already included in the flink-planner* modules.

## Brief change log

  - Fixed shading of `flink-sql-parser`
  - Removed unnecessary dependency on guava in `flink-sql-parser`

Verifying this change

  * For verification of sql-parser module e.g. run the query attached to jira issue.
  * Verify contents of flink-table-planner, flink-table-planner-blink, flink-table-runtime-blink, flink-sql-parser, flink-table-uber, flink-table-uber-blink do not contain unwanted classes.

## Documentation

### DIFF
--- a/flink-end-to-end-tests/run-pre-commit-tests.sh
+++ b/flink-end-to-end-tests/run-pre-commit-tests.sh
@@ -60,6 +60,7 @@ run_test "Modern Kafka end-to-end test" "$END_TO_END_DIR/test-scripts/test_strea
 run_test "Kinesis end-to-end test" "$END_TO_END_DIR/test-scripts/test_streaming_kinesis.sh"
 run_test "class loading end-to-end test" "$END_TO_END_DIR/test-scripts/test_streaming_classloader.sh"
 run_test "Distributed cache end-to-end test" "$END_TO_END_DIR/test-scripts/test_streaming_distributed_cache_via_blob.sh"
+run_test "Dependency shading of table module test" "$END_TO_END_DIR/test-scripts/test_table_shaded_dependencies.sh"
 
 printf "\n[PASS] All tests passed\n"
 exit 0
--- a/flink-end-to-end-tests/test-scripts/test_table_shaded_dependencies.sh
+++ b/flink-end-to-end-tests/test-scripts/test_table_shaded_dependencies.sh
@@ -19,26 +19,82 @@
 
 source "$(dirname "$0")"/common.sh
 
-function checkDependencies {
-  JAR=$1
-  CONTENTS_FILE=$TEST_DATA_DIR/contentsInJar.txt
-  jdeps $JAR > $CONTENTS_FILE
-  if [[ `grep -c "(?<!org.apache.flink.calcite.shaded.)com.google" $CONTENTS_FILE` -eq '0' && \
-      `grep -c "(?<!org.apache.flink.calcite.shaded.)com.jayway" $CONTENTS_FILE` -eq '0' && \
-      `grep -c "(?<!org.apache.flink.calcite.shaded.)org.apache.commons.codec" $CONTENTS_FILE` -eq '0' && \
-      `grep -c "(?<!org.apache.flink.table.shaded.)org.joda.time" $CONTENTS_FILE` -eq '0' && \
-      `grep -c "(?<!org.apache.flink.table.shaded.)net.jpountz" $CONTENTS_FILE` -eq '0' && \
-      `grep -c "(?<!org.apache.flink.calcite.shaded.)com.fasterxml" $CONTENTS_FILE` -eq '0' ]]; then
+FLINK_VERSION=`ls "${END_TO_END_DIR}/../flink-table/flink-table-runtime-blink/target" | sed -n "s/.*flink-table-runtime-blink_\(.*\)-tests\.jar/\1/p" | uniq`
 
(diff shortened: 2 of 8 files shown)


1u0:
This patch doesn't work for `flink-table-blink*.jar`, the code in `org.apache.flink.sql.parser.ddl.SqlCreateTable` still references original (unshaded) type of `org.apache.calcite.sql.SqlIdentifier.names`.

Although, for the old planner, the issue is fixed.

dawidwys:
Thx @1u0 for checking. I forgot to commit a change.

1u0:
`flink-table-planner*.jar` also looks fine.

flink-table/flink-table-uber-blink/pom.xml
```diff
@@ -103,7 +103,8 @@ under the License.
 							<artifactSet>
 								<includes combine.children="append">
 									<include>org.apache.flink:flink-table-common</include>
-									<include>org.apache.flink:flink-sql-parser</include>
+									<!--Sql parser is included in planners-->
+									<!--<include>org.apache.flink:flink-sql-parser</include>-->
```
1u0: Remove this commented include?
+ same for `flink-table-uber/pom.xml`.

dawidwys:
@zentol @twalthr Could you have a look at this PR, if you have a second? This just fixes the shading of `flink-sql-parser` without reworking the pom structure. 

> The problem is that when we relocate flink-sql-parser it uses the relocated guava classes. flink-table-planner* have direct dependency on calcite which pulls the unshaded versions of the same classes in tests (as tests are executed before shading). Therefore we cannot move the shading configuration to the flink-table module.

@zentol What do you think about the test added? Do you think it makes sense?

zentol:
What I would do is setup a new pom module in `flink-table` like `flink-table-uber-jars`, containing the shade plugin configuration that we wanted to add to `flink-table`, which then contains all modules we have that create uber-jars.

Then remove all relocations etc. from other modules in flink-table.
As in, the shade-plugin would only be used when creating the uber-jars and not anywhere else. This should isolate tests completely from any shading issue, while still providing us with a way to share shade-plugin configurations.

flink-end-to-end-tests/run-pre-commit-tests.sh
```diff
@@ -60,6 +60,7 @@ run_test "Modern Kafka end-to-end test" "$END_TO_END_DIR/test-scripts/test_strea
 run_test "Kinesis end-to-end test" "$END_TO_END_DIR/test-scripts/test_streaming_kinesis.sh"
 run_test "class loading end-to-end test" "$END_TO_END_DIR/test-scripts/test_streaming_classloader.sh"
 run_test "Distributed cache end-to-end test" "$END_TO_END_DIR/test-scripts/test_streaming_distributed_cache_via_blob.sh"
+run_test "Dependency shading of table module test" "$END_TO_END_DIR/test-scripts/test_table_shaded_dependencies.sh"
```
twalthr: We remove this again while merging, right?

flink-table/flink-table-uber/pom.xml
```diff
@@ -96,8 +95,8 @@ under the License.
 						<configuration>
 							<artifactSet>
 								<includes combine.children="append">
+									<!--Sql parser is included in planners-->
```
twalthr: Does it harm to include it here as well? Because you could also argue that `flink-table-api-java` is in `flink-table-api-scala` etc.

dawidwys: It does. We do not relocate any packages in the `table-uber` jar. The sql-parser is relocated and included in the uber jar of  `table-planner*` modules. This is not the case for the api modules. Planners just declare dependency on those, but they do not bundle them into the jars

flink-end-to-end-tests/run-pre-commit-tests.sh
dawidwys: Do you mean you wouldn't add this test to pre commit tests?

twalthr: I think the pre commit tests should just test the very core functionality. It is also ok, to wait a night for getting noticed about an violation.

flink-end-to-end-tests/test-scripts/test_table_shaded_dependencies.sh
```diff
@@ -0,0 +1,100 @@
...
+      grep -v "org/apache/flink" > $CONTENTS_FILE
+  if [[ `cat $CONTENTS_FILE | wc -l` -eq '0' ]]; then
+      echo "Success: There are no unwanted classes in the ${JAR} jar."
+  else
+      echo "Failure: There are unwanted classes in the ${JAR} jar: `cat $CONTENTS_FILE`"
+      exit 1
+  fi
+}
+
+checkCodeDependencies "${END_TO_END_DIR}/../flink-table/flink-table-planner/target/flink-table-planner_${FLINK_VERSION}.jar"
```
twalthr: Add the API modules as well? In case a user doesn't use the uber jars but adds planner and API separately to the job jar?

PR #9317, opened 2019-08-01
[FLINK-13530][qs][tests] Increase port range size

zentol:
Increase the size of the port range used in `AbstractServerTest#testPortRangeSuccess` by 4000%, which should make it quite likely that we find an open port; should the test fail again for the same reason we can conclude that there is an actual problem in the code.


tillrohrmann:
The test failure seems to be unrelated. Re-opened [FLINK-9900](https://issues.apache.org/jira/browse/FLINK-9900) for this.

PR #9318, opened 2019-08-01
[FLINK-13044][s3][fs] Fix handling of relocated amazon classes

zentol:
To provide credentials to S3 users may configure a credentials provider. For providers from amazon (which are relocated) we allow users to configure the original class name, and relocate it manually in the S3 filesystem factories.

The factories however
- used the wrong shading pattern
- could not correctly detect providers with a "com.amazon*" prefix, as the String constant containing this prefix was also being relocated.

This commit obfuscates the constant to prevent relocations, fixes the shading patterns, and sets up e2e tests to cover this case.

I've triggered a cron job: 
- ~~https://travis-ci.org/apache/flink/builds/566411631~~
- ~~https://travis-ci.org/apache/flink/builds/566798037~~
- ~~https://travis-ci.org/apache/flink/builds/566864128~~
- https://travis-ci.org/apache/flink/builds/566950961

### DIFF
--- a/tools/travis/splits/split_misc.sh
+++ b/tools/travis/splits/split_misc.sh
@@ -76,5 +76,7 @@ run_test "SQL Client end-to-end test for modern Kafka" "$END_TO_END_DIR/test-scr
 
 run_test "Dependency shading of table modules test" "$END_TO_END_DIR/test-scripts/test_table_shaded_dependencies.sh"
 
+run_test "Shaded Hadoop S3A with credentials provider end-to-end test" "$END_TO_END_DIR/test-scripts/test_batch_wordcount.sh hadoop_with_provider"
+
 printf "\n[PASS] All tests passed\n"
 exit 0
--- a/flink-end-to-end-tests/test-scripts/common_s3.sh
+++ b/flink-end-to-end-tests/test-scripts/common_s3.sh
@@ -68,6 +68,12 @@ function s3_setup {
   set_config_key "s3.secret-key" "$IT_CASE_S3_SECRET_KEY"
 }
 
+function s3_setup_with_provider {
+  add_optional_plugin "s3-fs-$1"
+  # reads (AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY)
+  set_config_key "$2" "com.amazonaws.auth.EnvironmentVariableCredentialsProvider"
+}
+
 ###################################
 # List s3 objects by full path prefix.
 #
--- a/flink-end-to-end-tests/test-scripts/test_batch_wordcount.sh
+++ b/flink-end-to-end-tests/test-scripts/test_batch_wordcount.sh
@@ -29,6 +29,11 @@ case $INPUT_TYPE in
         s3_setup hadoop
         INPUT_LOCATION="${S3_TEST_DATA_WORDS_URI}"
     ;;
+    (hadoop_with_provider)
+        source "$(dirname "$0")"/common_s3.sh
+        s3_setup_with_provider hadoop "fs.s3a.aws.credentials.provider"
+        INPUT_LOCATION="${S3_TEST_DATA_WORDS_URI}"
+    ;;
     (presto)
         source "$(dirname "$0")"/common_s3.sh
         s3_setup presto
--- a/flink-filesystems/flink-s3-fs-hadoop/src/main/java/org/apache/flink/fs/s3hadoop/S3FileSystemFactory.java
+++ b/flink-filesystems/flink-s3-fs-hadoop/src/main/java/org/apache/flink/fs/s3hadoop/S3FileSystemFactory.java
@@ -41,11 +41,13 @@ public class S3FileSystemFactory extends AbstractS3FileSystemFactory {
 
 	private static final Logger LOG = LoggerFactory.getLogger(S3FileSystemFactory.class);
 
-	private static final Set<String> PACKAGE_PREFIXES_TO_SHADE = Collections.singleton("com.amazonaws.");
+	// intentionally obfuscated to prevent relocations by the shade-plugin
+	private static final Set<String> PACKAGE_PREFIXES_TO_SHADE = Collections.singleton("com.UNSHADE.".replace("UNSHADE", "amazonaws"));
 
 	private static final Set<String> CONFIG_KEYS_TO_SHADE = Collections.singleton("fs.s3a.aws.credentials.provider");
 
-	private static final String FLINK_SHADING_PREFIX = "org.apache.flink.fs.s3hadoop.shaded.";
+	// keep this in sync with the relocation pattern applied to com.amazon in the shade-plugin configuration
+	private static final String FLINK_SHADING_PREFIX = "org.apache.flink.fs.s3base.shaded.";
 
 	private static final String[] FLINK_CONFIG_PREFIXES = { "s3.", "s3a.", "fs.s3a." };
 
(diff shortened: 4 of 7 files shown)


tools/travis/splits/split_misc.sh
```diff
@@ -72,5 +72,8 @@ run_test "SQL Client end-to-end test for Kafka 0.10" "$END_TO_END_DIR/test-scrip
 run_test "SQL Client end-to-end test for Kafka 0.11" "$END_TO_END_DIR/test-scripts/test_sql_client_kafka011.sh"
 run_test "SQL Client end-to-end test for modern Kafka" "$END_TO_END_DIR/test-scripts/test_sql_client_kafka.sh"
 
+run_test "Shaded Hadoop S3A with credentials provider end-to-end test" "$END_TO_END_DIR/test-scripts/test_batch_wordcount.sh hadoop_with_provider"
+run_test "Shaded Presto S3 with credentials provider end-to-end test" "$END_TO_END_DIR/test-scripts/test_batch_wordcount.sh presto_with_provider"
```
1u0: Afaik, the `run-nightly-tests.sh` is not used in CI runs, but it's mentioned in the `README.md`.
I think it would be nice to add those new tests into the `run-nightly-tests.sh` too, to keep the file in sync with tests splits.

zentol: The `run-nightly-tests.sh` should actually be removed entirely. It's already out-of-sync with the actually run tests, and acts as a trap for committers who aren't that familiar with the Travis setup. Multiple tests are defined in there that are missing from the splits, and this is likely not intentional.

zentol: If anything it should be reworked to call into all splits.

zentol:
According to the CI results there's a shading issue in the presto filesystem:
```
There's a shading issue in the Presto filesystem. When I setup the E2E test to use a credentials provider I run into a NoSuchMethodException: ```Caused by: java.lang.RuntimeException: Error creating an instance of org.apache.flink.fs.s3base.shaded.com.amazonaws.auth.EnvironmentVariableCredentialsProvider for URI s3://[secure]/static/words
	at org.apache.flink.fs.s3presto.shaded.com.facebook.presto.hive.s3.PrestoS3FileSystem.getCustomAWSCredentialsProvider(PrestoS3FileSystem.java:724)
	at org.apache.flink.fs.s3presto.shaded.com.facebook.presto.hive.s3.PrestoS3FileSystem.getAwsCredentialsProvider(PrestoS3FileSystem.java:708)
	at org.apache.flink.fs.s3presto.shaded.com.facebook.presto.hive.s3.PrestoS3FileSystem.createAmazonS3Client(PrestoS3FileSystem.java:632)
	at org.apache.flink.fs.s3presto.shaded.com.facebook.presto.hive.s3.PrestoS3FileSystem.initialize(PrestoS3FileSystem.java:216)
	at org.apache.flink.fs.s3.common.AbstractS3FileSystemFactory.create(AbstractS3FileSystemFactory.java:126)
	... 28 more
Caused by: java.lang.NoSuchMethodException: org.apache.flink.fs.s3base.shaded.com.amazonaws.auth.EnvironmentVariableCredentialsProvider.<init>(java.net.URI, org.apache.flink.fs.shaded.hadoop3.org.apache.hadoop.conf.Configuration)
	at java.lang.Class.getConstructor0(Class.java:3082)
	at java.lang.Class.getConstructor(Class.java:1825)
	at org.apache.flink.fs.s3presto.shaded.com.facebook.presto.hive.s3.PrestoS3FileSystem.getCustomAWSCredentialsProvider(PrestoS3FileSystem.java:720)
	... 32 more``` Branch: https://github.com/apache/flink/tree/s3_test2
```

1u0:
It looks like it's an another issue with Presto fs:
the `com.amazonaws.auth.EnvironmentVariableCredentialsProvider` is not suitable `AWSCredentialsProvider` for `com.facebook.presto.hive.s3.PrestoS3FileSystem`.

From [Prestodb documentation](https://prestodb.github.io/docs/current/connector/hive.html#amazon-s3-configuration):

> You can configure a custom S3 credentials provider by setting the Hadoop configuration property presto.s3.credentials-provider to be the fully qualified class name of a custom AWS credentials provider implementation. This class must implement the `AWSCredentialsProvider` interface and provide a two-argument constructor that takes a `java.net.URI` and a Hadoop `org.apache.hadoop.conf.Configuration` as arguments.

The versions of `EnvironmentVariableCredentialsProvider` we are using have only trivial constructor.

zentol:
@1u0 Do you know by chance any credentials provider that we _could_ use?

A custom implementation cannot work since it wouldn't use the relocated amazon dependencies; if there is no compatible provider in the presto/amazon dependencies then this feature is simply broken.

zentol:
The only ones I could find so far are `org.apache.hadoop.fs.s3a.SimpleAWSCredentialsProvider` and `org.apache.hadoop.fs.s3a.auth.AssumedRoleCredentialProvider`, but they both load hadoop properties with an `fs.s3a` prefix, which we don't mirror from the Flink configuration.

zentol:
Given that this isn't new, and the removal of shading logic in 1.10 would provide simple workarounds, I would be fine with excluding presto from this PR and just fix it for `s3-hadoop`.

"Fixing" this for presto would effectively be adding a new feature.

1u0:
Personally I'm **`+1`** for removing the end2end test for Presto in this PR.

If I understand correctly, there is no need to revert production code changes for Presto, as the (modified) settings has no effect (before change and after).

Also **`+1`** to address Presto "auto-prefix" as a separate effort.

tools/travis/splits/split_misc.sh
```diff
@@ -76,5 +76,7 @@ run_test "SQL Client end-to-end test for modern Kafka" "$END_TO_END_DIR/test-scr
 
 run_test "Dependency shading of table modules test" "$END_TO_END_DIR/test-scripts/test_table_shaded_dependencies.sh"
 
+run_test "Shaded Hadoop S3A with credentials provider end-to-end test" "$END_TO_END_DIR/test-scripts/test_batch_wordcount.sh hadoop_with_provider"
```
1u0: Unrelated to the changes in this PR, but could you also move:
```
run_test "Shaded Hadoop S3A end-to-end test" "$END_TO_END_DIR/test-scripts/test_batch_wordcount.sh hadoop"
run_test "Shaded Presto S3 end-to-end test" "$END_TO_END_DIR/test-scripts/test_batch_wordcount.sh presto"
```
from `run-pre-commit-tests.sh` here, so they are aggregated in one place?

Apparently, the above tests are not actually running during pre-commit tests run:
```
==============================================================================
Running 'Shaded Hadoop S3A end-to-end test'
==============================================================================
...
Did not find AWS environment variables, NOT running the e2e test.
...
[PASS] 'Shaded Hadoop S3A end-to-end test' passed after 0 minutes and 0 seconds! Test exited with exit code 0.

==============================================================================
Running 'Shaded Presto S3 end-to-end test'
==============================================================================
...
Did not find AWS environment variables, NOT running the e2e test.
...
[PASS] 'Shaded Presto S3 end-to-end test' passed after 0 minutes and 0 seconds! Test exited with exit code 0.
```

zentol: The s3 e2e tests rely on environment variables that are only available on actual branches, not pull requests. On a branch (like `master`) these tests will be run on each commit.

PR #9321, opened 2019-08-01
[FLINK-13486][tests] Optimize AsyncDataStreamITCase to alleviate the …

ifndef-SleePy:
…pain of timeout

* Alleviate the timeout problem of `AsyncDataStreamITCase`

## Brief change log

* Make a normal timeout threshold from 1s to 10s
* Avoid time-consuming testing
* Simplify redundant testing codes

## Verifying this change

  - The serializers: no

## Documentation

### DIFF
--- a/flink-streaming-scala/src/test/scala/org/apache/flink/streaming/api/scala/AsyncDataStreamITCase.scala
+++ b/flink-streaming-scala/src/test/scala/org/apache/flink/streaming/api/scala/AsyncDataStreamITCase.scala
@@ -18,13 +18,12 @@
 
 package org.apache.flink.streaming.api.scala
 
-import java.util.concurrent.TimeUnit
+import java.util.concurrent.{CountDownLatch, TimeUnit}
 
 import org.apache.flink.configuration.Configuration
-import org.apache.flink.streaming.api.functions.async.RichAsyncFunction.{RichAsyncFunctionIterationRuntimeContext, RichAsyncFunctionRuntimeContext}
 import org.apache.flink.streaming.api.functions.sink.SinkFunction
 import org.apache.flink.streaming.api.scala.AsyncDataStreamITCase._
-import org.apache.flink.streaming.api.scala.async.{AsyncFunction, ResultFuture, RichAsyncFunction}
+import org.apache.flink.streaming.api.scala.async.{ResultFuture, RichAsyncFunction}
 import org.apache.flink.test.util.AbstractTestBase
 import org.junit.Assert._
 import org.junit.Test
@@ -33,7 +32,6 @@ import scala.collection.mutable
 import scala.concurrent.{ExecutionContext, Future}
 
 object AsyncDataStreamITCase {
-  val timeout = 1000L
   private var testResult: mutable.ArrayBuffer[Int] = _
 }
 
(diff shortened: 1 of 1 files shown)


flink-streaming-scala/src/test/scala/org/apache/flink/streaming/api/scala/AsyncDataStreamITCase.scala
```diff
@@ -33,37 +32,39 @@ import scala.collection.mutable
...
-        source, new MyAsyncFunction(), timeout, TimeUnit.MILLISECONDS)
+        source, new AsyncFunctionWithTimeoutExpired(), fastTimeout, TimeUnit.MILLISECONDS)
     } else {
       AsyncDataStream.unorderedWait(
-        source, new MyAsyncFunction(), timeout, TimeUnit.MILLISECONDS)
+        source, new AsyncFunctionWithTimeoutExpired(), fastTimeout, TimeUnit.MILLISECONDS)
     }
 
-    executeAndValidate(ordered, env, asyncMapped, mutable.ArrayBuffer[Int](2, 6))
+    // the latch only works once, so the second element would multiply 2
```
kl0u: Here you have a race condition. You assume that the order of calls will be:
1) receive 1st element
2) call `timeout()` on the 1st element.
3) receive 2nd element

This may not be the case as you may have the following ordering given that the `timeout()` method is called by another thread:
1) receive 1st element
2) receive 2nd element
3) call `timeout()` on the 1st element.

This may lead to different results and actually if you run the test in a loop, you will see it is still flaky.

What you could do is have two latches, one as it is now, and the other waiting in the `asyncInvoke` when it sees the **second** element. This second latch will be triggered by the `timeout` of the first element. I think that this will guarantee the ordering that you want.

flink-streaming-scala/src/test/scala/org/apache/flink/streaming/api/scala/AsyncDataStreamITCase.scala
```diff
@@ -33,37 +32,39 @@ import scala.collection.mutable
 import scala.concurrent.{ExecutionContext, Future}
 
 object AsyncDataStreamITCase {
-  val timeout = 1000L
+  val timeout = 10000L
```
kl0u: These two variables can become local variables in the private methods (`testAsyncWaitTimeout` and `testAsyncWaitUsingAnonymousFunction`) that execute the tests. This will help the readability of the tests as the reader will not have to scroll up and down to see the values.

flink-streaming-scala/src/test/scala/org/apache/flink/streaming/api/scala/AsyncDataStreamITCase.scala
```diff
@@ -88,31 +89,16 @@ class AsyncDataStreamITCase extends AbstractTestBase {
...
 
   @Test
-  def testOrderedWaitUsingAnonymousFunction(): Unit = {
-    testAsyncWaitUsingAnonymousFunction(true)
-  }
-
-  @Test
-  def testUnorderedWaitUsingAnonymousFunction(): Unit = {
-    testAsyncWaitUsingAnonymousFunction(false)
+  def testOrderedWait(): Unit = {
```
kl0u: Please revert all the renamings to their previous names (`testOrderedWaitUsingAnonymousFunction`, `testUnorderedWaitUsingAnonymousFunction`, and `testAsyncWaitUsingAnonymousFunction`), as refactorings should not be part of this PR (they are not relevant to the attached JIRA issue) and also I think that, although not perfect, the previous names were giving more information about what the tests are trying to do.

ifndef-SleePy:
Oh, it's a horrible PR. I must be drunk at that time :(
Will rewrite the PR later. Thanks for pointing the mistake. @kl0u

kl0u:
No problem @ifndef-SleePy :) Let me know when you have a newer version.

kl0u:
Could you also move the JIRA issue to "in-progress" state?

ifndef-SleePy:
Hi @kl0u , sorry for the late response.
I have addressed the comments and revert the unnecessary changes.

ifndef-SleePy:
Hi @kl0u , thanks for nice reviewing!

PR #9322, opened 2019-08-01
[FLINK-13471][table] Add FlatAggregate support to stream Table API(blink planner)

sunjincheng121:
This pull request ports non-window flatAggregate from Flink planner to Blink planner.
The changes mainly include three parts: blink-planner, blink-runtime and some meta handers. The code of the API part can be reused as the two planner share the same API.

## Brief change log
- Add plan support for table aggregate. [279fea2b05](https://github.com/apache/flink/commit/279fea2b05d5fe6c428a1a315f9b903d3740aa52)
This commit adds RelNodes(Logical, FlinkLogical and StreamExec) and Rules in blink planner. The execution procedures are:
First, the TableAggregate QueryOperation node in API level will be converted to a LogicalTableAggregate relnode in QueryOperationConverter.
Second, the LogicalTableAggregate will be converted to a FlinkLogicalTableAggregate by the rule in logical optimization phase.
Third, the FlinkLogicalTableAggregate will be converted to a StreamExecGroupTableAggregate by the rule in physical optimization phase.

- Add runtime support for table aggregate. [b4c0231e3](https://github.com/apache/flink/commit/b4c0231e30c9d0f37c4eb3d7fa354cbfa6201db3)
This commit adds codegen and runtime process function for table aggregate. The AggCodeGen generates a TableAggsHandleFunction which is used in the GroupTableAggFunction KeyedProcessFunction. A TableAggsHandleFunction contains methods like createAccumulators(), accumulate(), emitValue() etc.
In this part, The implementation of TableAggregate is very similar to Aggregate. The biggest difference is the emit strategy of TableAggregate. For TableAggregate, it emits multi rows and multi columns, while Aggregate emits single row and single column each time.

- Support TableAggregate in some MetadataHandler. [e3d5e780a1](https://github.com/apache/flink/commit/e3d5e780a11ddd27ab6721772925db730d95c75c)
MetadataHandler is used to provide Metadata of a relational expression. For example, provide the unique key information of a relnode. The information can be used during optimization.
This commit adds TableAggregate support in FlinkRelMdColumnInterval, FlinkRelMdFilteredColumnInterval and FlinkRelMdModifiedMonotonicity. The absence of TableAggregate in other MetadataHandlers mainly because:

    - The TableAggregate can not provide uniqueness information which is required in FlinkRelMdColumnUniqueness, FlinkRelMdDistinctRowCount, FlinkRelMdUniqueGroups, FlinkRelMdUniqueKeys.
    - Most of the MetadataHandlers are not been used in streaming jobs，such as FlinkRelMdPercentageOriginalRows，FlinkRelMdPopulationSize, FlinkRelMdRowCount, etc.

## Verifying this change
  - Adds Table Aggregate Tests for  in blink planner.
  - Adds  Meta handler Tests in blink planner.
  - Adds Harness Tests in blink planner

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/codegen/CodeGenUtils.scala
+++ b/flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/codegen/CodeGenUtils.scala
@@ -25,7 +25,7 @@ import org.apache.flink.table.dataformat.util.BinaryRowUtil.BYTE_ARRAY_BASE_OFFS
 import org.apache.flink.table.dataformat.{BinaryStringUtil, Decimal, _}
 import org.apache.flink.table.functions.UserDefinedFunction
 import org.apache.flink.table.runtime.dataview.StateDataViewStore
-import org.apache.flink.table.runtime.generated.{AggsHandleFunction, HashFunction, NamespaceAggsHandleFunction}
+import org.apache.flink.table.runtime.generated.{AggsHandleFunction, HashFunction, NamespaceAggsHandleFunction, TableAggsHandleFunction}
 import org.apache.flink.table.runtime.types.ClassLogicalTypeConverter
 import org.apache.flink.table.runtime.types.ClassLogicalTypeConverter.getInternalClassForType
 import org.apache.flink.table.runtime.types.LogicalTypeDataTypeConverter.fromDataTypeToLogicalType
@@ -90,6 +90,8 @@ object CodeGenUtils {
   val SEGMENT: String = className[MemorySegment]
 
   val AGGS_HANDLER_FUNCTION: String = className[AggsHandleFunction]
+
+  val TABLE_AGGS_HANDLER_FUNCTION: String = className[TableAggsHandleFunction]
 
   val NAMESPACE_AGGS_HANDLER_FUNCTION: String = className[NamespaceAggsHandleFunction[_]]
 
--- a/flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/codegen/agg/AggsHandlerCodeGenerator.scala
+++ b/flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/codegen/agg/AggsHandlerCodeGenerator.scala
@@ -19,6 +19,7 @@ package org.apache.flink.table.planner.codegen.agg
 
 import org.apache.flink.table.api.TableException
 import org.apache.flink.table.dataformat.GenericRow
+import org.apache.flink.table.dataformat.util.BaseRowUtil
 import org.apache.flink.table.expressions._
 import org.apache.flink.table.planner.codegen.CodeGenUtils.{BASE_ROW, _}
 import org.apache.flink.table.planner.codegen.Indenter.toISC
@@ -29,7 +30,8 @@ import org.apache.flink.table.planner.expressions.{PlannerProctimeAttribute, Pla
 import org.apache.flink.table.planner.functions.aggfunctions.DeclarativeAggregateFunction
 import org.apache.flink.table.planner.plan.utils.AggregateInfoList
 import org.apache.flink.table.runtime.dataview.{StateListView, StateMapView}
-import org.apache.flink.table.runtime.generated.{AggsHandleFunction, GeneratedAggsHandleFunction, GeneratedNamespaceAggsHandleFunction, GeneratedTableAggsHandleFunction, NamespaceAggsHandleFunction}
+import org.apache.flink.table.runtime.generated.{AggsHandleFunction, GeneratedAggsHandleFunction, GeneratedNamespaceAggsHandleFunction, GeneratedTableAggsHandleFunction, NamespaceAggsHandleFunction, TableAggsHandleFunction}
+import org.apache.flink.table.runtime.types.PlannerTypeUtils
 import org.apache.flink.table.runtime.types.LogicalTypeDataTypeConverter.fromDataTypeToLogicalType
 import org.apache.flink.table.types.DataType
 import org.apache.flink.table.types.logical.{BooleanType, IntType, LogicalType, RowType}
@@ -420,13 +422,33 @@ class AggsHandlerCodeGenerator(
 
     // gen converter
     val aggExternalType = aggInfoList.getActualAggregateInfos(0).externalResultType
-    val collectorInputName = newName("collectorInput")
-    val converterCode = CodeGenUtils.genToInternal(ctx, aggExternalType, collectorInputName)
+    val recordInputName = newName("recordInput")
+    val converterCode = CodeGenUtils.genToInternal(ctx, aggExternalType, recordInputName)
+
+    def genRecordToBaseRow: String = {
+      val resultType = fromDataTypeToLogicalType(aggExternalType)
+      val resultBaseRowType = PlannerTypeUtils.toRowType(resultType)
+
+      val newCtx = CodeGeneratorContext(ctx.tableConfig)
+      val exprGenerator = new ExprCodeGenerator(newCtx, false).bindInput(resultType)
+      val resultExpr = exprGenerator.generateConverterResultExpression(
+        resultBaseRowType, classOf[GenericRow], "convertResult")
+
+      val resultTypeClass = boxedTypeTermForType(resultType)
+      s"""
+         |${newCtx.reuseMemberCode()}
+         |$resultTypeClass ${exprGenerator.input1Term} = ($resultTypeClass) $converterCode;
+         |${newCtx.reuseLocalVariableCode()}
+         |${newCtx.reuseInputUnboxingCode()}
+         |${resultExpr.code}
+         |return ${resultExpr.resultTerm};
+       """.stripMargin
+    }
 
     val functionName = newName(name)
     val functionCode =
       j"""
-        public final class $functionName implements $TABLE_AGGS_HANDLER_FUNCTION {
+        public final class $functionName implements ${className[TableAggsHandleFunction]} {
 
           ${ctx.reuseMemberCode()}
           private $CONVERT_COLLECTOR_TYPE_TERM $MEMBER_COLLECTOR_TERM;
@@ -481,8 +503,7 @@ class AggsHandlerCodeGenerator(
             $COLLECTOR<$BASE_ROW> $COLLECTOR_TERM, $BASE_ROW key, boolean isRetract)
             throws Exception {
 
-            $MEMBER_COLLECTOR_TERM.$COLLECTOR_TERM = $COLLECTOR_TERM;
-            $MEMBER_COLLECTOR_TERM.reset(key, isRetract);
+            $MEMBER_COLLECTOR_TERM.reset(key, isRetract, $COLLECTOR_TERM);
             $emitValueCode
           }
 
@@ -497,12 +518,10 @@ class AggsHandlerCodeGenerator(
           }
 
           private class $CONVERT_COLLECTOR_TYPE_TERM implements $COLLECTOR {
-            public $COLLECTOR<$BASE_ROW> $COLLECTOR_TERM;
+            private $COLLECTOR<$BASE_ROW> $COLLECTOR_TERM;
             private $BASE_ROW key;
             private $JOINED_ROW result;
             private boolean isRetract = false;
-            private static final byte ACCUMULATE_MSG = 0;
-            private static final byte RETRACT_MSG = 1;
             ${ctx.reuseMemberCode()}
 
             public $CONVERT_COLLECTOR_TYPE_TERM(java.lang.Object[] references) throws Exception {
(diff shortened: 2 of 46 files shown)


sunjincheng121:
@flinkbot attention @JingsongLi @wuchong @aljoscha @hequn8128 I appreciate if you can have look at this PR. :)

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/plan/utils/RelExplainUtil.scala
```diff
@@ -329,11 +329,16 @@ object RelExplainUtil {
...
 
+    val isTableAggregate =
+      AggregateUtil.isTableAggregate(aggInfoList.getActualAggregateCalls.toList)
     val outputFieldNames = if (isLocal) {
       grouping.map(inFieldNames(_)) ++ localAggOutputFieldNames(aggOffset, aggInfos, outFieldNames)
     } else if (isIncremental) {
       val accFieldNames = inputRowType.getFieldNames.toList.toArray
       grouping.map(inFieldNames(_)) ++ localAggOutputFieldNames(aggOffset, aggInfos, accFieldNames)
+    } else if (isTableAggregate) {
+      outFieldNames.toList.subList(0, grouping.length).toArray ++
```
JingsongLi: just use `outFieldNames.slice(0, grouping.length)` or dropRight?

flink-table/flink-table-planner/src/main/scala/org/apache/flink/table/expressions/aggregations.scala
```diff
@@ -358,10 +358,10 @@ case class VarSamp(child: PlannerExpression) extends Aggregation {
 }
 
 /**
-  * Expression for calling a user-defined aggregate function.
```
JingsongLi: the codes of flink-planner can be a separate commit?

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/runtime/operators/aggregate/GroupTableAggFunction.java
```diff
@@ -0,0 +1,172 @@
...
+import org.apache.flink.util.Collector;
+
+import static org.apache.flink.table.dataformat.util.BaseRowUtil.isAccumulateMsg;
+
+/**
+ * Aggregate Function used for the groupby (without window) table aggregate.
+ */
+public class GroupTableAggFunction extends KeyedProcessFunctionWithCleanupState<BaseRow, BaseRow, BaseRow> {
+
+	private static final long serialVersionUID = -4767158666069797704L;
```
JingsongLi: serialVersionUID be 1L?

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/runtime/operators/aggregate/GroupTableAggFunction.java
```diff
@@ -0,0 +1,172 @@
...
+		// instantiate function
+		function = genAggsHandler.newInstance(getRuntimeContext().getUserCodeClassLoader());
+		function.open(new PerKeyStateDataViewStore(getRuntimeContext()));
+
+		BaseRowTypeInfo accTypeInfo = new BaseRowTypeInfo(accTypes);
+		ValueStateDescriptor<BaseRow> accDesc = new ValueStateDescriptor<>("accState", accTypeInfo);
+		accState = getRuntimeContext().getState(accDesc);
+
+		initCleanupTimeState("GroupTableAggregateCleanupTime");
+
```
JingsongLi: remove empty line?

flink-table/flink-table-planner-blink/src/test/scala/org/apache/flink/table/planner/runtime/stream/table/TableAggregateITCase.scala
```diff
@@ -0,0 +1,196 @@
...
+import org.junit.Assert.assertEquals
+import org.junit.runner.RunWith
+import org.junit.runners.Parameterized
+import org.junit.{Before, Test}
+
+/**
+  * Tests of groupby (without window) table aggregations
+  */
+@RunWith(classOf[Parameterized])
+class TableAggregateITCase(mode: StateBackendMode) extends StreamingWithStateTestBase(mode) {
```
JingsongLi: Can you add a it case that agg after flatagg?

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/codegen/CodeGenUtils.scala
```diff
@@ -91,6 +91,8 @@ object CodeGenUtils {
 
   val AGGS_HANDLER_FUNCTION: String = className[AggsHandleFunction]
 
+  val TABLE_AGGS_HANDLER_FUNCTION: String = className[TableAggsHandleFunction]
```
JingsongLi: No need to add here, here just for repeated use.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/codegen/agg/AggsHandlerCodeGenerator.scala
```diff
@@ -700,14 +841,21 @@ class AggsHandlerCodeGenerator(
     """.stripMargin
   }
 
+  private def genEmitValue(): String = {
```
JingsongLi: this method can be internal in `generateTableAggsHandler`?

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/codegen/agg/ImperativeAggCodeGen.scala
```diff
@@ -503,5 +505,22 @@ class ImperativeAggCodeGen(
...
+        .getOrElse(
+          throw new CodeGenException(
+            s"No matching emitValue method found for " +
+              s"table aggregate ${function.getClass.getCanonicalName}'.")
+        )
+    }
+  }
+
+  def emitValue: String = {
+    val accTerm = if (isAccTypeInternal) accInternalTerm else accExternalTerm
```
JingsongLi: Maybe you can add a BaseRow(isAccTypeInternal) it case?

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/codegen/agg/ImperativeAggCodeGen.scala
```diff
@@ -503,5 +505,22 @@ class ImperativeAggCodeGen(
...
+    s"""
+       |$functionTerm.emitValue($accTerm, $MEMBER_COLLECTOR_TERM);
```
JingsongLi: Just `s""` is OK?

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/codegen/agg/AggsHandlerCodeGenerator.scala
```diff
@@ -397,6 +398,146 @@ class AggsHandlerCodeGenerator(
...
+          public $BASE_ROW createAccumulators() throws Exception {
+            $createAccumulatorsCode
+          }
+
+          @Override
+          public void emitValue(
+            $COLLECTOR<$BASE_ROW> $COLLECTOR_TERM, $BASE_ROW key, boolean isRetract)
+            throws Exception {
+
+            $MEMBER_COLLECTOR_TERM.$COLLECTOR_TERM = $COLLECTOR_TERM;
```
JingsongLi: can we add `COLLECTOR_TERM` to reset method?

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/codegen/agg/AggsHandlerCodeGenerator.scala
```diff
@@ -397,6 +398,146 @@ class AggsHandlerCodeGenerator(
...
+          public void close() throws Exception {
+            ${ctx.reuseCloseCode()}
+          }
+
+          private class $CONVERT_COLLECTOR_TYPE_TERM implements $COLLECTOR {
+            public $COLLECTOR<$BASE_ROW> $COLLECTOR_TERM;
+            private $BASE_ROW key;
+            private $JOINED_ROW result;
+            private boolean isRetract = false;
+            private static final byte ACCUMULATE_MSG = 0;
```
JingsongLi: use `BaseRowUtil.ACCUMULATE_MSG` instead of hard code?

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/plan/nodes/physical/stream/StreamExecGroupTableAggregate.scala
```diff
@@ -0,0 +1,188 @@
...
+import org.apache.calcite.rel.core.AggregateCall
+import org.apache.calcite.rel.{RelNode, RelWriter, SingleRel}
+import org.apache.flink.table.runtime.operators.aggregate.GroupTableAggFunction
+
+import scala.collection.JavaConversions._
+
+/**
+  * Stream physical RelNode for unbounded group table aggregate.
+  */
+class StreamExecGroupTableAggregate(
```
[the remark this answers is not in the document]
sunjincheng121: I have also thinking about the reuse problem. The biggest obstacle for reuse these two classes is `StreamExecGroupAggregate` extends
  `StreamExecGroupAggregateBase` now. The `StreamExecGroupAggregateBase` has the same semantic with `Aggregate`. For StreamExecGroupTableAggregate,
  it is a `TableAggregate`. As `Aggregate` and `TableAggregate` contains different semantics, StreamExecGroupTableAggregate could not extends `StreamExecGroupAggregateBase` or other base class which extends `StreamExecGroupAggregateBase`.

How over, we can also use some interface or Util class to reuse the two classes, but from my point of view, the benefit is small, not only because the class contains code less than 200 lines, but also because add interface or Util here seems making code more complicated.

What do you think?

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/codegen/agg/AggsHandlerCodeGenerator.scala
sunjincheng121: This method would also be used for the window flatAggregate, i.e., would also be used in generateWindowTableAggsHandler() later.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/plan/metadata/FlinkRelMdModifiedMonotonicity.scala
```diff
@@ -279,6 +292,27 @@ class FlinkRelMdModifiedMonotonicity private extends MetadataHandler[ModifiedMon
...
+      input: RelNode,
+      grouping: Array[Int],
+      rowSize: Int,
+      mq: RelMetadataQuery): RelModifiedMonotonicity = {
+
+    val fmq = FlinkRelMetadataQuery.reuseOrCreate(mq)
+    val inputMonotonicity = fmq.getRelModifiedMonotonicity(input)
+
+    // if group by an update field or group by a field mono is null, just return null
+    if (grouping.exists(e =>
```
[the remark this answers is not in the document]
JingsongLi: Ignore me, there is no method to judge grouping only.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/plan/nodes/physical/stream/StreamExecGroupTableAggregate.scala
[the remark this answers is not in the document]
JingsongLi: got it, agree

sunjincheng121:
Thanks for the review, most of comments are makes sense to me. only two of them I want share you my thoughts. Please let me know what do you think?

sunjincheng121:
Thanks for your confirm @JingsongLi ! 
I have update the PR accordingly. I appreciate if you take an other look :)
Best, Jincheng

sunjincheng121:
Thanks for your confirm @JingsongLi !
I have update the PR accordingly. I appreciate if you could take an other look :)
Best, Jincheng

aljoscha:
I now also reviewed `[FLINK-13471][table] Add runtime support for non-window flatAggregate on blink planner.`. It looks good from what I can tell, but I don't have much experience with that code.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/plan/nodes/calcite/TableAggregate.scala
```diff
@@ -0,0 +1,108 @@
...
+    })
+
+    // agg fields
+    val aggCall = aggCalls.get(0)
+    if (aggCall.`type`.isStruct) {
+      // only a structured type contains a field list.
+      aggCall.`type`.getFieldList.foreach(builder.add)
+    } else {
+      // A non-structured type does not have a field list, so get field name through
+      // TableEnvImpl.getFieldNames.
```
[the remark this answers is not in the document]
sunjincheng121: Good catch！`TableEnvImpl ` -> `FieldInfoUtils` .

sunjincheng121:
Thanks for the review @aljoscha 
I have update the PR accordingly.

I think you have a good points on share the test both for blink planner and flink planner. 
We need to have a complete test sharing solution, because there are many different places between blink planner and flink planner, including the logic of validation and function, and we have to consider whether to maintain 2 planners for a long time, and then consider sharing the test. So I recommend that the two Planner tests remain independent. 
What do you think?

Best, Jincheng

aljoscha:
Yes, I agree. For now we keep the two separate tests.

sunjincheng121:
Thanks all of you review @aljoscha @JingsongLi 
I'll merging this...
@flinkbot approve all

sunjincheng121:
Rebase code.

PR #9323, opened 2019-08-01
[FLINK-13535][kafka] do not abort transactions twice during KafkaProducer startup

NicoK:

During startup of a transactional Kafka producer from previous state, we recover in two steps:
1) in `TwoPhaseCommitSinkFunction`, we commit pending commit-transactions and abort pending transactions and then call into `finishRecoveringContext()`
2) in `FlinkKafkaProducer#finishRecoveringContext()` we iterate over all recovered transaction IDs and abort them.

This may lead to some transactions being worked on twice and there is quite some overhead from creating a `KafkaProducer` for each of these transactions.

## Brief change log

- provide `finishRecoveringContext()` with a collection of all transactions that `TwoPhaseCommitSinkFunction` already covered
- adapt `FlinkKafkaProducer` and `FlinkKafkaProducer011` to ignores transactional IDs from that set

## Verifying this change

This change is already covered by existing tests, such as `FlinkKafkaProducerITCase` and `KafkaProducerExactlyOnceITCase`.

  - The serializers: **no**

## Documentation

### DIFF
--- a/flink-connectors/flink-connector-kafka-0.11/src/main/java/org/apache/flink/streaming/connectors/kafka/FlinkKafkaProducer011.java
+++ b/flink-connectors/flink-connector-kafka-0.11/src/main/java/org/apache/flink/streaming/connectors/kafka/FlinkKafkaProducer011.java
@@ -882,20 +882,26 @@ public class FlinkKafkaProducer011<IN>
 	}
 
 	@Override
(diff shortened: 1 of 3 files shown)


becketqin:
@flinkbot approve all

becketqin:
Merged to master.

PR #9325, opened 2019-08-01
[FLINK-13542][metrics] Update datadog reporter to send metrics if the…

mans2singh:
… metrics are available

* Update datadog metrics reporter to only post metrics if they are available.  

## Brief change log

* Check series size before sending metrics to Datadog and if there are no series, just log a message.

## Verifying this change

  - The serializers: no

## Documentation

### DIFF
--- a/flink-metrics/flink-metrics-datadog/src/main/java/org/apache/flink/metrics/datadog/DatadogHttpReporter.java
+++ b/flink-metrics/flink-metrics-datadog/src/main/java/org/apache/flink/metrics/datadog/DatadogHttpReporter.java
@@ -155,7 +155,13 @@ public class DatadogHttpReporter implements MetricReporter, Scheduled {
 		}
 
 		try {
-			client.send(request);
+			int seriesSize = request.getSeries().getSeries().size();
+			if (seriesSize > 0) {
+				client.send(request);
+			}
+			else  {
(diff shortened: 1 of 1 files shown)


zentol:
This looks a bit like introducing unnecessary complexity; the time-frame in which no metrics exist is _incredibly_ small.

mans2singh:
Hi @zentol 

I have an application that registers only specific metrics which are based on data and not know up front when the application starts (eg: an exception with its type).  I saw that, if no exception for thrown and no metrics was registered for a long time, the datadog reporter was still periodically reporting empty series to the server.  So, there might be other scenarios where there are no metrics registered and we can check to see if there are metrics before posting the data datadog.

Also, do you think this check belongs to the client and not the reporter ?

Let me know your thoughts.

Thanks

zentol:
> There might be other scenarios where there are no metrics registered for some time

Once a cluster is started this case should never occur, as the cluster is also registering metrics. The only time where it should be possible for no metric to be registered is a brief moment during cluster initialization. This is independent of whether your application is registering metrics or not.

mans2singh:
@zentol - You are right.  I am filtering all default metrics and as I mentioned only registering application metrics which may not be registered if those senarios (eg: an exception being thrown) do not occur.

mans2singh:
Closing request.

PR #9328, opened 2019-08-02
[FLINK-13521][sql-client] Allow setting configurations in SQL CLI

TsReaper:

Currently, we provide a set of configurations in blink planner to support optimization or enable some advanced feature. However, all the configurations can't be set in SQL CLI.

This PR allows setting table environment configurations in SQL CLI by using the following methods:

* Modify config files:
```yaml
configuration:
  table.exec.spill-compression.enabled: true
  table.exec.spill-compression.block-size: 128kb
  table.optimizer.join-reorder-enabled: true
```
* Or use `set` command:
```
SET table.exec.spill-compression.enable=true;
SET table.exec.spill-compression.block-size=128kb;
SET table.optimizer.join-reorder-enabled=true;
```

Note that different from other options in SQL CLI, the `SET` command about table environment configurations does not start with `configuration.` prefix but directly uses the option name.

## Brief change log

 - Allow setting configurations in SQL CLI

## Verifying this change

This change added tests and can be verified as follows: run the newly added tests `LocalExecutorITCase::testTableEnvConfigurations` and `ExecutionContextTest::testConfigurations`.

  - The serializers: no

## Documentation

### DIFF
--- a/flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/config/Environment.java
+++ b/flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/config/Environment.java
@@ -20,6 +20,7 @@ package org.apache.flink.table.client.config;
 
 import org.apache.flink.table.client.SqlClientException;
 import org.apache.flink.table.client.config.entries.CatalogEntry;
+import org.apache.flink.table.client.config.entries.ConfigurationEntry;
 import org.apache.flink.table.client.config.entries.DeploymentEntry;
 import org.apache.flink.table.client.config.entries.ExecutionEntry;
 import org.apache.flink.table.client.config.entries.FunctionEntry;
@@ -50,6 +51,9 @@ public class Environment {
 
 	public static final String DEPLOYMENT_ENTRY = "deployment";
 
+	public static final String[] CONFIGURATION_ENTRIES = new String[]{
+		"table.exec", "table.optimizer"};
+
 	private Map<String, CatalogEntry> catalogs;
 
 	private Map<String, TableEntry> tables;
@@ -60,12 +64,15 @@ public class Environment {
 
 	private DeploymentEntry deployment;
 
+	private ConfigurationEntry configuration;
+
 	public Environment() {
 		this.catalogs = Collections.emptyMap();
 		this.tables = Collections.emptyMap();
 		this.functions = Collections.emptyMap();
 		this.execution = ExecutionEntry.DEFAULT_INSTANCE;
 		this.deployment = DeploymentEntry.DEFAULT_INSTANCE;
+		this.configuration = ConfigurationEntry.DEFAULT_INSTANCE;
 	}
 
 	public Map<String, CatalogEntry> getCatalogs() {
@@ -136,6 +143,14 @@ public class Environment {
 		return deployment;
 	}
 
+	public void setConfiguration(Map<String, Object> config) {
+		this.configuration = ConfigurationEntry.create(config);
+	}
+
+	public ConfigurationEntry getConfiguration() {
+		return configuration;
+	}
+
 	@Override
 	public String toString() {
 		final StringBuilder sb = new StringBuilder();
@@ -158,6 +173,8 @@ public class Environment {
 		execution.asTopLevelMap().forEach((k, v) -> sb.append(k).append(": ").append(v).append('\n'));
 		sb.append("=================== Deployment ===================\n");
 		deployment.asTopLevelMap().forEach((k, v) -> sb.append(k).append(": ").append(v).append('\n'));
+		sb.append("================== Configuration =================\n");
+		configuration.asMap().forEach((k, v) -> sb.append(k).append(": ").append(v).append('\n'));
 		return sb.toString();
 	}
 
@@ -212,6 +229,9 @@ public class Environment {
 		// merge deployment properties
 		mergedEnv.deployment = DeploymentEntry.merge(env1.getDeployment(), env2.getDeployment());
 
+		// merge execution properties
+		mergedEnv.configuration = ConfigurationEntry.merge(env1.getConfiguration(), env2.getConfiguration());
+
 		return mergedEnv;
 	}
 
@@ -239,6 +259,9 @@ public class Environment {
 
 		// enrich deployment properties
 		enrichedEnv.deployment = DeploymentEntry.enrich(env.deployment, properties);
+
+		// enrich configuration properties
+		enrichedEnv.configuration = ConfigurationEntry.enrich(env.configuration, properties);
 
 		return enrichedEnv;
 	}
--- a/flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/config/Environment.java
+++ b/flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/config/Environment.java
@@ -49,10 +49,9 @@ public class Environment {
 
 	public static final String EXECUTION_ENTRY = "execution";
 
-	public static final String DEPLOYMENT_ENTRY = "deployment";
+	public static final String CONFIGURATION_ENTRY = "table";
 
-	public static final String[] CONFIGURATION_ENTRIES = new String[]{
-		"table.exec", "table.optimizer"};
+	public static final String DEPLOYMENT_ENTRY = "deployment";
 
 	private Map<String, CatalogEntry> catalogs;
 
@@ -62,17 +61,17 @@ public class Environment {
 
 	private ExecutionEntry execution;
 
-	private DeploymentEntry deployment;
-
 	private ConfigurationEntry configuration;
+
+	private DeploymentEntry deployment;
 
 	public Environment() {
 		this.catalogs = Collections.emptyMap();
 		this.tables = Collections.emptyMap();
 		this.functions = Collections.emptyMap();
 		this.execution = ExecutionEntry.DEFAULT_INSTANCE;
-		this.deployment = DeploymentEntry.DEFAULT_INSTANCE;
 		this.configuration = ConfigurationEntry.DEFAULT_INSTANCE;
+		this.deployment = DeploymentEntry.DEFAULT_INSTANCE;
 	}
 
 	public Map<String, CatalogEntry> getCatalogs() {
@@ -135,20 +134,20 @@ public class Environment {
 		return execution;
 	}
 
-	public void setDeployment(Map<String, Object> config) {
-		this.deployment = DeploymentEntry.create(config);
-	}
-
-	public DeploymentEntry getDeployment() {
-		return deployment;
-	}
-
 	public void setConfiguration(Map<String, Object> config) {
 		this.configuration = ConfigurationEntry.create(config);
 	}
 
 	public ConfigurationEntry getConfiguration() {
 		return configuration;
+	}
+
+	public void setDeployment(Map<String, Object> config) {
+		this.deployment = DeploymentEntry.create(config);
+	}
+
+	public DeploymentEntry getDeployment() {
+		return deployment;
 	}
 
 	@Override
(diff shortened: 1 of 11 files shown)


flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/config/entries/ConfigurationEntry.java
```diff
@@ -0,0 +1,190 @@
...
+	protected ConfigurationEntry(DescriptorProperties properties) {
+		super(properties);
+	}
+
+	@Override
+	protected void validate(DescriptorProperties properties) {
+
+		// Source options
+		properties.validateDuration(
+			ExecutionConfigOptions.TABLE_EXEC_SOURCE_IDLE_TIMEOUT.key(),
```
wuchong: Could we simply verify it is a valid key, and skip to validate type and value? I think it is checked by the planner.

If we only check the keys, we can avoid sync configs between this file and `ExecutionConfigOptions` and `OptimizerConfigOptions`. We can use Java reflection to extract options.

flink-table/flink-sql-client/src/test/resources/test-sql-client-configurations.yaml
```diff
@@ -0,0 +1,154 @@
...
+configuration:
+  table:
+    exec:
+      sort:
+        default-limit: "$SORT_DEFAULT_LIMIT"
+      spill-compression:
+        enabled: true
+        block-size: 128kb
+    optimizer:
+      join-reorder-enabled: true
```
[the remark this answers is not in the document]
TsReaper: ok

flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/config/entries/ConfigurationEntry.java
TsReaper: ok

TsReaper:
Travis: https://travis-ci.com/TsReaper/flink/builds/121713778

flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/config/Environment.java
```diff
@@ -212,6 +226,9 @@ public static Environment merge(Environment env1, Environment env2) {
 		// merge deployment properties
 		mergedEnv.deployment = DeploymentEntry.merge(env1.getDeployment(), env2.getDeployment());
 
+		// merge execution properties
```
twalthr: update comment

flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/config/entries/ConfigurationEntry.java
```diff
@@ -0,0 +1,87 @@
...
+import org.apache.flink.table.client.config.ConfigUtil;
+import org.apache.flink.table.descriptors.DescriptorProperties;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.apache.flink.table.client.config.Environment.CONFIGURATION_ENTRIES;
+
+/**
+ * Configuration of a table environment.
```
twalthr: Replace the entire comment by `Configuration for configuring {@link TableConfig}.` We don't need reference to explicit option classes.

twalthr:
@TsReaper please avoid force pushes in PR. Especially if other people try to review your code at the moment.

flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/config/Environment.java
```diff
@@ -50,6 +51,9 @@
 
 	public static final String DEPLOYMENT_ENTRY = "deployment";
 
+	public static final String[] CONFIGURATION_ENTRIES = new String[]{
```
twalthr: I think a prefix of `table` is enough. If we add more options we don't need to change this location again.

flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/config/entries/ConfigurationEntry.java
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	// --------------------------------------------------------------------------------------------
+
+	public static ConfigurationEntry create(Map<String, Object> config) {
+		return new ConfigurationEntry(ConfigUtil.normalizeYaml(config));
+	}
+
+	/**
+	 * Merges two execution entries. The properties of the first execution entry might be
```
twalthr: update comment

flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/config/entries/ConfigurationEntry.java
```diff
@@ -0,0 +1,87 @@
...
+	public static final ConfigurationEntry DEFAULT_INSTANCE =
+		new ConfigurationEntry(new DescriptorProperties(true));
+
+	private ConfigurationEntry(DescriptorProperties properties) {
+		super(properties);
+	}
+
+	@Override
+	protected void validate(DescriptorProperties properties) {
+		// Nothing to validate as the planner will check the options
```
twalthr: Why not retrieving the options from the config and do early validation. We have the capabilities to do that and do the environment file validation early. The `TableConfig` can have a static method `getAvailableOptions()`.

flink-table/flink-sql-client/src/test/java/org/apache/flink/table/client/gateway/local/LocalExecutorITCase.java
```diff
@@ -505,6 +506,45 @@ public void testUseCatalogAndUseDatabase() throws Exception {
 		}
 	}
 
+	@Test(timeout = 30_000L)
+	public void testTableEnvConfigurations() throws Exception {
```
twalthr: The IT cases in this file are meant for testing runtime behavior. We don't need a full IT case just for a configuration entry. Please remove this test.

flink-table/flink-sql-client/src/test/resources/test-sql-client-configurations.yaml
```diff
@@ -0,0 +1,149 @@
...
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+# limitations under the License.
+################################################################################
+
+#==============================================================================
+# TEST ENVIRONMENT FILE
+# General purpose default environment file.
```
twalthr: Update comment.

flink-table/flink-sql-client/src/test/resources/test-sql-client-configurations.yaml
```diff
@@ -0,0 +1,149 @@
...
+#==============================================================================
+
+# this file has variables that can be filled with content by replacing $VAR_XXX
+
+tables:
```
twalthr: Most of the content of this file is just code duplication. This file should contain `configuration` only nothing more.

flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/config/entries/ConfigurationEntry.java
TsReaper: I had a discussion with @wuchong on this problem (see @wuchong 's 2nd comment in the last review).

In the 1st edition of this commit, I verified everything from the `ExecutionConfigOptions` and `OptimizerConfigOptions` and called the corresponding verification method according to their types. But as @wuchong mentioned, this will force syncing of these files with `ConfigurationEntry`. As the config options will be later verified by planner, we can simply validate them as strings here.

I decided to use Java reflection to fetch all options in these two option classes and validate them in my 2nd edition of the commit, but I soon discover that validating them as strings actually does nothing, so I leave the validating method to be empty and let planner verify the options later.

I'll discuss with @wuchong again to find a more proper solution.

wuchong: Yes. The thing I want to avoid is to do validation in several places. It will bring extra efforts when we adding new configurations.

twalthr: I'm also against syncing files. If the abstraction is right, this is not necessary. Ideally, it should be possible to add a new configuration without touching the SQL Client later. But I wasn't aware that the config options don't provide a validation method yet. So you are right, it does nothing at the moment. Ignore my comment.

flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/config/Environment.java
TsReaper: got-it

flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/config/entries/ConfigurationEntry.java
TsReaper: got-it

flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/config/Environment.java
TsReaper: got-it

flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/config/entries/ConfigurationEntry.java
TsReaper: got-it

flink-table/flink-sql-client/src/test/java/org/apache/flink/table/client/gateway/local/LocalExecutorITCase.java
TsReaper: got-it

flink-table/flink-sql-client/src/test/resources/test-sql-client-configurations.yaml
TsReaper: got-it

flink-table/flink-sql-client/src/test/resources/test-sql-client-configurations.yaml
TsReaper: got-it

TsReaper:
> @TsReaper please avoid force pushes in PR. Especially if other people try to review your code at the moment.

Thanks for the detailed review. 

I'm sorry that force pushes bother the reviewers and I just want to keep the commit log clean. sorry Please inform be before merging this PR so that I can clean up the commit log. Thanks.

flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/config/Environment.java
```diff
@@ -48,6 +49,8 @@
 
 	public static final String EXECUTION_ENTRY = "execution";
 
+	public static final String CONFIGURATION_ENTRY = "table";
```
docete: What about set CONFIGURATION_ENTRY to 'configuration' (Just like execution or deployment segments)
And user use "SET configuration.table.exec.x.y.z=a" to set "table.exec.x.y.z".

docete: And the ConfigurationEntry should provide an 'asTopLevelMap' interface for accessing entities with 'configuration' prefix

wuchong: Hi @docete ,  we would like to avoid add a `configuration` prefix when SET config options, so that users can have a similar experience with `TableConfig`. The `CONFIGURATION_ENTRY` here is used to match configuration key prefix (all the existing configuration key are started with `table.exec` or `table.optimizer`)

twalthr: I agree with Jark. We should not have another level of nesting when using the `SET` command.

twalthr:
Thanks for the update @TsReaper. I will do a final pass over the changes and merge this by the end of the day, if there are no objections.

PR #9331, opened 2019-08-02
[FLINK-13523][table-planner-blink] Verify and correct arithmetic function's semantic for Blink planner

docete:

Verify and correct arithmetic function's semantic for Blink planner

## Brief change log

- Refactor arithmetic divide to keep it compatible with old planner
- Refactor avg agg to keep it compatible with old planner
- Remove non-standard bitwise/div scalar function from Blink planner

## Verifying this change

## Documentation


wuchong:
@JingsongLi , could you help to review this PR too ?

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/plan/utils/AggFunctionFactory.scala
```diff
@@ -37,11 +37,11 @@ import org.apache.flink.table.runtime.types.TypeInfoLogicalTypeConverter
 import org.apache.flink.table.runtime.typeutils.DecimalTypeInfo
 import org.apache.flink.table.types.logical.LogicalTypeRoot._
 import org.apache.flink.table.types.logical._
-
 import org.apache.calcite.rel.`type`.RelDataType
 import org.apache.calcite.rel.core.AggregateCall
 import org.apache.calcite.sql.fun._
 import org.apache.calcite.sql.{SqlAggFunction, SqlKind, SqlRankFunction}
+import org.apache.flink.table.types.DataType
```
wuchong: Rearrange the package order.

docete:
> Hi @docete , there are some cases failed in blink planner. Please have a look.
> 
> Here are some other suggestions:
> 
> 1. `org.apache.flink.table.planner.functions.aggfunctions.IntegralAvgAggFunction` is never used now, we can remove it.
OK, I will remove it.

> 2. The `AvgAggFunction` under test package `org.apache.flink.table.planner.utils` can be removed too.
This class is used by org.apache.flink.table.planner.runtime.batch.sql.agg.WindowAggregateITCase

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/plan/utils/AggFunctionFactory.scala
docete: OK

JingsongLi:
Maybe this PR will affect TPC-H result too, @docete can you verify it manually?

docete:
> Maybe this PR will affect TPC-H result too, @docete can you verify it manually?

OK, I will verify it manually.

docete:
Hi @JingsongLi I verified tpch e2e test locally
`[PASS] 'flink-end-to-end-tests/test-scripts/test_tpch.sh' passed after 2 minutes and 17 seconds! Test exited with exit code 0.`

docete:
Passed pre-commit test on my travis: https://travis-ci.com/docete/flink/builds/122209856

wuchong:
The TPC-H e2e tests are not affected because all the `avg` or `/` are on double fields.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/plan/rules/logical/SplitAggregateRule.scala
```diff
@@ -303,7 +304,19 @@ class SplitAggregateRule extends RelOptRule(
...
+          // TODO
+          val equals = relBuilder.call(
+            FlinkSqlOperatorTable.EQUALS,
+            countInputRef,
+            relBuilder.getRexBuilder.makeBigintLiteral(JBigDecimal.valueOf(0)))
+          val falseT = relBuilder.call(FlinkSqlOperatorTable.DIVIDE, sumInputRef, countInputRef)
+          val trueT = relBuilder.cast(
+            relBuilder.getRexBuilder.constantNull(), aggCall.`type`.getSqlTypeName)
+          relBuilder.call(
+            FlinkSqlOperatorTable.IF,
```
wuchong: Flink planner supports `cast(null as bigint) / cast( 0 as bigint)` and get `null` result. If we supports this, can we remove this?

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/codegen/agg/DeclarativeAggCodeGen.scala
```diff
@@ -204,8 +204,13 @@ class DeclarativeAggCodeGen(
   }
 
   def getValue(generator: ExprCodeGenerator): GeneratedExpression = {
-    val resolvedGetValueExpression = function.getValueExpression
+    val expr = function.getValueExpression
       .accept(ResolveReference())
+    val resolvedGetValueExpression = ApiExpressionUtils.unresolvedCall(
```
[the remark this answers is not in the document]
docete: For DeclarativeAggregateFunction functions, there is no such guarantee that getResultType and getValueExpression(after type inference) have the same type. If they are different type, a RuntimeException occurs.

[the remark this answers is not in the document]
docete: Will add comments for readability

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/plan/rules/logical/SplitAggregateRule.scala
```diff
@@ -303,7 +304,19 @@ class SplitAggregateRule extends RelOptRule(
             aggGroupCount + index + avgAggCount + 1,
             finalAggregate.getRowType)
           avgAggCount += 1
-          relBuilder.call(FlinkSqlOperatorTable.DIVIDE, sumInputRef, countInputRef)
+          // TODO
```
JingsongLi: This modify is really confused... Maybe you should add some comments here. And modify comment of this class.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/plan/rules/logical/SplitAggregateRule.scala
docete: We use SUM0 in this rule， which many run into zero / zero and throw / zero exception.

JingsongLi: This should be a regular case. Should it return null when avg don't have input count?

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/codegen/agg/DeclarativeAggCodeGen.scala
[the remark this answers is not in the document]
JingsongLi: I don't understand. Which runtime exception?

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/plan/rules/logical/SplitAggregateRule.scala
JingsongLi: Div 0 should return null instead of exception?

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/codegen/agg/DeclarativeAggCodeGen.scala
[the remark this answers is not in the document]
JingsongLi: In which case their type is different, why not change it to the same?

JingsongLi:
@docete Thanks update, left some comments. Especially for 1941bc6003ebe1e4b5fea8cabcee57d3f740cf58 , I think this commit need more discussion.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/plan/rules/logical/SplitAggregateRule.scala
wuchong: I agree with @JingsongLi , we should never throw exception in blink planner. `null or zero / zero` should always return null, in that way, we can avoid to introduce this.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/codegen/agg/DeclarativeAggCodeGen.scala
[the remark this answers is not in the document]
wuchong: IIUC, for example, `AVG(int)` will be expressed `sum(int)/bigint_count` in blink planner which will get `bigint` result. There are many other cases similar to this.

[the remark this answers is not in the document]
JingsongLi: Without this PR, are there no cases? Maybe we can do these cast in functions?
I think getResultType and getValueExpression(after type inference) have different type is weird.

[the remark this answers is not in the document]
docete: Yes, but there is no guarantee about this. So I think add a cast there is more safe.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/plan/rules/logical/SplitAggregateRule.scala
docete: In SQL 2013 Part 2 Section 6.27:
The dyadic arithmetic operators <plus sign>, <minus sign>, <asterisk>, and <solidus> (+, –, *, and /,
respectively) specify addition, subtraction, multiplication, and division, respectively. **If the value of a
divisor is zero, then an exception condition is raised: data exception — division by zero.**

And I think it's THE SplitAggretateRule owns the responsibility of return null(not throw exception).

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/codegen/agg/DeclarativeAggCodeGen.scala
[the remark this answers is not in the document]
docete: Will try to fix this in AvgAggFunction

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/plan/rules/logical/SplitAggregateRule.scala
```diff
@@ -304,19 +302,22 @@ class SplitAggregateRule extends RelOptRule(
...
+          // @see Glossary#SQL2011 SQL:2011 Part 2 Section 6.27
           val equals = relBuilder.call(
             FlinkSqlOperatorTable.EQUALS,
             countInputRef,
             relBuilder.getRexBuilder.makeBigintLiteral(JBigDecimal.valueOf(0)))
-          val falseT = relBuilder.call(FlinkSqlOperatorTable.DIVIDE, sumInputRef, countInputRef)
-          val trueT = relBuilder.cast(
+          val ifTrue = relBuilder.cast(
             relBuilder.getRexBuilder.constantNull(), aggCall.`type`.getSqlTypeName)
+          val ifFalse = relBuilder.call(FlinkSqlOperatorTable.DIVIDE, sumInputRef, countInputRef)
```
wuchong: Do we need a cast here?

docete: Not Need! The following code(line 329) would cast if type is not match.

`    relBuilder.convert(originalAggregate.getRowType, false)
`

wuchong: Thanks.  Please rebase master and resolve the conflicts. Others looks good to me now.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/plan/rules/logical/SplitAggregateRule.scala
```diff
@@ -303,7 +302,22 @@ class SplitAggregateRule extends RelOptRule(
             aggGroupCount + index + avgAggCount + 1,
             finalAggregate.getRowType)
           avgAggCount += 1
-          relBuilder.call(FlinkSqlOperatorTable.DIVIDE, sumInputRef, countInputRef)
+          // Make a guarantee that the final aggregation returns NULL if underlying count is ZERO.
```
JingsongLi: Need modify comments of `SplitAggregateRule` too?

docete:
rebase and resolve conflicts

wuchong:
Merged with

4d560df50d08f22b68e8b8c9a0e2086f45b5f4b4
fbd3225e9d63a10eb6e55fca136d2acfcc777250
421f0a559a3038d2f9f56ba2cbab8e6d1832812a

PR #9332, opened 2019-08-02
[FLINK-13544][connectors] Set parallelism of table sink operator to input transformation parallelism

wuchong:

Currently, there are a lot of TableSink connectors uses `dataStream.addSink()` without `.setParallelism()` explicitly. This will use default parallelism of the environment. However, the parallelism of input transformation might not be `env.parallelism`, for example, global aggregation has 1 parallelism. In this case, it will lead to data reorder, and result in incorrect result.

## Brief change log

Checks all the implementation of `StreamTableSink#emitDataStream/consumeDataStream`, and set parallelism as input parallelism. 

## Verifying this change

This is covered by existing tests.

## Documentation


wuchong:
Hi @twalthr , I appreciate if you can have a look.

wuchong:
Hi @twalthr , the planner can't set parallelism for the sink because we don't know how many transformations are applied in `consumeDataStream(DataStream<T> dataStream)`. For example, `CsvTableSink` applies `CsvFormatter` -> `TextOutputFormat`, we can't touch both transformations.

In the future, when we propose a new TableSource with functions, I think we can do this by planner. 

I think this change is safe and makes sink chained as much as possible to protect correctness.

PR #9333, opened 2019-08-02
[FLINK-9782][e2e] Harden bucketing sink e2e test.

kl0u:

Hardens the bucketing sink e2e test. The cause of the instability is described in the JIRA and a travis build where the updated test runs ~45 times can be found here: https://travis-ci.org/kl0u/flink/builds/566401415

## Brief change log

Updated the bucketing sink e2e test.

## Verifying this change

This change is already covered by the existing `BucketingSink` end-to-end test.

## Documentation

### DIFF
--- a/flink-end-to-end-tests/flink-bucketing-sink-test/src/main/java/org/apache/flink/streaming/tests/BucketingSinkTestProgram.java
+++ b/flink-end-to-end-tests/flink-bucketing-sink-test/src/main/java/org/apache/flink/streaming/tests/BucketingSinkTestProgram.java
@@ -62,7 +62,7 @@ public class BucketingSinkTestProgram {
 
 		StreamExecutionEnvironment sEnv = StreamExecutionEnvironment.getExecutionEnvironment();
 		sEnv.setRestartStrategy(RestartStrategies.fixedDelayRestart(
-				3,
+				10,
 				Time.of(10, TimeUnit.SECONDS)
 			));
 		sEnv.enableCheckpointing(4000);
@@ -71,7 +71,11 @@ public class BucketingSinkTestProgram {
 
 		// define bucketing sink to emit the result
 		BucketingSink<Tuple4<Integer, Long, Integer, String>> sink = new BucketingSink<Tuple4<Integer, Long, Integer, String>>(outputPath)
-			.setBucketer(new KeyBucketer());
+				.setBucketer(new KeyBucketer())
+				.setBatchSize(Long.MAX_VALUE)
+				.setBatchRolloverInterval(Long.MAX_VALUE)
+				.setInactiveBucketCheckInterval(Long.MAX_VALUE)
+				.setInactiveBucketThreshold(Long.MAX_VALUE);
 
 		// generate data, shuffle, perform stateful operation, sink
 		sEnv.addSource(new Generator(10, idlenessMs, 60))
@@ -143,6 +147,7 @@ public class BucketingSinkTestProgram {
 		private final int durationMs;
 
 		private long ms = 0;
+		private volatile boolean canceled = false;
 
 		public Generator(int numKeys, int idlenessMs, int durationSeconds) {
 			this.numKeys = numKeys;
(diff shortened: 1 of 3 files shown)


flink-end-to-end-tests/test-scripts/common.sh
```diff
@@ -553,9 +553,9 @@ function kill_all {
 }
 
 function kill_random_taskmanager {
```
twalthr: Does this change also other tests? If yes, I would move it to a separate commit.

flink-end-to-end-tests/test-scripts/test_streaming_bucketing.sh
```diff
@@ -20,72 +20,136 @@
...
 
-echo "Starting TM"
-
-# start task manager again
+echo "Restarting 1 TM"
 $FLINK_DIR/bin/taskmanager.sh start
+wait_for_number_of_running_tms 4
 
-echo "Killing 2 TMs"
+sleep 10
```
twalthr: why is this additional sleep required?

flink-end-to-end-tests/flink-bucketing-sink-test/src/main/java/org/apache/flink/streaming/tests/BucketingSinkTestProgram.java
```diff
@@ -71,7 +71,11 @@ public static void main(String[] args) throws Exception {
 
 		// define bucketing sink to emit the result
 		BucketingSink<Tuple4<Integer, Long, Integer, String>> sink = new BucketingSink<Tuple4<Integer, Long, Integer, String>>(outputPath)
-			.setBucketer(new KeyBucketer());
+				.setBucketer(new KeyBucketer())
+				.setBatchSize(Long.MAX_VALUE)
```
twalthr: What it the reason for these changes? Should the cause of the problem be already solved by having an infinite source now?

flink-end-to-end-tests/flink-bucketing-sink-test/src/main/java/org/apache/flink/streaming/tests/BucketingSinkTestProgram.java
```diff
@@ -143,6 +147,7 @@ public void open(Configuration parameters) {
 		private final int durationMs;
 
 		private long ms = 0;
+		private volatile boolean canceled = false;
```
twalthr: Is the volatile really necessary? I thought every task is single threaded?

aljoscha: There are actually asynchronous calls that are executed from different threads.

flink-end-to-end-tests/test-scripts/common.sh
```diff
@@ -294,7 +294,7 @@ function query_number_of_running_tms {
 
 function wait_for_number_of_running_tms {
   local TM_NUM_TO_WAIT=${1}
-  local TIMEOUT_COUNTER=10
+  local TIMEOUT_COUNTER=30
```
aljoscha: Maybe revert.

flink-end-to-end-tests/test-scripts/test_streaming_bucketing.sh
```diff
@@ -20,72 +20,136 @@
...
+
+function bucketing_cleanup() {
+  stop_cluster
+  $FLINK_DIR/bin/taskmanager.sh stop-all
+}
+
+# Fix the necessary configuration parameters.
+
+set_conf_ssl
+set_config_key "heartbeat.timeout" "70000"
```
aljoscha: Double check why this value is as it is now.

flink-end-to-end-tests/test-scripts/test_streaming_bucketing.sh
aljoscha: It's to allow the job to make some progress after killing the TaskManager(s)

flink-end-to-end-tests/flink-bucketing-sink-test/src/main/java/org/apache/flink/streaming/tests/BucketingSinkTestProgram.java
kl0u: Actually these are the real instability fix. These make sure that there are no `pending` files between checkpoints which may be invalidated in case of a failure. These files may cause instabilities, given that we _cannot_ safely clean up invalidated `pending` files.

kl0u:
@aljoscha @twalthr I integrated / replied to your comments. Please have a look and let me know what you think.

aljoscha:
Ok, I think this is good to merge now 👌

kl0u:
Thanks a lot for the reviews. I will merge it.

PR #9334, opened 2019-08-02
  [FLINK-10257][FLINK-13463][table] Support VARCHAR type generalization consistently

twalthr:

The Blink planner generalizes multiple `CHAR` literals into `VARCHAR`. It seems likely that we will adopt this behavior in the future to prevent unwanted side-effects for users. This PR removes the exception in the SQL Client and updates the existing type generalization classes.

## Brief change log

See commit messages.

## Verifying this change

Manually tested.

  - The serializers: no

## Documentation

### DIFF
--- a/flink-table/flink-table-common/src/test/java/org/apache/flink/table/types/LogicalTypeDuplicatorTest.java
+++ b/flink-table/flink-table-common/src/test/java/org/apache/flink/table/types/LogicalTypeDuplicatorTest.java
@@ -1,0 +1,161 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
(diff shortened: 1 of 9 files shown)


flink-table/flink-table-common/src/test/java/org/apache/flink/table/types/LogicalTypeDuplicatorTest.java
```diff
@@ -0,0 +1,161 @@
...
+import static org.hamcrest.CoreMatchers.equalTo;
+import static org.junit.Assert.assertThat;
+
+/**
+ * Tests for {@link LogicalTypeDuplicator}.
+ */
+@RunWith(Parameterized.class)
+public class LogicalTypeDuplicatorTest {
+
+	private static final LogicalTypeDuplicator DUPLICATOR = new LogicalTypeDuplicator();
```
dawidwys: Wouldn't it be better to test with at least one method overriden? Right now it tests only the duplicated logic from `LogicalType#copy`?

twalthr: That's on purpose. The `IntReplacer` tests this behavior.

twalthr:
Thanks @dawidwys. Merging...

PR #9335, opened 2019-08-02
[FLINK-13503][API] Add contract in `LookupableTableSource` to specify the behavior when lookupKeys contains null.

beyond1920:

Add contract in `LookupableTableSource` to specify the behavior when lookupKeys contains null.
And update existed connector to comply with this contract.
## Brief change log

  - Add contract in `LookupableTableSource` to specify the behavior when lookupKeys contains null.
  - Update HBase lookupFunction
  - Update JDBC lookupFunction

## Verifying this change

existed IT.

  - The serializers: no

## Documentation

### DIFF
--- a/flink-connectors/flink-hbase/src/main/java/org/apache/flink/addons/hbase/HBaseLookupFunction.java
+++ b/flink-connectors/flink-hbase/src/main/java/org/apache/flink/addons/hbase/HBaseLookupFunction.java
@@ -20,6 +20,7 @@ package org.apache.flink.addons.hbase;
 
 import org.apache.flink.addons.hbase.util.HBaseConfigurationUtil;
 import org.apache.flink.addons.hbase.util.HBaseReadWriteHelper;
+import org.apache.flink.addons.hbase.util.HBaseTypeUtils;
 import org.apache.flink.annotation.VisibleForTesting;
 import org.apache.flink.api.common.typeinfo.TypeInformation;
 import org.apache.flink.table.functions.FunctionContext;
@@ -34,7 +35,9 @@ import org.apache.hadoop.hbase.TableName;
 import org.apache.hadoop.hbase.TableNotFoundException;
 import org.apache.hadoop.hbase.client.Connection;
 import org.apache.hadoop.hbase.client.ConnectionFactory;
+import org.apache.hadoop.hbase.client.Get;
 import org.apache.hadoop.hbase.client.HTable;
+import org.apache.hadoop.hbase.client.Mutation;
 import org.apache.hadoop.hbase.client.Result;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -72,7 +75,15 @@ public class HBaseLookupFunction extends TableFunction<Row> {
 	 */
 	public void eval(Object rowKey) throws IOException {
 		// fetch result
-		Result result = table.get(readHelper.createGet(rowKey));
+		byte[] row = readHelper.serialize(rowKey);
+		Get get;
+		try {
+			get = readHelper.createGet(row);
+		} catch (IllegalArgumentException e) {
+			LOG.warn("Ignore illegal rowKey : {}!", rowKey);
+			return;
+		}
+		Result result = table.get(get);
 		if (!result.isEmpty()) {
 			// parse and collect
 			collect(readHelper.parseToRow(result, rowKey));
--- a/flink-connectors/flink-hbase/src/main/java/org/apache/flink/addons/hbase/util/HBaseReadWriteHelper.java
+++ b/flink-connectors/flink-hbase/src/main/java/org/apache/flink/addons/hbase/util/HBaseReadWriteHelper.java
@@ -87,16 +87,26 @@ public class HBaseReadWriteHelper {
 	}
 
 	/**
+	 * Serializes a rowkey object into byte array.
+	 * @param rowKey rowkey object to serialize
+	 *
+	 * @return serialize bytes.
+	 */
+	public byte[] serialize(Object rowKey) {
+		byte[] key = HBaseTypeUtils.serializeFromObject(
+				rowKey,
+				rowKeyType,
+				charset);
+		return key;
+	}
+
+	/**
 	 * Returns an instance of Get that retrieves the matches records from the HBase table.
 	 *
 	 * @return The appropriate instance of Get for this use case.
 	 */
-	public Get createGet(Object rowKey) {
-		byte[] rowkey = HBaseTypeUtils.serializeFromObject(
-			rowKey,
-			rowKeyType,
-			charset);
-		Get get = new Get(rowkey);
+	public Get createGet(byte[] row) {
+		Get get = new Get(row);
 		for (int f = 0; f < families.length; f++) {
 			byte[] family = families[f];
 			for (byte[] qualifier : qualifiers[f]) {
--- a/flink-connectors/flink-jdbc/src/main/java/org/apache/flink/api/java/io/jdbc/JDBCLookupFunction.java
+++ b/flink-connectors/flink-jdbc/src/main/java/org/apache/flink/api/java/io/jdbc/JDBCLookupFunction.java
@@ -47,9 +47,19 @@ import static org.apache.flink.util.Preconditions.checkNotNull;
 
 /**
  * A {@link TableFunction} to query fields from JDBC by keys.
- * The query template like:
+ * If look up keys don't contain any null value, using the nonNullableQuery template like:
  * <PRE>
  * SELECT c, d, e, f from T where a = ? and b = ?
+ * </PRE>
+ *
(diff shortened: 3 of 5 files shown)


beyond1920:
cc @JingsongLi @wuchong @lincoln-lil

lincoln-lil:
If I understand this issue correctly, should it just convert the `null` key into `xx is null` clause for `JDBCLookupFunction`  instead of `is not distinct from` clause or other variant?

beyond1920:
The issue aims to specify the behavior when the lookupKeys contains null value. It has no relationship with `is not distinct from` clause.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/sources/LookupableTableSource.java
```diff
@@ -34,12 +34,29 @@
 
 	/**
 	 * Gets the {@link TableFunction} which supports lookup one key at a time.
+	 *
 	 * @param lookupKeys the chosen field names as lookup keys, it is in the defined order
+	 *
+	 * <p>IMPORTANT:
+	 * If the returned {@link TableFunction} receives a lookup request with null value in lookup keys, expect
```
lincoln-lil: Give more detailed comments here maybe more clearly for understanding, 
1. lookupKeys are nullable
2. should carefully deal with `null` value lookup in each concrete store engine, e.g., in MySQL this can a `is null` query
...

beyond1920: OK, I'll try to explain more clearly in the comment

flink-connectors/flink-jdbc/src/main/java/org/apache/flink/api/java/io/jdbc/JDBCLookupFunction.java
```diff
@@ -101,7 +105,7 @@ public JDBCLookupFunction(
 		this.maxRetryTimes = lookupOptions.getMaxRetryTimes();
 		this.keySqlTypes = Arrays.stream(keyTypes).mapToInt(JDBCTypeUtil::typeInformationToSqlType).toArray();
 		this.outputSqlTypes = Arrays.stream(fieldTypes).mapToInt(JDBCTypeUtil::typeInformationToSqlType).toArray();
-		this.query = options.getDialect().getSelectFromStatement(
+		this.query = options.getDialect().getSelectNotDistinctFromStatement(
```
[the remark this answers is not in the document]
beyond1920: Good question, I would add two prepareStatement with two query template, if arguments does not contain any null value, it using original query template; else use the new one. Ok?

flink-connectors/flink-hbase/src/main/java/org/apache/flink/addons/hbase/util/HBaseReadWriteHelper.java
```diff
@@ -86,17 +86,27 @@ public HBaseReadWriteHelper(HBaseTableSchema hbaseTableSchema) {
...
 	 *
 	 * @return The appropriate instance of Get for this use case.
 	 */
-	public Get createGet(Object rowKey) {
-		byte[] rowkey = HBaseTypeUtils.serializeFromObject(
-			rowKey,
-			rowKeyType,
-			charset);
-		Get get = new Get(rowkey);
+	public Get createGet(byte[] row) {
```
wuchong: `row` -> `rowkey`

flink-connectors/flink-jdbc/src/main/java/org/apache/flink/api/java/io/jdbc/dialect/JDBCDialect.java
```diff
@@ -115,6 +115,20 @@ default String getDeleteStatement(String tableName, String[] conditionFields) {
 		return "DELETE FROM " + quoteIdentifier(tableName) + " WHERE " + conditionClause;
 	}
 
+	/**
+	 * Get select fields statement by `is not distinct from` condition fields. Default use SELECT.
+	 */
+	default String getSelectNotDistinctFromStatement(String tableName, String[] selectFields, String[] conditionFields) {
```
wuchong: I would re-design the method names a bit. 

- Add a new method `getSelectStatement(String tableName, String[] selectFields)` which doesn't contain `WHERE` clause.
- Rename `getSelectFromStatement` to `getFilterStatement(String tableName, String[] selectFields, String[] conditionFields)` and forces `conditionFields` shouldn't be empty.
- Rename `getSelectNotDistinctFromStatement` to `getNullSafeFilterStatement(...)`, because the default implementation is not `is not distinct from` but is a null safe comparison different from `getFilterStatement`. We should also outline this in the method javadoc.

flink-connectors/flink-hbase/src/main/java/org/apache/flink/addons/hbase/HBaseLookupFunction.java
```diff
@@ -34,7 +35,9 @@
 import org.apache.hadoop.hbase.TableNotFoundException;
 import org.apache.hadoop.hbase.client.Connection;
 import org.apache.hadoop.hbase.client.ConnectionFactory;
+import org.apache.hadoop.hbase.client.Get;
 import org.apache.hadoop.hbase.client.HTable;
+import org.apache.hadoop.hbase.client.Mutation;
```
wuchong: useless import

flink-connectors/flink-hbase/src/main/java/org/apache/flink/addons/hbase/HBaseLookupFunction.java
```diff
@@ -20,6 +20,7 @@
 
 import org.apache.flink.addons.hbase.util.HBaseConfigurationUtil;
 import org.apache.flink.addons.hbase.util.HBaseReadWriteHelper;
+import org.apache.flink.addons.hbase.util.HBaseTypeUtils;
```
wuchong: useless import

flink-connectors/flink-jdbc/src/main/java/org/apache/flink/api/java/io/jdbc/JDBCLookupFunction.java
```diff
@@ -141,7 +169,12 @@ public void eval(Object... keys) {
 			try {
 				statement.clearParameters();
 				for (int i = 0; i < keys.length; i++) {
-					JDBCUtils.setField(statement, keySqlTypes[i], keys[i], i);
+					if (containsNull) {
+						JDBCUtils.setField(statement, keySqlTypes[i], keys[i], 2 * i);
+						JDBCUtils.setField(statement, keySqlTypes[i], keys[i], 2 * i + 1);
```
wuchong: We can keep this special logic for now. But please open a JIRA to improve this. We can introduced a custom `NamedPreparedStatement` to pass each field only once. see https://www.javaworld.com/article/2077706/named-parameters-for-preparedstatement.html

flink-connectors/flink-jdbc/src/main/java/org/apache/flink/api/java/io/jdbc/JDBCLookupFunction.java
```diff
@@ -101,7 +113,8 @@ public JDBCLookupFunction(
 		this.maxRetryTimes = lookupOptions.getMaxRetryTimes();
 		this.keySqlTypes = Arrays.stream(keyTypes).mapToInt(JDBCTypeUtil::typeInformationToSqlType).toArray();
 		this.outputSqlTypes = Arrays.stream(fieldTypes).mapToInt(JDBCTypeUtil::typeInformationToSqlType).toArray();
-		this.query = options.getDialect().getSelectFromStatement(
+		this.nonNullableQuery = options.getDialect().getSelectFromStatement(options.getTableName(), fieldNames, keyNames);
+		this.nullableQuery = options.getDialect().getSelectNotDistinctFromStatement(
```
wuchong: If we have a consensus on the new methods of `SqlDialect`. we can call the field `nullSafeQuery` and `nonNullableQuery` -> `nullUnsafeQuery`.

flink-connectors/flink-jdbc/src/main/java/org/apache/flink/api/java/io/jdbc/JDBCLookupFunction.java
```diff
@@ -76,7 +87,8 @@
 	private final int maxRetryTimes;
 
 	private transient Connection dbConn;
-	private transient PreparedStatement statement;
+	private transient PreparedStatement fastStatement;
+	private transient PreparedStatement slowStatement;
```
wuchong: If we have a consensus on the new methods of `SqlDialect`. We can rename these fields to `nullSafeStatemen` and `nullUnsafeStatement`, we can add comment on the fields that we should use `nullUnsafeStatement` as much as possible because it is faster.

PR #9336, opened 2019-08-02
[FLINK-13548][Deployment/YARN]Support priority of the Flink YARN application

wzhero1:
This PR fixes the issue described in [1] that support priority of the Flink YARN application.

[1] https://issues.apache.org/jira/browse/FLINK-13548

## Brief change log

  -  provide APPLICATION_PRIORITY config in YarnConfigOptions

## Verifying this change

  - The serializers: no
## Documentation

### DIFF
--- a/docs/_includes/generated/yarn_config_configuration.html
+++ b/docs/_includes/generated/yarn_config_configuration.html
@@ -23,6 +23,11 @@
             <td>With this configuration option, users can specify a port, a range of ports or a list of ports for the Application Master (and JobManager) RPC port. By default we recommend using the default value (0) to let the operating system choose an appropriate port. In particular when multiple AMs are running on the same physical host, fixed port assignments prevent the AM from starting. For example when running Flink on YARN on an environment with a restrictive firewall, this option allows specifying a range of allowed ports.</td>
         </tr>
         <tr>
+            <td><h5>yarn.application.priority</h5></td>
+            <td style="word-wrap: break-word;">-1</td>
+            <td>A non-negative integer indicating the priority for submitting a Flink YARN application. It will only take effect if YARN priority scheduling setting is enabled. Larger integer corresponds with higher priority. If priority is negative or set to '-1'(default), Flink will unset yarn priority setting and use cluster default priority. Please refer to YARN's official documentation for specific settings required to enable priority scheduling for the targeted YARN version.</td>
+        </tr>
+        <tr>
             <td><h5>yarn.appmaster.rpc.address</h5></td>
             <td style="word-wrap: break-word;">(none)</td>
             <td>The hostname or address where the application master RPC system is listening.</td>
--- a/docs/ops/deployment/yarn_setup.md
+++ b/docs/ops/deployment/yarn_setup.md
@@ -262,6 +262,15 @@ Flink's YARN client has the following configuration parameters to control how to
 
 - `yarn.application-attempts`: The number of ApplicationMaster (+ its TaskManager containers) attempts. If this value is set to 1 (default), the entire YARN session will fail when the Application master fails. Higher values specify the number of restarts of the ApplicationMaster by YARN.
 
+## Setup for application priority on YARN
+
+Flink's YARN client has the following configuration parameters to setup application priority. These parameters can be set either from the `conf/flink-conf.yaml` or when starting the YARN session, using `-D` parameters.
+
+- `yarn.application.priority`: A non-negative integer indicating the priority for submitting a Flink YARN application. 
+It will only take effect if YARN priority scheduling setting is enabled. Larger integer corresponds with higher priority. 
+If priority is negative or set to '-1'(default), Flink will unset yarn priority setting and use cluster default priority. 
+Please refer to YARN's official documentation for specific settings required to enable priority scheduling for the targeted YARN version.
+
 ## Debugging a failed YARN session
 
 There are many reasons why a Flink YARN session deployment can fail. A misconfigured Hadoop setup (HDFS permissions, YARN configuration), version incompatibilities (running Flink with vanilla Hadoop dependencies on Cloudera Hadoop) or other errors.
--- a/docs/ops/deployment/yarn_setup.zh.md
+++ b/docs/ops/deployment/yarn_setup.zh.md
@@ -262,6 +262,15 @@ Flink's YARN client has the following configuration parameters to control how to
 
 - `yarn.application-attempts`: The number of ApplicationMaster (+ its TaskManager containers) attempts. If this value is set to 1 (default), the entire YARN session will fail when the Application master fails. Higher values specify the number of restarts of the ApplicationMaster by YARN.
 
+## Setup for application priority on YARN
+
+Flink's YARN client has the following configuration parameters to setup application priority. These parameters can be set either from the `conf/flink-conf.yaml` or when starting the YARN session, using `-D` parameters.
+
+- `yarn.application.priority`: A non-negative integer indicating the priority for submitting a Flink YARN application. 
+It will only take effect if YARN priority scheduling setting is enabled. Larger integer corresponds with higher priority. 
+If priority is negative or set to '-1'(default), Flink will unset yarn priority setting and use cluster default priority. 
+Please refer to YARN's official documentation for specific settings required to enable priority scheduling for the targeted YARN version.
+
 ## Debugging a failed YARN session
 
 There are many reasons why a Flink YARN session deployment can fail. A misconfigured Hadoop setup (HDFS permissions, YARN configuration), version incompatibilities (running Flink with vanilla Hadoop dependencies on Cloudera Hadoop) or other errors.
--- a/flink-yarn-tests/src/test/java/org/apache/flink/yarn/YarnPrioritySchedulingITCase.java
+++ b/flink-yarn-tests/src/test/java/org/apache/flink/yarn/YarnPrioritySchedulingITCase.java
@@ -1,0 +1,85 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.yarn;
+
+import org.apache.hadoop.util.VersionInfo;
+import org.apache.hadoop.yarn.api.records.ApplicationReport;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+import java.lang.reflect.Method;
+
+import static org.apache.flink.yarn.YarnTestUtils.isHadoopVersionGreaterThanOrEquals;
+import static org.hamcrest.Matchers.is;
+import static org.junit.Assert.assertThat;
+import static org.junit.Assume.assumeTrue;
+
+/**
+ * Tests to Yarn's priority scheduling.
+ */
+public class YarnPrioritySchedulingITCase extends YarnTestBase {
+
+	@BeforeClass
+	public static void setup() {
+		assumeTrue(
+			"Priority scheduling is not supported by Hadoop: " + VersionInfo.getVersion(),
+			isHadoopVersionGreaterThanOrEquals(2, 8));
+
+		YARN_CONFIGURATION.setStrings("yarn.cluster.max-application-priority", "10");
+		startYARNWithConfig(YARN_CONFIGURATION);
+	}
+
+	@Test
+	public void yarnApplication_submissionWithPriority_shouldRespectPriority() throws Exception {
+		runTest(() -> {
+			final int priority = 5;
+			final Runner yarnSessionClusterRunner = startWithArgs(new String[]{
+					"-j", flinkUberjar.getAbsolutePath(),
+					"-t", flinkLibFolder.getAbsolutePath(),
+					"-t", flinkShadedHadoopDir.getAbsolutePath(),
+					"-jm", "768m",
+					"-tm", "1024m",
+					"-Dyarn.application.priority=" + priority},
+				"Flink JobManager is now running on ",
+				RunTypes.YARN_SESSION);
+
+			try {
+				final ApplicationReport applicationReport = getOnlyApplicationReport();
+
+				assertApplicationIsStartedWithPriority(applicationReport, priority);
+			} finally {
+				yarnSessionClusterRunner.sendStop();
+				yarnSessionClusterRunner.join();
+			}
+		});
+	}
+
+	private void assertApplicationIsStartedWithPriority(ApplicationReport applicationReport, int priority) throws Exception {
+		final String getPriorityMethodName = "getPriority";
+
+		final Class<? extends ApplicationReport> applicationReportClass = applicationReport.getClass();
+		final Method getPriorityMethod = applicationReportClass.getMethod(getPriorityMethodName);
+		final Object priorityResult = getPriorityMethod.invoke(applicationReport);
+
+		final Class<?> priorityClass = priorityResult.getClass();
+		final Method getPriorityPriorityMethod = priorityClass.getMethod(getPriorityMethodName);
+
+		assertThat(getPriorityPriorityMethod.invoke(priorityResult), is(priority));
+	}
+}
--- a/flink-yarn/src/main/java/org/apache/flink/yarn/configuration/YarnConfigOptions.java
+++ b/flink-yarn/src/main/java/org/apache/flink/yarn/configuration/YarnConfigOptions.java
@@ -181,6 +181,22 @@ public class YarnConfigOptions {
 			" allowed ports.");
 
 	/**
+	 * A non-negative integer indicating the priority for submitting a Flink YARN application. It will only take effect
+	 * if YARN priority scheduling setting is enabled. Larger integer corresponds with higher priority. If priority is
+	 * negative or set to '-1'(default), Flink will unset yarn priority setting and use cluster default priority.
+	 *
+	 * @see <a href="https://hadoop.apache.org/docs/r2.8.5/hadoop-yarn/hadoop-yarn-site/CapacityScheduler.html">YARN Capacity Scheduling Doc</a>
+	 */
+	public static final ConfigOption<Integer> APPLICATION_PRIORITY =
+		key("yarn.application.priority")
+			.defaultValue(-1)
+			.withDescription("A non-negative integer indicating the priority for submitting a Flink YARN application. It" +
+				" will only take effect if YARN priority scheduling setting is enabled. Larger integer corresponds" +
+				" with higher priority. If priority is negative or set to '-1'(default), Flink will unset yarn priority" +
+				" setting and use cluster default priority. Please refer to YARN's official documentation for specific" +
+				" settings required to enable priority scheduling for the targeted YARN version.");
+
+	/**
 	 * A comma-separated list of strings to use as YARN application tags.
 	 */
 	public static final ConfigOption<String> APPLICATION_TAGS =
--- a/flink-yarn-tests/src/test/java/org/apache/flink/yarn/YARNSessionCapacitySchedulerITCase.java
+++ b/flink-yarn-tests/src/test/java/org/apache/flink/yarn/YARNSessionCapacitySchedulerITCase.java
@@ -50,7 +50,6 @@ import org.apache.hadoop.yarn.api.records.ApplicationReport;
 import org.apache.hadoop.yarn.api.records.YarnApplicationState;
 import org.apache.hadoop.yarn.client.api.YarnClient;
 import org.apache.hadoop.yarn.conf.YarnConfiguration;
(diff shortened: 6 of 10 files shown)


docs/_includes/generated/yarn_config_configuration.html
```diff
@@ -22,6 +22,11 @@
             <td style="word-wrap: break-word;">"0"</td>
             <td>With this configuration option, users can specify a port, a range of ports or a list of ports for the Application Master (and JobManager) RPC port. By default we recommend using the default value (0) to let the operating system choose an appropriate port. In particular when multiple AMs are running on the same physical host, fixed port assignments prevent the AM from starting. For example when running Flink on YARN on an environment with a restrictive firewall, this option allows specifying a range of allowed ports.</td>
         </tr>
+        <tr>
```
walterddr: you might also need to change `yarn-setup.md` to add some examples into the launch CLI script.

docs/_includes/generated/yarn_config_configuration.html
```diff
@@ -22,6 +22,11 @@
...
+            <td><h5>yarn.application.priority</h5></td>
+            <td style="word-wrap: break-word;">-1</td>
+            <td>The priority which Flink used to submit YARN application. The priority is non negative, the bigger the number, the higher the priority. By default, we take -1. When the priority is negative, we use default yarn queue priority.</td>
```
walterddr: maybe it is nice to hyperlink with YARN's priority scheduling doc [here](https://hadoop.apache.org/docs/current/hadoop-yarn/hadoop-yarn-site/CapacityScheduler.html), 

Just out of top of my head.
```
A non-negative integer indicating the priority for submitting a Flink YARN application. 
It will only take effect if YARN priority scheduling setting is enabled. 
Larger integer corresponds with higher the priority. 
If priority is set to '-1', Flink will unset yarn priority setting and use cluster default priority, see [YARN Capacity Scheduling Doc](https://hadoop.apache.org/docs/current/hadoop-yarn/hadoop-yarn-site/CapacityScheduler.html) for more info.
```

wzhero1: It is very useful, I have changed the option description according to this.

docs/_includes/generated/yarn_config_configuration.html
wzhero1: ok, I have added the instruction in yarn-setup.md and yarn-setup.zh.md(used in English).

wzhero1:
@walterddr Thanks for your review, I have modified YARN config description according to the comments.

wzhero1:
@walterddr I found CI failed by timeout problem, how could I rebuild the CI~

docs/_includes/generated/yarn_config_configuration.html
walterddr: could you also add information indicate that this only works for newer version of YARN like we discussed in the JIRA ticket? based on what I researched this is only enabled on YARN 2.8+, not sure if it was backported to any older version.

docs/ops/deployment/yarn_setup.md
```diff
@@ -262,6 +262,12 @@ Flink's YARN client has the following configuration parameters to control how to
 
 - `yarn.application-attempts`: The number of ApplicationMaster (+ its TaskManager containers) attempts. If this value is set to 1 (default), the entire YARN session will fail when the Application master fails. Higher values specify the number of restarts of the ApplicationMaster by YARN.
 
+## Setup for application priority on YARN
+
+Flink's YARN client has the following configuration parameters to setup application priority. These parameters can be set either from the `conf/flink-conf.yaml` or when starting the YARN session, using `-D` parameters.
+
+- `yarn.application.priority`: A non-negative integer indicating the priority for submitting a Flink YARN application. It will only take effect if YARN priority scheduling setting is enabled. Larger integer corresponds with higher priority. If priority is negative or set to '-1'(default), Flink will unset yarn priority setting and use cluster default priority, see [YARN Capacity Scheduling Doc](https://hadoop.apache.org/docs/current/hadoop-yarn/hadoop-yarn-site/CapacityScheduler.html) for more info.
```
walterddr: I think we should use specific version that first supported this instead of `current` in the URL link.

wzhero1: Thanks for the discussion on the JIRA issue! I will modify the doc description to specify the current supported version of YARN. And since most of our current jobs are still running with flink 1.5 API with priority feature we add, so the API calls 'appContext.setPriority' should be available in later versions.

docs/_includes/generated/yarn_config_configuration.html
wzhero1: yes, I have added the supported version info in the config description.

docs/ops/deployment/yarn_setup.md
wzhero1: @walterddr Thanks for your review~ I have added the supported Hadoop version info according to the advice.

flink-yarn/src/main/java/org/apache/flink/yarn/configuration/YarnConfigOptions.java
```diff
@@ -180,6 +180,25 @@
...
+	 * cluster default priority.
+	 *
+	 * @see <a href="https://hadoop.apache.org/docs/r2.8.5/hadoop-yarn/hadoop-yarn-site/CapacityScheduler.html">YARN Capacity Scheduling Doc</a>
+	 */
+	public static final ConfigOption<Integer> APPLICATION_PRIORITY =
+		key("yarn.application.priority")
+			.defaultValue(-1)
+			.withDescription(Description.builder()
+				.text("A non-negative integer indicating the priority for submitting a Flink YARN application. It" +
+					" will only take effect if the Hadoop version >= 2.8.5 and YARN priority scheduling setting is enabled." +
```
tillrohrmann: Why does it only work with Hadoop version >= `2.8.5`? The `ApplicationSubmissionContext` has the `#setPriority` method since version `2.0.4-alpha`? Has the method never been used before `2.8.5`?

walterddr: @tillrohrmann Yes. I saw that the API was there for a while. However, according the documentation, priority scheduling only appears in YARN documentation after 2.8.x. In fact I tried the integration test in `flink-yarn-test` module by compiling against default (2.4.x?) and 2.8.x. 

My finding was: only 2.8.x tests has specific logs indicating that priority setting has taken effect:
```
2019-08-15 19:46:52,182 INFO  org.apache.flink.yarn.YarnResourceManager                     - Received new container: container_1565923587220_0001_01_000002 - Remaining pending container requests: 1
2019-08-15 19:46:52,184 INFO  org.apache.flink.yarn.YarnResourceManager                     - Removing container request Capability[<memory:1024, vCores:1>]**Priority[1]**. Pending container requests 0.
```
similar logs does not exist on 2.4.x.

walterddr: I could've been looking for logs in the wrong place. @wzhero1 can probably provide more details regarding what signals in the log we should be looking at.

tillrohrmann:
Just a comment for a possible test. I think it would be enough to test that we submit an `ApplicationSubmissionContext` with the priority set. Hence we could have a `YarnClient` test implementation which we pass to the `AbstractYarnClusterDescriptor`. Not sure, though, how much mocked functionality the `YarnClient` would need to be able to trigger the submit application call.

flink-yarn/src/main/java/org/apache/flink/yarn/configuration/YarnConfigOptions.java
wzhero1: I had tested in a non-priority queue, it took no effect. And in a priority queue, the priority value took effect and would be restricted according to the max-application-priority.

wzhero1:
ok, I will consider to add a test in flink-yarn-tests, maybe in YarnConfigurationITCase.

wzhero1:
ok, I will consider to add a test in flink-yarn-tests, maybe in YarnConfigurationITCase.

docs/_includes/generated/yarn_config_configuration.html
walterddr: @wzhero1 we might want to revert the 2.8.x constrain here. and say something like this.
```
Please refer to YARN official documentation for specific settings required to enable priority schedule for the targeted YARN version
```

sorry for the confusion :-(

walterddr:
for testing. I think we can just test the fact that appContext contains priority settings correctly when `yarnClient.submitApplication(appContext);` is called. this covers most of the code portion: e.g. convert a commander line setting into a specific setting in appContext. what do you think?

wzhero1:
@walterddr Thanks for your detailed research.  I will change the doc and add a test in `FlinkYarnSessionCliTest` to test `setPriority` operation is valid. And after that we can submit a version to provide this yarn config parameter first.

docs/_includes/generated/yarn_config_configuration.html
wzhero1: @walterddr I also found in 2.8.x,  the `ApplicationReport` which is generated by `yarnClient.getApplicationReport` has the method `set/getPriority`, but there is no such method in 2.6.x. 

> Hi here,
> 
> did a bit of research: the API is already there in: [ApplicationSubmissionContext](https://hadoop.apache.org/docs/r2.4.1/api/org/apache/hadoop/yarn/api/records/ApplicationSubmissionContext.html). However, the usage pattern is different: it is only introduced to Capacity scheduler in [2.8.x](https://hadoop.apache.org/docs/r2.8.5/hadoop-yarn/hadoop-yarn-site/CapacityScheduler.html#Setup_for_application_priority).
> 
> The priority setting was used in [ResourceRequest](http://hadoop.apache.org/docs/r2.4.1/hadoop-yarn/hadoop-yarn-site/WritingYarnApplications.html), which is a different usage pattern: after [2.6.x](https://hadoop.apache.org/docs/r2.6.5//hadoop-yarn/hadoop-yarn-site/WritingYarnApplications.html), YARN documentation indicates that `appContext.setPriority(...)` should be used instead of `rsrcRequest.setPriority(...)`
> 
> I am still not sure how prior to 2.6.x what is the effect of setting `appContext.setPriority`.

walterddr:
thanks for the update @wzhero1 . I just ran [some E2E test](https://github.com/walterddr/flink/commit/70c48d09d2d3c446b7146eaf98622c4170d41f14) in `flink-yarn-test` with priority settings in capacity scheduler [here](https://travis-ci.com/walterddr/flink/jobs/226132781). seems like it is taking effect correctly since Travis uses default `hadoop-2.8.3`. I will do another pass once tests are updated. thanks -Rong

wzhero1:
@walterddr Thanks for the test about ApplicationReport~ I have updated the doc and added a test in `FlinkYarnSessionCliTest` to check whether the command line is correctly converted into flinkConfiguration before deployment. Also I think the ApplicationReport test may be integrated into Flink later after the hadoop version is upgraded after 2.8.x.

docs/_includes/generated/yarn_config_configuration.html
```diff
@@ -22,6 +22,11 @@
...
         </tr>
+        <tr>
+            <td><h5>yarn.application.priority</h5></td>
+            <td style="word-wrap: break-word;">-1</td>
+            <td>A non-negative integer indicating the priority for submitting a Flink YARN application. It will only take effect if YARN priority scheduling setting is enabled. Larger integer corresponds with higher priority. If priority is negative or set to '-1'(default), Flink will unset yarn priority setting and use cluster default priority. Please refer to <a href="https://hadoop.apache.org/docs/r2.8.5/hadoop-yarn/hadoop-yarn-site/CapacityScheduler.html">YARN Capacity Scheduling Doc</a> for specific settings required to enable priority schedule for the targeted YARN version.</td>
```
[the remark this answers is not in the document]
wzhero1: Thanks for your review~ I have updated the yarn doc as the comment mentioned above.

flink-yarn/src/test/java/org/apache/flink/yarn/FlinkYarnSessionCliTest.java
```diff
@@ -137,6 +137,32 @@ public void testCorrectSettingOfDetachedMode() throws Exception {
...
+		// YarnClusterDescriptor will add dynamic properties to local flinkConfiguraton before deployment
+		Map<String, String> dynProperties =
+			FlinkYarnSessionCli.getDynamicProperties(descriptor.getDynamicPropertiesEncoded());
+		for (Map.Entry<String, String> dynProperty : dynProperties.entrySet()) {
+			flinkConfiguration.setString(dynProperty.getKey(), dynProperty.getValue());
+		}
+
+		int priority = flinkConfiguration.getInteger(YarnConfigOptions.APPLICATION_PRIORITY);
+		assertEquals(10, priority);
+	}
```
tillrohrmann: What do we test here exactly? Shouldn't we rather test that given `YarnConfigOptions.APPLICATION_PRIORITY` being set that we create an `ApplicationSubmissionContext` with this priority?

docs/ops/deployment/yarn_setup.md
```diff
@@ -262,6 +262,15 @@ Flink's YARN client has the following configuration parameters to control how to
...
 - `yarn.application-attempts`: The number of ApplicationMaster (+ its TaskManager containers) attempts. If this value is set to 1 (default), the entire YARN session will fail when the Application master fails. Higher values specify the number of restarts of the ApplicationMaster by YARN.
 
+## Setup for application priority on YARN
+
+Flink's YARN client has the following configuration parameters to setup application priority. These parameters can be set either from the `conf/flink-conf.yaml` or when starting the YARN session, using `-D` parameters.
+
+- `yarn.application.priority`: A non-negative integer indicating the priority for submitting a Flink YARN application. 
+It will only take effect if YARN priority scheduling setting is enabled. Larger integer corresponds with higher priority. 
+If priority is negative or set to '-1'(default), Flink will unset yarn priority setting and use cluster default priority. 
+Please refer to YARN official documentation for specific settings required to enable priority schedule for the targeted YARN version.
```
tillrohrmann:
```suggestion
Please refer to YARN's official documentation for specific settings required to enable priority scheduling for the targeted YARN version.
```

docs/ops/deployment/yarn_setup.zh.md
```diff
@@ -262,6 +262,15 @@ Flink's YARN client has the following configuration parameters to control how to
...
 - `yarn.application-attempts`: The number of ApplicationMaster (+ its TaskManager containers) attempts. If this value is set to 1 (default), the entire YARN session will fail when the Application master fails. Higher values specify the number of restarts of the ApplicationMaster by YARN.
 
+## Setup for application priority on YARN
+
+Flink's YARN client has the following configuration parameters to setup application priority. These parameters can be set either from the `conf/flink-conf.yaml` or when starting the YARN session, using `-D` parameters.
+
+- `yarn.application.priority`: A non-negative integer indicating the priority for submitting a Flink YARN application. 
+It will only take effect if YARN priority scheduling setting is enabled. Larger integer corresponds with higher priority. 
+If priority is negative or set to '-1'(default), Flink will unset yarn priority setting and use cluster default priority. 
+Please refer to YARN official documentation for specific settings required to enable priority schedule for the targeted YARN version.
```
tillrohrmann:
```suggestion
Please refer to YARN's official documentation for specific settings required to enable priority scheduling for the targeted YARN version.
```

flink-yarn/src/main/java/org/apache/flink/yarn/configuration/YarnConfigOptions.java
```diff
@@ -180,6 +180,22 @@
...
+	 *
+	 * @see <a href="https://hadoop.apache.org/docs/r2.8.5/hadoop-yarn/hadoop-yarn-site/CapacityScheduler.html">YARN Capacity Scheduling Doc</a>
+	 */
+	public static final ConfigOption<Integer> APPLICATION_PRIORITY =
+		key("yarn.application.priority")
+			.defaultValue(-1)
+			.withDescription("A non-negative integer indicating the priority for submitting a Flink YARN application. It" +
+				" will only take effect if YARN priority scheduling setting is enabled. Larger integer corresponds" +
+				" with higher priority. If priority is negative or set to '-1'(default), Flink will unset yarn priority" +
+				" setting and use cluster default priority. Please refer to YARN official documentation for specific" +
```
tillrohrmann:
```suggestion
				" setting and use cluster default priority. Please refer to YARN's official documentation for specific" +
```

flink-yarn/src/main/java/org/apache/flink/yarn/configuration/YarnConfigOptions.java
```diff
@@ -180,6 +180,22 @@
...
+				" settings required to enable priority schedule for the targeted YARN version.");
```
tillrohrmann:
```suggestion
				" settings required to enable priority scheduling for the targeted YARN version.");
```

tillrohrmann:
@flinkbot run travis

tillrohrmann:
@wzhero1 what do you mean with `ApplicationReport` test can be merged after the Hadoop version is upgraded?

flink-yarn/src/test/java/org/apache/flink/yarn/FlinkYarnSessionCliTest.java
tillrohrmann: Yes, we could do something similar to the `AbstractYarnClusterTest` where we use the `TestingYarnClient` as a mock instance.

wzhero1:
@tillrohrmann @walterddr Thank you very much for the discussion~ I think maybe there are two places we should test.
1. the configuration we configured in Cli could be correctly parsed into YARN dynamic configuration. I found in `YARNSessionCapacitySchedulerITCase#testVCoresAreSet...()`, the method would getFlinkConfig by quering the yarn cluster. I added a priority config in this method, and it could correctly be parsed, which means the flinkConfiguration has the property when `AbstractYarnClusterDescriptor#startAppMaster()`,  so we can infer the property can be well configured in the `ApplicationSubmissionContext`(and debugging can confirm this Inference ). Also I found `FlinkYarnSessionCliTest#testConfigurationBeforeDeployment()` I added before is redundant so I have deleted this test.
https://github.com/apache/flink/blob/17e2747e575cc4d9847be76b5ce9a75a7b6707f7/flink-yarn-tests/src/test/java/org/apache/flink/yarn/YARNSessionCapacitySchedulerITCase.java#L254
2. after sumbitting the application, we should get the priority report from `ApplicationReport` to check whether the priority config takes effect. `YARNSessionCapacitySchedulerITCase#testVCoresAreSet...()` also provides the way to check the `ApplicationReport`. But in hadoop-2.4.1 api, the `ApplicationReport` does not provide the `getPriority()` method(the method is supported after 2.8.0). So in current flink(hadoop version:2.4.1), we could not to test the `ApplicationReport`.

Since currently `ApplicationSubmissionContext` is wrapped in the `AbstractYarnClusterDescriptor#startAppMaster` and hard to get and test, so I think maybe it is proper to test as I talked above. I sumbitted a new commit as I said, looking forward to your better ideas, thanks~

flink-yarn/src/main/java/org/apache/flink/yarn/configuration/YarnConfigOptions.java
wzhero1: Hi @walterddr it's tested in our yarn cluster which is based on the 2.8.x.

docs/ops/deployment/yarn_setup.md
wzhero1: done

docs/ops/deployment/yarn_setup.zh.md
wzhero1: done

flink-yarn/src/main/java/org/apache/flink/yarn/configuration/YarnConfigOptions.java
wzhero1: done

wzhero1:
Besides, it seems that using mocks to test `AbstractYarnClusterDescriptor#startAppMaster()` is very difficult. That method relies on too many dependencies which cannot be mocked properly, e.g. `setupSingleLocalResource()` relies on a local jar and copies it to HDFS.

Could we just test the configuration parse in this commit (like other existing test cases), and refactor the `AbstractYarnClusterDescriptor#startAppMaster()` in another commit?

tillrohrmann:
I've updated the PR @wzhero1 so that it only tests the priority if the right Hadoop version is used. You can take a look at the last two commits to see how I did it. Merging once Travis gives green light.

flink-yarn-tests/src/test/java/org/apache/flink/yarn/YarnPrioritySchedulingITCase.java
```diff
@@ -0,0 +1,85 @@
...
+ */
+public class YarnPrioritySchedulingITCase extends YarnTestBase {
+
+	@BeforeClass
+	public static void setup() {
+		assumeTrue(
+			"Priority scheduling is not supported by Hadoop: " + VersionInfo.getVersion(),
+			isHadoopVersionGreaterThanOrEquals(2, 8));
+
+		YARN_CONFIGURATION.setStrings("yarn.cluster.max-application-priority", "10");
```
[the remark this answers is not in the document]
walterddr: my bad. CapacityScheduler is the default. Thought we are extending from `FIFOITCase`.

wzhero1:
Thanks for the supplementary test of YARN priority @tillrohrmann . I think it looks good to merge.

walterddr:
thanks for adding the ITCase test and the quick merge @tillrohrmann

Jesuitry:
Can this be used in the fair scheduler？I set yarn.application.priority, but it does not take effect.

tillrohrmann:
I looks to me as if this only works with Yarn's capacity scheduler: https://docs.cloudera.com/HDPDocuments/HDP3/HDP-3.0.1/data-operating-system/content/enable_priority_scheduling.html#:~:text=YARN%20allocates%20more%20resources%20to,and%20dynamically%20at%20run%20time.

PR #9339, opened 2019-08-02
[FLINK-13555][runtime] SlotPool fails batch slot requests immediately if they are unfulfillable.

xintongsong:

This PR make SlotPool to fail batch slot requests immediately if they are unfulfillable, which means they cannot be fulfilled by any slot (available or not) in the cluster.

There are two kinds of request slots from RM failures:
1. RM does not have available slots. All slots are in use at the moment. But they might become available later when the currently running tasks finish.
2. The slot request requires too many resources that can not be fulfilled by any slot (available or not) in the cluster. The request is also not likely to be fulfilled later.

For the 2nd kinds of failures, it doesn't make sense to wait for the timeout. SlotPool should fail the job immediately, with proper error messages describing the problem and suggesting the user to tune job or cluster configurations.

## Brief change log

- 6af614c1647429d76c8514a67e7c5439e90a50e1, 8b5e9ff88e6052dc5845a9fcff684261c697899f: Cleanups for existing codes.
- dbdace5ae68331bd1203525425b357639e334752: Remove `SlotManagerException`, replace with its base class `ResourceException`.
- bb2c7c92fdbafe3644ab76041b300a23a6c87744: Introduce `UnfulfillableSlotRequestException` and fail pending batch slot requests on it.
- dfe5ef072dadfc0e1510f1923911304c635e54af: Add test cases for failing unfulfillable batch slot requests.

## Verifying this change

- Add `SlotPoolBatchSlotRequestTest#testPendingBatchSlotRequestFailsIfAllocationFailsUnfulfillable` that validates that pending batch slot requests fail when allocation fail with unfulfillable exception.
- Add `SlotPoolBatchSlotRequestTest#testPendingBatchSlotRequestFailsIfRMRequestFailsUnfulfillable` that validates that pending batch slot requests fail when requesting slots from RM fail with unfulfillable exception. 

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/resourcemanager/exceptions/UnfulfillableSlotRequestException.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/resourcemanager/exceptions/UnfulfillableSlotRequestException.java
@@ -16,19 +16,21 @@
  * limitations under the License.
  */
 
-package org.apache.flink.runtime.resourcemanager.slotmanager;
+package org.apache.flink.runtime.resourcemanager.exceptions;
 
-import org.apache.flink.runtime.resourcemanager.exceptions.ResourceManagerException;
+import org.apache.flink.runtime.clusterframework.types.AllocationID;
+import org.apache.flink.runtime.clusterframework.types.ResourceProfile;
 
-public class SlotManagerException extends ResourceManagerException {
 
-	private static final long serialVersionUID = -3723028616920379071L;
+/**
+ * Exception denoting that a slot request can not be fulfilled by any slot in the cluster.
+ * This usually indicates that the slot request should not be pended or retried.
+ */
+public class UnfulfillableSlotRequestException extends ResourceManagerException {
+	private static final long serialVersionUID = 4453490263648758730L;
 
-	public SlotManagerException(String message) {
-		super(message);
-	}
-
-	public SlotManagerException(String message, Throwable cause) {
-		super(message, cause);
+	public UnfulfillableSlotRequestException(AllocationID allocationId, ResourceProfile resourceProfile) {
+		super("Could not fulfill slot request " + allocationId + ". "
+			+ "Requested resource profile (" + resourceProfile + ") is unfulfillable.");
 	}
 }
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/resourcemanager/slotmanager/SlotManagerImpl.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/resourcemanager/slotmanager/SlotManagerImpl.java
@@ -117,7 +117,6 @@ public class SlotManagerImpl implements SlotManager {
 
 	/**
 	 * If true, fail unfulfillable slot requests immediately. Otherwise, allow unfulfillable request to pend.
-	 *
 	 * A slot request is considered unfulfillable if it cannot be fulfilled by neither a slot that is already registered
 	 * (including allocated ones) nor a pending slot that the {@link ResourceActions} can allocate.
 	 * */
@@ -211,8 +210,7 @@ public class SlotManagerImpl implements SlotManager {
 	 * @param newResourceActions to use for resource (de-)allocations
 	 */
 	@Override
-	public void start(ResourceManagerId newResourceManagerId, Executor newMainThreadExecutor,
-					  ResourceActions newResourceActions) {
+	public void start(ResourceManagerId newResourceManagerId, Executor newMainThreadExecutor, ResourceActions newResourceActions) {
 		LOG.info("Starting the SlotManager.");
 
 		this.resourceManagerId = Preconditions.checkNotNull(newResourceManagerId);
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/resourcemanager/slotmanager/SlotManagerImpl.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/resourcemanager/slotmanager/SlotManagerImpl.java
@@ -31,6 +31,7 @@ import org.apache.flink.runtime.messages.Acknowledge;
 import org.apache.flink.runtime.resourcemanager.ResourceManagerId;
 import org.apache.flink.runtime.resourcemanager.SlotRequest;
 import org.apache.flink.runtime.resourcemanager.exceptions.ResourceManagerException;
+import org.apache.flink.runtime.resourcemanager.exceptions.UnfulfillableSlotRequestException;
 import org.apache.flink.runtime.resourcemanager.registration.TaskExecutorConnection;
 import org.apache.flink.runtime.taskexecutor.SlotReport;
 import org.apache.flink.runtime.taskexecutor.SlotStatus;
@@ -290,10 +291,10 @@ public class SlotManagerImpl implements SlotManager {
 	 *
 	 * @param slotRequest specifying the requested slot specs
 	 * @return true if the slot request was registered; false if the request is a duplicate
-	 * @throws SlotManagerException if the slot request failed (e.g. not enough resources left)
+	 * @throws ResourceManagerException if the slot request failed (e.g. not enough resources left)
 	 */
 	@Override
-	public boolean registerSlotRequest(SlotRequest slotRequest) throws SlotManagerException {
+	public boolean registerSlotRequest(SlotRequest slotRequest) throws ResourceManagerException {
 		checkInit();
 
 		if (checkDuplicateRequest(slotRequest.getAllocationId())) {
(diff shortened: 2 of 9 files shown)


flink-runtime/src/main/java/org/apache/flink/runtime/resourcemanager/exceptions/UnfulfillableSlotRequestException.java
```diff
@@ -31,4 +32,9 @@ public UnfulfillableSlotRequestException(PendingSlotRequest pendingSlotRequest)
 		super("Could not fulfill slot request " + pendingSlotRequest.getAllocationId() + ". "
 			+ "Requested resource profile (" + pendingSlotRequest.getResourceProfile() + ") is unfulfillable.");
 	}
+
+	@VisibleForTesting
+	public UnfulfillableSlotRequestException() {
+		super("UnfulfillableSlotRequestException for testing purpose.");
+	}
```
tillrohrmann: Instead of introducing this constructor for tests only, I would change the existing on to accept `UnfulfillableSlotRequestException(AllocationID, ResourceProfile)`.

flink-runtime/src/test/java/org/apache/flink/runtime/jobmaster/slotpool/SlotPoolBatchSlotRequestTest.java
```diff
@@ -147,14 +148,42 @@ public void testPendingBatchSlotRequestDoesNotFailIfAllocationFails() throws Exc
...
 		}
 	}
 
 	/**
-	 * Tests that a batch slot request won't fail if its resource manager request fails.
+	 * Tests that a batch slot request does react to {@link SlotPool#failAllocation(AllocationID, Exception)}
+	 * signals whose exception is {@link UnfulfillableSlotRequestException}.
+	 */
+	@Test
+	public void testPendingBatchSlotRequestFailsIfAllocationFailsUnfulfillable() throws Exception {
```
tillrohrmann: `..FailsUnfulfillably`

flink-runtime/src/test/java/org/apache/flink/runtime/jobmaster/slotpool/SlotPoolBatchSlotRequestTest.java
```diff
@@ -147,14 +148,42 @@ public void testPendingBatchSlotRequestDoesNotFailIfAllocationFails() throws Exc
...
+		final TestingResourceManagerGateway testingResourceManagerGateway = new TestingResourceManagerGateway();
+		final CompletableFuture<AllocationID> allocationIdFuture = new CompletableFuture<>();
+		testingResourceManagerGateway.setRequestSlotConsumer(slotRequest -> allocationIdFuture.complete(slotRequest.getAllocationId()));
+
+		final ComponentMainThreadExecutor directMainThreadExecutor = ComponentMainThreadExecutorServiceAdapter.forMainThread();
+
+		final Time batchSlotTimeout = Time.milliseconds(1000L);
```
tillrohrmann: No need to set this timeout. Per default this should be 2s if using the `SlotPoolBuilder`.

flink-runtime/src/test/java/org/apache/flink/runtime/jobmaster/slotpool/SlotPoolBatchSlotRequestTest.java
```diff
@@ -175,6 +204,28 @@ public void testPendingBatchSlotRequestDoesNotFailIfRMRequestFails() throws Exce
...
+	 * Tests that a batch slot request fails if its resource manager request fails with {@link UnfulfillableSlotRequestException}.
+	 */
+	@Test
+	public void testPendingBatchSlotRequestFailsIfRMRequestFailsUnfulfillable() throws Exception {
+		final TestingResourceManagerGateway testingResourceManagerGateway = new TestingResourceManagerGateway();
+		testingResourceManagerGateway.setRequestSlotFuture(FutureUtils.completedExceptionally(new UnfulfillableSlotRequestException()));
+
+		final ComponentMainThreadExecutor directMainThreadExecutor = ComponentMainThreadExecutorServiceAdapter.forMainThread();
+
+		final Time batchSlotTimeout = Time.milliseconds(1000L);
```
tillrohrmann: Same here.

flink-runtime/src/test/java/org/apache/flink/runtime/jobmaster/slotpool/SlotPoolBatchSlotRequestTest.java
```diff
@@ -147,14 +148,42 @@ public void testPendingBatchSlotRequestDoesNotFailIfAllocationFails() throws Exc
...
+			.setBatchSlotTimeout(batchSlotTimeout)
+			.setResourceManagerGateway(testingResourceManagerGateway)
+			.build()) {
+
+			final CompletableFuture<PhysicalSlot> slotFuture = SlotPoolUtils.requestNewAllocatedBatchSlot(slotPool, directMainThreadExecutor, resourceProfile);
+
+			SlotPoolUtils.failAllocation(slotPool, directMainThreadExecutor, allocationIdFuture.get(),
+				new UnfulfillableSlotRequestException());
+
+			assertThat(slotFuture.isDone(), is(true));
```
tillrohrmann: I guess we should check that the `slotFuture` has been completed exceptionally.

flink-runtime/src/test/java/org/apache/flink/runtime/jobmaster/slotpool/SlotPoolBatchSlotRequestTest.java
```diff
@@ -175,6 +204,28 @@ public void testPendingBatchSlotRequestDoesNotFailIfRMRequestFails() throws Exce
...
+
+		final Time batchSlotTimeout = Time.milliseconds(1000L);
+		try (final SlotPoolImpl slotPool = new SlotPoolBuilder(directMainThreadExecutor)
+			.setBatchSlotTimeout(batchSlotTimeout)
+			.setResourceManagerGateway(testingResourceManagerGateway)
+			.build()) {
+
+			final CompletableFuture<PhysicalSlot> slotFuture = SlotPoolUtils.requestNewAllocatedBatchSlot(slotPool, directMainThreadExecutor, resourceProfile);
+
+			assertThat(slotFuture.isDone(), is(true));
```
tillrohrmann: I guess we should check that the `slotFuture` has been completed exceptionally.

flink-runtime/src/main/java/org/apache/flink/runtime/resourcemanager/slotmanager/SlotManagerImpl.java
```diff
@@ -767,8 +767,7 @@ private void internalRequestSlot(PendingSlotRequest pendingSlotRequest) throws R
 				// request can not be fulfilled by any free slot or pending slot that can be allocated,
 				// check whether it can be fulfilled by allocated slots
 				if (failUnfulfillableRequest && !isFulfillableByRegisteredSlots(pendingSlotRequest.getResourceProfile())) {
-					throw new ResourceManagerException("Could not fulfill slot request " + pendingSlotRequest.getAllocationId() + ". "
-						+ "Requested resource profile (" + pendingSlotRequest.getResourceProfile() + ") is unfulfillable.");
+					throw new UnfulfillableSlotRequestException(pendingSlotRequest);
```
tillrohrmann: I think we also need to update the method `SlotManagerImpl#setFailUnfulfillableRequest` to fail the slots with the proper exception. Please also add a test because this is apparently not guarded.

flink-runtime/src/main/java/org/apache/flink/runtime/resourcemanager/slotmanager/SlotManagerImpl.java
```diff
@@ -496,8 +495,7 @@ public void setFailUnfulfillableRequest(boolean failUnfulfillableRequest) {
 					resourceActions.notifyAllocationFailure(
 						pendingSlotRequest.getJobId(),
 						pendingSlotRequest.getAllocationId(),
-						new ResourceManagerException("Could not fulfill slot request " + pendingSlotRequest.getAllocationId() + ". "
-							+ "Requested resource profile (" + pendingSlotRequest.getResourceProfile() + ") is unfulfillable.")
+						new UnfulfillableSlotRequestException(pendingSlotRequest.getAllocationId(), pendingSlotRequest.getResourceProfile())
```
tillrohrmann: Could we extend `SlotManagerFailUnfulfillableTest` to verify that we throw the correct exception in this method `setFailUnfulfillableRequest` and `internalRequestSlot`?

xintongsong: I think the updates have already extended `SlotManagerFailUnfulfillableTest` to verify the exception being thrown. It's in the last commit.

tillrohrmann: Great. Sorry I overlooked this.

PR #9340, opened 2019-08-02
 [FLINK-13384][runtime] Fix back pressure sampling for SourceStreamTask

GJL:
See #9271


flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SourceStreamTask.java
```diff
@@ -58,6 +61,7 @@
 
 	public SourceStreamTask(Environment env) {
 		super(env);
+		this.sourceThread = new LegacySourceFunctionThread(getName());
```
GJL: Here we are calling a non-final method from the constructor :(

GJL: The name could be set in `performDefaultAction()`, before `sourceThread.start();`

pnowojski: Not the best solution, but let's not overthink this. +1 for setting the name in `performDefaultAction`

GJL:
Merged

PR #9236, opened 2019-07-26
[FLINK-13283][FLINK-13490][jdbc] Fix JDBC connectors with DataTypes.DATE/TIME/TIMESTAMP support and null checking

tsreaper:
## What is the purpose of the change

Currently JDBC connector does not support DataTypes.DATE/TIME/TIMESTAMP with their default conversion classes to be LocalDate/LocalTime/LocalDateTime. This PR fixes the support of these data types for JDBC connector.

<del>Also, the null checking in `JDBCUtils::getFieldFromResultSet` is incorrect. Under the current implementation, if one column is null in the result set, the following calls to this method using the same result set will always return null, no matter what the content of the column is.</del> merged

## Brief change log

 - Fix JDBC connector with DataTypes.DATE/TIME/TIMESTAMP support
 - <del>Fix null checking in `JDBCUtils::getFieldFromResultSet`</del> merged

## Verifying this change

This change is already covered by existing tests, such as the existing tests in the jdbc connector package. This PR updates these tests with some time data types and some null values.

## Does this pull request potentially affect one of the following parts:

  - Dependencies (does it add or upgrade a dependency): no
  - The public API, i.e., is any changed class annotated with `@Public(Evolving)`: no
  - The serializers: no
  - The runtime per-record code paths (performance sensitive):no
  - Anything that affects deployment or recovery: JobManager (and its components), Checkpointing, Yarn/Mesos, ZooKeeper: no
  - The S3 file system connector: no

## Documentation

  - Does this pull request introduce a new feature? no
  - If yes, how is the feature documented? not applicable

PR #9342, opened 2019-08-02
[FLINK-13438][hive] Fix DataTypes.DATE/TIME/TIMESTAMP support for hive connectors

TsReaper:

Similar to JDBC connectors, Hive connectors communicate with Flink framework using TableSchema, which contains DataType. As the time data read from and write to Hive connectors must be `java.sql.*` types and the default conversion class of our time data types are `java.time.*`, we have to fix Hive connector with DataTypes.DATE/TIME/TIMESTAMP support.

This PR fixes `DataTypes.DATE/TIME/TIMESTAMP` support for hive connectors, and is similar to PR #9236 .

## Brief change log

 - Fix `DataTypes.DATE/TIME/TIMESTAMP` support for hive connectors.

## Verifying this change

This change added tests and can be verified as follows: run the newly added `HiveFullTest`. This new test is to test if the time data can correctly flow from hive source to flink system and finally to hive sink.

There seems to be other bugs in hive connectors. When running this tests along with other hive tests in idea, `HiveTableSinkTest` will fail; bug when running this tests along with others in maven the tests will all pass. @zjuwangg please take a look into this issue.

  - The serializers: no

## Documentation

### DIFF
--- a/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/connectors/hive/HiveTableSink.java
+++ b/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/connectors/hive/HiveTableSink.java
@@ -27,6 +27,7 @@ import org.apache.flink.table.catalog.exceptions.CatalogException;
 import org.apache.flink.table.catalog.hive.client.HiveMetastoreClientFactory;
 import org.apache.flink.table.catalog.hive.client.HiveMetastoreClientWrapper;
 import org.apache.flink.table.catalog.hive.descriptors.HiveCatalogValidator;
+import org.apache.flink.table.catalog.hive.util.HiveTableUtil;
 import org.apache.flink.table.sinks.OutputFormatTableSink;
 import org.apache.flink.table.sinks.OverwritableTableSink;
 import org.apache.flink.table.sinks.PartitionableTableSink;
@@ -74,10 +75,13 @@ public class HiveTableSink extends OutputFormatTableSink<Row> implements Partiti
 	public HiveTableSink(JobConf jobConf, ObjectPath tablePath, CatalogTable table) {
 		this.jobConf = jobConf;
 		this.tablePath = tablePath;
-		this.catalogTable = table;
+
+		this.catalogTable = HiveTableUtil.toHiveCatalogTable(table);
+
 		hiveVersion = Preconditions.checkNotNull(jobConf.get(HiveCatalogValidator.CATALOG_HIVE_VERSION),
 				"Hive version is not defined");
-		tableSchema = table.getSchema();
+
+		tableSchema = catalogTable.getSchema();
 	}
 
 	@Override
--- a/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/connectors/hive/HiveTableSink.java
+++ b/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/connectors/hive/HiveTableSink.java
@@ -76,7 +76,7 @@ public class HiveTableSink extends OutputFormatTableSink<Row> implements Partiti
 		this.jobConf = jobConf;
 		this.tablePath = tablePath;
 
-		this.catalogTable = HiveTableUtil.toHiveCatalogTable(table);
+		this.catalogTable = HiveTableUtil.convertTableSchemaForHive(table);
 
 		hiveVersion = Preconditions.checkNotNull(jobConf.get(HiveCatalogValidator.CATALOG_HIVE_VERSION),
 				"Hive version is not defined");
--- a/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/connectors/hive/HiveTableSource.java
+++ b/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/connectors/hive/HiveTableSource.java
@@ -73,7 +73,7 @@ public class HiveTableSource extends InputFormatTableSource<Row> implements Part
 		this.tablePath = Preconditions.checkNotNull(tablePath);
 
 		Preconditions.checkNotNull(catalogTable);
-		this.catalogTable = HiveTableUtil.toHiveCatalogTable(catalogTable);
+		this.catalogTable = HiveTableUtil.convertTableSchemaForHive(catalogTable);
 
 		this.hiveVersion = Preconditions.checkNotNull(jobConf.get(HiveCatalogValidator.CATALOG_HIVE_VERSION),
 				"Hive version is not defined");
--- a/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/connectors/hive/HiveTableSource.java
+++ b/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/connectors/hive/HiveTableSource.java
@@ -27,6 +27,7 @@ import org.apache.flink.table.catalog.ObjectPath;
 import org.apache.flink.table.catalog.hive.client.HiveMetastoreClientFactory;
 import org.apache.flink.table.catalog.hive.client.HiveMetastoreClientWrapper;
 import org.apache.flink.table.catalog.hive.descriptors.HiveCatalogValidator;
+import org.apache.flink.table.catalog.hive.util.HiveTableUtil;
 import org.apache.flink.table.sources.InputFormatTableSource;
 import org.apache.flink.table.sources.PartitionableTableSource;
 import org.apache.flink.table.sources.TableSource;
@@ -70,7 +71,10 @@ public class HiveTableSource extends InputFormatTableSource<Row> implements Part
 	public HiveTableSource(JobConf jobConf, ObjectPath tablePath, CatalogTable catalogTable) {
 		this.jobConf = Preconditions.checkNotNull(jobConf);
 		this.tablePath = Preconditions.checkNotNull(tablePath);
-		this.catalogTable = Preconditions.checkNotNull(catalogTable);
+
+		Preconditions.checkNotNull(catalogTable);
+		this.catalogTable = HiveTableUtil.toHiveCatalogTable(catalogTable);
+
 		this.hiveVersion = Preconditions.checkNotNull(jobConf.get(HiveCatalogValidator.CATALOG_HIVE_VERSION),
 				"Hive version is not defined");
 		initAllPartitions = false;
--- a/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/util/HiveTableUtil.java
+++ b/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/util/HiveTableUtil.java
@@ -19,7 +19,11 @@
 package org.apache.flink.table.catalog.hive.util;
 
 import org.apache.flink.table.api.TableSchema;
+import org.apache.flink.table.catalog.CatalogTable;
+import org.apache.flink.table.catalog.CatalogTableImpl;
 import org.apache.flink.table.types.DataType;
+import org.apache.flink.table.types.utils.LegacyTypeInfoDataTypeConverter;
+import org.apache.flink.util.Preconditions;
 
 import org.apache.commons.lang3.StringUtils;
 import org.apache.hadoop.hive.metastore.api.FieldSchema;
@@ -30,7 +34,11 @@ import org.apache.hadoop.hive.serde.serdeConstants;
 import org.apache.hadoop.hive.serde2.SerDeUtils;
 import org.apache.hadoop.hive.serde2.typeinfo.TypeInfoUtils;
 
+import java.sql.Date;
+import java.sql.Time;
+import java.sql.Timestamp;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
@@ -133,4 +141,48 @@ public class HiveTableUtil {
 		return partition;
 	}
 
+	/**
+	 * Create a new catalog table according to the given catalog table.
+	 * The two catalog tables are the same except for their table schemas.
+	 * DataTypes not supported by hive will be changed to the types that hive supports.
+	 *
+	 * <p>TODO:
+	 * Can we just change the table schema of the given catalog table
+	 * without creating a new catalog table?
+	 */
+	public static CatalogTable toHiveCatalogTable(CatalogTable oldTable) {
+		TableSchema oldSchema = oldTable.getSchema();
+		DataType[] fieldDataTypes = oldSchema.getFieldDataTypes();
+		String[] fieldNames = oldSchema.getFieldNames();
+
+		TableSchema.Builder builder = TableSchema.builder();
+		for (int i = 0; i < fieldDataTypes.length; i++) {
+			// Hive stores date and timestamp using java.sql.*,
+			// so we change our date / time / timestamp type here.
+			switch (fieldDataTypes[i].getLogicalType().getTypeRoot()) {
+				case DATE:
+					builder.field(fieldNames[i], fieldDataTypes[i].bridgedTo(Date.class));
+					break;
+				case TIME_WITHOUT_TIME_ZONE:
+					builder.field(fieldNames[i], fieldDataTypes[i].bridgedTo(Time.class));
+					break;
+				case TIMESTAMP_WITHOUT_TIME_ZONE:
+					Preconditions.checkArgument(
+						LegacyTypeInfoDataTypeConverter.canConvertToTimestampTypeInfoLenient(fieldDataTypes[i]),
+						"Flink only supports timestamp with precision not larger than 3 internally");
+					builder.field(fieldNames[i], fieldDataTypes[i].bridgedTo(Timestamp.class));
+					break;
+				default:
+					builder.field(fieldNames[i], fieldDataTypes[i]);
+					break;
+			}
+		}
+		TableSchema newSchema = builder.build();
+
+		return new CatalogTableImpl(
+			newSchema,
+			new ArrayList<>(oldTable.getPartitionKeys()),
+			new HashMap<>(oldTable.getProperties()),
+			oldTable.getComment());
+	}
 }
--- a/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/util/HiveTableUtil.java
+++ b/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/util/HiveTableUtil.java
@@ -150,7 +150,7 @@ public class HiveTableUtil {
 	 * Can we just change the table schema of the given catalog table
 	 * without creating a new catalog table?
 	 */
-	public static CatalogTable toHiveCatalogTable(CatalogTable oldTable) {
+	public static CatalogTable convertTableSchemaForHive(CatalogTable oldTable) {
 		TableSchema oldSchema = oldTable.getSchema();
 		DataType[] fieldDataTypes = oldSchema.getFieldDataTypes();
 		String[] fieldNames = oldSchema.getFieldNames();
--- a/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/util/HiveTypeUtil.java
+++ b/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/util/HiveTypeUtil.java
@@ -55,6 +55,7 @@ import org.apache.hadoop.hive.serde2.typeinfo.TypeInfoFactory;
 import org.apache.hadoop.hive.serde2.typeinfo.TypeInfoUtils;
 import org.apache.hadoop.hive.serde2.typeinfo.VarcharTypeInfo;
 
+import java.sql.Date;
 import java.util.ArrayList;
 import java.util.List;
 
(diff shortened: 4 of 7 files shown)


zjuwangg:
cc @xuefuz @lirui-apache

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/connectors/hive/HiveTableSource.java
```diff
@@ -70,7 +72,11 @@
 	public HiveTableSource(JobConf jobConf, ObjectPath tablePath, CatalogTable catalogTable) {
 		this.jobConf = Preconditions.checkNotNull(jobConf);
 		this.tablePath = Preconditions.checkNotNull(tablePath);
-		this.catalogTable = Preconditions.checkNotNull(catalogTable);
+
+		Preconditions.checkArgument(catalogTable instanceof CatalogTableImpl);
```
zjuwangg: Why `catalogTable` must be `CatalogTableImpl ` type?
If it is a must, we may change the constructor.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/connectors/hive/HiveTableSource.java
```diff
@@ -70,7 +72,11 @@
...
+		Preconditions.checkNotNull(catalogTable);
```
zjuwangg: `checkNotNull` should precede `checkArgument`.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/util/HiveTableUtil.java
```diff
@@ -133,4 +140,34 @@ public static Partition createHivePartition(String dbName, String tableName, Lis
 		return partition;
 	}
 
+	public static CatalogTable toHiveCatalogTable(CatalogTable oldTable) {
```
zjuwangg: We may need to add comments to explain the method.

flink-connectors/flink-connector-hive/src/test/java/org/apache/flink/connectors/hive/HiveFullTest.java
```diff
@@ -0,0 +1,139 @@
...
+import java.util.HashMap;
+import java.util.List;
+
+import static org.junit.Assert.assertEquals;
+
+/**
+ * Tests using both {@link HiveTableSource} and {@link HiveTableSink}.
+ */
+@RunWith(FlinkStandaloneHiveRunner.class)
+public class HiveFullTest {
```
zjuwangg: Rename HiveFullTest to HiveTimeRealtedTypeTest or sth else?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/connectors/hive/HiveTableSink.java
```diff
@@ -74,10 +76,13 @@
 	public HiveTableSink(JobConf jobConf, ObjectPath tablePath, CatalogTable table) {
 		this.jobConf = jobConf;
 		this.tablePath = tablePath;
-		this.catalogTable = table;
+
+		Preconditions.checkArgument(table instanceof CatalogTableImpl);
+		this.catalogTable = HiveTableUtil.toHiveCatalogTable(table);
```
xuefuz: It seems that all we  need is to change some field types to their bridged ones. This can be done by building a new TableSchema from the original TableSchema. In other word, we don't need to create a new CatalogTable instance.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/util/HiveTableUtil.java
xuefuz: +1, I think we should just convert the old table's TableSchema to a new TableSchema by changing the corresponding field types. This does more than what we need.

flink-connectors/flink-connector-hive/src/test/java/org/apache/flink/connectors/hive/HiveFullTest.java
lirui-apache: We have `TableEnvHiveConnectorTest` to accommodate general tests where source and sink are both needed. I guess we can either move this test into `TableEnvHiveConnectorTest`, or give it a more specific name as @zjuwangg suggested.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/connectors/hive/HiveTableSource.java
TsReaper: `HiveTableSource` is only initialized in `HiveTableFactory::createInputFormatTableSource`, which is only called from `HiveTableFactory::createTableSource`. This method guarantees that `table` is `CatalogTableImpl`.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/connectors/hive/HiveTableSink.java
```diff
@@ -74,10 +76,13 @@
 	public HiveTableSink(JobConf jobConf, ObjectPath tablePath, CatalogTable table) {
 		this.jobConf = jobConf;
 		this.tablePath = tablePath;
-		this.catalogTable = table;
+
+		Preconditions.checkArgument(table instanceof CatalogTableImpl);
```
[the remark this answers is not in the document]
TsReaper: `HiveTableSink` is only initialized in `HiveTableFactory::createOutputFromTableSink`, which is only called from `HiveTableFactory::createOutputFormatTableSink`. This method guarantees that `table` is a `CatalogTableImpl`.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/util/HiveTableUtil.java
TsReaper: You're right. But `CatalogTableImpl` currently does not have `setSchema` method. Should I add one?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/connectors/hive/HiveTableSink.java
TsReaper: You're right. But `CatalogTableImpl` currently does not have setSchema method. Should I add one?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/util/HiveTypeUtil.java
```diff
@@ -256,9 +258,9 @@ private static DataType toFlinkPrimitiveType(PrimitiveTypeInfo hiveType) {
 			case DOUBLE:
 				return DataTypes.DOUBLE();
 			case DATE:
-				return DataTypes.DATE();
+				return DataTypes.DATE().bridgedTo(Date.class);
 			case TIMESTAMP:
-				return DataTypes.TIMESTAMP();
+				return DataTypes.TIMESTAMP(3).bridgedTo(Timestamp.class);
```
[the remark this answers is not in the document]
TsReaper: `java.sql.Timestamp` record time in milliseconds, which is 10^{-3} seconds.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/connectors/hive/HiveTableSource.java
zjuwangg: Though this method guarantees that `table` is `CatalogTableImpl`, it is not a must.
It's better using `CatalogTable` here.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/util/HiveTypeUtil.java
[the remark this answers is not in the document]
lirui-apache: IIRC, `java.sql.Timestamp` has a separate field to store nanoseconds and supports precision 9. E.g. you can create a Timestamp like `Timestamp.valueOf('2019-08-06 14:00:00.123456789')`

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/util/HiveTableUtil.java
lirui-apache: Probably another way is to create a converted table schema and return it in `getTableSchema` method?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/util/HiveTypeUtil.java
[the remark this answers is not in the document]
TsReaper: Flink previously stores timestamp using a long variable, so only a timestamp of precision 3 can be converted to the `SqlTimeInfo.TIMESTAMP_INFO`.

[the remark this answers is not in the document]
lirui-apache: If that's the case, please add some comments here to explain why the precision has to be 3. Besides, does this mean we'll lose precision when reading a timestamp from Hive into Flink?

flink-connectors/flink-connector-hive/src/test/java/org/apache/flink/connectors/hive/TableEnvHiveConnectorTest.java
```diff
@@ -138,6 +158,71 @@ private void readWriteFormat(String format) throws Exception {
 		hiveShell.execute("drop database db1 cascade");
 	}
 
+	@Test
+	public void testHiveSourceSink() throws Exception {
```
zjuwangg: change method name or add comments to indicate test content of this method.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/util/HiveTableUtil.java
TsReaper: I think the current modification is the most proper modification under the current code base. The table schema in the catalog table is not only used in table source, but will also be passed into hive input format will be used there. So I think the most proper way is to modify the table schema here.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/util/HiveTypeUtil.java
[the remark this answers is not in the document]
TsReaper: Yes, precision will be lost. See `TimestampConverter` in `DataTypeConverters`.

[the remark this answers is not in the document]
xuefuz: @twalthr Do you have any comment on this loss of precision in conversion? Thanks.

flink-connectors/flink-connector-hive/src/test/java/org/apache/flink/connectors/hive/TableEnvHiveConnectorTest.java
```diff
@@ -138,6 +158,74 @@ private void readWriteFormat(String format) throws Exception {
...
+		// this test is to check whether data can successfully
+		// flow from hive source into flink system and then into hive sink.
+		// we especially want to test data types about time here.
```
zjuwangg: Test comments should appeare before the method definition.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/util/HiveTypeUtil.java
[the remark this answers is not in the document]
twalthr: Precision from Hive should be forwarded to Flink. It should be the responsibility of the planner to throw an exception if it does not support the given precision of a source.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/util/HiveTableUtil.java
```diff
@@ -133,4 +140,46 @@ public static Partition createHivePartition(String dbName, String tableName, Lis
...
+		TableSchema.Builder builder = TableSchema.builder();
+		for (int i = 0; i < fieldDataTypes.length; i++) {
+			// Hive stores date and timestamp using java.sql.*,
+			// so we change our date / time / timestamp type here.
+			switch (fieldDataTypes[i].getLogicalType().getTypeRoot()) {
+				case DATE:
+					builder.field(fieldNames[i], DataTypes.DATE().bridgedTo(Date.class));
+					break;
+				case TIME_WITHOUT_TIME_ZONE:
+					builder.field(fieldNames[i], DataTypes.TIME().bridgedTo(Time.class));
```
JingsongLi: You should invoke `fieldDataTypes[i].bridgedTo` to keep nullable and precision.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/util/HiveTypeUtil.java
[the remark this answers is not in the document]
JingsongLi: +1 to keep precision and throw exception.
Indeed there are somewhere to loss precision because of conversion between TypeInformation and DataType. But we do our best to use DataType and keep precision. And we just convert the timestamp with precision smaller than 3. (`LegacyTypeInfoDataTypeConverter.canConvertToTimestampTypeInfoLenient`)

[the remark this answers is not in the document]
lirui-apache: Hive's timestamp has precision 9. If Flink internally stores `java.sql.Timestamp` as a long (like @TsReaper mentioned in `TimestampConverter`), seems we cannot avoid the loss. Does Flink support precision 9 with `java.time.LocalDateTime`?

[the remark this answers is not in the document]
JingsongLi: Only support precision smaller(or equal) than 3.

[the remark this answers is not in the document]
TsReaper: Thanks for the discussion on the precision issue. As the default precision in hive is 9, the connector will be literally not functional if we directly throws an exception. So I think it will be better to log a warning message and add in the connector's document that timestamp precision may be lost.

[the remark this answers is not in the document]
lirui-apache: I agree it's better to warn and document this limitation than totally ban timestamp from Hive connector. And in the long run, shall we support timestamp with higher precision in Flink? Since both `java.sql.Timestamp` and `java.time.LocalDateTime` support up to precision 9, I don't see a good reason why Flink has to limit precision to 3.

TsReaper:
After some discussions, we decide not to support timestamp of higher precision currently (an exception will be thrown if a higher precision timestamp is provided). I've removed timestamp type from a few test cases so that they can currently pass. We should add them back once timestamp type is supported for hive connectors.

zjuwangg:
@TsReaper you need to rebase the master because of confilicts

TsReaper:
Travis passed: https://travis-ci.com/TsReaper/flink/builds/123144021

zjuwangg:
@xuefuz @lirui-apache @bowenli86 
Please have a look and we'd better merge this before flink 1.9 release.

twalthr:
@zjuwangg Flink 1.9 is close to being release. @bowenli86 please watch out what you merge into the `release-1.9` branch as it might break user programs when people upgrade to a minor version release.

bowenli86:
> @zjuwangg Flink 1.9 is close to being released. @bowenli86 please watch out what you merge into the `release-1.9` branch as it might break user programs when people upgrade to a minor version release.

@twalthr can you elaborate more? do you mean this PR can potentially break upgrade from 1.9.0 to 1.9.1 if we merge it into 1.9.1?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/util/HiveTableUtil.java
lirui-apache: I think this method can be renamed to something like `convertTableSchemaForHive`?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/util/HiveTableUtil.java
```diff
@@ -133,4 +141,48 @@ public static Partition createHivePartition(String dbName, String tableName, Lis
...
+			// so we change our date / time / timestamp type here.
+			switch (fieldDataTypes[i].getLogicalType().getTypeRoot()) {
+				case DATE:
+					builder.field(fieldNames[i], fieldDataTypes[i].bridgedTo(Date.class));
+					break;
+				case TIME_WITHOUT_TIME_ZONE:
+					builder.field(fieldNames[i], fieldDataTypes[i].bridgedTo(Time.class));
+					break;
+				case TIMESTAMP_WITHOUT_TIME_ZONE:
+					Preconditions.checkArgument(
```
lirui-apache: Why not just error out here since we don't support timestamp for now?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/util/HiveTypeUtil.java
```diff
@@ -165,9 +166,10 @@ private static DataType toFlinkPrimitiveType(PrimitiveTypeInfo hiveType) {
 			case DOUBLE:
 				return DataTypes.DOUBLE();
 			case DATE:
-				return DataTypes.DATE();
+				return DataTypes.DATE().bridgedTo(Date.class);
 			case TIMESTAMP:
-				return DataTypes.TIMESTAMP();
+				throw new UnsupportedOperationException(
```
lirui-apache: Don't we need to throw exception for timestamp in `HiveTypeUtil.toHiveTypeInfo` too?

flink-connectors/flink-connector-hive/src/test/java/org/apache/flink/connectors/hive/TableEnvHiveConnectorTest.java
```diff
@@ -165,6 +184,77 @@ public void testDecimal() throws Exception {
...
+		TableSchema schema = TableSchema.builder()
+			.field("a", DataTypes.INT())
+			.field("b", DataTypes.STRING())
+			.field("c", DataTypes.DOUBLE())
+			.field("d", DataTypes.DATE())
+			.build();
+
+		ObjectPath sourcePath = new ObjectPath("full_test_db", "input_tbl");
+		CatalogTable sourceTable1 = (CatalogTable) hiveCatalog.getTable(sourcePath);
+		HiveTableSource hiveTableSource1 = new HiveTableSource(new JobConf(hiveConf), sourcePath, sourceTable1);
```
lirui-apache: I don't think we still need to manually create source and sink in the tests. Similar to other test cases in this class, we can just use sql.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/util/HiveTableUtil.java
TsReaper: We actually do not support timestamp with precision larger than 3. We can support timestamp with precision not larger than 3.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/util/HiveTypeUtil.java
TsReaper: No. Only when converting hive timestamp to flink timestamp may we suffer from precision loss.

lirui-apache: Right. But it doesn't seem very useful if the conversion is only a one-way trip. For example, users are allowed to insert some low-precision timestamps into a Hive table. But they cannot read these timestamps back because we don't allow converting Hive timestamp to Flink timestamp, right?

twalthr:
@bowenli86 I haven't looked into the details of what this PR does. I just wanted to add this note because it was mentioned that it might find its way into the release-1.9 branch.

bowenli86:
Hi @TsReaper @lirui-apache , can you sync up quickly offline to come to a resolution?

As Timo mentioned, I'm afraid we may not be able to merge it into 1.9 branch anymore once 1.9.0 is officially released.

TsReaper:
From my perspective, as flink 1.9 is close to release, I think it might be more proper to move this pr into 1.10, and we should support high precision timestamp in flink so the precision problems will not occur.

TsReaper:
Closing as its mostly duplicated by #10401

PR #9354, opened 2019-08-05
[FLINK-13568][sql-parser] DDL create table doesn't allow STRING data …

danny0405:

After FLINK-13335, we have supported full data types in SQL CREATE TABLE, but the `SqlCreateTable` validation breaks this, this patch remove the `SqlColumnType` because we would check the data types in the `SqlValidator`.

## Brief change log

  - Remove the `SqlColumnType`
  - Change the tests of DDL varchar type to string

## Verifying this change

See tests in CatalogTableITCase

  - The serializers: (no)
  - Anything that affects deployment or recovery: JobManager (no),
  - Checkpointing, Yarn/Mesos, ZooKeeper: (no)

## Documentation


flink-table/flink-sql-parser/src/test/java/org/apache/flink/sql/parser/FlinkSqlParserImplTest.java
```diff
@@ -365,18 +365,18 @@ public void testCreateTableWithWatermarkStrategyFromSource() {
 	public void testCreateTableWithComplexType() {
 		check("CREATE TABLE tbl1 (\n" +
```
twalthr: I agree with Xuefu. All we need is actually one test `testCreateTableWithAllTypes` we could improve `testCreateTableWithComplexType` for that.

wuchong:
Hi @danny0405 , don't we support VARCHAR in DDL currently? I would suggest to keep some VARCHAR tests, the same as VARBINARY vs BYTES.

wuchong:
Seems that people have the same question before I submit the comment.

flink-table/flink-sql-parser/src/test/java/org/apache/flink/sql/parser/FlinkSqlParserImplTest.java
danny0405: The error happens during the validation phrase, not sql parsing, in `FlinkDDLDataTypesTest` we already test all the data types in the DDL. But i think add full types in `SqlToOperationTest` is more proper way to go.

danny0405:
@wuchong Thanks for the review, i have updated based on your review comment address.

flink-table/flink-table-planner-blink/src/test/java/org/apache/flink/table/sqlexec/SqlToOperationConverterTest.java
```diff
@@ -157,6 +158,73 @@ public void testSqlInsertWithStaticPartition() {
...
+			"  f52 ROW(f0 INT NOT NULL, f1 BOOLEAN),\n" +
+			"  f53 ROW<`f0` INT>,\n" +
+			"  f54 ROW(`f0` INT),\n" +
+			"  f55 ROW<>,\n" +
+			"  f56 ROW(),\n" +
+			"  f57 ROW<f0 INT NOT NULL 'This is a comment.', f1 BOOLEAN 'This as well.'>)";
+		final FlinkPlannerImpl planner = getPlannerBySqlDialect(SqlDialect.DEFAULT);
+		SqlNode node = planner.parse(sql);
+		assert node instanceof SqlCreateTable;
+		SqlToOperationConverter.convert(planner, node);
```
twalthr: Verify the results that you get from calling this method?

danny0405: I think we can, thanks.

danny0405: @twalthr I have added the tests, but found many problems, i have fired a new issue https://issues.apache.org/jira/browse/FLINK-13604 for them.

flink-table/flink-table-planner-blink/src/test/java/org/apache/flink/table/sqlexec/SqlToOperationConverterTest.java
```diff
@@ -157,6 +158,143 @@ public void testSqlInsertWithStaticPartition() {
...
+			"  f48 INT NOT NULL MULTISET,\n" +
+			"  f49 INT MULTISET NOT NULL,\n" +
+			"  f50 MAP<BIGINT, BOOLEAN>,\n" +
+			"  f51 ROW<f0 INT NOT NULL, f1 BOOLEAN>,\n" +
+			"  f52 ROW(f0 INT NOT NULL, f1 BOOLEAN),\n" +
+			"  f53 ROW<`f0` INT>,\n" +
+			"  f54 ROW(`f0` INT),\n" +
+			"  f55 ROW<>,\n" +
+			"  f56 ROW(),\n" +
+			"  f57 ROW<f0 INT NOT NULL 'This is a comment.', f1 BOOLEAN 'This as well.'>)";
```
wuchong: Add fields to test nested Row and Row in Array/Multiset/Map ?

flink-table/flink-table-planner/src/test/java/org/apache/flink/table/sqlexec/SqlToOperationConverterTest.java
```diff
@@ -163,6 +169,229 @@ public void testSqlInsertWithStaticPartition() {
 		assertEquals(expectedStaticPartitions, sinkModifyOperation.getStaticPartitions());
 	}
 
+	@Test(expected = AssertionError.class) // TODO: tweak the tests when FLINK-13604 is fixed.
```
wuchong: Enable this test and comment the unsupported types?

flink-table/flink-table-planner/src/test/java/org/apache/flink/table/sqlexec/SqlToOperationConverterTest.java
```diff
@@ -163,6 +169,229 @@ public void testSqlInsertWithStaticPartition() {
...
+		final FlinkPlannerImpl planner = getPlannerBySqlDialect(SqlDialect.DEFAULT);
+		SqlNode node = planner.parse(sql);
+		assert node instanceof SqlCreateTable;
+		SqlToOperationConverter.convert(planner, node);
+	}
+
+	@Test
+	public void testCreateTableWithUnSupportedDataType4() {
+		expectedEx.expect(TableException.class);
+		expectedEx.expectMessage("Type is not supported: VARBINARY");
```
wuchong: Why the exception message is not "BYTES"?

flink-table/flink-table-planner-blink/src/test/java/org/apache/flink/table/sqlexec/SqlToOperationConverterTest.java
danny0405: These tests are synced from `LogicalParserTest`, i think nested Row and Row in Array/Multiset/Map are too complex and there is no much value to do this, ignored.

flink-table/flink-table-planner/src/test/java/org/apache/flink/table/sqlexec/SqlToOperationConverterTest.java
danny0405: Agree, thanks ~

flink-table/flink-table-planner/src/test/java/org/apache/flink/table/sqlexec/SqlToOperationConverterTest.java
danny0405: I think it is because `BYTES` is recognized as `VARBINARY` internally, so there is no way to deduce which one the `VARBINARY` comes from.

twalthr: if the exception is generated with `LogicalType#asSummaryString` `BYTES` should be the representation

flink-table/flink-sql-parser/src/main/java/org/apache/flink/sql/parser/ddl/SqlCreateTable.java
```diff
@@ -136,10 +136,12 @@ public void validate() throws SqlParseException {
 					SqlTableColumn tableColumn = (SqlTableColumn) column;
 					columnName = tableColumn.getName().getSimple();
 					String typeName = tableColumn.getType().getTypeName().getSimple();
-					if (SqlColumnType.getType(typeName).isUnsupported()) {
+					if (typeName == null) {
 						throw new SqlParseException(
 							column.getParserPosition(),
-							"Not support type [" + typeName + "], at " + column.getParserPosition());
+							"Non supported type for column ["
```
twalthr: nit: `Unsupported type for column`

flink-table/flink-table-planner-blink/src/test/java/org/apache/flink/table/sqlexec/SqlToOperationConverterTest.java
twalthr: @danny0405  I think Jark's suggestion is reasonable. "too complex" is not an argument for not testing a behavior. Real world programs have way more complex structures.

flink-table/flink-table-planner-blink/src/test/java/org/apache/flink/table/sqlexec/SqlToOperationConverterTest.java
```diff
@@ -157,6 +158,143 @@ public void testSqlInsertWithStaticPartition() {
...
+		final FlinkPlannerImpl planner = getPlannerBySqlDialect(SqlDialect.DEFAULT);
+		SqlNode node = planner.parse(sql);
+		assert node instanceof SqlCreateTable;
+		Operation operation = SqlToOperationConverter.convert(planner, node);
+		assert operation instanceof CreateTableOperation;
+		TableSchema schema = ((CreateTableOperation) operation).getCatalogTable().getSchema();
+		assertArrayEquals(new DataType[] {DataTypes.CHAR(1),
```
twalthr: Some feedback about testability and readability of code. It would be nicer to introduce a test spec here like `TestSpec(String, DataType)` such as `TestSpec("CHAR(1) NOT NULL", DataTypes.CHAR(1).notNull())`. Such that input and expected values are next to each other. We can still generate one big SQL DDL statement out of this test declaration.

flink-table/flink-table-planner/src/test/java/org/apache/flink/table/sqlexec/SqlToOperationConverterTest.java
```diff
@@ -163,6 +169,235 @@ public void testSqlInsertWithStaticPartition() {
...
+		final String sql = "create table t1(\n" +
+			"  f10 VARBINARY)";
+		final FlinkPlannerImpl planner = getPlannerBySqlDialect(SqlDialect.DEFAULT);
+		SqlNode node = planner.parse(sql);
+		assert node instanceof SqlCreateTable;
+		SqlToOperationConverter.convert(planner, node);
+	}
+
+	@Test
+	public void testCreateTableWithUnSupportedDataType7() {
```
twalthr: Some feedback around readability of tests. We should avoid numbering tests and instead give them useful names e.g. instead of `testCreateTableWithUnSupportedDataType7` why not `testCreateTableUnsupportedBinary`? And same concern as before, a test spec would reduce code duplication. Take `org.apache.flink.table.types.LogicalTypeParserTest` as a reference for this class and future tests.

flink-table/flink-table-planner-blink/src/test/java/org/apache/flink/table/sqlexec/SqlToOperationConverterTest.java
danny0405: Sure, let's add in the nested structured type.

flink-table/flink-table-planner/src/test/java/org/apache/flink/table/sqlexec/SqlToOperationConverterTest.java
danny0405: No, the exception comes from Flink.

danny0405:
@twalthr I have updated based on your review comments, please review if you have time.

wuchong:
Thanks @twalthr and @danny0405 , I will merge this.

PR #9355, opened 2019-08-05
[FLINK-13577][ml] Add an util class to build result row and generate …

xuyang1706:

Add an util class to build result row and generate result schema, which will be used in later algorithm implementations.

## Brief change log

  - *Add class OutputColsHelper.*
  - *Add unit test*
 

## Verifying this change

- run test case pass

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/OutputColsHelper.java
+++ b/flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/OutputColsHelper.java
@@ -31,7 +31,7 @@ import java.util.Arrays;
 import java.util.HashSet;
 
 /**
- * Util for generating output schema when doing prediction or transformation.
+ * Utils for merging input data with output data.
  *
  * <p>Input:
  * 1) Schema of input data being predicted or transformed.
@@ -48,83 +48,93 @@ import java.util.HashSet;
  * 2）The reserved columns are arranged ahead of the operator's output columns in the final output.
  * 3) If some of the reserved column names overlap with those of operator's output columns, then the operator's
  * output columns override the conflicting reserved columns.
+ * 4) The reserved columns in the result table preserve their orders as in the input table.
  *
  * <p>For example, if we have input data schema of ["id":INT, "f1":FLOAT, "f2":DOUBLE], and the operator outputs
- * a column "label" with type STRING, and we want to preserve the column "id", then we get the output
+ * a column "label" with type STRING, and we want to preserve the column "id", then we get the result
  * schema of ["id":INT, "label":STRING].
  */
 public class OutputColsHelper implements Serializable {
-	private transient String[] dataColNames;
-	private transient TypeInformation[] dataColTypes;
-	private transient String[] outputColNames;
-	private transient TypeInformation[] outputColTypes;
+	private String[] inputColNames;
+	private TypeInformation[] inputColTypes;
+	private String[] outputColNames;
+	private TypeInformation[] outputColTypes;
 
-	private int resultLength;
-	private int[] reservedColIndices;
-	private int[] reservedToResultIndices;
-	private int[] outputToResultIndices;
+	/**
+	 * Column indices in the input data that would be forward to the result.
+	 */
+	private int[] reservedCols;
 
-	public OutputColsHelper(TableSchema dataSchema, String outputColName, TypeInformation outputColType) {
-		this(dataSchema, outputColName, outputColType, null);
+	/**
+	 * The positions of reserved columns in the result.
+	 */
+	private int[] reservedColsPosInResult;
+
+	/**
+	 * The positions of output columns in the result.
+	 */
+	private int[] outputColsPosInResult;
+
+	public OutputColsHelper(TableSchema inputSchema, String outputColName, TypeInformation outputColType) {
+		this(inputSchema, outputColName, outputColType, inputSchema.getFieldNames());
 	}
 
-	public OutputColsHelper(TableSchema dataSchema, String outputColName, TypeInformation outputColType,
+	public OutputColsHelper(TableSchema inputSchema, String outputColName, TypeInformation outputColType,
 							String[] reservedColNames) {
-		this(dataSchema, new String[] {outputColName}, new TypeInformation[] {outputColType}, reservedColNames);
+		this(inputSchema, new String[]{outputColName}, new TypeInformation[]{outputColType}, reservedColNames);
 	}
 
-	public OutputColsHelper(TableSchema dataSchema, String[] outputColNames, TypeInformation[] outputColTypes) {
-		this(dataSchema, outputColNames, outputColTypes, null);
+	public OutputColsHelper(TableSchema inputSchema, String[] outputColNames, TypeInformation[] outputColTypes) {
+		this(inputSchema, outputColNames, outputColTypes, inputSchema.getFieldNames());
 	}
 
 	/**
 	 * The constructor.
 	 *
-	 * @param dataSchema       Schema of input data being predicted or transformed.
+	 * @param inputSchema      Schema of input data being predicted or transformed.
 	 * @param outputColNames   Output column names of the prediction/transformation operator.
 	 * @param outputColTypes   Output column types of the prediction/transformation operator.
 	 * @param reservedColNames Reserved column names, which is a subset of input data's column names that we want to preserve.
 	 */
-	public OutputColsHelper(TableSchema dataSchema, String[] outputColNames, TypeInformation[] outputColTypes,
+	public OutputColsHelper(TableSchema inputSchema, String[] outputColNames, TypeInformation[] outputColTypes,
 							String[] reservedColNames) {
-		this.dataColNames = dataSchema.getFieldNames();
-		this.dataColTypes = dataSchema.getFieldTypes();
+		this.inputColNames = inputSchema.getFieldNames();
+		this.inputColTypes = inputSchema.getFieldTypes();
 		this.outputColNames = outputColNames;
 		this.outputColTypes = outputColTypes;
 
-		HashSet <String> toReservedCols = new HashSet <>(
-			Arrays.asList(
-				reservedColNames == null ? this.dataColNames : reservedColNames
-			)
+		HashSet<String> toReservedCols = new HashSet<>(
+			Arrays.asList(reservedColNames == null ? this.inputColNames : reservedColNames)
 		);
-
-		ArrayList <Integer> reservedColIndices = new ArrayList <>(toReservedCols.size());
-		ArrayList <Integer> reservedColToResultIndex = new ArrayList <>(toReservedCols.size());
-		outputToResultIndices = new int[outputColNames.length];
-		Arrays.fill(outputToResultIndices, -1);
+		//the indices of the columns which need to be reserved.
+		ArrayList<Integer> reservedColIndices = new ArrayList<>(toReservedCols.size());
+		ArrayList<Integer> reservedColToResultIndex = new ArrayList<>(toReservedCols.size());
+		outputColsPosInResult = new int[outputColNames.length];
+		Arrays.fill(outputColsPosInResult, -1);
 		int index = 0;
-		for (int i = 0; i < dataColNames.length; i++) {
-			int key = ArrayUtils.indexOf(outputColNames, dataColNames[i]);
+		for (int i = 0; i < inputColNames.length; i++) {
+			int key = ArrayUtils.indexOf(outputColNames, inputColNames[i]);
 			if (key >= 0) {
-				outputToResultIndices[key] = index++;
+				outputColsPosInResult[key] = index++;
 				continue;
 			}
-			if (toReservedCols.contains(dataColNames[i])) {
+			//add these interested column.
+			if (toReservedCols.contains(inputColNames[i])) {
 				reservedColIndices.add(i);
 				reservedColToResultIndex.add(index++);
 			}
 		}
-		for (int i = 0; i < outputToResultIndices.length; i++) {
-			if (outputToResultIndices[i] == -1) {
-				outputToResultIndices[i] = index++;
+		for (int i = 0; i < outputColsPosInResult.length; i++) {
+			if (outputColsPosInResult[i] == -1) {
+				outputColsPosInResult[i] = index++;
 			}
 		}
-		this.resultLength = index;
-		this.reservedColIndices = new int[reservedColIndices.size()];
-		this.reservedToResultIndices = new int[reservedColIndices.size()];
-		for (int i = 0; i < this.reservedColIndices.length; i++) {
-			this.reservedColIndices[i] = reservedColIndices.get(i);
-			this.reservedToResultIndices[i] = reservedColToResultIndex.get(i);
+		//write reversed column information in array.
+		this.reservedCols = new int[reservedColIndices.size()];
+		this.reservedColsPosInResult = new int[reservedColIndices.size()];
+		for (int i = 0; i < this.reservedCols.length; i++) {
+			this.reservedCols[i] = reservedColIndices.get(i);
+			this.reservedColsPosInResult[i] = reservedColToResultIndex.get(i);
 		}
 	}
 
@@ -133,12 +143,12 @@ public class OutputColsHelper implements Serializable {
 	 *
 	 * @return the reserved colNames.
 	 */
-	public String[] getReservedColNames() {
-		String[] reservedColNames = new String[reservedColIndices.length];
-		for (int i = 0; i < reservedColIndices.length; i++) {
-			reservedColNames[i] = dataColNames[reservedColIndices[i]];
+	public String[] getReservedColumns() {
+		String[] passThroughColNames = new String[reservedCols.length];
+		for (int i = 0; i < reservedCols.length; i++) {
+			passThroughColNames[i] = inputColNames[reservedCols[i]];
 		}
-		return reservedColNames;
+		return passThroughColNames;
 	}
 
 	/**
@@ -148,38 +158,40 @@ public class OutputColsHelper implements Serializable {
 	 * @return The result table schema.
 	 */
 	public TableSchema getResultSchema() {
+		int resultLength = reservedCols.length + outputColNames.length;
 		String[] resultColNames = new String[resultLength];
 		TypeInformation[] resultColTypes = new TypeInformation[resultLength];
-		for (int i = 0; i < reservedColIndices.length; i++) {
-			resultColNames[reservedToResultIndices[i]] = dataColNames[reservedColIndices[i]];
-			resultColTypes[reservedToResultIndices[i]] = dataColTypes[reservedColIndices[i]];
+		for (int i = 0; i < reservedCols.length; i++) {
+			resultColNames[reservedColsPosInResult[i]] = inputColNames[reservedCols[i]];
+			resultColTypes[reservedColsPosInResult[i]] = inputColTypes[reservedCols[i]];
 		}
-		for (int i = 0; i < outputToResultIndices.length; i++) {
-			resultColNames[outputToResultIndices[i]] = outputColNames[i];
-			resultColTypes[outputToResultIndices[i]] = outputColTypes[i];
+		for (int i = 0; i < outputColsPosInResult.length; i++) {
+			resultColNames[outputColsPosInResult[i]] = outputColNames[i];
+			resultColTypes[outputColsPosInResult[i]] = outputColTypes[i];
 		}
 		return new TableSchema(resultColNames, resultColTypes);
 	}
 
 	/**
-	 * Get the result row.
+	 * Merge the input row and the output row.
 	 *
-	 * @param data   The input row being predicted or transformed.
+	 * @param input  The input row being predicted or transformed.
 	 * @param output The output row of the prediction/transformation operator.
 	 * @return The result row, which is a combination of preserved columns and the operator's
 	 * output columns.
 	 */
-	public Row getResultRow(Row data, Row output) {
-		int numOutputs = outputToResultIndices.length;
+	public Row getResultRow(Row input, Row output) {
+		int numOutputs = outputColsPosInResult.length;
 		if (output.getArity() != numOutputs) {
 			throw new IllegalArgumentException("Invalid output size");
 		}
+		int resultLength = reservedCols.length + outputColNames.length;
 		Row result = new Row(resultLength);
-		for (int i = 0; i < reservedColIndices.length; i++) {
-			result.setField(reservedToResultIndices[i], data.getField(reservedColIndices[i]));
+		for (int i = 0; i < reservedCols.length; i++) {
+			result.setField(reservedColsPosInResult[i], input.getField(reservedCols[i]));
 		}
 		for (int i = 0; i < numOutputs; i++) {
-			result.setField(outputToResultIndices[i], output.getField(i));
+			result.setField(outputColsPosInResult[i], output.getField(i));
 		}
 		return result;
 	}
--- a/flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/OutputColsHelper.java
+++ b/flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/OutputColsHelper.java
@@ -37,13 +37,13 @@ import java.util.HashSet;
  * 1) data schema
  * 2) output column names
  * 3) output column types
- * 4) keep column names
+ * 4) reserved column names
  *
  * <p>The following roles are followed:
- * 1)If reserved columns is null, then reserve all columns from the origin dataSet.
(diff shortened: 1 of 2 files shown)


flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/OutputColsHelper.java
```diff
@@ -0,0 +1,190 @@
...
+import java.util.HashSet;
+
+/**
+ * Util for generating output schema.
+ *
+ * <p>Need following information:
+ * 1) data schema
+ * 2) output column names
+ * 3) output column types
+ * 4) keep column names
```
walterddr: this should either be `reserved column names` or the following java doc should change `reserved column` to `keep column`. 

also if we are using the word `keep` I think it is better to use `kept`?

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/OutputColsHelper.java
```diff
@@ -0,0 +1,190 @@
...
+ *
+ * <p>The following roles are followed:
+ * 1)If reserved columns is null, then reserve all columns from the origin dataSet.
+ * 2)If some of the reserved column names are the same as output column names, then they are
```
walterddr: does this means you select a particular subset of columns to output? (and if null, output all from `outputColumnName` ?

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/OutputColsHelper.java
```diff
@@ -0,0 +1,190 @@
...
+
+	/**
+	 * Get the outputting row, which is a combination of input data and output data.
+	 * The difference with getResultRow is that it's single column output.
+	 *
+	 * @param data   input data
+	 * @param output output data
+	 * @return
+	 */
+	public Row getResultRowSingle(Row data, Object output) {
```
walterddr: do we really need this API? it seems like the following are identical
```
getResultRowSingle(someData, myObj);
getResultRow(someData, Row.of(myObj));
```

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/OutputColsHelper.java
xuyang1706: Yes, user can select a particular subset of columns to output. Algorithm's outputs include 2 parts: the calc results(defined by outputColNames) and the reserved original columns(defined by reservedColNames). If the reservedColNames is null, means reserved all the original input columns.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/OutputColsHelper.java
xuyang1706: Yes, it should be `reserved column names`.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/OutputColsHelper.java
xuyang1706: Thanks, I removed the method getResultRowSingle().

xuyang1706:
> thanks for the contribution @xuyang1706 . This is a relatively small PR. i was wondering if it is better to combine with the PR which actually uses this utility.
> But regarding this contribution, I think overall it looks good. I left one comment regarding the API, and others are mostly questions regarding the JavaDoc and other APIs ( which prompts me to think it might be automatically answered if we combine with the actual contribution that uses the util).
> Please let me know what do you think

@walterddr , thanks for your comments and advice. I refined the code. Some PRs use this util, e.g. #9769 .

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/OutputColsHelper.java
walterddr: hmm in this case the javadoc is confusing to me still (especially with point #3). Let's use an example:
assume the output row is `["a":1, "b":2, "c":true, "d":3.4]`, is my assumptions below correct?
 
there's only one option of the `dataSchema`:
```
dataSchema: ["a":INT, "b":INT, "c":BOOL, "d":FLOAT]
```
2. `outputColumnName` should be a subset of `{"a", "b", "c", "d"}`, but preserves the ordering defined.
3. `outputColumnType` should match exactly the size of `outputColumnName` and should match the exact type at the corresponding index of `dataSchema` - (if point3 is true, why do we need outputColumnType?)
4. `reserveColumnName` should be a subset of `{"a", "b", "c", "d"}`, but ordering follows how `dataSchema` defines.
5. I am assuming you will output the result by first putting `reservedColumn`s then `outputColumn`s. (if point 2,4,5 are true, won't you have duplicate naming in your final output?

xuyang1706: > Hi @xuyang1706, I read the javadoc again and I think I have some confusion on what you are trying to achieve. please kindly take a look at my comments and see if you can provide some examples in test to better clarify? thanks -Rong

Thanks @walterddr, I have refined the javadoc with an example, and hope that it could make things clear. Further more, I’d like to give 3 use cases following your example.

Input data schema : ` ["a":INT, "b":INT, "c":BOOL, “d”:DOUBLE]`
Input data values:
```
{{1,3,true,1.0},
 {2,4,false,3.0},
 {3,5,false,5.0}}
```

Apply this operator:
```
new MinMaxScaler()
	.setSelectedCols("a","b","d")
	.setOutputCols("a_std","b_std","d_std");
```
For operator MinMaxScaler’s result types are DOUBLE, and if not defined reserved Column Names, it will reserve all the original input data cols.
output data schema : ` ["a":INT, "b":INT, "c":BOOL, “d”:DOUBLE, “a_std”:DOUBLE, “b_std”:DOUBLE, “d_std”:DOUBLE]`
output data values:
```
{{1,3,true,1.0, 0.0,0.0,0.0},
 {2,4,false,3.0, 0.5,0.5,0.5},
 {3,5,false,5.0, 1.0,1.0,1.0}}
```

Apply this operator for the origin input data
```
new MinMaxScaler()
	.setSelectedCols("a","b","d");
```

For operator MinMaxScaler’s result types are DOUBLE, and if not defined the output column names, it will use the selected operating columns to save the results, and changed the according column types.
output data schema : ` [“a”:DOUBLE, “b”:DOUBLE, "c":BOOL, “d”:DOUBLE]`
output data values:
```
{{0.0,0.0,true,0.0},
 {0.5,0.5,false,0.5},
 {1.0,1.0,false,1.0}}
```

Apply this operator for the origin input data
```
new LogisticRegressionModel()
	.setFeatureCols("a","b","c","d")
	.setPredictionCol("pred")
	.setReservedCols("a", “d”);
```

The prediction result type is decided by the model, e.g. BOOL type. 
output data schema :  `["a":INT, “d”:DOUBLE, “pred”:BOOL]`
output data values:
```
{{1,1.0,false},
 {2,3.0,true},
 {3,5.0,false}}
```

xuyang1706:
> Hi @xuyang1706, I read the javadoc again and I think I have some confusion on what you are trying to achieve. please kindly take a look at my comments and see if you can provide some examples in test to better clarify? thanks -Rong

Thanks @walterddr, I have refined the javadoc with an example, and hope that it could make things clear. Further more, I’d like to give 3 use cases following your example.

Input data schema : ` ["a":INT, "b":INT, "c":BOOL, “d”:DOUBLE]`
Input data values:
```
{{1,3,true,1.0},
 {2,4,false,3.0},
 {3,5,false,5.0}}
```

Apply this operator:
```
new MinMaxScaler()
	.setSelectedCols("a","b","d")
	.setOutputCols("a_std","b_std","d_std");
```
For operator MinMaxScaler’s result types are DOUBLE, and if not defined reserved Column Names, it will reserve all the original input data cols.
output data schema : ` ["a":INT, "b":INT, "c":BOOL, “d”:DOUBLE, “a_std”:DOUBLE, “b_std”:DOUBLE, “d_std”:DOUBLE]`
output data values:
```
{{1,3,true,1.0, 0.0,0.0,0.0},
 {2,4,false,3.0, 0.5,0.5,0.5},
 {3,5,false,5.0, 1.0,1.0,1.0}}
```

Apply this operator for the origin input data
```
new MinMaxScaler()
	.setSelectedCols("a","b","d");
```

For operator MinMaxScaler’s result types are DOUBLE, and if not defined the output column names, it will use the selected operating columns to save the results, and changed the according column types.
output data schema : ` [“a”:DOUBLE, “b”:DOUBLE, "c":BOOL, “d”:DOUBLE]`
output data values:
```
{{0.0,0.0,true,0.0},
 {0.5,0.5,false,0.5},
 {1.0,1.0,false,1.0}}
```

Apply this operator for the origin input data
```
new LogisticRegressionModel()
	.setFeatureCols("a","b","c","d")
	.setPredictionCol("pred")
	.setReservedCols("a", “d”);
```

The prediction result type is decided by the model, e.g. BOOL type. 
output data schema :  `["a":INT, “d”:DOUBLE, “pred”:BOOL]`
output data values:
```
{{1,1.0,false},
 {2,3.0,true},
 {3,5.0,false}}
```

walterddr:
Thanks for the quick response. @xuyang1706. I think I kinda understand what you were trying to achieve now. i think this PR seems to miss lots of context for someone who could perform a successful review (especially with the additional contents like `setPredictionCol` `setFeatureCol` you mentioned in the previous comment).

How about we merge this one with the first concrete algorithm/model PR together so that it can be better reviewed?

xuyang1706:
> Thanks for the quick response. @xuyang1706. I think I kinda understand what you were trying to achieve now. i think this PR seems to miss lots of context for someone who could perform a successful review (especially with the additional contents like `setPredictionCol` `setFeatureCol` you mentioned in the previous comment).
> 
> How about we merge this one with the first concrete algorithm/model PR together so that it can be better reviewed?

Thanks @walterddr, that is good idea. I have submitted another 2 PRs which use OutputColsHelper. 

[FLINK-13676][ml] Add class of Vector to Columns mapper
https://issues.apache.org/jira/browse/FLINK-13676 
https://github.com/apache/flink/pull/9413

[FLINK-13835][ml] Add class for FeatureHasherMapper.
https://issues.apache.org/jira/browse/FLINK-13835 
https://github.com/apache/flink/pull/9523

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/OutputColsHelper.java
```diff
@@ -0,0 +1,186 @@
...
+ * 1) If reserved columns are not given, then all columns of input data is reserved.
+ * 2）The reserved columns are arranged ahead of the operator's output columns in the final output.
+ * 3) If some of the reserved column names overlap with those of operator's output columns, then the operator's
+ * output columns override the conflicting reserved columns.
+ *
+ * <p>For example, if we have input data schema of ["id":INT, "f1":FLOAT, "f2":DOUBLE], and the operator outputs
+ * a column "label" with type STRING, and we want to preserve the column "id", then we get the output
+ * schema of ["id":INT, "label":STRING].
+ */
+public class OutputColsHelper implements Serializable {
```
walterddr: based on the usage #9413 and #9523 . this can be private to flink, yes? (I do not foresee user directly interacting with this Helper function)

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/OutputColsHelper.java
```diff
@@ -0,0 +1,186 @@
...
+ * 2) Output column names of the prediction/transformation operator.
+ * 3) Output column types of the prediction/transformation operator.
+ * 4) Reserved column names, which is a subset of input data's column names that we want to preserve.
+ *
+ * <p>Output:
+ * 1）The result data schema. The result data is a combination of the preserved columns and the operator's
+ * output columns.
+ *
+ * <p>Several rules are followed:
+ * 1) If reserved columns are not given, then all columns of input data is reserved.
```
walterddr: I think this is actually the one default that confuses me: why if reserved columns are not given then preserves all input data?

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/OutputColsHelper.java
```diff
@@ -0,0 +1,186 @@
...
+ * 2）The reserved columns are arranged ahead of the operator's output columns in the final output.
+ * 3) If some of the reserved column names overlap with those of operator's output columns, then the operator's
```
walterddr: I think we should throw an exception when output column names and the user-defined reserved column names have conflict (ultimately both are defined by user, this shouldn't occur on the users' perspective)

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/OutputColsHelper.java
xuyang1706: Yes, this helper function is for algo developers, not for users.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/OutputColsHelper.java
xuyang1706: In most of the cases, users want to preserve everything during prediction/transformation, so we make it as default.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/OutputColsHelper.java
xuyang1706: I think allowing such conflicts is more handy for users, since by default we are reserving every columns.

xuyang1706:
> Hi @xuyang1706 . After I read through #9413 I think I understood the intent a bit better.
> I think the outputColHelper did 2 things in the same class which made it very hard to review. please correct me if I were wrong.
> 
> 1. It tries provide a simple API to enrich the computed column (predict/transform) results - by allowing users to make some of the input columns as "pass through"
> 2. It tries to allow users to rename some of the columns in the output.
>    As a side effect of achieving both 1&2, It also allow users to reorder the outputs.
> 
> I think it is better to split the intent into 2 different utility classes:
> 
> * `ColumnExtractHelper` which provides the reserve/passthrough functionality.
>   
>   * allows users to define `passthrough` (or `reserve` in your word)
>   * allows users to define the output columns (e.g. the predict/transform result)
> * `ColumnFormatHelper` which provides the renaming/reordering functionality.
>   
>   * takes the column extract helper output as a whole row input (including the passthrough ones and function output ones, but doesn't actually know which is which)
>   * performs the renaming - this step is optional.
> 
> Please let me know if any use cases you had in mind was not covered by this split. thanks --Rong

Thanks for your suggestion @walterddr . The purpose of this Helper is to merge input columns with output columns. Your first point exactly describes what it is doing, and it does no more. It is not the purpose of this Helper to allow users to rename or reorder columns. I have renamed some of the variables to make this code more clear and accessible. Thanks  --Xu

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/OutputColsHelper.java
walterddr: replying to both comments above: the logic IMO assumes too much user behavior: the default reserving all columns in the input does seems convenient. but do you think adding a flag from the input argument (e.g. `reseverAllColumns`) makes more sense?

As a framework I think the responsibility is to strictly track the final result column name uniqueness. By silently drop an input column that users who write a transformation might not have the full knowledge of seems to be an unnecessary responsibility to hand it over to the user.

I would propose:
1. by default reserves no input columns (this is also the standard behavior of a transform in dataset/datastream - makes it consistent with user expectation.
2. create a `inputAsPassthrough` boolean flag when initializing `OutputColsHelper` so that users explicitly have to say "I want all input columns reserved"
3. when `inputAsPassthrough` flag enables; we will take `reservedColumn...` arguments into consideration. 

nit: I prefer the word `passthroughColumn...` over `reserve` makes more sense to me. It would've been nice if some native English speaker can provide more insights on this.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/OutputColsHelper.java
walterddr: Please see my comments below:

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/OutputColsHelper.java
walterddr: I would suggest putting this as part of the Javadoc
--> Unlike Scala Java has no module private modifier like `private[flink]`, but it is still good to point out IMO

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/OutputColsHelper.java
xuyang1706: Thanks for your advice. As far as we know from our practice, in the domain of ML, reserving input columns by default could be a better design. This is more than the framework itself, but also about user experience. For example, data preprocessing, always the first step to ML, often performs transformation on multiple columns one by one. So there may be PilelineStage_1 for feature scaling on column_1, PilelineStage_2 for binarizing on column_2 and more. This would be redundant to force user to add a flag for each of these steps.

xuyang1706:
> Thanks for the explanation @xuyang1706 .
> Now the PR makes much more sense --> especially when the ordering part of the Javadoc/code gets cleared up.
> 
> I have a small comment regarding how the default is being handled, but otherwise the PR looks good to me. Please kindly take a look --Rong

Thanks for your comments and advice @walterddr. I replied the comment in lines. Thanks -Xu

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/OutputColsHelper.java
walterddr: I see your point. I still reserves my opinion. but I think one point you made was very crucial: since this is an internal tooling (e.g. end-user does not directly interact with this class) I am ok with this behavior.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/OutputColsHelper.java
walterddr: In response to my previous comment. I would add a paragraph to make clear in the Javadoc that: end user should not directly interact with this helper class. instead it will be indirectly used via concrete algorithms.

xuyang1706:
> Thanks for the explanation @xuyang1706 . I think I agree to your presumption and general ML use cases. The only comment I have is regarding a java doc to specify that this class is suppose to be used for algorithm development.
> 
> I would do another pass on the javadoc/details before merging soon if there's no more questions regarding the APIs?

Thanks for your comments and advice @walterddr , we updated the JavaDoc.

walterddr:
I've made some final changes to the patch, please see them [here](https://github.com/walterddr/flink/commit/802fa4c6530d90b6322c7be548b0953f17cbf003)
If no more comments I will merge once travis runs green.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/OutputColsHelper.java
```diff
@@ -0,0 +1,200 @@
...
+			this.reservedColsPosInResult[i] = reservedColToResultIndex.get(i);
+		}
+	}
+
+	/**
+	 * Get the reserved columns' names.
+	 *
+	 * @return the reserved colNames.
+	 */
+	public String[] getReservedColumns() {
```
walterddr: API is not tested. I made some changes to the test in https://github.com/walterddr/flink/commit/802fa4c6530d90b6322c7be548b0953f17cbf003

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/OutputColsHelper.java
```diff
@@ -0,0 +1,200 @@
...
+import java.util.HashSet;
+
+/**
+ * Utils for merging input data with output data.
+ *
+ * <p>Input:
+ * 1) Schema of input data being predicted or transformed.
+ * 2) Output column names of the prediction/transformation operator.
+ * 3) Output column types of the prediction/transformation operator.
+ * 4) Reserved column names, which is a subset of input data's column names that we want to preserve.
```
walterddr: nit: list via `<li>`: see https://github.com/walterddr/flink/commit/802fa4c6530d90b6322c7be548b0953f17cbf003

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/OutputColsHelper.java
xuyang1706: Thanks.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/OutputColsHelper.java
xuyang1706: OK.

xuyang1706:
> I've made some final changes to the patch, please see them [here].([walterddr@802fa4c](https://github.com/walterddr/flink/commit/802fa4c6530d90b6322c7be548b0953f17cbf003))
> If no more comments I will merge once travis runs green. @xuyang1706

The patch looks good to me @walterddr. Thanks, -Xu

becketqin:
@walterddr Thanks for the review. Sorry I was quite busy with FLIP-27 and was not able to review the patch. Please feel free to commit the patch as long as you are happy with it :)

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/OutputColsHelper.java
```diff
@@ -0,0 +1,200 @@
...
+ *
+ * <p>For example, if we have input data schema of ["id":INT, "f1":FLOAT, "f2":DOUBLE], and the operator outputs
+ * a column "label" with type STRING, and we want to preserve the column "id", then we get the result
+ * schema of ["id":INT, "label":STRING].
+ *
+ * <p>end user should not directly interact with this helper class. instead it will be indirectly used via concrete algorithms.
+ */
+public class OutputColsHelper implements Serializable {
+	private String[] inputColNames;
+	private TypeInformation[] inputColTypes;
```
walterddr: Also added the type generic here `TypeInformation<?>[]`. see walterddr@802fa4c

flink-ml-parent/flink-ml-lib/src/test/java/org/apache/flink/ml/common/utils/OutputColsHelperTest.java
```diff
@@ -0,0 +1,261 @@
...
+				TypeInformation.of(Long.class),
+				TypeInformation.of(Integer.class),
+				TypeInformation.of(String.class)
+			}
+		);
+		OutputColsHelper helper = new OutputColsHelper(
+			tableSchema, "res",
+			TypeInformation.of(String.class)
+		);
+		Assert.assertEquals(helper.getResultSchema(), expectSchema);
```
walterddr: @xuyang1706 according to the tradition of `junit`, the expected value should be the first in the `assertXXX` calls. could you please fix this? 
See: http://junit.sourceforge.net/javadoc/org/junit/Assert.html

xuyang1706: Thanks, fixed.

xuyang1706:
> Hi @xuyang1706 sorry for not being careful with this earlier - I found a significant issue in the test, could you please take a look? I think this affects some of the other ML patches as well.

Thanks for pointing out this issue @walterddr. I have fixed in this PR, and I will fix this issue in other PRs. Thanks, -Xu

walterddr:
Looks good, thanks for the quick fix @xuyang1706 I will merge this once travis turns green

PR #9361, opened 2019-08-05
[FLINK-13540] DDL do not support properties key that contains number …

danny0405:

This patch changes the table option keys format to string literal, they are as sql identifiers before.

## Brief change log

  - Change the table option key to string literal

## Verifying this change

See tests in FlinkSqlParserImplTest.

  - The public API, i.e., is any changed class annotated with `@Public(Evolving)`:  no
  - The serializers: no

## Documentation


flink-table/flink-sql-parser/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -115,18 +115,18 @@ void UniqueKey(List<SqlNodeList> list) :
     }
 }
 
-SqlNode PropertyValue() :
+SqlNode TableOption() :
 {
-    SqlIdentifier key;
+    SqlNode key;
```
wuchong: Force key and value to be `StringLiteral`?

flink-table/flink-sql-parser/src/main/java/org/apache/flink/sql/parser/TableOption.java
```diff
@@ -0,0 +1,85 @@
...
+
+import java.util.List;
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * Table options of a DDL, a key-value pair
+ * with both key and value as string literal.
+ */
+public class TableOption extends SqlCall {
```
wuchong: Rename it to `SqlTableOption` or `SqlTableProperty` and move to `org.apache.flink.sql.parser.ddl`?

flink-table/flink-sql-parser/src/test/java/org/apache/flink/sql/parser/FlinkSqlParserImplTest.java
```diff
@@ -454,12 +454,36 @@ public void testCreateInvalidPartitionedTable() {
...
+			"  'a-b-c-d124' = 'ab',\n" +
+			"  'a.b-c-connector.e-f.g' = 'ada',\n" +
+			"  'a.b-c-d.e-1231.g' = 'ada',\n" +
+			"  'a.b-c-d.*' = 'adad')\n";
+		String expected = "CREATE TABLE `SOURCE_TABLE` (\n" +
+			"  `A`  INTEGER,\n" +
+			"  `B`  BIGINT,\n" +
+			"  `C`  STRING\n" +
+			") WITH (\n" +
+			"  'a-b-c-d124' = 'ab',\n" +
```
wuchong: nit: could you add a key `a.b.1.c` ?

flink-table/flink-sql-parser/src/main/codegen/includes/parserImpls.ftl
danny0405: The `StringLiteral()`'s return type is `SalNode`, i suggest to keep the type as it is(do not force type cast in parser), we have a type cast in `TableOption`.

flink-table/flink-sql-parser/src/main/java/org/apache/flink/sql/parser/TableOption.java
danny0405: Agree, done

flink-table/flink-sql-parser/src/test/java/org/apache/flink/sql/parser/FlinkSqlParserImplTest.java
danny0405: Added

danny0405:
@wuchong  Thanks for the review, i would rather to keep the `SqlProperty`, because it is a common pattern, we may still need to use this structure for future grammars, i.g. `CLUSTER BY` .

twalthr:
+1 to this from an API perspective.

PR #9364, opened 2019-08-06
[FLINK-13593][checkpointing] Prevent failing the wrong execution attempt in CheckpointFailureManager

carp84:

This PR fixes the issue as reported in FLINK_13593, that due to the asynchronously handling of checkpoint decline message in `LegacyScheduler#declineCheckpoint`, it's possible that the message is handled before job status transition thus `receiveDeclineMessage` grabbed the lock in `CheckpointCoordinator` before `pendingCheckpoints` got cleared by `stopCheckpointScheduler` (as triggered by the job status listener `CheckpointCoordinatorDeActivator`). If the job/tasks restarts quickly enough, the FailJobCallback in CheckpointFailureManager might unexpectedly fail the job again.

## Brief change log

Add a safe guard when failing the job, passing through the `ExecutionAttemptID` and checking against the current executions to make sure the to-be-failed one is still running, so we won't fail the newly restarted one by accident.

## Verifying this change

The issue is revealed by the unstable failure of `KafkaProducerExactlyOnceITCase` thus is already covered by existing tests.

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/checkpoint/CheckpointCoordinator.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/checkpoint/CheckpointCoordinator.java
@@ -446,7 +446,7 @@ public class CheckpointCoordinator {
 			long latestGeneratedCheckpointId = getCheckpointIdCounter().get();
 			// here we can not get the failed pending checkpoint's id,
 			// so we pass the negative latest generated checkpoint id as a special flag
-			failureManager.handleCheckpointException(e, -1 * latestGeneratedCheckpointId);
+			failureManager.handleJobLevelCheckpointException(e, -1 * latestGeneratedCheckpointId);
 			return false;
 		}
 	}
@@ -1331,7 +1331,7 @@ public class CheckpointCoordinator {
 	 *
 	 * @param pendingCheckpoint to discard
 	 * @param cause for discarding the checkpoint
-	 * @param executionAttemptID the execution attempt id as a safe guard.
+	 * @param executionAttemptID the execution attempt id of the failing task.
 	 */
 	private void discardCheckpoint(
 		PendingCheckpoint pendingCheckpoint,
@@ -1345,12 +1345,12 @@ public class CheckpointCoordinator {
 		LOG.info("Discarding checkpoint {} of job {}.", checkpointId, job, cause);
 
 		if (cause == null) {
-			failPendingCheckpoint(pendingCheckpoint, CheckpointFailureReason.CHECKPOINT_DECLINED, executionAttemptID);
+			failPendingCheckpointDueToTaskFailure(pendingCheckpoint, CheckpointFailureReason.CHECKPOINT_DECLINED, executionAttemptID);
 		} else if (cause instanceof CheckpointException) {
 			CheckpointException exception = (CheckpointException) cause;
-			failPendingCheckpoint(pendingCheckpoint, exception.getCheckpointFailureReason(), cause, executionAttemptID);
+			failPendingCheckpointDueToTaskFailure(pendingCheckpoint, exception.getCheckpointFailureReason(), cause, executionAttemptID);
 		} else {
-			failPendingCheckpoint(pendingCheckpoint, CheckpointFailureReason.JOB_FAILURE, cause, executionAttemptID);
+			failPendingCheckpointDueToTaskFailure(pendingCheckpoint, CheckpointFailureReason.JOB_FAILURE, cause, executionAttemptID);
 		}
 
 		rememberRecentCheckpointId(checkpointId);
@@ -1407,34 +1407,43 @@ public class CheckpointCoordinator {
 			final PendingCheckpoint pendingCheckpoint,
 			final CheckpointFailureReason reason) {
 
-		failPendingCheckpoint(pendingCheckpoint, reason, (ExecutionAttemptID) null);
+		failPendingCheckpoint(pendingCheckpoint, reason, null);
 	}
 
 	private void failPendingCheckpoint(
-			final PendingCheckpoint pendingCheckpoint,
-			final CheckpointFailureReason reason,
-			final Throwable cause) {
-		failPendingCheckpoint(pendingCheckpoint, reason, cause, null);
-	}
-
-	private void failPendingCheckpoint(
-			final PendingCheckpoint pendingCheckpoint,
-			final CheckpointFailureReason reason,
-			final ExecutionAttemptID executionAttemptID) {
-
-		failPendingCheckpoint(pendingCheckpoint, reason, null, executionAttemptID);
-	}
-
-	private void failPendingCheckpoint(
-			final PendingCheckpoint pendingCheckpoint,
-			final CheckpointFailureReason reason,
-			final Throwable cause,
-			@Nullable final ExecutionAttemptID executionAttemptID) {
+		final PendingCheckpoint pendingCheckpoint,
+		final CheckpointFailureReason reason,
+		final Throwable cause) {
 
 		CheckpointException exception = new CheckpointException(reason, cause);
 		pendingCheckpoint.abort(reason, cause);
-		failureManager.handleCheckpointException(exception, pendingCheckpoint.getCheckpointId(), executionAttemptID);
+		failureManager.handleJobLevelCheckpointException(exception, pendingCheckpoint.getCheckpointId());
 
+		checkAndResetCheckpointScheduler();
+	}
+
+	private void failPendingCheckpointDueToTaskFailure(
+		final PendingCheckpoint pendingCheckpoint,
+		final CheckpointFailureReason reason,
+		final ExecutionAttemptID executionAttemptID) {
+
+		failPendingCheckpointDueToTaskFailure(pendingCheckpoint, reason, null, executionAttemptID);
+	}
+
+	private void failPendingCheckpointDueToTaskFailure(
+			final PendingCheckpoint pendingCheckpoint,
+			final CheckpointFailureReason reason,
+			final Throwable cause,
+			final ExecutionAttemptID executionAttemptID) {
+
+		CheckpointException exception = new CheckpointException(reason, cause);
+		pendingCheckpoint.abort(reason, cause);
+		failureManager.handleTaskLevelCheckpointException(exception, pendingCheckpoint.getCheckpointId(), executionAttemptID);
+
+		checkAndResetCheckpointScheduler();
+	}
+
+	private void checkAndResetCheckpointScheduler() {
 		if (!shutdown && periodicScheduling && currentPeriodicTrigger == null) {
 			synchronized (lock) {
 				if (pendingCheckpoints.isEmpty() || allPendingCheckpointsDiscarded()) {
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/checkpoint/CheckpointCoordinator.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/checkpoint/CheckpointCoordinator.java
@@ -396,6 +396,7 @@ public class CheckpointCoordinator {
 			@Nullable final String targetLocation) {
 
 		final CheckpointProperties properties = CheckpointProperties.forSyncSavepoint();
+
 		return triggerSavepointInternal(timestamp, properties, advanceToEndOfEventTime, targetLocation).handle(
 				(completedCheckpoint, throwable) -> {
 					if (throwable != null) {
@@ -447,7 +448,7 @@ public class CheckpointCoordinator {
 			long latestGeneratedCheckpointId = getCheckpointIdCounter().get();
 			// here we can not get the failed pending checkpoint's id,
 			// so we pass the negative latest generated checkpoint id as a special flag
-			failureManager.handleCheckpointException(e, -1 * latestGeneratedCheckpointId);
+			failureManager.handleCheckpointException(e, -1 * latestGeneratedCheckpointId, null);
 			return false;
 		}
 	}
@@ -662,7 +663,7 @@ public class CheckpointCoordinator {
 						checkpointID, job, numUnsuccessful, t);
 
 				if (!checkpoint.isDiscarded()) {
-					failPendingCheckpoint(checkpoint, CheckpointFailureReason.TRIGGER_CHECKPOINT_FAILURE, t);
+					failPendingCheckpoint(checkpoint, CheckpointFailureReason.TRIGGER_CHECKPOINT_FAILURE, t, null);
 				}
 
 				try {
@@ -718,7 +719,7 @@ public class CheckpointCoordinator {
 					message.getTaskExecutionId(),
 					job,
 					taskManagerLocationInfo);
-				discardCheckpoint(checkpoint, message.getReason());
+				discardCheckpoint(checkpoint, message.getReason(), message.getTaskExecutionId());
 			}
 			else if (checkpoint != null) {
 				// this should not happen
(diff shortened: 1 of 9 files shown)


flink-runtime/src/main/java/org/apache/flink/runtime/checkpoint/CheckpointCoordinator.java
```diff
@@ -859,7 +860,7 @@ private void completePendingCheckpoint(PendingCheckpoint pendingCheckpoint) thro
 			catch (Exception e1) {
 				// abort the current pending checkpoint if we fails to finalize the pending checkpoint.
 				if (!pendingCheckpoint.isDiscarded()) {
-					failPendingCheckpoint(pendingCheckpoint, CheckpointFailureReason.FINALIZE_CHECKPOINT_FAILURE, e1);
+					failPendingCheckpoint(pendingCheckpoint, CheckpointFailureReason.FINALIZE_CHECKPOINT_FAILURE, e1, null);
```
pnowojski: ditto: add an overloaded `failPendingCheckpoint` method?

carp84:
Checked on my local travis (with some [modification](https://github.com/carp84/flink/commit/09b5137a2437223a7b8f76fc0503115772c84059) to run `KafkaProducerExactlyOnceITCase` repeatedly until failure) and confirmed [w/o patch](https://travis-ci.org/carp84/flink/jobs/568293467) it fails stably and [w/ patch](https://travis-ci.org/carp84/flink/builds/568275598) the issue disappears.

flink-runtime/src/main/java/org/apache/flink/runtime/checkpoint/CheckpointCoordinator.java
```diff
@@ -1407,14 +1412,23 @@ private void failPendingCheckpoint(
...
+
+		failPendingCheckpoint(pendingCheckpoint, reason, null, executionAttemptID);
+	}
+
 	private void failPendingCheckpoint(
 			final PendingCheckpoint pendingCheckpoint,
 			final CheckpointFailureReason reason,
-			final Throwable cause) {
+			final Throwable cause,
+			final ExecutionAttemptID executionAttemptID) {
```
pnowojski: you are missing `@Nullable` annotation.

Also per the recent `[DISCUSS][CODE STYLE] Usage of Java Optional` dev mailing list discussion, I would actually prefer to make this `cause ` and `executionAttemptID` parameters `Optional` (it's a private method) as this would avoid this missing `@Nullable` annotation errors.

flink-runtime/src/main/java/org/apache/flink/runtime/checkpoint/CheckpointFailureManager.java
```diff
@@ -59,6 +61,23 @@ public CheckpointFailureManager(int tolerableCpFailureNumber, FailJobCallback fa
...
+	 *
+	 * @param exception the checkpoint exception.
+	 * @param checkpointId the failed checkpoint id used to count the continuous failure number based on
+	 *                     checkpoint id sequence. In trigger phase, we may not get the checkpoint id when the failure
+	 *                     happens before the checkpoint id generation. In this case, it will be specified a negative
+	 *                      latest generated checkpoint id as a special flag.
+	 * @param executionAttemptID the execution attempt id, as a safe guard.
+	 */
+	public void handleCheckpointException(
+		CheckpointException exception,
```
pnowojski: nit: formatting. Usually we tend to double intend (with two tabs) the parameters list, to avoid first line of method's body "mangling" with the parameters.

flink-runtime/src/main/java/org/apache/flink/runtime/checkpoint/CheckpointFailureManager.java
```diff
@@ -59,6 +61,23 @@ public CheckpointFailureManager(int tolerableCpFailureNumber, FailJobCallback fa
...
+		long checkpointId,
+		@Nullable ExecutionAttemptID executionAttemptID) {
```
pnowojski: ditto: I would prefer `Optional` here.

flink-runtime/src/main/java/org/apache/flink/runtime/checkpoint/CheckpointFailureManager.java
```diff
@@ -151,7 +170,13 @@ private static boolean isPreFlightFailure(final Throwable cause) {
...
 	public interface FailJobCallback {
 
-		void failJob(final Throwable cause);
+		/**
+		 * Fails the whole job graph.
+		 *
+		 * @param cause The reason why the job is cancelled.
+		 * @param failingAttempt The failing attempt id to prevent failing the wrong job.
+		 */
+		void failJob(final Throwable cause, @Nullable final ExecutionAttemptID failingAttempt);
```
pnowojski: Can we make `failingAttempt` parameter obligatory? If not, it would be better to avoid null by overloading this method.

flink-runtime/src/main/java/org/apache/flink/runtime/checkpoint/CheckpointCoordinator.java
```diff
@@ -447,7 +448,7 @@ public boolean triggerCheckpoint(long timestamp, boolean isPeriodic) {
 			long latestGeneratedCheckpointId = getCheckpointIdCounter().get();
 			// here we can not get the failed pending checkpoint's id,
 			// so we pass the negative latest generated checkpoint id as a special flag
-			failureManager.handleCheckpointException(e, -1 * latestGeneratedCheckpointId);
+			failureManager.handleCheckpointException(e, -1 * latestGeneratedCheckpointId, null);
```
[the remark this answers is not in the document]
carp84: Actually I already have introduced a new `handleCheckpointException` method with 3 parameters, and adding another one will cause collision.

flink-runtime/src/main/java/org/apache/flink/runtime/checkpoint/CheckpointCoordinator.java
carp84: It's suggested to not using `Optional` as parameters, please refer to my [reply](https://s.apache.org/nbl83) here. And yes will add `@Nullable` annotation here.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/ExecutionGraph.java
```diff
@@ -575,8 +575,17 @@ public void enableCheckpointing(
...
-				cause -> getJobMasterMainThreadExecutor().execute(() -> failGlobal(cause))
+			chkConfig.getTolerableCheckpointFailureNumber(),
+			(cause, failingAttempt) -> getJobMasterMainThreadExecutor().execute(() -> {
+				if (failingAttempt == null) {
+					failGlobal(cause);
+				} else {
+					final Execution failedExecution = currentExecutions.get(failingAttempt);
+					if (failedExecution != null && failedExecution.getState() == ExecutionState.RUNNING) {
+						failGlobal(cause);
+					}
```
Myasuka: How about adding a `LOG.warn` here just like `CheckpointCoordinator` received a `UNKNOWN AcknowledgeCheckpoint` ?

flink-runtime/src/test/java/org/apache/flink/runtime/checkpoint/CheckpointCoordinatorTest.java
```diff
@@ -128,7 +128,7 @@
 
 	@Before
 	public void setUp() throws Exception {
-		failureManager = new CheckpointFailureManager(0, throwable -> {});
+		failureManager = new CheckpointFailureManager(0, (throwable,attemptID) -> {});
```
Myasuka: We should better leave a space between `throwable` and `attemptID`.

flink-runtime/src/main/java/org/apache/flink/runtime/checkpoint/CheckpointCoordinator.java
carp84: I already have introduced a new `handleCheckpointException` method with 3 parameters, and it requires to do a type casting on `null` when introducing another one and invoking with `null` as the last parameter, which I _thought_ to be not recommended. But after a [search and check](https://coderanch.com/t/382683/java/Casting-null-object) I believe it's fine for casting null to help compiler choose between overloading methods, so yes let me do this.

flink-runtime/src/main/java/org/apache/flink/runtime/checkpoint/CheckpointFailureManager.java
carp84: I'm afraid we cannot since there're cases like exception thrown when `triggerCheckpoint`, as well as the legacy `handleSynchronousSavepointFailure` method in `CheckpointFailureManager`, which are not caused by a single task failure. Will add a default interface to overload this method.

flink-runtime/src/main/java/org/apache/flink/runtime/checkpoint/CheckpointCoordinator.java
```diff
@@ -1401,21 +1404,46 @@ public void run() {
...
+			final PendingCheckpoint pendingCheckpoint,
+			final CheckpointFailureReason reason) {
 
 		failPendingCheckpoint(pendingCheckpoint, reason, null);
 	}
 
 	private void failPendingCheckpoint(
+		final PendingCheckpoint pendingCheckpoint,
+		final CheckpointFailureReason reason,
+		final Throwable cause) {
```
tillrohrmann: `@Nullable` annotation is missing.

flink-runtime/src/main/java/org/apache/flink/runtime/checkpoint/CheckpointCoordinator.java
```diff
@@ -1401,21 +1404,46 @@ public void run() {
...
+		final ExecutionAttemptID executionAttemptID) {
+
+		failPendingCheckpointDueToTaskFailure(pendingCheckpoint, reason, null, executionAttemptID);
+	}
+
+	private void failPendingCheckpointDueToTaskFailure(
 			final PendingCheckpoint pendingCheckpoint,
 			final CheckpointFailureReason reason,
-			final Throwable cause) {
+			final Throwable cause,
```
tillrohrmann: `@Nullable` is missing.

flink-runtime/src/main/java/org/apache/flink/runtime/checkpoint/CheckpointFailureManager.java
```diff
@@ -50,15 +51,46 @@ public CheckpointFailureManager(int tolerableCpFailureNumber, FailJobCallback fa
...
+	 * @param executionAttemptID the execution attempt id, as a safe guard.
+	 */
+	public void handleTaskLevelCheckpointException(
+			CheckpointException exception,
+			long checkpointId,
+			ExecutionAttemptID executionAttemptID) {
+		checkFailureCounter(exception, checkpointId);
+		if (continuousFailureCounter.get() > tolerableCpFailureNumber) {
+			clearCount();
+			failureCallback.failJobDueToTaskFailure(new FlinkRuntimeException("Exceeded checkpoint tolerable failure threshold."), executionAttemptID);
```
tillrohrmann: This is unrelated to this PR, but I think we should not throw a `RuntimeException` here. It is a recoverable exception and hence it should be checked.

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/ExecutionGraph.java
```diff
@@ -575,8 +575,18 @@ public void enableCheckpointing(
...
+			chkConfig.getTolerableCheckpointFailureNumber(),
+			new CheckpointFailureManager.FailJobCallback() {
+				@Override
+				public void failJob(Throwable cause) {
+					getJobMasterMainThreadExecutor().execute(() -> failGlobal(cause));
+				}
+
+				@Override
+				public void failJobDueToTaskFailure(Throwable cause, ExecutionAttemptID failingTask) {
+					getJobMasterMainThreadExecutor().execute(() -> failGlobal(cause, failingTask));
```
tillrohrmann: let's rename `failGlobal(Throwable, ExecutionAttemptId)` into `failGlobalIfTaskIsStillRunning`

flink-runtime/src/main/java/org/apache/flink/runtime/executiongraph/ExecutionGraph.java
```diff
@@ -1097,6 +1107,16 @@ public void suspend(Throwable suspensionCause) {
 		}
 	}
 
+	public void failGlobal(Throwable cause, ExecutionAttemptID failingAttempt) {
+		final Execution failedExecution = currentExecutions.get(failingAttempt);
+		if (failedExecution != null && failedExecution.getState() == ExecutionState.RUNNING) {
+			failGlobal(cause);
+		} else {
+			LOG.warn("The failing attempt {} belongs to an already not" +
```
tillrohrmann: Should not be a warning but debug log message since we expect this to happen (without having negative side effects).

flink-runtime/src/test/java/org/apache/flink/runtime/checkpoint/CheckpointCoordinatorFailureTest.java
```diff
@@ -66,7 +66,15 @@ public void testFailingCompletedCheckpointStoreAdd() throws Exception {
...
 
-		CheckpointFailureManager failureManager = new CheckpointFailureManager(0, throwable -> {});
+		CheckpointFailureManager failureManager = new CheckpointFailureManager(
+			0,
+			new CheckpointFailureManager.FailJobCallback() {
+				@Override
+				public void failJob(Throwable cause) {}
+
+				@Override
+				public void failJobDueToTaskFailure(Throwable cause, ExecutionAttemptID failingTask) {}
```
tillrohrmann: Nit: Why not using lambda here? Would be a bit shorter.

flink-runtime/src/test/java/org/apache/flink/runtime/checkpoint/CheckpointCoordinatorTest.java
```diff
@@ -3910,10 +3928,18 @@ public void jobFailsIfInFlightSynchronousSavepointIsDiscarded() throws Exception
...
+					0,
+					new CheckpointFailureManager.FailJobCallback() {
+						@Override
+						public void failJob(Throwable cause) {
+							invocationCounterAndException.f0 += 1;
+							invocationCounterAndException.f1 = cause;
+						}
+
+						@Override
+						public void failJobDueToTaskFailure(Throwable cause, ExecutionAttemptID failingTask) {}
```
tillrohrmann: Let's throw an exception here to make sure that this method is not being called.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/ExecutionGraphRestartTest.java
```diff
@@ -733,6 +733,51 @@ public void testFailureWhileRestarting() throws Exception {
 		assertEquals(JobStatus.RUNNING, executionGraph.getState());
 	}
 
+	@Test
+	public void testFailingOfTheSameExecutionWhileRestarting() throws Exception {
```
tillrohrmann: I think it would be nice if the test name states the unit under test, the state and the expected outcome. E.g. it could be named `failGlobalIfTaskIsStillRunning_failingAnExecutionTwice_ShouldTriggerOnlyOneFailover`

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/ExecutionGraphRestartTest.java
```diff
@@ -733,6 +733,51 @@ public void testFailureWhileRestarting() throws Exception {
...
+				assertNotNull("No assigned resource (test instability).", vertex.getCurrentAssignedResource());
+				vertex.getCurrentExecutionAttempt().switchToRunning();
+			}
+
+			// fail global with old finished execution, this should not affect the execution
+			eg.failGlobal(new Exception("This should have no effect"), finishedExecution.getAttemptId());
+
+			for (ExecutionVertex vertex: eg.getAllExecutionVertices()) {
+				vertex.getCurrentExecutionAttempt().markFinished();
+			}
```
tillrohrmann: I think this is not necessary. We could simply check that the job state is still `RUNNING`.

PR #9368, opened 2019-08-06
[FLINK-13592][e2e] Fix hardcoded flink version in tpch end-to-end test.

KurtYoung:

- Fix hardcoded flink version in tpch end-to-end test.

## Brief change log

- Use auto fetched flink version in tpch end-to-end test.
- Changes test parallelism to 2 to cover the situation that parallelism is higher than the slot number, since the testing cluster only have one task manager and contains only one slot.

## Verifying this change

Tested on a linux server and passed.

### DIFF
--- a/flink-end-to-end-tests/test-scripts/test_tpch.sh
+++ b/flink-end-to-end-tests/test-scripts/test_tpch.sh
@@ -72,6 +72,7 @@ execution:
   planner: blink
   type: batch
   result-mode: table
+  parallelism: 2
 EOF
 
     if [[ -e "$MODIFIED_QUERY_DIR/q$i.sql" ]]
--- a/flink-end-to-end-tests/test-scripts/test_tpch.sh
+++ b/flink-end-to-end-tests/test-scripts/test_tpch.sh
@@ -72,7 +72,6 @@ execution:
   planner: blink
   type: batch
   result-mode: table
-  parallelism: 2
 EOF
 
     if [[ -e "$MODIFIED_QUERY_DIR/q$i.sql" ]]
--- a/flink-end-to-end-tests/test-scripts/test_tpch.sh
+++ b/flink-end-to-end-tests/test-scripts/test_tpch.sh
@@ -31,9 +31,7 @@ echo "Generating test data..."
 
 TARGET_DIR="$END_TO_END_DIR/flink-tpch-test/target"
 TPCH_DATA_DIR="$END_TO_END_DIR/test-scripts/test-data/tpch"
-FLINK_VERSION=`ls "${END_TO_END_DIR}/../flink-table/flink-table-api-java/target" | sed -n "s/.*flink-table-api-java-\(.*\)-tests\.jar/\1/p" | uniq`
-
-java -cp "$TARGET_DIR/flink-tpch-test-"$FLINK_VERSION".jar:$TARGET_DIR/lib/*" org.apache.flink.table.tpch.TpchDataGenerator "$SCALE" "$TARGET_DIR"
+java -cp "$TARGET_DIR/TpchTestProgram.jar:$TARGET_DIR/lib/*" org.apache.flink.table.tpch.TpchDataGenerator "$SCALE" "$TARGET_DIR"
 
 ################################################################################
 # Prepare Flink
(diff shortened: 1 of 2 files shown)


flink-end-to-end-tests/test-scripts/test_tpch.sh
```diff
@@ -31,7 +31,9 @@ echo "Generating test data..."
 
 TARGET_DIR="$END_TO_END_DIR/flink-tpch-test/target"
 TPCH_DATA_DIR="$END_TO_END_DIR/test-scripts/test-data/tpch"
-java -cp "$TARGET_DIR/flink-tpch-test-1.10-SNAPSHOT.jar:$TARGET_DIR/lib/*" org.apache.flink.table.tpch.TpchDataGenerator "$SCALE" "$TARGET_DIR"
+FLINK_VERSION=`ls "${END_TO_END_DIR}/../flink-table/flink-table-api-java/target" | sed -n "s/.*flink-table-api-java-\(.*\)-tests\.jar/\1/p" | uniq`
```
wuchong: FLINK_VERSION has been defined in `common.sh`, maybe we can use it directly.

flink-end-to-end-tests/test-scripts/test_tpch.sh
```diff
@@ -31,7 +31,9 @@ echo "Generating test data..."
...
+
+java -cp "$TARGET_DIR/flink-tpch-test-"$FLINK_VERSION".jar:$TARGET_DIR/lib/*" org.apache.flink.table.tpch.TpchDataGenerator "$SCALE" "$TARGET_DIR"
```
wuchong:
```suggestion
java -cp "${TARGET_DIR}/flink-tpch-test-${FLINK_VERSION}.jar:$TARGET_DIR/lib/*" org.apache.flink.table.tpch.TpchDataGenerator "$SCALE" "$TARGET_DIR"
```

wuchong: nit

flink-end-to-end-tests/test-scripts/test_tpch.sh
```diff
@@ -72,6 +74,7 @@ execution:
   planner: blink
   type: batch
   result-mode: table
+  parallelism: 2
```
dawidwys: Do we need this change? It's unrelated to the subject of the commit.

flink-end-to-end-tests/test-scripts/test_tpch.sh
1u0: The `flink-tpch-test-"$FLINK_VERSION".jar` is test only artifact. You can change the `flink-end-to-end-tests/flink-tpch-test/pom.xml` to produce the final build artifact **without** version in the name, instead of detecting the Flink version.

For example, `flink-stream-sql-test` produces an artifact without version.

_Side note:_ your version checking depends on some other artifact (`flink-table-api-java`), this can silently brake in the future.

wuchong:
+1 for strip off version.

flink-end-to-end-tests/test-scripts/test_tpch.sh
KurtYoung: It's a supplement to this issue: https://issues.apache.org/jira/browse/FLINK-13441. I think this is also a good place to cover such scenarios, and don't need to open a jira for it.

dawidwys: How is this related to the linked issue? Honestly I still don't get what is the purpose of this change.

I disagree this is a good place to add such cases. It's impossible to link this change to the reason why it was introduced. I'm not saying we need a new JIRA issue, but it should at least  be in a separate hotfix commit with at least brief explanation what it does.

KurtYoung:
Striping off version sounds good, I will change it.

flink-end-to-end-tests/test-scripts/test_tpch.sh
KurtYoung: I explained this change in the commit message if you missed that: 

> Also changed test parallelism to 2 to cover the situation that parallelism is higher than the slot number, since the testing cluster only have one task manager and contains only one slot

I agree this can be isolated to another commit.

dawidwys: You are right I missed it. Now I get it. I think this would be a perfect description for the extracted commit.

KurtYoung: Yeah, I already did the exactly thing in my last commit.

KurtYoung:
@dawidwys Thanks for the reviewing, I tested locally again in my laptop and passed. I will merge this if you don't have any further comment.

PR #9373, opened 2019-08-06
[FLINK-13596][ml] Add two utils for Table transformations

xuyang1706:

For the input data type is Table, but during the ML training process, we need use DataSet and DataStream type. Thus, we need the type transformation utils in the ML algorithm implementations.

## Brief change log

- *Add the transformation between DataSet and Table*
- *Add the transformation between DataStream and Table*
- *Add unit test cases*
 

## Verifying this change

- run test case pass

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/DataSetConversionUtil.java
+++ b/flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/DataSetConversionUtil.java
@@ -34,16 +34,19 @@ import org.apache.flink.types.Row;
 /**
  * Provide functions of conversions between DataSet and Table.
  */
-public class RowTypeDataSet {
+public class DataSetConversionUtil {
 	/**
-	 * Convert the given Table to {@link DataSet DataSet&lt;Row&gt;}.
+	 * Convert the given Table to {@link DataSet}<{@link Row}>.
 	 *
 	 * @param sessionId the sessionId of {@link MLEnvironmentFactory}
 	 * @param table the Table to convert.
 	 * @return the converted DataSet.
 	 */
 	public static DataSet <Row> fromTable(Long sessionId, Table table) {
-		return MLEnvironmentFactory.get(sessionId).getBatchTableEnvironment().toDataSet(table, Row.class);
+		return MLEnvironmentFactory
+			.get(sessionId)
+			.getBatchTableEnvironment()
+			.toDataSet(table, Row.class);
 	}
 
 	/**
--- a/flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/DataSetConversionUtil.java
+++ b/flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/DataSetConversionUtil.java
@@ -29,6 +29,7 @@ import org.apache.flink.ml.common.MLEnvironment;
 import org.apache.flink.ml.common.MLEnvironmentFactory;
 import org.apache.flink.table.api.Table;
 import org.apache.flink.table.api.TableSchema;
+import org.apache.flink.table.api.ValidationException;
 import org.apache.flink.types.Row;
 
 /**
@@ -110,7 +111,7 @@ public class DataSetConversionUtil {
 				((TwoInputUdfOperator) data).returns(new RowTypeInfo(colTypes, colNames));
 			}
 			return toTable(session, data, colNames);
-		} catch (Exception ex) {
+		} catch (ValidationException ex) {
 			if (null == colTypes) {
 				throw ex;
 			} else {
--- a/flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/RowTypeDataSet.java
+++ b/flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/RowTypeDataSet.java
@@ -22,8 +22,11 @@ package org.apache.flink.ml.common.utils;
 import org.apache.flink.api.common.functions.MapFunction;
 import org.apache.flink.api.common.typeinfo.TypeInformation;
 import org.apache.flink.api.java.DataSet;
+import org.apache.flink.api.java.operators.SingleInputUdfOperator;
+import org.apache.flink.api.java.operators.TwoInputUdfOperator;
 import org.apache.flink.api.java.typeutils.RowTypeInfo;
-import org.apache.flink.ml.common.MLSession;
+import org.apache.flink.ml.common.MLEnvironment;
+import org.apache.flink.ml.common.MLEnvironmentFactory;
 import org.apache.flink.table.api.Table;
 import org.apache.flink.table.api.TableSchema;
 import org.apache.flink.types.Row;
@@ -32,51 +35,33 @@ import org.apache.flink.types.Row;
  * Provide functions of conversions between DataSet and Table.
  */
 public class RowTypeDataSet {
-
 	/**
 	 * Convert the given Table to {@link DataSet DataSet&lt;Row&gt;}.
 	 *
+	 * @param sessionId the sessionId of {@link MLEnvironmentFactory}
 	 * @param table the Table to convert.
 	 * @return the converted DataSet.
 	 */
-	public static DataSet <Row> fromTable(Table table) {
-		return MLSession.getBatchTableEnvironment().toDataSet(table, Row.class);
+	public static DataSet <Row> fromTable(Long sessionId, Table table) {
+		return MLEnvironmentFactory.get(sessionId).getBatchTableEnvironment().toDataSet(table, Row.class);
 	}
 
 	/**
 	 * Convert the given DataSet into a Table with specified TableSchema.
 	 *
+	 * @param sessionId the sessionId of {@link MLEnvironmentFactory}
 	 * @param data   the DataSet to convert.
 	 * @param schema the specified TableSchema.
 	 * @return the converted Table.
 	 */
-	public static Table toTable(DataSet <Row> data, TableSchema schema) {
-		return toTable(data, schema.getFieldNames(), schema.getFieldTypes());
-	}
-
-	/**
-	 * Convert the given DataSet into a Table with specified colNames.
-	 *
-	 * @param data     the DataSet to convert.
-	 * @param colNames the specified colNames.
-	 * @return the converted Table.
-	 */
-	public static Table toTable(DataSet <Row> data, String[] colNames) {
-		if (null == colNames || colNames.length == 0) {
-			return MLSession.getBatchTableEnvironment().fromDataSet(data);
-		} else {
-			StringBuilder sbd = new StringBuilder();
-			sbd.append(colNames[0]);
-			for (int i = 1; i < colNames.length; i++) {
-				sbd.append(",").append(colNames[i]);
-			}
-			return MLSession.getBatchTableEnvironment().fromDataSet(data, sbd.toString());
-		}
+	public static Table toTable(Long sessionId, DataSet <Row> data, TableSchema schema) {
+		return toTable(sessionId, data, schema.getFieldNames(), schema.getFieldTypes());
 	}
 
 	/**
 	 * Convert the given DataSet into a Table with specified colNames and colTypes.
 	 *
+	 * @param sessionId sessionId the sessionId of {@link MLEnvironmentFactory}.
 	 * @param data     the DataSet to convert.
 	 * @param colNames the specified colNames.
 	 * @param colTypes the specified colTypes. This variable is used only when the
@@ -84,16 +69,72 @@ public class RowTypeDataSet {
 	 *                 automatically what the produced type is.
 	 * @return the converted Table.
 	 */
-	public static Table toTable(DataSet <Row> data, String[] colNames, TypeInformation <?>[] colTypes) {
+	public static Table toTable(Long sessionId, DataSet <Row> data, String[] colNames, TypeInformation <?>[] colTypes) {
+		return toTable(MLEnvironmentFactory.get(sessionId), data, colNames, colTypes);
+	}
+
+	/**
+	 * Convert the given DataSet into a Table with specified colNames.
+	 *
+	 * @param sessionId sessionId the sessionId of {@link MLEnvironmentFactory}.
+	 * @param data     the DataSet to convert.
+	 * @param colNames the specified colNames.
+	 * @return the converted Table.
+	 */
+	public static Table toTable(Long sessionId, DataSet <Row> data, String[] colNames) {
+		return toTable(MLEnvironmentFactory.get(sessionId), data, colNames);
+	}
+
+	/**
+	 * Convert the given DataSet into a Table with specified colNames and colTypes.
+	 *
+	 * @param session the MLEnvironment using to convert DataSet to Table.
+	 * @param data     the DataSet to convert.
+	 * @param colNames the specified colNames.
+	 * @param colTypes the specified colTypes. This variable is used only when the
+	 *                 DataSet is produced by a function and Flink cannot determine
+	 *                 automatically what the produced type is.
+	 * @return the converted Table.
+	 */
+	public static Table toTable(MLEnvironment session, DataSet <Row> data, String[] colNames, TypeInformation <?>[] colTypes) {
 		try {
-			return toTable(data, colNames);
+			// Try to add row type information for the dataset to be converted.
+			// In most case, this keeps us from the rolling back logic in the catch block,
+			// which adds an unnecessary map function just in order to add row type information.
+			if (data instanceof SingleInputUdfOperator) {
+				((SingleInputUdfOperator) data).returns(new RowTypeInfo(colTypes, colNames));
+			} else if (data instanceof TwoInputUdfOperator) {
+				((TwoInputUdfOperator) data).returns(new RowTypeInfo(colTypes, colNames));
+			}
+			return toTable(session, data, colNames);
 		} catch (Exception ex) {
 			if (null == colTypes) {
 				throw ex;
 			} else {
 				DataSet <Row> t = getDataSetWithExplicitTypeDefine(data, colNames, colTypes);
-				return toTable(t, colNames);
+				return toTable(session, t, colNames);
 			}
+		}
+	}
+
+	/**
+	 * Convert the given DataSet into a Table with specified colNames.
+	 *
+	 * @param session the MLEnvironment using to convert DataSet to Table.
+	 * @param data     the DataSet to convert.
+	 * @param colNames the specified colNames.
+	 * @return the converted Table.
+	 */
+	public static Table toTable(MLEnvironment session, DataSet <Row> data, String[] colNames) {
+		if (null == colNames || colNames.length == 0) {
+			return session.getBatchTableEnvironment().fromDataSet(data);
+		} else {
+			StringBuilder sbd = new StringBuilder();
+			sbd.append(colNames[0]);
+			for (int i = 1; i < colNames.length; i++) {
+				sbd.append(",").append(colNames[i]);
+			}
+			return session.getBatchTableEnvironment().fromDataSet(data, sbd.toString());
 		}
 	}
 
--- a/flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/RowTypeDataSet.java
+++ b/flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/RowTypeDataSet.java
@@ -1,0 +1,127 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.flink.ml.common.utils;
+
+import org.apache.flink.api.common.functions.MapFunction;
+import org.apache.flink.api.common.typeinfo.TypeInformation;
+import org.apache.flink.api.java.DataSet;
+import org.apache.flink.api.java.typeutils.RowTypeInfo;
+import org.apache.flink.ml.common.MLSession;
+import org.apache.flink.table.api.Table;
+import org.apache.flink.table.api.TableSchema;
+import org.apache.flink.types.Row;
+
+/**
+ * Provide functions of conversions between DataSet and Table.
+ */
+public class RowTypeDataSet {
+
+	/**
+	 * Convert the given Table to {@link DataSet DataSet&lt;Row&gt;}.
+	 *
+	 * @param table the Table to convert.
+	 * @return the converted DataSet.
+	 */
+	public static DataSet <Row> fromTable(Table table) {
+		return MLSession.getBatchTableEnvironment().toDataSet(table, Row.class);
+	}
+
+	/**
+	 * Convert the given DataSet into a Table with specified TableSchema.
+	 *
+	 * @param data   the DataSet to convert.
+	 * @param schema the specified TableSchema.
+	 * @return the converted Table.
+	 */
+	public static Table toTable(DataSet <Row> data, TableSchema schema) {
+		return toTable(data, schema.getFieldNames(), schema.getFieldTypes());
+	}
+
+	/**
+	 * Convert the given DataSet into a Table with specified colNames.
+	 *
+	 * @param data     the DataSet to convert.
+	 * @param colNames the specified colNames.
+	 * @return the converted Table.
+	 */
+	public static Table toTable(DataSet <Row> data, String[] colNames) {
+		if (null == colNames || colNames.length == 0) {
+			return MLSession.getBatchTableEnvironment().fromDataSet(data);
+		} else {
+			StringBuilder sbd = new StringBuilder();
+			sbd.append(colNames[0]);
+			for (int i = 1; i < colNames.length; i++) {
+				sbd.append(",").append(colNames[i]);
+			}
+			return MLSession.getBatchTableEnvironment().fromDataSet(data, sbd.toString());
+		}
+	}
+
+	/**
+	 * Convert the given DataSet into a Table with specified colNames and colTypes.
+	 *
+	 * @param data     the DataSet to convert.
+	 * @param colNames the specified colNames.
+	 * @param colTypes the specified colTypes. This variable is used only when the
+	 *                 DataSet is produced by a function and Flink cannot determine
+	 *                 automatically what the produced type is.
+	 * @return the converted Table.
+	 */
+	public static Table toTable(DataSet <Row> data, String[] colNames, TypeInformation <?>[] colTypes) {
+		try {
+			return toTable(data, colNames);
+		} catch (Exception ex) {
+			if (null == colTypes) {
+				throw ex;
+			} else {
+				DataSet <Row> t = getDataSetWithExplicitTypeDefine(data, colNames, colTypes);
+				return toTable(t, colNames);
+			}
+		}
+	}
+
+	/**
+	 * Adds a type information hint about the colTypes with the Row to the DataSet.
+	 *
+	 * @param data     the DataSet to add type information.
+	 * @param colNames the specified colNames
+	 * @param colTypes the specified colTypes
+	 * @return the DataSet with type information hint.
+	 */
+	private static DataSet <Row> getDataSetWithExplicitTypeDefine(
+		DataSet <Row> data,
+		String[] colNames,
+		TypeInformation <?>[] colTypes) {
+
+		DataSet <Row> r = data
+			.map(
+				new MapFunction <Row, Row>() {
+					@Override
+					public Row map(Row t) throws Exception {
+						return t;
+					}
+				}
+			)
+			.returns(new RowTypeInfo(colTypes, colNames));
+
+		return r;
+	}
+
+}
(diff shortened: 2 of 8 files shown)


walterddr:
Thanks for the contribution @xuyang1706. looks like this PR depends on #9184 yes?

xuyang1706:
> Thanks for the contribution @xuyang1706. looks like this PR depends on #9184 yes?

@walterddr , yes, you are right. It need use `MLEnvironmentFactory` which is in PR #9184.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/RowTypeDataSet.java
```diff
@@ -0,0 +1,168 @@
...
+public class RowTypeDataSet {
+	/**
+	 * Convert the given Table to {@link DataSet DataSet&lt;Row&gt;}.
+	 *
+	 * @param sessionId the sessionId of {@link MLEnvironmentFactory}
+	 * @param table the Table to convert.
+	 * @return the converted DataSet.
+	 */
+	public static DataSet <Row> fromTable(Long sessionId, Table table) {
+		return MLEnvironmentFactory.get(sessionId).getBatchTableEnvironment().toDataSet(table, Row.class);
```
walterddr: did it throw exception of line too long? this should be captured by checkstyle though

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/RowTypeDataStream.java
```diff
@@ -0,0 +1,157 @@
...
+ */
+public class RowTypeDataStream {
+	/**
+	 * Convert the given Table to {@link DataStream DataStream&lt;Row&gt;}.
+	 *
+	 * @param sessionId the sessionId of {@link MLEnvironmentFactory}
+	 * @param table the Table to convert.
+	 * @return the converted DataStream.
+	 */
+	public static DataStream <Row> fromTable(Long sessionId, Table table) {
```
walterddr: nit: If we rename these two APIs to `convertToDataStream` & `convertToDataSet` we can probably combine the two files. since `MLEnvironment` encapsulates both Batch and Stream envs. it is also consistent IMO.

I am fine with keeping two files and separate these APIs as well. maybe some other folks can provide some feedbacks on this.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/RowTypeDataSet.java
xuyang1706: Thanks, changed

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/RowTypeDataSet.java
```diff
@@ -0,0 +1,168 @@
...
+import org.apache.flink.ml.common.MLEnvironment;
+import org.apache.flink.ml.common.MLEnvironmentFactory;
+import org.apache.flink.table.api.Table;
+import org.apache.flink.table.api.TableSchema;
+import org.apache.flink.types.Row;
+
+/**
+ * Provide functions of conversions between DataSet and Table.
+ */
+public class RowTypeDataSet {
```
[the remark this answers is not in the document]
xuyang1706: Thanks, changed

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/RowTypeDataSet.java
```diff
@@ -0,0 +1,168 @@
...
+	/**
+	 * Convert the given Table to {@link DataSet DataSet&lt;Row&gt;}.
```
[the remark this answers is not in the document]
xuyang1706: Thanks, changed

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/RowTypeDataSet.java
```diff
@@ -0,0 +1,168 @@
...
+	 *
+	 * @param sessionId sessionId the sessionId of {@link MLEnvironmentFactory}.
+	 * @param data     the DataSet to convert.
+	 * @param colNames the specified colNames.
+	 * @param colTypes the specified colTypes. This variable is used only when the
+	 *                 DataSet is produced by a function and Flink cannot determine
+	 *                 automatically what the produced type is.
+	 * @return the converted Table.
+	 */
+	public static Table toTable(Long sessionId, DataSet <Row> data, String[] colNames, TypeInformation <?>[] colTypes) {
```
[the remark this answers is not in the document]
xuyang1706: We totally agree to comply with FLIP-65 in long term. Meanwhile we depend heavily on `DataSet` which still uses legacy type system, `public TypeInformation<T> getType()` for example. Maybe we can take gradual steps to eliminate the lagacy in the future.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/RowTypeDataSet.java
```diff
@@ -0,0 +1,168 @@
...
+	/**
+	 * Convert the given DataSet into a Table with specified TableSchema.
+	 *
+	 * @param sessionId the sessionId of {@link MLEnvironmentFactory}
+	 * @param data   the DataSet to convert.
+	 * @param schema the specified TableSchema.
+	 * @return the converted Table.
+	 */
+	public static Table toTable(Long sessionId, DataSet <Row> data, TableSchema schema) {
+		return toTable(sessionId, data, schema.getFieldNames(), schema.getFieldTypes());
```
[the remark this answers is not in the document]
xuyang1706: refer the reply below.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/RowTypeDataStream.java
xuyang1706: Thanks for your suggestion, I prefer to keep two files.

xuyang1706:
> Thanks for the contribution @xuyang1706 and sorry for the late review.
> 
> Overall the utility looks good and to me this is definitely a need for the ML lib. I just left a few comments regarding the APIs and some of the type related handlings. please kindly take a look --Rong.

Thanks for your comments and advices @walterddr . I have renamed the two classes.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/DataSetConversionUtil.java
```diff
@@ -0,0 +1,171 @@
...
+
+	/**
+	 * Convert the given DataSet into a Table with specified TableSchema.
+	 *
+	 * @param sessionId the sessionId of {@link MLEnvironmentFactory}
+	 * @param data   the DataSet to convert.
+	 * @param schema the specified TableSchema.
+	 * @return the converted Table.
+	 */
+	public static Table toTable(Long sessionId, DataSet <Row> data, TableSchema schema) {
```
walterddr: This API seems to be not tested.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/DataSetConversionUtil.java
```diff
@@ -0,0 +1,171 @@
...
+			// Try to add row type information for the dataset to be converted.
+			// In most case, this keeps us from the rolling back logic in the catch block,
+			// which adds an unnecessary map function just in order to add row type information.
+			if (data instanceof SingleInputUdfOperator) {
+				((SingleInputUdfOperator) data).returns(new RowTypeInfo(colTypes, colNames));
+			} else if (data instanceof TwoInputUdfOperator) {
+				((TwoInputUdfOperator) data).returns(new RowTypeInfo(colTypes, colNames));
+			}
+			return toTable(session, data, colNames);
+		} catch (Exception ex) {
```
walterddr: Is there anyway to avoid doing this. Ideally speaking an exception is an exception and we shouldn't catch it and branch out to a different solution.

I am assuming you can do some pre-check to see if this `toTable` operation is valid (If I understood correctly, basically doing the type check between the `colTypes` and the actual dataset type right?)

walterddr: also this exception branch is not tested.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/DataSetConversionUtil.java
```diff
@@ -0,0 +1,171 @@
...
+	/**
+	 * Convert the given DataSet into a Table with specified colNames.
+	 *
+	 * @param session the MLEnvironment using to convert DataSet to Table.
+	 * @param data     the DataSet to convert.
+	 * @param colNames the specified colNames.
+	 * @return the converted Table.
+	 */
+	public static Table toTable(MLEnvironment session, DataSet <Row> data, String[] colNames) {
+		if (null == colNames || colNames.length == 0) {
```
walterddr: Shouldn't we combine this with the previous API?
```
public static Table toTable(MLEnvironment session, DataSet <Row> data, String[] colNames, TypeInformation <?>[] colTypes)
```

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/DataSetConversionUtil.java
walterddr: also if `schema.getFieldTypes()` is deprecated. I would rather we remove this function and just allow the user to handle deprecation in user code. (this would be much cleaner for future maintenance)

xuyang1706: Thanks, The case has been added.
we must set the table using the column types (see below) and the TypeInformation is used in DataSet Api, so the `TableSchema` is necessary.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/DataSetConversionUtil.java
xuyang1706: because getType() in https://github.com/apache/flink/blob/937356f786b5edac28a7ff44fc31410761c6d7f0/flink-java/src/main/java/org/apache/flink/api/java/DataSet.java#L171 can only be called once. so we can not do the pre-check.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/DataSetConversionUtil.java
xuyang1706: flink can also detect the type of the column when the `DataSet` carry the type information

xuyang1706:
> Hi @xuyang1706 thanks for the update. I just did another pass and find that some of the branches are missing test cases; also I think the various APIs in the utilities seems a bit hard to understand and I was not sure which of these APIs are considered to be general usage and which are less utilized.
> Please kindly take a look (I only comment on the DataSet side since DataStream side is similar).

Thanks for your comments @walterddr . I have added some test cases for both DataSet and DataStream sides. Flink can detect the type of the column when the DataSet carry the type information, thus in most cases we can use the simple API, but for some bad cases, we have to input the column type information. Thanks --Xu

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/DataSetConversionUtil.java
walterddr: I am actually a little confused on this. why is the `getType()` not able to be used more than once? It seems like you cannot alter the `typeInfo` once you have access the `getType()` API. 

however you can still access the type again yes? Also in the exception handle mechanism, `getDataSetWithExplicitTypeDefine` method, you are actually making a `map` first before you assign the return type?

walterddr: Also, it is probably better to list out the actual Exception you are trying to catch here instead of a generic `Exception` and then handle it. It is unknown to contributors who maintain the `DataSet` side code --> if they change and throw some new Exceptions that happens to cross path for this method, it will silently fail (since there's not test that actually touch the `catch` branch.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/DataSetConversionUtil.java
walterddr: I meant something like:
```
public static Table toTable(MLEnvironment session, DataSet <Row> data, String[] colNames, TypeInformation <?>[] colTypes) {
  if (colTypes == null) {
    // ... what you are doing here
  } 
  // the original 4 argument code
}

public static Table toTable(MLEnvironment session, DataSet <Row> data, String[] colNames) {
  toTable(session, data, colNames, null);
}
```

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/DataSetConversionUtil.java
qiuxiafei: The actual exception thrown here is this `ValidationException`:
https://github.com/apache/flink/blob/ef4055ab989db431e29e0071573619d267960000/flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/typeutils/FieldInfoUtils.java#L184-L189
That's because sometimes the type information of Row bassed dataset is not set.

As you mentioned, our initial solution is doing some pre-check, saying checking if the `TypeInformation` is `GenericTypeInfo<Row>`. To check the type, we need to `getType()`. As a side effect, `typeUsed` is set to `true`:

https://github.com/apache/flink/blob/937356f786b5edac28a7ff44fc31410761c6d7f0/flink-java/src/main/java/org/apache/flink/api/java/DataSet.java#L171-L182

 But what if the input dataset really has no specified type information? We have to set it's TypeInformation via the `.return()` methods, which calls the following code, which says "no modification after getType called":
https://github.com/apache/flink/blob/937356f786b5edac28a7ff44fc31410761c6d7f0/flink-java/src/main/java/org/apache/flink/api/java/DataSet.java#L155-L161

That's the deadlock we faced. So we just try the best to complete the type information (for `SingleInputUdfOperator` and `TwoInputUdfOperator` here) to avoid falling into the exception branch and let the `getDataSetWithExplicitTypeDefine` method work as a final guard. 

BTW, catching `ValidationException` instead of a generic `Exception` is a good idea I think.

walterddr: This seems to me is like a bug in `DataSet` - if an exception was thrown, then `typeUsed` should not be set (since an exception was actually thrown). 

However, I think I understand the situation a lot better, thanks for the clarification @qiuxiafei. I would dig deeper in the DataSet/TypeInfo side code and see if there's any better solution.

xuyang1706:
> Thanks for the explanation @xuyang1706 and sorry for not making my comments clear. please kindly take a look at my follow up if those make sense. thanks --Rong.

Thanks for your comments and advice @walterddr . I have changed generic Exception to the actual Exception, and feed back the follow up in lines.  Thanks --Xu

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/DataSetConversionUtil.java
```diff
@@ -0,0 +1,172 @@
...
+			// In most case, this keeps us from the rolling back logic in the catch block,
+			// which adds an unnecessary map function just in order to add row type information.
+			if (data instanceof SingleInputUdfOperator) {
+				((SingleInputUdfOperator) data).returns(new RowTypeInfo(colTypes, colNames));
+			} else if (data instanceof TwoInputUdfOperator) {
+				((TwoInputUdfOperator) data).returns(new RowTypeInfo(colTypes, colNames));
+			}
+			return toTable(session, data, colNames);
+		} catch (ValidationException ex) {
+			if (null == colTypes) {
```
walterddr: shouldn't we always honor users' explicit configuration of `colTypes` instead of trying to implicitly cast it?

xuyang1706: This function is used in all operators, so if we use the explicit configuration of colTypes , the `mapfunction` will be found everywhere, which rise concern for performance.

walterddr: oops.. that's a good point! this will in fact create performance issue in runtime.

xuyang1706: so, how about keep the original implementation?

walterddr: Yes. I think for now we can merge it as is and improve it later - I will add some comments.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/DataSetConversionUtil.java
walterddr: I would add some comments here.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/DataSetConversionUtil.java
walterddr: this can be resolved.

flink-ml-parent/flink-ml-lib/src/test/java/org/apache/flink/ml/common/utils/DataStreamConversionUtilTest.java
```diff
@@ -0,0 +1,101 @@
...
+
+/**
+ * Unit Test for DataStreamConversionUtil.
+ */
+public class DataStreamConversionUtilTest {
+	@Rule
+	public ExpectedException thrown = ExpectedException.none();
+
+	@Test
+	public void test() throws Exception {
```
walterddr: I would make sure this end to end test and the per-API test in DataSet Util exist in both side when merging.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/DataSetConversionUtil.java
```diff
@@ -0,0 +1,172 @@
...
+	 *                 DataSet is produced by a function and Flink cannot determine
+	 *                 automatically what the produced type is.
+	 * @return the converted Table.
+	 */
+	public static Table toTable(MLEnvironment session, DataSet <Row> data, String[] colNames, TypeInformation <?>[] colTypes) {
+		try {
+			// Try to add row type information for the dataset to be converted.
+			// In most case, this keeps us from the rolling back logic in the catch block,
+			// which adds an unnecessary map function just in order to add row type information.
+			if (data instanceof SingleInputUdfOperator) {
```
walterddr: After improving the test I found that this piece of heuristic improvement logic doesn't work.
I actually go ahead and removed this in https://github.com/walterddr/flink/commit/c55dfd899d17fb59e62211ab2dd05ff82ae1480f. Please kindly take a look

qiuxiafei: Hi, I'm just confused about the meaning of 'not work', could you explain it a bit more in detail?
I looked into the case `testE2E`, for the invocation on `table2`, our heuristic prevented it from falling back to the explicit type definition. You can see that if you remove the heuristic. But for `tableFromDataStreamWithTableSchema`, it fell to exception branch because the `input` dataset was reused and the `getType` API was accessed twice as your described in comments.

Or maybe I just missed something ……^_^

walterddr: It is very easy to reproduce, if you checkout the https://github.com/walterddr/flink/commit/c55dfd899d17fb59e62211ab2dd05ff82ae1480f and adding line 105-112 back. you would see the testE2E fails - interestingly the test failure is actually occurring in the 3rd `toTable` call, not the second one. please see https://github.com/walterddr/flink/commit/c55dfd899d17fb59e62211ab2dd05ff82ae1480f#r36615861

I am not sure whether the problem is with the test or with the logic itself (I copied the E2E test from the `DataStreamConversionUtilsTest` and I think we should expect feature parity between the two).


It could've been a mistake in the test case itself. Please kindly take a look and see if this is suppose to be a test problem or a conversion logic problem. if (1) test problem, then we need to fix the test and simultaneously fix the heuristic in the `DataStreamConversionUtils`; (2) if it is in fact a conversion problem we should remove the heuristic in both Utils. Please let me know .

qiuxiafei: There're two lines confuse me in `testE2E`, 
  1. The line `thrown.expect(ValidationException.class);` , why is an `ValidationException` expected?
  2. The line `env.execute();` seems have nothing to execute because of the previous `print()` ,am I right?

So I just comment them out, add the heuristic lines back and move ahead. I do reproduce the exception you mentioned, in the 3rd `toTable` invocation. I think that's because type information has been added to `input` in the 2nd invocation already. Flink complains about this when the same thing applied to the identical `input` object in the 3rd invocation. So I think at least it proves that our heuristic works. And it also reveals a corner case we missed, in which an identical `DataSet` object is passed to `toTable` more than once.  UT will pass after adding `IllegalStateException` to catch block. The new `toTable` mothod now looks like:
```java
	public static Table toTable(MLEnvironment session, DataSet <Row> data, String[] colNames, TypeInformation <?>[] colTypes) {
		try {
			if (data instanceof SingleInputUdfOperator) {
				((SingleInputUdfOperator) data).returns(new RowTypeInfo(colTypes, colNames));
			} else if (data instanceof TwoInputUdfOperator) {
				((TwoInputUdfOperator) data).returns(new RowTypeInfo(colTypes, colNames));
			}
			return toTable(session, data, colNames);
		} catch (ValidationException | IllegalStateException ex) {
			// currently ValidationException will be thrown and caught for further processing.
			// Because the getType() API of the Transformation can only be accessed once.
			// This can be improve if we add "isTypeSet()" API to the Transformation class.
			if (null == colTypes) {
				throw ex;
			} else {
				DataSet <Row> t = fallbackToExplicitTypeDefine(data, colNames, colTypes);
				return toTable(session, t, colNames);
			}
		}
```

I just wanna make sure it ok to add another exception to the origin anti-pattern code ^_^ ..

walterddr: I think the 3rd `toTable` was NOT suppose to work at all (e.g. the same DataStream/DataSet object should not be converted to table twice)

If so I think we should even remove the 3rd `toTable` operation all together instead of adding another catcher - especially because `IllegalStateException` is a `java.lang` exception instead of a Flink defined exception, there might be other ways it was thrown.

flink-ml-parent/flink-ml-lib/src/test/java/org/apache/flink/ml/common/utils/DataStreamConversionUtilTest.java
```diff
@@ -0,0 +1,101 @@
...
+
+		DataStream<Row> input = env.fromElements(Row.of("a"));
+
+		Table table1 = DataStreamConversionUtil.toTable(MLEnvironmentFactory.DEFAULT_ML_ENVIRONMENT_ID, input, new String[]{"word"});
+		Assert.assertEquals(
+			new TableSchema(new String[]{"word"}, new TypeInformation[]{TypeInformation.of(String.class)}),
+			table1.getSchema()
+		);
+
+		input = input.map(new GenericTypeMap());
```
walterddr: is this conversion suppose to remove the "string" type within the content: `Row.of("a")`?

flink-ml-parent/flink-ml-lib/src/test/java/org/apache/flink/ml/common/utils/DataStreamConversionUtilTest.java
```diff
@@ -0,0 +1,101 @@
...
+
+		Table table2 = DataStreamConversionUtil.toTable(
+			MLEnvironmentFactory.DEFAULT_ML_ENVIRONMENT_ID,
+			input,
+			new String[]{"word"},
+			new TypeInformation[]{TypeInformation.of(Integer.class)}
```
walterddr: `Integer.class` cannot be case to `String.class` obviously, however this `toTable` call is not throwing any exception, is this expected?

flink-ml-parent/flink-ml-lib/src/test/java/org/apache/flink/ml/common/utils/DataStreamConversionUtilTest.java
```diff
@@ -0,0 +1,101 @@
...
+		);
+
+		Assert.assertEquals(
+			new TableSchema(new String[]{"word"}, new TypeInformation[]{TypeInformation.of(Integer.class)}),
+			table2.getSchema()
+		);
+
+		Table tableFromDataStreamWithTableSchema = DataStreamConversionUtil.toTable(
+			MLEnvironmentFactory.DEFAULT_ML_ENVIRONMENT_ID,
+			input,
```
walterddr: Assuming this is used for testing passing in a TableSchema instead of colNames and colTypes. should we do another GenericTypeMap() conversion, like this?
```suggestion
			input.map(new GenericTypeMap()),
```

flink-ml-parent/flink-ml-lib/src/test/java/org/apache/flink/ml/common/utils/DataStreamConversionUtilTest.java
```diff
@@ -0,0 +1,101 @@
...
+			)
+		);
+
+		Assert.assertEquals(
+			new TableSchema(new String[]{"word"}, new TypeInformation[]{TypeInformation.of(Integer.class)}),
+			tableFromDataStreamWithTableSchema.getSchema()
+		);
+
+		thrown.expect(ValidationException.class);
+		DataStreamConversionUtil.toTable(MLEnvironmentFactory.DEFAULT_ML_ENVIRONMENT_ID, input, new String[]{"f0"});
```
walterddr: is this exception expected to be thrown because of "f0" not found? the exception thrown here is:
```
An input of GenericTypeInfo<Row> cannot be converted to Table. Please specify the type of the input with a RowTypeInfo.
```
instead of not found.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/DataSetConversionUtil.java
qiuxiafei: > If so I think we should even remove the 3rd toTable operation all together instead of adding another catcher - especially because IllegalStateException is a java.lang exception instead of a Flink defined exception, there might be other ways it was thrown.

make sence.

flink-ml-parent/flink-ml-lib/src/test/java/org/apache/flink/ml/common/utils/DataStreamConversionUtilTest.java
qiuxiafei: Yes, this removes type information on the original datastream.

flink-ml-parent/flink-ml-lib/src/test/java/org/apache/flink/ml/common/utils/DataStreamConversionUtilTest.java
qiuxiafei: This is intended, just another different type from the original one to make the following assert explicit. It throws no exception since it won't be executed.

flink-ml-parent/flink-ml-lib/src/test/java/org/apache/flink/ml/common/utils/DataStreamConversionUtilTest.java
qiuxiafei: Sorry for the misleading, `"f0"` should be `"word"`. The same exception will raise because type information of the column is not specified.

flink-ml-parent/flink-ml-lib/src/test/java/org/apache/flink/ml/common/utils/DataStreamConversionUtilTest.java
walterddr: Got it. I am assuming this converts a "generic row" input datastream/set into a typed table. which should not throw validation exception because we do not know it is going to pass in the data `Row.of("a")` until runtime. is that correct?

flink-ml-parent/flink-ml-lib/src/test/java/org/apache/flink/ml/common/utils/DataStreamConversionUtilTest.java
walterddr: I am going ahead and assume this is right.

flink-ml-parent/flink-ml-lib/src/test/java/org/apache/flink/ml/common/utils/DataStreamConversionUtilTest.java
qiuxiafei: Yes.

qiuxiafei:
Thanks for your work, especially the improvements on tests. It looks great to me. We can then come back to refine these bunches of code when the JIRAs are done.

walterddr:
thanks @qiuxiafei for the contribution. I've added https://issues.apache.org/jira/browse/FLINK-15475 to track the improvement on the API front.

PR #9375, opened 2019-08-06
[FLINK-13159] Prevent potential NPE of restored PojoSerializer when deserializing

Myasuka:

Prevent potential NPE when PojoSerializer restored.

## Brief change log

  - Ensure the `classloader` could be assigned when `PojoSerializer` restored.

## Verifying this change

  - Added test `testRestoreSerializerDeserialize` into `SerializerTestBase` to ensure all restored serializes could deserialize values as expected.

  - The serializers: **yes**
## Documentation

### DIFF
--- a/flink-core/src/main/java/org/apache/flink/api/java/typeutils/runtime/PojoSerializer.java
+++ b/flink-core/src/main/java/org/apache/flink/api/java/typeutils/runtime/PojoSerializer.java
@@ -123,7 +123,7 @@ public final class PojoSerializer<T> extends TypeSerializer<T> {
 			this.fields[i].setAccessible(true);
 		}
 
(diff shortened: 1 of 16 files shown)


tzulitai:
@flinkbot approve all

tzulitai:
@Myasuka thanks for the fast fix.
As discussed offline, I've reverted the changes to `SerializerTestBase` since the requirement of equality checks on the test data objects is out of scope of this fix, and would require a more proper discussion on how we want to do that.

On top of your fixes, I've extended the `PojoSerializerMigrationTest` to cover the NPE case.
Those tests fail without your fix, so that should confirm that the fix is correct.

Will merge once Travis is green:
https://github.com/tzulitai/flink/tree/FLINK-13159-premerge

PR #9377, opened 2019-08-07
[FLINK-13561][table-planner-blink] Verify and correct time function's semantic for Blink planner

wuchong:

Fix behavior of several datetime builtin functions in blink planner.

## Brief change log

1. af3db19573d91acae9cc15e2186680b49b659fdd  Fix NOW() should return TIMESTAMP instead of BIGINT
    This aligns the behavior to other systems (MySQL, Spark). Because NOW() is Synonyms for CURRENT_TIMESTAMP.

2. d3af9e280d705e51c1f3bc16908e5650c7a639ea  Fix UNIX_TIMESTAMP(string [,format]) should work in session time zone
    This aligns the behavior to other systems (MySQL, Spark). UNIX_TIMESTAMP(string [,format]) is an inverse of FROM_UNIXTIME(bigint [,format]). We also remove the support of UNIX_TIMESTAMP(timestamp) in this commit.

3. 5939a990cd86fc16141a39587e565988734162bc  Fix FROM_UNIXTIME(bigint [,format]) should work in session time zone
    This aligns the behavior to other systems (MySQL, Spark).

4. 584df370a4feb34b1ef1bdb4a81b60809dad7128  Drop TO_DATE(int) function support
    This commit drops TO_DATE(int) function support in blink planner to align with other systems. We only support TO_DATE(string [,format]) in this version.

5. 1a590b6d9c792b136a109cc5f3ad1ab3bf0f6916  Drop TO_TIMESTAMP(bigint) function support
    This commit drops TO_TIMESTAMP(bigint) function support in blink planner to align with other systems. We only support TO_TIMESTAMP(string [,format]) in this version.

6. 7dfb34225535a673c99baa4cd7b7dd0ed8a34783  Drop CONVERT_TZ(timestamp, format, from_tz, to_tz) function support
    This commit drops CONVERT_TZ(timestamp, format, from_tz, to_tz) function support in blink planner to align with other systems. We only support CONVERT_TZ(timestamp, from_tz, to_tz) in this version.

7. c2b09aa4f5c00eb5436e5f2e4ed1983657c9753e  Drop DATE_FORMAT(timestamp, from_format, to_format) function support
    This commit drops DATE_FORMAT(timestamp, from_format, to_format) function support in blink planner to align with other systems. We only support DATE_FORMAT(timestamp, to_format) and DATE_FORMAT(string, to_format) in this version.

8. 84be6c933af6b8a960df17f6767d620db7f3a59f  Remove some builtin datetime functions which can be covered by existing functions
    Removes DATE_FORMAT_TZ, DATE_ADD,DATE_SUB, DATEDIFF, FROM_TIMESTAMP, TO_TIMESTAMP_TZ builtin functions which can be covered by existing functions.

## Verifying this change

This change is already covered by existing tests. For the functions changed behavior, we modified existing tests and add some more tests.

## Documentation

### DIFF
--- a/flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/functions/sql/FlinkSqlOperatorTable.java
+++ b/flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/functions/sql/FlinkSqlOperatorTable.java
@@ -782,10 +782,7 @@ public class FlinkSqlOperatorTable extends ReflectiveSqlOperatorTable {
 		SqlKind.OTHER_FUNCTION,
 		ReturnTypes.cascade(ReturnTypes.explicit(SqlTypeName.VARCHAR), SqlTypeTransforms.FORCE_NULLABLE),
 		null,
-		OperandTypes.or(
-			OperandTypes.family(SqlTypeFamily.STRING, SqlTypeFamily.STRING, SqlTypeFamily.STRING),
-			OperandTypes.family(SqlTypeFamily.STRING, SqlTypeFamily.STRING,
-				SqlTypeFamily.STRING, SqlTypeFamily.STRING)),
+		OperandTypes.family(SqlTypeFamily.STRING, SqlTypeFamily.STRING, SqlTypeFamily.STRING),
 		SqlFunctionCategory.TIMEDATE);
 
 	public static final SqlFunction LOCATE = new SqlFunction(
--- a/flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/functions/sql/FlinkSqlOperatorTable.java
+++ b/flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/functions/sql/FlinkSqlOperatorTable.java
@@ -761,9 +761,8 @@ public class FlinkSqlOperatorTable extends ReflectiveSqlOperatorTable {
 		ReturnTypes.cascade(ReturnTypes.explicit(SqlTypeName.TIMESTAMP), SqlTypeTransforms.FORCE_NULLABLE),
 		null,
 		OperandTypes.or(
-			OperandTypes.family(SqlTypeFamily.NUMERIC),
-			OperandTypes.family(SqlTypeFamily.STRING),
-			OperandTypes.family(SqlTypeFamily.STRING, SqlTypeFamily.STRING)),
+			OperandTypes.family(SqlTypeFamily.CHARACTER),
+			OperandTypes.family(SqlTypeFamily.CHARACTER, SqlTypeFamily.CHARACTER)),
 		SqlFunctionCategory.TIMEDATE);
 
 	public static final SqlFunction TO_DATE = new SqlFunction(
--- a/flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/functions/sql/FlinkSqlOperatorTable.java
+++ b/flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/functions/sql/FlinkSqlOperatorTable.java
@@ -632,8 +632,7 @@ public class FlinkSqlOperatorTable extends ReflectiveSqlOperatorTable {
 		OperandTypes.or(
 			OperandTypes.NILADIC,
 			OperandTypes.family(SqlTypeFamily.STRING),
-			OperandTypes.family(SqlTypeFamily.STRING, SqlTypeFamily.STRING),
-			OperandTypes.family(SqlTypeFamily.TIMESTAMP)),
+			OperandTypes.family(SqlTypeFamily.STRING, SqlTypeFamily.STRING)),
 		SqlFunctionCategory.TIMEDATE) {
 
 		@Override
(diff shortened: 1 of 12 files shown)


wuchong:
CI passed in my own travis: https://travis-ci.org/wuchong/flink/builds/568707905

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/functions/sql/FlinkSqlOperatorTable.java
```diff
@@ -84,7 +85,7 @@ public static synchronized FlinkSqlOperatorTable instance() {
 	private static final SqlReturnTypeInference PROCTIME_TYPE_INFERENCE = createTimeIndicatorReturnType(false);
 
 	private static SqlReturnTypeInference createTimeIndicatorReturnType(boolean isRowTime) {
-		return ReturnTypes.explicit(factory -> {
```
JingsongLi: I think we should keep `ReturnTypes`, there are `ReturnTypes` and `InferTypes` and `OperandTypes`. It is confused if we remove class name.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/codegen/CodeGeneratorContext.scala
```diff
@@ -518,17 +518,6 @@ class CodeGeneratorContext(val tableConfig: TableConfig) {
     DEFAULT_TIMEZONE_TERM
   }
 
-  /**
-    * Adds a reusable Time ZoneId to the member area of the generated class.
-    */
-  def addReusableTimeZoneID(): String = {
```
JingsongLi: I think we can keep it... in future, we should use zoneID instead of timeZone.

flink-table/flink-table-planner-blink/src/test/scala/org/apache/flink/table/planner/expressions/ScalarFunctionsTest.scala
```diff
@@ -3971,7 +3971,7 @@ class ScalarFunctionsTest extends ScalarTypesTestBase {
   }
 
   @Test
-  def testFromUnixTimeInUTC(): Unit = {
+  def testFromUnixTime(): Unit = {
```
JingsongLi: should add FromUnixTime cases to `TemporalTypesTest`, there are many time zone env, like shanghai, DaylightSaving and etc.
Same to UnixTimestamp.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/codegen/calls/ScalarOperatorGens.scala
```diff
@@ -168,8 +168,15 @@ object ScalarOperatorGens {
         generateBinaryArithmeticOperator(ctx, op, left.resultType, left, right)
 
       case (DATE, INTERVAL_DAY_TIME) =>
-        generateOperatorIfNotNull(ctx, new DateType(), left, right) {
-          (l, r) => s"$l $op ((int) ($r / ${MILLIS_PER_DAY}L))"
+        resultType.getTypeRoot match {
```
JingsongLi: What is for?
Why in `[FLINK-13561][table-planner-blink] Drop TO_TIMESTAMP(bigint) function`?

wuchong: It is in order to fix `timestampadd(hour, 1, date '2019-08-08')` which will get `2019-08-08` as result.

I'm sorry I messed it in `to_timestamp` commit. I should split it out.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/codegen/CodeGeneratorContext.scala
wuchong: I would like to remove dead code and introduce them when we need them. Furthermore, these code does't work, for example, it uses the same member field name with timeZone, it doesn't add code to member are.

wuchong:
Thanks @JingsongLi  for the reviewing. I have updated the PR. I think the commit message explains the changes. Please have a look again when you are free.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/codegen/CodeGeneratorContext.scala
JingsongLi: OK

JingsongLi:
LGTM +1

wuchong:
Thanks @JingsongLi  for the reviewing. 

The travis failed on an unrelated case which has been fixed FLINK-13605. 
https://travis-ci.com/flink-ci/flink/builds/122406398

Will merge this.

wuchong:
Merged with commits:

28260cdd70aee0a4169a068887feae71a81ec4c0
e8fd892220f6575ef99339d274b812202e687f93
d01cfb0cdedea79b2b946b7c3367b29a787d3e79
b41b52a6d1a18eafa113de3abd14da519de8cf2f
80f17d5b16b139023ee63c1251491075000157f7
8dbc64105c13dfae467f2cfb75ec67d0d2d56c84
bde51f9e46600154bd60321d3d345bb539b30ed1
c365780e7527a917e9d127694cd80cf2cdabf247
499a0f49de47c61ad8ef6341402756468d25f680

PR #9382, opened 2019-08-07
[FLINK-13600][table] Rework TableEnvironment.connect() class hierarchy

twalthr:

This fixes the currently broken `TableEnvironment.connect()` API by reworking the descriptor class hierarchy. `TableDescriptor` now contains most of the logic and `ConnectTableDescriptor` is the return type of table environments (expect for legacy Batch/StreamTableDescriptors).

The API is still mostly source code compatible.

## Brief change log

- Return `ConnectTableDescriptor`
- Merge ConnectorFormatDescriptor, StreamableDescriptor, and TableDescriptor
- Merge RegistrableDescriptor and ConnectTableDescriptor

## Verifying this change

A new `TableEnvironmentTest` tests the implementation.

  - The serializers: no

## Documentation


flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/descriptors/ConnectTableDescriptor.java
```diff
@@ -80,51 +87,23 @@ public void registerTableSink(String name) {
...
-	public D withSchema(Schema schema) {
-		schemaDescriptor = Optional.of(Preconditions.checkNotNull(schema));
-		return (D) this;
-	}
-
 	/**
 	 * Converts this descriptor into a set of properties.
 	 */
 	@Override
 	public Map<String, String> toProperties() {
```
dawidwys: How about we make this method a bit safer to extend. Right now it adds implicit contract that you have to call `super.toProperties` in a child class.

Maybe we could change it to:
```
public final Map<String, String> toProperties() {
  ...
  properties.asMap().putAll(extraProperties());
}

protected abstract Map<String, String> extraProperties();
```

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/descriptors/StreamTableDescriptor.java
```diff
@@ -21,92 +21,18 @@
...
- * Descriptor for specifying a table source and/or sink in a streaming environment.
+ * Describes a table connected from a streaming environment.
  */
 @PublicEvolving
-public class StreamTableDescriptor
-	extends ConnectTableDescriptor<StreamTableDescriptor>
-	implements StreamableDescriptor<StreamTableDescriptor> {
-
-	private Optional<String> updateMode = Optional.empty();
+public final class StreamTableDescriptor extends ConnectTableDescriptor {
```
dawidwys: Probably we can't do it because of backwards compatibility, but I will put it here anyway.

Do we even need the `StreamTableDescriptor`? It has no additional methods compared to `ConnectTableDescriptor`

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/descriptors/ConnectTableDescriptor.java
dawidwys: I just saw it's done this way already for `FormatDescriptor`

flink-table/flink-table-api-java/src/test/java/org/apache/flink/table/utils/TableEnvironmentMock.java
```diff
@@ -0,0 +1,91 @@
...
+			boolean isStreamingMode) {
+		super(catalogManager, tableConfig, executor, functionCatalog, planner, isStreamingMode);
+
+		this.catalogManager = catalogManager;
+		this.executor = executor;
+		this.functionCatalog = functionCatalog;
+		this.planner = planner;
+	}
+
+	public static TableEnvironmentMock getInstance(boolean isStreamingMode) {
```
dawidwys: nit: `getInstance(Boolean)` -> `getStreamingInstance` + `getBatchInstance`

flink-table/flink-table-common/src/main/java/org/apache/flink/table/descriptors/TableDescriptor.java
```diff
@@ -19,11 +19,132 @@
...
+import static org.apache.flink.table.descriptors.StreamTableDescriptorValidator.UPDATE_MODE_VALUE_UPSERT;
 
 /**
-  * Common class for all descriptors describing table sources and sinks.
-  */
+ * Describes a table consisting of a connector (in a given update mode) and a format.
+ */
 @PublicEvolving
-public abstract class TableDescriptor extends DescriptorBase {
+public abstract class TableDescriptor<D extends TableDescriptor<D>> extends DescriptorBase {
```
dawidwys: Do we need both `TableDescriptor` & `ConnectTableDescriptor`?

flink-table/flink-table-common/src/main/java/org/apache/flink/table/descriptors/TableDescriptor.java
```diff
@@ -19,11 +19,132 @@
...
+	public D inUpsertMode() {
+		updateMode = UPDATE_MODE_VALUE_UPSERT;
+		return (D) this;
+	}
+
+	/**
+	 * Converts this descriptor into a set of properties.
+	 */
+	@Override
+	public Map<String, String> toProperties() {
```
dawidwys: The same problem as for `ConnectTableDescriptor`. Adds implicit contract.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/descriptors/StreamTableDescriptor.java
twalthr: Backwards compatibility is the answer. I will add a comment to the class.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/descriptors/TableDescriptor.java
twalthr: TableDescriptor is used in `CatalogTableBuilder`. The main reason of this class is the generic that is not present in `ConnectTableDescriptor`.

twalthr:
Merging...

PR #9386, opened 2019-08-07
[FLINK-13601][tests] Harden RegionFailoverITCase by recording info when checkpoint just completed

Myasuka:
# What is the purpose of the change

Harden `RegionFailoverITCase` by recording info when checkpoint just completed to avoid the race condition in [FLINK-13601].

## Brief change log

  - Harden `RegionFailoverITCase`.

## Verifying this change

  - We use `TestingCompletedCheckpointStore` to record information of checkpoints instead of previous `notifyCheckpointComplete` invoked on task side. This could avoid the race condition in [FLINK-13601](https://issues.apache.org/jira/browse/FLINK-13601).

  - The serializers: **no**

## Documentation

### DIFF
--- a/flink-tests/src/test/java/org/apache/flink/test/checkpointing/RegionFailoverITCase.java
+++ b/flink-tests/src/test/java/org/apache/flink/test/checkpointing/RegionFailoverITCase.java
@@ -442,6 +442,9 @@ public class RegionFailoverITCase extends TestLogger {
 		}
 	}
 
+	/**
+	 * Testing HA factory which needs to be public in order to be instantiatable.
+	 */
 	public static class TestingHAFactory implements HighAvailabilityServicesFactory {
 
 		@Override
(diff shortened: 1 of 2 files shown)


Myasuka:
@flinkbot attention @aljoscha

flink-tests/src/test/java/org/apache/flink/test/checkpointing/RegionFailoverITCase.java
```diff
@@ -102,8 +111,14 @@
 
 	@Before
 	public void setup() throws Exception {
+		HighAvailabilityServicesUtilsTest.TestHAFactory.haServices = new TestingHaServices(
```
zentol: please don't re-use the TestHAFactory class; with it's static mutable field it should be avoided if possible.
You can just define your own `HighAvailabilityServicesFactory` with a lambda
`HighAvailabilityServicesFactory haServiceFactory = (config, executor) -> new TestingHaServices...`

Myasuka:
@zentol Would you please take a look again at this PR?

zentol:
What would happen if a task starts checkpointing (execution `snapshotState`, modifying `snapshotIndicesOfSubTask`), but fails before the checkpoint is completed?

Myasuka:
@zentol  If current `lastCompletedCheckpointId` is `chk-3`, when task starts checkpointing, it would put `snapshotIndicesOfSubTask` with newer checkpoint `chk-4`. Then, the job fails before the checkpoint is completed, and the `chk-4` would not update in `TestingCompletedCheckpointStore` as `lastCompletedCheckpointId`. 
When the job recovers, it would verify with previous last checkpoint id `chk-3` from `snapshotIndicesOfSubTask` instead of `chk-4` [(ref here)](https://github.com/apache/flink/pull/9386/files#diff-bfb5100206bccf135ef75dc3df04ff88L310) , which means we could still verify the state correctness as well.

PR #9387, opened 2019-08-07
[FLINK-13231] [pubsub] Replace Max outstanding acknowledgement ids li…

Xeli:

Follow up on PubSub connector PR. This replaces the max acknowledgement id limit by a FlinkConnectorRateLimiter

## Verifying this change

I've made sure this change is covered by existing unit tests.

  - The public API, i.e., is any changed class annotated with `@Public(Evolving)`: Yes, unless we can add this to 1.9. The pubsub connector has not yet been released.
  - The serializers: no
  - The runtime per-record code paths (performance sensitive): Yes, but only for the pubsub connector, not the flink core code.

## Documentation

### DIFF
--- a/flink-connectors/flink-connector-gcp-pubsub/src/main/java/org/apache/flink/streaming/connectors/gcp/pubsub/PubSubSource.java
+++ b/flink-connectors/flink-connector-gcp-pubsub/src/main/java/org/apache/flink/streaming/connectors/gcp/pubsub/PubSubSource.java
@@ -18,6 +18,8 @@
 package org.apache.flink.streaming.connectors.gcp.pubsub;
 
 import org.apache.flink.api.common.functions.RuntimeContext;
+import org.apache.flink.api.common.io.ratelimiting.FlinkConnectorRateLimiter;
+import org.apache.flink.api.common.io.ratelimiting.GuavaFlinkConnectorRateLimiter;
 import org.apache.flink.api.common.serialization.DeserializationSchema;
 import org.apache.flink.api.common.typeinfo.TypeInformation;
 import org.apache.flink.api.java.typeutils.ResultTypeQueryable;
@@ -40,8 +42,6 @@ import com.google.cloud.pubsub.v1.Subscriber;
 import com.google.pubsub.v1.ProjectSubscriptionName;
 import com.google.pubsub.v1.PubsubMessage;
 import com.google.pubsub.v1.ReceivedMessage;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.io.Serializable;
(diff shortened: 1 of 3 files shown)


flink-connectors/flink-connector-gcp-pubsub/src/main/java/org/apache/flink/streaming/connectors/gcp/pubsub/PubSubSource.java
```diff
@@ -92,6 +93,10 @@ public void open(Configuration configuration) throws Exception {
 
 		getRuntimeContext().getMetricGroup().gauge("PubSubMessagesProcessedNotAcked", this::getOutstandingMessagesToAck);
 
+		//convert per-subtask-limit to global rate limit, as FlinkConnectorRateLimiter::setRate expects a global rate limit.
+		rateLimiter.setRate(messagePerSecondRateLimit * getRuntimeContext().getNumberOfParallelSubtasks());
```
becketqin: I am not sure if this is the correct logic. I think the rate limiter mentioned it is a "global" rate limit meaning the configuration itself is applied globally. But the effect of rate limit  is still per subtask. I don't think there is any communication in the rate limiter among different subtasks to ensure it is a rate limit shared by all the subtasks.

With the current code, if I have 2 subtasks. And I want to have each subtask to consume 100 records/second. I have to set the config to 50, this gives me 100 messages per subtask and 200 messages/s for the entire job. That seems very confusing.

becketqin: Talked to @Xeli offline. Actually the rate limit is a global rate as the value was divided by the parallelism in the rate limiter implementation. So the current patch does set the per task rate.

flink-connectors/flink-connector-gcp-pubsub/src/test/java/org/apache/flink/streaming/connectors/gcp/pubsub/PubSubSourceTest.java
```diff
@@ -120,10 +124,15 @@ public void testProcessMessage() throws Exception {
...
-		pubSubSource.processMessage(sourceContext, asList(receivedMessage("firstAckId", pubSubMessage(FIRST_MESSAGE)),
-															receivedMessage("secondAckId", pubSubMessage(SECOND_MESSAGE))));
+		List<ReceivedMessage> receivedMessages = asList(
+			receivedMessage("firstAckId", pubSubMessage(FIRST_MESSAGE)),
+			receivedMessage("secondAckId", pubSubMessage(SECOND_MESSAGE))
+		);
+		pubSubSource.processMessage(sourceContext, receivedMessages);
 
 		//verify handling messages
+		verify(rateLimiter, times(1)).acquire(receivedMessages.get(0).getSerializedSize() + receivedMessages.get(1).getSerializedSize());
```
becketqin: This test was left from the previous commit while the rate limit was still on bytes. I fixed it when merging the code.

flink-connectors/flink-connector-gcp-pubsub/src/main/java/org/apache/flink/streaming/connectors/gcp/pubsub/PubSubSource.java
```diff
@@ -264,12 +263,11 @@ private PubSubSourceBuilder(PubSubDeserializationSchema<OUT> deserializationSche
...
 		/**
-		 * Set a limit of the number of outstanding or to-be acknowledged messages.
-		 * default is 10000. Adjust this if you have high checkpoint intervals and / or run into memory issues
-		 * due to the amount of acknowledgement ids. Use {@link PubSubSource}.NO_MAX_MESSAGES_TO_ACKNOWLEDGE_LIMIT if you want to remove the limit.
+		 * Set a limit on the rate of messages per second received. This limit is per parallel instance of the source function.
+		 * Default is set to 100000 messages per second.
 		 */
-		public PubSubSourceBuilder<OUT> withMaxMessageToAcknowledge(int maxMessageToAcknowledge) {
-			this.maxMessageToAcknowledge = maxMessageToAcknowledge;
+		public PubSubSourceBuilder<OUT> withRateLimit(int messagePerSecondRateLimit) {
```
becketqin: I renamed this method to `withMessageRateLimit` in case we will have bytes rate limit in the future.

becketqin:
Merged 
1.9 b3ce2c76332da287ccf05d291fedb27517dd8c32
master f539300a03c101b4370d58186297d1d55829278f

PR #9388, opened 2019-08-07
[FLINK-13230] [pubsub] Add retry mechanism to acknowledging pubsub me…

Xeli:

Follow up on PubSub connector PR. This adds a retry mechanism to acknowledging pubsub messages. Google's PubSub is unstable enough for this to be a must-have imo.

## Verifying this change

Trivial change.

  - The public API, i.e., is any changed class annotated with `@Public(Evolving)`: Yes, unless we can add this to 1.9. The pubsub connector has not yet been released.
  - The serializers: no
  - The runtime per-record code paths (performance sensitive): Yes, but only for the pubsub connector, not the flink core code.

## Documentation

### DIFF
--- a/flink-connectors/flink-connector-gcp-pubsub/src/main/java/org/apache/flink/streaming/connectors/gcp/pubsub/BlockingGrpcPubSubSubscriber.java
+++ b/flink-connectors/flink-connector-gcp-pubsub/src/main/java/org/apache/flink/streaming/connectors/gcp/pubsub/BlockingGrpcPubSubSubscriber.java
@@ -95,9 +95,21 @@ public class BlockingGrpcPubSubSubscriber implements PubSubSubscriber {
 									.addAllAckIds(splittedAckIds.f0)
 									.build();
 
-			stub.withDeadlineAfter(60, SECONDS).acknowledge(acknowledgeRequest);
-
+			acknowledgeWithRetries(acknowledgeRequest, retries);
 			splittedAckIds = splitAckIds(splittedAckIds.f1);
+		}
+	}
+
+	private void acknowledgeWithRetries(AcknowledgeRequest acknowledgeRequest, int retriesRemaining) {
+		try {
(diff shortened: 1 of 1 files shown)


Xeli:
@becketqin as you've looked at the pubsub code the most: Would you be so kind to have a look at this PR? We can then close https://issues.apache.org/jira/browse/FLINK-13083

flink-connectors/flink-connector-gcp-pubsub/src/main/java/org/apache/flink/streaming/connectors/gcp/pubsub/BlockingGrpcPubSubSubscriber.java
```diff
@@ -95,12 +95,24 @@ public void acknowledge(List<String> acknowledgementIds) {
...
 			splittedAckIds = splitAckIds(splittedAckIds.f1);
 		}
 	}
 
+	private void acknowledgeWithRetries(AcknowledgeRequest acknowledgeRequest, int retriesRemaining) {
+		try {
+			stub.withDeadlineAfter(timeout.toMillis(), TimeUnit.MILLISECONDS).acknowledge(acknowledgeRequest);
+		} catch (StatusRuntimeException e) {
+			if (retriesRemaining > 0) {
+				acknowledgeWithRetries(acknowledgeRequest, retriesRemaining - 1);
```
becketqin: We probably don't want to have recursion in the retry logic. Can we replace it with a for loop instead?

RyanSkraba:
The changes in this PR have been applied (with attribution to @Xeli!) to the new FLIP-27 Pub/Sub connector in https://github.com/apache/flink-connector-gcp-pubsub/pull/2, and this PR might be closeable.  What do you think?

snuyanzin:
thanks @RyanSkraba 
yes, since pubsub connector is externalised it makes sense to close it

PR #9389, opened 2019-08-08
[FLINK-13645][table-planner] Error in code-gen when using blink planner in scala shell

zjffdu:

This is a trivial PR to fix the issue when using blink planner in scala shell. The root cause is that we didn't use the right ClassPathLoader. This PR just fix it.

## Brief change log

Set the ClassPathLoader properly.

## Verifying this change

This change is a trivial rework / code cleanup without any test coverage. I verify it manually in apache zeppelin which uses scala shell related code.

  - The serializers: ( no )
  - Anything that affects deployment or recovery: JobManager (and its components), Checkpointing, Yarn/Mesos, ZooKeeper: (no )
  - The S3 file system connector: ( no )

## Documentation

### DIFF
--- a/flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/codegen/CodeGeneratorContext.scala
+++ b/flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/codegen/CodeGeneratorContext.scala
@@ -603,7 +603,7 @@ class CodeGeneratorContext(val tableConfig: TableConfig) {
     val byteArray = InstantiationUtil.serializeObject(obj)
     val objCopy: AnyRef = InstantiationUtil.deserializeObject(
       byteArray,
-      obj.getClass.getClassLoader)
+      Thread.currentThread().getContextClassLoader)
     references += objCopy
 
     reusableMemberStatements.add(s"private transient $fieldTypeTerm $fieldTerm;")


flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/codegen/CodeGeneratorContext.scala
```diff
@@ -603,7 +603,7 @@ class CodeGeneratorContext(val tableConfig: TableConfig) {
     val byteArray = InstantiationUtil.serializeObject(obj)
     val objCopy: AnyRef = InstantiationUtil.deserializeObject(
       byteArray,
-      obj.getClass.getClassLoader)
+      Thread.currentThread().getContextClassLoader)
     references += objCopy
```
danny0405: Can we give a test case to prove that the `Thread.currentThread().getContextClassLoader` is the right one to choose ?

zjffdu: The proper place to put the test code is module `flink-scala-shell`. But after some investigation, I just found that it is not easy to do that because currently scala shell doesn't' support switch planner. I may need to create another PR to support that, but that seems too complicated. 
Since this is a straightforward fix, I think it is OK without test code.  What do you think @danny0405

danny0405: I'm fine with this, there seems some compile error.

zjffdu:
@wuchong Could you help take a look at it ? Thanks

wuchong:
Waiting for travis passing: https://travis-ci.org/wuchong/flink/builds/569696637

PR #9394, opened 2019-08-08
[FLINK-13547][table-planner-blink] Verify and correct string function's semantic for Blink planner

docete:

Currently many string builtin functions in blink planner follow hive/spark semantics, which should keep compatible with old planner. And some non-standard functions(Blink planner intros) should be removed.

## Brief change log

- concat/concat_ws function (null treatment)
- substring function (follow calcite/flink)
- from_base64 should return string not binary
- intro truncate function to blink planner
- uuid should be no-argument (remove the one-argument version)
- length/jsonvalue/keyvalue/substr (non-standard function should be removed)
- md5/sha1/sha2/sha224/sha256/sha384/sha512(remove the two-arguments version)
- ascii (operand type should beSqlTypeFamily.CHARACTER)

## Verifying this change

## Documentation

### DIFF
--- a/flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/functions/sql/FlinkSqlOperatorTable.java
+++ b/flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/functions/sql/FlinkSqlOperatorTable.java
@@ -615,7 +615,7 @@ public class FlinkSqlOperatorTable extends ReflectiveSqlOperatorTable {
 	public static final SqlFunction FROM_BASE64 = new SqlFunction(
 		"FROM_BASE64",
 		SqlKind.OTHER_FUNCTION,
-		ReturnTypes.cascade(ReturnTypes.explicit(SqlTypeName.BINARY), SqlTypeTransforms.TO_NULLABLE),
+		ReturnTypes.cascade(ReturnTypes.explicit(SqlTypeName.VARCHAR), SqlTypeTransforms.TO_NULLABLE),
 		null,
 		OperandTypes.family(SqlTypeFamily.STRING),
 		SqlFunctionCategory.STRING);
--- a/flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/functions/sql/FlinkSqlOperatorTable.java
+++ b/flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/functions/sql/FlinkSqlOperatorTable.java
@@ -658,8 +658,8 @@ public class FlinkSqlOperatorTable extends ReflectiveSqlOperatorTable {
 		ARG0_VARCHAR_FORCE_NULLABLE,
 		null,
 		OperandTypes.or(
-			OperandTypes.family(SqlTypeFamily.STRING, SqlTypeFamily.INTEGER),
-			OperandTypes.family(SqlTypeFamily.STRING, SqlTypeFamily.INTEGER, SqlTypeFamily.INTEGER)
+			OperandTypes.family(SqlTypeFamily.CHARACTER, SqlTypeFamily.INTEGER),
+			OperandTypes.family(SqlTypeFamily.CHARACTER, SqlTypeFamily.INTEGER, SqlTypeFamily.INTEGER)
 		),
 		SqlFunctionCategory.STRING);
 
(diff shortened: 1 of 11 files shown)


flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/dataformat/BinaryStringUtil.java
```diff
@@ -749,21 +749,21 @@ public static BinaryString concat(Iterable<BinaryString> inputs) {
 	}
 
 	/**
-	 * Concatenates input strings together into a single string using the separator.
-	 * A null input is skipped. For example, concat(",", "a", null, "c") would yield "a,c".
+	 * <p>Concatenates input strings together into a single string using the separator.
+	 * Returns NULL If the separator is NULL.</p>
+	 *
+	 * <p>Note: CONCAT_WS() does not skip any empty strings, however it does skip any NULL values after
+	 * the separator. For example, concat(",", "a", null, "c") would yield "a,c".</p>
```
wuchong:
```suggestion
	 * the separator. For example, concat_ws(",", "a", null, "c") would yield "a,c".</p>
```

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/functions/sql/FlinkSqlOperatorTable.java
```diff
@@ -502,69 +482,56 @@ public void lookupOperatorOverloads(
...
 
 	public static final SqlFunction SHA2 = new SqlFunction(
 		"SHA2",
 		SqlKind.OTHER_FUNCTION,
 		VARCHAR_2000_NULLABLE,
 		null,
-		OperandTypes.or(
-			OperandTypes.family(SqlTypeFamily.STRING, SqlTypeFamily.INTEGER),
-			OperandTypes.family(SqlTypeFamily.STRING, SqlTypeFamily.STRING, SqlTypeFamily.INTEGER)),
+		OperandTypes.sequence("'(DATA, HASH_LENGTH)'",
```
wuchong:
```suggestion
		OperandTypes.sequence("'SHA2(<STRING>, <HASH_LENGTH>)'",
```

docete:
> Thanks @docete , I left some suggestions.
> 
> 1. Remove "SUBSTR" builtin function definition and tests?
Agree! Forgot to remove this non-standard function.

> 2. Should "MD5" and "SHA" take SqlTypeFamily.CHARACTER instead of SqlTypeFamily.STRING ?
Sure. But now both Blink planner and old planner use SqlTypeFamily.STRING. I will open [a new ticket](https://issues.apache.org/jira/browse/FLINK-13664) to trace and fix both side.

> 3. Could you add a detailed description what builtin function are changed and the reason for commit "Refactor non-standard functions"? You can take [28260cd](https://github.com/apache/flink/commit/28260cdd70aee0a4169a068887feae71a81ec4c0) as an example.
OK, will fix in next hours.

> 4. Add `testInvalidTruncate1` and `testInvalidTruncate2` to `ScalarFunctionsValidationTest`
I don't understand. U mean put the decimal related truncate tests to a single test function?

> 5. We should also support truncate TableAPI. Please add the following case to `SqlExpressionTest#testArithmeticFunctions`
> 
> ```
>     testSqlApi("truncate(42.345)", "42")
>     testSqlApi("truncate(cast(42.345 as decimal(2, 3)), 2)", "42.34")
> ```
> 
> Btw, please add a prefix to the append commit message to describe which commit should be squashed into. For example: "fixup concat commit: fix test case". So that the committer can help to squash your commits easier.
OK, will fix in next hours.

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/functions/sql/FlinkSqlOperatorTable.java
docete: Keep the same as old planner.

wuchong: I don't think we need to keep the same with old planner for this. This is just an exception message. And the change will make the exception more readable.

wuchong:
Merge with:
5e697cf148f2d4ea09b0117d4c22102de98440c4
2ae0e8a0762732b2a87dbb8ebadd933ebeb4b96a
0196a95c74cba8a9fb072385dbc7e5920d357d70
7e13f5b43f417b8e29e1964aa109a463b152991a

PR #9396, opened 2019-08-08
[FLINK-13473][table-blink] Add stream Windowed FlatAggregate support for blink planner

sunjincheng121:

This pull request ports window flatAggregate from Flink planner to Blink planner. 

The changes mainly include two modules: blink-planner and blink-runtime. The code of the API part can be reused as the two planner share the same API.

## Brief change log
This pr is further split into the following 4 commits:

1. Add plan support for table aggregate. [0e875a2](https://github.com/sunjincheng121/flink/commit/0e875a20091a2b53a1614702ced7a7c3c9e117fa)
  This commit adds RelNodes(Logical, FlinkLogical and StreamExec) and Rules in blink planner. The execution procedures are: 
  First, the WindowAggregate QueryOperation node in API level will be converted to a LogicalWindowTableAggregate relnode in `QueryOperationConverter`.
  Second, the LogicalWindowTableAggregate will be converted to a FlinkLogicalWindowTableAggregate by the rule in logical optimization phase.
  Third, the FlinkLogicalWindowTableAggregate will be converted to a StreamExecGroupWindowTableAggregate by the rule in physical optimization phase.

2. Add runtime support for table aggregate. [a1120036](https://github.com/sunjincheng121/flink/commit/a1120036ae215e7831f6cdcc848f4188abc63780) 
  This commit adds codegen and window operator support for window table aggregete. 
  The AggCodeGen generates a NamespaceTableAggsHandleFunction which is used in the window operator.
  For window operator, this commit refactors the previous window operator by adding a base window operator(`WindowOperator`) and two sub window operators(`AggregateWindowOperator` and `TableAggregateWindowOperator`).
  Additionally, to make the refactor possible, this commit also change the window processing function(`InternalWindowProcessFunction`) by changing the `getWindowAggregationResult` method to `prepareAggregateAccumulatorForEmit`.
  The `prepareAggregateAccumulatorForEmit` method returns the accumulator, with which the `AggregateWindowOperator` can use getValue to return result and the `TableAggregateWindowOperator` can use emitValue to return results.

3. Add window operator tests for window table aggregate. [686ab9df](https://github.com/sunjincheng121/flink/commit/686ab9df908a914255fcc7bd6b447e42c79e2a9f
) 
  Change the `WindowOperatorTest` into a parameterized test which can test both aggregate and table aggregate.
  To simplify the testing logic, the table aggregate outputs same value with the aggregate except that the table aggregate outputs two same records each time.
  
4. Support WindowTableAggregate in some MetadataHandler. [9a5651f1d](https://github.com/sunjincheng121/flink/commit/9a5651f1dee8a7d5da17bc4e54831af55d7002c4)
  MetadataHandler is used to provide Metadata of a relational expression. For example, provide the unique key information of a relnode. The information can be used during optimization.
  This commit adds WindowTableAggregate support in `FlinkRelMdColumnInterval`, `FlinkRelMdFilteredColumnInterval` and `FlinkRelMdModifiedMonotonicity`. The absence of WindowTableAggregate in other MetadataHandlers mainly because:
   
- The WindowTableAggregate can not provide uniqueness information which is required in FlinkRelMdColumnUniqueness, FlinkRelMdDistinctRowCount, FlinkRelMdUniqueGroups, FlinkRelMdUniqueKeys. 
- Most of the MetadataHandlers are not been used in streaming jobs，such as FlinkRelMdPercentageOriginalRows，FlinkRelMdPopulationSize, FlinkRelMdRowCount, etc.

BTW: I do not add the emitpolicy for window aggregate just we want align with Flink Planner.
## Verifying this change
 - WindowOperatorContractTest
 - WindowOperatorTest

  - The serializers: (no)

## Documentation


sunjincheng121:
@flinkbot attention  @JingsongLi @wuchong @hequn8128 @aljoscha I appreciate if you can have look at this PR :)

wuchong:
I would suggest to modify the component name in commit message a bit. How about:

[FLINK-13473][table-planner-blink] Support windowed TableAggregate in some MetadataHandle
[FLINK-13473][table-runtime-blink] Add tests for window operator
[FLINK-13473][table-blink] Add runtime support for windowed flatAggregat on blink planner
[FLINK-13473][table-planner-blink] Add plan support for windowed flatAggregate on blink planner

And for the pull request title, I would suggest to use "[table-blink]" because it doesn't contain API changes. 

What do you think?

sunjincheng121:
Thanks for the review @wuchong !
This commit will be squashed when merge, So I think [table-blink] is make senses for me.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/codegen/agg/AggsHandlerCodeGenerator.scala
```diff
@@ -866,10 +993,65 @@ class AggsHandlerCodeGenerator(
...
     checkNeededMethods(needEmitValue = true)
-    aggBufferCodeGens(0).asInstanceOf[ImperativeAggCodeGen].emitValue
+    val methodName = "emitValue"
+    ctx.startNewLocalVariableStatement(methodName)
+
+    val windowCode =
+      if (isWindow) {
+        // no need to bind input
+        val exprGenerator = new ExprCodeGenerator(ctx, INPUT_NOT_NULL)
+        var valueExprs = Seq[GeneratedExpression]()
```
JingsongLi: Just use `val valueExprs = getWindowExpressions(windowProperties)`? Because `windowProperties` is empty when `hasNamespace` is false.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/codegen/agg/AggsHandlerCodeGenerator.scala
```diff
@@ -906,6 +1088,7 @@ object AggsHandlerCodeGenerator {
   val COLLECTOR_TERM = "out"
   val MEMBER_COLLECTOR_TERM = "convertCollector"
   val CONVERT_COLLECTOR_TYPE_TERM = "ConvertCollector"
+  val KEY_TERM = "groupKey"
```
JingsongLi: maybe you can name it `NAMESPACE_TABLE_AGG_KEY`

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/plan/nodes/physical/stream/StreamExecGroupWindowTableAggregate.scala
```diff
@@ -0,0 +1,98 @@
...
+
+    val aggsHandler = aggCodeGenerator.generateNamespaceTableAggsHandler(
+      "GroupingWindowTableAggsHandler",
+      aggInfoList,
+      namedProperties.map(_.property),
+      getWindowClass(window))
+
+    val builder = getWindowOperatorBuilder(inputFields, timeIdx)
+    builder
+      .aggregate(aggsHandler, accTypes, aggValueTypes, windowPropertyTypes)
```
JingsongLi: Should window table agg support `withAllowedLateness`?

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/runtime/operators/window/AggregateWindowOperator.java
```diff
@@ -0,0 +1,186 @@
...
+import org.apache.flink.table.runtime.operators.window.assigners.WindowAssigner;
+import org.apache.flink.table.runtime.operators.window.triggers.Trigger;
+import org.apache.flink.table.types.logical.LogicalType;
+
+import static org.apache.flink.util.Preconditions.checkNotNull;
+
+/**
+ * A {@link WindowOperator} that dedicates for group aggregate.
+ *
+ * <p>When an element arrives it gets assigned a key using a {@link KeySelector} and it gets
```
JingsongLi: repeated comments? can we remove it?

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/runtime/operators/window/TableAggregateWindowOperator.java
```diff
@@ -0,0 +1,121 @@
...
+import org.apache.flink.table.runtime.generated.NamespaceTableAggsHandleFunction;
+import org.apache.flink.table.runtime.operators.window.assigners.WindowAssigner;
+import org.apache.flink.table.runtime.operators.window.triggers.Trigger;
+import org.apache.flink.table.types.logical.LogicalType;
+import org.apache.flink.util.Collector;
+
+/**
+ * A {@link WindowOperator} that dedicates for group table aggregate.
+ *
+ * <p>When an element arrives it gets assigned a key using a {@link KeySelector} and it gets
```
JingsongLi: repeated comments? can we remove it?

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/runtime/operators/window/WindowOperatorBuilder.java
```diff
@@ -181,24 +194,79 @@ public WindowOperatorBuilder aggregate(
...
+
 	public WindowOperatorBuilder withSendRetraction() {
 		this.sendRetraction = true;
 		return this;
 	}
 
 	public WindowOperator build() {
 		checkNotNull(trigger, "trigger is not set");
-		if (generatedAggregateFunction != null && generatedEqualiser != null) {
+		if (generatedTableAggregateFunction != null) {
```
JingsongLi: There are a lot of branches, can we abstract this builder to multiple builders?

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/codegen/agg/AggsHandlerCodeGenerator.scala
sunjincheng121: Sounds good, and how about NAMESPACE_TABLE_AGG_KEY_TERM. It looks more consistent with other variables which end with _TERM.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/plan/nodes/physical/stream/StreamExecGroupWindowTableAggregate.scala
sunjincheng121: ou are right. We can support withAllowedLateness for table aggregate. However, there are two reasons that it is not supported in this PR.
- This PR mainly dedicate to port table aggregate from flink to blink. It maybe good to keep both align with each other first.
- withAllowedLateness is not a simple change for table aggregate. We also need to think about how to support emitStrategy in Table API.
So, in this PR we just align with flink planner table aggregate, What do you think?

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/runtime/operators/window/AggregateWindowOperator.java
sunjincheng121: If we remove this paragraph, the next paragraph("Each pane gets its own instance...") would be hard to understand.
Maybe we can keep this paragraph, or remove these two paragraphs in the base class? What do you think?

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/runtime/operators/window/WindowOperatorBuilder.java
sunjincheng121: Good idea. We can make the WindowOperatorBuilder as a base builer and add two more builders for aggregate and table aggregate.
What do you think?

sunjincheng121:
@JingsongLi Thanks for the review, I have update the PR, and I also have been shared my throughs on your comments. :)
please let me know what do you think?

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/plan/nodes/physical/stream/StreamExecGroupWindowAggregateBase.scala
```diff
@@ -0,0 +1,345 @@
...
+      generator.needMerge(mergedAccOffset = 0, mergedAccOnHeap = false)
+    }
+    if (needRetraction) {
+      generator.needRetract()
+    }
+
+    generator
+  }
+
+  protected def enrichWindowOperatorBuilder(
```
wuchong: I would like to finish the build logic in one place, rather than passing a half-work builder here and there. 
Can we use a if-else to generate the window operator in the base class depends on whether it is a table aggregate? like what we do in `DataStreamGroupWindowAggregateBase`?

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/runtime/operators/window/AggregateWindowOperator.java
```diff
@@ -0,0 +1,186 @@
...
+import org.apache.flink.table.runtime.generated.NamespaceAggsHandleFunction;
+import org.apache.flink.table.runtime.generated.RecordEqualiser;
+import org.apache.flink.table.runtime.operators.window.assigners.WindowAssigner;
+import org.apache.flink.table.runtime.operators.window.triggers.Trigger;
+import org.apache.flink.table.types.logical.LogicalType;
+
+import static org.apache.flink.util.Preconditions.checkNotNull;
+
+/**
+ * A {@link WindowOperator} that dedicates for group aggregate.
```
aljoscha:
```suggestion
 * A {@link WindowOperator} for grouped window aggregates.
```

flink-table/flink-table-runtime-blink/src/main/java/org/apache/flink/table/runtime/operators/window/TableAggregateWindowOperator.java
```diff
@@ -0,0 +1,121 @@
...
+import org.apache.flink.table.dataformat.BaseRow;
+import org.apache.flink.table.runtime.generated.GeneratedNamespaceTableAggsHandleFunction;
+import org.apache.flink.table.runtime.generated.NamespaceTableAggsHandleFunction;
+import org.apache.flink.table.runtime.operators.window.assigners.WindowAssigner;
+import org.apache.flink.table.runtime.operators.window.triggers.Trigger;
+import org.apache.flink.table.types.logical.LogicalType;
+import org.apache.flink.util.Collector;
+
+/**
+ * A {@link WindowOperator} that dedicates for group table aggregate.
```
aljoscha:
```suggestion
 * A {@link WindowOperator} for grouped and windowed table aggregates.
```

sunjincheng121:
Hi @wuchong and @aljoscha, thanks a lot for your nice review. 👍 
Glad to see the nice suggestions. I have addressed all your comments and updated the PR. 
Would be great if you can take another look! ;)

Best, Jincheng

sunjincheng121:
Thanks for the review @JingsongLi @wuchong @aljoscha 
@flinkbot approve all
Merging...

JingsongLi:
LGTM

PR #9399, opened 2019-08-09
[FLINK-13526][sql-client] Switching to a non existing catalog or data…

lirui-apache:
…base crashes sql-client

Avoid crashing sql-client when switching to non-existing catalog or database.

## Brief change log

  - Catch `CatalogException` thrown by TableEnvironment and throw a `SqlExecutionException` instead.
  - Add test cases.

## Verifying this change

New test cases.

  - The serializers: no

## Documentation

### DIFF
--- a/flink-table/flink-sql-client/src/test/java/org/apache/flink/table/client/gateway/local/LocalExecutorITCase.java
+++ b/flink-table/flink-sql-client/src/test/java/org/apache/flink/table/client/gateway/local/LocalExecutorITCase.java
@@ -506,6 +506,30 @@ public class LocalExecutorITCase extends TestLogger {
 		}
 	}
 
+	@Test
+	public void testUseNonExistingDatabase() throws Exception {
+		final Executor executor = createDefaultExecutor(clusterClient);
+		final SessionContext session = new SessionContext("test-session", new Environment());
+
+		try {
+			executor.useDatabase(session, "nonexistingdb");
+		} catch (SqlExecutionException e) {
+			// expected
+		}
+	}
+
+	@Test
+	public void testUseNonExistingCatalog() throws Exception {
+		final Executor executor = createDefaultExecutor(clusterClient);
+		final SessionContext session = new SessionContext("test-session", new Environment());
+
+		try {
+			executor.useCatalog(session, "nonexistingcatalog");
+		} catch (SqlExecutionException e) {
+			// expected
+		}
+	}
+
 	private void executeStreamQueryTable(
 			Map<String, String> replaceVars,
 			String query,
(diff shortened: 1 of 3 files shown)


lirui-apache:
@xuefuz @bowenli86 @zjuwangg please have a look. I suppose it's trivial change.

flink-table/flink-sql-client/src/test/java/org/apache/flink/table/client/gateway/local/LocalExecutorITCase.java
```diff
@@ -506,6 +506,30 @@ public void testUseCatalogAndUseDatabase() throws Exception {
...
 	}
 
+	@Test
+	public void testUseNonExistingDatabase() throws Exception {
+		final Executor executor = createDefaultExecutor(clusterClient);
+		final SessionContext session = new SessionContext("test-session", new Environment());
+
+		try {
+			executor.useDatabase(session, "nonexistingdb");
+		} catch (SqlExecutionException e) {
```
bowenli86: shall we just assert the exception is expected instead of try-catch-ignore?

flink-table/flink-sql-client/src/test/java/org/apache/flink/table/client/gateway/local/LocalExecutorITCase.java
```diff
@@ -506,6 +506,30 @@ public void testUseCatalogAndUseDatabase() throws Exception {
...
+	}
+
+	@Test
+	public void testUseNonExistingCatalog() throws Exception {
+		final Executor executor = createDefaultExecutor(clusterClient);
+		final SessionContext session = new SessionContext("test-session", new Environment());
+
+		try {
+			executor.useCatalog(session, "nonexistingcatalog");
+		} catch (SqlExecutionException e) {
```
bowenli86: shall we just assert the exception is expected instead of try-catch-ignore?

flink-table/flink-sql-client/src/test/java/org/apache/flink/table/client/gateway/local/LocalExecutorITCase.java
lirui-apache: It's following the pattern in `testValidateSession`. But I guess it's also fine to use `ExpectedException`.

lirui-apache:
@bowenli86 PR updated according to your suggestions. Please take another look. Thanks.

lirui-apache:
@bowenli86 The test in `LocalExecutorITCase` is to verify `LocalExecutor` throws a proper exception (i.e. `SqlExecutionException`) when db/catalog doesn't exist. Do you think we need a test for sql-client to make sure it doesn't crash?

bowenli86:
is there a contract that throwing SqlExecutionException guarantees executor won't crash? if so, we probably don't need extra test anymore. BTW, the example is named as `testUseNonExistingDatabaseDoesntCrush()`

lirui-apache:
@bowenli86 Throwing `SqlExecutionException` is part of the contract of `Executor`, so in general I expect SQL client to be able to handle such exceptions. However, since each "call command" is implemented independently (regarding exception handling), I added a test anyway in `CliClientTest` to make sure the exception is indeed handled.

PR #9404, opened 2019-08-09
[FLINK-13667][ml] Add the utility class for the Table

xuyang1706:

Add the utility class for the Table 

## Brief change log

  - *Add the operations on column name*
  - *Add the operations on column type*
  - *Add unit test case*
 

## Verifying this change

- run test case pass

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/TableUtil.java
+++ b/flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/TableUtil.java
@@ -38,7 +38,9 @@ import java.util.UUID;
  */
 public class TableUtil {
 	/**
-	 * Return a lower temp table name in lower case and use "-" as connector character.
+	 * Return a temp table named with prefix `temp_`, follow by a random UUID.
+	 *
+	 * <p>UUID hyphens ("-") will be replaced by underscores ("_").
 	 *
 	 * @return tableName
 	 */
@@ -66,6 +68,17 @@ public class TableUtil {
 	}
 
 	/**
+	 * Find the index of <code>targetCol</code> from the <code>tableSchema</code>.
+	 *
+	 * @param tableSchema the TableSchema among which to find the targetCol.
+	 * @param targetCol   the targetCols to find.
+	 * @return the index of the targetCol.
+	 */
+	public static int findColIndex(TableSchema tableSchema, String targetCol) {
+		return findColIndex(tableSchema.getFieldNames(), targetCol);
+	}
+
+	/**
 	 * Find the indices of <code>targetCols</code> in string array <code>tableCols</code>. If
 	 * <code>targetCols</code> is
 	 * null, it will be replaced by the <code>tableCols</code>
@@ -87,17 +100,6 @@ public class TableUtil {
 			indices[i] = findColIndex(tableCols, targetCols[i]);
 		}
 		return indices;
-	}
-
-	/**
-	 * Find the index of <code>targetCol</code> from the <code>tableSchema</code>.
-	 *
-	 * @param tableSchema the TableSchema among which to find the targetCol.
-	 * @param targetCol   the targetCols to find.
-	 * @return the index of the targetCol.
-	 */
-	public static int findColIndex(TableSchema tableSchema, String targetCol) {
-		return findColIndex(tableSchema.getFieldNames(), targetCol);
 	}
 
 	/**
@@ -148,7 +150,7 @@ public class TableUtil {
 	 * @param dataType the dataType to determine.
 	 * @return whether it is number type
 	 */
-	public static boolean isNumber(TypeInformation dataType) {
+	public static boolean isSupportedNumericType(TypeInformation dataType) {
 		return Types.DOUBLE == dataType
 			|| Types.LONG == dataType
 			|| Types.BYTE == dataType
@@ -208,7 +210,7 @@ public class TableUtil {
 		if (selectedCols != null && selectedCols.length != 0) {
 			for (String selectedCol : selectedCols) {
 				if (null != selectedCol) {
-					if (!isNumber(findColType(tableSchema, selectedCol))) {
+					if (!isSupportedNumericType(findColType(tableSchema, selectedCol))) {
 						throw new IllegalArgumentException("col type must be number " + selectedCol);
 					}
 				}
@@ -315,7 +317,7 @@ public class TableUtil {
 		TypeInformation<?>[] inColTypes = tableSchema.getFieldTypes();
 
 		for (int i = 0; i < inColNames.length; i++) {
-			if (isNumber(inColTypes[i])) {
+			if (isSupportedNumericType(inColTypes[i])) {
 				if (null == excludeColsList || !excludeColsList.contains(inColNames[i])) {
 					numericCols.add(inColNames[i]);
 				}
@@ -420,8 +422,9 @@ public class TableUtil {
 	}
 
 	/**
-	 * which method Convert column names array to SQL clause. For example, columns {a, b, c} will be converted
-	 * into "`a`,`b`,`c`".
+	 * Convert column name array to SQL clause.
+	 *
+	 * <p>For example, columns "{a, b, c}" will be converted into a SQL-compatible select string section: "`a`, `b`, `c`".
 	 *
 	 * @param colNames columns to convert
 	 * @return converted SQL clause.
--- a/flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/TableUtil.java
+++ b/flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/TableUtil.java
@@ -24,17 +24,18 @@ import org.apache.flink.api.common.typeinfo.Types;
 import org.apache.flink.table.api.TableSchema;
 import org.apache.flink.util.Preconditions;
 
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
 import java.util.UUID;
 
 /**
- * Utility class for the operations on Table, such as
- * the operations on column name, the operations on column
- * type and so on.
+ * Utility class for the operations on Table, such as the operations on column name, the operations on column type and
+ * so on.
  */
 public class TableUtil {
 	/**
-	 * return a lower temp table name in lower case and
-	 * use "-" as connector character.
+	 * Return a lower temp table name in lower case and use "-" as connector character.
 	 *
 	 * @return tableName
 	 */
@@ -44,15 +45,12 @@ public class TableUtil {
 	}
 
 	/**
-	 * Find the index of <code>targetCol</code> in string
-	 * array <code>tableCols</code>. It will ignore the
-	 * case of the tableCols.
+	 * Find the index of <code>targetCol</code> in string array <code>tableCols</code>. It will ignore the case of the
+	 * tableCols.
 	 *
-	 * @param tableCols a string array among which to find
-	 *                  the targetCol.
+	 * @param tableCols a string array among which to find the targetCol.
 	 * @param targetCol the targetCol to find.
-	 * @return the index of the targetCol, if not found,
-	 * returns -1.
+	 * @return the index of the targetCol, if not found, returns -1.
 	 */
 	public static int findColIndex(String[] tableCols, String targetCol) {
 		Preconditions.checkNotNull(targetCol, "targetCol is null!");
@@ -65,12 +63,11 @@ public class TableUtil {
 	}
 
 	/**
-	 * Find the indices of <code>targetCols</code> in string
-	 * array <code>tableCols</code>. If <code>targetCols</code>
-	 * is null, it will be replaced by the <code>tableCols</code>
+	 * Find the indices of <code>targetCols</code> in string array <code>tableCols</code>. If
+	 * <code>targetCols</code> is
+	 * null, it will be replaced by the <code>tableCols</code>
 	 *
-	 * @param tableCols  a string array among which to
-	 *                   find the targetCols.
+	 * @param tableCols  a string array among which to find the targetCols.
 	 * @param targetCols the targetCols to find.
 	 * @return the indices of the targetCols.
 	 */
@@ -90,11 +87,9 @@ public class TableUtil {
 	}
 
 	/**
-	 * Find the index of <code>targetCol</code> from
-	 * the <code>tableSchema</code>.
+	 * Find the index of <code>targetCol</code> from the <code>tableSchema</code>.
 	 *
-	 * @param tableSchema the TableSchema among which
-	 *                    to find the targetCol.
+	 * @param tableSchema the TableSchema among which to find the targetCol.
 	 * @param targetCol   the targetCols to find.
 	 * @return the index of the targetCol.
 	 */
@@ -103,11 +98,9 @@ public class TableUtil {
 	}
 
 	/**
-	 * Find the indices of <code>targetCols</code> from
-	 * the <code>tableSchema</code>.
+	 * Find the indices of <code>targetCols</code> from the <code>tableSchema</code>.
 	 *
-	 * @param tableSchema the TableSchema among which to
-	 *                    find the targetCols.
+	 * @param tableSchema the TableSchema among which to find the targetCols.
 	 * @param targetCols  the targetCols to find.
 	 * @return the indices of the targetCols.
 	 */
@@ -116,8 +109,7 @@ public class TableUtil {
 	}
 
 	/**
-	 * Find the types of the <code>targetCols</code>.
-	 * If the targetCol not exist, return null.
+	 * Find the types of the <code>targetCols</code>. If the targetCol not exist, return null.
 	 *
 	 * @param tableSchema TableSchema.
 	 * @param targetCols  the targetCols to find.
@@ -132,8 +124,7 @@ public class TableUtil {
 	}
 
 	/**
-	 * Find the type of the <code>targetCol</code>.
-	 * If the targetCol not exist, return null.
+	 * Find the type of the <code>targetCol</code>. If the targetCol not exist, return null.
 	 *
 	 * @param tableSchema TableSchema
 	 * @param targetCol   the targetCol to find.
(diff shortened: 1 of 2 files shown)


flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/TableUtil.java
```diff
@@ -0,0 +1,429 @@
...
+	/**
+	 * Find the types of the <code>targetCols</code>. If the targetCol not exist, return null.
+	 *
+	 * @param tableSchema TableSchema.
+	 * @param targetCols  the targetCols to find.
+	 * @return the corresponding types.
+	 */
+	public static TypeInformation[] findColTypes(TableSchema tableSchema, String[] targetCols) {
+		TypeInformation[] types = new TypeInformation[targetCols.length];
+		for (int i = 0; i < types.length; i++) {
```
walterddr: echo previous comment - check null for `targetCols` missing

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/TableUtil.java
```diff
@@ -0,0 +1,429 @@
...
+	 * Find the indices of <code>targetCols</code> in string array <code>tableCols</code>. If
+	 * <code>targetCols</code> is
+	 * null, it will be replaced by the <code>tableCols</code>
+	 *
+	 * @param tableCols  a string array among which to find the targetCols.
+	 * @param targetCols the targetCols to find.
+	 * @return the indices of the targetCols.
+	 */
+	public static int[] findColIndices(String[] tableCols, String[] targetCols) {
+		if (targetCols == null) {
```
walterddr: related to `findColTypes` --> this is not consistent. 

Also the chaining call to `findColIndex(..)` may result in some "-1" values in the return array. If this is expected, please put a test case to demonstrate.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/TableUtil.java
```diff
@@ -0,0 +1,429 @@
...
+		}
+	}
+
+	/**
+	 * Check whether colTypes of the <code>selectedCols</code> is numerical, if not, throw exception.
+	 *
+	 * @param tableSchema  TableSchema
+	 * @param selectedCols the selectedCols to assert.
+	 */
+	public static void assertNumericalCols(TableSchema tableSchema, String... selectedCols) {
```
walterddr: missing test for `null` type information

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/TableUtil.java
```diff
@@ -0,0 +1,429 @@
...
+	/**
+	 * Return the columns in the table whose types are string and are not included in the excludeCols.
+	 *
+	 * <p>If <code>excludeCols</code> is null, return all the string columns.
+	 *
+	 * @param tableSchema TableSchema.
+	 * @param excludeCols The columns who are not considered.
+	 * @return string columns.
+	 */
+	public static String[] getStringCols(TableSchema tableSchema, String[] excludeCols) {
```
walterddr: question on this and all below `getXXXCols` API. why is the fields being get needs to have an "excludeCols" instead of include?

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/TableUtil.java
```diff
@@ -0,0 +1,429 @@
...
+				}
+			}
+		}
+
+		return numericCols.toArray(new String[0]);
+	}
+
+	/**
+	 * Get the columns from featureCols who are included in the <code>categoricalCols</code>, and the columns whose
+	 * types are string or boolean.
```
walterddr: could you elaborate what scenarios would require this API?

flink-ml-parent/flink-ml-lib/src/test/java/org/apache/flink/ml/common/utils/TableUtilTest.java
```diff
@@ -0,0 +1,127 @@
...
+		Assert.assertTrue(TableUtil.isNumber(Types.LONG));
+		Assert.assertTrue(TableUtil.isNumber(Types.BYTE));
+		Assert.assertTrue(TableUtil.isNumber(Types.FLOAT));
+		Assert.assertTrue(TableUtil.isNumber(Types.SHORT));
+		Assert.assertFalse(TableUtil.isNumber(Types.STRING));
+		Assert.assertTrue(TableUtil.isString(Types.STRING));
+	}
+
+	@Test
+	public void assertTest() {
```
walterddr: better test naming is needed.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/TableUtil.java
xuyang1706: We keep it consistent with findColIndices.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/TableUtil.java
xuyang1706: "-1" values are expected and we have added the test case.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/TableUtil.java
xuyang1706: We have added the test case.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/TableUtil.java
xuyang1706: Take the "getNumericCols" API as an example: when there are a number of feature columns and we want to use all the numerical columns as feature columns except the label column, this API is used.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/TableUtil.java
xuyang1706: An example of use this API is https://issues.apache.org/jira/browse/FLINK-13835.

flink-ml-parent/flink-ml-lib/src/test/java/org/apache/flink/ml/common/utils/TableUtilTest.java
xuyang1706: We have modified the test name.

xuyang1706:
> Thanks for the contribution @xuyang1706 . I think overall the diff looks good to me. Just have some questions regarding the usage scenario of the utility class. please kindly take a look --Rong

Thanks for your comments and advice @walterddr . We have refined the code and added more test cases. Thanks -Xu

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/TableUtil.java
```diff
@@ -0,0 +1,432 @@
...
+	public static int[] findColIndices(String[] tableCols, String[] targetCols) {
+		if (targetCols == null) {
+			int[] indices = new int[tableCols.length];
+			for (int i = 0; i < tableCols.length; i++) {
+				indices[i] = i;
+			}
+			return indices;
+		}
+		int[] indices = new int[targetCols.length];
+		for (int i = 0; i < indices.length; i++) {
```
walterddr: one more question is: would this function be used (1) only once (e.g. during table program compile time) or (2) per-row (e.g. as part of the job graph and execute during runtime) ? 

if (2), there's ways to avoid the double loop for performance improvement but if (1) is true there's no need IMO.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/TableUtil.java
```diff
@@ -0,0 +1,432 @@
...
+	}
+
+	/**
+	 * Find the index of <code>targetCol</code> from the <code>tableSchema</code>.
+	 *
+	 * @param tableSchema the TableSchema among which to find the targetCol.
+	 * @param targetCol   the targetCols to find.
+	 * @return the index of the targetCol.
+	 */
+	public static int findColIndex(TableSchema tableSchema, String targetCol) {
```
walterddr: nit. please move all overload methods together. see [Google code style 3.4.2.1 - overloads](https://google.github.io/styleguide/javaguide.html#s3.4.2-ordering-class-contents)

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/TableUtil.java
```diff
@@ -0,0 +1,432 @@
...
+	}
+
+	/**
+	 * Determine whether it is number type, number type includes double, long, byte, int, float and short.
+	 *
+	 * @param dataType the dataType to determine.
+	 * @return whether it is number type
+	 */
+	public static boolean isNumber(TypeInformation dataType) {
+		return Types.DOUBLE == dataType
```
walterddr: what about: `Types.BIG_DEC` & `Types.BIG_INT`? they seems to be SQL only but should it be considered?

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/TableUtil.java
```diff
@@ -0,0 +1,432 @@
...
+
+		for (Row row : data) {
+			sbd.append("\n").append(formatRows(row));
+		}
+
+		return sbd.toString();
+	}
+
+	/**
+	 * which method Convert column names array to SQL clause. For example, columns {a, b, c} will be converted
```
walterddr:
```suggestion
	 * Convert column name array to SQL clause. 
	 * 
	 * <p>For example, columns "{a, b, c}" will be converted into a SQL-compatible select string section: "`a`, `b`, `c`".
```

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/TableUtil.java
```diff
@@ -0,0 +1,432 @@
...
+import java.util.List;
+import java.util.UUID;
+
+/**
+ * Utility class for the operations on Table, such as the operations on column name, the operations on column type and
+ * so on.
+ */
+public class TableUtil {
+	/**
+	 * Return a lower temp table name in lower case and use "-" as connector character.
```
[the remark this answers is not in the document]
xuyang1706: Thanks, changed

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/TableUtil.java
xuyang1706: Thanks, this function should be called at the flink client. so it belongs to (1)

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/TableUtil.java
xuyang1706: Thanks, changed

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/TableUtil.java
xuyang1706: Thanks, Types.BIG_DEC and Types.BIG_INT are not necessary for ML algorithms, so we change the function name to isSupportedNumericType

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/TableUtil.java
xuyang1706: Thanks, changed

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/TableUtil.java
xuyang1706: Thanks, this function should be called at the flink client. so it belongs to (1)

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/TableUtil.java
```diff
@@ -0,0 +1,435 @@
...
+
+import org.apache.flink.shaded.guava18.com.google.common.base.Joiner;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.UUID;
+
+/**
+ * Utility class for the operations on Table, such as the operations on column name, the operations on column type and
```
walterddr:
```suggestion
 * Utility to operator to interact with Table contents, such as rows and columns.
```

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/TableUtil.java
walterddr: thanks for the explanation. yes it looks good to me

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/TableUtil.java
walterddr: can you add a simple example in the javadoc?

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/TableUtil.java
xuyang1706: Thanks, changed

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/TableUtil.java
xuyang1706: Thanks, added

xuyang1706:
> Thanks for the contribution @xuyang1706 . I think overall the diff looks good to me. Just have some questions regarding the usage scenario of the utility class. please kindly take a look --Rong

Thanks for your comments @walterddr .  I have changed the JavaDoc for detail usage scenario. Thanks, -Xu

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/TableUtil.java
```diff
@@ -0,0 +1,439 @@
...
+ */
+public class TableUtil {
+	/**
+	 * Return a temp table named with prefix `temp_`, follow by a random UUID.
+	 *
+	 * <p>UUID hyphens ("-") will be replaced by underscores ("_").
+	 *
+	 * @return tableName
+	 */
+	public static synchronized String getTempTableName() {
```
walterddr: missing test for this API.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/TableUtil.java
```diff
@@ -0,0 +1,439 @@
...
+	}
+
+	/**
+	 * Check whether colTypes of the <code>selectedCols</code> is vector, if not, throw exception.
+	 *
+	 * @param tableSchema  TableSchema
+	 * @param selectedCols the selectedCol to assert.
+	 * @see #isVector(TypeInformation)
+	 */
+	public static void assertVectorCols(TableSchema tableSchema, String... selectedCols) {
```
walterddr: missing test for this API

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/TableUtil.java
```diff
@@ -0,0 +1,439 @@
...
+		}
+	}
+
+	/**
+	 * Return the columns in the table whose types are string.
+	 *
+	 * @param tableSchema TableSchema
+	 * @return String columns.
+	 */
+	public static String[] getStringCols(TableSchema tableSchema) {
```
walterddr: missing test for this API.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/TableUtil.java
```diff
@@ -0,0 +1,439 @@
...
+
+	/**
+	 * Convert column name array to SQL clause.
+	 *
+	 * <p>For example, columns "{a, b, c}" will be converted into a SQL-compatible select string section: "`a`, `b`, `c`".
+	 *
+	 * @param colNames columns to convert
+	 * @return converted SQL clause.
+	 */
+	public static String columnsToSqlClause(String[] colNames) {
```
walterddr: missing test for this API

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/utils/TableUtil.java
```diff
@@ -0,0 +1,439 @@
...
+			}
+		}
+
+		return sbd.toString();
+	}
+
+	/**
+	 * format the column names and rows in table as markdown.
+	 */
+	public static String format(String[] colNames, List<Row> data) {
```
walterddr: missing test for this API

xuyang1706:
> Sorry @xuyang1706 for the late follow up. I committed a change to fix the test issues in: [walterddr@c8e13a7](https://github.com/walterddr/flink/commit/c8e13a72b08f048c76052ad2e1fa3767193c09e7). Please kindly take a look.
> 
> If everything looks good I will merge it soon.

Thanks very much for your kindly help on adding more test cases! Yes, your change looks very good for me. Thanks, -Xu

walterddr:
thanks for the feedback. will merge it soon.

PR #9417, opened 2019-08-12
[FLINK-13688][hive] Limit the parallelism/memory of HiveCatalogUseBlinkITCase

JingsongLi:

limit the parallelism of HiveCatalogUseBlinkITCase to avoid too many slot requests by default parallelism (use the core size of machine).

## Verifying this change

  - The serializers: no

## Documentation

### DIFF
--- a/flink-connectors/flink-connector-hive/pom.xml
+++ b/flink-connectors/flink-connector-hive/pom.xml
@@ -616,7 +616,14 @@ under the License.
 			<scope>test</scope>
 		</dependency>
 
-	</dependencies>
+        <dependency>
+            <groupId>org.apache.flink</groupId>
+            <artifactId>flink-test-utils_2.11</artifactId>
+            <version>${project.version}</version>
+			<scope>test</scope>
+        </dependency>
+
+    </dependencies>
 
 	<build>
 		<plugins>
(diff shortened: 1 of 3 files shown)


KurtYoung:
Another way is to extend `AbstractTestBase`, which will configure mini cluster in a more predictable way.

KurtYoung:
And I'm not sure the root cause is the parallelism is too high for this case. We should support running very high parallelism batch job with only one slot by design.

JingsongLi:
@KurtYoung Yeah, the main problem is memory request is too high(128MB).  I think 8dacb85c1e1e1bbd7a31c91800b886eb62be2c80 will solve it.

KurtYoung:
How about we use a pre configured mini cluster instead of lower the memory request? Because it's not 100% safe to just rely on lowering the sort memory, because we might choose hash aggregate.

JingsongLi:
> How about we use a pre configured mini cluster instead of lower the memory request? Because it's not 100% safe to just rely on lowering the sort memory, because we might choose hash aggregate.

Yeah, that is a way too, but in this case, I think use lower memory is OK. Because the data is too few.
1.It must rely on sort now, because UDAF only can be support by sort agg.
2.I think I can extract the configuration in `BatchTestBase`, to config all memory things.

KurtYoung:
I would like to repeat my first question: why not just extending ` AbstractTestBase`

JingsongLi:
> I would like to repeat my first question: why not just extending ` AbstractTestBase`

OK.

JingsongLi:
This case makes me feel that the first time a user uses blink batch sql, he has to think about resources.....

bowenli86:
+1 for fixing it. @KurtYoung @JingsongLi

KurtYoung:
Verified locally, +1.

PR #9419, opened 2019-08-12
[FLINK-13599][e2e tests] Harden test_streaming_kinesis with kinesalite docker image download/run retries

azagrebin:

Add retry logic for docker run command to start kinesalite in Kinesis E2E test.

## Brief change log

  - add function `retry_until_passed_and_wait_condition` to `common.sh`
  - call `retry_until_passed_and_wait_condition` in `test_streaming_kinesis.sh`

## Verifying this change

FLINK_DIR=../build-target ./run-single-test.sh test-scripts/test_streaming_kinesis.sh

  - The serializers: (no)

## Documentation


flink-end-to-end-tests/test-scripts/common.sh
```diff
@@ -747,3 +747,47 @@ function retry_times() {
...
+###################################################################################
+function retry_until_passed_and_wait_condition {
+    local action=${@:1}
+    local check_done=${@:2}
+    local name=${3}
+    local max_retry="${4:-10}"
+    local backoff_delay="${5:-0}"
+
+    local attempt=1
+    local action_passed=1
```
GJL: The name implies a boolean, which it is not. How about `action_exit_code`?

flink-end-to-end-tests/test-scripts/common.sh
```diff
@@ -747,3 +747,47 @@ function retry_times() {
...
+# Globals:
+#   -
+# Arguments:
+#   $1 - action to run
+#   $2 - boolean function to check whether the action has been successfully done
+#   $3 - action name for logs
+#   $4 - (default: 10) max number of retries to run action until passed and check done
+#   $5 - (default: 0) backoff delay in seconds between retry attempts
+# Returns:
+#   Done or failed message
```
GJL: Not sure if it makes sense to document the return value. Technically this function also _"returns"_ _"Retry <...> after ... seconds, ..."_ so the documentation is already inaccurate.

flink-end-to-end-tests/test-scripts/test_streaming_kinesis.sh
```diff
@@ -28,9 +28,26 @@ export AWS_SECRET_KEY=flinkKinesisTestFakeAccessKey
...
+function KINESALITE_STARTED {
+    if [[ "$(docker ps --format '{{.Names}}' | grep flink-test-kinesis)" = "flink-test-kinesis" ]]; then
+        return 0
+    else
+        return 1
+    fi
+}
+
+START_KINESALITE_MAX_RETRY_NUMBER=10
+retry_until_passed_and_wait_condition START_KINESALITE KINESALITE_STARTED "start kinesalite" ${START_KINESALITE_MAX_RETRY_NUMBER}
```
GJL: Were you able to reproduce the problem, and does _"retrying"_ mitigate it?

flink-end-to-end-tests/test-scripts/common.sh
```diff
@@ -747,3 +747,47 @@ function retry_times() {
...
+###################################################################################
+function retry_until_passed_and_wait_condition {
```
1u0: I think, it should be possible to "simulate" your new function by already existing `retry_times` function above:
```bash
retry_until_passed_and_wait_condition START_KINESALITE KINESALITE_STARTED "start kinesalite" ${START_KINESALITE_MAX_RETRY_NUMBER}

### Can be translated to smt. like:

function check_and_start_kinesalite() {
  if ! KINESALITE_STARTED; then
    START_KINESALITE
  fi
  KINESALITE_STARTED
  return $?
}
if ! retry_times ${START_KINESALITE_MAX_RETRY_NUMBER} 0 check_and_start_kinesalite ; then
  exit 1
fi
```

I suggest to avoid functional duplication and try to consolidate both methods.

flink-end-to-end-tests/test-scripts/test_streaming_kinesis.sh
```diff
@@ -28,9 +28,26 @@ export AWS_SECRET_KEY=flinkKinesisTestFakeAccessKey
 
 KINESALITE_PORT=4567
 
-#docker run -d --rm --name flink-test-kinesis -p ${KINESALITE_PORT}:${KINESALITE_PORT} instructure/kinesalite
-# override entrypoint to enable SSL
-docker run -d --rm --entrypoint "/tini" --name flink-test-kinesis -p ${KINESALITE_PORT}:${KINESALITE_PORT} instructure/kinesalite -- /usr/src/app/node_modules/kinesalite/cli.js --path /var/lib/kinesalite --ssl
+function START_KINESALITE {
```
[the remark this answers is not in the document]
azagrebin: true, leftover

flink-end-to-end-tests/test-scripts/test_streaming_kinesis.sh
azagrebin: I have not looped it on travis, only tested it by turning off and on connection and seeing that it downloads docker image eventually. I can try looping it on Travis but not sure we can do much more against the download hiccups.

flink-end-to-end-tests/test-scripts/common.sh
azagrebin: We cannot retry docker run if it has succeeded, only checking that the container started,
but true, it is possible to reuse `retry_times` in 2 steps. I pushed the change.

azagrebin:
Thanks for the review @GJL @1u0 
I also tried to loop on Travis but it is hard to reproduce that download hiccup.
The testing of the retry itself should be enough.
We have to see if the original download problem sustains despite the retries.
Merging this.

PR #9421, opened 2019-08-12
[FLINK-13539][table-api] Improve CSV table factory doesn't need to require format.fields

wuchong:

Currently, the CSV table factory requires `format.fields.`, which makes it hard to use. This PR makes CSV table factory supports `deriveSchema` property to derive schema from DDL body part and avoid to pass `format.fields.` properties.

## Brief change log

- Add a `deriveSchema` property to derive schema from `schema.*` properties.

## Verifying this change

- Add a `CsvTableSinkFactoryTest` to verify source and sink table factory. 

## Documentation

### DIFF
--- a/flink-table/flink-table-api-java-bridge/src/main/java/org/apache/flink/table/descriptors/OldCsv.java
+++ b/flink-table/flink-table-api-java-bridge/src/main/java/org/apache/flink/table/descriptors/OldCsv.java
@@ -31,6 +31,7 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.stream.Collectors;
 
+import static org.apache.flink.table.descriptors.FormatDescriptorValidator.FORMAT_DERIVE_SCHEMA;
 import static org.apache.flink.table.descriptors.OldCsvValidator.FORMAT_COMMENT_PREFIX;
 import static org.apache.flink.table.descriptors.OldCsvValidator.FORMAT_FIELDS;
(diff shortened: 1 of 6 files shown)


wuchong:
cc @JingsongLi  @twalthr , I would appreciate if you could have a look.

flink-table/flink-table-api-java-bridge/src/main/java/org/apache/flink/table/descriptors/OldCsv.java
```diff
@@ -168,22 +170,40 @@ public OldCsv ignoreParseErrors() {
 		return this;
 	}
 
+	/**
+	 * Derives the format schema from the table's schema. Required if no format schema is defined.
+	 *
+	 * <p>This allows for defining schema information only once.
+	 *
+	 * <p>The names, types, and fields' order of the format are determined by the table's
+	 * schema. Time attributes are ignored if their origin is not a field. A "from" definition
```
twalthr: remove the comments about `from` and the ignoring of time attributes.

wuchong:
Hi @twalthr , do you think we should merge this back to 1.9 branch? This is not a bug but an improvement/feature for user experience.

twalthr:
Usually, we don't introduce new features to minor releases. You never know if it also introduces some bugs. I would not do that.

wuchong:
Thanks for the replying @twalthr , so I won't cherry-pick it as I didn't see users complaining this in ML.

PR #9422, opened 2019-08-12
 [FLINK-13697][table-api] Removed deprecated ExternalCatalog API 

dawidwys:

Drop the deprecated ExternalCatalog API and simplify the calls that can be simplified due to that.

## Brief change log

  - Dropped all ExternalCatalog API related classes
  - Unified CatalogManager operations
  - Used ObjectIdentifier in CatalogTableOperation

## Verifying this change

## Documentation

  - If yes, how is the feature documented? (not applicable / **docs** / **JavaDocs** / not documented)


bowenli86:
as you may guess, I'm so in favor of this PR!

just to double check - should classes be deleted after one release or two after they're marked as deprecated? I roughly remember there're some community rules around that but can't recall what exactly it is. e.g. it's marked deprecated in 1.9, can it be deleted in 1.10 or 1.11?

dawidwys:
Hi @bowenli86 
I don't think we have a rule about stability of `Evolving` interfaces. I think the rule you are referring to is about state migration gurantees and bugfix support for the last two versions.

twalthr:
@bowenli86 Dawid is right, we don't have a strict process but use common sense. For example, the old catalog interfaces were more "2 Level API" and had not many users, so we remove them after one release. Deprecated core API such as `cast(TypeInformation)` in Scala expression DSL was used by almost every program and should stay around for some time.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/internal/TableEnvironmentImpl.java
```diff
@@ -180,7 +177,8 @@ public void registerTable(String name, Table table) {
 		}
 
 		CatalogBaseTable tableTable = new QueryOperationCatalogView(table.getQueryOperation());
-		registerTableInternal(name, tableTable);
+		CatalogManagerCommands.createTable(tableTable, false)
+			.executeIn(catalogManager, getTemporaryTablePath(name));
```
twalthr: Why can't we integrate `CatalogManagerCommands` into `CatalogManager`? The first argument of `executeIn` is always the same and basically implements a member that should be part of the manager itself.

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/plan/QueryOperationConverter.java
```diff
@@ -296,7 +297,12 @@ public RelNode visit(SortQueryOperation sort) {
 
 		@Override
 		public RelNode visit(CatalogQueryOperation catalogTable) {
-			return relBuilder.scan(catalogTable.getTablePath()).build();
+			ObjectIdentifier tablePath = catalogTable.getTablePath();
```
twalthr: Rename `tablePath` to `objectIdentifier` in `CatalogQueryOperation`.

flink-table/flink-table-planner/src/main/scala/org/apache/flink/table/sinks/TableSinkUtils.scala
```diff
@@ -41,7 +42,7 @@ object TableSinkUtils {
   def validateSink(
       staticPartitions: JMap[String, String],
       query: QueryOperation,
-      sinkPath: JList[String],
+      sinkPath: ObjectIdentifier,
```
twalthr: nit: update all names to identifier

flink-table/flink-table-planner/src/test/java/org/apache/flink/table/runtime/batch/table/JavaTableEnvironmentITCase.java
```diff
@@ -138,7 +137,7 @@ public void testRegisterWithFields() throws Exception {
 		compareResultAsText(results, expected);
 	}
 
-	@Test(expected = TableAlreadyExistException.class)
+	@Test(expected = TableException.class)
```
twalthr: Shall we use `ValidationException` instead?

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/api/internal/TableEnvironmentImpl.java
dawidwys: Good idea

PR #9423, opened 2019-08-13
[FLINK-13699][table-api] Fix TableFactory doesn't work with DDL when containing TIMESTAMP/DATE/TIME types

wuchong:
The solution is encode DataTypes.TIMESTAMP() as "TIMESTAMP" when translating to properties.
And will be converted back to the old TypeInformation: Types.SQL_TIMESTAMP.
This would fix all factories at once.

Currently, in blink planner, we will convert DDL to TableSchema with new type system, i.e. DataTypes.TIMESTAMP()/DATE()/TIME() whose underlying TypeInformation are Types.LOCAL_DATETIME/LOCAL_DATE/LOCAL_TIME.

However, this makes the existing connector implementations (Kafka, ES, CSV, etc..) don't work because they only accept the old TypeInformations (Types.SQL_TIMESTAMP/SQL_DATE/SQL_TIME).

A simple solution is encode DataTypes.TIMESTAMP() as "TIMESTAMP" when translating to properties. And will be converted back to the old TypeInformation: Types.SQL_TIMESTAMP. This would fix all factories at once.

## Brief change log

- Convert Types.LOCAL_DATETIME/LOCAL_DATE/LOCAL_TIME to "TIMESTAMP"/"DATE"/"TIME" as well.

## Verifying this change

- Add unit test in `TypeStringUtilsTest` to verify the conversion from LocalDataTime/LocalDate/LocalTime.

## Documentation

### DIFF
--- a/flink-connectors/flink-hbase/src/test/java/org/apache/flink/addons/hbase/HBaseConnectorITCase.java
+++ b/flink-connectors/flink-hbase/src/test/java/org/apache/flink/addons/hbase/HBaseConnectorITCase.java
@@ -562,7 +562,7 @@ public class HBaseConnectorITCase extends HBaseTestBase {
 			CollectTableSink<Row> configuredSink = (CollectTableSink<Row>) sink.configure(
 				schema.getFieldNames(), types.toArray(new TypeInformation[0]));
 			return JavaScalaConversionUtil.toJava(
-				BatchTableEnvUtil.collect(t.getTableEnvironment(), table, configuredSink, Option.apply("JOB")));
+				BatchTableEnvUtil.collect(t.getTableEnvironment(), table, configuredSink, Option.apply("JOB"), "catalog", "database"));
 		}
 	}
 
--- a/flink-connectors/flink-hbase/src/test/java/org/apache/flink/addons/hbase/HBaseConnectorITCase.java
+++ b/flink-connectors/flink-hbase/src/test/java/org/apache/flink/addons/hbase/HBaseConnectorITCase.java
@@ -38,7 +38,9 @@ import org.apache.flink.table.api.java.StreamTableEnvironment;
 import org.apache.flink.table.descriptors.DescriptorProperties;
 import org.apache.flink.table.factories.TableFactoryService;
 import org.apache.flink.table.functions.ScalarFunction;
-import org.apache.flink.table.planner.runtime.utils.TableUtil;
+import org.apache.flink.table.planner.runtime.utils.BatchTableEnvUtil;
+import org.apache.flink.table.planner.sinks.CollectRowTableSink;
+import org.apache.flink.table.planner.sinks.CollectTableSink;
 import org.apache.flink.table.planner.utils.JavaScalaConversionUtil;
 import org.apache.flink.table.runtime.utils.StreamITCase;
 import org.apache.flink.table.sinks.TableSink;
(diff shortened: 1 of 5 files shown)


wuchong:
cc @JingsongLi  @twalthr , I would appreciate if you could have a look.

JingsongLi:
Such a change would cause `TypeStringUtils` to cease to be one that allows back-and-forth conversion.
Another choice is to only modify `DescriptorProperties.putTableSchema`. What do you think?
But consider `TypeStringUtils` is deprecated and it is only used in connector schema with logical presentation, I am OK too.

wuchong:
Hi @JingsongLi , IMO, the encoding of `TypeStringUtils` can not support compatible upgrade, no matter we add LocalDateTime or not.
However, we have the `property-version` which can be used to backward compatible. 

What's your suggestion about modifying `DescriptorProperties.putTableSchema`?

JingsongLi:
> What's your suggestion about modifying DescriptorProperties.putTableSchema?

Maybe should visit logical type and convert DATE/TIME/TIMESTAMP to `SqlTimeTypeInfo` instead of `LocalTimeTypeInfo`.

wuchong:
Sure @twalthr . Will update the PR soon.

wuchong:
Hi @twalthr , I added an HBase IT case to verify DDL with timestamp types. The case include DDL to create both source and sink. It passed in my local machine.

flink-connectors/flink-hbase/src/test/java/org/apache/flink/addons/hbase/HBaseConnectorITCase.java
```diff
@@ -320,6 +336,72 @@ public void testTableSink() throws Exception {
...
+
+		DataStream<Row> ds = execEnv.fromCollection(testData1).returns(testTypeInfo1);
+		tEnv.registerDataStream("src", ds);
+
+		// register hbase table
+		String quorum = getZookeeperQuorum();
+		String ddl = "CREATE TABLE hbase (\n" +
+			"    rowkey INT," +
+			"    family1 ROW<col1 INT>,\n" +
+			"    family2 ROW<col1 VARCHAR, col2 BIGINT>,\n" +
```
twalthr: Use `STRING` instead of `VARCHAR` and `TIMESTAMP(3)`/`TIME(3)` instead of `TIMESTAMP`/`TIME`.

wuchong: `STRING` is not allowed currently? #9354

twalthr: I thought this issue has been fixed for 1.9. This is bad :(

twalthr: This means that the DDL in 1.9 is instant legacy. Can we at least ensure the other comment around `TIMESTAMP(3)/TIME(3)`? Thanks.

wuchong: Sure. I will use `TIMESTAMP(3)/TIME(3)`.

twalthr:
@wuchong feel free to merge it once the built gives green light.

wuchong:
Travis passed in my own repo: https://travis-ci.org/wuchong/flink/builds/573858501

Merged with commits:

b837a589f1bda5d8352e9760af39937f9194c670 d20175ee62cd9b3ce8912745240b57c88c5af51c

PR #9426, opened 2019-08-13
[FLINK-12958] Integrate AsyncWaitOperator with mailbox (preserving compatibility with legacy sources)

1u0:

This PR integrates `AsyncWaitOperator` with task's mailbox, by coordinating `AsyncWaitOperator` with it's `Emitter` thread via mailbox (the `Emitter` now delegates the execution of the element processing to the main task thread).

## Brief change log

  - the `AsyncWaitOperator` may execute letters from the task's mailbox if the elements queue is full;
  - the `Emitter` thread is delegating elememnt processing and pop up from the `StreamElementQueue` to the main task thread (via mailbox).

## Verifying this change

## Documentation

### DIFF
--- a/flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/AsyncWaitOperator.java
+++ b/flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/AsyncWaitOperator.java
@@ -46,6 +46,7 @@ import org.apache.flink.streaming.runtime.streamrecord.StreamElementSerializer;
 import org.apache.flink.streaming.runtime.streamrecord.StreamRecord;
 import org.apache.flink.streaming.runtime.tasks.ProcessingTimeCallback;
 import org.apache.flink.streaming.runtime.tasks.StreamTask;
+import org.apache.flink.streaming.runtime.tasks.mailbox.execution.MailboxExecutor;
 import org.apache.flink.util.ExceptionUtils;
 import org.apache.flink.util.Preconditions;
 
@@ -113,6 +114,9 @@ public class AsyncWaitOperator<IN, OUT>
 	/** Thread running the emitter. */
 	private transient Thread emitterThread;
 
+	/** Executor frontend to the task mailbox. */
+	private transient MailboxExecutor mainThreadExecutor;
+
 	public AsyncWaitOperator(
 			AsyncFunction<IN, OUT> asyncFunction,
 			long timeout,
@@ -136,6 +140,7 @@ public class AsyncWaitOperator<IN, OUT>
 	public void setup(StreamTask<?, ?> containingTask, StreamConfig config, Output<StreamRecord<OUT>> output) {
 		super.setup(containingTask, config, output);
 
+		this.mainThreadExecutor = containingTask.getTaskMailboxExecutor();
 		this.checkpointingLock = getContainingTask().getCheckpointLock();
 
 		this.inStreamElementSerializer = new StreamElementSerializer<>(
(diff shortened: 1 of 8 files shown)


1u0:
cc @pnowojski, @zhijiangW, @AHeise
this is the rebased [branch](https://github.com/StefanRRichter/flink/tree/async-wait-rework) of @StefanRRichter work.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/AsyncWaitOperator.java
```diff
@@ -397,26 +404,41 @@ private void stopResources(boolean waitForShutdown) throws InterruptedException
...
 		pendingStreamElementQueueEntry = streamElementQueueEntry;
 
-		while (!queue.tryPut(streamElementQueueEntry)) {
-			// we wait for the emitter to notify us if the queue has space left again
-			checkpointingLock.wait();
+		if (!queue.tryPut(streamElementQueueEntry)) {
+			do {
+				// We will receive a notify on the lock at the end of each processing-letter execution (see {Emitter})
+				// so we can try to put again.
+				if (!tryYieldMainThreadExecutor()) {
```
pnowojski: This logic (difference of mailbox vs non mailbox thread) should be encapsulated in the mailbox. It's already duplicated here, it will be duplicated in other places (like `ContinousFileReaderOperator`) and mailbox should leak the details of the threading model to the operators.

something like `mailbox.yield()` called from here, should do this black magic in a secret.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/Emitter.java
```diff
@@ -138,19 +145,33 @@ private void output(AsyncResult asyncResult) throws InterruptedException {
...
-
-				// notify the main thread that there is again space left in the async collector
-				// buffer
-				checkpointLock.notifyAll();
+			} catch (InterruptedException e) {
+				Thread.currentThread().interrupt();
 			}
+		};
+
+		CompletableFuture<Void> writebackFuture = CompletableFuture.runAsync(processingRequest, mailboxExecutor);
```
pnowojski: Logic of synchronizing on `checkpointLock` and notifying it should also be hidden from the user.

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/StreamTaskTestHarness.java
```diff
@@ -111,8 +111,8 @@
 	protected StreamTestSingleInputGate[] inputGates;
 
 	public StreamTaskTestHarness(
-			Function<Environment, ? extends StreamTask<OUT, ?>> taskFactory,
-			TypeInformation<OUT> outputType) {
+		Function<Environment, ? extends StreamTask<OUT, ?>> taskFactory,
```
pnowojski: Could you revert the whitespace changes? They are questionable for me and even if we want them, they shouldn't be part of the same commit.

1u0:
@pnowojski, thanks for the review. I opened this PR mostly as quick alternative/unblocker for further work for checkpoints triggers and timers to be migrated to mailbox. Also, it should give some approach ideas for @AHeise.

Besides what you have mentioned, I think this PR needs additional work to be done (wrt to checkpoint lock).

I'm closing this PR, as I think the common agreement that @AHeise would continue working on this and I'd build checkpoints triggers migration upon his work.

PR #9434, opened 2019-08-14
[FLINK-13634][Formats]: added module to compress data using StreamingFileSink

eskabetxe:
This pull request is to create a CompressFileWriter which allows user to compressed the stream to filesystem using StreamingFileSink

## Brief change log
added a new module on flink-formats 

## Verifying this change
  - *Added test that validates that file is created compressed and can be read after creation*

  - The serializers: no

## Documentation

### DIFF
--- a/flink-formats/flink-compress/pom.xml
+++ b/flink-formats/flink-compress/pom.xml
@@ -1,0 +1,92 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+		xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+
+	<modelVersion>4.0.0</modelVersion>
+
+	<parent>
+		<groupId>org.apache.flink</groupId>
+		<artifactId>flink-formats</artifactId>
+		<version>1.10-SNAPSHOT</version>
+		<relativePath>..</relativePath>
+	</parent>
+
+	<artifactId>flink-compress</artifactId>
+	<name>flink-compress</name>
+
+	<packaging>jar</packaging>
+
+	<dependencies>
+
+		<dependency>
+			<groupId>org.apache.flink</groupId>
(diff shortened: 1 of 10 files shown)


eskabetxe:
@kl0u could you check this

eskabetxe:
@flinkbot attention @kl0u

eskabetxe:
hi @aljoscha,
any possibility of having this merged in the next release?

aljoscha:
Thanks for your contribution! I think this is a valuable addition.

I do have some comments about naming and structure:

I think the maven module should be called something like `org.apache.flink.formats.hadoopcompress` to show that this is for using Hadoop compression.

We should make the structure similar to `flink-parquet`, which is another format for the `StreamingFileSink` in order to give consistency to users. This means:
 - Add a `org.apache.flink.formats.hadoopcompress.HadoopCompressionWriters` similar to `ParquetAvroWriters` that has static methods for creating a writer using various forms of specifying a codec. You have these methods in `CompressFileWriterBuilder` right now, so they should be moved and that class should be removed.
 - `CompressFileWriter` should be called `CompressionWriterFactory`, similar to `ParquetWriterFactory`. This can be returned from the static methods above.
 - I don't see a need for the `Compressor` interface. `HadoopCompressor` should be called `CompressionBulkWriter`, because that's what it is and it would be created from the `CompressionWriterFactory`.
 - I don't see a need for a "no-compression" mode when the purpose of this addition is compression

If we iterate quickly we can try and manage to get this into the next release. :ok

eskabetxe:
Hi @aljoscha, 
I will do the changes..

Some points: 
HadoopCompressionWriters -> this was a builder because we have to set 2 fields, the extractor and the codec, making this static we multiplicate the number of combinations.

Compressor -> this was to allow other implementations (initial this is using apache codecs) and SequenceFiles, that are in flink-sequence-files and for that reason removed.

"no-compression" mode -> this has set to allow to change to a no-compression mode (for debug for example) without change the implementation of job

eskabetxe:
@aljoscha all changes done

aljoscha:
Thanks for the quick response!

I'm afraid I didn't express myself too clearly. We can still have the builder pattern, i.e. `CompressionWriterFactory` can have the builder methods for setting extractor and codec. The naming of the `BulkWriter.Factory` is unfortunately not very good and it could have been called `BulkWriter.Builder` instead. I thought that an `Extractor` was always needed so the static method on `HadoopCompressionWriters` could have been `forExtractor(Extractor)` and that returns the factory where you have the methods for setting the compression codec. If that's not the case you can also have a static method without parameters and do everything via the "builder".

Also, we shouldn't have both `hadoopcompress` and `hadoop`. We can either have only `hadoopcompress` or `compress` and then inside that `compress.hadoop`. Are you actually planning to add more compression codecs besides Hadoop? I think planning for those future additions would mean we have to potentially make the interfaces more independent of Hadoop.

Some Javadoc comments are outdated now, for example `CompressionWriterFactory` has
```
 * A {@link BulkWriter} implementation that compress file.
```

Sorry for the inconvenience for you!

eskabetxe:
I think we should be prepared to add other implementations without problem, for example the flink-sequence-files could be absorbed by this module

aljoscha:
Do you still want to add the `NoCompression` feature? I wasn't strictly against that, I meant it as a starting point for discussion and you don't have to agree to everything I suggest. 😅

aljoscha:
By the way, I'm on the ASF slack workspace: https://s.apache.org/slack-invite. You can ping me there so that we can resolve reviews quicker. Sometimes it takes quite some time with the back-and-forth on github.

eskabetxe:
@aljoscha updated the PR with the changes we talk

aljoscha:
Thanks a lot! I merged this 👌

PR #9439, opened 2019-08-14
[FLINK-13728][docs] fix wrong closing tag order in sidenav

NicoK:

The side navigation was closing some tags incorrectly as `</li></ul></div>` which should rather be `</ul></div></li>`.

Next PR in this series: #9440.

## Brief change log

- fix tag order as proposed

## Verifying this change

This change is a trivial rework which I checked in the generated HTML pages.

### DIFF
--- a/docs/_includes/sidenav.html
+++ b/docs/_includes/sidenav.html
@@ -88,7 +88,7 @@ level is determined by 'nav-pos'.
     {% else %}
       {% assign elementsPos = elementsPosStack | last %}
       {% assign pos = posStack | last %}
-</li></ul></div>
+</ul></div></li>
       {% assign elementsPosStack = elementsPosStack | pop %}
       {% assign posStack = posStack | pop %}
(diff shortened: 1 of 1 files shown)


docs/_includes/sidenav.html
```diff
@@ -88,7 +88,7 @@
     {% else %}
       {% assign elementsPos = elementsPosStack | last %}
       {% assign pos = posStack | last %}
-</li></ul></div>
+</ul></div></li>
```
zentol: where are the starting tags for div and li?

NicoK: here: https://github.com/apache/flink/pull/9439/files/530bf78fdb79319e0a9f28e0fdf679559cee7c4c#diff-24bed5845471ba1bdda01fe23f23aee8R118

StephanEwen:
Looks good to me, merging this in the next batch...

PR #9440, opened 2019-08-14
[FLINK-13724][docs] remove unnecessary whitespace from the generated pages

NicoK:

Code like the following in the side navigation before this PR creates quite some unnecessary whitespace as each line will have only the code removed:
```
    {% if parent_id %}
    {% assign parent_id = current[0].nav-parent_id %}
    {% else %}
    {% break %}
    {% endif %}
```

Starting command tags with `{%-` will drop all whitespace to the left and ending with `-%}` will drop all whitespace to the right (including newlines!).

Next PR in this series: #9441 

## Brief change log

Building on top of #9439, this PR adds:
- use `{%-` and `-%}` as appropriate in basic includes and layouts which will be included into all/most pages

## Verifying this change

This change is a trivial rework which I verified in the generated HTML pages.

### DIFF
--- a/docs/_includes/sidenav.html
+++ b/docs/_includes/sidenav.html
@@ -17,36 +17,36 @@ specific language governing permissions and limitations
 under the License.
 -->
 
-{% comment %}
+{%- comment -%}
 ==============================================================================
 Extract the active nav IDs.
 ==============================================================================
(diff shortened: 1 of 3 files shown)


StephanEwen:
Looks good, merging this...

PR #9441, opened 2019-08-14
[FLINK-13723][docs] use liquid-c for faster doc generation

NicoK:

Jekyll requires `liquid` and only optionally uses `liquid-c` if available. The latter uses natively-compiled code and reduces generation time by ~5% for me.

Next PR in this series: #9442

## Brief change log

Building on top of #9440, this PR adds:
- add one gem dependency to `liquid-c`

## Verifying this change

This change is a trivial rework which I verified in the generated HTML pages (nothing changes).


StephanEwen:
Looks good, merging this...

PR #9442, opened 2019-08-14
[FLINK-13729][docs] update website generation dependencies

NicoK:

The website generation dependencies are quite old. By upgrading some of them we get improvements like a much nicer code highlighting and prepare for the jekyll update of FLINK-13726 and FLINK-13727.

Next PR in this series: #9443

## Brief change log

Building on top of #9441, this PR adds:
- update these to the newest versions:
  - `addressable`
  - `octokit`
  - `therubyracer`
  - `json`

## Verifying this change

This change is a trivial rework and I verified the changes in the generated HTML pages. So far, I only saw changes/improvements in the syntax highlighting, e.g. non-primitive Java types also being highlighted.


zentol:
have we verified in any way whether all these changes work in the buildbot environment we're using?

NicoK:
What do you mean? How would I test this?
I could let travis build it once to verify (doc building seems to be a cron task now and not covered by the PR tests) but that's probably not what you mean.

PR #9444, opened 2019-08-14
[FLINK-13726][docs] build docs with jekyll 4.0.0.pre.beta1

NicoK:

Jekyll 4 is way faster in generating the docs than jekyll 3 - probably due to the newly introduced cache. Site generation time goes down by roughly a factor of 2.5 even with the current beta version!

Next PR in this series: #9487

## Brief change log

Building on top of #9443, this PR adds:
- temporarily disable liveserve inside `build_docs.sh -i` because the module it is based on is not (yet) available for jekyll 4
- upgrade gem dependency to `jekyll 4.0.0.pre.beta1`

## Verifying this change

This change is a trivial rework and I verified the changes in the generated HTML pages (nothing changed).


rmetzger:
@flinkbot approve description

rmetzger:
I've checked out your PR to test that the Docker setup still works.

Are these build times faster than before?
```
      Regenerating: 1 file(s) changed at 2019-08-16 09:53:39
                    release-notes/flink-1.8.md
                    ...done in 108.7836321 seconds.

      Regenerating: 1 file(s) changed at 2019-08-16 09:55:50
                    release-notes/flink-1.8.md
                    ...done in 85.8082178 seconds.
```

NicoK:
Every machine is different - these are my results building three times in a row:

before (9391a70a43):
```
215.728 seconds
187.969 seconds
183.09 seconds
```

after (c2a5438951):
```
71.132 seconds
69.715 seconds
69.3 seconds
```


commands to reproduce:
```
cd docs/docker
./run.sh
# ...
./build_docs.sh -p
```

rmetzger:
On https://github.com/apache/flink/commit/9391a70a4362f8a02cdd744a2a7f3ebbd1d577d6, these are my results.
```
      Regenerating: 1 file(s) changed at 2019-08-19 13:05:55
                    release-notes/flink-1.8.md
                    ...done in 246.926852143 seconds.

      Regenerating: 1 file(s) changed at 2019-08-19 13:14:11
                    release-notes/flink-1.8.md
                    ...done in 237.093417802 seconds.
```
My 2016 MacBookPro is not the fastest apparently (but yay, it looks nice) 

Looks like I was wrong :) Your PR is improving the performance. Thanks a lot!
I would still hope that we can get to a state where we can do an incremental page build in < 5 seconds :)

NicoK:
I guess, the biggest problem is the navbar to the left
https://github.com/apache/flink/blob/master/docs/_includes/sidenav.html

This is also what you'll see if you add `--profile` to the build options to see where time is spent.

sjwiesman:
When I previously profiled the docs build 99% of the time is spent rendering `_includes/sidenav.html`. This is because of all the liquid tags to build custom side navigation for each page. Specifically every time a page is rendered every other page must be loaded into memory and parsed to build that navigation. In my opinion, the single best thing we could do is re-write the side nav in either pure javascript or render it once to HTML with a little javascript to customize it for each page.

sjwiesman:
Ran locally and this is definitely an improvement. One question,  the notes for Jekyll 4.0 states that there is a breaking change for `{% link %}` that looks like it should break all the links in our docs but it doesn't. I ran `./check-links.sh` and clicked through a little bit and everything seems to work. Am I missing something? 

```
  * Our `link` tag now comes with the `relative_url` filter incorporated into it.
    You should no longer prepend `{{ site.baseurl }}` to `{% link foo.md %}`
    For further details: https://github.com/jekyll/jekyll/pull/6727
```

sjwiesman:
Please add `docs/.jekyll-cache/*` to flinks `.gitignore`.

NicoK:
yes, @sjwiesman a complete rewrite of the site navigation would ultimately fix the build times. I have a follow-up PR which uses the current sidenav but builds it in an optimized way that further helps (probably not as much but enough for now).

Regarding the link changes, I postponed that change for now and put it into https://issues.apache.org/jira/browse/FLINK-13727 as it is still working as it was before

PR #9448, opened 2019-08-15
[FLINK-13710] Fix JarListHandler always extract the jar package

yeweicheng:

(Standalone mode)
Solving the jobmanager submit page always unpacks the job jar, causing the temporary folder to be filled, if the jar contains a lib folder.
Now the page is not decompressing the jar except for submitting the job, and the same jar is only decompressed once, and the next submission is not decompressed.

## Brief change log

  - The PackagedProgram class adds jar list mapping variables, and the extractedJars variable is used to determine whether to extract
  - JarListHandler does not extract the jar
  - JarHandlerUtils remains as it is
  - CliFrontend remains as it is

## Verifying this change

  - The serializers: (no)

## Documentation


StephanEwen:
I think the problem and solution are valid, but the PR fails with basic checkstyle errors.
Closing this for now, please reopen when the PR is fixed.

PR #9450, opened 2019-08-15
[FLINK-13711][sql-client] Hive array values not properly displayed in…

lirui-apache:
… SQL CLI

Fix the issue that SQL CLI doesn't display arrays properly.

## Brief change log

  - Properly handle arrays when converting a Row to String.
  - Add test case.

## Verifying this change

New test case.

  - The serializers: no

## Documentation

### DIFF
--- a/flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/cli/CliUtils.java
+++ b/flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/cli/CliUtils.java
@@ -96,6 +96,15 @@ public final class CliUtils {
 			final Object field = row.getField(i);
 			if (field == null) {
 				fields[i] = CliStrings.NULL_COLUMN;
+			} else if (field.getClass().isArray()) {
+				if (field.getClass().getComponentType().isPrimitive()) {
+					// wrap primitive array in an object array so as to use Arrays.deepToString
+					String nested = Arrays.deepToString(new Object[]{field});
+					// remove leading and trailing brackets
+					fields[i] = nested.substring(1, nested.length() - 1);
+				} else {
+					fields[i] = Arrays.deepToString((Object[]) field);
+				}
 			} else {
 				fields[i] = field.toString();
 			}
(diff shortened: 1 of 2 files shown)


lirui-apache:
cc @xuefuz @bowenli86 @zjuwangg

lirui-apache:
> The change looks fine. However, I'm wondering if we need to convert recursively because otherwise nested non-primitive types will be still showed as non-meaningful stuff.

@xuefuz `Arrays.deepToString` will take care of that for us as indicated by the JavaDoc:
```
     * Returns a string representation of the "deep contents" of the specified
     * array.  If the array contains other arrays as elements, the string
     * representation contains their contents and so on.  This method is
     * designed for converting multidimensional arrays to strings.
```

lirui-apache:
@twalthr Better if you can also have a look. Thanks.

bowenli86:
> > The change looks fine. However, I'm wondering if we need to convert recursively because otherwise nested non-primitive types will be still showed as non-meaningful stuff.
> 
> @xuefuz `Arrays.deepToString` will take care of that for us as indicated by the JavaDoc:
> 
> ```
>      * Returns a string representation of the "deep contents" of the specified
>      * array.  If the array contains other arrays as elements, the string
>      * representation contains their contents and so on.  This method is
>      * designed for converting multidimensional arrays to strings.
> ```

in that case, can we add some unit tests to verify that?

lirui-apache:
@bowenli86 Just added test for nested Integer array.

flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/cli/CliUtils.java
```diff
@@ -96,6 +96,15 @@ public static void normalizeColumn(AttributedStringBuilder sb, String col, int m
 			final Object field = row.getField(i);
 			if (field == null) {
 				fields[i] = CliStrings.NULL_COLUMN;
+			} else if (field.getClass().isArray()) {
```
twalthr: Use `org.apache.flink.table.utils.EncodingUtils#objectToString`.

lirui-apache:
@twalthr Thanks for the pointer. PR updated accordingly.

bowenli86:
@twalthr given this is a bug, I think we can merge it into 1.9.1, what do you think?

flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/cli/CliUtils.java
```diff
@@ -96,6 +97,8 @@ public static void normalizeColumn(AttributedStringBuilder sb, String col, int m
...
```
twalthr: We don't need this if/else branch, `objectToString` can be the `else` part.

lirui-apache:
@bowenli86 Latest Travis build passed. Please help merge this one if no further comments. Thanks.

PR #9451, opened 2019-08-15
[FLINK-13718][hbase] Disable tests on Java 11

zentol:
Disables all HBase tests on Java 11; The HBaseConnectorITCase fails since it requires changes to the classpath that hbase uses, but this isn't possible on Java 9+.

(Note that tests are already disabled for java 9 in tools/travis/stage.sh)

### DIFF
--- a/flink-connectors/flink-hbase/pom.xml
+++ b/flink-connectors/flink-hbase/pom.xml
@@ -352,7 +352,7 @@ under the License.
 						<groupId>org.apache.maven.plugins</groupId>
 						<artifactId>maven-surefire-plugin</artifactId>
 						<configuration>
-							<!-- hive does not support Java 11 -->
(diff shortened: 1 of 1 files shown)


flink-connectors/flink-hbase/pom.xml
```diff
@@ -340,6 +340,25 @@ under the License.
...
+				<jdk>11</jdk>
+			</activation>
+
+			<build>
+				<plugins>
+					<plugin>
+						<groupId>org.apache.maven.plugins</groupId>
+						<artifactId>maven-surefire-plugin</artifactId>
+						<configuration>
+							<!-- hive does not support Java 11 -->
```
carp84: Minor: hive->hbase, and suggest to mark as "currently doesn't support Java 11" due to HBASE-21110 is ongoing.

PR #9457, opened 2019-08-15
[FLINK-13741][table] "SHOW FUNCTIONS" should include Flink built-in functions' names

bowenli86:

FunctionCatalog.getUserDefinedFunctions() only returns catalog functions and FunctionDefinitions registered in memory, but does not include Flink built-in functions' names.

It means currently {{show functions;}} thru SQL would not be able to see Flink's built-in functions.

Should be fixed to include Flink built-in functions' names

## Brief change log

- added `listFunctions()` API to TableEnv, LocalExecutor
- added `getFunctions()` API to FunctionCatalog
- made `SHOW FUNCTIONS` call the stack of `listFunctions` and eventually `getFunctions()`
- added UT

## Verifying this change

This change added tests and can be verified as follows: `FunctionCatalogTest.testGetBuiltInFunctions()`

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-table/flink-sql-client/src/test/java/org/apache/flink/table/client/gateway/local/LocalExecutorITCase.java
+++ b/flink-table/flink-sql-client/src/test/java/org/apache/flink/table/client/gateway/local/LocalExecutorITCase.java
@@ -222,7 +222,7 @@ public class LocalExecutorITCase extends TestLogger {
 
 		final List<String> actualTables = executor.listUserDefinedFunctions(session);
 
-		final List<String> expectedTables = Arrays.asList("aggregateUDF", "scalarUDF", "tableUDF");
+		final List<String> expectedTables = Arrays.asList("aggregateUDF", "tableUDF", "scalarUDF");
 		assertEquals(expectedTables, actualTables);
 	}
 
--- a/flink-table/flink-sql-client/src/test/java/org/apache/flink/table/client/gateway/local/LocalExecutorITCase.java
+++ b/flink-table/flink-sql-client/src/test/java/org/apache/flink/table/client/gateway/local/LocalExecutorITCase.java
@@ -222,7 +222,7 @@ public class LocalExecutorITCase extends TestLogger {
 
 		final List<String> actualTables = executor.listUserDefinedFunctions(session);
 
-		final List<String> expectedTables = Arrays.asList("aggregateUDF", "tableUDF", "scalarUDF");
+		final List<String> expectedTables = Arrays.asList("aggregateUDF", "scalarUDF", "tableUDF");
 		assertEquals(expectedTables, actualTables);
 	}
 
--- a/flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/FunctionCatalog.java
+++ b/flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/FunctionCatalog.java
@@ -19,6 +19,7 @@
 package org.apache.flink.table.catalog;
 
 import org.apache.flink.annotation.Internal;
+import org.apache.flink.annotation.VisibleForTesting;
 import org.apache.flink.api.common.typeinfo.TypeInformation;
 import org.apache.flink.table.api.TableException;
 import org.apache.flink.table.catalog.exceptions.DatabaseNotExistException;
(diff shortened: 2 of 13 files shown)


bowenli86:
cc release manager: @KurtYoung @tzulitai 

cc @xuefuz @lirui-apache @zjuwangg @twalthr @dawidwys

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/FunctionCatalog.java
```diff
@@ -226,7 +235,8 @@ private void registerFunction(String name, FunctionDefinition functionDefinition
 		userFunctions.put(normalizeName(name), functionDefinition);
 	}
 
-	private String normalizeName(String name) {
+	@VisibleForTesting
+	protected static String normalizeName(String name) {
```
xuefuz: It seems that the scope can be  "package" level. "protected" doesn't seem applicable since this is a static method.

zjuwangg:
Thanks for your effort!
Should we return flink built-in functions in `getUserDefinedFunctions ` method?
IMO, User-defined function meanings external function registered in funcationCatalog. 
Of course, it makes sense regard the function catalog as en empty container after first initialization and regards no matter what functions as a UDF. If so, we'd better add a notice in the method comments.
Just minor advice.

bowenli86:
> Thanks for your effort!
> Should we return flink built-in functions in `getUserDefinedFunctions ` method?
> IMO, User-defined function meanings external function registered in funcationCatalog.
> Of course, it makes sense regard the function catalog as en empty container after first initialization and regards no matter what functions as a UDF. If so, we'd better add a notice in the method comments.
> Just minor advice.

Hi @zjuwangg , you have we can discuss it in the JIRA ticket.

bowenli86:
cc @xuefuz @lirui-apache @zjuwangg @twalthr @sunjincheng121 

The UT failure of BatchFineGrainedRecoveryITCase.testProgram is irrelevant.

flink-table/flink-sql-client/src/test/java/org/apache/flink/table/client/gateway/local/LocalExecutorITCase.java
```diff
@@ -222,7 +222,7 @@ public void testListUserDefinedFunctions() throws Exception {
 
 		final List<String> actualTables = executor.listUserDefinedFunctions(session);
 
-		final List<String> expectedTables = Arrays.asList("aggregateUDF", "tableUDF", "scalarUDF");
+		final List<String> expectedTables = Arrays.asList("aggregateUDF", "scalarUDF", "tableUDF");
```
lirui-apache: I think either we should mention in the JavaDoc that the returned list is sorted, or the test should verify with a set?

bowenli86: I think I should move the sort logic to CliClient#callShowFunctions(), since neither TableEnv#listFunctions() nor FunctionCatalog#getFunctions() care about the sorted order. Sorted order is only for convenience of displaying to users in SQL CLI.

flink-table/flink-table-api-java/src/main/java/org/apache/flink/table/catalog/FunctionCatalog.java
```diff
@@ -134,7 +136,31 @@ public void registerScalarFunction(String name, ScalarFunction function) {
...
+		// Get built-in functions
+		result.addAll(
+			BuiltInFunctionDefinitions.getDefinitions()
+				.stream()
+				.map(f -> normalizeName(f.getName()))
+				.collect(Collectors.toSet())
+		);
+
+		String[] names = result.toArray(new String[0]);
+		Arrays.sort(names);
```
xuefuz: I'm not sure if we need to sort the returned array. (same for getUserDefinedFunctions()). If a caller needs a sorted result, it can do on its own. This is also for consistency, as we are not sorting other listX().

For this matter, I think it makes sense for sql-client to show a sorted list. In this case, sql-client can sort and show. However, other clients may not need a sorted list.

flink-table/flink-sql-client/src/test/java/org/apache/flink/table/client/gateway/local/LocalExecutorITCase.java
xuefuz: Personally I don't think we need to return a sorted array. SQL-client can sort the result set and present to the console.

bowenli86:
cc @xuefuz @lirui-apache

PR #9461, opened 2019-08-16
[FLINK-9900][tests] Harden ZooKeeperHighAvailabilityITCase

ifndef-SleePy:

* Waiting checkpoint id increasing to 5 can not guarantee there must be a checkpoint with state data included. An empty checkpoint would fail the restoration checking.

## Brief change log

* Use `checkpointCompletedIncludingData` to make sure we could start failing the job instead of checkpoint id increasing to 5

## Verifying this change

  - The serializers: no

## Documentation


ifndef-SleePy:
@tillrohrmann , thanks for reviewing.
I agree with you that it's a bit complicated. There are too many checking in once case. But it indeed found some bugs :)
After refactoring the `CheckpointCoordinator`, I will think about how to simplify it.

PR #9462, opened 2019-08-16
[FLINK-13738][blink-table-planner] Fix NegativeArraySizeException in LongHybridHashTable

JingsongLi:

The dense range is computed by (maxKey - minKey + 1) in LongHybridHashTable.
it may be negative. it mean the range is bigger than Long.MaxValue when range is negative...
If range is zero, maybe the max is Long.Max, and the min is Long.Min, so we should not use dense mode too.

## Verifying this change

JoinITCase.testLongJoinWithBigRange

  - The serializers: no

## Documentation


rmetzger:
I will test this change with my sample job.

rmetzger:
my testing job doesn't fail anymore with this PR applied.
I can't say whether this fix is good or not.

JingsongLi:
Thanks @rmetzger for verification, ping @wuchong to review

PR #9468, opened 2019-08-16
[FLINK-13689] [Connectors/ElasticSearch] Fix thread leak when elasticsearch6 rest high level cli…

Rishi55:
**(The sections below can be removed for hotfixes of typos)**
-->

To fix the thread leak when elasticsearch6 rest high level client ping is unsuccessful

## Brief change log

Added close statement for the rhlClient when ping is unsuccessful

## Verifying this change

This change is a trivial rework / code cleanup without any test coverage. Tested in local though

  - The serializers: (no)
  - The runtime per-record code paths (performance sensitive): don't know

## Documentation


tzulitai:
Thanks for the fix @Rishi55.

Instead of the proposed fix, I wonder if we should actually not ping and throw within the `createClient` implementation. My reasoning is that cleanup of the resources, including the client, is already handled within the `close` method of the `ElasticsearchSinkBase`, and IMO is the correct place to handle this.

So, essentially the fix would be to scope down the responsibility of the `createClient` method to simply return the instantiated instance.

tzulitai:
@flinkbot approve description
@flinkbot approve consensus

wuchong:
Hi @Rishi55 , are you still working on this issue? 
We are planning to release 1.8.2 soon, and it's nice to have this issue.

Rishi55:
Yeah, Was quite busy because of some other things. Will complete that soon.


Thanks,

Rishi55:
Hi @tzulitai, I modified my changes according to your suggestions. Please review.

flink-connectors/flink-connector-elasticsearch-base/src/main/java/org/apache/flink/streaming/connectors/elasticsearch/ElasticsearchSinkBase.java
```diff
@@ -300,6 +301,14 @@ public void open(Configuration parameters) throws Exception {
...
 		requestIndexer = callBridge.createBulkProcessorIndexer(bulkProcessor, flushOnCheckpoint, numPendingRequests);
 		failureRequestIndexer = new BufferingNoOpRequestIndexer();
+
+		if(client instanceof RestHighLevelClient) {
+			RestHighLevelClient rhlClient = (RestHighLevelClient) client;
+			if(!rhlClient.ping()) {
+				rhlClient.close();
+				throw new RuntimeException("There are no reachable Elasticsearch nodes!");
+			}
+		}
```
aljoscha: I don't think this is the solution @tzulitai had in mind. He was referring to the fact that `close()` is calling `close()` on the client. And `close()` on the operator is called when there is a failure. So if we set the client to the `client` field we guarantee that we successfully clean up.

Now the only question is whether to call `ping()` and how to call it. I don't think adding a concrete dependency for ES6 6.3.1 is the solution here. It's problematic when you use other connector versions.

aljoscha:
@Rishi55 Any updates on this?

Rishi55:
> @Rishi55 Any updates on this?

@aljoscha I need some help in how to call ping(). I have been contacting my friends but couldn't get anything. Can you please point me to someone who can help me?

aljoscha:
Sorry for the late response. I think the solution here is to add a method `verifyClientConnection(C client)` to `ElasticsearchApiCallBridge` and move the verification logic that the subclasses of `ElasticsearchApiCallBridge` have in `createClient(...)` to this new method. This way, the `ElasticsearchSinkBase` can always store the client in its internal field, making sure that the client will be closed when a failure happens.

static-max:
I just had the same problem with around 5000 unclosed threads.
As there hasn't been any progress in the last days I will start working on a fix, based on @aljoscha's proposal to store the client in `ElasticsearchSinkBase`.

aljoscha:
Cool, please ping me on the Jira issue so I can assign you: https://issues.apache.org/jira/browse/FLINK-13689. And ping be when you open a PR.

PR #9470, opened 2019-08-17
[FLINK-13380][k8s] Improve the usability of Flink session cluster on Kubernetes

Myasuka:
Previous template of Flink session cluster on Kubernetes has several shortcoming:
* flink-conf.yaml is not supported to configure, user have to rebuild docker 
* No explicit documentation of how to submit jobs to session cluster. There existed some related question on stackoverflow. I think we should add nodeport for 8081 port in deployment 
* Cannot view logs via Flink web UI.
This PR is to help improve this.

## Brief change log
Modify `docs/ops/deployment/kubernetes.md ` to provide more powerful templates.

## Verifying this change

  - The serializers: **no**

## Documentation

### DIFF
--- a/docs/ops/deployment/kubernetes.md
+++ b/docs/ops/deployment/kubernetes.md
@@ -69,9 +69,9 @@ You can then access the Flink UI via different ways:
 
 *  `kubectl port-forward`:
     1. Run `kubectl port-forward ${flink-jobmanager-pod} 8081:8081` to forward your jobmanager's web ui port to local 8081.
-    2. Navigate to [http://localhost:8001](http://localhost:8001) in your browser.
+    2. Navigate to [http://localhost:8081](http://localhost:8081) in your browser.
     3. Moreover, you could use the following command below to submit jobs to the cluster:
-    {% highlight bash %}./bin/flink run -m localhost:8001 ./examples/streaming/WordCount.jar{% endhighlight %}
+    {% highlight bash %}./bin/flink run -m localhost:8081 ./examples/streaming/WordCount.jar{% endhighlight %}
 
 *  Create a `NodePort` service on the rest service of jobmanager:
     1. Run `kubectl create -f jobmanager-rest-service.yaml` to create the `NodePort` service on jobmanager. The example of `jobmanager-rest-service.yaml` can be found in [appendix](#session-cluster-resource-definitions).
@@ -262,7 +262,7 @@ spec:
     component: jobmanager
 {% endhighlight %}
 
-`jobmanager-rest-service.yaml`. Optional service, which is used to add a `NodePort` on the jobmanager rest service so that user could submit job via `<public-node-ip>:<node-port>`.
+`jobmanager-rest-service.yaml`. Optional service, that exposes the jobmanager `rest` port as public Kubernetes node's port.
 {% highlight yaml %}
 apiVersion: v1
 kind: Service
--- a/docs/ops/deployment/kubernetes.md
+++ b/docs/ops/deployment/kubernetes.md
@@ -61,17 +61,24 @@ Using the resource definitions for a [session cluster](#session-cluster-resource
 
 Note that you could define your own customized options of `flink-conf.yaml` within `flink-configuration-configmap.yaml`.
 
-You can then access the Flink UI via `kubectl proxy`:
+You can then access the Flink UI via different ways:
+*  `kubectl proxy`:
 
-1. Run `kubectl proxy` in a terminal
-2. Navigate to [http://localhost:8001/api/v1/namespaces/default/services/flink-jobmanager:ui/proxy](http://localhost:8001/api/v1/namespaces/default/services/flink-jobmanager:ui/proxy) in your browser
+    1. Run `kubectl proxy` in a terminal.
+    2. Navigate to [http://localhost:8001/api/v1/namespaces/default/services/flink-jobmanager:ui/proxy](http://localhost:8001/api/v1/namespaces/default/services/flink-jobmanager:ui/proxy) in your browser.
 
-In order to view the Flink's web UI, it needs to be exposed via a Kubernetes service which exposes the port on the node's public ip. 
-The service `flink-rest-service` exactly achieves this. Using this service, the web ui can be accessed via `http://<public-node-ip>:<node-port>`.
-What's more, you could use the following command below to submit jobs to the cluster:
-{% highlight bash %}
-./bin/flink run -m <public-node-ip>:<node-port> ./examples/streaming/WordCount.jar
-{% endhighlight %}
+*  `kubectl port-forward`:
+    1. Run `kubectl port-forward ${flink-jobmanager-pod} 8081:8081` to forward your jobmanager's web ui port to local 8081.
+    2. Navigate to [http://localhost:8001](http://localhost:8001) in your browser.
+    3. Moreover, you could use the following command below to submit jobs to the cluster:
+    {% highlight bash %}./bin/flink run -m localhost:8001 ./examples/streaming/WordCount.jar{% endhighlight %}
+
+*  Create a `NodePort` service on the rest service of jobmanager:
+    1. Run `kubectl create -f jobmanager-rest-service.yaml` to create the `NodePort` service on jobmanager. The example of `jobmanager-rest-service.yaml` can be found in [appendix](#session-cluster-resource-definitions).
+    2. Run `kubectl get svc flink-jobmanager-rest` to know the `node-port` of this service and navigate to [http://&lt;public-node-ip&gt;:&lt;node-port&gt;](http://<public-node-ip>:<node-port>) in your browser.
+    3. Similarly to `port-forward` solution, you could also use the following command below to submit jobs to the cluster:
+
+        {% highlight bash %}./bin/flink run -m <public-node-ip>:<node-port> ./examples/streaming/WordCount.jar{% endhighlight %}
 
 In order to terminate the Flink session cluster, use `kubectl`:
 
@@ -253,15 +260,18 @@ spec:
   selector:
     app: flink
     component: jobmanager
----
+{% endhighlight %}
+
+`jobmanager-rest-service.yaml`. Optional service, which is used to add a `NodePort` on the jobmanager rest service so that user could submit job via `<public-node-ip>:<node-port>`.
+{% highlight yaml %}
 apiVersion: v1
 kind: Service
 metadata:
-  name: flink-rest-service
+  name: flink-jobmanager-rest
 spec:
   type: NodePort
   ports:
-  - name: ui
+  - name: rest
     port: 8081
     targetPort: 8081
   selector:
--- a/docs/ops/deployment/kubernetes.md
+++ b/docs/ops/deployment/kubernetes.md
@@ -54,21 +54,21 @@ A basic Flink session cluster deployment in Kubernetes has three components:
 
 Using the resource definitions for a [session cluster](#session-cluster-resource-definitions), launch the cluster with the `kubectl` command:
 
-    kubectl create -f configmap.yaml
+    kubectl create -f flink-configuration-configmap.yaml
     kubectl create -f jobmanager-service.yaml
     kubectl create -f jobmanager-deployment.yaml
     kubectl create -f taskmanager-deployment.yaml
 
-Note that you could define your own customized options of `flink-conf.yaml` within `configmap.yaml`.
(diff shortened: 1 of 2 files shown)


Myasuka:
After apply the scripts:
image
image

docs/ops/deployment/kubernetes.md
```diff
@@ -112,20 +153,38 @@ spec:
...
-        args:
-        - jobmanager
+        workingDir: /opt/flink
+        command: ["/bin/bash", "-c", "$FLINK_HOME/bin/jobmanager.sh start;\
+          while :;
+          do
+            if [[ -f $(find log -name '*jobmanager*.log' -print -quit) ]];
+              then tail -f -n +1 log/*jobmanager*.log;
+            fi;
+          done"]
```
tillrohrmann: Could this be part of the `entrypoint.sh`?

docs/ops/deployment/kubernetes.md
```diff
@@ -112,20 +153,38 @@ spec:
...
+        livenessProbe:
+          tcpSocket:
+            port: 6123
+          initialDelaySeconds: 30
+          periodSeconds: 60
+        volumeMounts:
+        - name: flink-config-volume
+          mountPath: /opt/flink/conf
+      volumes:
+      - name: flink-config-volume
```
tillrohrmann: Indentation seems to be wrong

docs/ops/deployment/kubernetes.md
```diff
@@ -145,18 +204,35 @@ spec:
...
-        args:
-        - taskmanager
+        workingDir: /opt/flink
+        command: ["/bin/bash", "-c", "$FLINK_HOME/bin/taskmanager.sh start; \
+          while :;
+          do
+            if [[ -f $(find log -name '*taskmanager*.log' -print -quit) ]];
+              then tail -f -n +1 log/*taskmanager*.log;
+            fi;
+          done"]
```
tillrohrmann: Same here, by adding it to the `entrypoint.sh` one could remove a bit of code duplication.

docs/ops/deployment/kubernetes.md
```diff
@@ -166,18 +242,31 @@ kind: Service
...
   - name: ui
     port: 8081
   selector:
     app: flink
     component: jobmanager
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: flink-jobmanager-svc
```
tillrohrmann: dumb question, does `svc` stand for service? If yes, then we could maybe rename it into `flink-web-ui-service`.

docs/ops/deployment/kubernetes.md
```diff
@@ -54,20 +54,31 @@ A basic Flink session cluster deployment in Kubernetes has three components:
...
     kubectl create -f taskmanager-deployment.yaml
 
+Note that you could define your own customized options of `flink-conf.yaml` within `configmap.yaml`.
+
 You can then access the Flink UI via `kubectl proxy`:
 
 1. Run `kubectl proxy` in a terminal
 2. Navigate to [http://localhost:8001/api/v1/namespaces/default/services/flink-jobmanager:ui/proxy](http://localhost:8001/api/v1/namespaces/default/services/flink-jobmanager:ui/proxy) in your browser
 
+Another way to view the Flink UI is via [service with NodePort](https://kubernetes.io/docs/tasks/access-application-cluster/service-access-application-cluster/#creating-a-service-for-an-application-running-in-two-pods)
```
tillrohrmann:
```suggestion
In order to view the Flink's web UI, it needs to be exposed via a Kubernetes service which exposes the port on the node's public ip. The service `flink-jobmanager-svc` achieves exactly this. Using this service, the web ui can be accessed via `http://<public-node-ip>:<node-port>`.
```

docs/ops/deployment/kubernetes.md
```diff
@@ -54,20 +54,31 @@ A basic Flink session cluster deployment in Kubernetes has three components:
...
+Note that you could define your own customized options of `flink-conf.yaml` within `configmap.yaml`.
+
 You can then access the Flink UI via `kubectl proxy`:
 
 1. Run `kubectl proxy` in a terminal
 2. Navigate to [http://localhost:8001/api/v1/namespaces/default/services/flink-jobmanager:ui/proxy](http://localhost:8001/api/v1/namespaces/default/services/flink-jobmanager:ui/proxy) in your browser
 
+Another way to view the Flink UI is via [service with NodePort](https://kubernetes.io/docs/tasks/access-application-cluster/service-access-application-cluster/#creating-a-service-for-an-application-running-in-two-pods) 
+(which has been defined as `flink-jobmanager-svc` in our template) of address `http://<public-node-ip>:<node-port>`. 
+What's more，your could use command below to submit your jobs:
```
tillrohrmann:
```suggestion
What's more, you could use the following command below to submit jobs to the cluster:
```

docs/ops/deployment/kubernetes.md
```diff
@@ -54,20 +54,31 @@ A basic Flink session cluster deployment in Kubernetes has three components:
 
 Using the resource definitions for a [session cluster](#session-cluster-resource-definitions), launch the cluster with the `kubectl` command:
 
+    kubectl create -f configmap.yaml
```
tillrohrmann: maybe rename `configmap.yaml` into `cluster-configuration-configmap.yaml` or `flink-configuration-configmap.yaml`.

docs/ops/deployment/kubernetes.md
```diff
@@ -166,18 +242,31 @@ kind: Service
...
   selector:
     app: flink
     component: jobmanager
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: flink-jobmanager-svc
+spec:
+  type: NodePort
```
1u0: @Myasuka, do you really need to expose the service as a `NodePort`?
Shouldn't the above `ClusterIP` type `flink-jobmanager` service be enough for cluster to work properly?

docs/ops/deployment/kubernetes.md
Myasuka: IMO, there should be another issue to discuss about modifying the `entrypoint.sh`. Even I include this part as `entrypoint.sh`, since I already use `command` to start the container, the default `entrypoint.sh` would be useless for me.

docs/ops/deployment/kubernetes.md
Myasuka: Since I already added `command` in deployment, the default `entrypoint.sh` [would be ignored](https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#notes).

docs/ops/deployment/kubernetes.md
Myasuka: Yes, `ClusterIP` should be enough for the cluster to work properly. But without a `NodePort` on jobmanager's rest RPC port, I couldn't find a way to submit my application via `bin/flink run -m <k8s-flink-cluster>`. It seems I could only submit my application via web UI which is not so convenient for developers.

docs/ops/deployment/kubernetes.md
Myasuka: Actually, this service enables user could submit jobs via shell script to rest API, how about `flink-rest-service`?

docs/ops/deployment/kubernetes.md
Myasuka: Actually, this is the right indentation. `containers` and `volumes` are both fields in [podspec](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.15/#podspec-v1-core).

Myasuka: Actually, this is the right indentation. `containers` and `volumes` are both fields in [podspec](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.15/#podspec-v1-core).

docs/ops/deployment/kubernetes.md
1u0: If your situation is in some test/development setup, you can try to:
 * run `kubectl port-forward ${flink-jobmanager-pod} 8081:8081` to make JM reachable from your local machine (where you run `bin/flink run -m ...`).
 * I don't know well the `bin/flink run -m ${endpoint}` functionality, but if it doesn't work with urls that have additional path, the script/program behind it could have been extended to support such endpoints (smt. like: `bin/flink -m "http://localhost:8081/api/v1/namespaces/default/services/flink-jobmanager:ui/proxy" ...`).

If the Kubernetes resources and the documentation here are served as some recommended/exemplar Flink cluster setup, then adding a service definition that exposes node port would be a bad example, imo.
I suggest to remove the new service definition and maybe just add some more options (in the documentation) how developers can make JM reachable outside of test K8s cluster.

Myasuka: Okay, I think we might add a separate service yaml file in the documentation to represent the `rest service` and tell users different ways to submit jobs: via `kubectl port-forward` or via the nodeport of this rest service.

docs/ops/deployment/kubernetes.md
tillrohrmann: Of course, one would have to use the entrypoint.sh then and not specify `command`. But I'm ok with discussing this as a separate issue and keeping it out of this PR's scope.

docs/ops/deployment/kubernetes.md
```diff
@@ -54,20 +54,38 @@ A basic Flink session cluster deployment in Kubernetes has three components:
...
-2. Navigate to [http://localhost:8001/api/v1/namespaces/default/services/flink-jobmanager:ui/proxy](http://localhost:8001/api/v1/namespaces/default/services/flink-jobmanager:ui/proxy) in your browser
+You can then access the Flink UI via different ways:
+*  `kubectl proxy`:
+
+    1. Run `kubectl proxy` in a terminal.
+    2. Navigate to [http://localhost:8001/api/v1/namespaces/default/services/flink-jobmanager:ui/proxy](http://localhost:8001/api/v1/namespaces/default/services/flink-jobmanager:ui/proxy) in your browser.
+
+*  `kubectl port-forward`:
+    1. Run `kubectl port-forward ${flink-jobmanager-pod} 8081:8081` to forward your jobmanager's web ui port to local 8081.
+    2. Navigate to [http://localhost:8001](http://localhost:8001) in your browser.
```
1u0: Should be `localhost:8081` in this example.

docs/ops/deployment/kubernetes.md
```diff
@@ -166,18 +249,34 @@ kind: Service
...
-  - name: query
-    port: 6125
   - name: ui
     port: 8081
   selector:
     app: flink
     component: jobmanager
 {% endhighlight %}
 
+`jobmanager-rest-service.yaml`. Optional service, which is used to add a `NodePort` on the jobmanager rest service so that user could submit job via `<public-node-ip>:<node-port>`.
```
1u0: Can be rephrased to smt. like:
```
Optional service, that exposes the jobmanager `ui` port as public Kubernetes node's port.
```

But personally, I'd avoid adding this at all, as poor recommendation/guidance for users (as mentioned before).

docs/ops/deployment/kubernetes.md
Myasuka: good catch.

docs/ops/deployment/kubernetes.md
Myasuka: Use `NodePort` to expose an application from K8S cluster is a general solution from official doc. Why you think this should be avoided to add?

1u0: My reasoning, that there maybe two types of readers of this docs:
 * people like devs, who have a test K8s cluster and want to setup test/experimental Flink cluster. For this, I assume they can run `kubectl port-forward|proxy` and also create resources by `kubectl`. For them the existing means should be already enough to make Flink JM available from a local dev machine. Showing them method that exposes `NodePort` would be like giving a poor example;
 * people like ops, that operate production K8s cluster. In that case they are probably knowledgeable enough how to expose a service properly.

There are some blog posts in internet, for example [this one](https://medium.com/google-cloud/kubernetes-nodeport-vs-loadbalancer-vs-ingress-when-should-i-use-what-922f010849e0) that give some information what are the different ways and in which situations to make service available outside of K8s cluster.

Myasuka: Why I want to tell users how to submit jobs is inspired from this [stackoverflow question](https://stackoverflow.com/questions/56910990/flink-on-kubernetes-how-to-submit-jobs-to-session-clusters). I think it's worthy to add this instruction to offer another choice.

tillrohrmann: I understand @1u0 concerns but I agree with @Myasuka that documenting the different ways might probably not hurt.

PR #9471, opened 2019-08-17
[FLINK-13754][task] Decouple OperatorChain with StreamStatusMaintainer

zhijiangW:

There are two motivations for this refactoring:
    
The current `OperatorChain` is heavy-weight to take some unrelated roles like `StreamStatusMaintainer`. If other components rely on the `StreamStatusMaintainer`, we have to pass the whole `OperatorChain`. From the design aspect of single function, we need to decouple `StreamStatusMaintainer` from `OperatorChain`.

Based on FLINK-13798 to break the cycle dependency between `RecordWriterOutput` and `StreamStatusMaintainer`, it is reasonable to create/maintain a separate instance of `StreamStatusMaintainer` on `StreamTask` side.

One reason is that it gets benefits of removing some duplicate logics while creating `RecordWriter` and `RecordWriterOutput` together.

Another reason is that it is easy to reference the `StreamStatusMaintainer` via `StreamTask` in the following refactoring work.

## Brief change log

  - *Refactor `OperatorChain` to not implement `StreamStatusProvider`*
  - *Introduce the `RecordWriterOutputWrapper` for creating the array of `RecordWriterOutputs` and handling some relevant logics*
  - *Refactor `StreamTask` to create `StreamStatusProvider` instance and pass it into `OperatorChain`*

## Verifying this change

## Documentation


flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/RecordWriterOutput.java
```diff
@@ -49,18 +48,17 @@
...
 	private final WatermarkGauge watermarkGauge = new WatermarkGauge();
 
+	private StreamStatus currentStatus;
+
 	@SuppressWarnings("unchecked")
 	public RecordWriterOutput(
 			RecordWriter<SerializationDelegate<StreamRecord<OUT>>> recordWriter,
 			TypeSerializer<OUT> outSerializer,
-			OutputTag outputTag,
-			StreamStatusProvider streamStatusProvider) {
```
pnowojski: I'm a little bit torn apart on this change. One one hand it indeed cuts a dependency, on another it duplicates the state of `StreamStatus` field. After this change, it is possible that because of some bug the value of `StreamStatus` will be different in `RecordWriterOutput` compared to `OperatorChain`.

Maybe the question should be, why does the `RecordWriterOutput` need `StreamStatus` in the first place? 

I asked @tzulitai about this, and this `if (currentStatus.isActive())` inside `emitWatermark` is probably redundant and unnecessary check. Contract is that nothing should be emitted if stream is not active, so this should be checked not only in the `emitWatermark`  but everywhere. However such check shouldn't be performed here, but somewhere on upper level, so if anything, this should/could be a `checkState` here. However adding an extra dependency for just a `checkState` seems like an overkill, so long story short:

Probably we can just remove this code altogether - at least we can remove `StreamStatusProvider` dependency/usage from `RecordWriterOutput`.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
```diff
@@ -1293,10 +1292,10 @@ private void checkpointStreamOperator(StreamOperator<?> op) throws Exception {
 			StreamEdge edge,
 			int outputIndex,
 			Environment environment,
-			String taskName,
```
pnowojski: Why do we need this change?

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
```diff
@@ -605,6 +610,50 @@ boolean isSerializingTimestamps() {
 		return tc == TimeCharacteristic.EventTime | tc == TimeCharacteristic.IngestionTime;
 	}
 
+	private void broadcastCheckpointBarrier(
```
pnowojski: Maybe those methods that were moved from `OperatorChain` shouldn't be now implemented in the `StreamTask` but encapsulated somehow? Now they are making `StreamTask` class even more complicated compared to how it was before.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
zhijiangW: There are two options:

- Not pass the heavy-height `Environment` here and pass the individual required fields from it. 

- Pass the big component `Environment` instead of every tiny fields.

I prefer the first option but it might bring more refactoring works, so I take the second way here. The task name could be got directly from `Environment`, no need to explicitly pass it.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
zhijiangW: Yes, I also thought the current `StreamTask` is overstaffing at the moment also besides `OperatorChain`. Let me think encapsulate it in another way.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/RecordWriterOutput.java
zhijiangW: Thanks for this investigation with @tzulitai .

1. After further reviewing the whole related process, the current idle `StreamStatus` only blocks the emitting of watermark, not works for `StreamRecord/LatencyMarker`.

2. The mechanism is different for source and non-source tasks to control emitting watermark based on stream status:

- For source task, the source context references the `StreamStatusMaintainer` to mark idle status if needed. Then the `RecordWriterOutput` also references this `StreamStatusMaintainer` to judge the status before really emitting watermark.

- For non-source task, we have a high-level `StatusWatermarkValve` component to handle watermark with `StreamStatus` together. In this case if the maintained status in some channel is idle, it would not really call the `Operator#processWatermark` while handling the watermark. So the existing judge on `RecordWriterOutput` side seems redundant for non-source case.

In conclusion, the current judge of status logic is still necessary for the case of source. I agree that it seems more reasonable to make this judge logic in upper layer like the current way for non-source task. If to do so, we need to define a similar `StatusWatermarkValve` component to be referenced by source context, and the watermark emitting must firstly go through `StatusWatermarkValve` to judge whether the status is idle. It would involve in more refactoring work on legacy source stack.

So the only feasible way currently might be done in PR. It seems redundant for non-source task, but we do not change the behavior and the redundant issue still exists before this refactoring.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
zhijiangW: I already updated the codes for addressing this issue.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/RecordWriterOutputWrapper.java
```diff
@@ -0,0 +1,162 @@
...
+		RecordWriter<SerializationDelegate<StreamRecord<?>>> output = new RecordWriterBuilder()
+			.setChannelSelector(outputPartitioner)
+			.setTimeout(bufferTimeout)
+			.setTaskName(taskName)
+			.build(bufferWriter);
+		output.setMetricGroup(environment.getMetricGroup().getIOMetricGroup());
+		return output;
+	}
+
+	public static RecordWriterOutputWrapper build(StreamConfig config, Environment environment) {
```
pnowojski: nit: Maybe this just should be another constructor?

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/StreamSourceContexts.java
```diff
@@ -383,8 +383,6 @@ public WatermarkContext(
 		@Override
 		public void collect(T element) {
 			synchronized (checkpointLock) {
-				streamStatusMaintainer.toggleStreamStatus(StreamStatus.ACTIVE);
```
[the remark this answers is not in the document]
zhijiangW: My previous thought was that the active status would be toggled in a separate logic in other places before collecting. After thinking through it again, it seems that this is only triggered via collecting elements or emitting watermarks. 

So you are right that we should not remove this action and keep it.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/streamstatus/StreamStatusMaintainerImpl.java
```diff
@@ -0,0 +1,58 @@
...
+ * {@link RecordWriterOutput} when the status has changed.
+ */
+@Internal
+public class StreamStatusMaintainerImpl implements StreamStatusMaintainer {
+
+	private final RecordWriterOutput<?>[] recordWriterOutputs;
+
+	private StreamStatus currentStatus = StreamStatus.ACTIVE;
+
+	public StreamStatusMaintainerImpl(RecordWriterOutput<?>[] recordWriterOutputs) {
```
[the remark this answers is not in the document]
zhijiangW: Yes, it could be. I guess the motivation to do this is we'd better not to expose `RecordWriterOutput` directly, and all the actions should be via `RecordWriterOutputWrapper`.

We can provide `RecordWriterOutputWrapper#emitWatermark` to further wrap this action done  my `RecordWriterOutput` previously. But `RecordWriterOutputWrapper#getRecordWriterOutputs` seems still be needed by `OperatorChain`.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/RecordWriterOutputWrapper.java
zhijiangW: Actually I tried to provide a factory role for constructing `RecordWriterOutputWrapper`.

I guess it is not suggested involving in complex processes in the constructor, and the better way is to build them in the factory and pass the final required arguments directly into the constructor. I am not sure whether it is a common rule.

zhijiangW:
Thanks for the further reviews @pnowojski . 
This PR is relying on #9495 and I already requested @tzulitai for reviewing in that PR page.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/operators/TimestampsAndPeriodicWatermarksOperator.java
```diff
@@ -41,6 +42,8 @@
 
 	private transient long currentWatermark;
 
+	private StreamStatusProvider streamStatusProvider;
+
 	public TimestampsAndPeriodicWatermarksOperator(AssignerWithPeriodicWatermarks<T> assigner) {
```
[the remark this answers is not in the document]
zhijiangW: I think we should also need this change in `TimestampsAndPunctuatedWatermarksOperator`. I was not aware of this path before.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/streamstatus/StreamStatusMaintainerImpl.java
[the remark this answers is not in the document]
pnowojski: I meant that we can use `RecordWriterOutputWrapper` here and in `toggleStreamStatus()` just call
```
wrapper.emitStreamStatus(streamStatus)
``` 
instead of
```
		for (RecordWriterOutput<?> output : recordWriterOutputs) {
			output.emitStreamStatus(streamStatus);
		}
```
(and move this for loop to the wrapper).

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/RecordWriterOutputWrapper.java
pnowojski: I don't think we have any hard rule in Flink about that. Probably in most places we are not using static methods if we can use just constructor. For me, builders/factories are a necessary overhead to overcome Java's language limitation of handling multiple parameters with default values, which is not the case here, so I would use just plain constructor. On it's own factories/builders for me do not have any added value. if java's syntax supported default parameters [(like python)](https://www.geeksforgeeks.org/default-arguments-in-python/), I wouldn't be using factories/builders   

However if you have strong feelings about this, we can try it out and maybe overtime you will be able to change our practice, so feel free to ignore my preference here :)

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/operators/TimestampsAndPeriodicWatermarksOperator.java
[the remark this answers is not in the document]
zhijiangW: I missed the path before. I guess we still need the check logic in `TimestampsAndPunctuatedWatermarksOperator` to keep the previous behavior.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/RecordWriterOutputWrapper.java
zhijiangW: Thanks for this explanation. I have no personal preference for this issue. Just because I was suggested that in previous PRs. Then I can refactor this as a constructor. :)

PR #9472, opened 2019-08-18
[FLINK-13759][builds] Fix builds for master branch are failed during compile stage

wuchong:

Fix builds for master branch are failed during compile stage.

## Brief change log

- `flink-connector-elasticsearch` doesn't exist at all. Fix the travis script to check `flink-connector-elasticsearch6`.

## Verifying this change

This change is a trivial work without any test coverage.

## Documentation

### DIFF
--- a/tools/travis_controller.sh
+++ b/tools/travis_controller.sh
@@ -114,11 +114,11 @@ if [ $STAGE == "$STAGE_COMPILE" ]; then
         EXIT_CODE=$(($EXIT_CODE+$?))
         check_shaded_artifacts_s3_fs presto
         EXIT_CODE=$(($EXIT_CODE+$?))
-        check_shaded_artifacts_connector_elasticsearch ""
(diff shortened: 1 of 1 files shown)


wuchong:
I have the same question. It should fail for a long time ago. Could this because of Travis caching mechanism?

tillrohrmann:
The reason why the build failed was indeed a cache inconsistency and not `check_shaded_artifacts_connector_elasticsearch` which checked for a non-existing connector directory. @zentol could it be that `check_shaded_artifacts_connector_elasticsearch` does not properly fail the build if it cannot find a directory?

zentol:
Based on the JIRA error the check does fail if the directory or file doesn't exist.

I would conclude that the directory _did_ exist, but as far as I can tell there shouldn't be a way for the cache to contain old date; Flink should always be compiled in a clean directory which is then cached separately from other builds.

PR #9475, opened 2019-08-18
[FLINK-13762][task] Add the exception throwable in the interface methods of ValveOutputHandler

zhijiangW:

It is only for simplifying the implementations of specific methods to not wrap the exception into `RuntimeException`.

## Brief change log

  - *Add the exception throwable in the interface methods of `ValveOutputHandler`*
  - *Adjust the implementations in `StreamOneInputProcessor` and `StreamTwoInputSelectableProcessor`.*

## Verifying this change

## Documentation

### DIFF
--- a/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamOneInputProcessor.java
+++ b/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamOneInputProcessor.java
@@ -215,26 +215,18 @@ public final class StreamOneInputProcessor<IN> implements StreamInputProcessor {
 		}
 
 		@Override
-		public void handleWatermark(Watermark watermark) {
-			try {
-				synchronized (lock) {
-					watermarkGauge.setCurrentWatermark(watermark.getTimestamp());
-					operator.processWatermark(watermark);
-				}
-			} catch (Exception e) {
-				throw new RuntimeException("Exception occurred while processing valve output watermark: ", e);
+		public void handleWatermark(Watermark watermark) throws Exception {
+			synchronized (lock) {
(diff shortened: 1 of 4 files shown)


flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/streamstatus/ForwardingValveOutputHandler.java
```diff
@@ -0,0 +1,56 @@
...
+
+import static org.apache.flink.util.Preconditions.checkNotNull;
+
+/**
+ * A unified implementation of {@link ValveOutputHandler} which is called by the
+ * {@link StatusWatermarkValve} only when it determines a new watermark or stream
+ * status can be propagated.
+ */
+@Internal
+public class ForwardingValveOutputHandler implements ValveOutputHandler {
```
pnowojski: What is the motivation behind changing this code from implementing `handleWatermark` `handleStreamStatus` methods into passing those methods' body via the constructor?

zhijiangW: The initial motivation to do this is for getting rid of dependency between two ForwardingValveOutputHandler instances while toggling the stream status. If one instance is preparing for toggling the idle status, it has to check another instance whether it is also idle.

So I tried to refactor this logic in upper layer for implementing a separate unified `ForwardingValveOutputHandler`, but the retry fails and I just keeps this form.

I agree it is not necessary to do this change now and I would revert it to only keep the exception as you suggested above.

zhijiangW:
Thanks for the review @pnowojski . I updated the codes for addressing the comments.

PR #9477, opened 2019-08-18
[FLINK-13765][task] Introduce TwoInputSelectionHandler for selecting input in StreamTwoInputSelectableProcessor

zhijiangW:

 In `StreamTwoInputSelectableProcessor` there are three fields {`InputSelectable`, `InputSelection`, `availableInputsMask`} to be used together for the function of selecting next available input index.
    
From design aspect, these fields can be abstracted into a separate component called `TwoInputSelectionHandler`, which is passed into the constructor of `StreamTwoInputSelectableProcessor` as a final field. So the internal implementation details  of `TwoInputSelectionHandler` is hidden from processor view which only needs to interact with exposed methods from selection handler.
    
Another tiny benefit is that we make the `StreamTwoInputSelectableProcessor` a bit because two methods are removed from it.

## Brief change log

  - *Introduce the `TwoInputSelectionHandler` for handling the logic of next available input index*
  - *Refactor the related process in `StreamTwoInputSelectableProcessor` based on `InputSelectionHandler`*

## Verifying this change

## Documentation

### DIFF
--- a/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamTwoInputSelectableProcessor.java
+++ b/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamTwoInputSelectableProcessor.java
@@ -25,8 +25,6 @@ import org.apache.flink.metrics.Counter;
 import org.apache.flink.runtime.io.disk.iomanager.IOManager;
 import org.apache.flink.runtime.io.network.partition.consumer.InputGate;
 import org.apache.flink.streaming.api.CheckpointingMode;
-import org.apache.flink.streaming.api.operators.InputSelectable;
-import org.apache.flink.streaming.api.operators.InputSelection;
 import org.apache.flink.streaming.api.operators.TwoInputStreamOperator;
 import org.apache.flink.streaming.api.watermark.Watermark;
 import org.apache.flink.streaming.runtime.metrics.WatermarkGauge;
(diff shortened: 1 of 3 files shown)


flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/InputSelectionHandler.java
```diff
@@ -0,0 +1,82 @@
...
+	void setAvailableInput(int inputIndex) {
+		availableInputsMask |= 1 << inputIndex;
+	}
+
+	void setUnavailableInput(int inputIndex) {
+		availableInputsMask &= ~(1 << inputIndex);
+	}
+
+	boolean isALLInputsSelected() {
+		checkState(inputSelection != null);
```
pnowojski: nitty nit: I think there is little value of this `checkState`, as `NullPointerException` would be thrown one line below and it doesn't safeguard any arguments. But you can keep this if you prefer.

zhijiangW:
Thanks for this review @pnowojski !

Actually this PR would not be a blocker for the following work as I misleading before.  If just from the design aspect, the benefits might not be very obvious as you mentioned. My previous thought:

- We could make `StreamTwoInputSelectableProcessor` a bit thin after refactoring, because we remove two non-final vars and two methods from the processor.

- In previous way, `StreamTwoInputSelectableProcessor` would involve in three fields {InputSelectable, InputSelection, availableInputsMask} to expose too many details. After refactoring, the processor only interacts with one component and some details are hidden behind. 

I could understand your concern and it would be better if we could further reduce some complex interactive logics provided by `TwoInputSelectionHandler`. 
So I think this pr can be pending now, and after we find something very confirmable to be done, we can merge then.

pnowojski:
As I wrote before, please feel free to merge this. Keeping it open for "rethinking" in the future it's probably a bad idea.

zhijiangW:
Thanks for the confirmation and I would merge it after travis green. @pnowojski

zhijiangW:
@flinkbot run travis

PR #9478, opened 2019-08-19
[FLINK-13766][task] Refactor the implementation of StreamInputProcessor based on StreamTaskInput#emitNext

zhijiangW:

The current data processing in task input processor is based on the way of `NullableAsyncDataInput#pollNext`. In order to unify the processing stack for new source operator, we introduce the new `PushBasedAsyncDataInput#emitNext(Output)` instead. Then we need to adjust the existing implementations of `StreamOneInputProcessor`/`StreamTwoInputSelectableProcessor` based on this new way. To do so, we could integrate all the task inputs from network/source in a unified processor on runtime side.

To do so, we could integrate all the task inputs from network/source in a unified processing on runtime side.

## Brief change log

  - *Introduce  relevant new interfaces `PushBasedAsyncDataInput` `InputStatus` and `DataOutput`
  - *Refactor the implementation of `StreamOneInputProcessor` based on new interface*
  - *Refactor the implementation of `StreamTwoInputSelectableProcessor` based on new interface*

## Verifying this change

## Documentation

### DIFF
--- a/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamOneInputProcessor.java
+++ b/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamOneInputProcessor.java
@@ -85,7 +85,6 @@ public final class StreamOneInputProcessor<IN> implements StreamInputProcessor {
 	private final WatermarkGauge watermarkGauge;
 	private final Counter numRecordsIn;
 
-	@SuppressWarnings("unchecked")
 	public StreamOneInputProcessor(
 			InputGate[] inputGates,
 			TypeSerializer<IN> inputSerializer,
--- a/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamOneInputProcessor.java
+++ b/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamOneInputProcessor.java
@@ -20,12 +20,8 @@ package org.apache.flink.streaming.runtime.io;
 
 import org.apache.flink.annotation.Internal;
 import org.apache.flink.api.common.typeutils.TypeSerializer;
-import org.apache.flink.configuration.Configuration;
 import org.apache.flink.metrics.Counter;
 import org.apache.flink.runtime.io.disk.iomanager.IOManager;
-import org.apache.flink.runtime.io.network.partition.consumer.InputGate;
-import org.apache.flink.runtime.metrics.groups.TaskIOMetricGroup;
-import org.apache.flink.streaming.api.CheckpointingMode;
 import org.apache.flink.streaming.api.operators.OneInputStreamOperator;
 import org.apache.flink.streaming.api.watermark.Watermark;
 import org.apache.flink.streaming.runtime.metrics.WatermarkGauge;
@@ -35,7 +31,6 @@ import org.apache.flink.streaming.runtime.streamstatus.StatusWatermarkValve;
 import org.apache.flink.streaming.runtime.streamstatus.StreamStatus;
 import org.apache.flink.streaming.runtime.streamstatus.StreamStatusMaintainer;
 import org.apache.flink.streaming.runtime.tasks.OperatorChain;
-import org.apache.flink.streaming.runtime.tasks.StreamTask;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -44,7 +39,6 @@ import java.io.IOException;
 import java.util.concurrent.CompletableFuture;
 
 import static org.apache.flink.util.Preconditions.checkNotNull;
-import static org.apache.flink.util.Preconditions.checkState;
 
 /**
  * Input reader for {@link org.apache.flink.streaming.runtime.tasks.OneInputStreamTask}.
@@ -86,44 +80,27 @@ public final class StreamOneInputProcessor<IN> implements StreamInputProcessor {
 	private final Counter numRecordsIn;
 
 	public StreamOneInputProcessor(
-			InputGate[] inputGates,
+			CheckpointedInputGate checkpointedInputGate,
 			TypeSerializer<IN> inputSerializer,
-			StreamTask<?, ?> checkpointedTask,
-			CheckpointingMode checkpointMode,
 			Object lock,
 			IOManager ioManager,
-			Configuration taskManagerConfig,
 			StreamStatusMaintainer streamStatusMaintainer,
 			OneInputStreamOperator<IN, ?> streamOperator,
-			TaskIOMetricGroup metrics,
 			WatermarkGauge watermarkGauge,
-			String taskName,
 			OperatorChain<?, ?> operatorChain,
-			Counter numRecordsIn) throws IOException {
+			Counter numRecordsIn) {
 
-		InputGate inputGate = InputGateUtil.createInputGate(inputGates);
-
-		CheckpointedInputGate barrierHandler = InputProcessorUtil.createCheckpointedInputGate(
-			checkpointedTask,
-			checkpointMode,
-			ioManager,
-			inputGate,
-			taskManagerConfig,
-			taskName);
-		this.input = new StreamTaskNetworkInput(barrierHandler, inputSerializer, ioManager, 0);
+		this.input = new StreamTaskNetworkInput(checkpointedInputGate, inputSerializer, ioManager, 0);
 
 		this.lock = checkNotNull(lock);
-
 		this.streamStatusMaintainer = checkNotNull(streamStatusMaintainer);
 		this.streamOperator = checkNotNull(streamOperator);
 
 		this.statusWatermarkValve = new StatusWatermarkValve(
-			inputGate.getNumberOfInputChannels(),
+			checkpointedInputGate.getNumberOfInputChannels(),
 			new ForwardingValveOutputHandler(streamOperator, lock));
 
-		this.watermarkGauge = watermarkGauge;
-		metrics.gauge("checkpointAlignmentTime", barrierHandler::getAlignmentDurationNanos);
-
+		this.watermarkGauge = checkNotNull(watermarkGauge);
 		this.operatorChain = checkNotNull(operatorChain);
 		this.numRecordsIn = checkNotNull(numRecordsIn);
 	}
--- a/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamOneInputProcessor.java
+++ b/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamOneInputProcessor.java
@@ -143,10 +143,7 @@ public final class StreamOneInputProcessor<IN> implements StreamInputProcessor {
 	public boolean processInput() throws Exception {
 		StreamElement recordOrMark = input.pollNextNullable();
 		if (recordOrMark != null) {
-			int channel = input.getLastChannel();
-			checkState(channel != StreamTaskInput.UNSPECIFIED);
-
-			processElement(recordOrMark, channel);
+			processElement(recordOrMark, input.getLastChannel());
 		}
 		checkFinished();
 
--- a/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamOneInputProcessor.java
+++ b/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamOneInputProcessor.java
@@ -19,17 +19,7 @@
 package org.apache.flink.streaming.runtime.io;
 
 import org.apache.flink.annotation.Internal;
-import org.apache.flink.api.common.typeutils.TypeSerializer;
-import org.apache.flink.metrics.Counter;
-import org.apache.flink.runtime.io.disk.iomanager.IOManager;
-import org.apache.flink.streaming.api.operators.OneInputStreamOperator;
-import org.apache.flink.streaming.api.watermark.Watermark;
-import org.apache.flink.streaming.runtime.metrics.WatermarkGauge;
-import org.apache.flink.streaming.runtime.streamrecord.StreamElement;
-import org.apache.flink.streaming.runtime.streamrecord.StreamRecord;
-import org.apache.flink.streaming.runtime.streamstatus.StatusWatermarkValve;
-import org.apache.flink.streaming.runtime.streamstatus.StreamStatus;
-import org.apache.flink.streaming.runtime.streamstatus.StreamStatusMaintainer;
+import org.apache.flink.streaming.runtime.io.PushingAsyncDataInput.DataOutput;
 import org.apache.flink.streaming.runtime.tasks.OperatorChain;
 
 import org.slf4j.Logger;
@@ -43,15 +33,6 @@ import static org.apache.flink.util.Preconditions.checkNotNull;
 /**
  * Input reader for {@link org.apache.flink.streaming.runtime.tasks.OneInputStreamTask}.
  *
- * <p>This internally uses a {@link StatusWatermarkValve} to keep track of {@link Watermark} and
- * {@link StreamStatus} events, and forwards them to event subscribers once the
- * {@link StatusWatermarkValve} determines the {@link Watermark} from all inputs has advanced, or
- * that a {@link StreamStatus} needs to be propagated downstream to denote a status change.
- *
- * <p>Forwarding elements, watermarks, or status status elements must be protected by synchronizing
- * on the given lock object. This ensures that we don't call methods on a
- * {@link OneInputStreamOperator} concurrently with the timer callback or other things.
- *
  * @param <IN> The type of the record that can be read with this record reader.
  */
 @Internal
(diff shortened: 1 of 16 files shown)


zhijiangW:
@pnowojski this pr is ready for review and I have not changed the `StreamTwoInputProcessor` yet because it would be removed soon by another PR.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/PushBasedAsyncDataInput.java
```diff
@@ -0,0 +1,60 @@
...
+import org.apache.flink.streaming.runtime.streamrecord.StreamRecord;
+import org.apache.flink.streaming.runtime.streamstatus.StreamStatus;
+
+/**
+ * The variant of {@link AsyncDataInput} that is defined for handling both network
+ * input and source input in a unified way via {@link #emitNext(DataOutput)} instead
+ * of returning {@code Optional.empty()} via {@link AsyncDataInput#pollNext()}.
+ */
+@Internal
+public interface PushBasedAsyncDataInput extends AvailabilityListener {
```
pnowojski: (loudly thinking) maybe rename `PushBasedAsyncDataInput` -> `PushingAsyncDataInput` and `AsyncDataInput` -> `PollingAsyncDataInput`?

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/PushBasedAsyncDataInput.java
```diff
@@ -0,0 +1,60 @@
...
+
+	/**
+	 * Pushes the next element to the output from current data input, and returns
+	 * the input status to indicate whether there are more available data in
+	 * current input.
+	 *
+	 * <p>This method should be non blocking.
+	 */
+	InputStatus emitNext(DataOutput output) throws Exception;
```
pnowojski: `PushBasedAsyncDataInput<T>`? and `emitNext(DataOutput<T>)`?

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamOneInputProcessor.java
```diff
@@ -188,26 +132,59 @@ public void close() throws IOException {
 		input.close();
 	}
 
-	private class ForwardingValveOutputHandler implements StatusWatermarkValve.ValveOutputHandler {
+	/**
+	 * The network data output implementation used for processing stream elements
+	 * from {@link StreamTaskNetworkInput} in one input processor.
+	 */
+	private class StreamTaskNetworkOutput implements DataOutput<IN> {
```
pnowojski: `private static class`?

pnowojski: Shouldn't the name be more generic and not connected with `Network`? Isn't the same class going to be used for the source reader?

Maybe if we are not sure about that, we can leave this as it is now and resolve it once we add `SourceReader` support?

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamTwoInputSelectableProcessor.java
```diff
@@ -381,12 +333,28 @@ private ForwardingValveOutputHandler(
...
 			this.inputIndex = inputIndex;
 		}
 
 		@Override
-		public void handleWatermark(Watermark watermark) throws Exception {
+		@SuppressWarnings("unchecked")
+		public void emitRecord(StreamRecord record) throws Exception {
+			synchronized (lock) {
+				numRecordsIn.inc();
+				if (inputIndex == 0) {
```
pnowojski: this might add some extra overhead compared to the previous code, as probably JIT won't be able to correctly handle this and branch prediction will not work (as far as I know, most/all optimisations are per class, not per class instance, so if there are two instances of the same class, where one has `inputIndex == 0` and other has `inputIndex == 1`, they will collide with one another).

Maybe it would be better to have this class abstract, with abstract method for `processElement`, which would be implemented concretely in `FirstStreamTaskNetworkOutput` and `SecondStreamTaskNetworkOutput` as
```
					operator.setKeyContextElement1(record);
					operator.processElement1(record);
```
and 
```
					operator.setKeyContextElement2(record);
					operator.processElement2(record);
```
respectively? We could leave watermarks/statuses/latency markers as they are, since they are not performance critical.

Normally I would suggest to relay on our benchmarks, but unfortunately our benchmarks for this selectable processor are very unstable :(

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/streamstatus/StatusWatermarkValveTest.java
```diff
@@ -403,22 +406,30 @@ public void testNoOutputWhenAllActiveChannelsAreUnaligned() throws Exception {
...
 		}
 
 		@Override
-		public void handleStreamStatus(StreamStatus streamStatus) {
+		public void emitStreamStatus(StreamStatus streamStatus) {
 			allOutputs.add(streamStatus);
 		}
 
+		@Override
+		public void emitRecord(StreamRecord record) {
```
pnowojski: `throw new UnsupportedOperationException`?

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/streamstatus/StatusWatermarkValveTest.java
```diff
@@ -403,22 +406,30 @@ public void testNoOutputWhenAllActiveChannelsAreUnaligned() throws Exception {
...
+		}
+
+		@Override
+		public void emitLatencyMarker(LatencyMarker latencyMarker) {
```
pnowojski: `throw new UnsupportedOperationException`?

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamOneInputProcessor.java
zhijiangW: It is hard to say. This implementation is for `OneInputStreamOperator` and I am not sure the new source reader could also couple with `OneInputStreamOperator`.

We can keep it as now and see how the formal `SourceReader` works.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamTwoInputSelectableProcessor.java
zhijiangW: Yes, I think you might be right for this performance concern. If we have two separate subclasses to implement abstract `emitRecords`, it could work.
Another possible option is to use the consumer function in my first version, then we could implement this function in `StreamTwoInputSelectableProcessor` and pass it into output constructor. This way seems a bit simple because it only relies on one class `StreamTaskNetworkOutput`, otherwise we need three classes (one abstract class and two subclasses). WDYT?

zhijiangW:
Thanks for the review @pnowojski !

I have submitted the fixup commit for addressing above comments. For the performance concern of `emitRecord` in two input processor, I took the consumer way temporarily. If you prefer the abstract class way, then I could refactor this part later. In addition I think the existing constructor of processor seems too long, so I prefer to refactor the constructor a bit later if you also agree.

I agree with your suggestion that before merging we firstly focus on a `SourceReader` example based on this PR.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamTwoInputSelectableProcessor.java
pnowojski: I think this is fine. Drawback is that stacktraces/debugging will be a little bit more difficult, but we can try this out.

zhijiangW: Yes, my only concern for this way is also for potential debugging issue.

zhijiangW:
Thanks for the further review. The failure compilation is caused by `StreamTwoInputProcessor` which I have adjusted before because it would be removed finally by another ongoing PR. I could wait for that PR merged and rebase on that. Also wait for some progress in `SourceReader`.

pnowojski:
FYI @zhijiangW, I have already merged the PR dropping non selectable TwoInput tasks/processors. The selectable counterparts were renamed by dropping `Selectable` infix (`StreamTwoInputSelectableProcessor` changed to `StreamTwoInputProcessor`)

zhijiangW:
Thanks for the notification and I already rebased the master. It is freiendly that no conflicts to be solved during rebasing. :)

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamTaskNetworkInput.java
```diff
@@ -111,7 +127,8 @@ public StreamElement pollNextNullable() throws Exception {
 				}
 
 				if (result.isFullRecord()) {
-					return deserializationDelegate.getInstance();
+					processElement(deserializationDelegate.getInstance(), output);
```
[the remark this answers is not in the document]
zhijiangW: It should never be null because of above condition of `result.isFullRecord()`.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamTwoInputProcessor.java
```diff
@@ -171,30 +193,23 @@ public boolean processInput() throws Exception {
...
 		}
 
-		return recordOrMark != null;
+		return status == InputStatus.MORE_AVAILABLE;
 	}
 
-	private void checkFinished(StreamTaskInput input, int inputIndex) throws Exception {
-		if (input.isFinished()) {
+	private void checkFinished(InputStatus status, int inputIndex) throws Exception {
+		if (status == InputStatus.END_OF_INPUT) {
```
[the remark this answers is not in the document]
zhijiangW: Actually I did that in first version to rename the `checkFinished` as `checkStatus` which also covers the 
```
if (status != InputStatus.MORE_AVAILABLE) {
	inputSelectionHandler.setUnavailableInput(readingInputIndex);
}
```

But in #9483 I integrated the above `if (status != InputStatus.MORE_AVAILABLE)` with other processes. So in this PR I kept the previous way to avoid repeated changes. :)

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/streamstatus/StatusWatermarkValve.java
```diff
@@ -29,24 +30,13 @@
...
-	 * to handle watermark and stream status outputs from the valve.
-	 */
-	public interface ValveOutputHandler {
-		void handleWatermark(Watermark watermark) throws Exception;
-
-		void handleStreamStatus(StreamStatus streamStatus) throws Exception;
-	}
-
-	private final ValveOutputHandler outputHandler;
+	private final DataOutput output;
```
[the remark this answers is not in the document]
zhijiangW: Thanks for this concern. I actually also hesitated to this issue before and passed the `output` in specific methods in `StatusWatermarkValve` instead of constructor in previous version.

I think there are two different possibilities here:

- Make `StatusWatermarkValve` completely coupled with `StreamTaskNetworkInput`, then we could even construct `StatusWatermarkValve` inside constructor of `StreamTaskNetworkInput`. This is a positive point. And we assume that the `output` is same for both `emitNext` and `StatusWatermarkValve` now. But if the `output` instances are different in future as you mentioned, then we need to pass another `output` into `StreamTaskNetworkInput` for `StatusWatermarkValve`. Especially it would cause many changes in `StatusWatermarkValve` because there are four internal methods involved in passing `output` argument in `StatusWatermarkValve`.

- Construct the different `output` instances in the upper layer of `StreamInputProcessor` for both `StreamTaskNetworkInput` and `StatusWatermarkValve` as the current implementation in this PR. In other words `StatusWatermarkValve` could be regarded as a `output` wrapper which is actually used in `StatusWatermarkValve`, and the wrapped `output` inside `StatusWatermarkValve` could be anything else different from the `output` in `emitNext`. If it changes we only need to construct a different `output` instance in `StreamInputProcessor` and pass it into `StatusWatermarkValve`, no need to touch four internal methods inside it.

I have no obvious preference for above two ways, because we could think of side effects for any way.

For the alternative of encapsulating `StatusWatermarkValve` behind `output`, I wonder it is not good to let `output` see any channel index concept, especially considering the source output future. And the current `StatusWatermarkValve` is a bit heavy-weight, so encapsulating it into `output` would complicate `output` implementation. In contrast, making `output` as an field of `StatusWatermarkValve` seems more light-weight.

zhijiangW:
Thanks for the review and good suggestions @1u0 !

I agree that `pull` seems more precise than `poll` as opposite to `push`, and I would submit a separate hotfix commit for renaming the previous class.

Could you help point out the code formatting issue inline codes then I could double confirm easily. :)

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamTwoInputProcessor.java
```diff
@@ -54,92 +50,98 @@
...
+
+		StatusWatermarkValve statusWatermarkValve1 = new StatusWatermarkValve(
+			checkpointedInputGates[0].getNumberOfInputChannels(), output1);
+		StatusWatermarkValve statusWatermarkValve2 = new StatusWatermarkValve(
+			checkpointedInputGates[1].getNumberOfInputChannels(), output2);
+
+		this.input1 = new StreamTaskNetworkInput<>(
+			checkpointedInputGates[0], inputSerializer1, ioManager, statusWatermarkValve1, 0);
+		this.input2 = new StreamTaskNetworkInput<>(
+			checkpointedInputGates[1], inputSerializer2, ioManager, statusWatermarkValve2, 1);
```
1u0: For code formatting: this and above code, related for arguments on new lines (see: https://flink.apache.org/contributing/code-style-and-quality-formatting.html#breaking-the-lines-of-too-long-statements).

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/OneInputStreamTask.java
```diff
@@ -68,31 +86,115 @@ public OneInputStreamTask(
...
+	}
+
+	private StreamTaskInput<IN> createTaskInput(CheckpointedInputGate inputGate, DataOutput<IN> output) {
+		TypeSerializer<IN> inSerializer = configuration.getTypeSerializerIn1(getUserCodeClassLoader());
+
+		StatusWatermarkValve statusWatermarkValve = new StatusWatermarkValve(
+			inputGate.getNumberOfInputChannels(), output);
+
+		return new StreamTaskNetworkInput<>(
+			inputGate, inSerializer, getEnvironment().getIOManager(), statusWatermarkValve, 0);
```
1u0: Same note for code formatting (and a few lines above).

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamTwoInputProcessor.java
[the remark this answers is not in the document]
1u0: Yes, I've read your other PR later.
There, you are moving the status check before reading the follow up element. Meaning, that the `processInput()` doesn't return actual current status, but rather the status before read.
But this is not a strict requirement/contract, afaik.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/streamstatus/StatusWatermarkValve.java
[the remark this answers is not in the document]
1u0: By encapsulating `StatusWatermarkValve` behind the `output`, I don't mean inlining the implementation of `StatusWatermarkValve` in the `output` class, but rather making it a private field (in similar way how you did it now, by moving the valve to the source).

Abstraction wise, I'm not sure (have no strong opinion) whether the valve should belong to either the source or output class.
But current situation, that one of the fields in the source have reference to the output and the contact of `PushingAsyncDataInput` that the output is passed via main `emitNext()` method is poor by design, in my opinion.

> there are four internal methods involved in passing output argument in `StatusWatermarkValve`.

That is not a lot of change, imo.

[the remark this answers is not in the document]
zhijiangW: It can not directly work well to make `StatusWatermarkValve` as a private field of `output`. 
The overall process is `StreamTaskNetworkInput#processElement` -> `if (recordOrMark.isWatermark())` -> `output.emitWatermark` -> StatusWatermarkValve#inputWatermark`.

The key problem is that the interface of `output.emitWatermark` can not expose any channel index info, but `StatusWatermarkValve` has to know the current channel index to judge the internal logics. So we make the `StatusWatermarkValve` in the upper layer of `output`. In other words, the `StatusWatermarkValve` can wrap the `output`, but `output` can not wrap the `StatusWatermarkValve`.

As for passing the `output` in the constructor of `StatusWatermarkValve` vs passing it in four internal methods of `StatusWatermarkValve`. I still prefer to the current way for two reasons:

- Once `StatusWatermarkValve` needs a different output future, we only need to change one place, otherwise we need to touch four internal methods.

- From design aspect, I think it is better to reduce the unnecessary dependency between `StatusWatermarkValve` and `StreamTaskNetworkInput`. Currently we have to rely on the channel index from `StreamTaskNetworkInput`. But if we regard the `output` as a more general concept which might not be consistent with the `output` in `emitNext`, then we can pass it into the constructor in free from outside. Just like the previous `ForwardingValveOutputHandler` which was also passed into the constructor of `StatusWatermarkValve` from outside.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamTwoInputProcessor.java
zhijiangW: Thanks. In my previous memories, it was both ok for arguments either in the same line or in separate line. Now the new formatting seems only accept the separate argument line if breaking. I should catch up with it and fix based on that. :)

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamTwoInputProcessor.java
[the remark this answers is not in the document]
zhijiangW: The motivation of moving `inputSelectionHandler.setUnavailableInput` before reading is only for consistent logics for better understanding, not related to whether to return actual current status.

In the past, `inputSelectionHandler.setUnavailableInput` was done after reading, and `inputSelectionHandler.setAvailableInput` was done before reading. From my understanding, no matter with `setUnavailableInput` or `setAvailableInput` for `inputSelectionHandler` , it is only used for selecting the next expected input index before reading, so we should integrate them together either before or after reading. It is hard to understand or trace the logic if spreading it anywhere.

zhijiangW:
Yes, it is better to merge this PR soon if no other concerns. Because there are additional two PRs #9483 #9646 relying on it,  and it actually wastes some efforts to append this PR every time.

pnowojski:
I'm running benchmark request against this code and will merge it at the beginning of next week if everything goes well.

pnowojski:
Benchmark request shown some potential regressions in pure network benchmarks, and only in some of them, not all, so that might be some random fluke. Especially that this PR barely touches anything used in those pure network benchmarks.

Let's merge it either way and observe the long term trends. I would wait couple of days before merging ( https://github.com/apache/flink/pull/9483 )

PR #9480, opened 2019-08-19
[Flink-13752] Only references necessary variables when bookkeeping result partitions on TM

gaoyunhaii:

This pull request fixes the problem that the TaskDeploymentDescriptor is referenced by an anonymous function that is going to be executed when task terminated. This causeTaskDeploymentDescriptor cannot be recycled by GC. Since TaskDeploymentDescriptor  has a reference to its serialized value and for some tasks this array might contain dozens of megabytes, unable to recycle TaskDeploymentDescriptor may have large impact on GC and further on the performance.

Since the anonymous function indeed only want to references the JobId instead of the whole TaskDeploymentDescriptor, it can thus only reference the JobId instead.

## Brief change log

a95596e548b62cfc71177f8c2da0be540566f976 acquires the JobId outside the anonymous function and only reference to the JobId in the anonymous function.

## Verifying this change
I think there might be no need to add a UT for this fix since it is not changing the interface of some components. If we want to add a UT, it might can only acquire the anonymous function and check if it has a field referencing TaskDeploymentDescriptor. However, the behavior to test is mainly implement related logic.

Currently it has been verified by manual test with an external job which triggers this problem.

  - The serializers: **no**
  - Anything that affects deployment or recovery: JobManager Checkpointing, Yarn/Mesos, ZooKeeper: **no**

## Documentation

### DIFF
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/taskexecutor/TaskExecutor.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/taskexecutor/TaskExecutor.java
@@ -618,7 +618,10 @@ public class TaskExecutor extends RpcEndpoint implements TaskExecutorGateway {
 				task.startTaskThread();
 				taskCompletionTracker.trackTaskCompletion(task);
 
-				setupResultPartitionBookkeeping(tdd, task.getTerminationFuture());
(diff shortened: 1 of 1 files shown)


gaoyunhaii:
Very thanks @tillrohrmann for the review!

PR #9483, opened 2019-08-19
[FLINK-13767][task] Refactor StreamInputProcessor#processInput based on InputStatus

zhijiangW:

`StreamInputProcessor#processInput` could return `InputStatus` instead of current boolean value to keep consistent with `PushingAsyncDataInput#emitNext`.
    
For the implementation of `StreamTwoInputProcessor#processInput`, we could maintain and judge the two input status together with the next selected input index to determine the final precise status. To do so we could avoid invalid `processInput` call except for the first call.
    
In addition, `AvailabilityProvider#isFinished` has the duplicated semantic with `InputStatus#END_OF_INPUT` for `PushingAsyncDataInput`, and it is only meaningful for `PullingAsyncDataInput` now. So we migrate the `#isFinished` method from `AvailabilityProvider` to `PullingAsyncDataInput`.

Note: This PR is based on #9478 , then only review the last four commits for this PR change.

## Brief change log

  - *Migrate `isFinished` method from `AvailabilityProvider` to `PullingAsyncDataInput`.*
  - *Make `StreamInputProcessor#processInput` return `InputStatus` instead.*
  - *Refactor the relevant processes in `StreamOneInputProcessor`.*
  - *Refactor the relevant processes in `StreamTwoInputProcessor`.*

## Verifying this change

## Documentation

### DIFF
--- a/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamTwoInputProcessor.java
+++ b/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamTwoInputProcessor.java
@@ -184,10 +184,6 @@ public final class StreamTwoInputProcessor<IN1, IN2> implements StreamInputProce
 		}
 		checkFinished(status, lastReadInputIndex);
 
-		if (status != InputStatus.MORE_AVAILABLE) {
-			inputSelectionHandler.setUnavailableInput(readingInputIndex);
-		}
-
 		return status == InputStatus.MORE_AVAILABLE;
 	}
 
@@ -253,11 +249,15 @@ public final class StreamTwoInputProcessor<IN1, IN2> implements StreamInputProce
 	}
 
 	private void updateAvailability() {
-		if (!input1.isFinished() && input1.isAvailable() == AVAILABLE) {
-			inputSelectionHandler.setAvailableInput(input1.getInputIndex());
-		}
-		if (!input2.isFinished() && input2.isAvailable() == AVAILABLE) {
-			inputSelectionHandler.setAvailableInput(input2.getInputIndex());
+		updateAvailability(input1);
+		updateAvailability(input2);
+	}
+
+	private void updateAvailability(StreamTaskInput input) {
+		if (!input.isFinished() && input.isAvailable() == AVAILABLE) {
+			inputSelectionHandler.setAvailableInput(input.getInputIndex());
+		} else {
+			inputSelectionHandler.setUnavailableInput(input.getInputIndex());
 		}
 	}
 
--- a/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamTwoInputProcessor.java
+++ b/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamTwoInputProcessor.java
@@ -279,7 +279,7 @@ public final class StreamTwoInputProcessor<IN1, IN2> implements StreamInputProce
 
 	private CompletableFuture<?> isAnyInputAvailable() {
 		if (input1.isFinished()) {
-			return input2.isFinished() ? AVAILABLE : input2.isAvailable();
+			return input2.isAvailable();
 		}
 
 		if (input2.isFinished()) {
--- a/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamTwoInputProcessor.java
+++ b/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamTwoInputProcessor.java
@@ -62,6 +62,10 @@ public final class StreamTwoInputProcessor<IN1, IN2> implements StreamInputProce
 	private final DataOutput<IN1> output1;
 	private final DataOutput<IN2> output2;
 
+	/** Input status to keep track for determining whether the input is finished or not. */
+	private InputStatus firstInputStatus = InputStatus.MORE_AVAILABLE;
+	private InputStatus secondInputStatus = InputStatus.MORE_AVAILABLE;
+
 	/**
 	 * Stream status for the two inputs. We need to keep track for determining when
 	 * to forward stream status changes downstream.
@@ -148,11 +152,6 @@ public final class StreamTwoInputProcessor<IN1, IN2> implements StreamInputProce
 	}
 
 	@Override
-	public boolean isFinished() {
-		return input1.isFinished() && input2.isFinished();
-	}
-
-	@Override
 	public CompletableFuture<?> isAvailable() {
 		if (inputSelectionHandler.areAllInputsSelected()) {
 			return isAnyInputAvailable();
@@ -163,28 +162,42 @@ public final class StreamTwoInputProcessor<IN1, IN2> implements StreamInputProce
 	}
 
 	@Override
-	public boolean processInput() throws Exception {
-		if (!isPrepared) {
+	public InputStatus processInput() throws Exception {
+		int readingInputIndex;
+		if (isPrepared) {
+			readingInputIndex = selectNextReadingInputIndex();
+			assert readingInputIndex != -1;
+		} else {
 			// the preparations here are not placed in the constructor because all work in it
 			// must be executed after all operators are opened.
-			prepareForProcessing();
+			readingInputIndex = selectFirstReadingInputIndex();
+			if (readingInputIndex == -1) {
+				return InputStatus.NOTHING_AVAILABLE;
+			}
 		}
 
-		int readingInputIndex = selectNextReadingInputIndex();
-		if (readingInputIndex == -1) {
-			return false;
-		}
 		lastReadInputIndex = readingInputIndex;
 
-		InputStatus status;
 		if (readingInputIndex == 0) {
-			status = input1.emitNext(output1);
+			firstInputStatus = input1.emitNext(output1);
+			checkFinished(firstInputStatus, lastReadInputIndex);
 		} else {
-			status = input2.emitNext(output2);
+			secondInputStatus = input2.emitNext(output2);
+			checkFinished(secondInputStatus, lastReadInputIndex);
 		}
-		checkFinished(status, lastReadInputIndex);
 
-		return status == InputStatus.MORE_AVAILABLE;
+		return getInputStatus();
+	}
+
+	private int selectFirstReadingInputIndex() throws IOException {
+		// Note: the first call to nextSelection () on the operator must be made after this operator
+		// is opened to ensure that any changes about the input selection in its open()
+		// method take effect.
+		inputSelectionHandler.nextSelection();
+
+		isPrepared = true;
+
+		return selectNextReadingInputIndex();
 	}
 
 	private void checkFinished(InputStatus status, int inputIndex) throws Exception {
@@ -194,6 +207,24 @@ public final class StreamTwoInputProcessor<IN1, IN2> implements StreamInputProce
 				inputSelectionHandler.nextSelection();
 			}
 		}
+	}
+
+	private InputStatus getInputStatus() {
+		if (firstInputStatus == InputStatus.END_OF_INPUT && secondInputStatus == InputStatus.END_OF_INPUT) {
+			return InputStatus.END_OF_INPUT;
+		}
+
+		if (inputSelectionHandler.areAllInputsSelected()) {
+			if (firstInputStatus == InputStatus.MORE_AVAILABLE || secondInputStatus == InputStatus.MORE_AVAILABLE) {
+				return InputStatus.MORE_AVAILABLE;
+			} else {
+				return InputStatus.NOTHING_AVAILABLE;
+			}
+		}
+
+		InputStatus selectedStatus = inputSelectionHandler.isFirstInputSelected() ? firstInputStatus : secondInputStatus;
+		InputStatus otherStatus = inputSelectionHandler.isFirstInputSelected() ? secondInputStatus : firstInputStatus;
+		return selectedStatus == InputStatus.END_OF_INPUT ? otherStatus : selectedStatus;
 	}
 
 	@Override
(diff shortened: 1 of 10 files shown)


zhijiangW:
My previous thought for this PR is migrating `isFinished` method from `AvailabilityListener` to `PollingAsyncDataInput`, and also refactoring the `StreamInputProcessor#processInput` to return `InputStatus` instead of boolean value as now.

But after thinking through it again, now I find that it would bring additional overheads for the implementation of `StreamTwoInputSelectableProcessor`.

-  If we do not maintain the `isFinished` flag in `StreamTaskNetworkInput`, we have to introduce another two variables for maintaining the first and second input status in `StreamTwoInputSelectableProcessor`, because the finished state needs to be checked in other processes before calling `emitNext`. 

- Furthermore, only if two inputs are both finished then we can return `InputStatus.END_OF_INPUT` to the `StreamTask` while calling `processInput`. So we need add additional check process in every call `StreamTwoInputSelectableProcessor#processInput`.

Therefore I do not think it is very necessary to do so now. WDYT? @pnowojski

zhijiangW:
Anyway, I already updated the codes for better understanding my above concerns.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamTwoInputSelectableProcessor.java
```diff
@@ -194,23 +193,28 @@ public boolean processInput() throws Exception {
...
+			secondInputStatus = status;
 		}
 		checkFinished(status, lastReadInputIndex);
 
 		if (status == InputStatus.NOTHING_AVAILABLE) {
 			inputSelectionHandler.setUnavailableInput(readingInputIndex);
 		}
 
-		return status == InputStatus.MORE_AVAILABLE;
+		if (status == InputStatus.END_OF_INPUT && secondInputStatus != InputStatus.END_OF_INPUT) {
```
pnowojski: I think this is not quite correct. What if `status == NOTHING_AVAILABLE` and for example `secondInputStatus == MORE_AVAILABLE`? Result should be `MORE_AVAILABLE` as long as input selection is correct. 

Similar, what about `status == END_OF_INPUT` and `secondInputStatus == MORE_AVAILABLE`? Currently that would return `NOTHING_AVAILABLE`?

Am I missing something, or is this wrong? If this is a bug, please also add a test coverage for those (and potentially other corner cases).

zhijiangW: I think we keep the same behavior as before. If `status == NOTHING_AVAILABLE` and `secondInputStatus == MORE_AVAILABLE`, it would return `status` and in `StreamTask#performDefaultAction` it would call `inputProcessor.isAvailable().thenRun()`. 

Before refactoring, if `status == NOTHING_AVAILABLE` which corresponds to the case of returning null `recordOrMark`, then in `StreamTask#performDefaultAction` it would still call `inputProcessor.isAvailable().thenRun()`. 

Also in previous behavior only two inputs are both finished, then we can return `status==END_OF_INPUT`. If only current `status== END_OF_INPUT` then the returned `recordOrMark` should be null, so in StreamTask#performDefaultAction` it would also enter the path of `inputProcessor.isAvailable().thenRun()`. In order to keep this same behavior we need to return `NOTHING_AVAILABLE` here to enter the same path in `StreamTask`.

So this part was my main concern before, whether it is worth doing this refactor here. Because we add this condition `if (status == InputStatus.END_OF_INPUT && secondInputStatus != InputStatus.END_OF_INPUT)` to always check in every call for keeping the previous behavior.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamTwoInputSelectableProcessor.java
```diff
@@ -194,23 +193,28 @@ public boolean processInput() throws Exception {
 
 		int readingInputIndex = selectNextReadingInputIndex();
 		if (readingInputIndex == -1) {
-			return false;
+			return InputStatus.NOTHING_AVAILABLE;
```
pnowojski: I think this might be not right? Shouldn't we return `END_OF_INPUT` in some cases here?

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamTwoInputSelectableProcessor.java
pnowojski: Hmmm, maybe we should clean this up?

I think previously it was not that bad, as it was more of an optimisation, to avoid checking `isAvailable()` when hot looping (touching `isAvailable()` means touching `volatile` field inside `CompletableFuture`). 

Now that we want to actually express `END_OF_INPUT` as well, we might need to be more precise here.

Do you think it would be difficult to fix this logic here?

zhijiangW: Let me think how to make this logic more clearly now.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamTwoInputSelectableProcessor.java
zhijiangW: I think it does not exist the case of return `END_OF_INPUT` here. Because if both inputs are finished, then `StreamTask` would call `context.allActionsCompleted()`, that means it would never call `inputProcessor.processInput()` any more. In other words, if `processInput` is called here that should mean at-least one input is not finished yet.

Actually I also felt this logic is difficult to understand. I guess the `readingInputIndex == -1` should only happen in the first call of `processInput`. In the following calls the `processInput` is triggered by available input, then the selected input index should never be -1. Is it right?

zhijiangW: After double checking the logic in `MailBoxProcessor#runMailboxLoop`, it has the optimization logic of warming hot path via 
```
if (!mailbox.hasMail()) {
	return true;
}
```
Then it would cause `selectNextReadingInputIndex` return -1 here, not only for the first call `processInput`. But it still does not exist the `END_OF_INPUT` case here.

Another concern is that are there any experiments to verify this hot path optimization has benefits in practice? I wonder it might waste some cpu time and also a bit delay the valid processing after available notification.

zhijiangW:
@pnowojski I have updated the codes for above concerns.

zhijiangW:
@flinkbot run travis

zhijiangW:
@flinkbot run travis

zhijiangW:
@flinkbot run travis

flink-runtime/src/main/java/org/apache/flink/runtime/io/AvailabilityListener.java
```diff
@@ -55,7 +50,7 @@
 	 * @return a future that is completed if there are more records available. If there are more
 	 * records available immediately, {@link #AVAILABLE} should be returned. Previously returned
 	 * not completed futures should become completed once there is more input available or if
-	 * the input {@link #isFinished()}.
+	 * the input is finished.
```
AHeise: Should we also have a constant for `FINISHED`? As is the `AvailabilityListener` depends on its implementations to provide some `isFinished()`.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamTwoInputProcessor.java
```diff
@@ -121,18 +115,46 @@ public StreamTwoInputProcessor(
...
 
-		this.statusWatermarkValve1 = new StatusWatermarkValve(
-			unionedInputGate1.getNumberOfInputChannels(),
-			new ForwardingValveOutputHandler(streamOperator, lock, streamStatusMaintainer, input1WatermarkGauge, 0));
-		this.statusWatermarkValve2 = new StatusWatermarkValve(
-			unionedInputGate2.getNumberOfInputChannels(),
-			new ForwardingValveOutputHandler(streamOperator, lock, streamStatusMaintainer, input2WatermarkGauge, 1));
+		this.output1 = new StreamTaskNetworkOutput<>(
+			streamOperator,
+			(StreamRecord<IN1> record) -> {
```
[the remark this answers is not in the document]
zhijiangW: Agree, the constructor seems too long now. In order to avoid this, I ever constructed the relevant arguments including `Output` via separate method in `StreamTwoInputTask`, and then pass these final arguments to the constructor of `StreamTwoInputProcessor` directly in my previous version. I would consider refactor this part later.

flink-runtime/src/main/java/org/apache/flink/runtime/io/AvailabilityListener.java
```diff
@@ -33,11 +33,6 @@
 	 */
 	CompletableFuture<?> AVAILABLE = CompletableFuture.completedFuture(null);
 
-	/**
```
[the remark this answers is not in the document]
zhijiangW: Yes, I agree with the current class naming seems a bit misnomer. It could be refactored in a separate hotfix commit.

flink-runtime/src/main/java/org/apache/flink/runtime/io/AvailabilityListener.java
zhijiangW: Not quite sure of it.  `isFinished()` is more proper to be provided by input interface to indicate whether this input ends or not. As for `AvailabilityListener`, we just explain that which cases would cause future complete. Or I did not get your key point?

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamTwoInputProcessor.java
```diff
@@ -158,43 +173,34 @@ public boolean isFinished() {
...
-		}
+	public InputStatus processInput() throws Exception {
+		// the preparations here are not placed in the constructor because all work in it
+		// must be executed after all operators are opened.
+		prepareForProcessing();
 
 		int readingInputIndex = selectNextReadingInputIndex();
 		if (readingInputIndex == -1) {
-			return false;
+			return InputStatus.NOTHING_AVAILABLE;
```
1u0: Should this clause actually check `firstInputStatus` and `secondInputStatus` that they are both `InputStatus.END_OF_INPUT` and return `InputStatus.END_OF_INPUT` in that case?

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamTwoInputProcessor.java
```diff
@@ -158,43 +173,34 @@ public boolean isFinished() {
...
-		if (recordOrMark == null) {
-			inputSelectionHandler.setUnavailableInput(readingInputIndex);
+			status = input2.emitNext(output2);
+			secondInputStatus = status;
+			anotherStatus = firstInputStatus;
 		}
+		checkFinished(status, lastReadInputIndex);
 
-		return recordOrMark != null;
+		return status == InputStatus.END_OF_INPUT ? anotherStatus : status;
```
1u0: I think more proper way (and as "performance optimization") you can rewrite the return value as:
```java
if (firstInputStatus == InputStatus.MORE_AVAILABLE || secondInputStatus == InputStatus.MORE_AVAILABLE) {
    return InputStatus.MORE_AVAILABLE;
} else if (firstInputStatus == InputStatus.END_OF_INPUT && secondInputStatus == InputStatus.END_OF_INPUT) {
    return InputStatus.END_OF_INPUT;
} else {
    return InputStatus.NOTHING_AVAILABLE;
}

Comment (zhijiangW) [reply flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamTwoInputProcessor.java]: Not quite confirm the performance optimization issue.

From the precise aspect, my previous way was trying to maintain the old behavior which only judges the current input status except for the end status. I think all these ways would bring bad cases.

E.g. if current status is equal to `MORE_AVAILABLE` but the next selection is the other input which is actually `NOTHING_AVAILABLE`, then after calling `processInput` again it would actually return -1 while `selectNextReadingInputIndex()` to waste processing time.

Before introducing `InputStatus` it could also return precise status if maintaining both last input states as the way I did (`firstInputStatus`, `secondInputStatus`) in this PR. If we want to do that we need also consider three factors together (`firstInputStatus`, `secondInputStatus` and next selected input) to determine the final precise status for `StreamTask`.

If doing so, the above `selectNextReadingInputIndex` could only return -1 for the first calling `processInput`, and we could also refactor that part together.

File: flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamTwoInputProcessor.java (hunk 6 above)
Comment (zhijiangW) [reply flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamTwoInputProcessor.java]: Actually it would not happen that case. If both inputs were ended before, then the `StreamTask` would call `context.allActionsCompleted()` in last calling `processInput`, so this calling `processInput` would never happen. 

In other words, if the `processInput` is called, that means at-least one of the inputs is not finished yet.

File: flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/StreamTwoInputProcessor.java [hunk 8]
```diff
@@ -184,10 +184,6 @@ public boolean processInput() throws Exception {
 		}
 		checkFinished(status, lastReadInputIndex);
 
-		if (status != InputStatus.MORE_AVAILABLE) {
```
pnowojski: Some side effect of removing this code is that now `inputSelectionHandler` will be updated not immediately during this `#processInput(...)` invocation, but only during the next one, right? Could this cause any (performance?) issues?

zhijiangW: Yes, I refactored this part in a separate hotfix commit for integrating the logics of `inputSelectionHandler#setAvailableIndex` and `setUnavailableIndex` together before `emitNext` call. In the past these two logics are scattered before and after `emitNext` separately.

I do not think it would impact the performance, and at-least in my benchmark results it has not caused.  I am curious that are there any purpose to update the selection immediately before?

pnowojski: > are there any purpose to update the selection immediately before?

I'm not sure if there is any difference now, with the updated return status code. Previously it was returning boolean "more available" for hot looping. Without up to date input availability, it could attempt to read from incorrect (empty) input and that would trigger a `volatile` `isAvailable()` access.

zhijiangW: Let me summary the changes a bit. There are mainly three changes:

1. Migrate `inputSelectionHandler.setUnavailableInput` to integrate with `selectNextReadingInputIndex#updateAvailability`. I think it has no behavior different with before. Just to put similar logics together for better tracing.

2. Return status instead of boolean for `StreamInputProcessor#processInput`. It is for replacing the `isFinished` method and keeping the consistent form with `emitNext` return type. It also has no behavior change, and only `StreamTask#processInput` adds two additional conditions check instead of previous `while` way. But we can also use `while` way or add the hot path to avoid performance concern.
```
if (status == InputStatus.MORE_AVAILABLE) {
     return;
}
```

3. The value returned by `StreamTwoInputProcessor#processInput` is different from before which might cause different behaviors. In the past it might return invalid status after current call, and then in the next call it might return -1 while `selectNextReadingInputIndex()` to trigger `isAvailable()` access. But now it would always return valid status by judging the next selected index with its status. So if it is not available for next selected input, it would trigger `isAvailable()` access in current call directly, no need to trigger it in next call. From this point, my previous thought was that it might save some unnecessary cost from mailbox schedule aspect.

Anyway, it is hard to verify the above points because of unstable benchmark. So we can only analyze the possible impacts in theory.

PR #9484, opened 2019-08-19
[FLINK-13774][table] Expression DSL use should resolve Expressions in blink

JingsongLi:

1.DeclarativeAggregate should resolve expressions and ResolvedAggInputReferences should be ResolvedExpression.
2.TimestampExtractor should use resolved expressions.
3.SourceFilterPushDown should use resolved expressions.

## Verifying this change

  - The serializers: no

## Documentation


flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/plan/rules/logical/PushFilterIntoTableSourceScanRule.scala
```diff
@@ -112,8 +112,8 @@ class PushFilterIntoTableSourceScanRule extends RelOptRule(
       call.transformTo(newScan)
     } else {
       relBuilder.push(scan)
-      val converter = new RexNodeConverter(relBuilder)
```
JingsongLi: One question is whether `UnsolvedExpression` or `ResolvedExpression` is passed to `FilterableTableSource`.

JingsongLi:
ping @dawidwys  @wuchong  to review~

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/expressions/DeclarativeAggregateExpressionConverter.java
```diff
@@ -0,0 +1,55 @@
...
+import java.util.List;
+import java.util.Optional;
+
+/**
+ * Convert expressions of {@link DeclarativeAggregateFunction} to calcite {@link RexNode}, inputs should only contains
+ * call unresolved expression, other expression should have been resolved.
+ */
+public class DeclarativeAggregateExpressionConverter {
+
+	public static RexNode toRexNode(
```
wuchong: It mixes resolve logic and convert logic in one method, but the method name is only about converting.

Can we split them? We already have a resolve logic in every agg code gen, can we make the `ResolveReference` as a `ResolverRule` and resolve with other rules together before passing the expressions to `RexNodeConverter`?

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/expressions/logic.scala
```diff
@@ -76,7 +78,9 @@ case class If(
 
   override private[flink] def validateInput(): ValidationResult = {
     if (condition.resultType == BasicTypeInfo.BOOLEAN_TYPE_INFO &&
-        ifTrue.resultType == ifFalse.resultType) {
+        ((isDecimal(fromTypeInfoToLogicalType(ifTrue.resultType)) &&
```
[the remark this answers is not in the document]
JingsongLi: I'll add test

JingsongLi:
@wuchong Thanks for your review, comments addressed.
Due to the large changes, rewrite 12a1965

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/expressions/ResolvedAggLocalReference.java
```diff
@@ -20,19 +20,23 @@
...
+
 /**
  * Special reference which represent a local filed, such as aggregate buffers or constants.
  * We are stored as class members, so the field can be referenced directly.
  * We should use an unique name to locate the field.
  *
  * <p>See {@link org.apache.flink.table.planner.codegen.ExprCodeGenerator#visitLocalRef}.
  */
-public class ResolvedAggLocalReference implements Expression {
+public class ResolvedAggLocalReference implements ResolvedExpression {
```
twalthr: I would like to keep the number of expression low. I'm wondering why this is needed. Can't we also use `FieldReferenceExpression` for these kinds of references? Having `nullTerm` and `fieldTerm` sounds very internal to me. Also for `ResolvedDistinctKeyReference`, why not expressing this like `'field.distinct`? I haven't looked deep into the code, people that are more familiar with the code base should decide.

dawidwys: I mostly agree with @twalthr. I think though rather than `FieldReferenceExpression` we should use `LocalReferenceExpression`.

JingsongLi:
Hi @twalthr @dawidwys , I will try to do two things as your suggestion:
1.Use `FieldReferenceExpression` to cover `ResolvedAggInputReference`.
2.Use `LocalReferenceExpression` to cover `ResolvedAggLocalReference` and `ResolvedDistinctKeyReference`.

Any thoughts about `DeclarativeAggregateFunction`? I have some tangle about using `UnresolvedReferenceExpression` or `LocalReferenceExpression` to represent `aggBufferAttributes`.

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/expressions/ResolvedAggLocalReference.java
JingsongLi: `LocalReferenceExpression` is never be null because it is used for a group window in window aggregation.
But if it is used for aggregate buffer, it should be nullable, so we need a `nullTerm`. If not, how do we identify its null attribute?

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/calcite/FlinkLocalRef.scala
```diff
@@ -23,18 +23,16 @@ import org.apache.calcite.rel.`type`.RelDataType
...
   */
-case class RexAggLocalVariable(
+case class LocalReference(
     fieldTerm: String,
-    nullTerm: String,
-    dataType: RelDataType,
-    internalType: LogicalType)
-  extends RexLocalRef(0, dataType)
+    dataType: RelDataType)
+    extends RexLocalRef(0, dataType)
```
twalthr: how about using a new RexNode instead of violating the contract of `RexLocalRef` with always index 0?

JingsongLi:
Hi @twalthr @dawidwys , I did two things to delete blink extended expressions:
1.Use `LocalReferenceExpression` to cover `ResolvedAggLocalReference`.
2.Use `RexNodeExpression(RexInputRef)` to cover `ResolvedAggInputReference`. Because now aggregate input use index to find this reference. And now Expression->InputRef do not support use index to InputRef because of the case of grouping keys (The groupings order will be changed by calcite `RelBuilder.aggregate`). So we can not use `FieldReferenceExpression` now.
3.Use `RexNodeExpression` to cover `ResolvedDistinctKeyReference`.  Now distinct keys dependents on special logical to generate nullTerm and it is a local variable instead of class member.

I think we can use `RexNodeExpression` to wrap RexNode here, because these place are internal codeGen. It did not introduce a new Expression class.

flink-table/flink-table-planner-blink/src/main/scala/org/apache/flink/table/planner/calcite/FlinkLocalRef.scala
JingsongLi: Good point! I think we can use new `RexVariable` instead of `RexLocalRef`.

PR #9485, opened 2019-08-19
[FLINK-13775][table-planner-blink] Refactor ExpressionConverter(RexNodeConverter) in blink

JingsongLi:

For subsequent development, refactor ExpressionConverter in the form of Rules.
Now the RexNodeConverter has 1000 lines, but these logic can do some better classification.
Later we will introduce more converter mode, refactoring can let us better maintain and modify.
This is based on #9484 

## Verifying this change

  - The serializers: no

## Documentation

### DIFF
--- a/flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/expressions/converter/CallExpressionConvertRule.java
+++ b/flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/expressions/converter/CallExpressionConvertRule.java
@@ -25,9 +25,7 @@ import org.apache.flink.table.planner.calcite.FlinkTypeFactory;
 import org.apache.calcite.rex.RexNode;
 import org.apache.calcite.tools.RelBuilder;
 
-import java.util.List;
 import java.util.Optional;
-import java.util.stream.Collectors;
 
 /**
  * Rule to convert {@link CallExpression}.
@@ -50,10 +48,6 @@ public interface CallExpressionConvertRule {
 		 * Convert expression to RexNode, used by children conversion.
 		 */
 		RexNode toRexNode(Expression expr);
-
-		default List<RexNode> toRexNodes(List<Expression> expr) {
-			return expr.stream().map(this::toRexNode).collect(Collectors.toList());
-		}
 
 		RelBuilder getRelBuilder();
 
--- a/flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/expressions/converter/CallExpressionConvertRule.java
+++ b/flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/expressions/converter/CallExpressionConvertRule.java
@@ -1,0 +1,62 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.table.planner.expressions.converter;
+
+import org.apache.flink.table.expressions.CallExpression;
+import org.apache.flink.table.expressions.Expression;
+import org.apache.flink.table.planner.calcite.FlinkTypeFactory;
+
+import org.apache.calcite.rex.RexNode;
+import org.apache.calcite.tools.RelBuilder;
+
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
+
+/**
+ * Rule to convert {@link CallExpression}.
+ */
+public interface CallExpressionConvertRule {
+
+	/**
+	 * Convert call expression with context to RexNode.
+	 *
+	 * @return Success return RexNode of {@link Optional#of}, Fail return {@link Optional#empty()}.
+	 */
+	Optional<RexNode> convert(CallExpression call, ConvertContext context);
+
+	/**
+	 * Context of {@link CallExpressionConvertRule}.
+	 */
+	interface ConvertContext {
+
+		/**
+		 * Convert expression to RexNode, used by children conversion.
+		 */
+		RexNode toRexNode(Expression expr);
+
+		default List<RexNode> toRexNodes(List<Expression> expr) {
+			return expr.stream().map(this::toRexNode).collect(Collectors.toList());
+		}
+
+		RelBuilder getRelBuilder();
+
+		FlinkTypeFactory getTypeFactory();
+	}
+}
--- a/flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/expressions/converter/CustomizedConvertRule.java
+++ b/flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/expressions/converter/CustomizedConvertRule.java
@@ -37,7 +37,6 @@ import org.apache.flink.table.types.DataType;
 import org.apache.flink.table.types.logical.ArrayType;
 import org.apache.flink.table.types.logical.LogicalType;
 import org.apache.flink.table.types.logical.RowType;
-import org.apache.flink.util.Preconditions;
 
 import com.google.common.collect.ImmutableList;
 import org.apache.calcite.rel.type.RelDataType;
@@ -177,7 +176,7 @@ public class CustomizedConvertRule implements CallExpressionConvertRule {
 	private static RexNode convertMap(CallExpression call, ConvertContext context) {
 		// TODO get type from CallExpression directly until introduce type inference on Expression
 		List<Expression> children = call.getChildren();
-		Preconditions.checkArgument(!children.isEmpty() && children.size() % 2 == 0);
+		checkArgument(call, !children.isEmpty() && children.size() % 2 == 0);
 		List<RexNode> childrenRexNode = toRexNodes(context, children);
 		RelDataType keyType = childrenRexNode.get(0).getType();
 		RelDataType valueType = childrenRexNode.get(childrenRexNode.size() - 1).getType();
@@ -311,13 +310,10 @@ public class CustomizedConvertRule implements CallExpressionConvertRule {
 		checkArgumentNumber(call, 2);
 		RexNode child = context.toRexNode(call.getChildren().get(0));
 		ValueLiteralExpression keyLiteral = (ValueLiteralExpression) call.getChildren().get(1);
-		Optional<Integer> indexOptional = ExpressionUtils.extractValue(keyLiteral, String.class).map(
-			child.getType().getFieldNames()::indexOf);
-		// Note: never replace the following code with :
-		// int index = indexOptional.orElseGet(() -> extractValue(keyLiteral, Integer.class));
-		// Because the logical in `orElseGet` always executed no matter whether indexOptional is present or not.
-		int index;
-		index = indexOptional.orElseGet(() -> extractValue(keyLiteral, Integer.class));
+		Optional<Integer> indexOptional = ExpressionUtils
+				.extractValue(keyLiteral, String.class)
+				.map(child.getType().getFieldNames()::indexOf);
+		int index = indexOptional.orElseGet(() -> extractValue(keyLiteral, Integer.class));
 		return context.getRelBuilder().getRexBuilder().makeFieldAccess(child, index);
 	}
 
@@ -328,8 +324,7 @@ public class CustomizedConvertRule implements CallExpressionConvertRule {
 			QueryOperation tableOperation = ((TableReferenceExpression) headExpr).getQueryOperation();
 			RexNode child = context.toRexNode(call.getChildren().get(0));
 			return RexSubQuery.in(((FlinkRelBuilder) context.getRelBuilder())
-					.queryOperation(tableOperation).build(),
-				ImmutableList.of(child));
+					.queryOperation(tableOperation).build(), ImmutableList.of(child));
 		} else {
 			List<RexNode> child = toRexNodes(context, call.getChildren());
 			return context.getRelBuilder().call(FlinkSqlOperatorTable.IN, child);
@@ -376,10 +371,7 @@ public class CustomizedConvertRule implements CallExpressionConvertRule {
 
 	private static void checkArgument(CallExpression call, boolean check) {
 		if (!check) {
-			throw new TableException(String.format(
-				"Invalid number of arguments for call: %s. %d arguments passed.",
-				call.getFunctionDefinition(),
-				call.getChildren().size()));
+			throw new TableException("Invalid arguments for call: " + call);
 		}
 	}
 
--- a/flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/expressions/converter/CustomizedConvertRule.java
+++ b/flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/expressions/converter/CustomizedConvertRule.java
@@ -18,6 +18,7 @@
 
 package org.apache.flink.table.planner.expressions.converter;
 
+import org.apache.flink.table.api.TableException;
 import org.apache.flink.table.expressions.CallExpression;
 import org.apache.flink.table.expressions.Expression;
 import org.apache.flink.table.expressions.ExpressionUtils;
@@ -53,6 +54,7 @@ import java.util.Optional;
 import static org.apache.calcite.sql.type.SqlTypeName.VARCHAR;
 import static org.apache.flink.table.planner.calcite.FlinkTypeFactory.toLogicalType;
 import static org.apache.flink.table.planner.expressions.converter.ExpressionConverter.extractValue;
+import static org.apache.flink.table.planner.expressions.converter.ExpressionConverter.toRexNodes;
 import static org.apache.flink.table.runtime.types.LogicalTypeDataTypeConverter.fromDataTypeToLogicalType;
 import static org.apache.flink.table.runtime.typeutils.TypeCheckUtils.isCharacterString;
 import static org.apache.flink.table.runtime.typeutils.TypeCheckUtils.isTemporal;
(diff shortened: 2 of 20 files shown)


flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/expressions/RexNodeConverter.java
```diff
@@ -865,6 +865,9 @@ private static Calendar valueAsCalendar(Object value) {
 
 	@Override
 	public RexNode visit(FieldReferenceExpression fieldReference) {
+		// We can not use inputCount+inputIndex+FieldIndex to construct field of calcite.
```
JingsongLi: @dawidwys @twalthr What do you think about this question? Should it be a bug of `QueryOperationConverter`? But I don't find any feasible in `RelBuilder`. Maybe we should keep same order in `AggregateQueryOperation` or adjust the order of existed `OperationTree`? Both ways look hard...

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/expressions/converter/ExpressionConverter.java
```diff
@@ -0,0 +1,322 @@
...
+			TimeUnitRange value = intervalUnitToUnitRange((TimeIntervalUnit) object);
+			return relBuilder.getRexBuilder().makeFlag(value);
+		} else {
+			return relBuilder.literal(extractValue(valueLiteral, Object.class));
+		}
+	}
+
+	@Override
+	public RexNode visit(FieldReferenceExpression fieldReference) {
+		// We can not use inputCount+inputIndex+FieldIndex to construct field of calcite.
```
JingsongLi: @dawidwys @twalthr What do you think about this question? Should it be a bug of QueryOperationConverter? But I don't find any feasible way in RelBuilder. Maybe we should keep same order in AggregateQueryOperation or adjust the order of existed OperationTree? Both ways look hard...

JingsongLi:
@wuchong @dawidwys @beyond1920 I hope you can have a look.

dawidwys:
Will try to do it today.

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/expressions/converter/ExpressionConverter.java
dawidwys: Yes, I think I also stumbled upon this problem myself. I did not figure out a good solution for it back then. But I was thinking about reordering the fields when converting the QueryOperation to RelNode. I am not sure how feasible that is.

dawidwys: I think it makes sense to open a JIRA issue for this issue.

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/expressions/converter/OverConvertRule.java
```diff
@@ -0,0 +1,195 @@
...
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import static org.apache.flink.table.planner.expressions.converter.ExpressionConverter.extractValue;
+import static org.apache.flink.table.runtime.types.LogicalTypeDataTypeConverter.fromDataTypeToLogicalType;
+
+/**
+ * {@link CallExpressionConvertRule} to convert over sql.
```
dawidwys: A {@link CallExpressionConvertRule} that converts {@link BuiltInFunctionDefinitions.OVER}.

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/expressions/converter/DefinedConvertRule.java
```diff
@@ -0,0 +1,158 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+
+/**
+ * Defined function {@link CallExpressionConvertRule}, it included conversions are one-to-one with
+ * calcite SqlOperator.
+ */
+public class DefinedConvertRule implements CallExpressionConvertRule {
```
dawidwys: `DirectConvertRule`? But not sure if this is much better name. `Defined` sounds a bit strange to me here. In fact all the rules are defined in some way.

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/expressions/converter/DefinedConvertRule.java
```diff
@@ -0,0 +1,158 @@
...
+
+import org.apache.calcite.rex.RexNode;
+import org.apache.calcite.sql.SqlOperator;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+
+/**
+ * Defined function {@link CallExpressionConvertRule}, it included conversions are one-to-one with
```
dawidwys: `A {@link CallExpressionConvertRule} that performs a simple one-to-one mapping between {@link Expression} and a corresponding {@link SqlOperator}.`

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/expressions/converter/DefinedConvertRule.java
```diff
@@ -0,0 +1,158 @@
...
+		DEFINITION_OPERATOR_MAP.put(BuiltInFunctionDefinitions.SHA512, FlinkSqlOperatorTable.SHA512);
+		DEFINITION_OPERATOR_MAP.put(BuiltInFunctionDefinitions.SHA1, FlinkSqlOperatorTable.SHA1);
+		DEFINITION_OPERATOR_MAP.put(
+			BuiltInFunctionDefinitions.STREAM_RECORD_TIMESTAMP, FlinkSqlOperatorTable.STREAMRECORD_TIMESTAMP);
+	}
+
+	@Override
+	public Optional<RexNode> convert(CallExpression call, ConvertContext context) {
+		SqlOperator operator = DEFINITION_OPERATOR_MAP.get(call.getFunctionDefinition());
+		if (operator == null) {
```
dawidwys:
```
return Optional.of(operator)
    .map(op -> context.getRelBuilder().call(op, context.toRexNodes(call.getChildren())))
```

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/expressions/converter/CallExpressionConvertRule.java
```diff
@@ -0,0 +1,62 @@
...
+	 * Context of {@link CallExpressionConvertRule}.
+	 */
+	interface ConvertContext {
+
+		/**
+		 * Convert expression to RexNode, used by children conversion.
+		 */
+		RexNode toRexNode(Expression expr);
+
+		default List<RexNode> toRexNodes(List<Expression> expr) {
```
dawidwys: Can you think of any reason we could ever want to override this method? If not can we put it somewhere else? If I understood it correctly it's just a helper method, no?

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/expressions/converter/CustomizedConvertRule.java
```diff
@@ -0,0 +1,361 @@
...
+			SqlThrowExceptionFunction function = new SqlThrowExceptionFunction(
+					context.getTypeFactory().createFieldTypeFromLogicalType(fromDataTypeToLogicalType(type)));
+			return context.getRelBuilder().call(function, context.toRexNode(children.get(0)));
+		});
+	}
+
+	@Override
+	public Optional<RexNode> convert(CallExpression call, ConvertContext context) {
+		Conversion conversion = DEFINITION_RULE_MAP.get(call.getFunctionDefinition());
+		if (conversion == null) {
```
dawidwys: `Optional.of(conversion).map(...)`

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/expressions/converter/CustomizedConvertRule.java
```diff
@@ -0,0 +1,361 @@
...
+	}
+
+	private static RexNode convertArrayElement(List<Expression> children, ConvertContext context) {
+		List<RexNode> childrenRexNode = context.toRexNodes(children);
+		return context.getRelBuilder().call(FlinkSqlOperatorTable.ELEMENT, childrenRexNode);
+	}
+
+	private static RexNode convertOrderAsc(List<Expression> children, ConvertContext context) {
+		List<RexNode> childrenRexNode = context.toRexNodes(children);
+		return childrenRexNode.get(0);
```
dawidwys: Can we assert for the number of arguments? Similar as in `PlannerExpressionConverter`? It actually applies to all the Conversions.

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/expressions/converter/CustomizedConvertRule.java
```diff
@@ -0,0 +1,361 @@
...
+		RexNode lowerBound = childrenRexNode.get(1);
+		RexNode upperBound = childrenRexNode.get(2);
+		return context.getRelBuilder().or(
+			context.getRelBuilder().call(FlinkSqlOperatorTable.LESS_THAN, expr, lowerBound),
+			context.getRelBuilder().call(FlinkSqlOperatorTable.GREATER_THAN, expr, upperBound));
+	}
+
+	private static RexNode convertBetween(List<Expression> children, ConvertContext context) {
+		List<RexNode> childrenRexNode = context.toRexNodes(children);
+		Preconditions.checkArgument(childrenRexNode.size() == 3);
```
dawidwys: How about we use `TableException` here? Also there are some Conversions that do not check the number of arguments. We should do that in all Conversions.

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/expressions/converter/CustomizedConvertRule.java
```diff
@@ -0,0 +1,361 @@
...
+		RexNode checkOverflow = context.toRexNode(children.get(2));
+		return context.getRelBuilder().getRexBuilder().makeReinterpretCast(
+			context.getTypeFactory().createFieldTypeFromLogicalType(
+				type.getOutputDataType().getLogicalType().copy(child.getType().isNullable())),
+			child,
+			checkOverflow);
+	}
+
+	private interface Conversion {
+		RexNode convert(List<Expression> children, ConvertContext context);
```
dawidwys: How about changing that to:
`RexNode convert(CallExpression children, ConvertContext context);`?

This would solve the comments:
`// TODO get type from CallExpression directly`

In the current interface we simplify just a rather simple call: `call.getChildren`

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/expressions/converter/CustomizedConvertRule.java
```diff
@@ -0,0 +1,361 @@
...
+		return childrenRexNode.get(0);
+	}
+
+	private static RexNode convertTimestampDiff(List<Expression> children, ConvertContext context) {
+		List<RexNode> childrenRexNode = context.toRexNodes(children);
+		return context.getRelBuilder().call(FlinkSqlOperatorTable.TIMESTAMP_DIFF, childrenRexNode.get(0), childrenRexNode.get(2),
+			childrenRexNode.get(1));
+	}
+
+	private static RexNode convertIsNull(List<Expression> children, ConvertContext context) {
```
dawidwys: Why isn't this in `DefinedConvertRule`?

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/expressions/converter/DefinedConvertRule.java
```diff
@@ -0,0 +1,158 @@
...
+
+/**
+ * Defined function {@link CallExpressionConvertRule}, it included conversions are one-to-one with
+ * calcite SqlOperator.
+ */
+public class DefinedConvertRule implements CallExpressionConvertRule {
+
+	private static final Map<FunctionDefinition, SqlOperator> DEFINITION_OPERATOR_MAP = new HashMap<>();
+	static {
+		DEFINITION_OPERATOR_MAP.put(BuiltInFunctionDefinitions.SHA1, FlinkSqlOperatorTable.SHA1);
```
dawidwys: Shouldn't this use the `standardSql` property as discussed in the `FLIP`?

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/expressions/converter/CustomizedConvertRule.java
```diff
@@ -0,0 +1,361 @@
...
+		TypeLiteralExpression type = (TypeLiteralExpression) children.get(1);
+		return context.getRelBuilder().getRexBuilder().makeAbstractCast(
+			context.getTypeFactory().createFieldTypeFromLogicalType(
+				type.getOutputDataType().getLogicalType().copy(child.getType().isNullable())),
+			child);
+	}
+
+	private static RexNode convertArrayElement(List<Expression> children, ConvertContext context) {
+		List<RexNode> childrenRexNode = context.toRexNodes(children);
+		return context.getRelBuilder().call(FlinkSqlOperatorTable.ELEMENT, childrenRexNode);
```
dawidwys: Why isn't this in `DefinedConvertRule`?

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/expressions/converter/ExpressionConverter.java
JingsongLi: I create https://issues.apache.org/jira/browse/FLINK-13902

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/expressions/converter/DefinedConvertRule.java
JingsongLi: Good point, I think it is `Optional.ofNullable`

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/expressions/converter/CallExpressionConvertRule.java
JingsongLi: Right, it should be a static helper method.

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/expressions/converter/CustomizedConvertRule.java
JingsongLi: The idea was to let calcite assert, but active assert would be better.

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/expressions/converter/CustomizedConvertRule.java
JingsongLi: It should be, now we have removed the `UnresolvedCallExpression`, we can get all output types.

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/expressions/converter/CustomizedConvertRule.java
JingsongLi: I'll remove it.

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/expressions/converter/DefinedConvertRule.java
JingsongLi: Yeah, I will introduce a new Rule and remove these after I introduce `standardSql`.

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/expressions/converter/CustomizedConvertRule.java
JingsongLi: I'll move it.

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/expressions/converter/CustomizedConvertRule.java
JingsongLi: But we still use PlannerExpressions to infer output type, that will lead to wrong precision of DecimalType.
So I will keep this TODO until we finish Expression type inference.

JingsongLi:
Hi @dawidwys , Thanks for your review. Comments addressed.
Sorry for the force push, I try to fix the checkstyle of c3a0c9b .

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/expressions/converter/DefinedConvertRule.java
dawidwys: Yes you are right, it should be `ofNullable`.

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/expressions/converter/OverConvertRule.java
```diff
@@ -0,0 +1,195 @@
...
+				upperBound,
+				isPhysical,
+				true,
+				false,
+				false));
+		}
+		return Optional.empty();
+	}
+
+	private RexNode createCollation(RexNode node, RelFieldCollation.Direction direction,
```
dawidwys: Could we break the parameters according to our code style guidelines: https://flink.apache.org/contributing/code-style-and-quality-formatting.html#breaking-the-lines-of-too-long-statements ?

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/expressions/converter/CustomizedConvertRule.java
```diff
@@ -0,0 +1,389 @@
...
+		}
+
+		checkArgument(call, find);
+	}
+
+	private static void checkArgument(CallExpression call, boolean check) {
+		if (!check) {
+			throw new TableException(String.format(
+				"Invalid number of arguments for call: %s. %d arguments passed.",
+				call.getFunctionDefinition(),
```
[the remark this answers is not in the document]
JingsongLi: +1, can be simplified to :
`"Invalid arguments for call: " + call`

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/expressions/converter/OverConvertRule.java
JingsongLi: You are right, got it.

flink-table/flink-table-planner-blink/src/main/java/org/apache/flink/table/planner/expressions/converter/CustomizedConvertRule.java
```diff
@@ -0,0 +1,389 @@
...
+
+	private static RexNode convertGet(CallExpression call, ConvertContext context) {
+		checkArgumentNumber(call, 2);
+		RexNode child = context.toRexNode(call.getChildren().get(0));
+		ValueLiteralExpression keyLiteral = (ValueLiteralExpression) call.getChildren().get(1);
+		Optional<Integer> indexOptional = ExpressionUtils.extractValue(keyLiteral, String.class).map(
+			child.getType().getFieldNames()::indexOf);
+		// Note: never replace the following code with :
+		// int index = indexOptional.orElseGet(() -> extractValue(keyLiteral, Integer.class));
+		// Because the logical in `orElseGet` always executed no matter whether indexOptional is present or not.
```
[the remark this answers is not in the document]
JingsongLi: I think this previous comment want to tell: The logical in `Optional.orElse` always executed no matter whether original optional is present or not. (So I change `isPresent` code to `orElseGet`)

`extractValue` will throw exception when extract fail, so we need avoid the evaluated here when the literal is string instead of integer.

I will remove this comment.

JingsongLi:
@dawidwys Thanks for you review, comments addressed. @wuchong can you take a look to 7052fdc ?

wuchong:
The travis was passed: https://travis-ci.org/wuchong/flink/builds/580167972
I will merge this.

PR #9486, opened 2019-08-19
[FLINK-13789] move prefix out of the format string

haodang:

Fixing https://issues.apache.org/jira/browse/FLINK-13789 by moving `prefix` out of the format string.

## Brief change log

- moving `prefix` out of the format string

## Verifying this change

This change is almost a trivial rework and we've been running such a version on our production environment and have verified the fix.

  - The serializers: no

## Documentation


haodang:
@carp84, Thank you, Yu!  Let me add a unit test.
Not sure of why the build failed, will investigate.

StephanEwen:
If we are now doing string concatenation anyways, should we just remove `String.format` all together here? In this way (mixing the two) it seems a bit strange (and is probably also a bit less efficient).

TisonKun:
make sense to change it into

```java
    return prefix + "-" + transactionalId;
```

since we don't want to do more format job here.

zentol:
@flinkbot run travis

PR #9487, opened 2019-08-19
[FLINK-13791][docs] Speed up sidenav by using group_by

NicoK:

`_includes/sidenav.html` parses through `pages_by_language` over and over again trying to find children when building the (recursive) side navigation. By doing this once with a `group_by`, we can gain considerable savings in building the docs via `./build_docs.sh` without any change to the generated HTML pages:

before: ~54s
after: ~37s

Next PR in this series: #9489

## Brief change log

Building on top of #9444, this PR adds:
- use `group_by` to create an easier-iterable array for determining page children

## Verifying this change

I verified the changes in the generated HTML pages (nothing changed).


StephanEwen:
Looks good, merging this...

PR #9489, opened 2019-08-20
[FLINK-13793][docs] build each language in a separate subprocess

NicoK:

When not serving docs and just building them, e.g. via `./build_docs.sh`, we can spawn sub-processes for each language build instead of running them single-threaded (jekyll currently does not support parallel builds).

before: ~37s
after: ~20s

## Brief change log

Building on top of #9487, this PR adds:
- change `build_docs.sh` to support parallel builds and merge resulting files together
(admittedly this seems a bit hacky)
- enhance `build_docs.sh`: show usage if wrong option is given
- exclude build scripts from generated contents

## Verifying this change

I verified the changes in the generated HTML pages (nothing changed except for the build scripts which I removed from the generated pages via a hotfix).

### DIFF
--- a/docs/_config_dev_en.yml
+++ b/docs/_config_dev_en.yml
@@ -17,6 +17,9 @@
 
 exclude:
   - "*.zh.md"
+  - "build_docs.sh"
+  - "build_docs.bat"
+  - "check_links.sh"
   - "content"
   - "content_en"
   - "content_zh"
--- a/docs/_config_dev_en.yml
+++ b/docs/_config_dev_en.yml
@@ -17,3 +17,6 @@
 
 exclude:
   - "*.zh.md"
+  - "content"
+  - "content_en"
+  - "content_zh"
--- a/docs/build_docs.sh
+++ b/docs/build_docs.sh
@@ -72,6 +72,8 @@ while getopts "piez" opt; do
 		z)
 		JEKYLL_CONFIG="--config _config.yml,_config_dev_zh.yml"
 		;;
+		*) echo "usage: $0 [-e|-z] [-i|-p]" >&2
+		exit 1 ;;
 	esac
 done
 
--- a/docs/build_docs.sh
+++ b/docs/build_docs.sh
@@ -53,6 +53,8 @@ JEKYLL_CMD="build"
 
 JEKYLL_CONFIG=""
 
+DOC_LANGUAGES="en zh"
+
 # if -p flag is provided, serve site on localhost
 # -i is like -p, but incremental (only rebuilds the modified file)
 # -e builds only english documentation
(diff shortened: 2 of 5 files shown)


NicoK:
I just rebased this PR now that we haveJekyll 4 changes in from #9487.

docs/build_docs.sh
```diff
@@ -72,6 +72,8 @@ while getopts "piez" opt; do
 		z)
 		JEKYLL_CONFIG="--config _config.yml,_config_dev_zh.yml"
 		;;
+    *) echo "usage: $0 [-e|-z] [-i|-p]" >&2
```
zentol: indentation is different

docs/_config_dev_en.yml
```diff
@@ -17,3 +17,6 @@
 
 exclude:
   - "*.zh.md"
+  - "content"
```
zentol: are these required for correctness?

docs/build_docs.sh
```diff
@@ -78,4 +80,28 @@ while getopts "piez" opt; do
 done
 
 # use 'bundle exec' to insert the local Ruby dependencies
-bundle exec jekyll ${JEKYLL_CMD} ${JEKYLL_CONFIG} --source "${DOCS_SRC}" --destination "${DOCS_DST}"
+
+if [ "${JEKYLL_CMD}" = "build" ] && [ -z "${JEKYLL_CONFIG}" ]; then
```
zentol: if we modified the e/z cases above to instead set DOC_LANGUAGES to en/zh we wouldn't need this condition nor else branch

zentol:
@NicoK ping

docs/_config_dev_en.yml
NicoK: if not excluded, they may end up in the generated folder (as yet another subdirectory) again, e.g. when the zh version builds, if would not see the `content_en` as a target dir (and exclude it)

docs/build_docs.sh
NicoK: maybe a branch for the DOC_LANGUAGES (to get all languages if none of the e/z options was used) but simpler anyway; I'll give it a try

NicoK: oh, the branch is still always needed since there is no parallel `serve`...so not sure whether this change makes sense

docs/build_docs.sh
```diff
@@ -72,8 +74,34 @@ while getopts "piez" opt; do
...
+  # run parallel builds for all languages if not serving or creating a single language only
+
+  # run processes and store pids
+  echo "Spawning parallel builds for ${DOC_LANGUAGES}..."
+  pids=
```
zentol: my god is this actually valid in bash? It looks soo much like an incomplete statement.

NicoK: yes it it ;)
but I can also replace it with `pids=""`

NicoK:
Serving  `en` and building `zh` in the background would be a nice trick but the result depends on the timing of when the `zh` build finishes. If it is first, then the `en` build will remove the `contents` directory and replace it with its own (thus removing the `zh` files) :(

NicoK:
Ideally, of course, these tricks would not be necessary and jekyll would just use more system resources, i.e. do a parallel build on its own. That would give an even higher speed-up but is, unfortunately, not available, afaik

NicoK:
FYI: I rebased onto latest master again and verified that the result from building the old way vs. this PR is the same

PR #9490, opened 2019-08-20
[hotfix][python] add System.exit() at the end of PythonGatewayServer to ensure the JVM will exit if its parent process dies.

WeiZhong94:

*This pull request add System.exit() at the end of PythonGatewayServer to ensure the JVM will exit if its parent process dies.*

## Brief change log

  - *add System.exit() at the end of PythonGatewayServer.*

## Verifying this change

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-python/src/main/java/org/apache/flink/client/python/PythonGatewayServer.java
+++ b/flink-python/src/main/java/org/apache/flink/client/python/PythonGatewayServer.java
@@ -78,7 +78,7 @@ public class PythonGatewayServer {
 			}
 			gatewayServer.shutdown();
 			System.exit(0);
-		} catch (IOException e) {
+		} finally {
 			System.exit(1);
 		}
 	}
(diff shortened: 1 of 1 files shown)


flink-python/src/main/java/org/apache/flink/client/python/PythonGatewayServer.java
```diff
@@ -70,11 +70,16 @@ public static void main(String[] args) throws IOException {
...
+		try {
+			// Exit on EOF or broken pipe.  This ensures that the server dies
+			// if its parent program dies.
+			while (System.in.read() != -1) {
+				// Do nothing
+			}
+			gatewayServer.shutdown();
+			System.exit(0);
+		} catch (IOException e) {
+			System.exit(1);
```
hequn8128: Maybe we should add this exit in the finally block and also print the error message of the error.

WeiZhong94:
@hequn8128  Thanks a lot for your review. This change is to make sure the JVM could exit after the Python process exits even if there are some threads running in the JVM. System.exit is a special method provided to terminates the currently running JVM, and the test seems not necessary for me as it seems that we are trying to test the behavior of the JVM. Does that make sense to you?

hequn8128:
@WeiZhong94 Thanks a lot for the explanation. LGTM, merging...

PR #9493, opened 2019-08-20
[FLINK-13797] Add missing format argument

Fokko:

Add missing format argument. We expect an argument, but it is missing.

https://jira.apache.org/jira/browse/FLINK-13797

## Brief changelog

We produce log lines to help to debug the workflow. I've noticed that there are some issues with one of the loglines. Doing a quick search, I've noticed a few more.

## Verifying this change

This change is a trivial rework/code cleanup without any test coverage.

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-filesystems/flink-hadoop-fs/src/main/java/org/apache/flink/runtime/fs/hdfs/HadoopFsFactory.java
+++ b/flink-filesystems/flink-hadoop-fs/src/main/java/org/apache/flink/runtime/fs/hdfs/HadoopFsFactory.java
@@ -118,7 +118,12 @@ public class HadoopFsFactory implements FileSystemFactory {
 				initUri = fsUri;
 			}
 			else {
-				LOG.debug("URI {} does not specify file system authority, trying to load default authority (fs.defaultFS)");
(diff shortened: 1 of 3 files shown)


flink-filesystems/flink-hadoop-fs/src/main/java/org/apache/flink/runtime/fs/hdfs/HadoopFsFactory.java
```diff
@@ -118,7 +118,12 @@ else if (flinkConfig != null) {
 				initUri = fsUri;
 			}
 			else {
-				LOG.debug("URI {} does not specify file system authority, trying to load default authority (fs.defaultFS)");
+				if (LOG.isDebugEnabled()) {
+					LOG.debug(
+						"URI {} does not specify file system authority, trying to load default authority (fs.defaultFS)",
+						fsUri.toString()
```
TisonKun: `.toString` is no need here.

Fokko:
Thanks for the pointer, I wasn't aware of the updated process.

PR #9495, opened 2019-08-20
[FLINK-13798][task] Refactor the process of checking stream status while emitting watermark in source

zhijiangW:
##  What is the purpose of the change

As we know, the watermark could be emitted to downstream only when the stream status is active. For the downstream task we already have the component of `StatusWatermarkValve` in StreamInputProcessor to handle this logic. But for the source task the current implementation of this logic seems a bit tricky.

There are two scenarios for the source case:

1. In the source `WatermarkContext`, it would toggle the status as active while collecting/emitting and the status is checked in `RecordWriterOutput`. If the watermark is triggered by timer for `AutomaticWatermarkContext`, the timer task would check the status before emitting watermark.

2. `TimestampsAndPeriodicWatermarksOperator`: The watermark is triggered by timer, but it still relies on `RecordWriterOutput` to check the status before emitting.

So the check logic in `RecordWriterOutput` only makes sense for the last scenario, and seems redundant for the first scenario. Even worse, this logic in `RecordWriterOutput` would bring cycle dependency with `StreamStatusMaintainer`, which is a blocker for the following work of integrating source processing on runtime side. To solve above issues, the basic idea is to refactor this check logic in upper layer instead of current low level `RecordWriterOutput`. The solution is migrating the check logic from `RecordWriterOutput` to `TimestampsAndPeriodicWatermarksOperator`. And we could further remove the logic of toggling active in `WatermarkContext`.

## Brief change log

  - *Introduce the `StreamStatusProvider` into `TimestampsAndPeriodicWatermarksOperator`*
  - *Remove the logic of checking status from `RecordWriterOutput`*
  - *Remove the logic of toggling active status in `WatermarkSourceContext`*

## Verifying this change

## Documentation

### DIFF
--- a/flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/StreamSourceContexts.java
+++ b/flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/StreamSourceContexts.java
@@ -383,8 +383,6 @@ public class StreamSourceContexts {
 		@Override
 		public void collect(T element) {
 			synchronized (checkpointLock) {
-				streamStatusMaintainer.toggleStreamStatus(StreamStatus.ACTIVE);
-
 				if (nextCheck != null) {
 					this.failOnNextCheck = false;
 				} else {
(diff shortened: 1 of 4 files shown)


zhijiangW:
close it as explained in the jira ticket.

PR #9497, opened 2019-08-21
[FLINK-13794][client] Remove unused logic of printStatusDuringExecution

TisonKun:

Remove unused logic of printStatusDuringExecution. Because it accidentally depends on remove methods in `ExecutionConfig`(In any way the methods relevant are no power but the class is marked as `@Public`), we might fall back to mark the methods relevant as `@Deprecated` and still live with them.

Please let me know what is the proper approach to treat these "breaking changes".

## Verifying this change

  - The public API, i.e., is any changed class annotated with `@Public(Evolving)`: (yes, ExecutionConfig, accidentally.)
  - The serializers: (no)

## Documentation

cc @mxm

### DIFF
--- a/flink-clients/src/main/java/org/apache/flink/client/program/ClusterClient.java
+++ b/flink-clients/src/main/java/org/apache/flink/client/program/ClusterClient.java
@@ -86,9 +86,6 @@ public abstract class ClusterClient<T> {
 
 	private final boolean sharedHaServices;
 
-	/** Flag indicating whether to sysout print execution updates. */
-	private boolean printStatusDuringExecution = true;
-
 	/**
 	 * For interactive invocations, the job results are only available after the ContextEnvironment has
 	 * been run inside the user JAR. We pass the Client to every instance of the ContextEnvironment
@@ -162,23 +159,6 @@ public abstract class ClusterClient<T> {
 	// ------------------------------------------------------------------------
 	//  Configuration
 	// ------------------------------------------------------------------------
-
-	/**
-	 * Configures whether the client should print progress updates during the execution to {@code System.out}.
-	 * All updates are logged via the SLF4J loggers regardless of this setting.
-	 *
-	 * @param print True to print updates to standard out during execution, false to not print them.
-	 */
-	public void setPrintStatusDuringExecution(boolean print) {
-		this.printStatusDuringExecution = print;
-	}
-
-	/**
-	 * @return whether the client will print progress updates during the execution to {@code System.out}
-	 */
-	public boolean getPrintStatusDuringExecution() {
-		return this.printStatusDuringExecution;
-	}
 
 	/**
 	 * Gets the current cluster connection info (may change in case of a HA setup).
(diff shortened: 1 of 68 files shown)


TisonKun:
Travis should fail on japicmp. I don't add `exclude` to verify the expected failure.

TisonKun:
Travis failed at 

>08:54:01.376 [ERROR] Failed to execute goal com.github.siom79.japicmp:japicmp-maven-plugin:0.11.0:cmp (default) on project flink-core: Breaking the build because there is at least one incompatibility: org.apache.flink.api.common.ExecutionConfig.disableSysoutLogging():METHOD_REMOVED,org.apache.flink.api.common.ExecutionConfig.enableSysoutLogging():METHOD_REMOVED,org.apache.flink.api.common.ExecutionConfig.isSysoutLoggingEnabled():METHOD_REMOVED -> [Help 1]

as expected.

Here is the problem: shall we add excludes to suppress errors or replace with dummy implementation and mark it as `@Deprecated`?

TisonKun:
I would assume there is few or none user make use of these method but the compatibility is promised by `@Public`.

flink-clients/src/main/java/org/apache/flink/client/program/ClusterClient.java
```diff
@@ -163,23 +160,6 @@ public void shutdown() throws Exception {
...
 
-	/**
-	 * Configures whether the client should print progress updates during the execution to {@code System.out}.
-	 * All updates are logged via the SLF4J loggers regardless of this setting.
-	 *
-	 * @param print True to print updates to standard out during execution, false to not print them.
-	 */
-	public void setPrintStatusDuringExecution(boolean print) {
-		this.printStatusDuringExecution = print;
-	}
```
mxm: Could we deprecate this method and just log a deprecation warning? We could even remove the printing logic, so this would just be a placeholder not to break the @Public contract.

flink-clients/src/main/java/org/apache/flink/client/program/ClusterClient.java
```diff
@@ -163,23 +160,6 @@ public void shutdown() throws Exception {
...
-
-	/**
-	 * @return whether the client will print progress updates during the execution to {@code System.out}
-	 */
-	public boolean getPrintStatusDuringExecution() {
-		return this.printStatusDuringExecution;
-	}
```
mxm: Could we deprecate this method and just log a deprecation warning? We could even remove the printing logic, so this would just be a placeholder not to break the @Public contract.

TisonKun:
Thanks for your review @mxm !

In fact the only "break change" is in `ExecutionConfig`. I just added a fixup to add back those method and mark as deprecated. Since they have no power any more we can dummy the body.

Please check the final commit.

TisonKun:
@flinkbot run travis

TisonKun:
@flinkbot run travis

TisonKun:
@mxm now we pass api compatibility test also other tests. I think this pull request has reached a mergeable state. Please take another pass of review :-)

TisonKun:
@kl0u @aljoscha could you take a look of this?

TisonKun:
Thanks for your review @mxm! Could you help merge this pull request?

mxm:
@TisonKun Could you squash the commits?

TisonKun:
@mxm done.

PR #9498, opened 2019-08-21
[FLINK-13810][Docs]Update `Elasticsearch Connector` label

hehuiyuan:

For document: https://ci.apache.org/projects/flink/flink-docs-release-1.8/dev/table/connect.html#elasticsearch-connector
 
<span class="label label-primary">Sink: Streaming Append Mode</span>
<span class="label label-primary">Sink: Streaming Upsert Mode</span>
<span class="label label-info">Format: JSON-only</span>
 
The first lable for `Elasticsearch Connector` should be Source.

## Documentation

  - Does this pull request introduce a new feature? (  no)

### DIFF
--- a/docs/dev/table/connect.md
+++ b/docs/dev/table/connect.md
@@ -771,7 +771,7 @@ Make sure to add the version-specific Kafka dependency. In addition, a correspon
 
 ### Elasticsearch Connector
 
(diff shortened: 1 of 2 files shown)


StephanEwen:
I believe the docs are correct as they are. ES is a sink in both cases not a source.

PR #9499, opened 2019-08-21
[FLINK-13809][python] Set more relaxed permission on "log" directory when installing Flink Python API via "sudo".

WeiZhong94:

*This pull request sets more relaxed permission on "log" directory when installing Flink Python API via "sudo".*

## Brief change log

  - *Add custom command class in setup.py to set a more relaxed permission on "log" directory if current user is "root".*

## Verifying this change

  *1. pip install the python package via "sudo"*
  *2. execute "pyflink-shell.sh". If no exception is thrown, it is ok.*

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-python/setup.py
+++ b/flink-python/setup.py
@@ -17,17 +17,33 @@
 ################################################################################
 from __future__ import print_function
 
+import getpass
 import io
 import os
 import sys
 from shutil import copytree, copy, rmtree
 
 from setuptools import setup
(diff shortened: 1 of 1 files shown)


StephanEwen:
I think this fix would work, but I am wondering: Is that the common way to do this?

For many UNIX programs, when they are installed shared (under root) , then I would guess that the logs should be somewhere under int user's directory.

PR #9504, opened 2019-08-22
[FLINK-13799][web]: fix error message web submit is disables in the c…

vthinkxie:
…onfig

fix https://issues.apache.org/jira/browse/FLINK-13799

## Brief change log

there is no method to get if web.submit.enable is false in rest API

I hide the jar error message from rest by modifying the frontend interceptor

## Verifying this change

If you put web.submit.enable: false into the configuration, no errors will continuously pop up.

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-runtime-web/web-dashboard/src/app/app.interceptor.ts
+++ b/flink-runtime-web/web-dashboard/src/app/app.interceptor.ts
@@ -32,7 +32,7 @@ export class AppInterceptor implements HttpInterceptor {
      * Error response from below url should be ignored
      */
     const ignoreErrorUrlEndsList = ['checkpoints/config', 'checkpoints'];
-    const ignoreErrorMessage = ['File not found.'];
+    const ignoreErrorMessage = ['File not found.', 'Unable to load requested file /jars.'];
     return next.handle(req).pipe(
       catchError(res => {
         const errorMessage = res && res.error && res.error.errors && res.error.errors[0];


sorahn:
LGTM

flink-runtime-web/web-dashboard/src/app/app.interceptor.ts
```diff
@@ -32,7 +32,7 @@ export class AppInterceptor implements HttpInterceptor {
      * Error response from below url should be ignored
      */
     const ignoreErrorUrlEndsList = ['checkpoints/config', 'checkpoints'];
-    const ignoreErrorMessage = ['File not found.'];
+    const ignoreErrorMessage = ['File not found.', 'Unable to load requested file /jars.'];
```
zentol: While this does cover the case of multiple error messages when accessing the page, the user still gets odd error messages if he tries to use the submit page in any way.

If we want to backport the proper solution (See subtasks of FLINK-13799) to 1.9, then I would skip this PR.
If we don't want to backport it, then we need a more extensive fix imo.

vthinkxie: the upload button in the submit page is already hidden when getting the error message, so the user could not use the submit page in the case.
I think we should wait for the proper solution.

vthinkxie:
Hi @zentol 
I found that the web can get web.submit.enable from the jobmanager configuration, is that correct to judge if the submit is enabled from the key `web.submit.enable`?
屏幕快照 2019-09-19 下午4 42 13

zentol:
@vthinkxie FLINK-13817 exposes whether submissions are enabled in a more reliable way. The configuration is is unreliable since it only lists explicitly configured keys; if the key is not explicitly configured you'd have to make assumptions on the default.

zentol:
I will close this PR for the time being. If we do not manage to implement the approach outlined in FLINK-13818 in time for a release we can fallback on this proposal.

PR #9474, opened 2019-08-18
[FLINK-13761][ScalaAPI]`SplitStream` should be deprecated because `SplitJavaStream` is deprecated

izhangzhihao:
## What is the purpose of the change

`SplitStream` should be deprecated because `SplitJavaStream` is deprecated.

## Brief change log

Deprecated `SplitStream` scala api.

## Verifying this change

This change does not affect any tests.

## Does this pull request potentially affect one of the following parts:

  - Dependencies (does it add or upgrade a dependency): (yes / **no**)
  - The public API, i.e., is any changed class annotated with `@Public(Evolving)`: (**yes**/ no)
  - The serializers: (yes / **no** / don't know)
  - The runtime per-record code paths (performance sensitive): (yes / **no** / don't know)
  - Anything that affects deployment or recovery: JobManager (and its components), Checkpointing, Yarn/Mesos, ZooKeeper: (yes / **no** / don't know)
  - The S3 file system connector: (yes / **no** / don't know)

## Documentation

  - Does this pull request introduce a new feature? (yes / **no**)
  - If yes, how is the feature documented? (not applicable / docs / **JavaDocs** / not documented)

PR #9505, opened 2019-08-22
[FLINK-13761][datastream] Deprecate SplitStream

zentol:
The split() API was deprecated, but the scala SplitStream class was missed when this was done.


izhangzhihao:
There already have a pr: https://github.com/apache/flink/pull/9474

flink-streaming-scala/src/main/scala/org/apache/flink/streaming/api/scala/SplitStream.scala
```diff
@@ -29,6 +29,7 @@ import org.apache.flink.streaming.api.datastream.{ SplitStream => SplitJavaStrea
  * the appropriate method on this stream.
  */
 @Public
+@deprecated
```
tillrohrmann: Could we add a deprecation message explaining why it shouldn't be used anymore? `@deprecated("reason", "deprecated since.")`

PR #9508, opened 2019-08-22
[FLINK-13817][rest] Expose whether web submissions are enabled

zentol:
Extends the REST API's `/config` endpoint to expose enabled/disabled features. For this PR in particular, the only exposed feature are web-submissions, but we could extend this in the future.


xintongsong:
@TisonKun You can refer to flink-docs/README.md.

TisonKun:
@xintongsong Thanks for your answer :-)

PR #9511, opened 2019-08-22
[FLINK-13356][table][docs] Add documentation for TopN and Deduplication in blink planner

wuchong:

Add documentation for TopN and Deduplication operations which are supported in blink planner. 

## Brief change log

- Add introduction for TopN and Deduplication in `sql.md`.

## Verifying this change
only add docs without test

## Documentation

### DIFF
--- a/docs/dev/table/sql.md
+++ b/docs/dev/table/sql.md
@@ -24,7 +24,7 @@ under the License.
 
 This is a complete list of Data Definition Language (DDL) and Data Manipulation Language (DML) constructs supported in Flink.
 * This will be replaced by the TOC
-{:toc} 
+{:toc}
 
 ## Query
 SQL queries are specified with the `sqlQuery()` method of the `TableEnvironment`. The method returns the result of the SQL query as a `Table`. A `Table` can be used in [subsequent SQL and Table API queries](common.html#mixing-table-api-and-sql), be [converted into a DataSet or DataStream](common.html#integration-with-datastream-and-dataset-api), or [written to a TableSink](common.html#emit-a-table)). SQL and Table API queries can be seamlessly mixed and are holistically optimized and translated into a single program.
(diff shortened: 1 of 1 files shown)


docs/dev/table/sql.md
```diff
@@ -813,6 +813,222 @@ LIMIT 3
 
 {% top %}
 
+### Top-N
+
+TopN is used to calculate the maximum/minimum N records in a stream. It can be flexibly completed based on OVER window aggregation. The grammar is shown as below:
```
JingsongLi: `The grammar is shown as below:`? but below is sentence too? where is the grammar?

docs/dev/table/sql.md
```diff
@@ -813,6 +813,222 @@ LIMIT 3
...
+
+Top-N queries ask for the N smallest or largest values ordered by columns. Both smallest and largest values sets are considered Top-N queries. Top-N queries are useful in cases where the need is to display only the N bottom-most or the N top-
+most records from batch/stream table on a condition. This result set can be used for further analysis.
+
+Flink uses the combination of a OVER window clause and a filter condition to express a Top-N query. With the power of OVER window `PARTITION BY` clause, Flink also supports per group Top-N. For example, the top five products per category that have the maximum sales in realtime. Top-N queries are supported for SQL on batch and streaming tables.
```
JingsongLi: some places is batch/stream, some others are streaming? maybe we just use stream?

docs/dev/table/sql.md
```diff
@@ -813,6 +813,222 @@ LIMIT 3
...
+   SELECT [column_list],
+     ROW_NUMBER() OVER ([PARTITION BY col1[, col2...]]
+       ORDER BY col1 [asc|desc][, col2 [asc|desc]...]) AS rownum
+   FROM table_name)
+WHERE rownum <= N [AND conditions]
+{% endhighlight %}
+
+**Parameter Specification:**
+
+- `ROW_NUMBER()`: Assigns an unique, sequential number to each row, starting with one, according to the ordering of rows within the partition. Currently, we only support `ROW_NUMBER` as the over window function. In the future, we will support `RANK()` and `DENSE_RANK()`.
```
JingsongLi: Explain rowNumber is from 1?

docs/dev/table/sql.md
```diff
@@ -813,6 +813,222 @@ LIMIT 3
...
+      |   SELECT *,
+      |       ROW_NUMBER() OVER (PARTITION BY category ORDER BY sales DESC) as row_num
+      |   FROM ShopSales)
+      |WHERE row_num <= 5
+    """.stripMargin)
+{% endhighlight %}
+</div>
+</div>
+
+<span class="label label-danger">Attention</span> In order to output the above query to an external storage and have a correct result, the external storage must have the same primary key with the Top-N query. In the above example query, if the `product_id` is the primary key of the query, then the external table should also has `product_id` as the primary key.
```
JingsongLi: `if the top N records need to be stored in external storage, the result table must have the same primary key with the Top-N query. By default, the primary key of Top-N query is the combination of partition columns + rownum column.` above.
Confuse here, so what is the real pks of top-n query?

docs/dev/table/sql.md
wuchong: Oh! Good catch! I missed to remove this line.

docs/dev/table/sql.md
wuchong: I think we have done this. "starting with one".

docs/dev/table/sql.md
wuchong: I mean the pk of topn query is (partition columns + rownum column), for example, if partition column is `a`, `b`,  so the pk of topn is the combination of `a`, `b`, `rownum`

docs/dev/table/sql.md
wuchong: I think we also support this in batch? 
The attention part and optimization part is only relative to streaming. I can explain it more detailed.

docs/dev/table/sql.md
JingsongLi: Oh, I see. When the user omit the rownumber, it can't be output directly to the external storage with pk.

JingsongLi: It can be explained that the PK changes caused by optimization and non-optimization are difficult to understand.

leonardBang: > It can be explained that the PK changes caused by optimization and non-optimization are difficult to understand.

+1, the top-N implements with optimization and non-optimization are difficult to understand according to the description.

wuchong:
Thanks @JingsongLi , I will merge this then.

PR #9512, opened 2019-08-22
[FLINK-13819][coordination] Introduce State for RpcEndpoint

azagrebin:

To better reflect the lifecycle of `RpcEndpoint`, we can introduce its running state.
We can use the non-running state e.g. to make decision about how to react on API calls if it is already known that the RpcEndpoint is terminating.

## Brief change log

  - Introduce `RpcEndpoint.isRunning` flag and getter
  - Introduce non-overridable `internalCallOnStart` and `internalCallOnStop` methods in `RpcEndpoint` to set the running flag
  - add unit tests

## Verifying this change

run `RpcEndpointTest`

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/rpc/RpcEndpoint.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/rpc/RpcEndpoint.java
@@ -56,6 +56,36 @@ import static org.apache.flink.util.Preconditions.checkNotNull;
  *
  * <p>The RPC endpoint provides {@link #runAsync(Runnable)}, {@link #callAsync(Callable, Time)}
  * and the {@link #getMainThreadExecutor()} to execute code in the RPC endpoint's main thread.
+ *
+ * <h1>Lifecycle</h1>
+ *
+ * <p>The RPC endpoint has the following stages:
+ * <ul>
+ *    <li>
+ *        The RPC endpoint is created in a non-running state and does not serve any RPC requests.
+ *    </li>
+ *    <li>
+ *        Calling the {@link #start()} method triggers the start of the RPC endpoint and schedules overridable
+ *        {@link #onStart()} method call to the main thread.
+ *    </li>
+ *    <li>
+ *        When the start operation ends the RPC endpoint is moved to the running state
+ *        and starts to serve and complete RPC requests.
+ *    </li>
+ *    <li>
+ *        Calling the {@link #closeAsync()} method triggers the termination of the RPC endpoint and schedules overridable
+ *        {@link #onStop()} method call to the main thread.
+ *    </li>
+ *    <li>
+ *        When {@link #onStop()} method is called, it triggers an asynchronous stop operation.
+ *        The RPC endpoint is not in the running state anymore but it continues to serve RPC requests.
+ *    </li>
+ *    <li>
+ *        When the asynchronous stop operation ends, the RPC endpoint terminates completely
+ *        and does not serve RPC requests anymore.
+ *    </li>
+ * </ul>
+ * The running state can be queried in a RPC method handler or in the main thread by calling {@link #isRunning()} method.
  */
 public abstract class RpcEndpoint implements RpcGateway, AutoCloseableAsync {
 
@@ -78,6 +108,13 @@ public abstract class RpcEndpoint implements RpcGateway, AutoCloseableAsync {
 	/** The main thread executor to be used to execute future callbacks in the main thread
 	 * of the executing rpc server. */
 	private final MainThreadExecutor mainThreadExecutor;
+
+	/**
+	 * Indicates whether the RPC endpoint is started and not stopped or being stopped.
+	 *
+	 * <p>IMPORTANT: the running state is not thread safe and can be used only in the main thread of the rpc endpoint.
+	 */
+	private boolean isRunning;
 
 	/**
 	 * Initializes the RPC endpoint.
@@ -112,12 +149,22 @@ public abstract class RpcEndpoint implements RpcGateway, AutoCloseableAsync {
 		return endpointId;
 	}
 
+	/**
+	 * Returns whether the RPC endpoint is started and not stopped or being stopped.
+	 *
+	 * @return whether the RPC endpoint is started and not stopped or being stopped.
+	 */
+	protected boolean isRunning() {
+		validateRunsInMainThread();
+		return isRunning;
+	}
+
 	// ------------------------------------------------------------------------
 	//  Start & shutdown & lifecycle callbacks
 	// ------------------------------------------------------------------------
 
 	/**
-	 * Starts the rpc endpoint. This tells the underlying rpc server that the rpc endpoint is ready
+	 * Triggers start of the rpc endpoint. This tells the underlying rpc server that the rpc endpoint is ready
 	 * to process remote procedure calls.
 	 *
 	 * @throws Exception indicating that something went wrong while starting the RPC endpoint
@@ -127,20 +174,33 @@ public abstract class RpcEndpoint implements RpcGateway, AutoCloseableAsync {
 	}
 
 	/**
-	 * User overridable callback.
+	 * Internal method which is called by the RpcService implementation to start the RpcEndpoint.
+	 *
+	 * @throws Exception indicating that the rpc endpoint could not be started. If an exception occurs,
+	 * then the rpc endpoint will automatically terminate.
+	 */
+	public final void internalCallOnStart() throws Exception {
+		validateRunsInMainThread();
+		isRunning = true;
+		onStart();
+	}
+
+	/**
+	 * User overridable callback which is called from {@link #internalCallOnStart()}.
 	 *
 	 * <p>This method is called when the RpcEndpoint is being started. The method is guaranteed
 	 * to be executed in the main thread context and can be used to start the rpc endpoint in the
 	 * context of the rpc endpoint's main thread.
 	 *
 	 * <p>IMPORTANT: This method should never be called directly by the user.
+	 *
 	 * @throws Exception indicating that the rpc endpoint could not be started. If an exception occurs,
 	 * then the rpc endpoint will automatically terminate.
 	 */
-	public void onStart() throws Exception {}
+	protected void onStart() throws Exception {}
 
 	/**
-	 * Stops the rpc endpoint. This tells the underlying rpc server that the rpc endpoint is
+	 * Triggers stop of the rpc endpoint. This tells the underlying rpc server that the rpc endpoint is
 	 * no longer ready to process remote procedure calls.
 	 */
 	protected final void stop() {
@@ -148,7 +208,20 @@ public abstract class RpcEndpoint implements RpcGateway, AutoCloseableAsync {
 	}
 
 	/**
-	 * User overridable callback.
+	 * Internal method which is called by the RpcService implementation to stop the RpcEndpoint.
+	 *
+	 * @return Future which is completed once all post stop actions are completed. If an error
+	 * occurs this future is completed exceptionally
+	 */
+	public final CompletableFuture<Void> internalCallOnStop() {
+		validateRunsInMainThread();
+		CompletableFuture<Void> stopFuture = onStop();
+		isRunning = false;
+		return stopFuture;
+	}
+
+	/**
+	 * User overridable callback which is called from {@link #internalCallOnStop()}.
 	 *
 	 * <p>This method is called when the RpcEndpoint is being shut down. The method is guaranteed
 	 * to be executed in the main thread context and can be used to clean up internal state.
@@ -158,7 +231,7 @@ public abstract class RpcEndpoint implements RpcGateway, AutoCloseableAsync {
 	 * @return Future which is completed once all post stop actions are completed. If an error
 	 * occurs this future is completed exceptionally
 	 */
-	public CompletableFuture<Void> onStop() {
+	protected CompletableFuture<Void> onStop() {
 		return CompletableFuture.completedFuture(null);
 	}
 
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/rpc/akka/AkkaRpcActor.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/rpc/akka/AkkaRpcActor.java
@@ -506,7 +506,7 @@ class AkkaRpcActor<T extends RpcEndpoint & RpcGateway> extends AbstractActor {
 
 			CompletableFuture<Void> terminationFuture;
 			try {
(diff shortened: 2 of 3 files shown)


flink-runtime/src/main/java/org/apache/flink/runtime/rpc/RpcEndpoint.java
```diff
@@ -323,6 +355,15 @@ public void validateRunsInMainThread() {
 		assert MainThreadValidatorUtil.isRunningInExpectedThread(currentMainThread.get());
 	}
 
+	/**
+	 * Rpc endpoint's state.
+	 */
+	public enum State {
```
TisonKun: Do we have a context why State in `RpcEndpoint` designed to be `CREATED`, `STARTED` and `STOPPING`?

At least document as javadoc to express what the state exactly means and transitions could be help.

flink-runtime/src/main/java/org/apache/flink/runtime/rpc/RpcEndpoint.java
```diff
@@ -79,6 +79,9 @@
 	 * of the executing rpc server. */
 	private final MainThreadExecutor mainThreadExecutor;
 
+	/** Rpc endpoint's state. */
+	private State state;
```
TisonKun: Maybe `volatile`?

flink-runtime/src/main/java/org/apache/flink/runtime/rpc/RpcEndpoint.java
```diff
@@ -137,18 +153,26 @@ public final void start() {
 	 * @throws Exception indicating that the rpc endpoint could not be started. If an exception occurs,
 	 * then the rpc endpoint will automatically terminate.
 	 */
-	public void onStart() throws Exception {}
+	public final void doStart() throws Exception {
```
TisonKun: Regarding the changes in "AkkaRpcActor.java", why not merge `doStart` into `RpcEndpoint#onStart`? We confuse further reader(maybe ourselves) that we have `start` `doStart` and `onStart`.

flink-runtime/src/main/java/org/apache/flink/runtime/rpc/RpcEndpoint.java
```diff
@@ -323,6 +355,15 @@ public void validateRunsInMainThread() {
 		assert MainThreadValidatorUtil.isRunningInExpectedThread(currentMainThread.get());
 	}
 
+	/**
+	 * Rpc endpoint's state.
+	 */
+	public enum State {
+		CREATED,
+		STARTED,
+		STOPPING
```
zentol: What's the reasoning for `STOPPING`, set at the beginning of the stop operation, opposed to `STOPPED`, set after the stop operation?

flink-runtime/src/main/java/org/apache/flink/runtime/rpc/RpcEndpoint.java
azagrebin: `doStart` is supposed to be called only from `AkkaRpcActor` and never overriden because it does internal staff like modifying the state.
`onStart` is an abstract method to override and do endpoint specific actions.
This way we make sure that the state change is not overriden and not lost.

flink-runtime/src/main/java/org/apache/flink/runtime/rpc/RpcEndpoint.java
azagrebin: As the getter says the state is supposed to be read/written only in the main thread of the endpoint then there is no need for the synchronisation and I want to avoid confusion but adding volatile.

flink-runtime/src/main/java/org/apache/flink/runtime/rpc/RpcEndpoint.java
azagrebin: The states basically reflect the lifecycle of the RPC Endpoint (as written in JIRA) and what one can do with it by calling the available methods: start/stop/close and getting callbacks: onStart and onStop.
Then the concrete endpoints can easier judge about the state in concrete API calls (also in JIRA).
I can add a jdoc comment about when the transitions happen.

flink-runtime/src/main/java/org/apache/flink/runtime/rpc/RpcEndpoint.java
TisonKun: Thanks for the clarification. I see the document below. We can also document it here or use a somewhat `@GuardBy` annotation. Although I'm not sure how to express guarded by MainThread with `@GuardBy`.

flink-runtime/src/main/java/org/apache/flink/runtime/rpc/RpcEndpoint.java
TisonKun: How about inline state transition in `AkkaRpcActor`? If only we can reduce these `start`s.

flink-runtime/src/main/java/org/apache/flink/runtime/rpc/RpcEndpoint.java
azagrebin: Good question. The start/stop happen asynchronously.
From what I understand, the main thread is ready to operate at the end of the `doStart`.
But `onStop` can additionally trigger some extra asynchronous actions. Only after they are done the endpoint is fully stopped until then it continues to operate but it is known that it is being stopped.
Not sure whether we need more states to reflect the lifecycle better. 
@tillrohrmann probably could help with the input here.

flink-runtime/src/main/java/org/apache/flink/runtime/rpc/RpcEndpoint.java
TisonKun: Thanks for your reply. 

From JIRA it says "we suggest to introduce its state:..." without why the exactly states.

It would be helpful if you write down the transitions. As @zentol's comment, you can find `STOPPING` is not yet well-defined.

Let's say we prefer reaching a consensus in JIRA first on which states we will introduce, accurate meaning of state, and when transition happens. Here is one way to implement it and maybe it'd better to discuss the requirement on JIRA first.

flink-runtime/src/main/java/org/apache/flink/runtime/rpc/RpcEndpoint.java
azagrebin: Not sure we have something like that, I can only duplicate the comment from the getter here.
AFAIK, it is our usual implicit assumption about non-final variables in endpoints that they are confined to the main thread.

flink-runtime/src/main/java/org/apache/flink/runtime/rpc/RpcEndpoint.java
azagrebin: Well, the state transition belongs to the `RpcEndpoint`, not `AkkaRpcActor`.

flink-runtime/src/main/java/org/apache/flink/runtime/rpc/RpcEndpoint.java
azagrebin: It says `To better reflect the lifecycle of RpcEndpoint..`
I will write it down after we fully confirm them here. I expect a review from @tillrohrmann as we agreed offline.

flink-runtime/src/main/java/org/apache/flink/runtime/rpc/RpcEndpoint.java
TisonKun: Make sense.

flink-runtime/src/main/java/org/apache/flink/runtime/rpc/RpcEndpoint.java
```diff
@@ -112,12 +117,23 @@ public String getEndpointId() {
...
 	}
 
+	/**
+	 * Returns the rpc endpoint's state.
+	 *
+	 * <p>IMPORTANT: the state is not thread safe and can be used only in the main thread of the rpc endpoint.
+	 *
+	 * @return Rpc endpoint's state.
+	 */
+	public State getState() {
```
tillrohrmann: Instead of exposing the internal state machine of the `RpcEndpoint` via `State`, I think it would be better to define a method `public final boolean isRunning()` which returns `true` if the state is `Running` and if not, then it returns `false`. This method could validate that it is only called from the main thread by calling `validateRunsInMainThread`.

flink-runtime/src/main/java/org/apache/flink/runtime/rpc/RpcEndpoint.java
```diff
@@ -137,18 +153,26 @@ public final void start() {
...
 	 * then the rpc endpoint will automatically terminate.
 	 */
-	public void onStart() throws Exception {}
+	public final void doStart() throws Exception {
+		state = State.STARTED;
+		onStart();
+	}
+
+	/**
+	 * User overridable callback called from {@link #doStart()}.
```
tillrohrmann:
```suggestion
	 * User overridable callback which is called from {@link #doStart()}.
```

flink-runtime/src/main/java/org/apache/flink/runtime/rpc/RpcEndpoint.java
tillrohrmann: This method should be guarded by `validateRunsInMainThread`

flink-runtime/src/main/java/org/apache/flink/runtime/rpc/RpcEndpoint.java
```diff
@@ -158,7 +182,15 @@ protected final void stop() {
 	 * @return Future which is completed once all post stop actions are completed. If an error
 	 * occurs this future is completed exceptionally
 	 */
-	public CompletableFuture<Void> onStop() {
+	public final CompletableFuture<Void> doStop() {
```
tillrohrmann: Better to guard via `validateRunsInMainThread`

flink-runtime/src/main/java/org/apache/flink/runtime/rpc/RpcEndpoint.java
tillrohrmann: If we only expose `isRunning`, then we could simplify the state machine to `boolean isRunning` which is set to `true` in `doStart` and to `false` in `doStop`.

flink-runtime/src/main/java/org/apache/flink/runtime/rpc/RpcEndpoint.java
tillrohrmann: Maybe let's rename this method into `internalCallOnStart` so that users see that this method is an internal method. It is a bit unfortunate that with Java's accessors we are not able to restrict the accessibility of this method to only the `AkkaRpcActor`.

flink-runtime/src/test/java/org/apache/flink/runtime/rpc/akka/AkkaRpcActorTest.java
```diff
@@ -609,6 +611,7 @@ public void onStart() throws Exception {
 
 		public void awaitUntilOnStartCalled() throws InterruptedException {
 			countDownLatch.await();
+			assertThat(getState(), is(State.STARTED));
```
tillrohrmann: Please don't piggy back on existing tests to check for the newly introduced functionality. It is usually a better idea to add targeted tests for the new behaviour.

flink-runtime/src/main/java/org/apache/flink/runtime/rpc/RpcEndpoint.java
```diff
@@ -127,7 +143,7 @@ public final void start() {
 	}
 
 	/**
-	 * User overridable callback.
+	 * Starts the rpc endpoint.
```
tillrohrmann: This JavaDoc is misleading because the user might think that he needs to call this method to start the `RpcEndpoint`. Better to say `Internal method which is called by the RpcService implementation to start the RpcEndpoint` or so.

flink-runtime/src/main/java/org/apache/flink/runtime/rpc/RpcEndpoint.java
```diff
@@ -137,18 +153,26 @@ public final void start() {
...
 	 */
-	public void onStart() throws Exception {}
+	public final void doStart() throws Exception {
+		state = State.STARTED;
+		onStart();
+	}
+
+	/**
+	 * User overridable callback called from {@link #doStart()}.
+	 */
```
tillrohrmann: `@throws` clause missing

flink-runtime/src/main/java/org/apache/flink/runtime/rpc/RpcEndpoint.java
```diff
@@ -137,18 +153,26 @@ public final void start() {
 	 * @throws Exception indicating that the rpc endpoint could not be started. If an exception occurs,
 	 * then the rpc endpoint will automatically terminate.
```
tillrohrmann: The two paragraphs and the `@throws` clause should go to the `onStart` method because that's what users will override.

azagrebin:
Thanks for the reviews @tillrohrmann @TisonKun 
I have updated the PR

flink-runtime/src/main/java/org/apache/flink/runtime/rpc/RpcEndpoint.java
```diff
@@ -112,12 +149,22 @@ public String getEndpointId() {
 		return endpointId;
 	}
 
+	/**
+	 * Returns whether the RPC endpoint is started and not stopped or being stopped.
+	 *
+	 * @return whether the RPC endpoint is started and not stopped or being stopped.
+	 */
+	public boolean isRunning() {
```
tillrohrmann: let's make this protected

flink-runtime/src/main/java/org/apache/flink/runtime/rpc/RpcEndpoint.java
```diff
@@ -127,28 +174,53 @@ public final void start() {
...
 	/**
-	 * User overridable callback.
+	 * Internal method which is called by the RpcService implementation to stop the RpcEndpoint.
+	 *
+	 * @return Future which is completed once all post stop actions are completed. If an error
+	 * occurs this future is completed exceptionally
+	 */
+	public final CompletableFuture<Void> internalCallOnStop() {
+		validateRunsInMainThread();
+		isRunning = false;
```
tillrohrmann: I would move it after the `onStop` to make it symmetric with the `onStart`

azagrebin:
Thanks @tillrohrmann @TisonKun
Merging the PR when the Travis is green

azagrebin:
merged into master by aabde88e5280d9fef9c911265e270ebb101b58e4

PR #9516, opened 2019-08-23
[FLINK-13815][State Backends] Implement the SpaceAllocator

carp84:

Adding allocators for off-heap byte buffer allocation, which is required by storing state on disk.

## Brief change log

Mainly introduces two types of buffer allocators: `DirectBucketAllocator` and `PowerTwoBucketAllocator`. The former is for relatively big data, and the latter is a simple buddy-like allocator for small data. Also assures java 9+ compatibility referring to Hadoop's [CleanerUtil](https://github.com/apache/hadoop/blob/trunk/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/util/CleanerUtil.java).

## Verifying this change

This change added tests under `org.apache.flink.runtime.state.heap.space` package.

  - The serializers: (no)

## Documentation

     - Added a [design doc](https://drive.google.com/file/d/1xdOLDnRCBLEP0xrlqydFAN1BrBIkokFb/view?usp=sharing) for `PowerTwoBucketAllocator`


flink-state-backends/flink-statebackend-heap-spillable/pom.xml
```diff
@@ -37,5 +37,20 @@ under the License.
...
+		<!-- core dependencies -->
+
+		<dependency>
+			<groupId>org.apache.flink</groupId>
+			<artifactId>flink-streaming-java_${scala.binary.version}</artifactId>
+			<version>${project.version}</version>
+			<scope>provided</scope>
+		</dependency>
+
+                <!-- test dependencies -->
```
StephanEwen: Tabs / spaces

flink-state-backends/flink-statebackend-heap-spillable/src/main/java/org/apache/flink/runtime/state/heap/space/ChunkAllocator.java
```diff
@@ -0,0 +1,64 @@
...
+	 * @param chunkId id of the chunk.
+	 * @param allocateStrategy strategy to allocate chunk.
+	 * @return a chunk.
+	 */
+	Chunk createChunk(int chunkId, AllocateStrategy allocateStrategy);
+
+	/**
+	 * Type of space where to allocate chunks.
+	 */
+	enum SpaceType {
```
StephanEwen: Maybe reuse `MemoryType` here?

https://github.com/apache/flink/blob/master/flink-core/src/main/java/org/apache/flink/core/memory/MemoryType.java

flink-state-backends/flink-statebackend-heap-spillable/src/main/java/org/apache/flink/runtime/state/heap/space/CleanerUtil.java
```diff
@@ -0,0 +1,182 @@
...
+/**
+ * Direct buffer cleaner tool, mainly copied from Hadoop project (see org.apache.hadoop.util.CleanerUtil).
+ *
+ * <p/>sun.misc.Cleaner has moved in OpenJDK 9 and
+ * sun.misc.Unsafe#invokeCleaner(ByteBuffer) is the replacement.
+ * This class is a hack to use sun.misc.Cleaner in Java 8 and
+ * use the replacement in Java 9+.
+ * This implementation is inspired by LUCENE-6989.
+ */
+final class CleanerUtil {
```
StephanEwen: What is this class used for? I could not find a reference to it in the code.

The `MemorySegment` with unsafe memory allocation also has a cleaner solution already - would that cover the use cases originally handled by this util?

flink-state-backends/flink-statebackend-heap-spillable/src/test/resources/log4j-test.properties
```diff
@@ -0,0 +1,32 @@
...
+
+
+# testlogger is set to be a ConsoleAppender.
+log4j.appender.testlogger=org.apache.log4j.ConsoleAppender
+log4j.appender.testlogger.target=System.err
+log4j.appender.testlogger.layout=org.apache.log4j.PatternLayout
+log4j.appender.testlogger.layout.ConversionPattern=%-4r [%t] %-5p %c %x - %m%n
+
+# Suppress the irrelevant (wrong) warnings from the Netty channel handler
+log4j.logger.org.apache.flink.shaded.akka.org.jboss.netty.channel.DefaultChannelPipeline=ERROR, testlogger
```
StephanEwen: Do any of the ITCases spawn a full Flink? If not, we could remove these logger configuration lines.

flink-state-backends/flink-statebackend-heap-spillable/src/main/java/org/apache/flink/runtime/state/heap/space/SpaceConstants.java
```diff
@@ -0,0 +1,37 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.runtime.state.heap.space;
+
+/**
+ * Some constants used for space.
+ */
+public class SpaceConstants {
```
StephanEwen: Minor: We often give util classes with static methods only a private constructor, do prevent instantiation.

flink-state-backends/flink-statebackend-heap-spillable/pom.xml
carp84: The changes on pom.xml are no more needed with PR #9501 in

flink-state-backends/flink-statebackend-heap-spillable/src/main/java/org/apache/flink/runtime/state/heap/space/CleanerUtil.java
carp84: Yes, there were some more clean up work to do after #9501 and this is fixed in the updated commit.

flink-state-backends/flink-statebackend-heap-spillable/src/test/resources/log4j-test.properties
carp84: Yes we plan to also add spillable as a new backend type in `EventTimeWindowCheckpointingITCase`, `LocalRecoveryITCase` and `StreamOperatorSnapshotRestoreTest`.

flink-state-backends/flink-statebackend-heap-spillable/src/main/java/org/apache/flink/runtime/state/heap/space/ChunkAllocator.java
carp84: The major type of chunk/space allocator we plan to use in spill-able backend is the memory-mapped file, and we will introduce the `FILE` type when implementing `MmapManager`, so it actually differs with `MemoryType`.

The existing `SpaceType` truly looks like `MemoryType`, but let me explain different meanings of the existing types:

- `HEAP` is just for unit testing

- `OFFHEAP` is reserved if user would like to use memory instead of disk for the spilled part (serialized data, less memory consumption, won't affect GC, and better control comparing to page cache for mmap)

PR #9517, opened 2019-08-23
[FLINK-13779][metrics] PrometheusPushGatewayReporter support push metrics with groupingKey

kaibozhou:

The pushgateway already support send metrics with [groupingKey](https://prometheus.github.io/client_java/io/prometheus/client/exporter/PushGateway.html#push-io.prometheus.client.CollectorRegistry-java.lang.String-java.util.Map-), PrometheusPushGatewayReporter can support push metrics with groupingKey, which will be convenient for users to identify, group or filter their metrics.

## Brief change log

  - *parser groupingKey from user's config: k1=v1,k2=v2 to Map<String, String>*
  - *Use PushGateway Java SDK to push with groupingKey*
  - *Use PushGateway Java SDK to delete with groupingKey**

## Verifying this change
Run *PrometheusPushGatewayReporterTest* to verify the groupingKey config.

  - The serializers: (no)

## Documentation

  - If yes, how is the feature documented? (docs will auto generated)


flink-metrics/flink-metrics-prometheus/src/test/java/org/apache/flink/metrics/prometheus/PrometheusPushGatewayReporterTest.java
```diff
@@ -0,0 +1,55 @@
...
+		Map<String, String> groupingKey = PrometheusPushGatewayReporter.parserGroupingKey("k1=");
+		assert groupingKey != null;
+		Assert.assertEquals("", groupingKey.get("k1"));
+
+		groupingKey = PrometheusPushGatewayReporter.parserGroupingKey("=v1");
+		assert groupingKey != null;
+		Assert.assertEquals("v1", groupingKey.get(""));
+
+		groupingKey = PrometheusPushGatewayReporter.parserGroupingKey("k1");
+		assert groupingKey != null;
```
StephanEwen: Replace with `Assert.assertNotNull`

flink-metrics/flink-metrics-prometheus/src/main/java/org/apache/flink/metrics/prometheus/PrometheusPushGatewayReporter.java
```diff
@@ -67,25 +73,44 @@ public void open(MetricConfig config) {
...
+			return groupingKey;
+		}
+		return null;
 	}
 
 	@Override
 	public void report() {
 		try {
-			pushGateway.push(CollectorRegistry.defaultRegistry, jobName);
+			pushGateway.push(CollectorRegistry.defaultRegistry, jobName, groupingKey);
```
zentol: is there a difference in behavior between passing null and an empty collection?

flink-metrics/flink-metrics-prometheus/src/main/java/org/apache/flink/metrics/prometheus/PrometheusPushGatewayReporter.java
```diff
@@ -67,25 +73,44 @@ public void open(MetricConfig config) {
 		}
 
 		pushGateway = new PushGateway(host + ':' + port);
-		log.info("Configured PrometheusPushGatewayReporter with {host:{}, port:{}, jobName: {}, randomJobNameSuffix:{}, deleteOnShutdown:{}}", host, port, jobName, randomSuffix, deleteOnShutdown);
+		log.info("Configured PrometheusPushGatewayReporter with {host:{}, port:{}, jobName:{}, randomJobNameSuffix:{}, deleteOnShutdown:{}, groupingKey:{}}",
+			host, port, jobName, randomSuffix, deleteOnShutdown, groupingKey);
+	}
+
+	@VisibleForTesting
+	static Map<String, String> parserGroupingKey(final String groupingKeyConfig) {
```
zentol: rename to `parseGroupingKey`

flink-metrics/flink-metrics-prometheus/src/main/java/org/apache/flink/metrics/prometheus/PrometheusPushGatewayReporter.java
```diff
@@ -67,25 +73,44 @@ public void open(MetricConfig config) {
...
+		if (!groupingKeyConfig.isEmpty()) {
+			Map<String, String> groupingKey = new HashMap<>();
+			String[] kvs = groupingKeyConfig.split(",");
+			for (String kv : kvs) {
+				int idx = kv.indexOf("=");
+				if (idx < 0) {
+					continue;
```
zentol: this cast implies an invalid configuration and should be logged as a warning

zentol:
> Could you explain a bit more what the groupingKey actually accomplishes? Is it just a global tag that is applied to all metrics?

To clarify, if this is all the groupingKey is then it shouldn't implemented this way; there's no reason to make this a feature specific this reporter.

flink-metrics/flink-metrics-prometheus/src/main/java/org/apache/flink/metrics/prometheus/PrometheusPushGatewayReporter.java
kaibozhou: I did some test and see the code of pushGateway SDK, both null and an empty collection will be ignored.

kaibozhou:
Hi, @StephanEwen @zentol  
Thanks for your review. The groupingKey is just a global tag that applied to all metrics. And if users do not specify groupingKey, the reporter will pass null which will be ignored by pushGateway. So it has no impact on the way of using the push gateway without grouping keys.

kaibozhou:
All comments addressed, and I add a detailed doc description.

Sorry for the misunderstood about the doc, I thought it would be generated automatically -)

StephanEwen:
@zentol The way I understand this, it is simply some labels that can be attached to any set of metrics that are pushed so that promotheus can internally store the metrics under these group labels. Seems very prometheus specific.

kaibozhou:
@flinkbot  run travis

flink-metrics/flink-metrics-prometheus/src/main/java/org/apache/flink/metrics/prometheus/PrometheusPushGatewayReporter.java
zentol: In that case `parseGroupingKey` should not return null but an empty map instead.

flink-metrics/flink-metrics-prometheus/src/main/java/org/apache/flink/metrics/prometheus/PrometheusPushGatewayReporterOptions.java
```diff
@@ -63,4 +63,14 @@
...
 			.build());
+
+	public static final ConfigOption<String> GROUPING_KEY = ConfigOptions
+		.key("groupingKey")
+		.defaultValue("")
+			.withDescription(Description.builder()
+				.text("Specifies the grouping key which is the group and global labels of all metrics." +
+					" The label name and value is separated with '=', and labels are separated with ';', e.g. k1=v1;k2=v2." +
+					" Please ensure that your grouping key meet the %s.",
+					LinkElement.link("https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels", "Prometheus metric labels requirements"))
```
zentol: The options just above this one links to the same page, but has a different link name. I'd prefer them to be consistent.

flink-metrics/flink-metrics-prometheus/src/test/java/org/apache/flink/metrics/prometheus/PrometheusPushGatewayReporterTest.java
```diff
@@ -0,0 +1,57 @@
...
+		Assert.assertEquals("v1", groupingKey.get("k1"));
+		Assert.assertEquals("v2", groupingKey.get("k2"));
+	}
+
+	@Test
+	public void testParseIncompleteGroupingKey() {
+		PrometheusPushGatewayReporter reporter = new PrometheusPushGatewayReporter();
+		Map<String, String> groupingKey = reporter.parseGroupingKey("k1=");
+		Assert.assertNotNull(groupingKey);
+		Assert.assertEquals("", groupingKey.get("k1"));
```
zentol: we shouldn't allow empty labels, from what I've gathered they aren't particularly useful.

https://github.com/prometheus/prometheus/issues/3706

flink-metrics/flink-metrics-prometheus/src/test/java/org/apache/flink/metrics/prometheus/PrometheusPushGatewayReporterTest.java
```diff
@@ -0,0 +1,57 @@
...
+
+		groupingKey = reporter.parseGroupingKey("=v1");
+		Assert.assertNotNull(groupingKey);
+		Assert.assertEquals("v1", groupingKey.get(""));
```
zentol: Does prometheus actually accept such a grouping key?

flink-metrics/flink-metrics-prometheus/src/main/java/org/apache/flink/metrics/prometheus/PrometheusPushGatewayReporterOptions.java
```diff
@@ -63,4 +63,14 @@
...
 				" label values meet the %s.", LinkElement.link("https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels", "Prometheus requirements"))
 			.build());
+
+	public static final ConfigOption<String> GROUPING_KEY = ConfigOptions
+		.key("groupingKey")
+		.defaultValue("")
+			.withDescription(Description.builder()
+				.text("Specifies the grouping key which is the group and global labels of all metrics." +
+					" The label name and value is separated with '=', and labels are separated with ';', e.g. k1=v1;k2=v2." +
+					" Please ensure that your grouping key meet the %s.",
```
zentol:
```suggestion
					" Please ensure that your grouping key meets the %s.",
```

flink-metrics/flink-metrics-prometheus/src/main/java/org/apache/flink/metrics/prometheus/PrometheusPushGatewayReporterOptions.java
```diff
@@ -63,4 +63,14 @@
 				" Before disabling this option please ensure that your" +
 				" label values meet the %s.", LinkElement.link("https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels", "Prometheus requirements"))
 			.build());
+
+	public static final ConfigOption<String> GROUPING_KEY = ConfigOptions
+		.key("groupingKey")
+		.defaultValue("")
+			.withDescription(Description.builder()
+				.text("Specifies the grouping key which is the group and global labels of all metrics." +
+					" The label name and value is separated with '=', and labels are separated with ';', e.g. k1=v1;k2=v2." +
```
zentol:
```suggestion
					" The label name and value are separated by '=', labels are separated by ';', e.g., k1=v1;k2=v2." +
```

Bonus points for replacing the example with a `code` element.

flink-metrics/flink-metrics-prometheus/src/test/java/org/apache/flink/metrics/prometheus/PrometheusPushGatewayReporterTest.java
kaibozhou: Agree. Although the empty label key and value will not throw an exception, it does not make sense. So I will make a check to ignore empty labels.

flink-metrics/flink-metrics-prometheus/src/test/java/org/apache/flink/metrics/prometheus/PrometheusPushGatewayReporterTest.java
kaibozhou: I think we can ignore the empty key, it does not make sense.

flink-metrics/flink-metrics-prometheus/src/main/java/org/apache/flink/metrics/prometheus/PrometheusPushGatewayReporterOptions.java
kaibozhou: Do you mean e.g., < code > k1=v1;k2=v2 < /code >. ? I found the generated doc display it as plain text. Or if I have a mistake with it?

kaibozhou:
@zentol Thanks for your review.  Comments addressed.

flink-metrics/flink-metrics-prometheus/src/main/java/org/apache/flink/metrics/prometheus/PrometheusPushGatewayReporterOptions.java
zentol: You have to use `TextElement#code` for that, works just like adding a link.

flink-metrics/flink-metrics-prometheus/src/main/java/org/apache/flink/metrics/prometheus/PrometheusPushGatewayReporter.java
```diff
@@ -83,15 +85,22 @@ public void open(MetricConfig config) {
...
-					log.warn("Invalid prometheusPushGateway groupingKey {}, will be ignored", kv);
+					log.warn("Invalid prometheusPushGateway groupingKey:{}, will be ignored", kv);
 					continue;
 				}
-				groupingKey.put(kv.substring(0, idx), kv.substring(idx + 1));
+
+				String labelKey = kv.substring(0, idx);
+				String labelValue = kv.substring(idx + 1);
+				if (StringUtils.isNullOrWhitespaceOnly(labelKey) || StringUtils.isNullOrWhitespaceOnly(labelValue)) {
+					log.warn("Invalid groupingKey {labelKey:{}, labelValue:{}} will be ignored", labelKey, labelValue);
```
zentol: Explicitly say that name/value must not be empty

flink-metrics/flink-metrics-prometheus/src/main/java/org/apache/flink/metrics/prometheus/PrometheusPushGatewayReporterOptions.java
```diff
@@ -63,4 +63,14 @@
 				" Before disabling this option please ensure that your" +
 				" label values meet the %s.", LinkElement.link("https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels", "Prometheus requirements"))
 			.build());
+
+	public static final ConfigOption<String> GROUPING_KEY = ConfigOptions
+		.key("groupingKey")
+		.defaultValue("")
+			.withDescription(Description.builder()
+				.text("Specifies the grouping key which is the group and global labels of all metrics." +
+					" The label name and value are separated with '=', and labels are separated by ';', e.g., k1=v1;k2=v2." +
```
zentol:
```suggestion
					" The label name and value are separated by '=', and labels are separated by ';', e.g., k1=v1;k2=v2." +
```

kaibozhou:
Thanks for the careful review.  @zentol 
Comments address.

zentol:
@flinkbot run travis

PR #9520, opened 2019-08-23
[hotfix][docs] wrong method name in state-processor-api example

alpinegizmo:

The State Processor API docs have an example of implementing a KeyedStateReaderFunction. The wrong method name is used in this example -- the class has no processKey method, but rather a readKey method.

## Verifying this change

This change is a trivial typo fix.


1u0:
@flinkbot run travis

StephanEwen:
Thank you for the fixing this!

Looks good, merging...

PR #9527, opened 2019-08-24
[FLINK-13832] Rename DefaultRollingPolicy create to builder

gyfora:

Deprecate and rename DefaultRollingPolicy create() method to builder()

## Brief change log

  - Renamed and deprecated old method
  - Replaced existing usage
  - Added test to validate deprecated method still works

## Verifying this change

This change is a trivial rework / code cleanup with minimum test coverage for the deprecated method

  - The serializers: no

## Documentation

  - If yes, how is the feature documented?  JavaDocs

### DIFF
--- a/flink-streaming-java/src/main/java/org/apache/flink/streaming/api/functions/sink/filesystem/rollingpolicies/DefaultRollingPolicy.java
+++ b/flink-streaming-java/src/main/java/org/apache/flink/streaming/api/functions/sink/filesystem/rollingpolicies/DefaultRollingPolicy.java
@@ -83,10 +83,34 @@ public final class DefaultRollingPolicy<IN, BucketID> implements RollingPolicy<I
 	}
 
 	/**
-	 * Initiates the instantiation of a {@code DefaultRollingPolicy}.
-	 * To finalize it and have the actual policy, call {@code .create()}.
+	 * Returns the maximum part file size before rolling.
+	 * @return Max size in bytes
 	 */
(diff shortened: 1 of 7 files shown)


StephanEwen:
It would be nice to have a project-wide standard for how we call the "builder creator" method.

Shall we [DISCUSS] that and add it to the code style guide?

gyfora:
Sounds like a good approach @StephanEwen , I haven't checked through the project but this one felt very much off but having standards would be much better. I will open a discussion thread on the ml

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/functions/sink/filesystem/rollingpolicies/DefaultRollingPolicy.java
```diff
@@ -83,18 +83,51 @@ public boolean shouldRollOnProcessingTime(final PartFileInfo<BucketID> partFileS
...
+	 */
+	public static DefaultRollingPolicy.PolicyBuilder builder() {
 		return new DefaultRollingPolicy.PolicyBuilder(
 				DEFAULT_MAX_PART_SIZE,
 				DEFAULT_ROLLOVER_INTERVAL,
 				DEFAULT_INACTIVITY_INTERVAL);
 	}
 
+	/**
+	 * This method is deprecated, use {@link DefaultRollingPolicy#builder()} instead.
```
kl0u: Use the `@deprecated` annotation in the javadoc.

PR #9534, opened 2019-08-26
[FLINK-13853][e2e] Update common_ha.sh test expression to count recoveries

tillrohrmann:

With FLINK-13573 we removed the SubmittedJobGraph and replaced it with the JobGraph.
Consequently, we no longer see the log statement "Recovered SubmittedJobGraph" which
was used by the common_ha.sh script to count the number of job recoveries. Now we need
to look for "Recovered JobGraph" instead.

## Verifying this change

- Manually tested that failing e2e test passes

  - The serializers: (no)

## Documentation


tillrohrmann:
Thanks for the review @TisonKun. The problem is that we don't run the cron jobs which run the e2e tests with every commit. Hence, we only saw the failure later.

Merging this PR.

PR #9541, opened 2019-08-27
[FLINK-13828][configuration] Deprecate ConfigConstants.LOCAL_START_WEBSERVER

TisonKun:

Maybe backport to 1.9.1 because I heard this issue from a user of 1.9.0

Exactly ConfigConstants.LOCAL_START_WEBSERVER has no power any more but we don't mark it as deprecated or give an investigate to revive it. Since it is part of public interface we cannot remove it in minor version but deprecate it is necessary.

## Brief change log

- Remove use points of `ConfigConstants.LOCAL_START_WEBSERVER`
- Deprecate `ConfigConstants.LOCAL_START_WEBSERVER`
- hotfix for test style in `WebFrontendITCase`

## Verifying this change

This change is a trivial rework / code cleanup without any test coverage

  - The public API, i.e., is any changed class annotated with `@Public(Evolving)`: (yes, mark a stale field as deprecated.)
  - The serializers: (no)

## Documentation

cc @twalthr

### DIFF
--- a/flink-core/src/main/java/org/apache/flink/configuration/ConfigConstants.java
+++ b/flink-core/src/main/java/org/apache/flink/configuration/ConfigConstants.java
@@ -1866,6 +1866,10 @@ public final class ConfigConstants {
 	@Deprecated
 	public static final int DEFAULT_LOCAL_NUMBER_RESOURCE_MANAGER = 1;
 
+	/**
+	 * @deprecated Has no effect; the web-server is always started. Will be removed in 2.0.
+	 */
+	@Deprecated
 	public static final String LOCAL_START_WEBSERVER = "local.start-webserver";
 
 	// --------------------------- High Availability ---------------------------------
--- a/flink-runtime-web/src/test/java/org/apache/flink/runtime/webmonitor/WebFrontendITCase.java
+++ b/flink-runtime-web/src/test/java/org/apache/flink/runtime/webmonitor/WebFrontendITCase.java
@@ -115,15 +115,10 @@ public class WebFrontendITCase extends TestLogger {
 	}
 
 	@Test
-	public void getFrontPage() {
-		try {
-			String fromHTTP = TestBaseUtils.getFromHTTP("http://localhost:" + getRestPort() + "/index.html");
-			String text = "Apache Flink Web Dashboard";
-			assertTrue("Startpage should contain " + text, fromHTTP.contains(text));
-		} catch (Exception e) {
-			e.printStackTrace();
-			fail(e.getMessage());
-		}
+	public void getFrontPage() throws Exception {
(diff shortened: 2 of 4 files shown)


flink-core/src/main/java/org/apache/flink/configuration/ConfigConstants.java
```diff
@@ -1866,6 +1866,10 @@
 	@Deprecated
 	public static final int DEFAULT_LOCAL_NUMBER_RESOURCE_MANAGER = 1;
 
+	/**
+	 * Ineffective any more. Will be removed in 2.0.
```
zentol:
```suggestion
	 * @deprecated Has no effect; the web-server is always started. Will be removed in 2.0.
```

flink-runtime-web/src/test/java/org/apache/flink/runtime/webmonitor/WebFrontendITCase.java
```diff
@@ -209,45 +198,37 @@ public void getLogAndStdoutFiles() throws Exception {
...
 
-			Map<String, String> conf = WebMonitorUtils.fromKeyValueJsonArray(config);
-			assertEquals(
-				CLUSTER_CONFIGURATION.getString(ConfigConstants.LOCAL_START_WEBSERVER, null),
-				conf.get(ConfigConstants.LOCAL_START_WEBSERVER));
-		} catch (Exception e) {
-			e.printStackTrace();
-			fail(e.getMessage());
-		}
+		String expected = CLUSTER_CONFIGURATION.getString(TaskManagerOptions.MANAGED_MEMORY_SIZE);
```
zentol: Please add a comment to the setup section pointing out that setting this value is required for the validity of some tests.

flink-runtime-web/src/test/java/org/apache/flink/runtime/webmonitor/WebFrontendITCase.java
```diff
@@ -209,45 +198,37 @@ public void getLogAndStdoutFiles() throws Exception {
...
+		String actual = conf.get(TaskManagerOptions.MANAGED_MEMORY_SIZE.key());
+
+		assertNotNull(actual);
```
zentol: seems unnecessary, assertEquals also handles this case.

TisonKun: I think this prevent the case `null == null` but if we add comment above and assume we always set `TaskManagerOptions.MANAGED_MEMORY_SIZE` I agree with can remove this one.

zentol:
@TisonKun Could you squash your last commit (61306fa) into the first one (933c1ed)? I would then merge it afterwards.

TisonKun:
@zentol Thanks for your review. Updated.

PR #9547, opened 2019-08-27
[FLINK-13877][hive] Support Hive version 2.1.0 and 2.1.1

xuefuz:

Support  Hive 2.1.x versions (2.1.0 and 2.1.1).

## Brief change log

  - Added new shims for each version
  - Adjusted shim inheritance heirarchy
  - Relocated one shimmed method

## Verifying this change

This change is already covered by existing tests, and manually ran the tests for all versions.

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/client/HiveShimV210.java
+++ b/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/client/HiveShimV210.java
@@ -1,0 +1,76 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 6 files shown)


flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/client/HiveShimV210.java
```diff
@@ -0,0 +1,76 @@
...
+ * Shim for Hive version 2.1.0.
+ */
+public class HiveShimV210 extends HiveShimV122 {
+
+	@Override
+	public IMetaStoreClient getHiveMetastoreClient(HiveConf hiveConf) {
+		try {
+			Class<?>[] constructorArgTypes = {HiveConf.class};
+			Object[] constructorArgs = {hiveConf};
+			Method method = RetryingMetaStoreClient.class.getMethod("getProxy", HiveConf.class,
```
lirui-apache: Can we call `public static IMetaStoreClient getProxy(HiveConf hiveConf, boolean allowEmbedded)` instead? Then `HiveShimV230` probably doesn't have to implement its own `getHiveMetastoreClient `?

bowenli86:
can we add test profiles for new Hive versions? 

I’d be more comfortable if we can at least run them manually.

xuefuz:
Added profile. @bowenli86 Can you take another look?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/table/catalog/hive/client/HiveShimV210.java
lirui-apache: Thanks for the details. Maybe we can add some comments for that?

PR #5031, opened 2017-11-17
[FLINK-8099] Reduce default restart delay to 1 second

aljoscha:
R: @tillrohrmann 

We could also introduce an extra entry in `ConfigConstants` for the `"1 s"` default but I opted against that. What do you think?

PR #9552, opened 2019-08-28
[FLINK-13884] Set default failure rate restart strategy delay to 0s

tillrohrmann:

Set default failure rate restart strategy delay to `0 s`.

This PR breaks the public evolving `RESTART_STRATEGY_FAILURE_RATE_DELAY` configuration by changing its type from `String` to `ConfigOption`. I believe that this is ok given that it is public evolving. We should add a release note when closing the JIRA issue, though.

This PR is based on #9562.

## Verifying this change

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-core/src/main/java/org/apache/flink/configuration/ConfigConstants.java
+++ b/flink-core/src/main/java/org/apache/flink/configuration/ConfigConstants.java
@@ -56,20 +56,29 @@ public final class ConfigConstants {
 	 * be "failurerate", "failure-rate" to use FailureRateRestartStrategy. You can also
 	 * specify a class name which implements the RestartStrategy interface and has a static
 	 * create method which takes a Configuration object.
+	 *
+	 * @deprecated use {@link RestartStrategyOptions#RESTART_STRATEGY} instead.
 	 */
+	@Deprecated
 	@PublicEvolving
 	public static final String RESTART_STRATEGY = "restart-strategy";
 
 	/**
 	 * Maximum number of attempts the fixed delay restart strategy will try before failing a job.
+	 *
+	 * @deprecated use {@link RestartStrategyOptions#RESTART_STRATEGY_FIXED_DELAY_ATTEMPTS} instead.
 	 */
+	@Deprecated
 	@PublicEvolving
 	public static final String RESTART_STRATEGY_FIXED_DELAY_ATTEMPTS = "restart-strategy.fixed-delay.attempts";
 
 	/**
 	 * Delay between two consecutive restart attempts in FixedDelayRestartStrategy. It can be specified using Scala's
 	 * FiniteDuration notation: "1 min", "20 s"
+	 *
+	 * @deprecated use {@link RestartStrategyOptions#RESTART_STRATEGY_FIXED_DELAY_DELAY} instead.
 	 */
+	@Deprecated
 	@PublicEvolving
 	public static final ConfigOption<String> RESTART_STRATEGY_FIXED_DELAY_DELAY =
 		key("restart-strategy.fixed-delay.delay").defaultValue("0 s");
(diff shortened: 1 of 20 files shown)


flink-core/src/main/java/org/apache/flink/configuration/ConfigConstants.java
```diff
@@ -93,7 +93,12 @@
 	 * It can be specified using Scala's FiniteDuration notation: "1 min", "20 s".
 	 */
 	@PublicEvolving
-	public static final String RESTART_STRATEGY_FAILURE_RATE_DELAY = "restart-strategy.failure-rate.delay";
```
zentol: For converting constants to options so far we followed a different approach: add new ConfigOptions into the appropriate `*Options` class, deprecate the existing constant in `ConfigConstants` and link to the new option.
Ultimately, the goal is for ConfigConstants to be fully deprecated _at some point_.

This may be a good opportunity to move all restart-related options into a dedicated or existing *Options class. This would have the added benefit that they'd be covered by the docs generator.

tillrohrmann: I think this deserves to be a separate issue because it also requires updating the failover strategy documentation where we currently list the configuration values. The thing I want to improve is simply setting the default value to `0s`.

tillrohrmann:
@zentol the reason why we had this delay was for all tasks to be canceled and their slots to be freed before restarting. Back in the days when we didn't support queued scheduling, this was required because otherwise we would fail with a not enough resource exception. Nowadays this is different since it is ok to wait a bit for the slots being freed.

flink-core/src/main/java/org/apache/flink/configuration/ConfigConstants.java
tillrohrmann: I agree that changing the type is not a nice thing and I mainly followed #5031. But in order to do it properly I guess you are right that we need to create `RestartStrategyOptions`. I'll create a JIRA issue for it and base this PR on it.

zentol:
Isn't the delay also used to prevent restart storms?

tillrohrmann:
The restart storms should no longer happen with queued scheduling because we wait for the slot request timeout before failing again.

tillrohrmann:
@zentol I've migrated the restart config constants to `RestartStrategyOptions` and rebased this PR on it.

zentol:
Are you saying that the slot timeout effectively acts as a restart delay in _all cases_?

zentol:
FLINK-11218 looks related

tillrohrmann:
> FLINK-11218 looks related

Yes, it seems to be related. I've written to the JIRA issue and reached out to the community to ask whether the `0 s` default delay is a problem. Independent of that I think it would be good to unify the behaviour of the `FixedDelay` and `FailureRateRestartStrategy` wrt the restart delay. Both values can be adjusted later if the community sees the `0s` delay as a problem.

zhuzhurk:
> The restart storms should no longer happen with queued scheduling because we wait for the slot request timeout before failing again.

Hi @tillrohrmann , we once experienced restart storms in which case a task failed fast once it is deployed to TM, e.g. failed due to task internal errors or failed when trying to connecting to a bad external node. Queued scheduling might not help in such cases.

tillrohrmann:
> > The restart storms should no longer happen with queued scheduling because we wait for the slot request timeout before failing again.
> 
> Hi @tillrohrmann , we once experienced restart storms in which case a task failed fast once it is deployed to TM, e.g. failed due to task internal errors or failed when trying to connecting to a bad external node. Queued scheduling might not help in such cases.

You are right @zhuzhurk. Restart storms caused by external system overload won't be solved with the slot request timeouts.

Then let's see this PR as a unification of the restart delays of the fixed delay and failure rate restart strategies. Whether to increase the default delay or not will be dealt with in a follow up issue. I've started a [survey on the ML](https://lists.apache.org/thread.html/107b15de6b8ac849610d99c4754715d2a8a2f32ddfe9f8da02af2ccc@%3Cdev.flink.apache.org%3E). Depending on the outcome of it, I'll create a new issue to update the delay.

tillrohrmann:
Closing this PR in favour of setting the default delay to `1 s` as agreed on in [FLIP-62](https://cwiki.apache.org/confluence/display/FLINK/FLIP-62%3A+Set+default+restart+delay+for+FixedDelay-+and+FailureRateRestartStrategy+to+1s). I'll open another PR shortly.

PR #9553, opened 2019-08-28
[FLINK-13883] Remove unused AkkaOptions related to Akka's death watch

tillrohrmann:

We no longer use Akka's death watch. Hence, this commit removes the death
watch related configuration options AkkaOptions#WATCH_HEARTBEAT_INTERVAL,
AkkaOptions#WATCH_THRESHOLD and AkkaOptions#WATCH_HEARTBEAT_PAUSE.

This PR changes the public evolving API and should, hence, add a release note about the removal of the configuration options.

This PR is based on #9552.

## Verifying this change

  - The serializers: (no)

## Documentation


StephanEwen:
+1 to the change

Concerning the point to keep the ConfigOptions for the death watch: I see the point, and it seems to be in line with the handling of the old `ConfigConstants` class. In this case, I would vote to go for cleanup here, to not accumulate too much dead code.
The keeping of previous options was one of the reasons that got us into the weird state in the first place (referring to unused death watch options for restart strategy defaults).
We still do not break existing config files, we only break programs that refer to the config option fields.

tillrohrmann:
Thanks for the review @zentol and @StephanEwen.

Thinking about it again, I would slightly lean towards a graceful removal of the configuration options via deprecation. I guess that the main part of the confusion for users comes from this option still being documented. This should no longer be the case once the options are deprecated.

The other problem Stephan raised is that we misused the config option (in the context of the `FailureRateRestartStrategy`). Deprecating the option won't prevent this in the future to happen again but developers should at least see a deprecation notice telling him not to use it. I hope that this warning would be strong enough in order to prevent the usage of these options until the next release where we could then actually remove the option.

The benefit I see is that we don't break existing programs w/o prior notice (I guess it will be a very small set of affected programs). One could argue that breaking programs would be a kind of notification that one should no longer use this `ConfigOption` which in itself is a good thing. However, this would not work for custom `flink-conf.yamls` where this option is specified. Moreover, we will break these programs in the release after the next one.

Can you live with this proposal @zentol and @StephanEwen?

zentol:
yes.

tillrohrmann:
Why did you close this PR @zentol?

PR #9558, opened 2019-08-29
[FLINK-13892][hs] Harden HistoryServerTest

zentol:
The HistoryServerTest can fail if the HS attempts to unpack a partially written archive. The test only checked that the HS made 1 _attempt_ at fetching archives, and just assumed that the polling cannot fail.

The test now instead tracks how many archives have been successfully fetched instead. Since we are now taking scenarios into account where the HS polls archives multiple times (in case of errors) the polling interval was reduced to 100ms to prevent excessive test durations.

### DIFF
--- a/flink-runtime-web/src/test/java/org/apache/flink/runtime/webmonitor/history/HistoryServerTest.java
+++ b/flink-runtime-web/src/test/java/org/apache/flink/runtime/webmonitor/history/HistoryServerTest.java
@@ -116,11 +116,12 @@ public class HistoryServerTest extends TestLogger {
 		}
 		createLegacyArchive(jmDirectory.toPath());
 
-		CountDownLatch numFinishedPolls = new CountDownLatch(1);
+		CountDownLatch numExpectedArchivedJobs = new CountDownLatch(numJobs + 1);
 
 		Configuration historyServerConfig = new Configuration();
 		historyServerConfig.setString(HistoryServerOptions.HISTORY_SERVER_ARCHIVE_DIRS, jmDirectory.toURI().toString());
 		historyServerConfig.setString(HistoryServerOptions.HISTORY_SERVER_WEB_DIR, hsDirectory.getAbsolutePath());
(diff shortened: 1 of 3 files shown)


flink-runtime-web/src/test/java/org/apache/flink/runtime/webmonitor/history/HistoryServerTest.java
```diff
@@ -131,11 +132,11 @@ public void testHistoryServerIntegration() throws Exception {
 			archives = jmDirectory.listFiles();
```
azagrebin: Is this while loop still needed after changing the verification approach or `HistoryServer` will not monitor job files created after its start?

Also a side note, would it be a more incapsulated approach to poll `MultipleJobsDetails` similar way in this loop until the expected, non-failed state reached instead of instrumenting `HistoryServer` internals for testing?

zentol: Regarding the loop, I suppose we don't technically need it but it can speed up the test.

We had the polling approach in the past but it wasn't particularly stable and way more complicated than the current approach since you have to deal with quite a few error cases.

azagrebin: How does the loop speed up the test? because it even introduces a sleep while `numExpectedArchivedJobs` just immediately gives a callback when everything is done. Does `numExpectedArchivedJobs` not basically subsume the loop time-wise? i.e. `numExpectedArchivedJobs` is done only when the loop is done

Regarding the polling approach I'm fine with it if it has been already discussed before.
It would be a separate issue anyways.

zentol: With the current settings it can speed things up since the first poll of the HS is more likely to fetch all archives.

Another benefit is that it is easier to tell whether the archiving has failed or not. Without the first loop, if the archiving on the JM fails the HS we'd never execute the callback; in this case we'd only know that _something_ went wrong. With the dedicated loop to check the JM archiving it is easier to pinpoint which component is experiencing problems.

PR #9560, opened 2019-08-29
[FLINK-13897][oss] Move NOTICE file into META-INF directory

zentol:
Moves the NOTICE file for flink-oss-fs-hadoop to the correct directory (META-INF), and updates the NOTICE-binary.


NOTICE-binary
```diff
@@ -4732,11 +4851,6 @@ Copyright 2014-2019 The Apache Software Foundation
 flink-fs-hadoop-shaded
 Copyright 2014-2019 The Apache Software Foundation
 
-This project includes software developed at
-The Apache Software Foundation (http://www.apache.org/).
-
-This project bundles the following dependencies under the Apache Software License 2.0 (http://www.apache.org/licenses/LICENSE-2.0.txt)
-
```
tillrohrmann: Why did you remove these lines?

zentol: I didn't do any manual changes; I just built everything, ran the collect_licenses script and removed some unrelated changes.

I'll double check whether an older jar sneaked in, and how the notice file in the oss jar looks like. AFAIK the shade-plugin isn't deduplicating entries, so this is indeed odd.

zentol: nope, happens on a clean build. Is also consistent with the entries for other filesystems that bundle flink-fs-shaded-hadoop.

tillrohrmann: Ok, then I guess it should be correct.

PR #9563, opened 2019-08-29
[FLINK-13514] Fix instability in StreamTaskTest.testAsyncCheckpointingConcurrentCloseAfterAcknowledge

aljoscha:

Before, thread pool shutdown would interrupt our waiting method.
Production code cannot throw an InterruptedException here and would also
not be correct if one is thrown.

We now swallow interrupted exceptions and wait until we successfully
return from await().

## Brief change log

 - we swallow interrupted exceptions in test code

## Verifying this change

 - this is covered by the existing test that we fix

  - this affects checkpointing (only test code is changed, though)


aljoscha:
Merged

PR #9564, opened 2019-08-30
[FLINK-12481][FLINK-12482][FLINK-12958] Streaming runtime: integrate mailbox for timer triggers, checkpoints and AsyncWaitOperator

1u0:

This PR contains changes for
 * FLINK-12481: to migrate timer triggers to mailbox execution model;
 * FLINK-12482: to migrate checkpoint triggers to mailbox execution model;
 * FLINK-12958: modifies `AsyncWaitOperator` to be compatible with mailbox execution model.

## Verifying this change

## Documentation

### DIFF
--- a/flink-formats/flink-parquet/src/test/java/org/apache/flink/formats/parquet/avro/ParquetStreamingFileSinkITCase.java
+++ b/flink-formats/flink-parquet/src/test/java/org/apache/flink/formats/parquet/avro/ParquetStreamingFileSinkITCase.java
@@ -39,7 +39,9 @@ import org.apache.parquet.avro.AvroParquetReader;
 import org.apache.parquet.hadoop.ParquetReader;
 import org.apache.parquet.hadoop.util.HadoopInputFile;
 import org.apache.parquet.io.InputFile;
+import org.junit.Rule;
 import org.junit.Test;
+import org.junit.rules.Timeout;
 
 import java.io.File;
 import java.io.IOException;
@@ -61,6 +63,9 @@ import static org.junit.Assert.assertTrue;
  */
 @SuppressWarnings("serial")
 public class ParquetStreamingFileSinkITCase extends AbstractTestBase {
+
+	@Rule
+	public final Timeout timeoutPerTest = Timeout.seconds(20);
 
 	@Test
 	public void testWriteParquetAvroSpecific() throws Exception {
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/jobgraph/tasks/AbstractInvokable.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/jobgraph/tasks/AbstractInvokable.java
@@ -257,10 +257,9 @@ public abstract class AbstractInvokable {
 	 * Invoked when a checkpoint has been completed, i.e., when the checkpoint coordinator has received
 	 * the notification from all participating tasks.
 	 *
-	 * @param checkpointId The ID of the checkpoint that is complete..
-	 * @throws Exception The notification method may forward its exceptions.
+	 * @param checkpointId The ID of the checkpoint that is complete.
 	 */
-	public void notifyCheckpointComplete(long checkpointId) throws Exception {
+	public void notifyCheckpointComplete(long checkpointId) {
 		throw new UnsupportedOperationException(String.format("notifyCheckpointComplete not supported by %s", this.getClass().getName()));
 	}
 }
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/jobgraph/tasks/AbstractInvokable.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/jobgraph/tasks/AbstractInvokable.java
@@ -27,6 +27,7 @@ import org.apache.flink.runtime.execution.Environment;
 import org.apache.flink.runtime.taskmanager.Task;
 
 import java.util.Optional;
+import java.util.concurrent.Future;
 
 import static org.apache.flink.util.Preconditions.checkNotNull;
 
@@ -53,7 +54,7 @@ import static org.apache.flink.util.Preconditions.checkNotNull;
  * the initial state structure by the Garbage Collector.
  *
  * <p>Any subclass that supports recoverable state and participates in
- * checkpointing needs to override {@link #triggerCheckpoint(CheckpointMetaData, CheckpointOptions, boolean)},
+ * checkpointing needs to override {@link #triggerCheckpointAsync(CheckpointMetaData, CheckpointOptions, boolean)},
  * {@link #triggerCheckpointOnBarrier(CheckpointMetaData, CheckpointOptions, CheckpointMetrics)},
  * {@link #abortCheckpointOnBarrier(long, Throwable)} and {@link #notifyCheckpointComplete(long)}.
  */
@@ -219,10 +220,13 @@ public abstract class AbstractInvokable {
 	 * @param advanceToEndOfEventTime Flag indicating if the source should inject a {@code MAX_WATERMARK} in the pipeline
 	 *                          to fire any registered event-time timers
 	 *
-	 * @return {@code false} if the checkpoint can not be carried out, {@code true} otherwise
+	 * @return future with value of {@code false} if the checkpoint was not carried out, {@code true} otherwise
 	 */
-	public boolean triggerCheckpoint(CheckpointMetaData checkpointMetaData, CheckpointOptions checkpointOptions, boolean advanceToEndOfEventTime) throws Exception {
-		throw new UnsupportedOperationException(String.format("triggerCheckpoint not supported by %s", this.getClass().getName()));
+	public Future<Boolean> triggerCheckpointAsync(
+			CheckpointMetaData checkpointMetaData,
+			CheckpointOptions checkpointOptions,
+			boolean advanceToEndOfEventTime) {
+		throw new UnsupportedOperationException(String.format("triggerCheckpointAsync not supported by %s", this.getClass().getName()));
 	}
 
 	/**
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/jobgraph/tasks/AbstractInvokable.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/jobgraph/tasks/AbstractInvokable.java
@@ -56,7 +56,7 @@ import static org.apache.flink.util.Preconditions.checkNotNull;
  * <p>Any subclass that supports recoverable state and participates in
  * checkpointing needs to override {@link #triggerCheckpointAsync(CheckpointMetaData, CheckpointOptions, boolean)},
  * {@link #triggerCheckpointOnBarrier(CheckpointMetaData, CheckpointOptions, CheckpointMetrics)},
- * {@link #abortCheckpointOnBarrier(long, Throwable)} and {@link #notifyCheckpointComplete(long)}.
+ * {@link #abortCheckpointOnBarrier(long, Throwable)} and {@link #notifyCheckpointCompleteAsync(long)}.
  */
 public abstract class AbstractInvokable {
 
@@ -262,8 +262,10 @@ public abstract class AbstractInvokable {
 	 * the notification from all participating tasks.
 	 *
 	 * @param checkpointId The ID of the checkpoint that is complete.
+	 *
+	 * @return future that completes when the notification has been processed by the task.
 	 */
-	public void notifyCheckpointComplete(long checkpointId) {
-		throw new UnsupportedOperationException(String.format("notifyCheckpointComplete not supported by %s", this.getClass().getName()));
+	public Future<Void> notifyCheckpointCompleteAsync(long checkpointId) {
+		throw new UnsupportedOperationException(String.format("notifyCheckpointCompleteAsync not supported by %s", this.getClass().getName()));
 	}
 }
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/Task.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/Task.java
@@ -92,8 +92,6 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.Executor;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.atomic.AtomicBoolean;
@@ -259,19 +257,13 @@ public class Task implements Runnable, TaskActions, PartitionProducerStateProvid
 	/** The observed exception, in case the task execution failed. */
 	private volatile Throwable failureCause;
 
-	/** Serial executor for asynchronous calls (checkpoints, etc), lazily initialized. */
-	private volatile ExecutorService asyncCallDispatcher;
-
 	/** Initialized from the Flink configuration. May also be set at the ExecutionConfig */
 	private long taskCancellationInterval;
 
 	/** Initialized from the Flink configuration. May also be set at the ExecutionConfig */
 	private long taskCancellationTimeout;
 
-	/**
-	 * This class loader should be set as the context class loader of the threads in
-	 * {@link #asyncCallDispatcher} because user code may dynamically load classes in all callbacks.
-	 */
+	/** This class loader should be set as the context class loader for threads that may dynamically load user code. */
 	private ClassLoader userCodeClassLoader;
 
 	/**
@@ -806,13 +798,6 @@ public class Task implements Runnable, TaskActions, PartitionProducerStateProvid
 				// to the invokable and its structures in cases where this Task object is still referenced
 				this.invokable = null;
 
-				// stop the async dispatcher.
-				// copy dispatcher reference to stack, against concurrent release
-				ExecutorService dispatcher = this.asyncCallDispatcher;
-				if (dispatcher != null && !dispatcher.isShutdown()) {
-					dispatcher.shutdownNow();
-				}
-
 				// free the network resources
 				releaseNetworkResources();
 
@@ -1137,35 +1122,26 @@ public class Task implements Runnable, TaskActions, PartitionProducerStateProvid
 		final CheckpointMetaData checkpointMetaData = new CheckpointMetaData(checkpointID, checkpointTimestamp);
 
 		if (executionState == ExecutionState.RUNNING && invokable != null) {
-
-			Runnable runnable = new Runnable() {
-				@Override
-				public void run() {
-					try {
-						invokable.triggerCheckpointAsync(checkpointMetaData, checkpointOptions, advanceToEndOfEventTime);
-					}
-					catch (RejectedExecutionException ex) {
-						// This may happen if the mailbox is closed. It means that the task is shutting down, so we just ignore it.
-						LOG.debug(
-							"Triggering checkpoint {} for {} ({}) was rejected by the mailbox",
-							checkpointID, taskNameWithSubtask, executionId);
-					}
-					catch (Throwable t) {
-						if (getExecutionState() == ExecutionState.RUNNING) {
-							failExternally(new Exception(
-								"Error while triggering checkpoint " + checkpointID + " for " +
-									taskNameWithSubtask, t));
-						} else {
-							LOG.debug("Encountered error while triggering checkpoint {} for " +
-								"{} ({}) while being not in state running.", checkpointID,
-								taskNameWithSubtask, executionId, t);
-						}
-					}
+			try {
+				invokable.triggerCheckpointAsync(checkpointMetaData, checkpointOptions, advanceToEndOfEventTime);
+			}
+			catch (RejectedExecutionException ex) {
+				// This may happen if the mailbox is closed. It means that the task is shutting down, so we just ignore it.
+				LOG.debug(
+					"Triggering checkpoint {} for {} ({}) was rejected by the mailbox",
+					checkpointID, taskNameWithSubtask, executionId);
+			}
+			catch (Throwable t) {
+				if (getExecutionState() == ExecutionState.RUNNING) {
+					failExternally(new Exception(
+						"Error while triggering checkpoint " + checkpointID + " for " +
+							taskNameWithSubtask, t));
+				} else {
+					LOG.debug("Encountered error while triggering checkpoint {} for " +
+						"{} ({}) while being not in state running.", checkpointID,
+						taskNameWithSubtask, executionId, t);
 				}
-			};
-			executeAsyncCallRunnable(
-					runnable,
-					String.format("Checkpoint Trigger for %s (%s).", taskNameWithSubtask, executionId));
+			}
 		}
 		else {
 			LOG.debug("Declining checkpoint request for non-running task {} ({}).", taskNameWithSubtask, executionId);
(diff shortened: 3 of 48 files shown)


1u0:
@flinkbot run travis

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
```diff
@@ -1368,13 +1369,19 @@ private void checkpointStreamOperator(StreamOperator<?> op) throws Exception {
...
+				mailboxProcessor.getMailboxExecutor(TaskMailbox.MAX_PRIORITY).execute(() -> {
+					synchronized (getCheckpointLock()) {
+						try {
+							callback.onProcessingTime(timestamp);
+						} catch (Throwable t) {
+							handleAsyncException("Caught exception while processing timer.", new TimerException(t));
+						}
+					}
+				});
+			} catch (Throwable t) {
```
AHeise: Do we need the outer exception handling? If the mailbox is incorrect state, then I'd image that the stack trace is already good enough.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
```diff
@@ -428,6 +429,10 @@ public final void invoke() throws Exception {
 				// See FLINK-7430
 				isRunning = false;
 			}
+			MailboxExecutor mainMailboxExecutor = mailboxProcessor.getMainMailboxExecutor();
+			while (mainMailboxExecutor.tryYield()) {
+				// Run until we have last letters.
```
AHeise: Suggestion: // Run until we have processed all pending letters.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
```diff
@@ -64,6 +64,7 @@
 import org.apache.flink.streaming.runtime.streamstatus.StreamStatusMaintainer;
```
AHeise: Commit message should be a bit more expressive, for example: Drain mailbox while shutting down StreamTask

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
```diff
@@ -428,6 +429,10 @@ public final void invoke() throws Exception {
 				// See FLINK-7430
 				isRunning = false;
 			}
+			MailboxExecutor mainMailboxExecutor = mailboxProcessor.getMainMailboxExecutor();
```
AHeise: I'd rather add a method drain to mailboxProcessor that does the following lines.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
```diff
@@ -841,6 +845,10 @@ protected void declineCheckpoint(long checkpointId) {
 			new CheckpointException("Task Name" + getName(), CheckpointFailureReason.CHECKPOINT_DECLINED_TASK_NOT_READY));
 	}
 
+	protected void handleCheckpointException(Exception exception) {
+		handleAsyncException(exception);
```
AHeise: Why not use existing handleAsyncException and add a message?

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
```diff
@@ -962,6 +970,10 @@ public ProcessingTimeService getProcessingTimeService() {
 	 */
 	@Override
 	public void handleAsyncException(String message, Throwable exception) {
+		handleAsyncException(exception);
+	}
+
+	private void handleAsyncException(Throwable exception) {
 		if (isRunning) {
 			// only fail if the task is still running
 			asyncExceptionHandler.handleAsyncException(message, exception);
```
AHeise: Where is the message coming from?

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/SynchronousCheckpointITCase.java
```diff
@@ -151,11 +152,15 @@ protected void performDefaultAction(DefaultActionContext context) throws Excepti
...
-			eventQueue.put(Event.POST_TRIGGER_CHECKPOINT);
-			return result;
+		public Future<Boolean> triggerCheckpointAsync(CheckpointMetaData checkpointMetaData, CheckpointOptions checkpointOptions, boolean advanceToEndOfEventTime) {
+			try {
+				eventQueue.put(Event.PRE_TRIGGER_CHECKPOINT);
+				Future<Boolean> result = super.triggerCheckpointAsync(checkpointMetaData, checkpointOptions, advanceToEndOfEventTime);
+				eventQueue.put(Event.POST_TRIGGER_CHECKPOINT);
+				return result;
+			} catch (InterruptedException e) {
+				throw new RuntimeException(e);
```
AHeise: Swallowing of thread interruption. 
Would probably be better to either allow triggerCheckpointAsync to throw (interrupt)exceptions or at least also call Thread.interrupt().

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/Task.java
```diff
@@ -27,7 +27,6 @@
 import org.apache.flink.configuration.TaskManagerOptions;
```
AHeise: Could you add a longer description to the commit message, why it's not needed anymore?

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/Task.java
```diff
@@ -1183,8 +1183,12 @@ public void notifyCheckpointComplete(final long checkpointID) {
 				@Override
 				public void run() {
 					try {
-						invokable.notifyCheckpointComplete(checkpointID);
-					} catch (Throwable t) {
+						invokable.notifyCheckpointCompleteAsync(checkpointID);
```
AHeise: Is the future used at all? Do we need a #get here?

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/Task.java
1u0: Do you mean `FileSystemSafetyNet` usage? I can do.
My reasoning is because the thread is now only passing execution from one thread (RPC) to another (main task thread) and not calling user code anymore.

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/Task.java
1u0: Correct me if I'm wrong, but I think, we don't need to wait for a result here.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SystemProcessingTimeService.java
```diff
@@ -249,89 +239,53 @@ int getNumTasksScheduled() {
 	// ------------------------------------------------------------------------
 
 	/**
-	 * Internal task that is invoked by the timer service and triggers the target.
+	 * A context to which {@link ProcessingTimeCallback} would be passed to be invoked when a timer is up.
 	 */
-	private static final class TriggerTask implements Runnable {
+	public interface ScheduledCallbackExecutionContext {
```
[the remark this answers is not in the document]
1u0: Can do. But then, I'd prefer to mark `SystemProcessingTimeService` as `@Internal` instead. What do you think?

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SystemProcessingTimeService.java
```diff
@@ -249,89 +239,53 @@ int getNumTasksScheduled() {
...
 
-	/**
-	 * Internal task which is repeatedly called by the processing time service.
-	 */
-	private static final class RepeatedTriggerTask implements Runnable {
+	private Runnable wrapOnTimerCallback(ProcessingTimeCallback callback, long nextTimestamp, long period) {
+		return new TimeTrackingDelegate(status, callbackExecutionContext, callback, nextTimestamp, period);
+	}
 
+	private static final class TimeTrackingDelegate implements Runnable {
```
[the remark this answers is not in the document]
1u0: Suggestion for a better name?

flink-streaming-java/src/test/java/org/apache/flink/streaming/api/operators/async/ExecutorServiceRule.java
```diff
@@ -0,0 +1,152 @@
+/*
```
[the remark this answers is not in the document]
1u0: I'll remove it, this was included with your old commit.

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/SynchronousCheckpointITCase.java
1u0: This `InterruptedException` is from `LinkiedBlockingQueue.put()` (`eventQueue.put()`).
The queue is not bounded and I hope this exception won't be thrown here during tests runs.
But even if thrown, I'd prefer to propagate them.

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/SystemProcessingTimeServiceTest.java
```diff
@@ -391,61 +384,6 @@ public void onProcessingTime(long timestamp) throws Exception {}
 		}
 	}
 
-	@Test
```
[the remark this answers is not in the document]
1u0: I find those two tests not applicable to `SystemProcessingTimeService`, as `AsyncExceptionHandler` is not part of `SystemProcessingTimeService` api contract anymore.
But I'll try to add something similar for `StreamTask.TimerInvocationContext` as a test.

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/Task.java
pnowojski: if we do not wait, then we will miss the error handling? 

But please first respond to my comment about double/nested mailbox/enqueuing pattern.

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/Task.java
```diff
@@ -1139,17 +1138,9 @@ public void triggerCheckpointBarrier(
...
 		if (executionState == ExecutionState.RUNNING && invokable != null) {
 
-			// build a local closure
-			final SafetyNetCloseableRegistry safetyNetCloseableRegistry =
-				FileSystemSafetyNet.getSafetyNetCloseableRegistryForThread();
-
 			Runnable runnable = new Runnable() {
 				@Override
 				public void run() {
-					// set safety net from the task's context for checkpointing thread
```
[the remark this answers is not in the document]
1u0: > Are you sure we can remove this? What was the purpose of this safety net?

Do you have any reasons to keep it?

For the second part of your comment, I don't think it would be correct to allow any other (non main task thread) to get letters from the mailbox. And so far, we haven't raised any reason to have such requirement (that letters can be run by some other threads).

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/Task.java
```diff
@@ -1151,7 +1151,10 @@ public void run() {
 					FileSystemSafetyNet.setSafetyNetCloseableRegistryForThread(safetyNetCloseableRegistry);
 
 					try {
-						invokable.triggerCheckpoint(checkpointMetaData, checkpointOptions, advanceToEndOfEventTime);
+						invokable.triggerCheckpointAsync(checkpointMetaData, checkpointOptions, advanceToEndOfEventTime);
```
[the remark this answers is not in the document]
1u0: I'm not so knowledgeable about RPC execution model. If the RPC requests are guaranteed to be handled in ordered matter (at least per task basis), then we can get rid of `asyncCallDispatcher`.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/AsyncWaitOperatorFactory.java
```diff
@@ -0,0 +1,86 @@
...
+			long timeout,
+			int capacity,
+			AsyncDataStream.OutputMode outputMode) {
+		this.asyncFunction = asyncFunction;
+		this.timeout = timeout;
+		this.capacity = capacity;
+		this.outputMode = outputMode;
+	}
+
+	@Override public void setMailboxExecutor(MailboxExecutor mailboxExecutor) {
```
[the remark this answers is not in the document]
1u0: If it's not required by the project code style, I'd prefer to keep it as is.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/AsyncWaitOperatorFactory.java
```diff
@@ -0,0 +1,86 @@
...
+
+	@Override public StreamOperator createStreamOperator(StreamTask containingTask, StreamConfig config,
+			Output output) {
+		AsyncWaitOperator asyncWaitOperator = new AsyncWaitOperator(
+				asyncFunction,
+				timeout,
+				capacity,
+				outputMode,
+				mailboxExecutor);
+		asyncWaitOperator.setup(containingTask, config, output);
```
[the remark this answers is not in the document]
1u0: > whole `AsyncWaitOperatorFactory` could be cleaned up: all fields could be made non `transient` and `final`.

Do you mean `AsyncWaitOperator` here? Because `AsyncWaitOperatorFactory` doesn't have any `transient` fields and other fields are `final` as much as possible.

For `AsyncWaitOperator`, changing it here would be yak shaving and jumping into a rabbit hole, imo. I can create a follow up ticket for that (for `AsyncWaitOperator`!).

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/Task.java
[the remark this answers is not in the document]
1u0: No.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/AsyncWaitOperatorFactory.java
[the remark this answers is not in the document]
pnowojski: Yes, I meant AsyncWaitOperator :) I copied/pated wrong class name.

> I can create a follow up ticket for that 

Please do so and can you CC me in the ticket?

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/Task.java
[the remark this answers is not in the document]
pnowojski: I would guess that there is only one RPC thread, however I do not see an issue here even if they weren't? Those actions would be ordered on the mailbox, the same way they were ordered on the `asyncCallDispatcher`. If there was a race on the RPC thread between requests, there would be the same situation now, just against enqueuing into a different single threaded executor.

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/Task.java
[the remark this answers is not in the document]
pnowojski: > Do you have any reasons to keep it?

I don't know this code and I'm not sure how it's suppose to work, but I see that it's being used (`FileSystemSafetyNet#initializeSafetyNetForThread/closeSafetyNetAndGuardedResourcesForThread/wrapWithSafetyNetWhenActivated`). Since you were removing it, I thought that you have investigated how is it suppose to be working and that's safe to remove it (as I'm pretty sure that such safety feature has a very poor test coverage).

> For the second part of your comment, I don't think it would be correct to allow any other (non main task thread) to get letters from the mailbox. And so far, we haven't raised any reason to have such requirement (that letters can be run by some other threads).

I guess that's right as long as we do not allow for yielding in the legacy source threads. Do we have a `checkstate` somewhere for that? Somewhere Inside yield a `checkState(isMailboxThread())`?

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SystemProcessingTimeService.java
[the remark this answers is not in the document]
AHeise: Yes, fine by me. Can you also double-check if packet-private may also work (hard to tell in PR)?

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/SystemProcessingTimeServiceTest.java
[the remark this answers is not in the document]
AHeise: Yes, good idea.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SystemProcessingTimeService.java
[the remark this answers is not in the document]
AHeise: Getting inspired by ScheduledThreadPoolExecutor, I'd call it ScheduledTask.

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/SynchronousCheckpointITCase.java
AHeise: Yes, do both, interrupt() and throw the exception.
Didn't find the article that I was looking for, but I think this [one](https://www.yegor256.com/2015/10/20/interrupted-exception.html) is good enough.

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/Task.java
AHeise: Tbh I'm not deep enough to understand that. I just found it odd to remove a safety net that was placed for some reasons. Maybe just add your response to the commit message.

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/Task.java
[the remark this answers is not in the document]
1u0: My assumption that we are not calling any user code and file system operations in the dispatcher thread anymore. Hence, we can remove the `FileSystemSafetyNet` guard usage there.
I'm open for corrections, if someone finds this to be not true.

> Do we have a checkstate somewhere for that? Somewhere Inside yield a checkState(isMailboxThread())?

There are checks that you can find in the code (and you should have read it from the preceding PRs).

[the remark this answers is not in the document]
pnowojski: To me it looks like `FileSystemSafetyNet.setSafetyNetCloseableRegistryForThread` should be called by any thread that will be executing some/any part of the code that used to be invoked in `invokable.triggerCheckpoint`. This is the entry point, where safety net registry was being set, and then it was being used implicitly via static variables in `SnapshotDirectory` via `FileSystem#getLocalFileSystem`. 

> we are not calling any user code and file system operations in the dispatcher thread anymore.

I guess yes, probably setting it here now doesn't make sense, but it shouldn't be just removed from here, but moved used inside the mailbox action?

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
1u0: This have been wrongly rebased (intermediate commit was not compiling, but the end PR was correct).

Regarding `handleAsyncException()`, I think there needs to be some refactoring/more thinking about the final exception that thrown. The doc string for `AsynchronousException` says:
```
 * An exception for wrapping exceptions that are thrown by an operator in threads other than the
 * main compute thread of that operator.
```
but moving timers and checkpoint triggers to the mailbox execution, this no longer true.

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/Task.java
1u0: I've added a description to the commit.

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/Task.java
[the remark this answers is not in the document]
1u0: Mailbox action is executed by the main task's thread. I think that thread has already `FileSystemSafetyNet` guard set.

flink-formats/flink-parquet/src/test/java/org/apache/flink/formats/parquet/avro/ParquetStreamingFileSinkITCase.java
```diff
@@ -62,6 +64,9 @@
 @SuppressWarnings("serial")
 public class ParquetStreamingFileSinkITCase extends AbstractTestBase {
 
+	@Rule
+	public final Timeout timeoutPerTest = Timeout.seconds(20);
```
AHeise: +1 for using rule

flink-streaming-java/src/test/java/org/apache/flink/streaming/api/operators/async/AsyncWaitOperatorTest.java
```diff
@@ -700,11 +696,11 @@ private void testAsyncTimeout(LazyAsyncFunction lazyAsyncFunction,
 			2,
 			AsyncDataStream.OutputMode.ORDERED);
 
-		final MockEnvironment mockEnvironment = createMockEnvironment();
-		mockEnvironment.setExpectedExternalFailureCause(Throwable.class);
-
 		final OneInputStreamOperatorTestHarness<Integer, Integer> testHarness =
-			new OneInputStreamOperatorTestHarness<>(operator, IntSerializer.INSTANCE, mockEnvironment);
+			new OneInputStreamOperatorTestHarness<>(operator, IntSerializer.INSTANCE);
```
AHeise: +1 for unification

flink-streaming-java/src/test/java/org/apache/flink/streaming/api/operators/async/AsyncWaitOperatorTest.java
```diff
@@ -1127,4 +1073,15 @@ public void testEndInput() throws Exception {
 			}
 		}
 	}
+
+	private static <OUT> OneInputStreamOperatorTestHarness<Integer, OUT> createTestHarness(
```
AHeise: +1 to reduce boilerblate

flink-streaming-java/src/test/java/org/apache/flink/streaming/api/operators/async/AsyncWaitOperatorTest.java
```diff
@@ -937,19 +894,16 @@ private void testTimeoutExceptionHandling(AsyncDataStream.OutputMode outputMode)
...
-		final OneInputStreamOperatorTestHarness<Integer, Integer> snapshotHarness = new OneInputStreamOperatorTestHarness<>(
-			new AsyncWaitOperator<>(
-				controllableAsyncFunction, // the NoOpAsyncFunction is like a blocking function
-				1000L,
-				capacity,
-				AsyncDataStream.OutputMode.ORDERED),
-			IntSerializer.INSTANCE);
+
+		final OneInputStreamOperatorTestHarness<Integer, Integer> snapshotHarness = createTestHarness(
+			new ControllableAsyncFunction<>(trigger), // the NoOpAsyncFunction is like a blocking function
```
AHeise: nit: comment outdated?

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/tasks/SystemProcessingTimeServiceTest.java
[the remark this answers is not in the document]
1u0: I've migrated similar tests into `StreamTaskTimerTest`.

flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/Task.java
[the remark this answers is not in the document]
1u0: I've removed `asyncCallDispatcher` from the `Task`.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/Emitter.java
```diff
@@ -80,9 +88,14 @@ public void run() {
...
 			while (running) {
 				LOG.debug("Wait for next completed async stream element result.");
-				AsyncResult streamElementEntry = streamElementQueue.peekBlockingly();
-
-				output(streamElementEntry);
+				AsyncResult asyncResult = streamElementQueue.peekBlockingly();
+				executor.submit(() -> {
+					try {
+						output(asyncResult);
+					} catch (InterruptedException e) {
```
pnowojski: Why do we need this `InterruptedException`? I see that it's coming (and it is the same in master branch) from `streamElementQueue.poll()`, however I don't understand it:
1. We shouldn't be blocking neither in the mailbox thread (this version) nor under the checkpoint lock (master version), so this already looks suspicious.
2. `poll` method that blocks is already "mildly" surprising contract on it's own
3. From the code it looks like this `poll` call should never block and changing it to a proper poll (returning `Optional` or `null`) and adding a check state that the element is present, would work just as well but would be better/cleaner.

@AHeise I remember that you were also rising some objections about the `Emitter` code and that it could be simplified and you wanted to work on that later. Did you have this in mind as well?

Am I missing something? If not can you @AHeise create a JIRA ticket for cleaning up those things?

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/AsyncWaitOperatorFactory.java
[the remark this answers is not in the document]
1u0: The created ticket: [FLINK-14011](https://issues.apache.org/jira/browse/FLINK-14011).

flink-runtime/src/main/java/org/apache/flink/runtime/jobgraph/tasks/AbstractInvokable.java
```diff
@@ -261,10 +261,9 @@ public void abortCheckpointOnBarrier(long checkpointId, Throwable cause) throws
 	 * Invoked when a checkpoint has been completed, i.e., when the checkpoint coordinator has received
```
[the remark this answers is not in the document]
1u0: I've added a commit message.

> Maybe it should be squashed with the following commit?

I prefer to separate grunt work against commits that actually introduce new behavior/feature.
That commit is considered as grunt work (and also the reason why I had difficulties to find it better commit message.)

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/StreamTask.java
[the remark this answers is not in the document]
1u0: There is an existing `AsyncWaitOperator` test that uses timers and blocks until it's elements are not resolved (by timeout). Without this change, the test would fail.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/Emitter.java
AHeise: Yes, I would like to get rid of any blocking methods inside emitter and `streamElementQueue`.
The main idea is to transfer the results of async operations via letters to the mailbox thread and then only have non-blocking, non-synchronized data structures that ensure the ordering of inputs/results. At that point, no interruptions can occur.
I'm opening a ticket for that.

1u0:
I'm looking into `SystemProcessingTimeServiceTest.testImmediateShutdown()` test that became flaky.

1u0:
@flinkbot run travis

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/Emitter.java
AHeise: https://issues.apache.org/jira/browse/FLINK-14044

pnowojski:
Could you @1u0 close the relevant tickets?

PR #9565, opened 2019-08-30
[FLINK-12501] Use SpecificRecord.getSchema in AvroFactory

aljoscha:

Before, we were using SpecificData.getSchema(type) which was not working
for types that were generated using Avrohugger (for Scala) because
the SCHEMA was generated in the companion object. Now we use a method
that must be available on all SpecificRecord(s).

We still use the old method as a fallback if we cannot instantiate or
call getSchema() on the instance.

## Brief change log

  - We create an instance of the `SpecificRecord` using `Class.newInstance` and then call `SpecificRecord.getSchema()`. We use the old method as a fallback

## Verifying this change

This is covered by existing tests.

  - The serializers: yes, the avro serializer

## Documentation

### DIFF
--- a/flink-formats/flink-avro/src/main/java/org/apache/flink/formats/avro/typeutils/AvroFactory.java
+++ b/flink-formats/flink-avro/src/main/java/org/apache/flink/formats/avro/typeutils/AvroFactory.java
@@ -37,6 +37,8 @@ import org.apache.avro.specific.SpecificData;
 import org.apache.avro.specific.SpecificDatumReader;
 import org.apache.avro.specific.SpecificDatumWriter;
 import org.apache.avro.specific.SpecificRecord;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.util.Optional;
 
@@ -49,6 +51,8 @@ import static org.apache.flink.util.Preconditions.checkNotNull;
  */
 @Internal
 final class AvroFactory<T> {
+
+	private static final Logger LOG = LoggerFactory.getLogger(AvroFactory.class);
 
 	private final DataOutputEncoder encoder = new DataOutputEncoder();
 	private final DataInputDecoder decoder = new DataInputDecoder();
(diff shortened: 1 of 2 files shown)


aljoscha:
R: @igalshilman could you maybe review this?

flink-formats/flink-avro/src/main/java/org/apache/flink/formats/avro/typeutils/AvroFactory.java
```diff
@@ -94,7 +94,8 @@ static Schema parseSchemaString(@Nullable String schemaString) {
 	@SuppressWarnings("OptionalUsedAsFieldOrParameterType")
 	private static <T> AvroFactory<T> fromSpecific(Class<T> type, ClassLoader cl, Optional<Schema> previousSchema) {
 		SpecificData specificData = new SpecificData(cl);
-		Schema newSchema = specificData.getSchema(type);
+		Optional<Schema> newSchemaOptional = tryExtractAvroSchema(type);
+		Schema newSchema = newSchemaOptional.orElse(specificData.getSchema(type));
```
debasishg: I faced a problem here which comes from the use of `Optional.orElse`. It's a known gotcha with `orElse`. Even though `newSchemaOptional` is non-empty, `orElse` will evaluate the argument passed as default. Hence it tries to evaluate `specificData.getSchema(type))` and fails with the earlier exception. The workaround is to change `orElse` to `orElseGet` and pass a lambda instead ..

```java
Schema newSchema = newSchemaOptional.orElseGet(() -> specificData.getSchema(type));
```

This fixes the problem and runs ok. There are a few other invocations of `orElse` in this class which you may want to change to `orElseGet` as well ..

I cherrypicked your commit into the `release-1.9` branch and did a build which runs fine. Thanks for the help.

aljoscha: Thanks! That's very helpful. I'll update the PR.

aljoscha:
I updated the PR.

debasishg:
Thanks a lot!

aljoscha:
@flinkbot run travis

aljoscha:
@flinkbot run travis

aljoscha:
@debasishg We did some changes because I forgot to also adapt the `AvroSerializerSnapshot`. Could you again check if this works with your Avrohugger classes? Maybe also check whether you can successfully take a savepoint and restore, because that part was not previously covered by the changes.

debasishg:
@aljoscha Sure I will check all of them ..

debasishg:
works ok .. Thanks!

aljoscha:
@debasishg thanks again! I merged this now.

rsheldiii:
would a similar wrapper around [AvroDeserializationSchema's checkAvroInitialized](https://github.com/apache/flink/blob/master/flink-formats/flink-avro/src/main/java/org/apache/flink/formats/avro/AvroDeserializationSchema.java#L145) function allow for avro-hugger deserialization support? https://github.com/julianpeeters/avrohugger/issues/111

aljoscha:
We could probably do the same as in https://github.com/apache/flink/pull/9565

PR #9567, opened 2019-08-30
[FLINK-13887] [core] Ensure defaultInputDependencyConstraint to be non-null when setting it in ExecutionConfig

zhuzhurk:

*If a user invokes ExecutionConfig#setDefaultInputDependencyConstraint(null) to set the defaultInputDependencyConstraint to be null, the scheduling topology building will throw NPE in ExecutionGraph creating stage, causing a master node fatal error.*

## Brief change log

*Setting defaultInputDependencyConstraint to ANY in ExecutionConfig if ExecutionConfig#setDefaultInputDependencyConstraint is invoked with a `null` param.*

## Verifying this change

This change is trivial so no test case is added to reduce code redundancy.
*Manually verified the change by invoking ExecutionConfig#setDefaultInputDependencyConstraint in BatchFineGrainedRecoveryITCase. Without this change there will be job submission error and the case will fail. With this change, the case can pass.*

## Documentation

### DIFF
--- a/flink-core/src/main/java/org/apache/flink/api/common/ExecutionConfig.java
+++ b/flink-core/src/main/java/org/apache/flink/api/common/ExecutionConfig.java
@@ -546,7 +546,13 @@ public class ExecutionConfig implements Serializable, Archiveable<ArchivedExecut
 	 */
 	@PublicEvolving
 	public void setDefaultInputDependencyConstraint(InputDependencyConstraint inputDependencyConstraint) {
-		this.defaultInputDependencyConstraint = inputDependencyConstraint;
+		if (inputDependencyConstraint != null) {
+			this.defaultInputDependencyConstraint = inputDependencyConstraint;
+		} else {
+			// defaultInputDependencyConstraint is not allowed to be null
+			// setting it to ANY to not break existing jobs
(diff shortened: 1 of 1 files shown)


flink-core/src/main/java/org/apache/flink/api/common/ExecutionConfig.java
```diff
@@ -546,7 +546,13 @@ public ExecutionMode getExecutionMode() {
...
 	public void setDefaultInputDependencyConstraint(InputDependencyConstraint inputDependencyConstraint) {
-		this.defaultInputDependencyConstraint = inputDependencyConstraint;
+		if (inputDependencyConstraint != null) {
+			this.defaultInputDependencyConstraint = inputDependencyConstraint;
+		} else {
+			// defaultInputDependencyConstraint is not allowed to be null
+			// setting it to ANY to not break existing jobs
+			this.defaultInputDependencyConstraint = InputDependencyConstraint.ANY;
+		}
 	}
```
zhuzhurk: I think this change is trivial, to reduce code redundancy I did not add any test case for it.
I manually verified the change by invoking ExecutionConfig#setDefaultInputDependencyConstraint in BatchFineGrainedRecoveryITCase. 
Without this change there will be job submission error and the case will fail. 
With this change, the case can pass.

PR #9575, opened 2019-08-30
[FLINK-13912][client] Remove ClusterClient#getClusterConnectionInfo

TisonKun:

As discussed in [FLINK-13750](https://issues.apache.org/jira/browse/FLINK-13750), we actually doesn't need this method any more. All configuration needed is `WebMonitor` address and port in standalone HA mode. We can safely remove this method and replace its usages with `ClusterClient#getWebInterfaceURL`

## Brief change log

- Remove `ClusterClient#getClusterConnectionInfo`
- Replace its usages with `ClusterClient#getWebInterfaceURL`
- Adjust logs and tests

## Verifying this change

This change remove an existing method so that existing tests should guard nothing breaks.

  - The serializers: (no)

## Documentation

cc @tillrohrmann

### DIFF
--- a/flink-scala-shell/src/main/scala/org/apache/flink/api/scala/FlinkShell.scala
+++ b/flink-scala-shell/src/main/scala/org/apache/flink/api/scala/FlinkShell.scala
@@ -19,12 +19,12 @@
 package org.apache.flink.api.scala
 
 import java.io._
+import java.net.{InetSocketAddress, URL}
 
 import org.apache.flink.client.cli.{CliFrontend, CliFrontendParser}
 import org.apache.flink.client.deployment.ClusterDescriptor
 import org.apache.flink.client.program.ClusterClient
 import org.apache.flink.configuration.{Configuration, GlobalConfiguration, JobManagerOptions}
-import org.apache.flink.runtime.akka.AkkaUtils
 import org.apache.flink.runtime.minicluster.{MiniCluster, MiniClusterConfiguration}
 
 import scala.collection.mutable.ArrayBuffer
@@ -277,8 +277,9 @@ object FlinkShell {
 
     val cluster = clusterDescriptor.deploySessionCluster(clusterSpecification)
 
-    val inetSocketAddress = AkkaUtils.getInetSocketAddressFromAkkaURL(
-      cluster.getClusterConnectionInfo.getAddress)
+    val webMonitorUrl = new URL(cluster.getWebInterfaceURL)
+
+    val inetSocketAddress = new InetSocketAddress(webMonitorUrl.getHost, webMonitorUrl.getPort)
 
     val address = inetSocketAddress.getAddress.getHostAddress
     val port = inetSocketAddress.getPort
@@ -317,10 +318,9 @@ object FlinkShell {
       throw new RuntimeException("Yarn Cluster could not be retrieved.")
     }
 
-    val jobManager = AkkaUtils.getInetSocketAddressFromAkkaURL(
-      cluster.getClusterConnectionInfo.getAddress)
+    val webMonitorUrl = new URL(cluster.getWebInterfaceURL)
 
-    (jobManager.getHostString, jobManager.getPort, None)
+    (webMonitorUrl.getHost, webMonitorUrl.getPort, None)
   }
 
   def ensureYarnConfig(config: Config) = config.yarnConfig match {
--- a/flink-scala-shell/src/main/scala/org/apache/flink/api/scala/FlinkShell.scala
+++ b/flink-scala-shell/src/main/scala/org/apache/flink/api/scala/FlinkShell.scala
@@ -279,12 +279,7 @@ object FlinkShell {
 
     val webMonitorUrl = new URL(cluster.getWebInterfaceURL)
 
-    val inetSocketAddress = new InetSocketAddress(webMonitorUrl.getHost, webMonitorUrl.getPort)
-
-    val address = inetSocketAddress.getAddress.getHostAddress
-    val port = inetSocketAddress.getPort
-
-    (address, port, Some(Right(cluster)))
+    (webMonitorUrl.getHost, webMonitorUrl.getPort, Some(Right(cluster)))
   }
 
   def fetchDeployedYarnClusterInfo(
--- a/flink-scala-shell/src/main/scala/org/apache/flink/api/scala/FlinkShell.scala
+++ b/flink-scala-shell/src/main/scala/org/apache/flink/api/scala/FlinkShell.scala
@@ -19,7 +19,7 @@
 package org.apache.flink.api.scala
 
 import java.io._
-import java.net.{InetSocketAddress, URL}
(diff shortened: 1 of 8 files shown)


flink-scala-shell/src/main/scala/org/apache/flink/api/scala/FlinkShell.scala
```diff
@@ -277,8 +277,9 @@ object FlinkShell {
 
     val cluster = clusterDescriptor.deploySessionCluster(clusterSpecification)
 
-    val inetSocketAddress = AkkaUtils.getInetSocketAddressFromAkkaURL(
-      cluster.getClusterConnectionInfo.getAddress)
```
tillrohrmann: Did this ever work with Flip-6?

flink-scala-shell/src/main/scala/org/apache/flink/api/scala/FlinkShell.scala
```diff
@@ -277,8 +277,9 @@ object FlinkShell {
...
+    val webMonitorUrl = new URL(cluster.getWebInterfaceURL)
+
+    val inetSocketAddress = new InetSocketAddress(webMonitorUrl.getHost, webMonitorUrl.getPort)
```
tillrohrmann: I guess we should make it consistent with `fetchDeployedYarnClusterInfo`. In fact, I think that not trying to resolve the host name might actually be better because Akka expects that messages arrive with the same hostname which it was configured with at start up time.

flink-yarn/src/main/java/org/apache/flink/yarn/AbstractYarnClusterDescriptor.java
```diff
@@ -513,12 +509,8 @@ private void validateClusterSpecification(ClusterSpecification clusterSpecificat
 			yarnApplication,
 			validClusterSpecification);
 
-		String host = report.getHost();
-		int port = report.getRpcPort();
-
-		// Correctly initialize the Flink config
-		flinkConfiguration.setString(JobManagerOptions.ADDRESS, host);
-		flinkConfiguration.setInteger(JobManagerOptions.PORT, port);
```
tillrohrmann: I'm not 100% sure that we can remove this update operation. A quick search revealed that the `ResultStore` uses the `JobManagerOptions.ADDRESS`. Since I could not rule out that we implicitly rely on the fact that the `flinkConfiguration` is shared, I would not remove this setting. Only if we are sure, ideally guarded by a test, then we should remove it.

flink-scala-shell/src/main/scala/org/apache/flink/api/scala/FlinkShell.scala
TisonKun: I don't know whether there is any user connect YARN cluster by Scala Shell. Theoretically in standalone ha(a.k.a. non ha) it is no difference with WebMonitor. But with ZooKeeper enabled it will retrieve `Dispatcher` address and later it should failed to send REST request to `Dispatcher`.

flink-scala-shell/src/main/scala/org/apache/flink/api/scala/FlinkShell.scala
TisonKun: Make sense, will update

flink-yarn/src/main/java/org/apache/flink/yarn/AbstractYarnClusterDescriptor.java
TisonKun: This configuration is passed to `RestClusterClient` and never pass to `ResultStore` IIUC.

An investigation of generally how we configure and use JM(DispatcherRMComponent) and WebMonitor host&port would be a worthy follow-up. I don't stick to remove this since it does no harm in production but given its outputs are only when constructs `RestClusterClient` in `createYarnClusterClient` we can ensure that it would not escape.

TisonKun:
@sunjincheng121 python profile fails as below. I have no idea on what happened. Could you please figure out why tests failed?

```
=================================== FAILURES ===================================
__________________ ExecutionConfigTests.test_equals_and_hash ___________________
self = <pyflink.common.tests.test_execution_config.ExecutionConfigTests testMethod=test_equals_and_hash>
    def test_equals_and_hash(self):
    
        config1 = ExecutionEnvironment.get_execution_environment().get_config()
    
        config2 = ExecutionEnvironment.get_execution_environment().get_config()
    
        self.assertEqual(config1, config2)
    
>       self.assertEqual(hash(config1), hash(config2))
E       AssertionError: 897378335 != 1596606912
pyflink/common/tests/test_execution_config.py:277: AssertionError
=============================== warnings summary ===============================
.tox/py37/lib/python3.7/site-packages/py4j/java_collections.py:13
.tox/py37/lib/python3.7/site-packages/py4j/java_collections.py:13
.tox/py37/lib/python3.7/site-packages/py4j/java_collections.py:13
.tox/py37/lib/python3.7/site-packages/py4j/java_collections.py:13
.tox/py37/lib/python3.7/site-packages/py4j/java_collections.py:13
  /home/travis/build/TisonKun/flink/flink-python/.tox/py37/lib/python3.7/site-packages/py4j/java_collections.py:13: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
    from collections import (
-- Docs: https://docs.pytest.org/en/latest/warnings.html
```

flink-yarn/src/main/java/org/apache/flink/yarn/AbstractYarnClusterDescriptor.java
TisonKun: Aha! Even I found a reason that we "need" these stmts because when build HAService at client-side we have to build one just like on cluster client thus if `JobManagerOption` not set it will fail with

```
org.apache.flink.client.deployment.ClusterDeploymentException: Could not deploy Yarn job cluster.
	at org.apache.flink.yarn.YARNITCase.lambda$testPerJobMode$0(YARNITCase.java:109)
	at org.apache.flink.yarn.YARNITCase.testPerJobMode(YARNITCase.java:69)
Caused by: org.apache.flink.util.ConfigurationException: Config parameter 'Key: 'jobmanager.rpc.address' , default: null (fallback keys: [])' is missing (hostname/address of JobManager to connect to).
	at org.apache.flink.yarn.YARNITCase.lambda$testPerJobMode$0(YARNITCase.java:109)
	at org.apache.flink.yarn.YARNITCase.testPerJobMode(YARNITCase.java:69)
```

this is what FLINK-13750 want to resolve, i.e., do not instance service that is never needed.

TisonKun: Reverted removals and create FLINK-13929 for further discussion.

TisonKun:
@tillrohrmann I think this pr is mergeable. Please take a look :-)

PR #9576, opened 2019-08-30
[FLINK-13915][ml] Add several base classes of summarizer.

xuyang1706:

Summarizer is the class for calculating statistics, summary is the result class of summarizer. Summary defines methods to get statistics. Statistics include count, mean, variance, min, max, standardDeviation, normL1 and normL2.

## Brief change log

- *Add BaseSummarizer which is the base class of summarizer.*
- *Add BaseSummary which is the base class of summary.*
- *Add BaseVectorSummarizer which defined functions to calculate statistics.*
- *Add BaseVectorSummary which defined functions to get statistics.*

## Verifying this change

- run test case pass

  - The serializers: (no)

## Documentation


flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/statistics/basicstatistic/BaseSummarizer.java
```diff
@@ -0,0 +1,62 @@
...
+ */
+
+package org.apache.flink.ml.common.statistics.basicstatistic;
+
+import org.apache.flink.ml.common.linalg.DenseMatrix;
+
+import java.io.Serializable;
+
+/**
+ * Summarizer is the base class to calculate summary and store , and Summary is the result of Summarizer.
```
walterddr: This needs more explanation.
```
Summarizer provides summary statistics about a given object (Matrix/Vector)
```

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/statistics/basicstatistic/BaseSummarizer.java
```diff
@@ -0,0 +1,62 @@
...
+ * It will compute sum, squareSum = sum(x_i*x_i), min, max, normL1.
```
walterddr: better to explain what the base class do. because this base class does not provide any of these functionalities.
```
Base Summarizer class holds basic stats (element count, outerproduct, etc) 
```

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/statistics/basicstatistic/BaseSummarizer.java
```diff
@@ -0,0 +1,62 @@
...
+ * All statistics can use summary to calculate from these statistics.
+ */
+public abstract class BaseSummarizer implements Serializable {
```
walterddr: Also higher level question: Given this class is very basic, I was wondering why we need this. If this is for interface sake. then better rework this as interface and explain what are the APIs to be supproted.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/statistics/basicstatistic/BaseSummarizer.java
walterddr: also. I don't see `normL1` anywhere in this diff.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/statistics/basicstatistic/BaseSummarizer.java
xuyang1706: Thanks, added.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/statistics/basicstatistic/BaseSummarizer.java
xuyang1706: Thanks, I have refined the JavaDoc.

xuyang1706: Thanks, we have refined the JavaDoc.

xuyang1706: Thanks, we have refined the JavaDoc.

xuyang1706:
> Thanks for the contribution @xuyang1706 . I think this PR lacks information for a successful review. Is it possible to combine with a concrete summarizer impl PR? for example: #9577

Thanks, @walterddr , I have rebased the [FLINK-13667][FLINK-13923][FLINK-13924] to this pr

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/statistics/basicstatistic/BaseSummarizer.java
```diff
@@ -0,0 +1,94 @@
...
+ */
+
+package org.apache.flink.ml.common.statistics.basicstatistic;
+
+import org.apache.flink.ml.common.linalg.DenseMatrix;
+
+import java.io.Serializable;
+
+/**
+ * Summarizer is the base class to calculate summary and store intermediate results, and Summary is the result of Summarizer.
```
walterddr: I actually have a higher level questions regarding the type of summary we are going to support and what's the road map in the future.

In general: a "summary" to me should be something that can be serialized into a format that an external system can read and use it to produce meaningful insights. such as visualization e.g. [tf.summary](https://www.tensorflow.org/api_docs/python/tf/summary), or for transforming into some object that can be used to export data [R.summary](https://www.rdocumentation.org/packages/base/versions/3.6.1/topics/summary). 

In this case, the class "summary" in this PR is more like a wrapper around: `sum()`, `count()`, `correlate()`, `covariance()`. I am not sure what's the value add to the ML-library. Maybe I was understanding this entire intention in the wrong way.

walterddr: To add to this discussion. Some of the summaries used by Tensorflow for example, is meant to be used by Tensorboard - which understands the format of the summary in order to plot visualization via [protobuf](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/framework/summary.proto)), same thing with [tf.Event](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/util/event.proto). 

If this is the case, is there an existing discussion regarding what needs to be done in terms of the public APIs of a `Summary` class should have in Flink-ML? I haven't follow up closely with the ML but I cant find any in the [FLIP-39 documentation](https://cwiki.apache.org/confluence/display/FLINK/FLIP-39+Flink+ML+pipeline+and+ML+libs)

xuyang1706: Tranks for your comments @, We just provide a simple way to view the statistics, similar to summarizer in SparkML. The value can be printed. For example,
```
Summarizer summarizer = new Summarizer(data)
  .setSelectedCols("f_double", "f_int");
TableSummary srt = summarizer.collectResult();
System.out.println(srt.mean("f_double”));
```

Summarizer is also a tool, which can be used by other algorithms.  Call mean () and standarddeviation () for the standardized data of logical regression.

xuyang1706:
> Hi @xuyang1706 @qiuxiafei, seems like after #9404 there are some compilation failures, could you take a look and rebase the change? thanks --Rong

Thanks, I have rebased the changes.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/operator/common/statistics/basicstatistic/BaseSummarizer.java
```diff
@@ -0,0 +1,94 @@
...
+ */
+
+package org.apache.flink.ml.operator.common.statistics.basicstatistic;
+
+import org.apache.flink.ml.common.linalg.DenseMatrix;
+
+import java.io.Serializable;
+
+/**
+ * Summarizer is the base class to calculate summary and store intermediate results, and Summary is the result of Summarizer.
```
walterddr:
```suggestion
 * BaseSummarizer is the base class to calculate summary and store intermediate results.
```

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/operator/common/statistics/basicstatistic/BaseSummarizer.java
```diff
@@ -0,0 +1,94 @@
...
+ *           /         \
+ * TableSummarizer   BaseVectorSummarizer
+ *                     /            \
+ *                    /              \
+ *     SparseVectorSummarizer    DenseVectorSummarizer
+ *
+ * <p>TableSummarizer is for table data, BaseVectorSummarizer is for vector data.
+ *  SparseVectorSummarizer is for sparse vector, DenseVectorSummarizer is for dense vector.
+ *
+ *  <p>Summary Inheritance relationship as follow:
```
walterddr:
```suggestion
 * <p>Summary is the result of Summarizer. {@link BaseSummary} is the base class of summaries.
 *
 * <p>Summary Inheritance relationship as follow:
```

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/operator/common/statistics/basicstatistic/BaseSummary.java
```diff
@@ -0,0 +1,73 @@
...
+ *           /         \
+ *  TableSummary     BaseVectorSummary
+ *                     /            \
+ *                    /              \
+ *      SparseVectorSummary    DenseVectorSummary
+ *
+ * <p>You can get statistics value from summary.
+ *
+ * <p>example:
+ * <pre>
```
walterddr: This is duplicate from `BaseSummarizer` I would suggest do something like this
```
/**
 * Summary is ....
 * 
 * <p>Summary Inheritance relationship as follow:
 * ...
 * @see BaseSummarizer
```

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/operator/common/statistics/basicstatistic/BaseSummarizer.java
walterddr: Also we can move Summary part to BaseSummary.java

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/operator/common/statistics/basicstatistic/BaseSummarizer.java
```diff
@@ -0,0 +1,94 @@
...
+ * <pre>
+ * {@code
+ *      Row data =  Row.of("a", 1L, 1, 2.0, true)
+ *      TableSummarizer summarizer = new TableSummarizer(selectedColNames, numberIdxs, bCov);
+ *      summarizer.visit(data);
+ *      TableSummary summary = summarizer.toSummary()
+ *      double mean = summary.mean("col")
+ * }
+ * </pre>
+ */
```
walterddr:
```suggestion
 * @see BaseSummary
 */
```

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/operator/common/statistics/basicstatistic/BaseVectorSummary.java
```diff
@@ -0,0 +1,153 @@
...
+ * <p>Summary Inheritance relationship as follow:
+ *            BaseSummary
+ *             /       \
+ *            /         \
+ *   TableSummary     BaseVectorSummary
+ *                     /            \
+ *                    /              \
+ *      SparseVectorSummary    DenseVectorSummary
+ *
+ * <p>It can use toSummary() to get the result BaseVectorSummary.
```
walterddr: This is redundant IMO. Simply this should work:
```
 *            BaseVectorSummary
 *            /            \
 *           /              \
 *  SparseVectorSummary    DenseVectorSummary
```

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/operator/common/statistics/basicstatistic/CorrelationResult.java
```diff
@@ -0,0 +1,68 @@
...
+import org.apache.flink.ml.common.linalg.DenseMatrix;
+import org.apache.flink.ml.common.utils.TableUtil;
+import org.apache.flink.types.Row;
+
+import java.util.ArrayList;
+
+/**
+ * It is correlation result, which has colNames and correlation values.
+ */
+public class CorrelationResult {
```
walterddr: This class seems to be a fit for an inner class of basesummarizer/basesummary ? I do not see this class being used outside of the summary perspective. 

Althought the toString() part can be a Utility function for DenseMatrix IMO and can be factor out?

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/operator/common/statistics/basicstatistic/BaseVectorSummary.java
```diff
@@ -0,0 +1,153 @@
...
+ * {@code
+ *   DenseVector data = new DenseVector(new double[]{1.0, -1.0, 3.0})
+ *   DenseVectorSummarizer summarizer = new DenseVectorSummarizer(false);
+ *   summarizer = summarizer.visit(data);
+ *   BaseVectorSummary summary = summarizer.toSummary()
+ *   double mean = summary.mean(0)
+ *}
+ *</pre>
+ */
+public abstract class BaseVectorSummary extends BaseSummary {
```
walterddr: Missing test cases for Vector Summary, could you add?

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/operator/common/statistics/basicstatistic/TableSummarizer.java
```diff
@@ -0,0 +1,404 @@
...
+import org.apache.flink.ml.common.linalg.VectorUtil;
+import org.apache.flink.types.Row;
+
+import java.util.Arrays;
+
+/**
+ * It is summarizer for table， it will calculate statistics and return TableSummary.
+ * You can get statistics from Summary.
+ */
+public class TableSummarizer extends BaseSummarizer {
```
walterddr: Can we factor out Table part into a separated PR? seems like they should belong to [FLINK-13923](https://issues.apache.org/jira/browse/FLINK-13923)

walterddr:
@xuyang1706 @qiuxiafei can you guys take another look at moving the Table part to #9577 ?
Otherwise I think the base part is good to go: https://github.com/walterddr/flink/commit/f812b99a74560b1d534eee50141335920618684a

PR #9581, opened 2019-08-30
[FLINK-13864][streaming]: Modify the StreamingFileSink Builder interface to allow for easier subclassing of StreamingFileSink 

yxu-valleytider:

This patch makes it easier to extend `StreamingFileSink`, which adopts the builder pattern as a way to construct a new `StreamingFileSink` object. It follows the technique of "recurring generic pattern" which is documented in [[1](https://community.oracle.com/blogs/emcmanus/2010/10/24/using-builder-pattern-subclasses)] and [[2](https://stackoverflow.com/questions/17164375/subclassing-a-java-builder-class)].  

## Brief change log

**StreamingFileSink**
  - Added the builder type as the class signature for `BucketsBuilder`, `RowFormatBuilder` and `BulkFormatBuilder`. 
  - Added a `self()` method inside `BucketsBuilder` to return the proper builder type for base classes and subclasses in the path of inheritance. 
  - Change the construction methods to return the builder object (`self()`), instead of a new instance of the builder.  
  - Removed `final` for a few variables as they are now set by the construction methods. 

## Verifying this change

This change is merely an interface change and is already covered by existing tests, such as *LocalStreamingFileSinkTest* and *BulkWriterTest*.

  - The serializers: (no)
  - The S3 file system connector: (yes) it affects *StreamingFileSink* but merely an interface change.

## Documentation

### DIFF
--- a/flink-streaming-java/src/main/java/org/apache/flink/streaming/api/functions/sink/filesystem/StreamingFileSink.java
+++ b/flink-streaming-java/src/main/java/org/apache/flink/streaming/api/functions/sink/filesystem/StreamingFileSink.java
@@ -110,7 +110,7 @@ public class StreamingFileSink<IN>
 
 	private final long bucketCheckInterval;
 
-	private final StreamingFileSink.BucketsBuilder<IN, ?> bucketsBuilder;
+	private final StreamingFileSink.BucketsBuilder<IN, ?, ? extends BucketsBuilder<IN, ?, ?>> bucketsBuilder;
 
 	// --------------------------- runtime fields -----------------------------
 
@@ -125,11 +125,24 @@ public class StreamingFileSink<IN>
 	private transient ListState<Long> maxPartCountersState;
 
 	/**
-	 * Creates a new {@code StreamingFileSink} that writes files to the given base directory.
+	 * Creates a new {@code StreamingFileSink} that writes files in row-based format to the given base directory.
 	 */
 	protected StreamingFileSink(
-			final StreamingFileSink.BucketsBuilder<IN, ?> bucketsBuilder,
-			final long bucketCheckInterval) {
+		final RowFormatBuilder<IN, ?, ? extends BucketsBuilder<IN, ?, ?>> bucketsBuilder,
+		final long bucketCheckInterval) {
+
+		Preconditions.checkArgument(bucketCheckInterval > 0L);
+
+		this.bucketsBuilder = Preconditions.checkNotNull(bucketsBuilder);
+		this.bucketCheckInterval = bucketCheckInterval;
+	}
+
+	/**
+	 * Creates a new {@code StreamingFileSink} that writes files in bulk-encoded format to the given base directory.
+	 */
+	protected StreamingFileSink(
+		final BulkFormatBuilder<IN, ?, ? extends BucketsBuilder<IN, ?, ?>> bucketsBuilder,
+		final long bucketCheckInterval) {
 
 		Preconditions.checkArgument(bucketCheckInterval > 0L);
 
@@ -149,7 +162,7 @@ public class StreamingFileSink<IN>
 	 * @return The builder where the remaining of the configuration parameters for the sink can be configured.
 	 * In order to instantiate the sink, call {@link RowFormatBuilder#build()} after specifying the desired parameters.
 	 */
-	public static <IN> StreamingFileSink.RowFormatBuilder<IN, String> forRowFormat(
+	public static <IN> StreamingFileSink.RowFormatBuilder<IN, String, ? extends RowFormatBuilder<IN, String, ?>> forRowFormat(
 			final Path basePath, final Encoder<IN> encoder) {
 		return new StreamingFileSink.RowFormatBuilder<>(basePath, encoder, new DateTimeBucketAssigner<>());
 	}
@@ -162,7 +175,7 @@ public class StreamingFileSink<IN>
 	 * @return The builder where the remaining of the configuration parameters for the sink can be configured.
 	 * In order to instantiate the sink, call {@link RowFormatBuilder#build()} after specifying the desired parameters.
 	 */
-	public static <IN> StreamingFileSink.BulkFormatBuilder<IN, String> forBulkFormat(
+	public static <IN> StreamingFileSink.BulkFormatBuilder<IN, String, ? extends BulkFormatBuilder<IN, String, ?>> forBulkFormat(
 			final Path basePath, final BulkWriter.Factory<IN> writerFactory) {
 		return new StreamingFileSink.BulkFormatBuilder<>(basePath, writerFactory, new DateTimeBucketAssigner<>());
 	}
@@ -170,9 +183,17 @@ public class StreamingFileSink<IN>
 	/**
 	 * The base abstract class for the {@link RowFormatBuilder} and {@link BulkFormatBuilder}.
 	 */
-	protected abstract static class BucketsBuilder<IN, BucketID> implements Serializable {
+	private abstract static class BucketsBuilder<IN, BucketID, T extends BucketsBuilder<IN, BucketID, T>> implements
+		Serializable {
 
 		private static final long serialVersionUID = 1L;
+
+		protected static final long DEFAULT_BUCKET_CHECK_INTERVAL = 60L * 1000L;
+
+		@SuppressWarnings("unchecked")
+		protected T self() {
+			return (T) this;
+		}
 
 		abstract Buckets<IN, BucketID> createBuckets(final int subtaskIndex) throws IOException;
 	}
@@ -181,31 +202,31 @@ public class StreamingFileSink<IN>
 	 * A builder for configuring the sink for row-wise encoding formats.
 	 */
 	@PublicEvolving
-	public static class RowFormatBuilder<IN, BucketID> extends StreamingFileSink.BucketsBuilder<IN, BucketID> {
+	public static class RowFormatBuilder<IN, BucketID, T extends RowFormatBuilder<IN, BucketID, T>> extends StreamingFileSink.BucketsBuilder<IN, BucketID, T> {
 
 		private static final long serialVersionUID = 1L;
 
-		private final long bucketCheckInterval;
+		private long bucketCheckInterval;
 
 		private final Path basePath;
 
-		private final Encoder<IN> encoder;
+		private Encoder<IN> encoder;
 
-		private final BucketAssigner<IN, BucketID> bucketAssigner;
+		private BucketAssigner<IN, BucketID> bucketAssigner;
 
-		private final RollingPolicy<IN, BucketID> rollingPolicy;
+		private RollingPolicy<IN, BucketID> rollingPolicy;
 
-		private final BucketFactory<IN, BucketID> bucketFactory;
+		private BucketFactory<IN, BucketID> bucketFactory;
 
-		private final String partFilePrefix;
+		private String partFilePrefix;
 
-		private final String partFileSuffix;
+		private String partFileSuffix;
 
-		RowFormatBuilder(Path basePath, Encoder<IN> encoder, BucketAssigner<IN, BucketID> bucketAssigner) {
-			this(basePath, encoder, bucketAssigner, DefaultRollingPolicy.builder().build(), 60L * 1000L, new DefaultBucketFactoryImpl<>(), PartFileConfig.DEFAULT_PART_PREFIX, PartFileConfig.DEFAULT_PART_SUFFIX);
+		protected RowFormatBuilder(Path basePath, Encoder<IN> encoder, BucketAssigner<IN, BucketID> bucketAssigner) {
+			this(basePath, encoder, bucketAssigner, DefaultRollingPolicy.builder().build(), DEFAULT_BUCKET_CHECK_INTERVAL, new DefaultBucketFactoryImpl<>(), PartFileConfig.DEFAULT_PART_PREFIX, PartFileConfig.DEFAULT_PART_SUFFIX);
 		}
 
-		private RowFormatBuilder(
+		protected RowFormatBuilder(
 				Path basePath,
 				Encoder<IN> encoder,
 				BucketAssigner<IN, BucketID> assigner,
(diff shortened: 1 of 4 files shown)


yxu-valleytider:
**Example subclass interface**

```
public class CustomizedStreamingFileSink<INT> extends StreamingFileSink<INT> {

  private CustomizedStreamingFileSink(
    final BulkFormatBuilder<INT, ?, ?> bucketsBuilder, final long bucketCheckInterval) {
    super(bucketsBuilder, bucketCheckInterval);
    ...
  }

  // Retrieve a new bulk format builder 
  public static <INT> StreamingFileSink.BulkFormatBuilder<INT, String,
      ? extends CustomizedBulkFormatBuilder<INT, String, ?>> forBulkFormat(Path basePath, Factory<INT> writerFactory) {
    ...
    return new CustomizedBulkFormatBuilder(...)
  }

  .
  .
  .

  public static class CustomizedBulkFormatBuilder<INT, BucketIDT, TT extends CustomizedBulkFormatBuilder<INT, BucketIDT, TT>> extends StreamingFileSink.BulkFormatBuilder<INT, BucketIDT, TT> {
    public CustomizedBulkFormatBuilder(
        Path basePath, Factory<INT> writerFactory, BucketAssigner<INT, BucketIDT> assigner) {
      super(basePath, writerFactory, assigner);
    }

    public CustomizedStreamingFileSink<INT> build() {
      return new CustomizedStreamingFileSink(this, bucketCheckInterval);
    }
  }
}
```

One alternative is to make `private` members in `StreamingFileSink.BulkFormatBuilder` (or `StreamingFileSink.RowFormatBuilder`) `public` or `protected`, which is more hackish.

yxu-valleytider:
PTAL @kailashhd @tweise

yxu-valleytider:
PTAL @kl0u

kl0u:
Thanks for the example! I am having a look now @yxu-valleytider and @kailashhd

kl0u:
@yxu-valleytider in the example you present what you want to be able to do with the changes in terms of code. But can you describe the functionality that you think is missing from the `StreamingFileSink` and makes you want to subclass it instead of providing an alternative rolling policy for example or writer factory? 

In general I think that the idea of making the `StreamingFileSink` extendable is good. I am just wondering what additional functionality users may want.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/functions/sink/filesystem/StreamingFileSink.java
```diff
@@ -314,25 +332,30 @@ private BulkFormatBuilder(
...
-		public StreamingFileSink.BulkFormatBuilder<IN, BucketID> withBucketCheckInterval(long interval) {
-			return new BulkFormatBuilder<>(basePath, writerFactory, bucketAssigner, interval, bucketFactory, partFilePrefix, partFileSuffix);
+		public T withBucketCheckInterval(long interval) {
+			this.bucketCheckInterval = interval;
+			return self();
 		}
 
-		public <ID> StreamingFileSink.BulkFormatBuilder<IN, ID> withBucketAssigner(BucketAssigner<IN, ID> assigner) {
-			return new BulkFormatBuilder<>(basePath, writerFactory, Preconditions.checkNotNull(assigner), bucketCheckInterval, new DefaultBucketFactoryImpl<>(), partFilePrefix, partFileSuffix);
+		public T withBucketAssigner(BucketAssigner<IN, BucketID> assigner) {
```
kl0u: Here you are missing the parameter type `ID`. 

As described also in the comments here https://issues.apache.org/jira/browse/FLINK-13843, this method for the bulk formats and the `withBucketAssignerAndPolicy ()` for the row formats, are the ones that allow the user to change the default type of the `BucketID`, so they should stay as they are.

kl0u: @yxu-valleytider  Yes, this is why I am mentioned it here. The same holds for the `rowFormat` builder above where you explicitly cast. I am not so sure if this approach would work and we may need to investigate this further.

kl0u:
If you simply want to extend the `StreamingFileSink`, is it really necessary to also be able to extend the builders?

yxu-valleytider:
> If you simply want to extend the `StreamingFileSink`, is it really necessary to also be able to extend the builders?

The primary reason is our use case needs customized bucket assigner to construct the extended class, and we would like to follow the existing construction pattern. The application logic looks like: 
```
CustomizedStreamingFileSink.forBulkFormat(...)
            .withBucketAssigner(new CustomizedBucketAssigner())
            .build();
```
If we don't override the builder in the way this PR does, `withBucketAssigner()` would return a `StreamingFileSink.BulkFormatBuilder` and  `build()` would return a parent `StreamingFileSink` class but not the `CustomizedStreamingFileSink` class. 

The linked articles discussed how to perform subclassing for a parent class using the builder pattern. This PR follows suggestions there on the best practice.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/functions/sink/filesystem/StreamingFileSink.java
yxu-valleytider: @kl0u Did a small experiment with a non-string BucketID assigner :    
```
class NonStringBucketIdAssigner implements BucketAssigner<Message, Integer> {
    ...
}
```
and 
```
CustomizedStreamingFileSink.forBulkFormat(...)
            .withBucketAssigner(new NonStringBucketIdAssigner())
            .build();
```

The casting appears to be successful as long as the `assigner` parameter can be casted to `BucketAssigner`.  It makes sense considering the effect of unbounded type erasure.  

Also updating the PR to reflect this in the casting statement.   If the casting is truly a concern we could keep the original method with different method names, e.g., `newBuilderWithBucketAssigner`.

yxu-valleytider:
PTAL @kl0u

tweise:
@kl0u could you please comment if this solution is acceptable? We would ideally like to wrap this up and use the same patch in our fork.

kl0u:
I will have a look first thing in the morning.

In general changes look good, but I would like to have another pass. Also I think it is worth adding a test that verifies that we can change the type of the `BucketID` without problems. Essentially testing the `withBucketAssigner()` and the corresponding `withBucketAssignerAndPolicy()` for the `rowFormat` builder. This is mainly to guarantee that nobody will break it in the future. Such a test should have been added earlier when the initial `StreamingFileSink` was written :( .

yxu-valleytider:
Hi @kl0u thanks a lot for the suggestions. 

Did more investigations when writing tests. I suspect there could be potential type clashing if `bucketFactory` and `bucketAssigner` have different `BucketID` type.  For example:

```
TestBucketFactory<IN, String> bucketFactory;
TestBucketAssigner<IN, Integer> bucketAssigner;

StreamingFileSink.forBulkFormat(path, writer)
			.withBucketFactory(bucketFactory)
			.withBucketAssigner(bucketAssigner)
			.build();
``` 

Hence in the [new commit](https://github.com/apache/flink/pull/9581/commits/df82b55379806ac8502ef92999a1abc7f9a0056b) I'm putting back the old methods, which construct a new builder when altering the `BucketID` type.  Changed the function names to reflect that explicitly.  If a subclass also wants this capability, it has to override these methods (to return the correct subclass type).  

I can put up a few more tests for the new functions. Do you think they can be in a separate PR ? Thanks.


> I will have a look first thing in the morning.
> 
> In general changes look good, but I would like to have another pass. Also I think it is worth adding a test that verifies that we can change the type of the `BucketID` without problems. Essentially testing the `withBucketAssigner()` and the corresponding `withBucketAssignerAndPolicy()` for the `rowFormat` builder. This is mainly to guarantee that nobody will break it in the future. Such a test should have been added earlier when the initial `StreamingFileSink` was written :( .

kl0u:
Hi @yxu-valleytider , I am reviewing the PR now.
The tests should be also part of the PR as this PR adds a new feature (subclassing the `StreamingFileSink`).

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/functions/sink/filesystem/StreamingFileSink.java
```diff
@@ -162,47 +162,54 @@ protected StreamingFileSink(
...
+	public static <IN> StreamingFileSink.BulkFormatBuilder<IN, String, ? extends BulkFormatBuilder<IN, String, ?>> forBulkFormat(
 			final Path basePath, final BulkWriter.Factory<IN> writerFactory) {
 		return new StreamingFileSink.BulkFormatBuilder<>(basePath, writerFactory, new DateTimeBucketAssigner<>());
 	}
 
 	/**
 	 * The base abstract class for the {@link RowFormatBuilder} and {@link BulkFormatBuilder}.
 	 */
-	protected abstract static class BucketsBuilder<IN, BucketID> implements Serializable {
+	protected abstract static class BucketsBuilder<IN, BucketID, T extends BucketsBuilder<IN, BucketID, T>> implements Serializable {
```
kl0u: This class should become `private` as we do not expect users to subclass it directly.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/functions/sink/filesystem/StreamingFileSink.java
```diff
@@ -162,47 +162,54 @@ protected StreamingFileSink(
...
 	 * A builder for configuring the sink for row-wise encoding formats.
 	 */
 	@PublicEvolving
-	public static class RowFormatBuilder<IN, BucketID> extends StreamingFileSink.BucketsBuilder<IN, BucketID> {
+	public static class RowFormatBuilder<IN, BucketID, T extends RowFormatBuilder<IN, BucketID, T>> extends StreamingFileSink.BucketsBuilder<IN, BucketID, T> {
 
 		private static final long serialVersionUID = 1L;
 
-		private final long bucketCheckInterval;
+		protected long bucketCheckInterval;
```
kl0u: This can be `private`

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/functions/sink/filesystem/StreamingFileSink.java
```diff
@@ -162,47 +162,54 @@ protected StreamingFileSink(
...
-		private final String partFileSuffix;
+		private String partFileSuffix;
 
-		RowFormatBuilder(Path basePath, Encoder<IN> encoder, BucketAssigner<IN, BucketID> bucketAssigner) {
-			this(basePath, encoder, bucketAssigner, DefaultRollingPolicy.builder().build(), 60L * 1000L, new DefaultBucketFactoryImpl<>(), PartFileConfig.DEFAULT_PART_PREFIX, PartFileConfig.DEFAULT_PART_SUFFIX);
+		protected RowFormatBuilder(Path basePath, Encoder<IN> encoder, BucketAssigner<IN, BucketID> bucketAssigner) {
+			this(basePath, encoder, bucketAssigner, DefaultRollingPolicy.builder().build(), DEFAULT_BUCKET_CHECK_INTERVAL, new DefaultBucketFactoryImpl<>(), PartFileConfig.DEFAULT_PART_PREFIX, PartFileConfig.DEFAULT_PART_SUFFIX);
 		}
 
 		private RowFormatBuilder(
```
kl0u: This should also become `protected` because users will need it if they want to override the `newBuilderWithBucketAssignerAndPolicy()`.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/functions/sink/filesystem/StreamingFileSink.java
```diff
@@ -224,35 +231,52 @@ private RowFormatBuilder(
...
 
-		public StreamingFileSink.RowFormatBuilder<IN, BucketID> withPartFileSuffix(final String partSuffix) {
-			return new RowFormatBuilder<>(basePath, encoder, bucketAssigner, rollingPolicy, bucketCheckInterval, bucketFactory, partFilePrefix, partSuffix);
+		public T withPartFileSuffix(final String partSuffix) {
+			this.partFileSuffix = partSuffix;
+			return self();
+		}
+
+		public <ID> StreamingFileSink.RowFormatBuilder<IN, ID, ? extends RowFormatBuilder<IN, ID, ?>> newBuilderWithBucketAssignerAndPolicy(final BucketAssigner<IN, ID> assigner, final RollingPolicy<IN, ID> policy) {
+			return new RowFormatBuilder<>(basePath, encoder, Preconditions.checkNotNull(assigner), Preconditions.checkNotNull(policy), bucketCheckInterval, new DefaultBucketFactoryImpl<>(), partFilePrefix, partFileSuffix);
```
kl0u: We can add a  check `checkState(bucketFactory.getClass() == DefaultBucketFactoryImpl.class, "withBucketAssignerAndPolicy() should be called before specifying a custom BucketFactory.")`. Currently there is no way to change the `bucketFactory` but it could be added in the future.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/functions/sink/filesystem/StreamingFileSink.java
```diff
@@ -264,37 +288,32 @@ private RowFormatBuilder(
...
-			return new RowFormatBuilder<>(basePath, encoder, bucketAssigner, rollingPolicy, bucketCheckInterval, Preconditions.checkNotNull(factory), partFilePrefix, partFileSuffix);
-		}
 	}
 
 	/**
 	 * A builder for configuring the sink for bulk-encoding formats, e.g. Parquet/ORC.
 	 */
 	@PublicEvolving
-	public static class BulkFormatBuilder<IN, BucketID> extends StreamingFileSink.BucketsBuilder<IN, BucketID> {
+	public static class BulkFormatBuilder<IN, BucketID, T extends BulkFormatBuilder<IN, BucketID, T>> extends StreamingFileSink.BucketsBuilder<IN, BucketID, T> {
```
kl0u: Same comments as for the `RowFormatBuilder`.

yxu-valleytider:
@kl0u Agree with most of the review comments. Just want to re-iterate the main reason why introducing a third generics parameter in `RowFormatBuilder` and `BulkFormatBuilder` is due to the above https://github.com/apache/flink/pull/9581#issuecomment-527534328.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/functions/sink/filesystem/StreamingFileSink.java
yxu-valleytider: DONE.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/functions/sink/filesystem/StreamingFileSink.java
yxu-valleytider: DONE.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/functions/sink/filesystem/StreamingFileSink.java
yxu-valleytider: DONE.

yxu-valleytider:
PTAL @kl0u

kl0u:
Thanks a lot @yxu-valleytider . I will have a look.

kl0u:
Hi @yxu-valleytider , I still have the following concerns that make me feel uncomfortable about merging this PR:
1) The user should know that he has to write his own `forRowFormat` / `forBulkFormat` in order for them to return the correct builder, and this can only be "implicit" knowledge.
2) The user should override the `build` method of the builder he provides. 
3) It may break user-code, if the user was explicitly stating in his program the return type of the `forRowFormat` or `forBulkFormat`. This is because we add a new type parameter.

Also some other issues that can be tackled easily though are:
1) The constructor should not take a `BucketsBuilder` but we have to have 2 constructors, one for row and one for bulk. If not, then the user may create an invalid `StreamingFileSink`.
2) The user cannot change the type of the `BucketID` (although this may not be that important).

kl0u:
I think I need more time to see what can be done and how we can find an elegant solution.

aljoscha:
This PR seems very contentious because it makes the code of the sink somewhat more complicated.

Taking a step back, could we look at alternative options? You mentioned that the reason for extending this is that you want to extend metrics support. Would a metrics extension point help here, i.e. we add a `MetricsExtension` (name TBD) that we can provide to the `StreamingFileSink` and that is invoked at certain points and the implementation can spit out extended metrics.

yxu-valleytider:
@aljoscha  The metrics collection needs be done inside `notifyCheckpointComplete()` and `snapshotState()`.  In addition, we may add customized logic into these two function as well.  Some of the logic are specific to our use cases, and probably better to stay inside our own customized `StreamingFileSink`.  I'm following the cited articles' suggestion in order to make the builder pattern extensible to the subclasses.

yxu-valleytider:
> Hi @yxu-valleytider , I still have the following concerns that make me feel uncomfortable about merging this PR:
> 
> 1. The user should know that he has to write his own `forRowFormat` / `forBulkFormat` in order for them to return the correct builder, and this can only be "implicit" knowledge.
> 2. The user should override the `build` method of the builder he provides.
> 3. It may break user-code, if the user was explicitly stating in his program the return type of the `forRowFormat` or `forBulkFormat`. This is because we add a new type parameter.

@kl0u I can see  3) is more or less a concern.  This PR does introduce quite a bit of changes to the interface of the StreamingFileSink -- the original hope is there are not so many users of `StreamingFileSink` yet hence impact is limited. 

> Also some other issues that can be tackled easily though are:
> 
> 1. The constructor should not take a `BucketsBuilder` but we have to have 2 constructors, one for row and one for bulk. If not, then the user may create an invalid `StreamingFileSink`.
> 2. The user cannot change the type of the `BucketID` (although this may not be that important).

@aljoscha In terms of a step back, it would something extending [this branch](https://github.com/apache/flink/compare/master...kailashhd:FLINK-12539).  If you are OK with adding a bunch of getter functions to retrieve the `private` member fields of the class.

yxu-valleytider:
> @yxu-valleytider I think we can merge this PR as is for now and re-iterate within this release cycle some of the details that need to be ironed out.

Thanks @kl0u  for the review, much appreciated!   Yes definitely open to further improvements/consolidations around the `StreamingFileSink` interface.

yxu-valleytider:
🚀

tweise:
@kl0u thanks for working through this. Should the constructor issue be addressed before merging or you prefer to take it up separately?

@yxu-valleytider please rebase your branch and squash.

kl0u:
@tweise It would be nice if it can be addressed before merging. I was thinking that I will be the one to merge this PR so I was planning to do it then but if you are willing to do it, it would be much appreciated if you could add the two separate constructors (one for row and one for bulk) before merging.

yxu-valleytider:
@tweise Did rebase and squash 

@kl0u Done splitting the constructor into two: one for row-based and the other for bulk-encoded. Feel free to modify in case there's anything missing.

kl0u:
Merged

PR #9595, opened 2019-09-02
[FLINK-13586] Make ClosureCleaner.clean() backwards compatible with 1.8.0

aljoscha:

We revert earlier changes that make (among other things) connectors incompatible between different 1.8.x versions. See https://issues.apache.org/jira/browse/FLINK-13586 for a more thorough description

## Brief change log

 - introduce a new method `ClosureCleaner.clean()` that has the same signature as the 1.8.0 method.
 - change call sites in connectors to use the "new" method

## Verifying this change

 - This is covered by existing tests, the default of `RECURSIVE` is still in place
 - actually verifying that we don't break binary compatibility in this way again is prohibitively difficult, IMO

  - Technically, this does affect public API but only because the connectors internally use `ClosureCleaner`

## Documentation

  - No new feature, and no documentation


StephanEwen:
+1 looks correct to me

PR #9596, opened 2019-09-02
[FLINK-13445] Add 'jobmanager.containerized.heap-cutoff-{min,ratio}' …

dmvk:

* Distinguish Memory Configuration for TaskManager and JobManager 

## Brief change log

*(for example:)*
  - Allow to override jobmanager containerized.heap-cutoff-{min,ratio}

## Verifying this change

*added test for config fallback

  - The serializers: (no)

## Documentation

### DIFF
--- a/docs/_includes/generated/resource_manager_configuration.html
+++ b/docs/_includes/generated/resource_manager_configuration.html
@@ -18,6 +18,16 @@
             <td>Percentage of heap space to remove from containers (YARN / Mesos), to compensate for other JVM memory usage.</td>
         </tr>
         <tr>
+            <td><h5>jobmanager.containerized.heap-cutoff-min</h5></td>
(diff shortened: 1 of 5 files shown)


dmvk:
@tillrohrmann would you have time to look at this one? Thanks

D.

StephanEwen:
As mentioned in the JIRA discussion, we may want to solve this a bit differently.
for the TM, we are moving away from this "cutoff" value, because it seems not flexible/specific enough.

We may want to do the same on the JM as well.

aljoscha:
@StephanEwen @dmvk @tillrohrmann what's the state on this? Do we close the PR since we want to solve it a different way?

dmvk:
@aljoscha closing this as there was no consensus on jira. Can you please close the jira as well?

PR #9598, opened 2019-09-03
[FLINK-13633][coordination] Move submittedJobGraph and completedCheckpoint to cluster-id subdirectory of high-availability storage

wangyang0918:

This pull request moves submittedJobGraph and completedCheckpoint to cluster-id subdirectory of high-availability storage. If the flink cluster terminates exceptionally, some external tools could be used to clean up these residual files.

## Brief change log

Use cluster-id sub directory instead of ha storage in `ZookeeperUtils#createCompletedCheckpoints()` and `ZookeeperUtils#createJobGraphs()`.

## Verifying this change

  - Added integration tests `ZooKeeperHaStorageITCase` to check the high availability storage directory structure.

## Documentation

### DIFF
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/blob/BlobUtils.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/blob/BlobUtils.java
@@ -22,10 +22,9 @@ import org.apache.flink.api.common.JobID;
 import org.apache.flink.configuration.BlobServerOptions;
 import org.apache.flink.configuration.Configuration;
 import org.apache.flink.configuration.ConfigurationUtils;
-import org.apache.flink.configuration.HighAvailabilityOptions;
-import org.apache.flink.configuration.IllegalConfigurationException;
 import org.apache.flink.core.fs.FileSystem;
 import org.apache.flink.core.fs.Path;
+import org.apache.flink.runtime.highavailability.HighAvailabilityServicesUtils;
 import org.apache.flink.runtime.jobmanager.HighAvailabilityMode;
 import org.apache.flink.util.InstantiationUtil;
 import org.apache.flink.util.StringUtils;
@@ -46,8 +45,6 @@ import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 import java.util.Random;
 import java.util.UUID;
-
-import static org.apache.flink.util.StringUtils.isNullOrWhitespaceOnly;
 
 /**
  * Utility class to work with blob data.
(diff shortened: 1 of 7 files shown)


flink-tests/pom.xml
```diff
@@ -189,6 +189,22 @@ under the License.
 			<scope>test</scope>
 		</dependency>
 
+		<dependency>
```
TisonKun: I notice you mentioned 

>Dependencies (does it add or upgrade a dependency): (yes / **no**)

in pull request description but it actually does upgrade a dependency. Could you please explain why this upgrade needed and correspondingly update the description?

flink-runtime/src/main/java/org/apache/flink/runtime/util/ZooKeeperUtils.java
```diff
@@ -354,13 +362,23 @@ public static ZooKeeperCheckpointIDCounter createCheckpointIDCounter(
...
+	 */
+	private static String getClusterHighAvailabilityStoragePath(Configuration configuration) {
 		String rootPath = configuration.getValue(HighAvailabilityOptions.HA_STORAGE_PATH);
 
 		if (rootPath == null || StringUtils.isBlank(rootPath)) {
 			throw new IllegalConfigurationException("Missing high-availability storage path for metadata." +
 					" Specify via configuration key '" + HighAvailabilityOptions.HA_STORAGE_PATH + "'.");
 		} else {
-			return new FileSystemStateStorageHelper<T>(rootPath, prefix);
+			final String clusterId = configuration.getValue(HighAvailabilityOptions.HA_CLUSTER_ID);
```
TisonKun: What if `HighAvailabilityOptions.HA_CLUSTER_ID` configured wrongly? Shall we perform a runtime checker to guard `clusterId` is valid?

TisonKun:
Any updates?

Also ping @azagrebin could you take a look at this?

flink-tests/pom.xml
wangyang0918: We have to add the hdfs related dependencies to use MiniDFSCluster in the tests. And the scope is only for test.

flink-runtime/src/main/java/org/apache/flink/runtime/util/ZooKeeperUtils.java
wangyang0918: What do you mean configured wrongly? In standalone mode, it is configured by user. In Yarn/mesos mode, it will be set automatically. Also it have default value and will not be null.

TisonKun: Yes your right. It's my mistake.

flink-tests/pom.xml
TisonKun: >For the hdfs related dependencies, we have to add to use MiniDFSCluster in the tests. And the scope is only for test.

ok I see

> I think they stay in ZooKeeperHaServices is more reasonable.

make sense to me

tillrohrmann:
I've pushed an update which simplifies the test and makes sure that `BlobUtils` also uses the cluster wide high available storage directory. Please take a look @wangyang0918 and @TisonKun.

flink-runtime/src/main/java/org/apache/flink/runtime/blob/BlobUtils.java
```diff
@@ -23,9 +23,9 @@
 import org.apache.flink.configuration.Configuration;
 import org.apache.flink.configuration.ConfigurationUtils;
 import org.apache.flink.configuration.HighAvailabilityOptions;
```
TisonKun: Unused import that checkstyle complains with

tillrohrmann: Good catch. Will fix it.

tillrohrmann:
I've updated the PR to remove the unused import.

flink-runtime/src/main/java/org/apache/flink/runtime/util/ZooKeeperUtils.java
TisonKun: I think it is done partially because Till check the clusterId is not empty or whitespace-only. I remember my concern here previous that if the path concatenate is not a valid path it might fail later.

tillrohrmann:
Travis passed. I will merge this PR now. Thanks a lot for the initial authoring @wangyang0918 and the review @TisonKun.

PR #9601, opened 2019-09-03
[FLINK-13894][web]Web Ui add log url for subtask of vertex

jinglining:
This pull request adds log link for subtask in the page of job's vertex. Then it's easy for user seeing log.

## Brief change log
- The JobSubTaskInterface in job-subtask.ts add resource-id .
- job-overview-drawer-subtasks.component.html add log logic.

## Verifying this change
no

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-runtime-web/web-dashboard/src/app/pages/job/overview/subtasks/job-overview-drawer-subtasks.component.html
+++ b/flink-runtime-web/web-dashboard/src/app/pages/job/overview/subtasks/job-overview-drawer-subtasks.component.html
@@ -26,8 +26,7 @@
   [nzShowPagination]="false">
   <thead (nzSortChange)="sort($event)" nzSingleSort>
     <tr>
-      <th nzWidth="35px" nzLeft="0px">ID</th>
-      <th nzWidth="95px" nzLeft="35px">TaskManager</th>
+      <th nzWidth="80px" nzLeft="0px">ID</th>
       <th nzSortKey="metrics.read-bytes" nzShowSort nzWidth="140px">Bytes Received</th>
       <th nzSortKey="metrics.read-records" nzShowSort nzWidth="150px">Records Received</th>
       <th nzSortKey="metrics.write-bytes" nzShowSort nzWidth="120px">Bytes Sent</th>
@@ -37,19 +36,14 @@
       <th nzSortKey="start_time" nzShowSort nzWidth="150px">Start Time</th>
       <th nzSortKey="duration" nzShowSort nzWidth="100px">Duration</th>
       <th nzSortKey="end-time" nzShowSort nzWidth="150px">End Time</th>
-      <th nzSortKey="status" nzShowSort nzWidth="120px" nzRight="0px">Status</th>
+      <th nzSortKey="status" nzShowSort nzWidth="120px" nzRight="50px">Status</th>
+      <th nzWidth="50px" nzRight="0px">More</th>
     </tr>
   </thead>
   <tbody>
     <tr *ngFor="let task of listOfTask; trackBy:trackTaskBy;">
       <td nzLeft="0">
         {{ task.subtask }}
-      </td>
-      <td nzLeft="35px">
-        <span *ngIf="!task['taskmanager-id'] || task['taskmanager-id'] === '(unassigned)'; else hrefTpl">-</span>
-        <ng-template #hrefTpl>
-          <a [routerLink]="['/task-manager',task['taskmanager-id'],'metrics']" target="_blank">TaskManager</a>
-        </ng-template>
       </td>
       <td>
         <span *ngIf="task.metrics['read-bytes-complete'];else loadingTemplate">
@@ -76,8 +70,21 @@
       <td>{{ task["start_time"] | humanizeDate: 'yyyy-MM-dd HH:mm:ss' }}</td>
       <td>{{ task.duration | humanizeDuration }}</td>
       <td>{{ task["end-time"] | humanizeDate: 'yyyy-MM-dd HH:mm:ss' }}</td>
-      <td nzRight="0px">
+      <td nzRight="50px">
         <flink-job-badge [state]="task.status"></flink-job-badge>
+      </td>
+      <td nzRight="0px">
+        <nz-dropdown [nzPlacement]="'bottomRight'">
+          <span nz-dropdown><i nz-icon nzType="ellipsis" nzTheme="outline"></i></span>
+              <ul nz-menu nzSelectable>
+                <li nz-menu-item>
+              <span *ngIf="!task['taskmanager-id'] || task['taskmanager-id'] === '(unassigned)'; else hrefTpl">-</span>
+              <ng-template #hrefTpl>
+                <a [routerLink]="['/task-manager',task['taskmanager-id'],'logs']" target="_blank">Taskamanger Log</a>
+              </ng-template>
+            </li>
+          </ul>
+        </nz-dropdown>
       </td>
     </tr>
   </tbody>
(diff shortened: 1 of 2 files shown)


zentol:
As usual, please include a screenshot of the changes.

jinglining:
image


> As usual, please include a screenshot of the changes.

image
This is the screenshot of the changes.

zentol:
This is rather unintuitive. The column key is "ID" but the values links to some log file.
We need a better column key, and imo shouldn't directly link to the log file but instead the page of the host TE.

jinglining:
> This is rather unintuitive. The column key is "ID" but the values links to some log file.
> We need a better column key, and imo shouldn't directly link to the log file but instead the page of the host TE.
This is a part for log change, then user can easily see log from job. So add this log link. And I have discussed with till

zentol:
I know what the goal is; but this doesn't change the fact that `ID` is not an appropriate column key for a link to the TaskExecutors log page. For that "Log file" would make more sense.

flink-runtime-web/web-dashboard/src/app/pages/job/overview/subtasks/job-overview-drawer-subtasks.component.html
```diff
@@ -42,7 +42,10 @@
   <tbody>
     <tr *ngFor="let task of listOfTask; trackBy:trackTaskBy;">
       <td nzLeft="0">
-        {{ task.subtask }}
+        <span *ngIf="!task['resource-id'] || task['resource-id'] === '(unassigned)'; else hrefTpl">{{ task.subtask }}</span>
+        <ng-template #hrefTpl>
+          <a [routerLink]="['/task-manager',task['resource-id'],'logs']" target="_blank">LOG {{ task.subtask }}</a>
```
zentol: why are we including the subtask ID in the link label? IMO this will just confuse people as they will assume that there are different logs.

flink-runtime-web/web-dashboard/src/app/pages/job/overview/subtasks/job-overview-drawer-subtasks.component.html
```diff
@@ -42,7 +43,10 @@
   <tbody>
     <tr *ngFor="let task of listOfTask; trackBy:trackTaskBy;">
       <td nzLeft="0">
```
vthinkxie: should add nzWidth="40px"

jinglining:
image
screenshot

jinglining:
#9696

rmetzger:
Thanks a lot for this pull request.
I agree with Chesnay, that the initial approach of adding a "Logs" link next to the ID does not make sense conceptually.
The current approach...

image

... is also not good in my opinion, for the following reasons:
- This view shows metrics per subtask ID, the "log" link is a little bit lost there
- I don't understand the decision to use a lowercase "l". I would write "Log".
- I'm not sure if it is intuitive for users to understand that this "Log" is the TaskManager log. As a user, I would expect a page that shows the log messages only from this specific SubTask, which is not true.

I don't have much experience with UX design, and I don't have much time to really think this through (I would recommend you to do that), but one idea I had was to solve this similar to how Google does their UI:
Screenshot 2019-10-17 at 14 08 32
With that approach, we could have the three "..." at the end of the line, and then a dropdown that says: "View TaskManager logs"

zentol:
Could we not update the `host` column to include the port (like on the taskmanagers page) and link to the TaskManager page? Wouldn't this effectively work like the current iteration, just without an additional column?

jinglining:
> Could we not update the `host` column to include the port (like on the taskmanagers page) and link to the TaskManager page? Wouldn't this effectively work like the current iteration, just without an additional column?
Like this:
image

jinglining:
> Could we not update the `host` column to include the port (like on the taskmanagers page) and link to the TaskManager page? Wouldn't this effectively work like the current iteration, just without an additional column?

Like this:
image

flink-runtime-web/web-dashboard/src/app/pages/job/overview/subtasks/job-overview-drawer-subtasks.component.html
```diff
@@ -69,9 +70,22 @@
...
         <flink-job-badge [state]="task.status"></flink-job-badge>
       </td>
+      <td nzRight="0px">
+        <nz-dropdown [nzPlacement]="'bottomRight'">
+          <span nz-dropdown><i nz-icon nzType="ellipsis" nzTheme="outline"></i></span>
+              <ul nz-menu nzSelectable>
+                <li nz-menu-item>
+              <span *ngIf="!task['taskmanager-id'] || task['taskmanager-id'] === '(unassigned)'; else hrefTpl">-</span>
+              <ng-template #hrefTpl>
+                <a [routerLink]="['/task-manager',task['taskmanager-id'],'logs']" target="_blank">Taskamanger Log</a>
```
vthinkxie:
```suggestion
                <a [routerLink]="['/task-manager',task['taskmanager-id'],'logs']" target="_blank">View taskmanger log</a>
```

flink-runtime-web/web-dashboard/src/app/pages/job/overview/subtasks/job-overview-drawer-subtasks.component.html
```diff
@@ -69,9 +70,22 @@
...
-      <td nzRight="0px">
+      <td nzRight="50px">
         <flink-job-badge [state]="task.status"></flink-job-badge>
       </td>
+      <td nzRight="0px">
+        <nz-dropdown [nzPlacement]="'bottomRight'">
+          <span nz-dropdown><i nz-icon nzType="ellipsis" nzTheme="outline"></i></span>
+              <ul nz-menu nzSelectable>
+                <li nz-menu-item>
+              <span *ngIf="!task['taskmanager-id'] || task['taskmanager-id'] === '(unassigned)'; else hrefTpl">-</span>
```
vthinkxie: should not inside the dropdown

flink-runtime-web/web-dashboard/src/app/pages/job/overview/subtasks/job-overview-drawer-subtasks.component.html
```diff
@@ -69,9 +70,22 @@
...
-      <td nzRight="0px">
+      <td nzRight="50px">
         <flink-job-badge [state]="task.status"></flink-job-badge>
       </td>
+      <td nzRight="0px">
+        <span *ngIf="!task['taskmanager-id'] || task['taskmanager-id'] === '(unassigned)'; else hrefTpl">-</span>
+        <ng-template #hrefTpl>
+          <nz-dropdown [nzPlacement]="'bottomRight'">
+            <span nz-dropdown><i nz-icon nzType="ellipsis" nzTheme="outline"></i></span>
+            <ul nz-menu nzSelectable>
```
vthinkxie:
```suggestion
            <ul nz-menu>
```

jinglining:
screenshot

running

image

created

image

jinglining:
The previous design in ths PR is following the blink version UX which may not get enough feedbacks before. 

After trying both Robert and zentol's UX suggestion in my local development, I think "..." with dropdown is better since it would give a much more clear meanings for users and save the limited space at the same time.

I have updated the screenshot and code, thanks a lot for your advices.

cc @zentol @rmetzger

jinglining:
@flinkbot run travis

jinglining:
CI failed because FlinkKafkaProducerBase error, ignore.

jinglining:
> screenshot
> 
> running
> 
> image
> 
> created
> 
> image

@zentol I updated the screenshot, is there any problems? Could you give me some advice if possible?

zentol:
This looks fine to me; however I'd recommend to revise 60b5698aa2b9fe9ae233957688bd67a63325b7db to follow the same approach.

PR #9602, opened 2019-09-03
[FLINK-13856][checkpoint] Reduce the delete file api when the checkpo…

chendonglin521:

*Reduce the delete file api when the checkpoint is completed*

## Brief change log

  - *In full checkpoint, to delete the snapshot(chk-) folder directly*
  - *In incremental checkpoint, deleting the shared state first, and then dropping the exclusive location as a whole.*

## Verifying this change

  - *refator CompletedCheckpoint.doDiscard, add private method  getIncrementKeyedStates to get increment key*
  - *add method void disposeStorageLocation(boolean recursive) throws IOException  in  CompletedCheckpointStorageLocation*
 

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/checkpoint/CompletedCheckpoint.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/checkpoint/CompletedCheckpoint.java
@@ -26,6 +26,8 @@ import org.apache.flink.runtime.state.CompletedCheckpointStorageLocation;
 import org.apache.flink.runtime.state.SharedStateRegistry;
 import org.apache.flink.runtime.state.StateUtil;
 import org.apache.flink.runtime.state.StreamStateHandle;
+import org.apache.flink.runtime.state.KeyedStateHandle;
+import org.apache.flink.runtime.state.IncrementalRemoteKeyedStateHandle;
 import org.apache.flink.util.ExceptionUtils;
 
 import org.slf4j.Logger;
@@ -40,6 +42,8 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.HashSet;
+import java.util.Set;
 
 import static org.apache.flink.util.Preconditions.checkArgument;
 import static org.apache.flink.util.Preconditions.checkNotNull;
@@ -254,14 +258,14 @@ public class CompletedCheckpoint implements Serializable {
 
 			// discard private state objects
 			try {
-				StateUtil.bestEffortDiscardAllStateObjects(operatorStates.values());
+				StateUtil.bestEffortDiscardAllStateObjects(getSharedStateHandle());
 			} catch (Exception e) {
 				exception = ExceptionUtils.firstOrSuppressed(e, exception);
 			}
 
 			// discard location as a whole
 			try {
-				storageLocation.disposeStorageLocation();
+				storageLocation.disposeStorageLocation(true);
 			}
 			catch (Exception e) {
 				exception = ExceptionUtils.firstOrSuppressed(e, exception);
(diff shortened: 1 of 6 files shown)


StephanEwen:
I think this needs a bit more work. Will follow up with a more detailed review in a few days, but the basic points are:

  - no tests
  - not all FS implementations actually support recursive deletes well (object stores are slower in recursive delete than file-by-file delete)
  - the extracting of shared keyed state handles could use a different approach

chendonglin521:
> I think this needs a bit more work. Will follow up with a more detailed review in a few days, but the basic points are:
> 
> * no tests
> * not all FS implementations actually support recursive deletes well (object stores are slower in recursive delete than file-by-file delete)
> * the extracting of shared keyed state handles could use a different approach
Sorry Stephan！ I missed your comments！ I will investigate FS in detail again.

klion26:
@chendonglin521 I think reduce the RPC call is really import for HDFS, we encounter the RPC pressure also(keyed states and operator states). And we solved the problem by using shared the same file for different state handle, currently we have filed a issue[1] for this, and @StephanEwen Step is helping review the filed PR, the PR  currently is just for shared state. We have also implemented this feature for exclusive state currently in our production, from the observation from our production, this can reduce much RPC count(including create RPC, delete RPC, create block RPC and so on…), the create file RPC can reduce 1/3 for all the cluster, and request queue length reduced from thousands to less than hundred.
And I think the problem you encountered here can benefit from our proposal also.

[1] https://issues.apache.org/jira/browse/FLINK-11937

flink-runtime/src/main/java/org/apache/flink/runtime/checkpoint/CompletedCheckpoint.java
```diff
@@ -281,6 +285,34 @@ private void doDiscard() throws Exception {
 		}
 	}
 
+	private Set<KeyedStateHandle> getIncrementKeyedStates() {
```
Myasuka: The real target you want is actually `getSharedStates`

flink-runtime/src/main/java/org/apache/flink/runtime/checkpoint/CompletedCheckpoint.java
```diff
@@ -281,6 +285,34 @@ private void doDiscard() throws Exception {
...
+					for (KeyedStateHandle managedKeyedState : operatorSubtaskState.getManagedKeyedState()) {
+						if (managedKeyedState instanceof IncrementalRemoteKeyedStateHandle) {
+							incrementKeyedState.add(managedKeyedState);
+						}
+					}
+				}
+
+				if (operatorSubtaskState != null && operatorSubtaskState.getRawKeyedState() != null) {
+					for (KeyedStateHandle rawKeyedState : operatorSubtaskState.getRawKeyedState()) {
+						if (rawKeyedState instanceof IncrementalRemoteKeyedStateHandle) {
```
Myasuka: raw keyed state handle cannot be a `IncrementalRemoteKeyedStateHandle`

flink-runtime/src/main/java/org/apache/flink/runtime/checkpoint/CompletedCheckpoint.java
chendonglin521: yes，do you a better suggestion？

Myasuka: I think I have offered the method name as `getSharedStates`

flink-runtime/src/main/java/org/apache/flink/runtime/checkpoint/CompletedCheckpoint.java
```diff
@@ -281,6 +285,30 @@ private void doDiscard() throws Exception {
...
+	private Set<StreamStateHandle> getSharedStateHandle() {
+		Set<StreamStateHandle> sharedStateHandles = new HashSet<>(operatorStates.size());
+
+		//reference SavepointV2Serializer.serializeKeyedStateHandle
+		operatorStates.values().forEach(operatorState -> {
+			operatorState.getStates().forEach(operatorSubtaskState -> {
+				if (operatorSubtaskState != null) {
+					for (KeyedStateHandle managedKeyedState : operatorSubtaskState.getManagedKeyedState()) {
+						if (managedKeyedState instanceof IncrementalRemoteKeyedStateHandle) {
+							sharedStateHandles.addAll(((IncrementalRemoteKeyedStateHandle) managedKeyedState).getSharedState().values());
```
1996fanrui, line 297: Cleaning up all shared state(FileStateHandle) will cause SharedStateRegistry to become invalid, and all shared state will be cleaned up?

Please confirm, has the sst file used by subsequent Checkpoint been deleted?

PR #9607, opened 2019-09-04
[FLINK-13946] Remove job session related code from ExecutionEnvironment

kl0u:

This PR removes code related to JobSessions from the `ExecutionEnvironment` and the `PlanExecutor`s. This code was added in the context of [FLINK-2097](https://issues.apache.org/jira/browse/FLINK-2097) but it was never activated, as illustrated by the comment at [ExecutionEnvironment.java#L285 ](https://github.com/apache/flink/blob/master/flink-java/src/main/java/org/apache/flink/api/java/ExecutionEnvironment.java#L285). The work in this PR is part of the preparation for the upcoming re-design of the whole Client/Executor API.

## Brief change log

The changes in the subclasses of the `ExecutionEnvironment` remove methods that were setting session-related parameters and reflect the simplification of the `PlanExecutor` _lifecycle_ explained below (for `Local` and `RemoteEnvironment`).

The changes to the `PlanExecutors` have to do with the executor's lifecycle. Now the executor itself controls its lifecycle (`start()` and `stop()` are `private`) and we instantiate an executor for each call to `executePlan()`. This allows to get rid of the reapers from the `Local` and `RemoteEnvironments` and the `lock` that protected concurrent access to the executor's state.

The lifecycle is more explicit now and aligned with the current use of the `ExecutionEnvironment`. If in the future we choose to change this and decide to re-use execution environments, then we can add this functionality back, potentially under a different design/architecture.

## Verifying this change

This change is a code cleanup so it is covered by existing tests.

## Documentation

Please have a look at this one @aljoscha and @tillrohrmann.

### DIFF
--- a/flink-clients/src/main/java/org/apache/flink/client/LocalExecutor.java
+++ b/flink-clients/src/main/java/org/apache/flink/client/LocalExecutor.java
@@ -182,6 +182,17 @@ public class LocalExecutor extends PlanExecutor {
 			throw new IllegalArgumentException("The plan may not be null.");
 		}
 
+		stop();
+
+		// configure the number of local slots equal to the parallelism of the local plan
+		if (this.taskManagerNumSlots == DEFAULT_TASK_MANAGER_NUM_SLOTS) {
+			int maxParallelism = plan.getMaximumParallelism();
+			if (maxParallelism > 0) {
+				this.taskManagerNumSlots = maxParallelism;
+			}
+		}
+		start();
+
 		synchronized (this.lock) {
 
 			// check if we start a session dedicated for this execution
@@ -189,14 +200,6 @@ public class LocalExecutor extends PlanExecutor {
 
 			if (jobExecutorService == null) {
 				shutDownAtEnd = true;
-
-				// configure the number of local slots equal to the parallelism of the local plan
-				if (this.taskManagerNumSlots == DEFAULT_TASK_MANAGER_NUM_SLOTS) {
-					int maxParallelism = plan.getMaximumParallelism();
-					if (maxParallelism > 0) {
-						this.taskManagerNumSlots = maxParallelism;
-					}
-				}
 
 				// start the cluster for us
 				start();
--- a/flink-clients/src/main/java/org/apache/flink/client/LocalExecutor.java
+++ b/flink-clients/src/main/java/org/apache/flink/client/LocalExecutor.java
@@ -41,6 +41,9 @@ import org.apache.flink.runtime.minicluster.RpcServiceSharing;
 
 import java.util.List;
 
+import static org.apache.flink.util.Preconditions.checkNotNull;
+import static org.apache.flink.util.Preconditions.checkState;
+
 /**
  * A PlanExecutor that runs Flink programs on a local embedded Flink runtime instance.
  *
@@ -77,11 +80,11 @@ public class LocalExecutor extends PlanExecutor {
 	// ------------------------------------------------------------------------
 
 	public LocalExecutor() {
-		this(null);
+		this(new Configuration());
 	}
 
 	public LocalExecutor(Configuration conf) {
-		this.baseConfiguration = conf != null ? conf : new Configuration();
+		this.baseConfiguration = checkNotNull(conf);
 	}
 
 	// ------------------------------------------------------------------------
@@ -106,19 +109,13 @@ public class LocalExecutor extends PlanExecutor {
 
 	// --------------------------------------------------------------------------------------------
 
-	@Override
-	public void start() throws Exception {
-		synchronized (lock) {
-			if (jobExecutorService == null) {
-				// create the embedded runtime
-				jobExecutorServiceConfiguration = createConfiguration();
+	private void start() throws Exception {
+		Thread.holdsLock(lock);
+		checkState(jobExecutorService == null);
 
-				// start it up
-				jobExecutorService = createJobExecutorService(jobExecutorServiceConfiguration);
-			} else {
-				throw new IllegalStateException("The local executor was already started.");
-			}
-		}
+		// start the embedded runtime
+		jobExecutorServiceConfiguration = createConfiguration();
+		jobExecutorService = createJobExecutorService(jobExecutorServiceConfiguration);
 	}
 
 	private JobExecutorService createJobExecutorService(Configuration configuration) throws Exception {
@@ -146,20 +143,11 @@ public class LocalExecutor extends PlanExecutor {
 		return miniCluster;
 	}
 
-	@Override
-	public void stop() throws Exception {
-		synchronized (lock) {
-			if (jobExecutorService != null) {
-				jobExecutorService.close();
-				jobExecutorService = null;
-			}
-		}
-	}
-
-	@Override
-	public boolean isRunning() {
-		synchronized (lock) {
-			return jobExecutorService != null;
+	private void stop() throws Exception {
+		Thread.holdsLock(lock);
+		if (jobExecutorService != null) {
+			jobExecutorService.close();
+			jobExecutorService = null;
 		}
 	}
 
@@ -178,38 +166,13 @@ public class LocalExecutor extends PlanExecutor {
 	 */
 	@Override
 	public JobExecutionResult executePlan(Plan plan) throws Exception {
-		if (plan == null) {
-			throw new IllegalArgumentException("The plan may not be null.");
-		}
-
-		stop();
-
-		// configure the number of local slots equal to the parallelism of the local plan
-		if (this.taskManagerNumSlots == DEFAULT_TASK_MANAGER_NUM_SLOTS) {
-			int maxParallelism = plan.getMaximumParallelism();
-			if (maxParallelism > 0) {
-				this.taskManagerNumSlots = maxParallelism;
-			}
-		}
-		start();
+		checkNotNull(plan);
 
 		synchronized (this.lock) {
-
-			// check if we start a session dedicated for this execution
-			final boolean shutDownAtEnd;
-
-			if (jobExecutorService == null) {
-				shutDownAtEnd = true;
-
-				// start the cluster for us
-				start();
-			}
-			else {
-				// we use the existing session
-				shutDownAtEnd = false;
-			}
-
 			try {
+				configureNoOfLocalSlots(plan);
+				start();
+
 				// TODO: Set job's default parallelism to max number of slots
 				final int slotsPerTaskManager = jobExecutorServiceConfiguration.getInteger(TaskManagerOptions.NUM_TASK_SLOTS, taskManagerNumSlots);
 				final int numTaskManagers = jobExecutorServiceConfiguration.getInteger(ConfigConstants.LOCAL_NUMBER_TASK_MANAGER, 1);
@@ -222,11 +185,17 @@ public class LocalExecutor extends PlanExecutor {
 				JobGraph jobGraph = jgg.compileJobGraph(op, plan.getJobId());
 
 				return jobExecutorService.executeJobBlocking(jobGraph);
+			} finally {
+				stop();
 			}
-			finally {
-				if (shutDownAtEnd) {
-					stop();
-				}
+		}
+	}
+
+	private void configureNoOfLocalSlots(final Plan plan) {
+		if (this.taskManagerNumSlots == DEFAULT_TASK_MANAGER_NUM_SLOTS) {
+			int maxParallelism = plan.getMaximumParallelism();
+			if (maxParallelism > 0) {
+				this.taskManagerNumSlots = maxParallelism;
 			}
 		}
 	}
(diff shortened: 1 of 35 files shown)


flink-clients/src/main/java/org/apache/flink/client/LocalExecutor.java
```diff
@@ -146,21 +129,9 @@ private JobExecutorService createJobExecutorService(Configuration configuration)
...
-			}
-		}
-	}
-
-	@Override
-	public boolean isRunning() {
-		synchronized (lock) {
-			return jobExecutorService != null;
-		}
+	private void stopExecutorService(final JobExecutorService executorService) throws Exception {
```
TisonKun: I'm highly suspect the value brought by `JobExecutorService`. Its only implementation is `MiniCluster`. Also, it might have name conflict with the ongoing "Executor".

flink-clients/src/main/java/org/apache/flink/client/LocalExecutor.java
```diff
@@ -178,65 +149,71 @@ public boolean isRunning() {
...
+			if (executorService != null) {
+				stopExecutorService(executorService);
 			}
-			else {
-				// we use the existing session
-				shutDownAtEnd = false;
+		}
+	}
+
+	private Configuration configureExecution(final Plan executable) {
```
TisonKun: Why the parameter named `executable`  that different from `plan` anywhere others in this file?

flink-clients/src/main/java/org/apache/flink/client/RemoteExecutor.java
```diff
@@ -178,49 +127,30 @@ public boolean isRunning() {
...
-			try {
-				return client.run(program, defaultParallelism).getJobExecutionResult();
-			}
-			finally {
-				if (shutDownAtEnd) {
-					stop();
-				}
+		ClusterClient<?>  client = null;
+		try {
+			client = startClusterClient();
```
TisonKun: inline `startClusterClient` and `stopClusterClient` make more sense to me.

flink-clients/src/main/java/org/apache/flink/client/RemoteExecutor.java
```diff
@@ -178,49 +127,30 @@ public boolean isRunning() {
...
+			return client.run(program, defaultParallelism).getJobExecutionResult();
+		} finally {
+			if (client != null) {
+				stopClusterClient(client);
```
TisonKun: Maybe we can rename the method `shutdown` to `close` so that we can make use of all AutoClosable provides.

flink-java/src/main/java/org/apache/flink/api/java/ExecutionEnvironment.java
```diff
@@ -245,68 +237,6 @@ public JobExecutionResult getLastJobExecutionResult(){
...
-	// --------------------------------------------------------------------------------------------
-
-	/**
-	 * Gets the JobID by which this environment is identified. The JobID sets the execution context
-	 * in the cluster or local environment.
-	 *
-	 * @return The JobID of this environment.
-	 * @see #getIdString()
-	 */
-	@PublicEvolving
```
TisonKun: Notice that your pull request description mentioned

```
The public API, i.e., is any changed class annotated with @Public(Evolving): (yes / **no**)
```

However, this pull request actually remove some of `@PublicEvolving` methods. I think they are not in used so that we can safely removed them but it worth to follow the required process.

flink-java/src/main/java/org/apache/flink/api/java/LocalEnvironment.java
```diff
@@ -71,150 +63,34 @@ public LocalEnvironment(Configuration config) {
...
 
-		// Session management is disabled, revert this commit to enable
-		//p.setJobId(jobID);
-		//p.setSessionTimeout(sessionTimeout);
-
-		JobExecutionResult result = executor.executePlan(p);
-
-		this.lastJobExecutionResult = result;
-		return result;
+		// TODO: 31.08.19 make the executor autocloseable
```
TisonKun: Another followup could be whether we unify stream & batch PlanExecutor or both inline it, given that in `PlanExecutor` can be only used in batch case and `RemoteStreamEnvironment` just inline the execution.

In the context of client APIs refactoring it seems to finally have one env thus I'm afraid more works to `PlanExecutor` will be in vain.

kl0u: I agree on both points:
1) We should unify the executors for both batch and streaming but this is a bigger issue and will be addressed later.
2) It is not worth spending a lot of time in the `PlanExecutor`s, but at least let's make the code clearer.

flink-java/src/main/java/org/apache/flink/api/java/ExecutionEnvironment.java
kl0u: I agree. I updated the PR description.

flink-clients/src/main/java/org/apache/flink/client/LocalExecutor.java
kl0u: I am not so familiar with this code and maybe @tillrohrmann has more to say on the value of the `JobExecutorService`. In any case, I believe this is not related to this PR and can be added in the future. What do you think?

flink-clients/src/main/java/org/apache/flink/client/LocalExecutor.java
```diff
@@ -77,11 +70,11 @@
 	// ------------------------------------------------------------------------
 
 	public LocalExecutor() {
-		this(null);
+		this(new Configuration());
 	}
 
 	public LocalExecutor(Configuration conf) {
-		this.baseConfiguration = conf != null ? conf : new Configuration();
+		this.baseConfiguration = checkNotNull(conf);
```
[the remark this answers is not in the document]
kl0u: I am used to using the `checkNotNull` but it is true that the `@Nonnull` is part of the whole java ecosystem so it may be worth using it instead. I would like to see other arguments as well and I am always up for uniformizing coding styles :)

kl0u:
Hi @TisonKun ! Thanks for the review. I integrated your comments. Please have a look and let me know what you think.

flink-clients/src/main/java/org/apache/flink/client/LocalExecutor.java
tillrohrmann: The `JobExecutor` and the `JobExecutorService` have been introduced to bridge between the Flip-6 `MiniCluster` and the legacy `FlinkMiniCluster`. They should be obsolete now and could be removed if needed.

flink-clients/src/main/java/org/apache/flink/client/LocalExecutor.java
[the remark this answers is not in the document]
aljoscha: I think you can only omit the `checkNotNull()` calls when your build system verifies that the `@Nonnull` contract is actually followed. Outside code, that is calling your function might still pass in a `null`.

flink-clients/src/main/java/org/apache/flink/client/program/ContextEnvironment.java
```diff
@@ -72,15 +71,9 @@ public String getExecutionPlan() throws Exception {
...
-	@Override
-	public void startNewSession() throws Exception {
-		jobID = JobID.generate();
-	}
-
 	@Override
 	public String toString() {
-		return "Context Environment (parallelism = " + (getParallelism() == ExecutionConfig.PARALLELISM_DEFAULT ? "default" : getParallelism())
-				+ ") : " + getIdString();
+		return "Context Environment (parallelism = " + (getParallelism() == ExecutionConfig.PARALLELISM_DEFAULT ? "default" : getParallelism()) + ").";
```
aljoscha: I wouldn't put a `.` here, I don't think it's a sentence

flink-clients/src/main/java/org/apache/flink/client/LocalExecutor.java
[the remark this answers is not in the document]
TisonKun: @aljoscha thanks for your input. I agree that we could both make use of compile time constraint for internal development and place runtime checker when input may be from outside.

flink-clients/src/main/java/org/apache/flink/client/LocalExecutor.java
TisonKun: Thanks for you information @tillrohrmann. @kl0u I'd like to make this removal a separated issue, fyi FLINK-13961.

flink-clients/src/main/java/org/apache/flink/client/RemoteExecutor.java
TisonKun: I think the latest CI fails on uncompleted refactor

```
12:40:29.667 [ERROR] /home/travis/build/flink-ci/flink/flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/gateway/local/ProgramDeployer.java:[134,54] cannot find symbol
  symbol:   method shutdown()
  location: variable clusterClient of type org.apache.flink.client.program.ClusterClient<T>
12:40:29.668 [ERROR] /home/travis/build/flink-ci/flink/flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/gateway/local/ProgramDeployer.java:[175,54] cannot find symbol
  symbol:   method shutdown()
  location: variable clusterClient of type org.apache.flink.client.program.ClusterClient<T>
12:40:29.669 [ERROR] /home/travis/build/flink-ci/flink/flink-table/flink-sql-client/src/main/java/org/apache/flink/table/client/gateway/local/LocalExecutor.java:[434,62] cannot find symbol
  symbol:   method shutdown()
  location: variable clusterClient of type org.apache.flink.client.program.ClusterClient<T>
```

flink-java/src/main/java/org/apache/flink/api/java/LocalEnvironment.java
TisonKun: Thanks for the explanation. Sounds reasonable.

flink-clients/src/main/java/org/apache/flink/client/LocalExecutor.java
kl0u: Sounds good @TisonKun .

flink-clients/src/main/java/org/apache/flink/client/RemoteExecutor.java
kl0u: Yes, I fixed it. For some reason `flink-table` was excluded from my IDE and I did not see the failure locally.

flink-clients/src/main/java/org/apache/flink/client/LocalExecutor.java
```diff
@@ -178,65 +139,65 @@ public boolean isRunning() {
...
 
-			try {
-				// TODO: Set job's default parallelism to max number of slots
-				final int slotsPerTaskManager = jobExecutorServiceConfiguration.getInteger(TaskManagerOptions.NUM_TASK_SLOTS, taskManagerNumSlots);
-				final int numTaskManagers = jobExecutorServiceConfiguration.getInteger(ConfigConstants.LOCAL_NUMBER_TASK_MANAGER, 1);
-				plan.setDefaultParallelism(slotsPerTaskManager * numTaskManagers);
+	private Configuration createExecutorServiceConfig() {
+		final Configuration newConfiguration = new Configuration();
+		newConfiguration.setInteger(TaskManagerOptions.NUM_TASK_SLOTS, taskManagerNumSlots);
+		newConfiguration.setBoolean(CoreOptions.FILESYTEM_DEFAULT_OVERRIDE, defaultOverwriteFiles);
```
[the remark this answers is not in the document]
kl0u: You mean the `setTaskManagerNumSlots()` setter?

kl0u:
I integrated your comments @aljoscha . Please have a look and let me know what you think.

flink-yarn-tests/src/test/java/org/apache/flink/yarn/util/FakeClusterClient.java
```diff
@@ -93,4 +93,9 @@ public String stopWithSavepoint(JobID jobId, boolean advanceToEndOfEventTime, @N
 	public Map<String, OptionalFailure<Object>> getAccumulators(JobID jobID, ClassLoader loader) {
 		return Collections.emptyMap();
 	}
+
+	@Override
+	public void close() throws Exception {
```
TisonKun: Wried. The latest CI complaint with

```
/home/travis/build/flink-ci/flink/flink-yarn-tests/src/test/java/org/apache/flink/yarn/util/FakeClusterClient.java:[43,8] org.apache.flink.yarn.util.FakeClusterClient is not abstract and does not override abstract method close() in java.lang.AutoCloseable
```

but actually we implement it.

flink-clients/src/main/java/org/apache/flink/client/program/ClusterClient.java
```diff
@@ -101,17 +101,6 @@ public ClusterClient(Configuration flinkConfig) {
...
 	}
 
-	// ------------------------------------------------------------------------
-	//  Startup & Shutdown
-	// ------------------------------------------------------------------------
-
-	/**
-	 * User overridable hook to close the client, possibly closes internal services.
-	 */
-	public void shutdown() throws Exception {
```
TisonKun: As I have discussed with @tillrohrmann in FLINK-13750, maybe it is also reasonable we have a empty impl for `close` in `ClusterClient` and let only those who actually requires cleanups override it.

TisonKun:
Hi @kl0u, thanks for your update. I notice the strange state in travis that reports a compile error which should not be there. Please take a look.

Also you can mark conversations that addressed or reached a consensus above as "resolved" to fold them, which make ongoing reviews more clear.

flink-yarn-tests/src/test/java/org/apache/flink/yarn/util/FakeClusterClient.java
kl0u: Let's see what travis says now.

kl0u:
Thanks for the reviews @TisonKun and @aljoscha . Merged.

PR #9618, opened 2019-09-04
[FLINK-13927] Add note about hadoop dependencies for local debug

azagrebin:

Currently if user tries to run the job locally (e.g. from IDE) and uses Hadoop fs, it will not work if hadoop dependencies are not on the class path which is the case for the example from the quick start.

## Brief change log

We can add a hint about adding provided hadoop dependencies in this case (+ Idea hint about provided included checkbox in run/debug app setup) to `docs/ops/deployment/hadoop.md`.

## Verifying this change

Run `./docs/build_docs.sh` -i and [open](http://localhost:4000/ops/deployment/hadoop.html) in browser to check the doc changes.

  - The serializers: (no)

## Documentation


azagrebin:
Thanks for the review @aljoscha 
merged into master by 3c991dc507b1d9ab07a24cda00c91881907d8e25

PR #9620, opened 2019-09-04
[FLINK-13959] Consolidate DetachedEnvironment and ContextEnvironment

kl0u:

The purpose of this PR is to remove the `DetachedEnvironment` and integrate its functionality in its superclass, the `ContextEnvironment`.

## Verifying this change

This change is a rework / code cleanup covered by existing tests.

## Documentation

R @aljoscha


flink-clients/src/main/java/org/apache/flink/client/program/MiniClusterClient.java
```diff
@@ -61,7 +61,10 @@ public JobSubmissionResult submitJob(JobGraph jobGraph, ClassLoader classLoader)
 
 		if (isDetached()) {
 			try {
-				return jobSubmissionResultFuture.get();
+				final JobSubmissionResult jobSubmissionResult = jobSubmissionResultFuture.get();
+
+				this.lastJobExecutionResult = new DetachedJobExecutionResult(jobSubmissionResult.getJobID());
```
TisonKun: Nice catch! I just notice that we not set `lastJobExecutionResult` in MiniCluster case. I think it might be hopefully we eventually get rid of it(we will unify all submission code path by `submitJob` instead of `run`s which the only method require `lastJobExecutionResult`), but let's keep consistency now which also helps convinced refactors.

flink-core/src/main/java/org/apache/flink/api/common/JobSubmissionResult.java
```diff
@@ -56,10 +56,6 @@ public boolean isJobExecutionResult() {
 	 * @throws ClassCastException if this is not a JobExecutionResult
 	 */
 	public JobExecutionResult getJobExecutionResult() {
-		if (isJobExecutionResult()) {
-			return (JobExecutionResult) this;
-		} else {
-			throw new ClassCastException("This JobSubmissionResult is not a JobExecutionResult.");
-		}
+		throw new ClassCastException("This JobSubmissionResult is not a JobExecutionResult.");
```
TisonKun: In fact this infer a bad inheritance...

kl0u: I agree, but this is `public` unfortunately.

flink-clients/src/main/java/org/apache/flink/client/program/MiniClusterClient.java
kl0u: Yes, I fixed that in this PR. Before it was not failing because in the `ClusterClient.run(...)` we had the second check in the `if (lastJobExecutionResult == null) ...` which was returning `false`.

kl0u:
Thanks for the review @aljoscha and @TisonKun . I will merge as soon as Travis gives green.

PR #9630, opened 2019-09-05
[FLINK-13964] Let HighAvailabilityServices create cluster and client side HA services

tillrohrmann:

This PR is based on #9629.

Let HighAvailabilityServices extend ClientHighAvailabilityServices. This ensures that
the HighAvailabilityServices can create cluster and client side HA services. This again
is useful for components which need to create both types of services such as the MiniCluster.

Replace #getWebMonitorLeaderRetriever implementations with #getClusterRestEndpointLeaderRetriever.

cc @TisonKun 

## Verifying this change

  - The serializers: (no)

## Documentation


flink-runtime/src/main/java/org/apache/flink/runtime/highavailability/HighAvailabilityServicesFactory.java
```diff
@@ -47,19 +46,6 @@
...
-			public LeaderRetrievalService getClusterRestEndpointLeaderRetriever() {
-				return haServices.getWebMonitorLeaderRetriever();
-			}
-
-			@Override
-			public void close() throws Exception {
-				haServices.close();
-			}
-		};
+		return createHAServices(configuration, UnsupportedOperationExecutor.INSTANCE);
```
TisonKun: I think with current and semantically valid code path we should never use `executor` from client side. Even we do in the future, then we will pass a valid executor at that moment. Here my concerns is whether we need to place a `UnsupportedOperationExecutor.INSTANCE` here which will never be used to executed. Although the utility `UnsupportedOperationExecutor` seems reasonably but we can pass anything here because it is not to use. Either `UnsupportedOperationExecutor` or `Executor.directExecutor()` even `null` requires comments for those who doesn't familiar with this part, i.e., we don't get more readability use `UnsupportedOperationExecutor` here.

tillrohrmann: The idea was to explicitly state that this executor should not be used by any of the client side services. At the moment this holds true. If in the future this should change, it will be visible due to the `UnsupportedOperationException`. Then one should think about introducing a proper `Executor` whose lifecycle is properly managed. This will most likely require changes to the `HighAvailabilityServicesFactory` because we need to pass it to the `#createClientHAServices` method.

The reason why I would not like to pass `Executors.directExecutor()` or something else is that someone who changes that client side services don't use the `Executor` won't notice this because it won't fail. This might be problematic because he might execute blocking operations in the `Executor` for which the direct executor is not suitable.

tillrohrmann:
Thanks for the initial review @TisonKun. I hope that I could answer your question.

tillrohrmann:
Thanks for the review @TisonKun. Merging this PR.

PR #9632, opened 2019-09-05
[FLINK-13978] Add azure-pipelines.yml

rmetzger:

* Introduce an `azure-pipelines.yml` file, so that we can trigger [Azure Pipelines](https://dev.azure.com/rmetzger/Flink/_build?definitionId=4&_a=summary&view=branches) builds from `flink-ci/flink`.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,0 +1,35 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+trigger:
+  branches:
+    include:
+    - '*' 
+
+resources:
+  containers:
+  # Container with Maven 3.2.5 to have the same environment everywhere.
+  - container: flink-build-container
+    image: rmetzger/flink-ci:3
+  repositories:
+    - repository: templates
+      type: github
+      name: flink-ci/flink-azure-builds
+      endpoint: flink-ci
+
+jobs:
+- template: flink-build-jobs.yml@templates
+


azure-pipelines.yml
```diff
@@ -0,0 +1,79 @@
...
+  container: flink-build-container
+  timeoutInMinutes: 0
+  steps:
+  # Azure pipelines can only evaluate conditions with the build repo name in the steps.
+  # if the repo == "flink-ci/flink", we stop
+  - script: exit 1
+    condition: eq(variables['Build.Repository.Name'], 'flink-ci/flink')
+  - script: sudo chown -R user:user /home/user
+  - script: sudo chown -R user:user .
+  - task: Maven@3
```
zentol: can we just call into a bash script here that lives somewhere in flink-ci, so that we can change what is being run without having to touch the flink repo? (to a certain limit...)

azure-pipelines.yml
```diff
@@ -0,0 +1,79 @@
...
+    include:
+    - '*' 
+  paths:
+    exclude:
+    - docs/*
+    - '*.md'
+
+resources:
+  containers:
+  # Container with Maven 3.6.1 to have the same environment everywhere.
```
zentol: can we reduce this to 3.2.5 to have something comparable to the current setup?

azure-pipelines.yml
```diff
@@ -0,0 +1,79 @@
...
+  - container: flink-build-container
+    image: rmetzger/flink-ci:2
+
+jobs:
+  # use 'Default' pool
+- job: runOnDefaultAgentPool
+  pool:
+    name: Default
```
zentol: do I understand it correctly that this here runs on free machines, while the other job runs on sponsored ones?

azure-pipelines.yml
rmetzger: Yes, you can clone multiple repositories, and call bash scripts.
If we use the Maven task, we'll get test reports: https://dev.azure.com/rmetzger/Flink/_test/analytics?definitionId=1&contextType=build

I'm not sure if we will have this many changes to this particular task.
What we could consider doing is the following dirty hack: Azure pipelines has a "templates" feature, that allows to include other yml files into the main file.
We could put the main body of this file into a file in a separate repository, and then load it here.

I'll think about this.

azure-pipelines.yml
rmetzger: The "Default" pool are the sponsored machines. The name is a bit unfortunate, but we can change it

azure-pipelines.yml
zentol: My idea wasn't meant for the long-term, just short-term so we can play around a bit (For example, run e2e tests). Referring to an external script is the easiest way that I could think of.

azure-pipelines.yml
```diff
@@ -0,0 +1,39 @@
...
+# limitations under the License.
+
+
+trigger:
+  branches:
+    include:
+    - '*' 
+  paths:
+    exclude:
+    - docs/*
```
zentol: There are some tests that check the contents of this directory; modifications to _includes/generated should still be checked.

rmetzger:
Thanks

PR #9634, opened 2019-09-06
[hotfix][runtime] Streaming runtime 1.9 improvements for debuggability

1u0:

This are minor changes to improve streaming runtime debuggability for the changes introduced with mailbox in `release-1.9` branch.

## Brief change log

  - set a name (that includes task's name) for legacy source thread. This is a port of change that already present in the `master` branch;
  - rename `StreamTask`s `performDefaultAction()` method to `processInput()` to better reflect method's intent.

## Verifying this change

## Documentation

### DIFF
--- a/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SourceStreamTask.java
+++ b/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SourceStreamTask.java
@@ -114,6 +114,7 @@ public class SourceStreamTask<OUT, SRC extends SourceFunction<OUT>, OP extends S
 	protected void performDefaultAction(ActionContext context) throws Exception {
 		// Against the usual contract of this method, this implementation is not step-wise but blocking instead for
 		// compatibility reasons with the current source interface (source functions run as a loop, not in steps).
+		sourceThread.setTaskDescription(getName());
 		sourceThread.start();
 
 		// We run an alternative mailbox loop that does not involve default actions and synchronizes around actions.
@@ -205,6 +206,10 @@ public class SourceStreamTask<OUT, SRC extends SourceFunction<OUT>, OP extends S
 			} finally {
 				mailbox.clearAndPut(SOURCE_POISON_LETTER);
 			}
+		}
+
+		public void setTaskDescription(final String taskDescription) {
+			setName("Legacy Source Thread - " + taskDescription);
 		}
 
 		void checkThrowSourceExecutionException() throws Exception {
--- a/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SourceStreamTask.java
+++ b/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SourceStreamTask.java
@@ -111,7 +111,7 @@ public class SourceStreamTask<OUT, SRC extends SourceFunction<OUT>, OP extends S
 	}
 
 	@Override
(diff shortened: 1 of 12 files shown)


flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SourceStreamTask.java
```diff
@@ -114,6 +114,7 @@ protected void cleanup() {
 	protected void performDefaultAction(ActionContext context) throws Exception {
 		// Against the usual contract of this method, this implementation is not step-wise but blocking instead for
 		// compatibility reasons with the current source interface (source functions run as a loop, not in steps).
+		sourceThread.setTaskDescription(getName());
```
zhijiangW: I wonder why not do it via `this.sourceThread = new LegacySourceFunctionThread(getName())` directly in master branch. Then 
```
LegacySourceFunctionThread(String name) {
       super(name);
}
```

Although `performDefaultAction` would be executed only once for source task, it seems better to do it in constructor. WDYT?

1u0: Thanks for review @zhijiangW.

Afair, the story behind why it's done this way in the `master` branch, because the `LegacySourceFunctionThread` is instantiated in the constructor of `SourceStreamTask` and `StreamTask::getName()` method is not final.
To avoid a virtual method call in the constructor, @GJL has moved thread's name setup to the `performDefaultAction()` method just before starting the thread.

zhijiangW: Thanks for the explanation! 
Another option is via `env.getTaskInfo().getTaskNameWithSubtasks()` instead of `getName()` in constructor?
I could also accept the current way if you prefer since it was already accepted before in master. :)

1u0: Using `env.getTaskInfo().getTaskNameWithSubtasks()` directly was also rejected in that story because of code duplication. 😀

I'd prefer to keep this change as similar to `master` branch as possible, to avoid backport conflicts for other PRs (if happens).

zhijiangW:
@1u0 I guess only the hotfix for debugging issue would be merging into 1.9 and the other hotfix should be only merged into master? 
If so it is better to only cover one hotfix commit in this PR, and then we could merge the other hotfix into master in a separate PR.

1u0:
The change with the first commit is already present in the `master` branch.
If the second commit (with renaming of `performDefaultAction()`) is also ok to introduce in the `release-1.9` branch, then I recommend to merge this PR as is. But not backport it into `master` branch (!).

I have opened another PR #9635 to port the second commit into `master` branch. Otherwise, directly rebasing the second commit against the `master` branch should have merge conflicts (imo), as there was some refactoring that removed some `performDefaultAction()` overrides.

zhijiangW:
@1u0 After confirming with @pnowojski , I think we could only make the hotfix of thread name in release-1.9, so could you retain only this commit in PR then I could merge it?

pnowojski:
@zhijiangW sorry for misunderstanding, but I wrote that we can backport the `perfomDefaultAction` rename if it's not too much hassle (no conflicts etc.). As @1u0 has already done it, it's better to merge both commits, as it will make 1.9 and 1.10 branches a little bit more similar, so it will be easier to backport future changes.

PR #9636, opened 2019-09-06
[FLINK-13921] Simplify cluster level RestartStrategy configuration

tillrohrmann:

This PR simplifies the cluster `RestartStrategy` configuration. The configuration works now the following way:

* If the config option `restart-strategy` is configured, then Flink uses this `RestartStrategy`
* If the config option `restart-strategy` is not configured, then
** If checkpointing is disabled, then choose `NoRestartStrategy`
** If checkpointing is enabled, then choose `FixedDelayRestartStrategy(Integer.MAX_VALUE, "0 s")`

## Verifying this change

- Added `RestartStrategyFactoryTest`

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/restart/RestartStrategyFactoryTest.java
+++ b/flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/restart/RestartStrategyFactoryTest.java
@@ -1,0 +1,74 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.runtime.executiongraph.restart;
+
+import org.apache.flink.configuration.ConfigConstants;
+import org.apache.flink.configuration.Configuration;
+import org.apache.flink.util.TestLogger;
+
+import org.junit.Test;
+
+import java.time.Duration;
+
+import static org.hamcrest.CoreMatchers.instanceOf;
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+/**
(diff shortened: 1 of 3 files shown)


flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/restart/RestartStrategyFactoryTest.java
```diff
@@ -0,0 +1,74 @@
...
+	public void createRestartStrategyFactory_noRestartStrategyConfigured_returnsNoOrFixedIfCheckpointingEnabledRestartStrategyFactory() throws Exception {
+		final Configuration configuration = new Configuration();
+
+		final RestartStrategyFactory restartStrategyFactory = RestartStrategyFactory.createRestartStrategyFactory(configuration);
+
+		assertThat(restartStrategyFactory, instanceOf(NoOrFixedIfCheckpointingEnabledRestartStrategyFactory.class));
+	}
+
+	@Test
+	public void createRestartStrategyFactory_noRestartStrategyButAttemptsConfigured_returnsNoOrFixedIfCheckpointingEnabledRestartStrategyFactory() throws Exception {
```
zhuzhurk: I can understand why we have this case.
Just not sure whether it's necessary to have tests for removed logics, as we have to mock the inputs for a previous code path which does not exist any more.
This seems to introduce redundant tests in a long term sense. 
Or should we remove this kinds of tests a few versions later?

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/restart/RestartStrategyFactoryTest.java
```diff
@@ -0,0 +1,74 @@
...
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+/**
+ * Tests for the {@link RestartStrategyFactory}.
+ */
+public class RestartStrategyFactoryTest extends TestLogger {
+
+	@Test
+	public void createRestartStrategyFactory_noRestartStrategyConfigured_returnsNoOrFixedIfCheckpointingEnabledRestartStrategyFactory() throws Exception {
```
zhuzhurk: Maybe "createRestartStrategyFactory_noRestartStrategyConfigured" is OK and this case can have a shorter name.
I think it's not necessary to expose the expected result in the name.

tillrohrmann: I think it is actually beneficial to have an expressive name because it tells the reader what to expect.

flink-runtime/src/test/java/org/apache/flink/runtime/executiongraph/restart/RestartStrategyFactoryTest.java
tillrohrmann: I admit that this test is not perfect but I would like to keep it because it validates this PR's change.

zhuzhurk:
No more questions.
+1 for merging it.

tillrohrmann:
Thanks for the review @zhuzhurk and @GJL. Merging this PR now.

PR #9637, opened 2019-09-06
[FLINK-13884] Set default delay for restart strategies to 1s

tillrohrmann:
## What is the purpose of the change

This PR is based on #9562.

This commit sets the default delay for all restart strategies to 1s in order
to prevent restart storms from happening.

cc @zentol @zhuzhurk 

## Verifying this change

This change is a trivial rework / code cleanup without any test coverage.

## Does this pull request potentially affect one of the following parts:

  - Dependencies (does it add or upgrade a dependency): (no)
  - The public API, i.e., is any changed class annotated with `@Public(Evolving)`: (yes)
  - The serializers: (no)
  - The runtime per-record code paths (performance sensitive): (no)
  - Anything that affects deployment or recovery: JobManager (and its components), Checkpointing, Yarn/Mesos, ZooKeeper: (no)
  - The S3 file system connector: (no)

## Documentation

  - Does this pull request introduce a new feature? (no)
  - If yes, how is the feature documented? (docs)

PR #9638, opened 2019-09-06
[FLINK-13883] Deprecate unused AkkaOptions related to Akka's death watch

tillrohrmann:

This PR is based on #9637.

We no longer use Akka's death watch. Hence, this commit deprecates the death
watch related configuration options AkkaOptions#WATCH_HEARTBEAT_INTERVAL,
AkkaOptions#WATCH_THRESHOLD and AkkaOptions#WATCH_HEARTBEAT_PAUSE. Moreover,
it removes the death watch verification logic in AkkaUtils.

cc @zentol.

## Verifying this change

  - The serializers: (no)

## Documentation


flink-runtime/src/main/scala/org/apache/flink/runtime/akka/AkkaUtils.scala
```diff
@@ -505,12 +492,6 @@ object AkkaUtils {
          |      threshold = $transportThreshold
          |    }
          |
-         |    watch-failure-detector{
```
zentol: Just to understand things, if we use these options here, why do they have no effect on akka?

flink-connectors/flink-connector-kafka-base/src/test/java/org/apache/flink/streaming/connectors/kafka/KafkaTestBase.java
```diff
@@ -118,8 +117,6 @@ public static void shutDownServices() throws Exception {
 
 	protected static Configuration getFlinkConfiguration() {
 		Configuration flinkConfig = new Configuration();
-		flinkConfig.setString(AkkaOptions.WATCH_HEARTBEAT_PAUSE, "5 s");
```
zentol: shouldn't we update these instead to use the new heartbeat options?

flink-runtime/src/test/java/org/apache/flink/runtime/testutils/ZooKeeperTestUtils.java
```diff
@@ -81,10 +81,6 @@ public static Configuration configureZooKeeperHA(
 		config.setString(CheckpointingOptions.CHECKPOINTS_DIRECTORY, fsStateHandlePath + "/checkpoints");
 		config.setString(HighAvailabilityOptions.HA_STORAGE_PATH, fsStateHandlePath + "/recovery");
 
-		// Akka failure detection and execution retries
-		config.setString(AkkaOptions.WATCH_HEARTBEAT_INTERVAL, "1000 ms");
```
zentol: I suppose the idea behind removing them, instead of an update to the new options, is that since they didn't fail with ineffective options they obviously don't needed them in the first place? (same question for kafka)

flink-runtime/src/main/scala/org/apache/flink/runtime/akka/AkkaUtils.scala
tillrohrmann: The `watch-failure-detector` configuration is relevant for Akka's deathwatch mechanism. Since we don't use this feature anymore, we can remove the configuration.

flink-connectors/flink-connector-kafka-base/src/test/java/org/apache/flink/streaming/connectors/kafka/KafkaTestBase.java
tillrohrmann: The test is using an unused configuration option for several releases. Hence, it seems to work with the default of the heartbeat options for quite some time. Hence, I would not touch if there is not a very good reason for it.

flink-runtime/src/test/java/org/apache/flink/runtime/testutils/ZooKeeperTestUtils.java
tillrohrmann: Yes exactly. They weren't used for several releases now.

tillrohrmann:
Thanks for the review @zentol. Merging this PR now.

PR #9643, opened 2019-09-06
[FLINK-13961][client] Remove obsolete abstraction JobExecutor(Service)

TisonKun:

Remove obsolete abstraction JobExecutor(Service)

## Verifying this change

  - The serializers: (no)

## Documentation

cc @kl0u @aljoscha


TisonKun:
travis fails on NOTICE-binary checker. I don't think it is relevant to this pull request.

Instead, it looks like an instance of [FLINK-14009](https://issues.apache.org/jira/browse/FLINK-14009). @tillrohrmann is it relevant?

tillrohrmann:
@flinkbot run travis.

TisonKun:
Hi @tillrohrmann, thanks for your review. I don't think this is an issue with high priority, but let me share you some of my concerns.

- `JobExecutor` and `JobExecutorServices` provide poor abstraction based on our codebase.

`JobExecutor` has only one method `executeJobBlocking(JobGraph)`. In my opinion, in order to submit a job and request the result we are supposed to use the client instead of directly access the cluster. `JobExecutorServices` is introduced for the usage in `LocalExecutor#executePlan(Plan)`. 

- A real demand will require clear abstraction.

Back to the introduction of `JobExecutor` and `JobExecutorService`, we are required to provide an abstraction of MiniCluster and FlinkMiniCluster, which clearly points to the interface `executeJobBlocking(JobGraph)`. Now and in the near future, we don't have another `JobExecutor`. Anytime when we find another class takes similar responsibility like `MiniCluster`, we can always add an abstraction bases on that similarity.

TisonKun:
@flinkbot run travis

tillrohrmann:
Even though `JobExecutor` and `JobExecutorService` are not perfect it is clearly better than depending directly on the `MiniCluster` because it follows the [interface segregation principle](https://en.wikipedia.org/wiki/Interface_segregation_principle). 

As long as a user depends on `executeJobBlocking`, we should not remove it. Whether to use the client for communication with the `MiniCluster` or directly talking to it should not be topic of this PR. The same applies to the usage of the interface in the future. You might be right but at the moment we still have users of this interface and, hence, we should not do any second or third step before having done even the first.

I still think that we should close this PR.

TisonKun:
Now I agree with that we don't have to rush into the removal since there are still users of this interface. I will revisit this topic when the prerequisite in my mind ready.

Closed pull request and closed the corresponding JIRA as not a problem.

Thanks for your review @tillrohrmann !

PR #9646, opened 2019-09-08
 [FLINK-14004][runtime] Define SourceReaderOperator to verify the integration with StreamOneInputProcessor

zhijiangW:

We already refactored the task input and output in runtime stack for considering the requirements of FLIP-27. In order to further verify that the new source could work well with the unified `StreamOneInputProcessor` in mailbox model, we define the `SourceReaderOperator` as task input and implement a unit test for passing through the whole process.

## Brief change log

- Define `AbstractDataOutput` as the base class for both source output and network output.
- Refactor the previous processes based on `AbstractDataOutput`.
- Define `SourceReaderOperator` base class.
- Define `SourceReaderStreamTask` to generate the source input and output for `StreamOneInputProcessor`.

## Verifying this change

  - Added unit test in `SourceReaderStreamTaskTest` for verifying that the source reader operator could be integrated well with `StreamOneInputProcessor`.

## Documentation

### DIFF
--- a/flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/SourceReaderOperator.java
+++ b/flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/SourceReaderOperator.java
@@ -1,0 +1,34 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.streaming.api.operators;
+
+import org.apache.flink.annotation.Internal;
+import org.apache.flink.streaming.runtime.io.PushingAsyncDataInput;
+
+/**
+ * Base source operator only used for integrating the source reader which is proposed by FLIP-27. It implements
+ * the interface of {@link PushingAsyncDataInput} for naturally compatible with one input processing in runtime
+ * stack.
+ *
+ * <p>Note: We are expecting this to be changed to the concrete class once SourceReader interface is introduced.
+ *
+ * @param <OUT> The output type of the operator
+ */
+@Internal
+public abstract class SourceReaderOperator<OUT> extends AbstractStreamOperator<OUT> implements PushingAsyncDataInput<OUT> {
+}
--- a/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/AbstractDataOutput.java
+++ b/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/io/AbstractDataOutput.java
@@ -1,0 +1,51 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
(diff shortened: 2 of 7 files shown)


zhijiangW:
Thanks for this kind notification @StephanEwen !

I agree with your point. The final version of `SourceReader` should be given by FLIP-27, not by runtime side. The motivation of this PR is just an initial retry to verify the previous runtime refactoring tasks (#9478 #9483) which could integrate well with the future `SourceReader`. Then we could take it easy to merge previous PRs.

This PR would be rebased again to only focus on runtime modifications after FLIP-27 is voted and `SourceReader` is merged by source api stack.

zhijiangW:
Sorry for missing the following comments. I agree with the general above points.

On runtime side it does not care about the specific `SourceReader` interface. The runtime only interacts with source operator via `#emitNext` and `#available` methods atm. As for how the source operator further interacts with the internal source reader, the runtime does not need to understand it.

We can ignore the SourceReader related details in this PR. The possible meaningful point in this PR is that I try to pass through whether the runtime proposed `DataOutput` argument via `operator#emitNext` can satisfy well for the future `SourceReader`. So I construct a unit test for verifying it.

pnowojski:
I agree with @StephanEwen
> I think in that case there should be a placeholder interface in flink-runtime that is used, instead of a preliminary version of SourceReader. That placeholder gets refactored/replaced by the proper source reader then.
>
> That placeholder should also have only the necessary methods, nothing about splits, start, etc

With defining this place holder interface, I think we could work on this PR. We might need to reconsider regarding the splits if we want also to follow up with runtime support for stateful `SourceReader` in some next PR.

pnowojski:
I had a chat with @StephanEwen and a proposal to unblock this is to avoid defining `SourceReader` interface here. We can just implement the `SourceReaderOperator` class, as a for example a stateful/stateless infinite/finite source of integers. We could integrate it with the runtime and it would be generating the integers and emitting them directly to the `DataOutput`. Later, someone implementing a final `SourceReader` interface would just reimplement `SourceReaderOperator` to use `SourceReader` interface.

More specifically, we could just drop the `SourceReader` interface from this PR, embed the logic of `TestingFiniteSourceReader` inside the `SourceReaderOperator` and that should be it. 

The following would be to implement state snapshotting in this `SourceReaderOperator` (to test our integration with stateful readers).

zhijiangW:
Thanks for the further confirmation and pushing forward this work.

I agree with the conclusion that we could only touch `SourceReaderOperator` in this PR and not care about the interactive details between `SourceReaderOperator` and `SourceReader`.

In detail I think it is better to only give an abstract `SourceReaderOperator` to provide `emitNext` and `isAvailable` methods for runtime call, and implement this operator only in unit tests to verify the whole process ATM. Then after FLIP-27 ready, it could implement the former `SourceReaderOperator` with `SourceReader` together.

I already updated the codes based on the above confirmation.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/SourceReaderOperator.java
```diff
@@ -0,0 +1,32 @@
...
+import org.apache.flink.streaming.runtime.io.PushingAsyncDataInput;
+
+/**
+ * Base source operator only used for integrating the source reader which is proposed by FLIP-27. It implements
+ * the interface of {@link PushingAsyncDataInput} for naturally compatible with one input processing in runtime
+ * stack.
+ *
+ * @param <OUT> The output type of the operator
+ */
+@PublicEvolving
```
pnowojski: I think I would mark it as `Internal` and let the API folks later decide whether to expose this, or just the `SourceReader`. (Imo there is no need to expose the `SourceReaderOperator`).

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/SourceReaderOperator.java
```diff
@@ -0,0 +1,32 @@
...
+public abstract class SourceReaderOperator<OUT> extends AbstractStreamOperator<OUT> implements PushingAsyncDataInput<OUT> {
```
pnowojski: nit: maybe add a note, that we are expecting this to be changed to the concrete class once `SourceReader` interface is introduced?

zhijiangW:
@flinkbot run travis

zhijiangW:
Thanks for the review @pnowojski and I fixed the above nit comments.

The travis failed because of known fragile `Kafka010ProducerITCase`, so ignore it to merge.

PR #9652, opened 2019-09-09
[FLINK-14029][mesos] Update Mesos scheduling behavior to reject all expired offers

piyushnarang:

While digging into why our Flink jobs weren't being scheduled on our internal Mesos setup we noticed that we were hitting Mesos quota limits tied to the way we've set up the Fenzo (https://github.com/Netflix/Fenzo/) library defaults in the Flink project.

Behavior we noticed was that we got a bunch of offers from our Mesos master (50+) out of which only 1 or 2 of them were super skewed and took up a huge chunk of our disk resource quota. Thanks to this we were not sent any new / different offers (as our usage at the time + resource offers reached our Mesos disk quota). As the Flink / Fenzo Mesos scheduling code was not using the 1-2 skewed disk offers they end up expiring. The way we've set up the Fenzo scheduler is to use the default values on when to expire unused offers (120s) and maximum number of unused offer leases at a time (4). Unfortunately as we have a considerable number of outstanding expired offers (50+) we end up in a situation where we reject only 4 or so every 2 mins and we never get around to rejecting the super skewed disk ones which are stopping us from scheduling our Flink job. Thanks to this we end up in a situation where our job is waiting to be scheduled for more than an hour.

An option to work around this is to reject all expired offers at 2 minute expiry time rather than hold on to them. This will allow Mesos to send alternate offers that might be scheduled by Fenzo.

We could also make this aspect of the behavior configurable if needed. Though I'm not sure if it's worth it to add a config knob for this setting to allow users to have Fenzo hold on to expired offers for longer. 

## Brief change log

  - Include missing details on networkMbps, disk resources in the offer and Mesos task startup logs to help debugging. 
  - Update `TaskSchedulerBuilder` to setup the Fenzo `TaskScheduler` to reject all expired offers (rather than just 4). 

## Verifying this change

  - Dependencies (does it add or upgrade a dependency): No
  - The public API, i.e., is any changed class annotated with `@Public(Evolving)`: No
  - The serializers: No
  - The runtime per-record code paths (performance sensitive): No
  - Anything that affects deployment or recovery: JobManager (and its components), Checkpointing, Yarn/Mesos, ZooKeeper: Yes (Mesos deployment)
  - The S3 file system connector: No

## Documentation

  - Does this pull request introduce a new feature? No
  - If yes, how is the feature documented? Not applicable

### DIFF
--- a/flink-mesos/src/main/java/org/apache/flink/mesos/runtime/clusterframework/LaunchableMesosWorker.java
+++ b/flink-mesos/src/main/java/org/apache/flink/mesos/runtime/clusterframework/LaunchableMesosWorker.java
@@ -189,6 +189,8 @@ public class LaunchableMesosWorker implements LaunchableTask {
 				"cpus=" + getCPUs() +
 				", memory=" + getMemory() +
 				", gpus=" + getGPUs() +
+				", disk=" + getDisk() +
+				", network=" + getNetworkMbps() +
 				"}";
 		}
 	}
--- a/flink-mesos/src/main/java/org/apache/flink/mesos/runtime/clusterframework/MesosResourceManager.java
+++ b/flink-mesos/src/main/java/org/apache/flink/mesos/runtime/clusterframework/MesosResourceManager.java
@@ -445,8 +445,9 @@ public class MesosResourceManager extends ResourceManager<RegisteredMesosWorkerN
 
 			LaunchableMesosWorker launchable = createLaunchableMesosWorker(worker.taskID());
 
-			LOG.info("Scheduling Mesos task {} with ({} MB, {} cpus).",
-				launchable.taskID().getValue(), launchable.taskRequest().getMemory(), launchable.taskRequest().getCPUs());
+			LOG.info("Scheduling Mesos task {} with ({} MB, {} cpus, {} gpus, {} disk MB, {} Mbps).",
+				launchable.taskID().getValue(), launchable.taskRequest().getMemory(), launchable.taskRequest().getCPUs(),
+				launchable.taskRequest().getScalarRequests().get("gpus"), launchable.taskRequest().getDisk(), launchable.taskRequest().getNetworkMbps());
 
 			// tell the task monitor about the new plans
 			taskMonitor.tell(new TaskMonitor.TaskGoalStateUpdated(extractGoalState(worker)), selfActor);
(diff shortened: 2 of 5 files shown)


piyushnarang:
Thanks for getting back @tillrohrmann. I was actually thinking about adding a couple of other knobs as well (still in the process of testing these ones - https://github.com/criteo-forks/flink/pull/11/). Basically I wanted to also allow users to configure how long the Fenzo scheduler holds on to unused expired offers (currently defaults to 120s) and allow users to configure how long we want Mesos to not resend a rejected offers (currently defaults to 5s). The folks in our Mesos team suggested lowering the unused expired rejection from 120s -> 1s and the not resending rejected offers from 5s -> 1 min. 

As there are kind of related to this, if you'd be open I could update this review to add those knobs as well (after testing on our end today). Else, I can break this one up into the logging improvements commit and the expired offer change and follow up with a different jira for the other ones. 

Wdyt?

tillrohrmann:
Thanks for getting back so quickly @piyushnarang. I'm a fan of small changes because it makes it easier to review and roll back if needed. Hence, I would suggest to split this PR up as I've suggested and then tackle the other configuration improvements as separate PRs.

Concerning the default values, I guess that you are the experts because you are running it. Hence I would trust you to pick sane default values.

tillrohrmann:
Just out of curiosity what's preventing you from using vanilla Flink? The criteo fork seems to be Flink 1.5. Is it the Mesos support?

piyushnarang:
Ok that sounds good @tillrohrmann. I'll take a stab at updating this with the logging + expire and follow up with a different one for the other two timeout. 

We've been lagging a bit on bumping our Flink version internally (1.6 as of today). We have a couple of custom patches but by and large we've tried to push them upstream (like the AzureFS change we made a while back) so I don't there's a huge difference in our fork and whats on master. It's largely 1.6 with a few changes cherrypicked (also cause some were released in newer Flink versions that we have yet to migrate to..)

piyushnarang:
@tillrohrmann - updated the review. Let me know if this matches what you were thinking regarding the hotfix / breakup.

piyushnarang:
Thanks @tillrohrmann I'll spin up a new jira on the Mesos / Fenzo config knob settings after we've tested a bit and cc you on that.

tillrohrmann:
Sounds good Piyush.

PR #9653, opened 2019-09-09
[FLINK-14014][python] Introduce PythonScalarFunctionRunner to handle the communication with Python worker for Python ScalarFunction execution

dianfu:

*This pull request introduces PythonScalarFunctionRunner to handle the communication with Python worker for Python ScalarFunction execution. Internally, it uses Apache Beam's portability framework for Python UDF execution.*

## Brief change log

  - *It bumps maven-shade-plugin to 3.2.1 as version below 3.1.0 could not work well with ASM 6.0([MSHADE-258](https://issues.apache.org/jira/browse/MSHADE-258)) which is dependent by Beam*
  - *It introduces basic classes such as PythonEnv, PythonFunction, PythonFunctionInfo into flink-table-common which hold the information such as the Python execution environment, the serialized Python functions, etc. These classes are located in flink-table-common because they will be used by the added RelNode which will be introduced later*
  - *Introduces a series of PythonFunctionRunners. They are responsible for using Beam's portability framework for Python UDF execution*
  - *Handle the license statements as it introduces a few third-part dependences are introduced because of we depends on Beam*

## Verifying this change

  - *The added PythonScalarFunctionRunnerTest, BeamTypeUtilsTest*

  - The serializers: (no)

## Documentation


pom.xml
```diff
@@ -1667,7 +1669,7 @@ under the License.
 				<plugin>
 					<groupId>org.apache.maven.plugins</groupId>
 					<artifactId>maven-shade-plugin</artifactId>
-					<version>3.0.0</version>
+					<version>3.1.0</version>
```
hequn8128: Should we bump it to 3.2.1? see FLINK-13453. And would be great if we also update the document about maven-shade-plugin in [dependencies.md](https://ci.apache.org/projects/flink/flink-docs-master/dev/projectsetup/dependencies.html)

flink-table/flink-table-common/src/main/java/org/apache/flink/table/functions/python/PythonFunctionInfo.java
```diff
@@ -0,0 +1,60 @@
...
+package org.apache.flink.table.functions.python;
+
+import org.apache.flink.annotation.Internal;
+import org.apache.flink.util.Preconditions;
+
+import java.io.Serializable;
+
+/**
+ * PythonFunctionInfo contains the execution information of a Python function, such as:
+ * the actual Python function, the input arguments, the input types, the result type, etc.
```
hequn8128: Should we make the comment consistent with the implementation of the class? It seems this class only contains the python function and it's input indexes, i.e., does not contain input types, result type.

flink-python/src/main/java/org/apache/flink/table/functions/python/coders/RowCoder.java
```diff
@@ -0,0 +1,81 @@
...
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * A {@link Coder} for {@link Row}.
+ */
+@Internal
+public class RowCoder extends Coder<Row> {
```
hequn8128: override `consistentWithEquals` to true?

flink-python/src/main/java/org/apache/flink/table/functions/python/coders/RowCoder.java
```diff
@@ -0,0 +1,81 @@
...
+	@Override
+	public void encode(Row row, OutputStream outStream) throws IOException {
+		int len = fieldCoders.length;
+
+		if (row.getArity() != len) {
+			throw new RuntimeException("Row arity of input element does not match coders.");
+		}
+
+		for (int i = 0; i < row.getArity(); i++) {
+			fieldCoders[i].encode(row.getField(i), outStream);
```
hequn8128: Should take the null field into consideration. Coder in Beam can not encode null values. Same for `BaseRowCoder`.

flink-python/src/main/java/org/apache/flink/table/functions/python/BeamTypeUtils.java
```diff
@@ -0,0 +1,548 @@
...
+			return null;
+		}
+
+		@Override
+		public Coder visit(LogicalType other) {
+			return null;
+		}
+	}
+
+	private static class LogicalTypeToBlinkCoderConverter implements LogicalTypeVisitor<Coder> {
```
hequn8128: Make this class extends from the `LogicalTypeToCoderConverter`? Most parts of the two classes are same.

flink-python/src/main/java/org/apache/flink/table/functions/python/BeamTypeUtils.java
```diff
@@ -0,0 +1,548 @@
...
+import org.apache.flink.table.types.logical.YearMonthIntervalType;
+import org.apache.flink.table.types.logical.ZonedTimestampType;
+
+import org.apache.beam.sdk.coders.Coder;
+
+/**
+ * Utilities for converting Flink data types to Beam data types.
+ */
+@Internal
+public final class BeamTypeUtils {
```
hequn8128: Why only support BigIntType and RowType in these Converter classes?

flink-python/src/test/java/org/apache/flink/table/functions/python/BeamTypeUtilsTest.java
```diff
@@ -0,0 +1,65 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+/**
+ * Tests for {@link BeamTypeUtils}.
+ */
+public class BeamTypeUtilsTest {
```
hequn8128: Also add tests for `BaseRowCoder`

flink-python/src/main/java/org/apache/flink/python/PythonFunctionExecutionResultReceiver.java
```diff
@@ -0,0 +1,35 @@
...
+
+package org.apache.flink.python;
+
+import org.apache.flink.annotation.Internal;
+
+/**
+ * A receiver for the execution results of Python functions.
+ */
+@Internal
+public interface PythonFunctionExecutionResultReceiver<T> {
```
hequn8128: Use `FnDataReceiver` in beam directly.

flink-python/src/main/java/org/apache/flink/python/PythonFunctionRunner.java
```diff
@@ -0,0 +1,60 @@
...
+ *
+ * @param <IN> Type of the input elements.
+ */
+@Internal
+public interface PythonFunctionRunner<IN> {
+
+	/**
+	 * Prepares the Python function runner, such as preparing the Python execution environment, etc.
+	 */
+	void open() throws Exception;
```
hequn8128: Can we add these methods later? I am asking because these methods have been implemented in its child class however have not been covered in any tests. 
Or can we add some tests to cover these methods?

hequn8128:
@dianfu I have finished my review now. Thank you very much for the nice PR. 

I have not reviewed the last pr as there are related failures in the CI report. I will look at it once the CI passed. Is it ok with you?

Also, would be great if @sunjincheng121 can also take a look.

Best, Hequn

flink-table/flink-table-common/src/main/java/org/apache/flink/table/functions/python/PythonEnv.java
```diff
@@ -0,0 +1,82 @@
...
+		// python function is executed in a separate process
+		PROCESS,
+
+		// python function is executed in a docker container
+		DOCKER,
+
+		// python function is executed in an external non-managed process
+		EXTERNAL,
+
+		UNKNOWN,
```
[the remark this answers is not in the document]
dianfu: What about removing the UNKNOWN type?

flink-table/flink-table-common/src/main/java/org/apache/flink/table/functions/python/PythonFunctionInfo.java
dianfu: Good catch. Originally it contains the input types and result type. Forgot to update the comments when removing those fields.

flink-python/pyflink/proto/flink-fn-execution.proto
```diff
@@ -0,0 +1,96 @@
...
+
+package org.apache.flink.fn_execution.v1;
+
+option java_package = "org.apache.flink.fnexecution.v1";
+option java_outer_classname = "FlinkFnApi";
+
+// User-defined function definition. It supports chaining functions, that's, the execution
+// result of one user-defined function as the input of another user-defined function.
+message UserDefinedFunction {
+    message Input {
```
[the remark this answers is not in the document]
dianfu: Makes sense to me.

flink-python/pyflink/proto/flink-fn-execution.proto
```diff
@@ -0,0 +1,96 @@
...
+        DATETIME = 9;
+        BOOLEAN = 10;
+        BINARY = 11;
+        VARBINARY = 12;
+        CHAR = 13;
+        VARCHAR = 14;
+        ARRAY = 15;
+        MAP = 16;
+        MULTISET = 17;
+        ROW = 18;
```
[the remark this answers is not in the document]
dianfu: Row is a special type and it could be used to represent the type of a table. So it seems more reasonable to put it at last of the type list. What do you think?

flink-python/src/main/java/org/apache/flink/table/functions/python/BeamTypeUtils.java
dianfu: I want to make sure this PR only focus on the runner part. For the types, I want to add them in a separate PR. BigIntType and RowType are added in this PR because they are the minimum types needed to run end to end use cases. Does that make sense to you? If so, I will create a follow up JIRA.

flink-python/src/main/java/org/apache/flink/python/PythonFunctionRunner.java
dianfu: These methods can only be tested in integration test. What about revisiting these methods when adding the integration tests in FLINK-14018. However, I'm in favor of keeping them as they are part of the functionality of PythonFunctionRunner and are also very important part. Just removing them because they are not test right now seems wired for me.

dianfu:
@hequn8128 Thanks a lot for your review. Very appreciated! Most comments are reasonable for me with just a few exceptions and I have explained under the comment. Please see if that makes sense to you. Thanks again!

flink-table/flink-table-common/src/main/java/org/apache/flink/table/functions/python/PythonEnv.java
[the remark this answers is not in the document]
hequn8128: Sounds good. How about remove DOCKER and EXTERNAL for now? It seems we are going to only support PROCESS first?

flink-table/flink-table-common/src/main/java/org/apache/flink/table/functions/python/PythonEnv.java
```diff
@@ -0,0 +1,82 @@
...
+import java.io.Serializable;
+
+/**
+ * Python execution environments.
+ */
+public final class PythonEnv implements Serializable {
+
+	private static final long serialVersionUID = 1L;
+
+	private final String pythonExec;
```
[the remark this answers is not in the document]
hequn8128: Thanks a lot for adding the comments. Maybe rename the variable name to `pythonExecFilePath`?  Same for `pipExec`

flink-table/flink-table-common/src/main/java/org/apache/flink/table/functions/python/PythonEnv.java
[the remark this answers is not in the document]
dianfu: +1

flink-python/pyflink/proto/flink-fn-execution.proto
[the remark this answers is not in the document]
hequn8128: New types would be added later, and we can't change the tag number for ROW at that time. In this case, we can't make sure it is at the last. 
If we want to separate ROW from the others, we can put ROW at the front, this can also make encode ROW one less byte.(My main concern is the performance, see [here](https://developers.google.com/protocol-buffers/docs/pythontutorial) about the tag number)
What do you think?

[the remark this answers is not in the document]
dianfu: Good catch for the performance issue. Make sense to me. +1 to put it at the first of the list.

flink-python/src/main/java/org/apache/flink/python/AbstractPythonFunctionRunner.java
```diff
@@ -0,0 +1,296 @@
...
+			remoteBundle.close();
+		} catch (Throwable t) {
+			throw new RuntimeException("Failed to close remote bundle", t);
+		}
+	}
+
+	@Override
+	public void processElement(IN element) {
+		try {
+			baos.reset();
```
hequn8128: Add some log logic here, e.g., LOG.debug("Process element: {}", element)

flink-python/src/main/java/org/apache/flink/python/PythonFunctionRunnerContext.java
```diff
@@ -0,0 +1,39 @@
...
+import org.apache.beam.runners.core.construction.graph.ExecutableStage;
+import org.apache.beam.runners.fnexecution.control.RemoteBundle;
+import org.apache.beam.runners.fnexecution.control.StageBundleFactory;
+
+/**
+ * The context used to manage the Python execution environment. It can be used to create
+ * the {@link StageBundleFactory} which is the entry point to manage the Python execution environment.
+ */
+@Internal
+public interface PythonFunctionRunnerContext extends AutoCloseable {
```
hequn8128: Remove this interface? It seems not be used.

flink-python/src/main/java/org/apache/flink/python/AbstractPythonFunctionRunner.java
```diff
@@ -0,0 +1,296 @@
...
+	public abstract Coder<OUT> getOutputCoder();
+
+	/**
+	 * An {@link StateRequestHandler} implementation which does nothing.
+	 */
+	public static class NoOpStateRequestHandler implements StateRequestHandler {
+
+		@Override
+		public CompletionStage<BeamFnApi.StateResponse.Builder> handle(BeamFnApi.StateRequest request) {
+			return null;
```
hequn8128: Should we throw exceptions here(to inform that state request could not be handled correctly)?  Maybe, we can use `StateRequestHandler.unsupported()` directly. What do you think?

flink-python/src/main/java/org/apache/flink/python/AbstractPythonFunctionRunner.java
```diff
@@ -0,0 +1,296 @@
...
+	}
+
+	private BundleProgressHandler getProgressHandler() {
+		return new NoOpBundleProgressHandler();
+	}
+
+	/**
+	 * An {@link BundleProgressHandler} implementation which does nothing.
+	 */
+	private static class NoOpBundleProgressHandler implements BundleProgressHandler {
```
hequn8128: Remove this class. Use `BundleProgressHandler.ignored()` directly.

flink-python/src/main/java/org/apache/flink/table/functions/python/AbstractPythonScalarFunctionRunner.java
```diff
@@ -0,0 +1,225 @@
...
+	private static final String TRANSFORM_ID = "transform";
+
+	private static final String MAIN_INPUT_NAME = "input";
+	private static final String MAIN_OUTPUT_NAME = "output";
+
+	private static final String INPUT_CODER_ID = "input_coder";
+	private static final String OUTPUT_CODER_ID = "output_coder";
+	private static final String WINDOW_CODER_ID = "window_coder";
+
+	private static final String WINDOW_STRATEGY = "windowing-strategy";
```
hequn8128: change to `windowing_strategy`?  Make it consistent with other names.

flink-python/src/main/java/org/apache/flink/table/functions/python/coders/BaseRowCoder.java
```diff
@@ -0,0 +1,145 @@
...
+			throw new RuntimeException("Row arity of input element does not match coders.");
+		}
+
+		// write a null mask
+		reusableOutputStream.reset(outStream);
+		writeNullMask(len, row, reusableOutputStream);
+
+		for (int i = 0; i < row.getArity(); i++) {
+			if (!row.isNullAt(i)) {
+				fieldCoders[i].encode(TypeGetterSetters.get(row, i, fieldTypes[i]), outStream);
```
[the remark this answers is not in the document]
dianfu: +1

flink-python/src/main/java/org/apache/flink/python/AbstractPythonFunctionRunner.java
dianfu: Personally I think it is meaningless:
1. For UDF, there may be multiple udfs in one operator, each udf may refer to different columns of the original inputs. It's difficult for users to determine which column printed here is corresponding to which column from the original input after all kinds of rules optimization.
2. Seems that there is no built-in operator which logged the input elements information.

flink-python/src/main/java/org/apache/flink/python/AbstractPythonFunctionRunner.java
dianfu: +1 to use StateRequestHandler.unsupported()

dianfu:
@hequn8128 Thanks a lot for the review. I have updated the PR accordingly. Looking forward to your feedback.

flink-python/src/main/java/org/apache/flink/python/AbstractPythonFunctionRunner.java
hequn8128: OK, make sense.

hequn8128:
@dianfu Thanks a lot for the update. LGTM from my side.
Would be great if @sunjincheng121 can take another look.

flink-python/src/main/java/org/apache/flink/python/AbstractPythonFunctionRunner.java
```diff
@@ -0,0 +1,270 @@
...
+	 */
+	public abstract Coder<IN> getInputCoder();
+
+	/**
+	 * Returns the coder for execution results.
+	 */
+	public abstract Coder<OUT> getOutputCoder();
+
+	private String createEmptyRetrievalToken() throws Exception {
+		final File retrievalToken = Files.createTempFile("retrieval_token", "json").toFile();
```
WeiZhong94: If this part is running in TaskManager, the better locations to store temporary files is `getContainingTask().getEnvironment().getTaskManagerInfo().getTmpDirectories()`. And maybe we should delete it after finish bundle? At least we should call `retrievalToken.deleteOnExit()` to ensure that it will be deleted when jvm exits.

dianfu:
@WeiZhong94 Thanks for the review. Make sense and have updated the PR.

WeiZhong94:
@dianfu Thanks a lot for the update. LGTM and +1 to merge from my side.

hequn8128:
@dianfu Great job! Merging...

tillrohrmann:
Please be a bit more careful when bumping build plugin versions. The introduced version `3.2.1` seems to fail the builds for our end-to-end tests: https://travis-ci.org/apache/flink/builds/591085384. When doing these kind of changes I would suggest to do it as a separate PR and to run at least every end-to-end test on Travis before merging.

dianfu:
@tillrohrmann Very appreciated for your suggestions. That makes sense to me. Will be more careful  for this kind of changes next time, e.g. doing it in a separate PR and making sure end-to-end test passed.

hequn8128:
@tillrohrmann Thanks a lot for the notice. Will be more careful next time.

PR #9664, opened 2019-09-10
[FLINK-14043] Speed up SavepointMigrationTestBase sub classes

tillrohrmann:

Since all SavepointMigrationTestBase sub classes rely on the
MigrationTestUtils.AccumulatorCountingSink which uses user code accumulators
in order to communicate with the test driver, we set the heartbeat interval
to 300ms in order to speed the test execution up. The reason this works is
that Flink transports user code accumulators from the TM to the JM via
the heartbeats. Hence, the heartbeat interval represents the lower boundary
for the test completion.

## Verifying this change

  - The serializers: (no)

## Documentation


tillrohrmann:
Thanks for the review @aljoscha and @GJL. Merging this PR now.

PR #9675, opened 2019-09-12
[FLINK-13953] [runtime] Facilitate enabling new scheduler in MiniCluster Tests

zhuzhurk:

Currently, tests using the `MiniCluster` use the legacy scheduler by default. Once the new scheduler is implemented, we should run tests with the new scheduler enabled. However, it is not expected that all tests will pass immediately. Therefore, it should be possible to enable the new scheduler for a subset of tests. 

In the first step the tests should be able to run manually against new scheduler.

*Acceptance Criteria*
 * A junit test category `AlsoRunWithSchedulerNG` can be used to mark MiniCluster tests.
 * A new maven profile `scheduler-ng` will be enabled to support running `AlsoRunWithSchedulerNG` annotated tests with the new scheduler.

## Brief change log

The change is inspired by the prototype from @tillrohrmann at https://github.com/tillrohrmann/flink/commit/dee9e7d82f65fa1697489de743e84dcf02d76711.
  - *Added a category marker for new scheduler testing*
  - *Enabled setting scheduler type in MiniClusterConfiguration via environment variable*
  - *Added a maven profile to run annotated MiniCluster tests for the new scheduler*
  - *other hotfixes, details see each hotfix commit*

## Verifying this change

  - *Added UTs for MiniClusterConfiguration changes*
  - *Manually verified the change by running command "mvn verify -Dscheduler-ng". Verified that only the tests annotated with `AlsoRunWithSchedulerNG` are executed in this way. Verified that the scheduler type can be specified by setting maven profiles or properties.*

## Documentation

### DIFF
--- a/flink-connectors/flink-connector-elasticsearch6/pom.xml
+++ b/flink-connectors/flink-connector-elasticsearch6/pom.xml
@@ -195,7 +195,6 @@ under the License.
 			<plugin>
 				<groupId>org.apache.maven.plugins</groupId>
 				<artifactId>maven-surefire-plugin</artifactId>
-				<version>2.12.2</version>
 				<configuration>
 					<systemPropertyVariables>
 						<jna.nosys>true</jna.nosys>
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/jobmaster/JobMaster.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/jobmaster/JobMaster.java
@@ -261,8 +261,10 @@ public class JobMaster extends FencedRpcEndpoint<JobMasterId> implements JobMast
 
 		this.shuffleMaster = checkNotNull(shuffleMaster);
 
-		this.jobManagerJobMetricGroup = jobMetricGroupFactory.create(jobGraph);
-		this.schedulerNG = createScheduler(jobManagerJobMetricGroup);
+		JobManagerJobMetricGroup jobManagerJobMetricGroup = jobMetricGroupFactory.create(jobGraph);
+		SchedulerNG schedulerNG = createScheduler(jobManagerJobMetricGroup);
+		assignScheduler(schedulerNG, jobManagerJobMetricGroup);
+
 		this.jobStatusListener = null;
 
 		this.resourceManagerConnection = null;
@@ -810,15 +812,24 @@ public class JobMaster extends FencedRpcEndpoint<JobMasterId> implements JobMast
 			log);
 	}
 
-	private void assignScheduler(
+	private void reassignScheduler(
 			SchedulerNG newScheduler,
 			JobManagerJobMetricGroup newJobManagerJobMetricGroup) {
 		validateRunsInMainThread();
 		checkState(schedulerNG.requestJobStatus().isTerminalState());
 		checkState(jobManagerJobMetricGroup == null);
 
-		schedulerNG = newScheduler;
-		jobManagerJobMetricGroup = newJobManagerJobMetricGroup;
+		assignScheduler(newScheduler, newJobManagerJobMetricGroup);
+	}
+
+	private void assignScheduler(
+			SchedulerNG scheduler,
+			JobManagerJobMetricGroup jobManagerJobMetricGroup) {
+
+		this.schedulerNG = scheduler;
+		this.jobManagerJobMetricGroup = jobManagerJobMetricGroup;
+
+		log.info("Scheduler {} is used for job {}.", schedulerNG, jobGraph.getJobID());
 	}
 
 	private void resetAndStartScheduler() throws Exception {
@@ -837,7 +848,7 @@ public class JobMaster extends FencedRpcEndpoint<JobMasterId> implements JobMast
 			schedulerAssignedFuture = schedulerNG.getTerminationFuture().handle(
 				(ignored, throwable) -> {
 					newScheduler.setMainThreadExecutor(getMainThreadExecutor());
-					assignScheduler(newScheduler, newJobManagerJobMetricGroup);
+					reassignScheduler(newScheduler, newJobManagerJobMetricGroup);
 					return null;
 				}
 			);
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/minicluster/MiniClusterConfiguration.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/minicluster/MiniClusterConfiguration.java
@@ -26,6 +26,7 @@ import org.apache.flink.configuration.TaskManagerOptions;
 import org.apache.flink.configuration.UnmodifiableConfiguration;
 import org.apache.flink.runtime.akka.AkkaUtils;
 import org.apache.flink.util.Preconditions;
+import org.apache.flink.util.StringUtils;
 
 import javax.annotation.Nullable;
 
@@ -38,7 +39,7 @@ import static org.apache.flink.runtime.minicluster.RpcServiceSharing.SHARED;
  */
 public class MiniClusterConfiguration {
 
-	static final String SCHEDULER_TYPE_KEY = "schedulerType";
+	static final String SCHEDULER_TYPE_KEY = JobManagerOptions.SCHEDULER.key();
 
 	private final UnmodifiableConfiguration configuration;
 
@@ -66,7 +67,10 @@ public class MiniClusterConfiguration {
 	}
 
 	private UnmodifiableConfiguration generateConfiguration(final Configuration configuration) {
-		final String schedulerType = System.getProperty(SCHEDULER_TYPE_KEY, JobManagerOptions.SCHEDULER.defaultValue());
+		String schedulerType = System.getProperty(SCHEDULER_TYPE_KEY);
+		if (StringUtils.isNullOrWhitespaceOnly(schedulerType)) {
+			schedulerType = JobManagerOptions.SCHEDULER.defaultValue();
+		}
 
 		if (!configuration.contains(JobManagerOptions.SCHEDULER)) {
 			configuration.setString(JobManagerOptions.SCHEDULER, schedulerType);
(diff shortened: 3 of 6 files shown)


pom.xml
```diff
@@ -637,6 +639,18 @@ under the License.
 	</dependencyManagement>
 
 	<profiles>
+		<profile>
```
zhuzhurk: This profile is not necessary but simplifies the way to run tests for annotated MiniClusters again the new scheduler.
We can run "mvn verify -Dscheduler.type=ng -Dtest.groups=org.apache.flink.testutils.junit.category.AlsoRunWithSchedulerNG" instead.
And to find out whether all the MiniCluster tests(even not annotated with `AlsoRunWithSchedulerNG`) can pass against the new scheduler, we can run "mvn verify -Dscheduler.type=ng".

flink-connectors/flink-connector-elasticsearch6/pom.xml
```diff
@@ -195,7 +195,6 @@ under the License.
 			<plugin>
 				<groupId>org.apache.maven.plugins</groupId>
 				<artifactId>maven-surefire-plugin</artifactId>
-				<version>2.12.2</version>
```
zhuzhurk: Without this fix, the testing for this module fails when `groups` is specified in maven building. The error is as below:

> initializationError(org.junit.runner.manipulation.Filter)  Time elapsed: 0.002 sec  <<< ERROR!
> java.lang.Exception: No tests found matching *org.apache.flink.testutils.junit.category.AlsoRunWithSchedulerNG from org.junit.runner.Request$1@7e0ea639
> 	at org.junit.internal.requests.FilterRequest.getRunner(FilterRequest.java:40)
> 	at org.junit.runner.JUnitCore.run(JUnitCore.java:115)
> 	at org.apache.maven.surefire.junitcore.JUnitCoreWrapper.execute(JUnitCoreWrapper.java:62)
> 	at org.apache.maven.surefire.junitcore.JUnitCoreProvider.invoke(JUnitCoreProvider.java:139)
> 	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
> 	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
> 	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
> 	at java.lang.reflect.Method.invoke(Method.java:498)
> 	at org.apache.maven.surefire.util.ReflectionUtils.invokeMethodWithArray(ReflectionUtils.java:189)
> 	at org.apache.maven.surefire.booter.ProviderFactory$ProviderProxy.invoke(ProviderFactory.java:165)
> 	at org.apache.maven.surefire.booter.ProviderFactory.invokeProvider(ProviderFactory.java:85)
> 	at org.apache.maven.surefire.booter.ForkedBooter.runSuitesInProcess(ForkedBooter.java:113)
> 	at org.apache.maven.surefire.booter.ForkedBooter.main(ForkedBooter.java:75)

flink-test-utils-parent/flink-test-utils-junit/src/main/java/org/apache/flink/testutils/junit/category/AlsoRunWithSchedulerNG.java
```diff
@@ -0,0 +1,25 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.testutils.junit.category;
+
+/**
+ * Category marker interface to run tests with SchedulerNG.
+ */
+public interface AlsoRunWithSchedulerNG {
```
zhuzhurk: No test is annotated since the `DefaultScheduler` has not implemented the `startScheduling` interface yet.

zhuzhurk: Have verified the annotation manually with a temporary patch. The annotated test can be found out and executed with new scheduler, although it hangs in the scheduling process.

flink-runtime/src/main/java/org/apache/flink/runtime/jobmaster/JobMaster.java
```diff
@@ -810,15 +812,24 @@ private void startHeartbeatServices() {
...
+	}
+
+	private void assignScheduler(
+			SchedulerNG scheduler,
+			JobManagerJobMetricGroup jobManagerJobMetricGroup) {
+
+		this.schedulerNG = scheduler;
+		this.jobManagerJobMetricGroup = jobManagerJobMetricGroup;
+
+		log.info("Scheduler {} is used for job {}.", schedulerNG, jobGraph.getJobID());
```
GJL: I wouldn't log this information here yet:
-  Scheduler configuration is per cluster at the moment. However, here it is logged per job.
- Potentially logged multiples times per job (if JobMaster loses and regains leadership)
- One can differentiate between the schedulers by looking at the cluster configuration, which is already being logged

Bottom line is that I would drop this commit.

flink-runtime/src/main/java/org/apache/flink/runtime/minicluster/MiniClusterConfiguration.java
```diff
@@ -38,6 +38,8 @@
  */
 public class MiniClusterConfiguration {
 
+	static final String SCHEDULER_TYPE_KEY = "schedulerType";
```
GJL: Not sure if there are any conventions about how to name system properties. Maybe this should be also named `jobmanager.scheduler` to keep it aligned with `JobManagerOptions.SCHEDULER`.

flink-runtime/src/test/java/org/apache/flink/runtime/minicluster/MiniClusterConfigurationTest.java
```diff
@@ -0,0 +1,59 @@
...
+/**
+ * Tests for the {@link MiniClusterConfiguration}.
+ */
+public class MiniClusterConfigurationTest extends TestLogger {
+
+	private static final String TEST_SCHEDULER_NAME = "test-scheduler";
+
+	@Test
+	public void testSchedulerType_setViaSystemProperty() {
+		System.setProperty(MiniClusterConfiguration.SCHEDULER_TYPE_KEY, TEST_SCHEDULER_NAME);
```
GJL: Modifying global state is potentially dangerous. Whether the tests will pass at all, depends on [surefire's JVM fork settings](https://maven.apache.org/surefire/maven-surefire-plugin/examples/fork-options-and-parallel-execution.html). We should restore the system properties to its previous state after running the tests.

See:
* https://stackoverflow.com/questions/54025381/system-setproperty-in-junit-tests
* https://stefanbirkner.github.io/system-rules/#RestoreSystemProperties

pom.xml
GJL: I think the profile is fine.

pom.xml
```diff
@@ -148,6 +148,8 @@ under the License.
 		<hivemetastore.hadoop.version>2.7.5</hivemetastore.hadoop.version>
 		<jamicmp.referenceVersion>1.8.0</jamicmp.referenceVersion>
 		<japicmp.outputDir>tools/japicmp-output</japicmp.outputDir>
+		<scheduler.type>legacy</scheduler.type>
```
GJL: Is it necessary to set a default scheduler in the build configuration again? We already have a default set in Java code:

```
final String schedulerType = System.getProperty(SCHEDULER_TYPE_KEY, JobManagerOptions.SCHEDULER.defaultValue());
```

zhuzhurk: You are right. I think we can remove it.

flink-runtime/src/test/java/org/apache/flink/runtime/minicluster/MiniClusterConfigurationTest.java
zhuzhurk: Good catch! Will restore the state in the tests.

flink-runtime/src/main/java/org/apache/flink/runtime/minicluster/MiniClusterConfiguration.java
zhuzhurk: No such convention yet. But I think it's fine to use the JobManagerOptions.SCHEDULER key as the property key as they take the same values.

flink-runtime/src/main/java/org/apache/flink/runtime/jobmaster/JobMaster.java
zhuzhurk: Agreed that we can infer from the cluster config log for the scheduler used. But it is not very straight forward, especially when it is not configured and the default config is used.
As we are already logging the failover strategy for each job, I think logging the scheduler for each job is acceptable.
To avoid logging it for multiple times, we can just log it the JobMaster constructor.

GJL: You are right that in the absence of an explicit configuration entry/system variable, one has to check the config option's default value. However, there are still reasons against logging in the JobMaster constructor:
* Potentially logged multiple times in sesssion clusters.
* It is not expected that the end-user has to change the scheduler. Hopefully `legacy` will be removed soon and there are no concrete plans to include additional schedulers in the repository at the moment.
* No proper `toString()` implementation at the moment (classname@hexcode does not look nice).
* Side effects in constructor should be avoided.

Since the rest of the PR looks good, I will drop the commit for now.

pom.xml
```diff
@@ -148,6 +148,8 @@ under the License.
 		<hivemetastore.hadoop.version>2.7.5</hivemetastore.hadoop.version>
 		<jamicmp.referenceVersion>1.8.0</jamicmp.referenceVersion>
 		<japicmp.outputDir>tools/japicmp-output</japicmp.outputDir>
+		<scheduler.type></scheduler.type>
```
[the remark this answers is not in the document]
zhuzhurk: Good idea.

flink-runtime/src/main/java/org/apache/flink/runtime/jobmaster/JobMaster.java
zhuzhurk: Fine.

PR #9679, opened 2019-09-14
[hotfix] Fix the comments related to side output.

gaoyunhaii:

The comments related to side output in the `Output#collect` method seem to be inconsistent with the logic. Therefore, this PR tries to address it.

## Brief change log
-  ca01daf37331ff493f367143af6a9a9c2cb28866 fixes the comments.

## Verifying this change

  - Dependencies (does it add or upgrade a dependency):  **no**
  - The serializers: **no**

## Documentation

### DIFF
--- a/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/OperatorChain.java
+++ b/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/OperatorChain.java
@@ -539,7 +539,7 @@ public class OperatorChain<OUT, OP extends StreamOperator<OUT>> implements Strea
 		@Override
 		public void collect(StreamRecord<T> record) {
 			if (this.outputTag != null) {
-				// we are only responsible for emitting to the main input
+				// we are not responsible for emitting to the main output.
 				return;
 			}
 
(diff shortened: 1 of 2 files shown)


flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/OperatorChain.java
```diff
@@ -549,7 +549,7 @@ public void collect(StreamRecord<T> record) {
 		@Override
 		public <X> void collect(OutputTag<X> outputTag, StreamRecord<X> record) {
 			if (this.outputTag == null || !this.outputTag.equals(outputTag)) {
-				// we are only responsible for emitting to the side-output specified by our
+				// we are not responsible for emitting to the side-output specified by this
```
KarmaGYZ: I don't think the origin comment is wrong. Same to `RecordWriterOutput.java`

gaoyunhaii: I think if the output tag is null, this output should be responsible for emitting to the main output instead of one side output.

zentol: The original comment was correct, but I like the new wording and think it is more helpful than the previous one.

PR #9681, opened 2019-09-16
[FLINK-13450][test] Adjust tests to tolerate arithmetic differences between x86 and ARM

wangxiyuan:

`math.log` and `math.pow` can actually return slightly different results across platforms because of hardware optimizations. We should use StrictMath instead to ensure the answers are already the same. 

## Brief change log

  - Changed Math.log to StrictMath.log
  - Changed Math.pow to StrictMath.pow

## Verifying this change

This change is already covered by existing tests

## Documentation

### DIFF
--- a/flink-table/flink-table-planner/src/main/scala/org/apache/flink/table/plan/cost/DataSetCost.scala
+++ b/flink-table/flink-table-planner/src/main/scala/org/apache/flink/table/plan/cost/DataSetCost.scala
@@ -109,7 +109,7 @@ class DataSetCost(val rowCount: Double, val cpu: Double, val io: Double) extends
     if (n == 0) {
       return 1.0
     }
-    Math.pow(d, 1 / n)
+    StrictMath.pow(d, 1 / n)
   }
 
   def plus(other: RelOptCost): RelOptCost = {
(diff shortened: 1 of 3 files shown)


wangxiyuan:
@TisonKun, hi, thanks for your comment. I don't know what scope is better, since this pr changed many modules.  any suggestion? Thanks.

TisonKun:
Hi @wangxiyuan I revoke my change request above. After taking a look at the corresponding issue I know the root cause of this pull request is to adjust test so sort it into test component makes sense to me. Besides, what do you think if we change the title of the pull request and the commit to "Adjust tests to tolerate arithmetic differences between x86 and ARM" as the original JIRA written? I think it is more descriptive.

TisonKun:
@flinkbot run travis

StephanEwen:
Could we reduce the changes to the scope where we really need strict results?
It looks like this mainly matters for result verification (some tests) and for the SQL functions.

In other parts, like exponential backoff, strict precision doesn't actually matter.
In some parts, like samplers, the performance is more important than the precision, so we should not use StrictMath there at all.

StephanEwen:
To clarify how this relates to my previous comment ("have just StrictMath"):
Within a module (like SQL scalar functions) I would not try to switch back and forth between Math and StrictMath. But between different uses (sampler versus SQL function) we can differentiate where we need it and where not.

wangxiyuan:
@StephanEwen Thanks for your review. ThenI just change the place where affect the test.

wangxiyuan:
can anyone review this PR? Thanks.

wangxiyuan:
@StephanEwen may this PR be visited again?

StephanEwen:
Looks good to me.
Merging this...

PR #9684, opened 2019-09-16
[FLINK-13992][coordination] Refactor Optional parameter in InputGateWithMetrics#updateMetrics

TisonKun:

As consensus from community code style discussion, in `InputGateWithMetrics#updateMetrics` we can refactor to reduce the usage of Optional parameter.

cc @azagrebin 

## Verifying this change

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/InputGateWithMetrics.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/taskmanager/InputGateWithMetrics.java
@@ -67,12 +67,12 @@ public class InputGateWithMetrics extends InputGate {
 
 	@Override
 	public Optional<BufferOrEvent> getNext() throws IOException, InterruptedException {
-		return updateMetrics(inputGate.getNext());
+		return inputGate.getNext().map(this::updateMetrics);
 	}
 
 	@Override
 	public Optional<BufferOrEvent> pollNext() throws IOException, InterruptedException {
-		return updateMetrics(inputGate.pollNext());
+		return inputGate.pollNext().map(this::updateMetrics);
 	}
 
 	@Override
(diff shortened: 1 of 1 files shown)


TisonKun:
Thanks for your review @azagrebin!

I'm not very sure about your addition. IMO it is good to separate code style changes with feature efforts. In your words it seems you also agree this opinion but there is an "although" and following this statement

> The small code style improvements are not associated with any feature efforts and clutter the history which makes understanding the original reason for the code lines harder.

What best practice do you suggest here exactly? Could you explain a bit?

TisonKun:
ping @azagrebin as a reminder :-)

azagrebin:
Sorry @TisonKun something is messed up with my Github notifications :)

> IMO it is good to separate code style changes with feature efforts. In your words it seems you also agree this opinion but there is an "although" and following this statement

I did not mean that we should separate code style changes with feature efforts. On the contrary, we usually avoid changes which are not related to some behavioural change, like feature or bug fix.

If you do a lot of changes to some method or class to actually change something in Flink behaviour like feature or bug fix, you can also refactor its code style. This kind of changes are interesting to follow in the git history (annotate lines in IDE: commit, who changed and why) and they have value to understand the code.

If you change just code style, nothing actually changes in Flink behaviour but you rewrite the git history for the changed lines. When somebody looks into the history of the changed lines, she is probably more interested in the previous changes and their description which is hidden after the code style improvement.

All-in-all, it is again a trade-off. We want to have a good code style in the code for readability. On the other hand, we want to have a clear git history for the same purpose of readability. We usually choose the git history over the minor changes like code style because the git history is more valuable for understanding. That is why we prefer not to do unrelated changes as well and do not refactor older parts of code without a related purpose.

Hope it helps.

cc @tillrohrmann @pnowojski

PR #9686, opened 2019-09-16
[FLINK-14069] [core] Enable TimeUtils for all time units labels supported by scala Duration

zhuzhurk:

*Currently we are using scala Duration to parse duration configs. *
*The supported time unit labels are: 
{ DAYS -> "d day", HOURS -> "h hour", MINUTES -> "min minute", SECONDS -> "s sec second", MILLISECONDS -> "ms milli millisecond", MICROSECONDS -> "µs micro microsecond", NANOSECONDS -> "ns nano nanosecond" }*
*We want to use Flink TimeUtils to parse the duration configuration, as a step to let flink core get rid of scala dependencies. *
*In order not to break existing jobs, TimeUtils must be able to parse all time unit labels supported by scala Duration.*

## Brief change log

  - *Improved TimeUtils to able to parse more time units*

## Verifying this change

  - *Added unit tests*

## Documentation

### DIFF
--- a/flink-core/src/main/java/org/apache/flink/util/TimeUtils.java
+++ b/flink-core/src/main/java/org/apache/flink/util/TimeUtils.java
@@ -19,7 +19,11 @@
 package org.apache.flink.util;
 
 import java.time.Duration;
+import java.time.temporal.ChronoUnit;
+import java.util.Collections;
+import java.util.HashMap;
 import java.util.Locale;
+import java.util.Map;
 
 import static org.apache.flink.util.Preconditions.checkArgument;
 import static org.apache.flink.util.Preconditions.checkNotNull;
@@ -29,14 +33,28 @@ import static org.apache.flink.util.Preconditions.checkNotNull;
  */
 public class TimeUtils {
 
+	private static final Map<String, ChronoUnit> LABEL_TO_UNIT_MAP = Collections.unmodifiableMap(initMap());
+
 	/**
 	 * Parse the given string to a java {@link Duration}.
-	 * The string is like "123ms", "321s", "12min" and such.
+	 * The string is in format "{length value}{time unit label}", e.g. "123ms", "321 s".
+	 * If no time unit label is specified, it will be considered as milliseconds.
+	 *
+	 * <p>Supported time unit labels are:
+	 * <ul>
+	 *     <li>DAYS： "d", "day"</li>
+	 *     <li>HOURS： "h", "hour"</li>
+	 *     <li>MINUTES： "min", "minute"</li>
+	 *     <li>SECONDS： "s", "sec", "second"</li>
+	 *     <li>MILLISECONDS： "ms", "milli", "millisecond"</li>
+	 *     <li>MICROSECONDS： "µs", "micro", "microsecond"</li>
+	 *     <li>NANOSECONDS： "ns", "nano", "nanosecond"</li>
+	 * </ul>
 	 *
 	 * @param text string to parse.
 	 */
 	public static Duration parseDuration(String text) {
-		checkNotNull(text, "text");
+		checkNotNull(text);
 
 		final String trimmed = text.trim();
 		checkArgument(!trimmed.isEmpty(), "argument is an empty- or whitespace-only string");
@@ -50,7 +68,7 @@ public class TimeUtils {
 		}
 
 		final String number = trimmed.substring(0, pos);
-		final String unit = trimmed.substring(pos).trim().toLowerCase(Locale.US);
+		final String unitLabel = trimmed.substring(pos).trim().toLowerCase(Locale.US);
 
 		if (number.isEmpty()) {
 			throw new NumberFormatException("text does not start with a number");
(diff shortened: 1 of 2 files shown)


flink-core/src/main/java/org/apache/flink/util/TimeUtils.java
```diff
@@ -29,9 +33,23 @@
...
 
+	private static final Map<String, ChronoUnit> LABEL_TO_UNIT_MAP = Collections.unmodifiableMap(initMap());
+
 	/**
 	 * Parse the given string to a java {@link Duration}.
-	 * The string is like "123ms", "321s", "12min" and such.
+	 * The string is in format "{length value}{time unit label}", e.g. "123ms", "321 s".
+	 * If no time unit label is specified, it will be considered as milliseconds.
+	 *
+	 * <p>Supported time unit labels are:
```
TisonKun: Is there a source/standard support this format? If so, could you refer it?

flink-core/src/main/java/org/apache/flink/util/TimeUtils.java
```diff
@@ -29,9 +33,23 @@
  */
 public class TimeUtils {
 
+	private static final Map<String, ChronoUnit> LABEL_TO_UNIT_MAP = Collections.unmodifiableMap(initMap());
+
 	/**
 	 * Parse the given string to a java {@link Duration}.
-	 * The string is like "123ms", "321s", "12min" and such.
+	 * The string is in format "{length value}{time unit label}", e.g. "123ms", "321 s".
+	 * If no time unit label is specified, it will be considered as milliseconds.
```
TisonKun: Where is this default comes from?

How about fail with a message that time unit must be configured?

flink-core/src/main/java/org/apache/flink/util/TimeUtils.java
zhuzhurk: The supported labels are acquired from the source code of [scala.concurrent.duration.Duration](https://github.com/scala/scala/blob/f7b5b312ca5dd1333245d8174a8dbadc89bd40fc/src/library/scala/concurrent/duration/Duration.scala#L76). 
And I think there is no need to refer it, as we want to be compatible with scala duration labels at the moment but not with scala duration's future changes.

The standard for users would only be the javadoc of flink TimeUtils then.

flink-core/src/main/java/org/apache/flink/util/TimeUtils.java
zhuzhurk: TimeUtils previously support non-labeled duration parsed as millisecond duration.
I think we should keep it to not break anything which is already using it.

TisonKun: Thanks for your explanation :-)

flink-core/src/main/java/org/apache/flink/util/TimeUtils.java
TisonKun: Make sense. "refer it" means we keep the context at least in this pull request so that the origin of the change can be sourced. I just notice that it is already mentioned in the title and description :P

flink-core/src/test/java/org/apache/flink/util/TimeUtilsTest.java
```diff
@@ -123,9 +154,4 @@ public void testParseDurationInvalid() {
 	public void testParseDurationNumberOverflow() {
 		TimeUtils.parseDuration("100000000000000000000000000000000 ms");
 	}
-
-	@Test(expected = IllegalArgumentException.class)
-	public void testParseDurationNumberTimeUnitOverflow() {
```
TisonKun: With your patch we still possibly throw an `IllegalArgumentException` if the number overflow. Why do you remove this test?

flink-core/src/main/java/org/apache/flink/util/TimeUtils.java
```diff
@@ -64,64 +82,71 @@ public static Duration parseDuration(String text) {
...
+		MILLISECONDS(ChronoUnit.MILLIS, "ms", "milli", "millisecond"),
+		MICROSECONDS(ChronoUnit.MICROS, "µs", "micro", "microsecond"),
+		NANOSECONDS(ChronoUnit.NANOS, "ns", "nano", "nanosecond");
+
+		private String[] labels;
+
+		private ChronoUnit unit;
+
+		TimeUnit(ChronoUnit unit, String... labels) {
+			this.unit = checkNotNull(unit);
```
TisonKun: nit: `TimeUnit` has the only private constructor, it is ok to reduce code complexity by directly assign without runtime checker.

flink-core/src/test/java/org/apache/flink/util/TimeUtilsTest.java
zhuzhurk: This test was for some removed logics in TimeUtils.

Currently this case can cause `ArithmeticException` thrown from `java.time.Duration`.
But I think we do not need it, just like we do not need to add tests for formats validation, which should be tested by `java.time.Duration` itself.

flink-core/src/main/java/org/apache/flink/util/TimeUtils.java
zhuzhurk: Agreed.

flink-core/src/main/java/org/apache/flink/util/TimeUtils.java
```diff
@@ -64,64 +82,68 @@ public static Duration parseDuration(String text) {
...
-				MINUTES.getUnit(),
-				HOURS.getUnit()
+			return String.join(", ", new String[]{
+				"DAYS: (" + String.join(" | ", DAYS.getLabels()) + ")",
+				"HOURS: (" + String.join(" | ", HOURS.getLabels()) + ")",
+				"MINUTES: (" + String.join(" | ", MINUTES.getLabels()) + ")",
+				"SECONDS: (" + String.join(" | ", SECONDS.getLabels()) + ")",
+				"MILLISECONDS: (" + String.join(" | ", MILLISECONDS.getLabels()) + ")",
+				"MICROSECONDS: (" + String.join(" | ", MICROSECONDS.getLabels()) + ")",
+				"NANOSECONDS: (" + String.join(" | ", NANOSECONDS.getLabels()) + ")"
```
tillrohrmann: Could be generated from `TimeUnit.values()`.

zhuzhurk: Good catch and thanks for improving it. 
The change looks good to me.

PR #9690, opened 2019-09-16
[FLINK-14067] Decouple PlanExecutor from JSON plan generation

aljoscha:

{{PlanExecutor}} has a method {{getOptimizerPlanAsJSON()}} that is used by DataSet environments to get a JSON version of the execution plan. To ease future work and to make it more maintainable we should get rid of that method and instead have a dedicated utility for generating JSON plans that the environments can use.

(The only reason this method is on the executor is because only {{flink-clients}} via {{flink-optimizer}} has the required components to derive a JSON plan.)

## Brief change log
Please look at the list of commits for a changelog, each commit is an isolated refactoring on the path to the final result.

## Verifying this change

* Added a test for the new `ExecutionPlanUtil`

## Documentation

### DIFF
--- a/flink-clients/src/main/java/org/apache/flink/client/program/ContextEnvironment.java
+++ b/flink-clients/src/main/java/org/apache/flink/client/program/ContextEnvironment.java
@@ -24,8 +24,6 @@ import org.apache.flink.api.common.JobExecutionResult;
 import org.apache.flink.api.common.JobSubmissionResult;
 import org.apache.flink.api.common.Plan;
 import org.apache.flink.api.java.ExecutionEnvironment;
-import org.apache.flink.optimizer.plan.OptimizedPlan;
-import org.apache.flink.optimizer.plandump.PlanJSONDumpGenerator;
 import org.apache.flink.runtime.jobgraph.SavepointRestoreSettings;
 
 import java.net.URL;
@@ -72,15 +70,6 @@ public class ContextEnvironment extends ExecutionEnvironment {
 
 		lastJobExecutionResult = jobSubmissionResult.getJobExecutionResult();
 		return lastJobExecutionResult;
-	}
-
-	@Override
-	public String getExecutionPlan() throws Exception {
-		Plan plan = createProgramPlan("unnamed job");
-
-		OptimizedPlan op = ClusterClient.getOptimizedPlan(client.compiler, plan, getParallelism());
-		PlanJSONDumpGenerator gen = new PlanJSONDumpGenerator();
-		return gen.getOptimizerPlanAsJSON(op);
 	}
 
 	private void verifyExecuteIsCalledOnceWhenInDetachedMode() {
--- a/flink-java/src/main/java/org/apache/flink/api/java/ExecutionEnvironment.java
+++ b/flink-java/src/main/java/org/apache/flink/api/java/ExecutionEnvironment.java
@@ -837,7 +837,7 @@ public abstract class ExecutionEnvironment {
 	 */
 	@Internal
 	public Plan createProgramPlan() {
-		return createProgramPlan(null);
+		return createProgramPlan(getDefaultName());
 	}
 
 	/**
--- a/flink-java/src/main/java/org/apache/flink/api/java/ExecutionEnvironment.java
+++ b/flink-java/src/main/java/org/apache/flink/api/java/ExecutionEnvironment.java
@@ -771,7 +771,7 @@ public abstract class ExecutionEnvironment {
 	 * @throws Exception Thrown, if the compiler could not be instantiated.
 	 */
 	public String getExecutionPlan() throws Exception {
-		Plan p = createProgramPlan(null, false);
+		Plan p = createProgramPlan(getDefaultName(), false);
 		return ExecutionPlanUtil.getExecutionPlanAsJSON(p);
 	}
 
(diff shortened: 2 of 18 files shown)


kl0u:
Thanks @aljoscha !  I will have a look

flink-java/src/main/java/org/apache/flink/api/java/ExecutionPlanUtil.java
```diff
@@ -0,0 +1,76 @@
...
+		} catch (Throwable t) {
+			throw new RuntimeException("An error occurred while loading the plan generator ("
+					+ PLAN_GENERATOR_CLASS_NAME + ").", t);
+		}
+	}
+
+	private static Class<? extends ExecutionPlanJSONGenerator> loadJSONGeneratorClass(String className) {
+		try {
+			Class<?> generatorClass = Class.forName(
+					"org.apache.flink.optimizer.plandump.ExecutionPlanJSONGenerator");
```
kl0u: You can use the `PLAN_GENERATOR_CLASS_NAME` instead of re-hardcoding the classname and remove the argument or use the argument.

flink-clients/src/main/java/org/apache/flink/client/program/ContextEnvironment.java
```diff
@@ -74,15 +72,6 @@ public JobExecutionResult execute(String jobName) throws Exception {
 		return lastJobExecutionResult;
 	}
 
-	@Override
-	public String getExecutionPlan() throws Exception {
-		Plan plan = createProgramPlan("unnamed job");
-
-		OptimizedPlan op = ClusterClient.getOptimizedPlan(client.compiler, plan, getParallelism());
```
kl0u: I think that now we can make the `client.compiler` field `private` to avoid leaking information about internal state.

kl0u: And maybe other `ClusterClient` fields can become `private` as well.

flink-java/src/main/java/org/apache/flink/api/java/ExecutionEnvironment.java
```diff
@@ -766,13 +766,14 @@ public JobExecutionResult execute() throws Exception {
...
 	 * a String using a JSON representation of the execution data flow graph.
-	 * Note that this needs to be called, before the plan is executed.
 	 *
 	 * @return The execution plan of the program, as a JSON String.
-	 * @throws Exception Thrown, if the compiler could not be instantiated, or the master could not
-	 *                   be contacted to retrieve information relevant to the execution planning.
+	 * @throws Exception Thrown, if the compiler could not be instantiated.
 	 */
-	public abstract String getExecutionPlan() throws Exception;
+	public String getExecutionPlan() throws Exception {
```
kl0u: I think you can remove the `Exception`.

flink-java/src/main/java/org/apache/flink/api/java/ExecutionEnvironment.java
```diff
@@ -766,13 +766,14 @@ public JobExecutionResult execute() throws Exception {
...
+		Plan p = createProgramPlan(null, false);
+		return ExecutionPlanUtil.getExecutionPlanAsJSON(p);
```
kl0u: I would be up for removing passing `null` values by calling here `getDefaultName()` and `checkNotNull` in the `createProgramPlan`.

flink-java/src/main/java/org/apache/flink/api/java/ExecutionPlanUtil.java
```diff
@@ -0,0 +1,76 @@
...
+		}
+	}
+
+	/**
+	 * Internal interface for the JSON plan generator that has to reside in the optimizer package.
+	 * We load the actual subclass using reflection.
+	 */
+	@Internal
+	public interface ExecutionPlanJSONGenerator {
+		String getExecutionPlan(Plan plan);
```
kl0u: Add a javadoc to the method.

flink-java/src/main/java/org/apache/flink/api/java/ExecutionPlanUtil.java
```diff
@@ -0,0 +1,76 @@
...
+@Internal
+public class ExecutionPlanUtil {
+
+	private static final String PLAN_GENERATOR_CLASS_NAME = "org.apache.flink.optimizer.plandump.ExecutionPlanJSONGenerator";
+
+	/**
+	 * Extracts the execution plan (as JSON) from the given {@link Plan}.
+	 */
+	public static String getExecutionPlanAsJSON(Plan plan) {
+		ExecutionPlanJSONGenerator jsonGenerator = getJSONGenerator();
```
kl0u: I think this needs a check `checkNotNull(plan)`.

flink-java/src/main/java/org/apache/flink/api/java/ExecutionPlanUtil.java
aljoscha: fixing

flink-java/src/main/java/org/apache/flink/api/java/ExecutionEnvironment.java
aljoscha: I didn't want to remove it here because it is a `@Public` class.

flink-java/src/main/java/org/apache/flink/api/java/ExecutionEnvironment.java
aljoscha: fixing

flink-java/src/main/java/org/apache/flink/api/java/ExecutionPlanUtil.java
aljoscha: adding javadoc

aljoscha:
merged

PR #9691, opened 2019-09-16
[FLINK-13965] Keep hasDeprecatedKeys and deprecatedKeys methods in ConfigOption and mark it with @Deprecated annotation

yanghua:

*This pull request reversed two deleted method `hasDeprecatedKeys` and `deprecatedKeys`, then marked them with `@Deprecated`*

## Brief change log

  - *Reversed two deleted methods*
  - *Added test case for the two methods*

## Verifying this change

This change is already covered by existing tests, such as *testDeprecationForDeprecatedKeys*.

## Documentation

### DIFF
--- a/flink-core/src/main/java/org/apache/flink/configuration/ConfigOption.java
+++ b/flink-core/src/main/java/org/apache/flink/configuration/ConfigOption.java
@@ -23,6 +23,7 @@ import org.apache.flink.configuration.description.Description;
 
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import static org.apache.flink.util.Preconditions.checkNotNull;
@@ -193,6 +194,31 @@ public class ConfigOption<T> {
 	 */
 	public T defaultValue() {
 		return defaultValue;
+	}
+
+	/**
+	 * Checks whether this option has deprecated keys.
+	 * @return True if the option has deprecated keys, false if not.
+	 * @deprecated Replaced by {@link #hasFallbackKeys()}
+	 */
+	@Deprecated
+	public boolean hasDeprecatedKeys() {
+		return fallbackKeys == EMPTY ? false :
+			Arrays.stream(fallbackKeys).anyMatch(FallbackKey::isDeprecated);
+	}
+
+	/**
+	 * Gets the deprecated keys, in the order to be checked.
+	 * @return The option's deprecated keys.
+	 * @deprecated Replaced by {@link #fallbackKeys()}
+	 */
+	@Deprecated
+	public Iterable<String> deprecatedKeys() {
+		return fallbackKeys == EMPTY ? Collections.<String>emptyList() :
+			Arrays.stream(fallbackKeys)
+				.filter(FallbackKey::isDeprecated)
+				.map(FallbackKey::getKey)
+				.collect(Collectors.toList());
 	}
 
 	/**
--- a/flink-core/src/test/java/org/apache/flink/configuration/ConfigOptionTest.java
+++ b/flink-core/src/test/java/org/apache/flink/configuration/ConfigOptionTest.java
@@ -20,10 +20,15 @@ package org.apache.flink.configuration;
 
 import org.apache.flink.util.TestLogger;
 
+import org.apache.flink.shaded.guava18.com.google.common.collect.Sets;
+
 import org.junit.Test;
 
 import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 
 import static org.hamcrest.Matchers.containsInAnyOrder;
 import static org.junit.Assert.assertEquals;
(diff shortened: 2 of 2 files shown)


yanghua:
@flinkbot run travis

yanghua:
@tillrohrmann Can you review this PR? thanks.

flink-core/src/main/java/org/apache/flink/configuration/ConfigOption.java
```diff
@@ -195,6 +196,27 @@ public T defaultValue() {
...
 	}
 
+	/**
+	 * Checks whether this option has deprecated keys.
+	 * @return True if the option has deprecated keys, false if not.
+	 * @deprecated Replaced by {@link #hasFallbackKeys()}
+	 */
+	@Deprecated
+	public boolean hasDeprecatedKeys() {
+		return fallbackKeys != EMPTY;
```
TisonKun: What if we just delegate it to `hasFallbackKeys`? Concretely

```java
/**
  * @deprecated Replaced by {@link #hasFallbackKeys()}
  */
@Deprecated
public boolean hasDeprecatedKeys() {
  return hasFallbackKeys();
}
```

I think it would not cause compile error when you bump the version and it is better that we keep code deprecated like deprecated code.

yanghua: Agree reuse `hasFallbackKeys` and I would like to remain the java doc of `hasDeprecatedKeys `, because it is used to describe the method of `hasDeprecatedKeys `, not `FallbackKey`.

flink-core/src/main/java/org/apache/flink/configuration/ConfigOption.java
```diff
@@ -195,6 +196,27 @@ public T defaultValue() {
...
+
+	/**
+	 * Gets the deprecated keys, in the order to be checked.
+	 * @return The option's deprecated keys.
+	 * @deprecated Replaced by {@link #fallbackKeys()}
+	 */
+	@Deprecated
+	public Iterable<String> deprecatedKeys() {
+		return fallbackKeys == EMPTY ? Collections.<String>emptyList() :
+			Arrays.stream(fallbackKeys).map(fallbackKey -> fallbackKey.getKey()).collect(Collectors.toList());
```
tillrohrmann:
```suggestion
			Arrays.stream(fallbackKeys)
			    .filter(FallbackKey::isDeprecated)
			    .map(FallbackKey::getKey)
			    .collect(Collectors.toList());
```

flink-core/src/test/java/org/apache/flink/configuration/ConfigOptionTest.java
```diff
@@ -86,4 +87,19 @@ public void testDeprecationFlagForMixedAlternativeKeys() {
...
 		assertThat(deprecatedKeys, containsInAnyOrder("deprecated1", "deprecated2"));
 	}
+
+	@Test
+	public void testDeprecationForDeprecatedKeys() {
+		String[] deprecatedKeys = new String[] { "deprecated1", "deprecated2" };
+		final ConfigOption<Integer> optionWithDeprecatedKeys = ConfigOptions
+			.key("key")
+			.defaultValue(0)
+			.withDeprecatedKeys(deprecatedKeys);
```
tillrohrmann: Let's add a fallback key and test that it gets filtered out.

flink-core/src/main/java/org/apache/flink/configuration/ConfigOption.java
yanghua: I think you are right. Additionally, we may also need to change the implementation of the`hasDeprecatedKeys` method, right? Only based on the `fallbackKeys != EMPTY` condition, can not judge a config option has deprecated keys if the fallback key is not marked as deprecated.

flink-core/src/test/java/org/apache/flink/configuration/ConfigOptionTest.java
```diff
@@ -86,4 +87,32 @@ public void testDeprecationFlagForMixedAlternativeKeys() {
...
+			counter++;
+			assertTrue(Arrays.stream(deprecatedKeys).anyMatch(item -> item.equals(deprecatedKey)));
+		}
+
+		assertEquals(2, counter);
+
+		final ConfigOption<Integer> optionWithFallbackKeys = ConfigOptions
+			.key("key")
+			.defaultValue(0)
+			.withFallbackKeys("fallback1");
```
tillrohrmann: Let's split this test up into a separate one. It is usually better to have smaller and more targeted tests.

flink-core/src/test/java/org/apache/flink/configuration/ConfigOptionTest.java
```diff
@@ -86,4 +87,32 @@ public void testDeprecationFlagForMixedAlternativeKeys() {
...
+			.defaultValue(0)
+			.withDeprecatedKeys(deprecatedKeys)
+			.withFallbackKeys("fallback1");
+
+		assertTrue(optionWithDeprecatedKeys.hasDeprecatedKeys());
+		int counter = 0;
+		for (final String deprecatedKey : optionWithDeprecatedKeys.deprecatedKeys()) {
+			counter++;
+			assertTrue(Arrays.stream(deprecatedKeys).anyMatch(item -> item.equals(deprecatedKey)));
+		}
```
tillrohrmann: nit: Instead of manually going over the `deprecatedKeys` one could convert them and the input keys into a `Set` and use `assertEquals`.

flink-core/src/main/java/org/apache/flink/configuration/ConfigOption.java
```diff
@@ -195,6 +196,31 @@ public T defaultValue() {
...
 	}
 
+	/**
+	 * Checks whether this option has deprecated keys.
+	 * @return True if the option has deprecated keys, false if not.
+	 * @deprecated Replaced by {@link #hasFallbackKeys()}
+	 */
+	@Deprecated
+	public boolean hasDeprecatedKeys() {
+		return fallbackKeys == EMPTY ? true :
```
tillrohrmann: I think one needs to return `false` if `fallbackKeys` are `EMPTY`.

PR #9694, opened 2019-09-16
[hotfix] [docs] Fixes in glossar

afedulov:


docs/concepts/glossary.md
```diff
@@ -79,7 +79,7 @@ whole [Flink Master](#flink-master) was called JobManager.
 #### Logical Graph
 
 A logical graph is a directed graph describing the high-level logic of a stream processing program.
-The nodes are [Operators](#operator) and the edges indicate input/output-relationships or
```
StephanEwen: I think this sounded correct as it was.

docs/concepts/glossary.md
```diff
@@ -161,6 +161,6 @@ subsequent Tasks.
 A Transformation is applied on one or more data streams or data sets and results in one or more
 output data streams or data sets. A transformation might change a data stream or data set on a
 per-record basis, but might also only change its partitioning or perform an aggregation. While
-[Operators](#operator) and [Functions](#function)) are the "physical" parts of Flink's API,
-Transformations are only an API concept. Specifically, most - but not all - transformations are
```
StephanEwen: Was that wrong, or is this mainly a language improvement?

docs/concepts/glossary.md
afedulov: I am pretty sure it is a copy-paste typo from the Physical Graph where the wrong preposition got removed along with "partitions".
Physical graph:
"The nodes are Tasks and the edges indicate input/output-relationships **or** partitions **of** data streams or data sets." 

Also the construct "this OR that OR that" sounds awkward and people usually write "this, that OR that" when listing more than two items.

docs/concepts/glossary.md
afedulov: Language improvement. "most" already implies "not all", hence this is redundant and does not bring any additional information. Additionally I believe it should be separated with the commas, not hyphens from the punctuation perspective.

docs/concepts/glossary.md
fhueske: I agree with Stephan. "... input/output-relationships of data streams or data sets" does not seem right to me. The edges describe the relationships of *operators* but not of *data streams/sets*. 

You're probably correct that this part was copy/pasted. In the original sentence, the words belong together as follows: "(input/output-relationships) or (partitions of data streams or data sets)". Since, a logical graph is not parallelized, the "partitions of" was removed. 

Although correct, I agree that "input/output-relationships or data streams or data sets." doesn't sound good and could be improved. How about:

> ... the edges indicate input/output-relationships of the operators and correspond to data streams or data sets.

docs/concepts/glossary.md
fhueske: LGTM

docs/concepts/glossary.md
afedulov: @fhueske changed accordingly and rebased on the current master.

fhueske:
Thanks for the update @afedulov. Will merge this.

PR #9695, opened 2019-09-17
[FLINK-13748][S3][build] Fix jaxb relocation for S3.

kl0u:

Fixes shading on Presto and Hadoop S3 FileSystems.

## Brief change log

Changes the `pom.xml` in the `flink-s3-fs-hadoop` and `flink-s3-fs-presto`.

## Verifying this change

This change is already covered by existing tests, such as the s3 e2e tests and can be verified by running the cron jobs on Travis.

  - The S3 file system connector: (**yes** although its build process, not the logic / no / don't know)

## Documentation

### DIFF
--- a/flink-filesystems/flink-s3-fs-presto/pom.xml
+++ b/flink-filesystems/flink-s3-fs-presto/pom.xml
@@ -271,10 +271,6 @@ under the License.
 									<shadedPattern>org.apache.flink.fs.s3presto.shaded.io.airlift</shadedPattern>
 								</relocation>
 								<relocation>
-									<pattern>javax.xml.bind</pattern>
-									<shadedPattern>org.apache.flink.fs.s3presto.shaded.javax.xml.bind</shadedPattern>
-								</relocation>
-								<relocation>
 									<pattern>org.HdrHistogram</pattern>
 									<shadedPattern>org.apache.flink.fs.s3presto.shaded.org.HdrHistogram</shadedPattern>
 								</relocation>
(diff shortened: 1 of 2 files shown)


kl0u:
To reviewers: This is the build that verified that the change works: https://travis-ci.org/kl0u/flink/builds/585693473. As you can see something goes wrong during the building of the `e2e-misc-jdk11` profile. 

I will relaunch but at least the tests seem to work for java 8 as shown in https://api.travis-ci.org/v3/job/585693500/log.txt and https://api.travis-ci.org/v3/job/585693501/log.txt

flink-filesystems/flink-s3-fs-presto/pom.xml
```diff
@@ -342,6 +338,130 @@ under the License.
...
+												<include>com/facebook/presto/hadoop/**</include>
+											</includes>
+										</filter>
+									</filters>
+								</configuration>
+							</execution>
+						</executions>
+					</plugin>
+				</plugins>
+			</build>
```
tillrohrmann: Same here. I think maintaining this shading logic can be quite a pain in the ass.

kl0u:
@tillrohrmann I agree but adding by default the dependency seems to be failing to compile as shown here: https://travis-ci.org/kl0u/flink/builds/586431721 (this is the branch this travis run refers to https://github.com/kl0u/flink/tree/fix-jaxb-alt)

kl0u:
It may be just a versioning issue. Java 8 needs `jaxb` `2.2.11`.

tillrohrmann:
@kl0u, would it work to add `jaxb:2.2.11` as the dependency or does Java 11 require `jaxb` `2.3.x`?

kl0u:
@tillrohrmann I am running it on Travis to see

StephanEwen:
We were planning to completely remove shading of the file systems for Flink 1.10 and put them into the plugins directory.

Would this be the time to do that? It seems a bit like wasted effort to fix shading when we want to remove it soon anyways.

kl0u:
@StephanEwen I agree that if this is the goal, then this effort is "instant legacy".

That said, I do not feel comfortable doing the whole migration to the plugin mechanism for the filesystems myself. This is both due to time constraints, as I am already involved in other features, and because of non-familiarity with our plugin mechanism.

I can close this PR if you think that we should not pursue this solution any further.

tillrohrmann:
If adding the `jaxb` dependency works, then I'd be in favour of doing it because it solves an immediate problem which are failing cron jobs. Moreover, this kind of fix should be a 5 liner. What is the status here @kl0u?

Once we get to removing shading, this fix then becomes obsolete and can be removed.

kl0u:
The status can be seen here https://travis-ci.org/kl0u/flink/builds/586896522 and the branch with the changes where we only use `jaxb 2.2.11` is here https://github.com/kl0u/flink/tree/fix-jaxb-alt-3 if you want to see what we are testing.

PR #9697, opened 2019-09-17
[FLINK-14094] [runtime] [metric] Fix OperatorIOMetricGroup repeat register problem

maqingxiang:

There will be OperatorIOMetricGroup duplicate registration in the TaskMetricGroup's getOrAddOperator() method.

### DIFF
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/metrics/groups/TaskMetricGroup.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/metrics/groups/TaskMetricGroup.java
@@ -139,24 +139,19 @@ public class TaskMetricGroup extends ComponentMetricGroup<TaskManagerJobMetricGr
 	}
 
 	public OperatorMetricGroup getOrAddOperator(OperatorID operatorID, String name) {
+		final String metricName;
 		if (name != null && name.length() > METRICS_OPERATOR_NAME_MAX_LENGTH) {
 			LOG.warn("The operator name {} exceeded the {} characters length limit and was truncated.", name, METRICS_OPERATOR_NAME_MAX_LENGTH);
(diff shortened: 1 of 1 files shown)


flink-runtime/src/main/java/org/apache/flink/runtime/metrics/groups/TaskMetricGroup.java
```diff
@@ -143,20 +143,17 @@ public OperatorMetricGroup getOrAddOperator(OperatorID operatorID, String name)
...
-			if (previous == null) {
-				// no operator group so far
-				return operator;
-			} else {
-				// already had an operator group. restore that one.
-				operators.put(key, previous);
-				return previous;
-			}
+			operators.put(key, operator);
+			return operator;
```
tillrohrmann: I'd suggest to do it the following way:

```
public OperatorMetricGroup getOrAddOperator(OperatorID operatorID, String name) {
		final String metricName;
		if (name != null && name.length() > METRICS_OPERATOR_NAME_MAX_LENGTH) {
			LOG.warn("The operator name {} exceeded the {} characters length limit and was truncated.", name, METRICS_OPERATOR_NAME_MAX_LENGTH);
			metricName = name.substring(0, METRICS_OPERATOR_NAME_MAX_LENGTH);
		} else {
			metricName = name;
		}
		// unique OperatorIDs only exist in streaming, so we have to rely on the name for batch operators
		final String key = operatorID + metricName;

		synchronized (this) {
			return operators.computeIfAbsent(key, ignored -> new OperatorMetricGroup(this.registry, this, operatorID, metricName));
		}
	}
```

maqingxiang: Thanks for your review @tillrohrmann.  I had improved it.

PR #9702, opened 2019-09-17
[FLINK-14107][kinesis] Erroneous queue selection in record emitter may lead to deadlock

tweise:

When the emitter reaches the max timestamp for the current queue, it stops emitting and waits for the max timestamp to advance. Since it simultaneously selects the next queue as new "minimum" queue, it may subsequently deadlock if the previous min queue represents the new global lower bound after the max timestamp advanced. This occurs very infrequently and we were finally able to reproduce.

## Verifying this change

This change was verified on our internal system and reproduced with the unit test.


tweise:
@haltwise PTAL. I added a unit test on top of the changes we tested yesterday.

jgrier:
@flinkbot approve all

PR #9708, opened 2019-09-18
[FLINK-14107][kinesis] Erroneous queue selection in record emitter may lead to deadlock

tweise:
backport from master


tweise:
unrelated test failure

PR #9711, opened 2019-09-19
[FLINK-14033][yarn] upload user artifacts for yarn job cluster

HuangZhenQiu:

Upload user artifacts that registered in distributed caches for yarn job clusters. Currently, distributed caches are not handled correctly. The solution proposed is to upload local files onto the remote file system and update the registered file in cache entry. 

## Brief change-log
  - Modified the AbstractClusterDescriptor to handle with artifacts for per job mode before starting 
     application master.

## Verifying this change

The change is end to end testing in YarnDistributedCacheITCase

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/jobgraph/JobGraph.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/jobgraph/JobGraph.java
@@ -573,6 +573,15 @@ public class JobGraph implements Serializable {
 		));
 	}
 
+	public void setUserArtifactRemotePath(String entryName, String remotePath) {
+		userArtifacts.computeIfPresent(entryName, (key, originalEntry) -> new DistributedCache.DistributedCacheEntry(
+			remotePath,
+			originalEntry.isExecutable,
+			null,
+			originalEntry.isZipped
+		));
+	}
+
 	public void writeUserArtifactEntriesToConfiguration() {
 		for (Map.Entry<String, DistributedCache.DistributedCacheEntry> userArtifact : userArtifacts.entrySet()) {
 			DistributedCache.writeFileInfoToConfig(
--- a/flink-yarn-tests/src/test/java/org/apache/flink/yarn/YARNITCase.java
+++ b/flink-yarn-tests/src/test/java/org/apache/flink/yarn/YARNITCase.java
@@ -18,7 +18,6 @@
 
 package org.apache.flink.yarn;
 
-import org.apache.flink.api.common.time.Deadline;
 import org.apache.flink.client.deployment.ClusterSpecification;
 import org.apache.flink.client.program.ClusterClient;
 import org.apache.flink.configuration.AkkaOptions;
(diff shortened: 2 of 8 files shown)


flink-yarn/src/main/java/org/apache/flink/yarn/AbstractYarnClusterDescriptor.java
```diff
@@ -738,6 +738,25 @@ public ApplicationReport startAppMaster(
 			// add user code jars from the provided JobGraph
 			: jobGraph.getUserJars().stream().map(f -> f.toUri()).map(File::new).collect(Collectors.toSet());
 
+		// only for per job mode
+		if (jobGraph != null) {
+			final Collection<Tuple2<String, org.apache.flink.core.fs.Path>> userArtifacts = jobGraph.getUserArtifacts().entrySet().stream()
+				.map(entry -> Tuple2.of(entry.getKey(), new org.apache.flink.core.fs.Path(entry.getValue().filePath)))
+				.collect(Collectors.toList());
+
+			for (Tuple2<String, org.apache.flink.core.fs.Path> userArtifact : userArtifacts) {
```
wangyang0918: Can we use `for (Map.Entry<String, DistributedCache.DistributedCacheEntry> entry : jobGraph.getUserArtifacts().entrySet())` for the loop here? We may not need to create the userArtifacts list.

HuangZhenQiu: @wangyang0918 
I updated the loop and added an integration test in yarn test module. Would you please take one more round of review?

HuangZhenQiu:
@wangyang0918 @tillrohrmann 
Would you please help to review the PR. One of our customers wants to use this feature for his job.

HuangZhenQiu:
@wangyang0918 @suez1224 @tillrohrmann I just rebased lastest master, would you please take one more round of review?

HuangZhenQiu:
@aljoscha Given @tillrohrmann Is on vacation, would you please help to review this PR?

aljoscha:
Thanks for this contribution, I think it is a valuable feature! Could you please rebase this on top of the latest master and clean up the commits? I noticed some dangling/unused changes, for example you add a `CacheBasedDataSource` that is not used, and in `YarnTestCacheJob` the mapper that does the actual testing is not used, meaning the test does not test what it is supposed to test.

Did you think about how the uploaded files are being cleaned up? What is the lifecycle? Will they stay in the DFS after the application finishes?

HuangZhenQiu:
@aljoscha Thanks for pointing out issues in test case. I rebased latest master and removed unused classes. The distributed cache functionality should be covered in the map function.  For the life cycle of the distributed cache files, we should split into two parts. For the local copies in each TM, they are managed as a rolling array in the FileCache class. Any time one acces to a file not in FileCache, the file will be read from remote. For the remote file, it is the same as other local resources files for yarn application, they are under the ./flink/application_id fodler. If user kill the job, the folder will be deleted.

HuangZhenQiu:
@aljoscha During debugging, I found another bug introduced by a recent change. https://issues.apache.org/jira/browse/FLINK-14982. I would like to fix it, please assign to me after confirming it is a real issue.

aljoscha:
Thanks! This looks good now. I'm currently building some things and debugging, but once I'm done with that I'll review this again and then merge. 👌

sunjincheng121:
Just a reminder: This PR is the blocker of FLINK-14019 for YARN pre job mode. :)

HuangZhenQiu:
@sunjincheng121 Would you please also help to review the PR.

aljoscha:
Thanks a lot for this work! I merged it. 🍵

PR #9715, opened 2019-09-19
[FLINK-13993][runtime] Using FlinkUserCodeClassLoaders to load the user class in the perjob mode

guoweiM:

Currently, Flink has the FlinkUserCodeClassLoader, which is using to load user’s class. However, the user class and the system class are all loaded by the system classloader in the perjob mode. This introduces some conflicts.

This pull request makes the FlinkUserClassLoader load the user class in perjob mode. 

## Brief change log
- *Introduce the `AbstractUserClassPathJobGraphRetriever`  to get the user classpath.*
- *`ClassPathJobGraphRetriever` sets the user classpath to JobGraph.*
- *`StandaloneJobClusterEntryPoint` use the "FLINK_HOME/job" as the default user classpath.*
- *`FlinkJobGraphRetirever` uses the "job" as the default user classpath.*
- *The system classpath does not include the user jars in yarn perjob mode if setting the yarn.per-job-cluster.include-user-jar to "DISABLE".*
- *`MesosJobClusterEntrypoint` uses the "job" directory as the default user classpath.*

## Verifying this change

 - *Add some tests for the classloader.*
 - *The Standalone perjob mode can be covered by test_docker_embedded end-to-end test.*
 - *Add two integration tests for Yarn perjob mode.*
 - *Manually verifed Mesos perjob mode(Mesos/ Marathon).*

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-container/src/main/java/org/apache/flink/container/entrypoint/StandaloneJobClusterEntryPoint.java
+++ b/flink-container/src/main/java/org/apache/flink/container/entrypoint/StandaloneJobClusterEntryPoint.java
@@ -77,7 +77,7 @@ public final class StandaloneJobClusterEntryPoint extends JobClusterEntrypoint {
 	protected DispatcherResourceManagerComponentFactory<?> createDispatcherResourceManagerComponentFactory(Configuration configuration) {
 		return new JobDispatcherResourceManagerComponentFactory(
 			StandaloneResourceManagerFactory.INSTANCE,
-			new ClassPathJobGraphRetriever(jobId, savepointRestoreSettings, programArguments, jobClassName));
+			new ClassPathJobGraphRetriever(jobId, savepointRestoreSettings, programArguments, jobClassName, null));
 	}
 
 	public static void main(String[] args) {
--- a/flink-container/src/main/java/org/apache/flink/container/entrypoint/StandaloneJobClusterEntryPoint.java
+++ b/flink-container/src/main/java/org/apache/flink/container/entrypoint/StandaloneJobClusterEntryPoint.java
@@ -36,9 +36,11 @@ import org.apache.flink.runtime.util.SignalHandler;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 
+import java.nio.file.Paths;
 import java.util.Optional;
 
 import static java.util.Objects.requireNonNull;
+import static org.apache.flink.configuration.ConfigConstants.ENV_FLINK_HOME_DIR;
 
 /**
  * {@link JobClusterEntrypoint} which is started with a job in a predefined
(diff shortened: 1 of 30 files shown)


flink-runtime/src/main/java/org/apache/flink/runtime/entrypoint/component/AbstractUserClassPathJobGraphRetriever.java
```diff
@@ -0,0 +1,109 @@
...
+import java.util.LinkedList;
+import java.util.List;
+
+/**
+ *  Abstract class for the JobGraphRetriever, which wants to get classpath user's code depends on.
+ */
+
+public abstract class AbstractUserClassPathJobGraphRetriever implements JobGraphRetriever {
+
+	private final Logger log = LoggerFactory.getLogger(getClass());
```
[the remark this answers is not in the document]
zhuzhurk: Or maybe we can make it a private static logger for itself.
And then we can make `scanJarsInJobClassDir` static.

flink-container/src/main/java/org/apache/flink/container/entrypoint/StandaloneJobClusterEntryPoint.java
```diff
@@ -77,7 +81,9 @@ private StandaloneJobClusterEntryPoint(
 	protected DispatcherResourceManagerComponentFactory<?> createDispatcherResourceManagerComponentFactory(Configuration configuration) {
 		return new JobDispatcherResourceManagerComponentFactory(
 			StandaloneResourceManagerFactory.INSTANCE,
-			new ClassPathJobGraphRetriever(jobId, savepointRestoreSettings, programArguments, jobClassName));
+			new ClassPathJobGraphRetriever(jobId, savepointRestoreSettings, programArguments, jobClassName,
+				Paths.get(System.getenv(ENV_FLINK_HOME_DIR) == null ? DEFAULT_FLINK_HOME : System.getenv(ENV_FLINK_HOME_DIR),
```
zhuzhurk: Hard coding "/opt/flink" looks a bit tricky. 
Is it possible to we set "FLINK_HOME" in all cases? So that we can throw exception here for unexpected cases.

flink-container/src/test/java/org/apache/flink/container/entrypoint/ClassPathJobGraphRetrieverTest.java
```diff
@@ -55,6 +66,62 @@
...
+	private Path userJarHasEntryClass;
+	private Path userEntryClassDependJarAtUserDirHasEntryClass;
+	private Path textAtUserDirHasEntryClass;
+
+	private Path userDirHasNotEntryClass;
+	private Path userJarHasNotEntryClass;
+	private Path userEntryClassDependJarAtUserDirHasNotEntryClass;
+	private Path textAtUserDirHasNotEntryClass;
+
+
```
zhuzhurk: These duplicated empty lines can be removed.

zhuzhurk:
@guoweiM could you split this PR to into a few smaller tasks?
This PR contains a variety of changes, including those for stand alone job, yarn and mesos.
It can be helpful to track the changes under separate tickets and the boundaries of changes can be more clear.

guoweiM:
Split the original Jira into 4 sub-task to simplify the code review.
So I close this pull request.

PR #9716, opened 2019-09-19
[FLINK-14116][docs] Translate changes on Getting Started Overview to …

rongzhg:
…Chinese

## Brief change log

*(for example:)*

## Verifying this change

*(or)*

*(or)*

*(example:)*

  - Dependencies (does it add or upgrade a dependency): (  no)
  - The serializers: (no)
  - The runtime per-record code paths (performance sensitive): (no )
  - The S3 file system connector: ( no )

## Documentation

### DIFF
--- a/docs/getting-started/index.zh.md
+++ b/docs/getting-started/index.zh.md
@@ -46,7 +46,7 @@ under the License.
 <!--
 * The [**DataStream API**]() code walkthrough shows how to implement a simple DataStream application and how to extend it to be stateful and use timers.
 -->
-* [**DataStream API 示例**](./tutorials/datastream_api.html) 展示了如何编写一个基本的 DataStream 应用程序。 DataStream API 是 Flink 的主要抽象，用于通过 Java 或 Scala 实现具有复杂时间语义的有状态数据流处理的应用程序。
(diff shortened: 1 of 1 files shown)


docs/getting-started/index.zh.md
```diff
@@ -46,7 +46,7 @@ under the License.
 <!--
 * The [**DataStream API**]() code walkthrough shows how to implement a simple DataStream application and how to extend it to be stateful and use timers.
 -->
-* [**DataStream API 示例**](./tutorials/datastream_api.html) 展示了如何编写一个基本的 DataStream 应用程序。 DataStream API 是 Flink 的主要抽象，用于通过 Java 或 Scala 实现具有复杂时间语义的有状态数据流处理的应用程序。
+* [**DataStream API 示例**](./walkthroughs/datastream_api.html) 展示了如何编写一个基本的 DataStream 应用程序，以及如何在 DataStream 程序中使用状态或者timer。 DataStream API 是 Flink 的主要抽象，用于通过 Java 或 Scala 实现具有复杂时间语义的有状态数据流处理的应用程序。
```
klion26: Why do you change the link for "DataStream API 示例", if the link needs to be updated indeed, do we need to update the English version also?

rongzhg: English version has been updated already on master branch.  So i changed the URL in Chinese version

klion26:
@rongzhg Do you still work on this PR, could you please rebase the master to resolve the conflict?

PR #9717, opened 2019-09-19
[FLINK-14044] [runtime] Reducing synchronization in AsyncWaitOperator

AHeise:

Currently, AsyncWaitOperator receives computed results from a user thread and puts it in StreamElementQueue. This queue is synchronized as the emitter thread pulls available elements in a blocking fashion. With the recent migration to mailbox model, these elements are then put into a letter to be outputted in the mailbox thread.

If we directly send the results from the user thread to the mailbox, we can get rid of any synchronization as everything happens in the user thread. That will slim down the code heavily.

## Brief change log

*(for example:)*
  - Put results from asyncUserFunction into mailbox thread.
  - Remove all synchronizations from StreamElementQueue
  - Remove emitter and emitter thread
  - Remove executor from AsyncWaitOperator

## Verifying this change

This change is already covered by existing tests, such as.
AsyncWaitOperatorTest
(Ordered/Unordered)StreamElementQueueTest
Many IT tests

  - The serializers: no
  - The runtime per-record code paths (performance sensitive): yes

## Documentation


flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/AsyncWaitOperator.java
```diff
@@ -209,41 +176,31 @@ else if (element.isLatencyMarker()) {
...
+		userFunction.asyncInvoke(element.getValue(), resultHandler);
 	}
 
 	@Override
 	public void processWatermark(Watermark mark) throws Exception {
-		WatermarkQueueEntry watermarkBufferEntry = new WatermarkQueueEntry(mark);
+		addToWorkQueue(mark);
 
-		addAsyncBufferEntry(watermarkBufferEntry);
+		outputCompletedElements();
```
1u0: Is there a reason to "pump" results here? Maybe add a code comment here?

AHeise: Added a comment to explain the rationale.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/AsyncWaitOperator.java
```diff
@@ -293,140 +246,120 @@ public void close() throws Exception {
...
-
-			/*
-			 * FLINK-5638: If we have the checkpoint lock we might have to free it for a while so
-			 * that the emitter thread can complete/react to the interrupt signal.
-			 */
-			if (Thread.holdsLock(checkpointingLock)) {
-				while (emitterThread.isAlive()) {
-					checkpointingLock.wait(100L);
+	private void outputCompletedElements() {
+		Collection<StreamElement> completed = queue.popCompleted();
```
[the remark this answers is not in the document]
AHeise: I'd like to avoid pulling the elements one-by-one. But your comment gave me the idea to simply pass the collector into the queue (which passes it to the entries). This will avoid creating and copying any intermediate results.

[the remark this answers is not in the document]
1u0: What I've proposed is in fact would be very similar to the old code. The loop (traversal) overhead should be the same as it's now (with small "optimisation" that you would not check `.isWatermark()` for every output element but for every input element).
With this approach, I think it's possible to preserve more of old behavior and introduce back the `TimestampedCollector` here.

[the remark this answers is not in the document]
AHeise: Not sure what the benefit of the old code was, but I have improved the code even further. There is no need for the check anymore. So please check it one more time.

`TimestampedCollector` seems to be too limited at the moment (should actually delegate all `Output` methods). I could extend it in that way and use it (basically it's missing `emitWatermark` for our use case).

AHeise:
@flinkbot run travis

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/AsyncWaitOperator.java
```diff
@@ -209,41 +176,34 @@ else if (element.isLatencyMarker()) {
 	}
 
 	@Override
-	public void processElement(StreamRecord<IN> element) throws Exception {
-		final StreamRecordQueueEntry<OUT> streamRecordBufferEntry = new StreamRecordQueueEntry<>(element);
+	public void processElement(final StreamRecord<IN> element) throws Exception {
```
zhijiangW: Nit: this change might not be suggested mixing with the core changes. Unless this line was also touched by the core change, then we could add the `final` in passing. From the consistent view,  in the following `processWatermark(Watermark mark)` we also have no final argument.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/AsyncWaitOperator.java
```diff
@@ -209,41 +176,34 @@ else if (element.isLatencyMarker()) {
...
-		}
+			final long timeoutTimestamp = timeout + getProcessingTimeService().getCurrentProcessingTime();
 
-		addAsyncBufferEntry(streamRecordBufferEntry);
+			final ScheduledFuture<?> timeoutTimer = getProcessingTimeService().registerTimer(
+					timeoutTimestamp,
+					timestamp -> userFunction.timeout(element.getValue(), resultHandler));
 
-		userFunction.asyncInvoke(element.getValue(), streamRecordBufferEntry);
+			resultHandler.setTimeoutTimer(timeoutTimer);
```
zhijiangW: I have potential worries of setting timer delay here. I mean if above `registerTimer` happens before this setting, and during `ResultHandler#complete` if `mailboxExecutor.execute` also happens immediately, then the un-set timers would not be canceled any more.  

I know it would not happen atm because `mailboxExecutor.execute` just inserts into queue and would not execute right now. Maybe my worry could be ignored.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/AsyncWaitOperator.java
zhijiangW: Sorry for touching this issue again. I regard it as a separate improvement out of this pr scope, but I could also accept it if you wish. I am not quite sure whether it is worth verifying the effect of this improvement. Maybe not very necessary. :)

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/AsyncWaitOperator.java
```diff
@@ -276,10 +232,10 @@ public void snapshotState(StateSnapshotContext context) throws Exception {
 	@Override
 	public void initializeState(StateInitializationContext context) throws Exception {
 		super.initializeState(context);
+
```
zhijiangW: nit: unnecessary change. If we want to reformat the codes it is better for a separate hotfix commit.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/AsyncWaitOperator.java
```diff
@@ -276,10 +232,10 @@ public void snapshotState(StateSnapshotContext context) throws Exception {
...
 		recoveredStreamElements = context
 			.getOperatorStateStore()
 			.getListState(new ListStateDescriptor<>(STATE_NAME, inStreamElementSerializer));
-
```
zhijiangW: ditto

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/AsyncWaitOperator.java
```diff
@@ -293,140 +249,113 @@ public void close() throws Exception {
...
-		} catch (InterruptedException interrupted) {
-			exception = interrupted;
+		pendingStreamElement = streamElement;
 
-			Thread.currentThread().interrupt();
-		} catch (Exception e) {
-			exception = e;
+		Optional<ResultFuture<OUT>> queueEntry;
+		while (!(queueEntry = queue.tryPut(streamElement)).isPresent()) {
+			mailboxExecutor.yield();
```
zhijiangW: How to guarantee that `yield()` could consume the elements from `queue` finally?

In previous way the `EmitterThread` could consume the queue, but now I did not see any runnable to be submitted into mailbox for consuming the queue. I know `outputCompletedElements()` could empty the queue, but it can not happen before calling `userFunction.asyncInvoke(element.getValue(), resultHandler)`?

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/AsyncWaitOperator.java
```diff
@@ -293,140 +249,113 @@ public void close() throws Exception {
...
-	}
 
-	@Override
-	public void failOperator(Throwable throwable) {
-		getContainingTask().getEnvironment().failExternally(throwable);
+		@Override
+		public void completeExceptionally(Throwable error) {
+			// signal failure through task
+			getContainingTask().getEnvironment().failExternally(new Exception(
+						"Could not complete the stream element: " + inputRecord + '.',
```
zhijiangW: decrease indentation?

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/queue/StreamElementQueue.java
```diff
@@ -19,68 +19,54 @@
...
-	<T> void put(StreamElementQueueEntry<T> streamElementQueueEntry) throws InterruptedException;
-
-	/**
-	 * Try to put the given element in the queue. This operation succeeds if the queue has capacity
-	 * left and fails if the queue is full.
-	 *
-	 * @param streamElementQueueEntry to be inserted
-	 * @param <T> Type of the entries future value
+	 * @param streamElement to be inserted
 	 * @return True if the entry could be inserted; otherwise false
```
zhijiangW: We need to adjust the return javadoc because it is not boolean value now.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/AsyncWaitOperator.java
AHeise: `outputCompletedElements` is called after an element has been completed by the asyncUserFunction. Only at that point, the queue entries have actually a changed state. There is no other code path that can actually make any queue entry completed. So consuming the queue at any other location will not yield a different result.

The only other way an item can be actually be completed is if we insert a watermark into an empty queue, as watermarks are always completed.

AHeise:
> Thanks for this improvement @AHeise
> 
> Sorry for entering this PR a bit delay and actually I only finished the partial review now. I would continue on it later.

Incorporated your feedback and we have two open discussions (mostly me trying to convince you that it works as expected).

AHeise:
After discussion with @1u0 , I changed how elements are emitted.

Previously: whenever n elements are completed in the correct order, `outputCompletedElements` emits all of them. That may have blocked the mailbox thread unnecessarily if the output is full. A checkpoint event would have been only processed once all elements are handled by the `Collector`.

Now: We emit only one element in `outputCompletedElements` to keep latency low. Whenever, there are more elements available, directly enqueue a new mail to the mailbox. Checkpointing mails can now interleave the emission of records.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/queue/OrderedStreamElementQueue.java
```diff
@@ -19,117 +19,66 @@
...
-		try {
-			StreamElementQueueEntry<?>[] array = new StreamElementQueueEntry[queue.size()];
-
-			array = queue.toArray(array);
-
-			return Arrays.asList(array);
-		} finally {
-			lock.unlock();
+	public List<StreamElement> values() {
+		List<StreamElement> list = new ArrayList<>();
```
zhijiangW: new ArrayList<>(queue.size())?

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/queue/OrderedStreamElementQueue.java
```diff
@@ -19,117 +19,66 @@
...
+		for (StreamElementQueueEntry e : this.queue) {
```
zhijiangW: I guess `this` keyword is only suggested in the constructor scope, not quite sure whether it has the specific rule.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/queue/OrderedStreamElementQueue.java
```diff
@@ -143,88 +92,31 @@ public int size() {
...
-	public <T> void put(StreamElementQueueEntry<T> streamElementQueueEntry) throws InterruptedException {
-		lock.lockInterruptibly();
-
-		try {
-			while (queue.size() >= capacity) {
-				notFull.await();
-			}
+	public Optional<ResultFuture<OUT>> tryPut(StreamElement streamElement) {
+		if (queue.size() < capacity) {
+			StreamElementQueueEntry queueEntry = createEntry(streamElement);
```
zhijiangW: `StreamElementQueueEntry<OUT> queueEntry` to avoid unchecked.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/queue/OrderedStreamElementQueue.java
```diff
@@ -143,88 +92,31 @@ public int size() {
...
-		lock.lockInterruptibly();
-
-		try {
-			if (!queue.isEmpty() && queue.peek().isDone()) {
-				LOG.debug("Signal ordered stream element queue has completed head element.");
-				headIsCompleted.signalAll();
-			}
-		} finally {
-			lock.unlock();
+	private StreamElementQueueEntry createEntry(StreamElement streamElement) {
```
zhijiangW: ditto : StreamElementQueueEntry<OUT>

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/queue/OrderedStreamElementQueue.java
```diff
@@ -143,88 +92,31 @@ public int size() {
...
+		if (streamElement.isRecord()) {
+			return new StreamRecordQueueEntry<>((StreamRecord<?>) streamElement);
+		}
+		if (streamElement.isWatermark()) {
+			return new WatermarkQueueEntry((Watermark) streamElement);
```
zhijiangW: ditto:  new WatermarkQueueEntry<>

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/queue/StreamElementQueue.java
```diff
@@ -19,68 +19,56 @@
...
+ * Interface for a non-blocking stream element queues for the {@link AsyncWaitOperator}.
  */
 @Internal
-public interface StreamElementQueue {
+public interface StreamElementQueue<OUT> {
 
 	/**
-	 * Put the given element in the queue if capacity is left. If not, then block until this is
-	 * the case.
+	 * Try to put the given element in the queue. This operation succeeds if the queue has capacity left and fails if
```
zhijiangW: Try -> Tries

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/queue/StreamElementQueue.java
```diff
@@ -19,68 +19,56 @@
...
-	 * @param <T> Type of the entries future value
-	 * @throws InterruptedException if the calling thread has been interrupted while waiting to
-	 * 	insert the given element
-	 */
-	<T> void put(StreamElementQueueEntry<T> streamElementQueueEntry) throws InterruptedException;
-
-	/**
-	 * Try to put the given element in the queue. This operation succeeds if the queue has capacity
-	 * left and fails if the queue is full.
+	 * <p>This method returns a handle to the inserted element that allows to set the result of the computation.</p>
```
zhijiangW: I guess we do not need the tail </p> before in code formatting rule. Anything changed now?

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/queue/StreamElementQueue.java
```diff
@@ -19,68 +19,56 @@
...
-	 * Peek at the head of the queue and return the first completed {@link AsyncResult}. This
-	 * operation is a blocking operation and only returns once a completed async result has been
-	 * found.
+	 * Emits one completed element from the head of this queue into the given output.
 	 *
-	 * @return Completed {@link AsyncResult}
-	 * @throws InterruptedException if the current thread has been interrupted while waiting for a
-	 * 	completed async result.
+	 * <p>Will not emit any element if no element has been completed (check {@link #hasCompleted()} before entering
+	 * any critical section).</p>
```
zhijiangW: ditto: </p>

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/queue/StreamElementQueue.java
```diff
@@ -19,68 +19,56 @@
...
+	 *
+	 * @param output the output into which to emit
 	 */
-	AsyncResult peekBlockingly() throws InterruptedException;
+	void emitCompleted(TimestampedCollector<OUT> output);
 
 	/**
-	 * Poll the first completed {@link AsyncResult} from the head of this queue. This operation is
-	 * blocking and only returns once a completed async result has been found.
+	 * Checks if there is at least one completed element to be emitted.
```
zhijiangW: I think it is better to remove "to be emitted". `hasCompleted()` should not know/care of the purpose of upper caller whether to emit or not. In other words, if `hasCompleted()` is defined only used for  `emitCompleted`, we could judge it implicitly inside `emitCompleted`, no need to define an explicit interface method.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/queue/StreamElementQueue.java
```diff
@@ -19,68 +19,56 @@
...
 	 *
-	 * @return Completed {@link AsyncResult} which has been removed from the queue
-	 * @throws InterruptedException if the current thread has been interrupted while waiting for a
-	 * 	completed async result.
+	 * @return true if there is an element to be emitted.
```
zhijiangW: `true if there is a completed element.`

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/queue/StreamElementQueue.java
```diff
@@ -19,68 +19,56 @@
...
 	 */
-	AsyncResult poll() throws InterruptedException;
+	boolean hasCompleted();
 
 	/**
-	 * Return the collection of {@link StreamElementQueueEntry} currently contained in this queue.
+	 * Returns the collection of {@link StreamElement} currently contained in this queue for checkpointing.
+	 *
+	 * <p>This includes all non-emitted, completed and non-completed elements.</p>
```
zhijiangW: ditto: </p>

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/queue/StreamElementQueueEntry.java
```diff
@@ -19,79 +19,47 @@
...
+	boolean isDone();
 
 	/**
-	 * Register the given complete function to be called once this queue entry has been completed.
+	 * Emits the results associated with this queue entry.
 	 *
-	 * @param completeFunction to call when the queue entry has been completed
-	 * @param executor to run the complete function
+	 * @param output the output into which to emit.
+	 * @throws IllegalStateException if this entry is not done yet.
```
zhijiangW: We can not see this exception explicitly, so might not need this javadoc

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/queue/StreamRecordQueueEntry.java
```diff
@@ -21,65 +21,55 @@
...
-	public StreamRecordQueueEntry(StreamRecord<?> streamRecord) {
-		super(streamRecord);
+	@Nonnull
+	private final StreamRecord<?> inputRecord;
 
-		hasTimestamp = streamRecord.hasTimestamp();
-		timestamp = streamRecord.getTimestamp();
-
-		resultFuture = new CompletableFuture<>();
+	StreamRecordQueueEntry(@Nonnull StreamRecord<?> inputRecord) {
```
zhijiangW: We could remove `@Nonnull` because the below `Preconditions.checkNotNull` has the same effect.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/queue/StreamRecordQueueEntry.java
```diff
@@ -21,65 +21,55 @@
...
-	implements AsyncCollectionResult<OUT>, ResultFuture<OUT> {
-
-	/** Timestamp information. */
-	private final boolean hasTimestamp;
-	private final long timestamp;
-
-	/** Future containing the collection result. */
-	private final CompletableFuture<Collection<OUT>> resultFuture;
+class StreamRecordQueueEntry<OUT> implements StreamElementQueueEntry<OUT> {
+	private Collection<OUT> completed;
```
zhijiangW: It is better to give a more descriptive naming for `completed` for better understanding in following relevant processes. `completedOutputs/completedResults`?

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/queue/StreamRecordQueueEntry.java
```diff
@@ -21,65 +21,55 @@
...
-	public Collection<OUT> get() throws Exception {
-		return resultFuture.get();
-	}
+	public void emitResult(TimestampedCollector<OUT> output) {
+		Preconditions.checkState(completed != null, "Not done yet");
 
-	@Override
-	protected CompletableFuture<Collection<OUT>> getFuture() {
-		return resultFuture;
+		output.setTimestamp(this.inputRecord);
```
zhijiangW: Remove `this`

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/queue/StreamRecordQueueEntry.java
```diff
@@ -21,65 +21,55 @@
...
 
 	@Override
 	public void complete(Collection<OUT> result) {
-		resultFuture.complete(result);
-	}
-
-	@Override
-	public void completeExceptionally(Throwable error) {
-		resultFuture.completeExceptionally(error);
+		this.completed = result;
```
zhijiangW: ditto: `this`. Do we need to `checkNotNull(result)`?

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/queue/UnorderedStreamElementQueue.java
```diff
@@ -19,192 +19,146 @@
...
-	/** Locks and conditions for the blocking queue. */
-	private final ReentrantLock lock;
-	private final Condition notFull;
-	private final Condition hasCompletedEntries;
-
-	public UnorderedStreamElementQueue(
-			int capacity,
-			Executor executor,
-			OperatorActions operatorActions) {
+	private int numberEntries;
```
zhijiangW: numberEntries -> numberOfEntries?

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/queue/UnorderedStreamElementQueue.java
```diff
@@ -218,88 +172,99 @@ public int size() {
...
+		}
+	}
 
-					Iterator<StreamElementQueueEntry<?>> it = firstSet.iterator();
+	/**
+	 * A stage is a collection of queue entries that can be completed in arbitrary order.
+	 */
+	static class Stage<OUT> {
+		/** Undrained finished elements. */
+		private final Queue<StreamElementQueueEntry<OUT>> completedQueue;
```
zhijiangW: completedQueue -> completedElements

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/queue/UnorderedStreamElementQueue.java
```diff
@@ -218,88 +172,99 @@ public int size() {
...
+			super(inputRecord);
+			this.stage = stage;
+		}
 
-				while (firstSet.isEmpty() && firstSet != lastSet) {
-					firstSet = uncompletedQueue.poll();
+		@Override
+		public void complete(Collection<OUT> result) {
+			super.complete(result);
+			this.stage.completed(this);
```
zhijiangW: remove `this`

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/queue/UnorderedStreamElementQueue.java
```diff
@@ -218,88 +172,99 @@ public int size() {
...
 	 */
-	public void onCompleteHandler(StreamElementQueueEntry<?> streamElementQueueEntry) throws InterruptedException {
-		lock.lockInterruptibly();
+	static class StagedStreamRecordQueueEntry<OUT> extends StreamRecordQueueEntry<OUT> {
+		private final Stage stage;
 
-		try {
-			if (firstSet.remove(streamElementQueueEntry)) {
-				completedQueue.offer(streamElementQueueEntry);
+		public StagedStreamRecordQueueEntry(StreamRecord<?> inputRecord, Stage stage) {
```
zhijiangW: package private

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/queue/UnorderedStreamElementQueue.java
```diff
@@ -218,88 +172,99 @@ public int size() {
...
-	 *
-	 * @param streamElementQueueEntry which has been completed
-	 * @throws InterruptedException if the current thread has been interrupted while performing the
-	 * 	on complete callback.
+	 * An entry that notifies the respective stage upon completion.
 	 */
-	public void onCompleteHandler(StreamElementQueueEntry<?> streamElementQueueEntry) throws InterruptedException {
-		lock.lockInterruptibly();
+	static class StagedStreamRecordQueueEntry<OUT> extends StreamRecordQueueEntry<OUT> {
+		private final Stage stage;
```
zhijiangW: `Stage<OUT>`

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/queue/UnorderedStreamElementQueue.java
```diff
@@ -218,88 +172,99 @@ public int size() {
...
 
-				LOG.debug("Signal unordered stream element queue has completed entries.");
-				hasCompletedEntries.signalAll();
+		/**
+		 * Signal that an entry finished computation.
+		 */
+		void completed(StreamElementQueueEntry<OUT> elementQueueEntry) {
+			// adding only to completed queue if not completed before
+			// there may be a real result coming after the actual result, which is updated in the queue entry but
+			// the entry is not readded to the complete queue
```
zhijiangW: typo: readded

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/queue/UnorderedStreamElementQueue.java
```diff
@@ -218,88 +172,99 @@ public int size() {
...
-				}
-			},
-			executor);
-
-		numberEntries++;
+		/**
+		 * Adds the given entry to this stage. If the element is completed (watermark), it is directly moved into the
+		 * completed queue.
+		 */
+		void add(StreamElementQueueEntry queueEntry) {
```
zhijiangW: `StreamElementQueueEntry<OUT>`

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/queue/UnorderedStreamElementQueue.java
```diff
@@ -19,192 +19,146 @@
...
+	}
 
-			LOG.debug("Polled element from unordered stream element queue. New filling degree " +
-				"({}/{}).", numberEntries, capacity);
+	@Override
+	public void emitCompleted(TimestampedCollector<OUT> output) {
+		if (this.stages.isEmpty()) {
+			return;
+		}
+		final Stage currentStage = this.stages.getFirst();
```
zhijiangW: ditto: `this`

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/queue/UnorderedStreamElementQueue.java
```diff
@@ -19,192 +19,146 @@
...
+		numberEntries -= currentStage.emitCompleted(output);
 
-			return completedQueue.poll();
-		} finally {
-			lock.unlock();
+		// remove any stage if there are further stages, if not leave it as an optimization even if empty
+		if (this.stages.size() > 1 && currentStage.isEmpty()) {
+			this.stages.pop();
```
zhijiangW: ditto: `this`

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/queue/UnorderedStreamElementQueue.java
```diff
@@ -19,192 +19,146 @@
...
 
 	private static final Logger LOG = LoggerFactory.getLogger(UnorderedStreamElementQueue.class);
 
 	/** Capacity of this queue. */
 	private final int capacity;
 
-	/** Executor to run the onComplete callbacks. */
-	private final Executor executor;
+	/** Queue of queue entries segmented by watermarks. */
+	private final Deque<Stage> stages;
```
zhijiangW: Deque<Stage<OUT>>

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/queue/UnorderedStreamElementQueue.java
```diff
@@ -19,192 +19,146 @@
...
+			return Optional.empty();
 		}
 	}
 
-	@Override
-	public AsyncResult peekBlockingly() throws InterruptedException {
-		lock.lockInterruptibly();
+	private StreamElementQueueEntry<OUT> addRecord(StreamRecord<?> record) {
+		// ensure that there is at least one stage
+		Stage lastStage;
```
zhijiangW: `Stage<OUT>`

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/queue/UnorderedStreamElementQueue.java
```diff
@@ -19,192 +19,146 @@
...
-			}
+		// entry is bound to stage to notify it easily upon completion
+		StreamElementQueueEntry<OUT> queueEntry = new StagedStreamRecordQueueEntry<>(record, lastStage);
+		lastStage.add(queueEntry);
+		return queueEntry;
+	}
 
-			LOG.debug("Peeked head element from unordered stream element queue with filling degree " +
-				"({}/{}).", numberEntries, capacity);
+	private Stage addStage(int capacity) {
```
zhijiangW: ditto: `<OUT>`

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/queue/UnorderedStreamElementQueue.java
```diff
@@ -19,192 +19,146 @@
...
-import java.util.concurrent.locks.ReentrantLock;
 
 /**
- * Unordered implementation of the {@link StreamElementQueue}. The unordered stream element queue
- * emits asynchronous results as soon as they are completed. Additionally it maintains the
- * watermark-stream record order. This means that no stream record can be overtaken by a watermark
- * and no watermark can overtake a stream record. However, stream records falling in the same
- * segment between two watermarks can overtake each other (their emission order is not guaranteed).
+ * Unordered implementation of the {@link StreamElementQueue}. The unordered stream element queue provides
+ * asynchronous results as soon as they are completed. Additionally it maintains the watermark-stream record order.
```
zhijiangW: The previous descriptions for order/unorder issues are not very clearly in my thought.  Especially for `no stream record can be overtaken by a watermark and no watermark can overtake a stream record`, the sayings before/after `and` are the same.
I think it is better to involve in the stage concept for better understanding. If I understood correctly, elements in one stage are kept the order, and elements in different stages are not in order which means overtaking each other.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/queue/UnorderedStreamElementQueue.java
```diff
@@ -218,88 +172,99 @@ public int size() {
...
+	 * A stage is a collection of queue entries that can be completed in arbitrary order.
+	 */
+	static class Stage<OUT> {
+		/** Undrained finished elements. */
+		private final Queue<StreamElementQueueEntry<OUT>> completedQueue;
 
-					while (it.hasNext()) {
-						StreamElementQueueEntry<?> bufferEntry = it.next();
+		/** Finished and unfinished input elements. Used solely for checkpointing. */
+		private final Set<StreamElementQueueEntry<OUT>> pendingElements;
```
zhijiangW: I think it is better to rename `uncompletedElements` instead, because the semantic `pending` in below `addPendingElements` covers both completed/uncompleted elements.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/queue/UnorderedStreamElementQueue.java
```diff
@@ -218,88 +172,99 @@ public int size() {
...
+	/**
+	 * A stage is a collection of queue entries that can be completed in arbitrary order.
+	 */
+	static class Stage<OUT> {
+		/** Undrained finished elements. */
+		private final Queue<StreamElementQueueEntry<OUT>> completedQueue;
 
-					while (it.hasNext()) {
-						StreamElementQueueEntry<?> bufferEntry = it.next();
+		/** Finished and unfinished input elements. Used solely for checkpointing. */
```
zhijiangW: Remove `Finished` for the comments.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/queue/UnorderedStreamElementQueue.java
```diff
@@ -218,88 +172,99 @@ public int size() {
...
-				}
+		Stage(int initialCapacity) {
+			completedQueue = new ArrayDeque<>(initialCapacity);
+			pendingElements = new HashSet<>(initialCapacity);
+		}
 
-				LOG.debug("Signal unordered stream element queue has completed entries.");
-				hasCompletedEntries.signalAll();
+		/**
+		 * Signal that an entry finished computation.
```
zhijiangW: nit: Signal -> Signals

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/queue/UnorderedStreamElementQueue.java
```diff
@@ -218,88 +172,99 @@ public int size() {
...
+		}
 
-			if (firstSet.isEmpty()) {
-				firstSet.add(streamElementQueueEntry);
-			} else {
-				Set<StreamElementQueueEntry<?>> watermarkSet = new HashSet<>(1);
-				watermarkSet.add(streamElementQueueEntry);
-				uncompletedQueue.offer(watermarkSet);
+		/**
+		 * Adds the stages input elements for checkpointing including completed but not yet emitted elements.
```
zhijiangW: stages->staged?

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/queue/WatermarkQueueEntry.java
```diff
@@ -19,31 +19,44 @@
...
  * {@link StreamElementQueueEntry} implementation for the {@link Watermark}.
  */
 @Internal
-public class WatermarkQueueEntry extends StreamElementQueueEntry<Watermark> implements AsyncWatermarkResult {
+class WatermarkQueueEntry<OUT> implements StreamElementQueueEntry<OUT> {
+	@Nonnull
+	private final Watermark watermark;
 
-	private final CompletableFuture<Watermark> future;
+	WatermarkQueueEntry(@Nonnull Watermark watermark) {
```
zhijiangW: ditto: `@Nonnull` issue.
I guess it is the common sense for `Nonnull`. Then it is only necessary to tag it if `Nullable`?

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/queue/UnorderedStreamElementQueue.java
```diff
@@ -218,88 +172,99 @@ public int size() {
...
-				hasCompletedEntries.signalAll();
+		/**
+		 * Signal that an entry finished computation.
+		 */
+		void completed(StreamElementQueueEntry<OUT> elementQueueEntry) {
+			// adding only to completed queue if not completed before
+			// there may be a real result coming after the actual result, which is updated in the queue entry but
+			// the entry is not readded to the complete queue
+			if (pendingElements.remove(elementQueueEntry)) {
+				completedQueue.add(elementQueueEntry);
```
zhijiangW: In the past, after handling the current notified completed entry, it would fetch the next set and iterate to check whether there are completed entries in the next set. I do not know why it needs to do that before.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/execution/MailboxProcessor.java
```diff
@@ -200,7 +200,7 @@ private boolean processMail(TaskMailbox mailbox) throws MailboxStateException, I
 			try {
 				maybeLetter.get().run();
 			} catch (Exception e) {
-				e.printStackTrace();
+				throw new IllegalStateException("Cannot process mail " + maybeLetter.get(), e);
```
zhijiangW: If we want to interrupt the execution for any exceptions during looping the letters, I am wondering whether this wrapped exception provides any helpful debug information. If not, we could remove `try..catch` clauses here for simple.

zhijiangW:
Some review progress updates: I almost finished the core codes and left some tiny comments mainly for formatting. I would continue on the left tests codes later.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/queue/UnorderedStreamElementQueue.java
AHeise: It's actually the other way around: only elements within a stage can overtake each other. I will add a description to Stage and rewrite this description.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/queue/UnorderedStreamElementQueue.java
AHeise: The old implementation needed that for two reasons:
* it maintained an explicit first/last set, which needed to be updated when the current set is drained. I got rid of that because the added complexity didn't improve performance much (constant access to first/last).
* it needed to add completed elements from the next stage(s) to the completed queue, such that further elements could be pulled. I split the completed queue into smaller queues per stage, so that it can be directly updated and pulled. Since we have only one active stage most of the time, the memory overhead is smaller and the code should be easier now.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/execution/MailboxProcessor.java
AHeise: Can we actually use the async exception handler? I stumbled upon that in another PR.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/queue/StreamElementQueue.java
AHeise: No, IDE just automatically inserts it.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/queue/StreamElementQueue.java
AHeise: You are right. I added `hasCompleted()` mostly to avoid acquiring checkpoint lock unnecessarily, but the interface should be more abstract from the specific use case.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/queue/StreamRecordQueueEntry.java
AHeise: Yes. I was previously using lombok which generates the check and I got used to the style.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/queue/StreamRecordQueueEntry.java
AHeise: Added a check to AsyncWaitOperator where we receive use input/

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/queue/UnorderedStreamElementQueue.java
```diff
@@ -19,287 +19,261 @@
...
-	/** Queue of uncompleted stream element queue entries segmented by watermarks. */
-	private final ArrayDeque<Set<StreamElementQueueEntry<?>>> uncompletedQueue;
-
-	/** Queue of completed stream element queue entries. */
-	private final ArrayDeque<StreamElementQueueEntry<?>> completedQueue;
-
-	/** First (chronologically oldest) uncompleted set of stream element queue entries. */
-	private Set<StreamElementQueueEntry<?>> firstSet;
+	/** Queue of queue entries segmented by watermarks. */
+	private final Deque<Stage<OUT>> stages;
```
1u0: Personally I find "segment" as a more appropriate word here (also was used in the old java doc here) compared to word "stage".

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/AsyncWaitOperator.java
1u0: I think it's still possible that a timer callback can be executed later: it may happen, that the timer has triggered and posted a letter to the mailbox, but that letter would be executed later than the letter corresponding to proper result completion. Cancelling the `timeoutTimer` won't have much effect in that case.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/queue/StreamElementQueue.java
```diff
@@ -19,68 +19,56 @@
...
-	 * blocking and only returns once a completed async result has been found.
+	 * Checks if there is at least one completed element to be emitted.
 	 *
-	 * @return Completed {@link AsyncResult} which has been removed from the queue
-	 * @throws InterruptedException if the current thread has been interrupted while waiting for a
-	 * 	completed async result.
+	 * @return true if there is an element to be emitted.
 	 */
-	AsyncResult poll() throws InterruptedException;
+	boolean hasCompleted();
```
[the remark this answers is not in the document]
AHeise: His suggestion was just to make the doc of `hasCompleted` less dependent on `emitCompleted`.

I like your proposals concerning the name. I renamed them `hasCompletedElements` and `emitCompletedElement`.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/queue/UnorderedStreamElementQueue.java
AHeise: Okay, will change.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/AsyncWaitOperator.java
AHeise: It's unfortunately necessary and not an improvement in the closest sense.
If we don't "pump" elements here, then an empty stream consisting just of watermarks, will never emit anything. The only other way to emit elements is upon completion, which is imho the best place for normal records, but watermarks never take this code path. We could ofc trigger some kind of artificial completion on them, but that's more complicated imho.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/AsyncWaitOperator.java
AHeise: Hm that's definitively not what I expected from timers and I'm pretty sure it's hard for end-users to understand. My basic idea is: if I cancel a timer from 1 thread, then it will not execute in the same thread.

I would suggest hotfixing that in general by adding a safeguard at the start of the timer letter. WDYT? @1u0

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/functions/async/AsyncFunction.java
```diff
@@ -89,6 +89,7 @@
 	/**
 	 * {@link AsyncFunction#asyncInvoke} timeout occurred.
 	 * By default, the result future is exceptionally completed with a timeout exception.
+	 * Note that timeout and asyncInvoke might be invoked at the same time.
```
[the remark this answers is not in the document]
AHeise: I also thought about it, but the question is on how you define invoke:
if I have an asyncInvoke to an external system that takes a long time, then the timeout will be called when asyncInvoke is still in execution. so both are active at the same time, but ofc if you define invocation as the act of trigger the call, then it's not possible.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/AsyncWaitOperator.java
```diff
@@ -150,43 +139,26 @@ public void setup(StreamTask<?, ?> containingTask, StreamConfig config, Output<S
...
-		this.emitterThread = new Thread(emitter, "AsyncIO-Emitter-Thread (" + getOperatorName() + ')');
-		emitterThread.setDaemon(true);
-		emitterThread.start();
-
-		// process stream elements from state, since the Emit thread will start as soon as all
-		// elements from previous state are in the StreamElementQueue, we have to make sure that the
-		// order to open all operators in the operator chain proceeds from the tail operator to the
-		// head operator.
+		// Process stream elements from state. We have to make sure that the order to open all operators in the
+		// operator chain proceeds from the tail operator to the head operator.
```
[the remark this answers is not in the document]
AHeise: Yes, sounds like a good hotfix.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/functions/async/AsyncFunction.java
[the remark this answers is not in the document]
1u0: This method is currently supposed to be called in the task's main thread. It may happen that the timer would be triggered during `asyncInvoke()` call.
But (with the current changes in the `master` branch) the trigger itself won't call the `timeout()`, it would just post a mailbox letter to delegate the call to the task's main thread.
Before, this comment made sense, because both methods were called from different (independent) threads.

From API usage point of view, the `asyncInvoke()` is supposed to be "lightweight", it should initiate an actual asynchronous operation (maybe using some user defined execution pool) and give control back. Otherwise, it won't differ from normal "flat map" operator and would block the task's progress.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/execution/MailboxProcessor.java
zhijiangW: From the description of `AsyncExceptionHandler`, it is proposed to used by other threads except the main task thread. AFAIK it is only used by `SplitReader` in `ContinuousFileReaderOperator` now. For this case if the `processMail` is executed in the main task thread, maybe it is out of scope of `AsyncExceptionHandler`.

zhijiangW: Actually I concerned two issues here:

- `IllegalStateException` seems not very accurate here.

- If we want to provide more message for debugging, maybe we could use `ExceptionUtils.rethrow(e, "Cannot process mail " + maybeLetter.get());`. But I am not sure whether `maybeLetter.get()` could give any helpful info here. If the debug info is useless, it seems simple to remove `try...catch` directly.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/queue/UnorderedStreamElementQueue.java
zhijiangW: Yes. Elements within a stage could overtake each other. One more thing for further confirmation: how to guarantee that two stages can not overtake each other? 

e.g. assuming we have four stages : {a, b, c} {watermark1} {d, e, f} {watermark2}, and the first stage is the {a, b, c}. If the user function completes the element `e` in third stage firstly, that means the collection of `e` would be emitted to output before the first stage? Or I missed something else.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/TimestampedCollector.java
```diff
@@ -71,4 +73,19 @@ public void eraseTimestamp() {
...
 		output.close();
 	}
+
+	@Override
+	public void emitWatermark(Watermark mark) {
+		output.emitWatermark(mark);
+	}
+
+	@Override
+	public <X> void collect(OutputTag<X> outputTag, StreamRecord<X> record) {
```
zhijiangW: It is involved in three new methods by implementing `Output` interface instead. But actually we only need the `emitWatermark` method, and the left two methods would never be touched in codes. If so, another option is that we can only define `emitWatermark` method here and still implement `Collector` interface.

flink-streaming-java/src/test/java/org/apache/flink/streaming/api/operators/async/AsyncWaitOperatorTest.java
```diff
@@ -100,6 +102,9 @@
 public class AsyncWaitOperatorTest extends TestLogger {
 	private static final long TIMEOUT = 1000L;
 
+	@Rule
+	public Timeout timeoutRule = new Timeout(10, TimeUnit.SECONDS);
```
zhijiangW: Why we need this change? 
Some previous tests were configured separately as 2 seconds, but now we unify it for 10 seconds for all the cases.

flink-streaming-java/src/test/java/org/apache/flink/streaming/api/operators/async/queue/OrderedStreamElementQueueTest.java
```diff
@@ -18,110 +18,82 @@
...
-import java.util.concurrent.Executors;
-import java.util.concurrent.TimeUnit;
 
-import static org.mockito.Matchers.any;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.never;
-import static org.mockito.Mockito.verify;
+import static org.apache.flink.streaming.api.operators.async.queue.QueueUtil.popCompleted;
+import static org.apache.flink.streaming.api.operators.async.queue.QueueUtil.putSucessfully;
+
```
zhijiangW: remove extra empty line

flink-streaming-java/src/test/java/org/apache/flink/streaming/api/operators/async/queue/QueueUtil.java
```diff
@@ -0,0 +1,62 @@
...
+import java.util.Optional;
+
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+/**
+ * Utility for putting elements inside a {@link StreamElementQueue}.
+ */
+class QueueUtil {
+	static ResultFuture<Integer> putSucessfully(StreamElementQueue<Integer> queue, StreamElement streamElement) {
```
zhijiangW: typo: Sucess -> Success

flink-streaming-java/src/test/java/org/apache/flink/streaming/api/operators/async/queue/QueueUtil.java
```diff
@@ -0,0 +1,62 @@
...
+		Optional<ResultFuture<Integer>> resultFuture = queue.tryPut(streamElement);
+		assertTrue(resultFuture.isPresent());
+		return resultFuture.get();
+	}
+
+	static void putUnsucessfully(StreamElementQueue<Integer> queue, StreamElement streamElement) {
```
zhijiangW: typo: sucess

flink-streaming-java/src/test/java/org/apache/flink/streaming/api/operators/async/queue/QueueUtil.java
```diff
@@ -0,0 +1,62 @@
...
+		Optional<ResultFuture<Integer>> resultFuture = queue.tryPut(streamElement);
+		assertFalse(resultFuture.isPresent());
+	}
+
+	/**
+	 * Pops all completed elements from the head of this queue.
+	 *
+	 * @return Completed elements or empty list of none exists.
```
zhijiangW: `empty list if none exists`

flink-streaming-java/src/test/java/org/apache/flink/streaming/api/operators/async/queue/QueueUtil.java
```diff
@@ -0,0 +1,62 @@
...
+	 */
+	static List<StreamElement> popCompleted(StreamElementQueue<Integer> queue) {
+		final List<StreamElement> completed = new ArrayList<>();
+		TimestampedCollector<Integer> collector = new TimestampedCollector<>(new CollectorOutput<>(completed));
+		while (queue.hasCompletedElements()) {
+			queue.emitCompletedElement(collector);
+		}
+		collector.close();
```
zhijiangW: ATM the `close` actually did nothing, but the proposed semantic from `Collector#close` would flush buffered data. If so I wonder it might have potential risk to clear the `completed` list before return if the implementation is changed future.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/queue/StreamElementQueue.java
```diff
@@ -19,68 +19,56 @@
...
+	 *
+	 * @param output the output into which to emit
 	 */
-	AsyncResult peekBlockingly() throws InterruptedException;
+	void emitCompletedElement(TimestampedCollector<OUT> output);
 
 	/**
-	 * Poll the first completed {@link AsyncResult} from the head of this queue. This operation is
-	 * blocking and only returns once a completed async result has been found.
+	 * Checks if there is at least one completed element.
```
zhijiangW: This description is not very accurate for the ordered case. For order case, it must be the head element completed in queue.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/execution/MailboxProcessor.java
1u0: `+1` to remove the `try..catch` here.

flink-streaming-java/src/test/java/org/apache/flink/streaming/api/operators/async/queue/OrderedStreamElementQueueTest.java
```diff
@@ -18,110 +18,82 @@
...
+		Assert.assertEquals(2, queue.size());
+		Assert.assertFalse(queue.isEmpty());
+
+		Assert.assertEquals(Arrays.asList(
+				new Watermark(2L),
+				new Watermark(5L)),
+				popCompleted(queue));
+		Assert.assertEquals(0, queue.size());
+		Assert.assertTrue(queue.isEmpty());
+		Assert.assertEquals(Arrays.asList(), popCompleted(queue));
```
zhijiangW: Intellij inspection suggestions `Collections.emptyList()` instead.

flink-streaming-java/src/test/java/org/apache/flink/streaming/api/operators/async/queue/OrderedStreamElementQueueTest.java
```diff
@@ -18,110 +18,82 @@
...
+		final StreamElementQueue<Integer> queue = new OrderedStreamElementQueue<>(8);
+
+		putSucessfully(queue, new Watermark(2L));
+		putSucessfully(queue, new Watermark(5L));
+
+		Assert.assertEquals(2, queue.size());
+		Assert.assertFalse(queue.isEmpty());
+
+		Assert.assertEquals(Arrays.asList(
+				new Watermark(2L),
```
zhijiangW: ditto: indentation

flink-streaming-java/src/test/java/org/apache/flink/streaming/api/operators/async/queue/OrderedStreamElementQueueTest.java
```diff
@@ -18,110 +18,82 @@
...
 
-		Assert.assertEquals(expected, pollOperation.get());
+		Assert.assertEquals(Collections.emptyList(), popCompleted(queue));
+		Assert.assertEquals(4, queue.size());
+		Assert.assertFalse(queue.isEmpty());
+
+		entry1.complete(Collections.singleton(10));
+
+		List<StreamElement> expected = Arrays.asList(
+				new StreamRecord<>(10, 0L),
```
zhijiangW: reduce one indentation.

flink-streaming-java/src/test/java/org/apache/flink/streaming/api/operators/async/queue/OrderedStreamElementQueueTest.java
```diff
@@ -18,110 +18,82 @@
...
+		Assert.assertTrue(queue.isEmpty());
+	}
 
-		verify(operatorActions, never()).failOperator(any(Exception.class));
+	/**
+	 * Tests two adjacent watermarks can be processed successfully.
+	 */
+	@Test
+	public void testWatermarkOnly() {
+		final StreamElementQueue<Integer> queue = new OrderedStreamElementQueue<>(8);
```
zhijiangW: nit : 8 -> 2

flink-streaming-java/src/test/java/org/apache/flink/streaming/api/operators/async/queue/StreamElementQueueTest.java
```diff
@@ -18,253 +18,117 @@
...
+		this.outputMode = Preconditions.checkNotNull(outputMode);
 	}
 
-	public StreamElementQueue createStreamElementQueue(int capacity, OperatorActions operatorActions) {
-		switch (streamElementQueueType) {
-			case OrderedStreamElementQueueType:
-				return new OrderedStreamElementQueue(capacity, executor, operatorActions);
-			case UnorderedStreamElementQueueType:
-				return new UnorderedStreamElementQueue(capacity, executor, operatorActions);
+	public StreamElementQueue<Integer> createStreamElementQueue(int capacity) {
```
zhijiangW: public -> private

flink-streaming-java/src/test/java/org/apache/flink/streaming/api/operators/async/queue/StreamElementQueueTest.java
```diff
@@ -18,253 +18,117 @@
...
-
-	enum StreamElementQueueType {
-		OrderedStreamElementQueueType,
-		UnorderedStreamElementQueueType
-	}
-
 	@Parameterized.Parameters
-	public static Collection<StreamElementQueueType> streamElementQueueTypes() {
-		return Arrays.asList(OrderedStreamElementQueueType, UnorderedStreamElementQueueType);
+	public static Collection<AsyncDataStream.OutputMode> streamElementQueueTypes() {
```
zhijiangW: streamElementQueueTypes -> outputModes

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/AsyncWaitOperator.java
1u0: Allowing cancellation of timer callbacks that are already in mailbox queue may help in this situation, but still won't have strict guarantees in general (as it was before with multithreaded code).

Even with a hotfix, you may still need to somehow track/allow only one result (either successfull or failed) per `ResultHandler`. With current implementation, an ill written `AsyncFunction` may call multiple times `resultHandler.complete()` (on the same ``resultHandler` instance) and every such call would post a new mailbox letter.

flink-streaming-java/src/test/java/org/apache/flink/streaming/api/operators/async/queue/UnorderedStreamElementQueueTest.java
```diff
@@ -18,174 +18,108 @@
...
-					return queue.poll();
-				} catch (InterruptedException e) {
-					throw new CompletionException(e);
-				}
-			},
-			executor);
+		Assert.assertEquals(Arrays.asList(new StreamRecord<>(15, 6L)), popCompleted(queue));
+		Assert.assertEquals(0, queue.size());
+		Assert.assertTrue(queue.isEmpty());
+		Assert.assertEquals(Arrays.asList(), popCompleted(queue));
```
zhijiangW: `Collections.emptyList()` instead

flink-streaming-java/src/test/java/org/apache/flink/streaming/api/operators/async/queue/UnorderedStreamElementQueueTest.java
```diff
@@ -18,174 +18,108 @@
...
+	}
 
-		Thread.sleep(10L);
 
-		Assert.assertFalse(thirdPoll.isDone());
+	/**
+	 * Tests two adjacent watermarks can be processed successfully.
+	 */
+	@Test
+	public void testWatermarkOnly() {
```
zhijiangW: This test is totally the same with `OrderedStreamElementQueueTest#testWatermarkOnly`, so we could migrate it into `StreamElementQueueTest` instead.

zhijiangW:
@AHeise I have finished the first round review. Almost LGTM only for some tiny comments.

In addition, for the `hotfix` commit which would be merged separately, but I noticed `[hotfix] [runtime] Moving specification on how to properly implement` include some `AsyncWaitOperator` codes which should be excluded.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/AsyncWaitOperator.java
```diff
@@ -293,140 +252,120 @@ public void close() throws Exception {
...
-				}
-			} catch (InterruptedException e) {
-				executor.shutdownNow();
-
-				Thread.currentThread().interrupt();
+	private void outputCompletedElements() {
+		if (queue.hasCompletedElements()) {
+			// emit one element
+			synchronized (checkpointingLock) {
+				queue.emitCompletedElement(this.timestampedCollector);
```
zhijiangW: remove `this`

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/AsyncWaitOperator.java
```diff
@@ -293,140 +252,120 @@ public void close() throws Exception {
...
-			/*
-			 * FLINK-5638: If we have the checkpoint lock we might have to free it for a while so
-			 * that the emitter thread can complete/react to the interrupt signal.
-			 */
-			if (Thread.holdsLock(checkpointingLock)) {
-				while (emitterThread.isAlive()) {
-					checkpointingLock.wait(100L);
-				}
+			// if there are more completed elements, emit them with subsequent mails
+			if (queue.hasCompletedElements()) {
```
zhijiangW: I guess if we do not do this, it could still work? Because the result would finally be completed by async function.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/AsyncWaitOperator.java
```diff
@@ -293,140 +252,120 @@ public void close() throws Exception {
...
-					executor.shutdownNow();
-				}
-			} catch (InterruptedException e) {
-				executor.shutdownNow();
-
-				Thread.currentThread().interrupt();
+	private void outputCompletedElements() {
+		if (queue.hasCompletedElements()) {
+			// emit one element
+			synchronized (checkpointingLock) {
```
zhijiangW: It seems that the lock could not be removed completely while output, so maybe we need to adjust the commit title "Reducing synchronization.....". I guess the main motivation is to remove `EmitterThread`?

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/functions/async/AsyncFunction.java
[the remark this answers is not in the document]
AHeise: Okay I'm removing it.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/TimestampedCollector.java
AHeise: Yes, ideally we move all `emit*` to the `TimestampedCollector` and avoid going through the unwrapped output. I'm checking how much code would be affected.

flink-streaming-java/src/test/java/org/apache/flink/streaming/api/operators/async/queue/QueueUtil.java
AHeise: We need to call close here exactly because it flushes data (in this case it could flush buffered data into the list). Compare with Kryo's `Output`.

Flushing the list on `close` would be an odd behavior that breaks with the expectations that we have from `java.io.StringWriter`, `ByteArrayOutputStream`, etc.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/AsyncWaitOperator.java
AHeise: Unfortunately, it would not work completely.

Let's say we have two items i1, i2 in an ordered queue and they complete in reverse order. Then the async callback of i2 will not emit any element. The async callback of i2 will emit only the first element i1. So, i2 is stuck. Of course, it will be emitted as soon as any later item completed, but we have an artificial lag and for bounded inputs, we would actually never emit the last few elements.

Originally, `emitCompletedElement` emitted all elements, so we wouldn't need to enqueue an additional mail. However, that may block the mailbox for a longer time, such that interleaved events will not be processed.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/AsyncWaitOperator.java
AHeise: Yes, I will do so.

flink-streaming-java/src/test/java/org/apache/flink/streaming/api/operators/async/AsyncWaitOperatorTest.java
AHeise: These timeouts are just a safe-guard whenever we break stuff. All tests should always complete without timeout. But it will be especially useful on Travis as we do not need to wait for Travis to timeout to get a test result.

Instead of just annotating the two tests that just failed (as we did in the past), I'd annotate the whole suite with the junit rule, which makes it more readable. Imho @1u0 even started with that rule and I just copied him ;).

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/TimestampedCollector.java
AHeise: Opening a separate ticket for that as many code paths are touched.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/mailbox/execution/MailboxProcessor.java
AHeise: I will remove commit and address it in #9772 .

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/AsyncWaitOperator.java
AHeise: Added a guard now in `ResultHandler`.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/AsyncWaitOperator.java
```diff
@@ -293,140 +252,138 @@ public void close() throws Exception {
...
 
-	@Override
-	public void failOperator(Throwable throwable) {
-		getContainingTask().getEnvironment().failExternally(throwable);
+		@Override
+		public void completeExceptionally(Throwable error) {
+			// already completed, so ignore exception
+			if(completed) {
+				return;
+			}
```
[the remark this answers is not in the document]
AHeise: Using AtomicBoolean now.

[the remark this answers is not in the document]
AHeise: I'd say any callback from user code should be considered final. If user code first invokes `completeExceptionally()`, then `complete()` is unexpected and will be ignored.

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/AsyncWaitOperator.java
```diff
@@ -293,140 +245,136 @@ public void close() throws Exception {
...
 
-		pendingStreamElementQueueEntry = null;
-	}
+		@Override
+		public void complete(Collection<OUT> results) {
+			Preconditions.checkNotNull(results, "Results must not be null, use empty collection to emit nothing");
 
-	private void waitInFlightInputsFinished() throws InterruptedException {
-		assert(Thread.holdsLock(checkpointingLock));
+			// already completed, so ignore exception
```
1u0: This comment is not relevant for `complete(results)` method. :)

flink-streaming-java/src/main/java/org/apache/flink/streaming/api/operators/async/AsyncWaitOperator.java
```diff
@@ -293,140 +252,138 @@ public void close() throws Exception {
...
+	 * @return a handle that allows to set the result of the async computation for the given element.
+	 */
+	private ResultFuture<OUT> addToWorkQueue(StreamElement streamElement) throws InterruptedException {
+		assert(Thread.holdsLock(checkpointingLock));
 
-		try {
-			super.dispose();
-		} catch (InterruptedException interrupted) {
-			exception = interrupted;
+		pendingStreamElement = streamElement;
```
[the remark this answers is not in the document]
AHeise: Fixed in hotfix.

pnowojski:
Thanks for the contribution and the reviews. Merging.

PR #9719, opened 2019-09-19
[FLINK-14010][coordination] Shutdown cluster if ResourceManager terminated unexpectedly

TisonKun:

Context can be found at [FLINK-14010](https://issues.apache.org/jira/browse/FLINK-14010)

## Brief change log

Add callback logic when ResourceManager terminates unexpectedly, complete `DispatcherResourceManagerComponent#shutdownFuture` exceptionally.

## Verifying this change

The change is technically trivial and can be easily reasoned. However, I'd like to introduce a test of it, but cannot find a good way.

  - The serializers: (no)
  - Anything that affects deployment or recovery: JobManager (and its components), Checkpointing, Yarn/Mesos, ZooKeeper: (yes, now in YARN deployment Flink App can correctly response to shutdownRequest)

## Documentation

### DIFF
--- a/flink-yarn/src/main/java/org/apache/flink/yarn/YarnResourceManager.java
+++ b/flink-yarn/src/main/java/org/apache/flink/yarn/YarnResourceManager.java
@@ -96,6 +96,8 @@ public class YarnResourceManager extends ResourceManager<YarnWorkerNode> impleme
 	 * In task executor we use the hostnames given by YARN consistently throughout akka */
 	static final String ENV_FLINK_NODE_ID = "_FLINK_NODE_ID";
 
+	static final String ERROR_MASSAGE_ON_SHUTDOWN_REQUEST = "Received shutdown request from YARN ResourceManager.";
+
 	/** Default heartbeat interval between this resource manager and the YARN ResourceManager. */
 	private final int yarnHeartbeatIntervalMillis;
 
@@ -452,7 +454,7 @@ public class YarnResourceManager extends ResourceManager<YarnWorkerNode> impleme
 
 	@Override
 	public void onShutdownRequest() {
-		closeAsync();
+		onFatalError(new ResourceManagerException(ERROR_MASSAGE_ON_SHUTDOWN_REQUEST));
 	}
 
 	@Override
--- a/flink-yarn/src/test/java/org/apache/flink/yarn/YarnResourceManagerTest.java
+++ b/flink-yarn/src/test/java/org/apache/flink/yarn/YarnResourceManagerTest.java
@@ -39,6 +39,7 @@ import org.apache.flink.runtime.registration.RegistrationResponse;
 import org.apache.flink.runtime.resourcemanager.JobLeaderIdService;
 import org.apache.flink.runtime.resourcemanager.ResourceManagerGateway;
 import org.apache.flink.runtime.resourcemanager.SlotRequest;
+import org.apache.flink.runtime.resourcemanager.exceptions.ResourceManagerException;
 import org.apache.flink.runtime.resourcemanager.slotmanager.SlotManager;
 import org.apache.flink.runtime.resourcemanager.utils.MockResourceManagerRuntimeServices;
 import org.apache.flink.runtime.rpc.FatalErrorHandler;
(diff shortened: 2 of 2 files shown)


flink-runtime/src/main/java/org/apache/flink/runtime/entrypoint/component/DispatcherResourceManagerComponent.java
```diff
@@ -106,6 +107,16 @@ private void registerShutDownFuture() {
...
 					}
 				});
+
+		resourceManager
+			.getTerminationFuture()
+			.whenComplete(
+				(aVoid, throwable) -> {
+					if (isRunning.get()) {
+						shutDownFuture.completeExceptionally(
+							new FlinkException("ResourceManager got closed when DispatcherResourceManagerComponent is running"));
```
tillrohrmann:
```suggestion
							new FlinkException("ResourceManager was unexpectedly terminated while DispatcherResourceManagerComponent was running. This indicates either a failure of the ResourceManager or that it was asked to shut down by the external resource management system."));
```

TisonKun:
Hi @tillrohrmann Thanks for your review.

I add a `DispatcherResourceManagerComponentTest` and expose some visibility, please double check it doesn't cause other issues.

Besides, I find another code change required by this approach that actually we are unable to `deregisterApplication` on `DispatcherResourceManager` closed because RM is already terminated.

It seems we twist codes a bit more so although the logic is correct at the moment I still prefer just call `onFatalError` in `YarnResourceManager#onShutdownRequest` which fixes the original issue specifically. We cannot do anything when we lost connection to YARN RM and lost leadership semantically. Anything I miss we can gain by gracefully closing ClusterEntrypoint?

flink-runtime/src/test/java/org/apache/flink/runtime/entrypoint/component/DispatcherResourceManagerComponentTest.java
```diff
@@ -0,0 +1,80 @@
...
+					}
+				}
+			);
+			component.getResourceManager().closeAsync();
+
+			String errorMessage = "ResourceManager was unexpectedly terminated while DispatcherResourceManagerComponent was running.";
+			Throwable t = future.get(2000L, TimeUnit.MILLISECONDS);
+			assertThat(ExceptionUtils.findThrowable(t, FlinkException.class).isPresent(), is(true));
+			assertThat(ExceptionUtils.findThrowableWithMessage(t, errorMessage).isPresent(), is(true));
+		}
```
tillrohrmann: I think a more targeted test could be possible by directly testing a `DispatcherResourceManagerComponent` without starting a `MiniCluster`.

TisonKun:
Thanks for your insight @tillrohrmann. I put on the version causing fatal error in `YarnResourceManager#onShutdownRequest` as well as the test. Please review and share the idea which one you prefer after both implementation are present.

flink-runtime/src/test/java/org/apache/flink/runtime/entrypoint/component/DispatcherResourceManagerComponentTest.java
TisonKun: Sounds reasonable. When I said "I'd like to introduce a test of it, but cannot find a good way." in pull request description actually I mean I cannot find a testing `DispatcherResourceManagerComponent` 😄 . However, now I think we can just use `DispatcherResourceManagerComponent` because it is well-design for testing also.

If we eventually adopt the way changes in `DispatcherResourceManagerComponent`, I will do the update.

flink-yarn/src/main/java/org/apache/flink/yarn/YarnResourceManager.java
```diff
@@ -452,7 +454,7 @@ private void removeContainerRequest(AMRMClient.ContainerRequest pendingContainer
 
 	@Override
 	public void onShutdownRequest() {
-		closeAsync();
+		onFatalError(new FlinkException(ERROR_MASSAGE_ON_SHUTDOWN_REQUEST));
```
tillrohrmann:
```suggestion
		onFatalError(new ResourceManagerException(ERROR_MASSAGE_ON_SHUTDOWN_REQUEST));
```

TisonKun: Thanks for your review and fix @tillrohrmann!

PR #9721, opened 2019-09-19
[FLINK-14129][hive] HiveTableSource should implement ProjectableTable…

lirui-apache:
…Source

Implement ProjectableTableSource for HiveTableSource.

## Brief change log

  - Make HiveTableSource implement ProjectableTableSource.
  - HiveTableInputFormat won't inspect/return columns that are not needed.
  - New test case make sure it works.

## Verifying this change

Added test case

  - The serializers: no
  - The runtime per-record code paths (performance sensitive): yes

## Documentation

### DIFF
--- a/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/connectors/hive/HiveTableInputFormat.java
+++ b/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/connectors/hive/HiveTableInputFormat.java
@@ -89,8 +89,8 @@ public class HiveTableInputFormat extends HadoopInputFormatCommonBase<Row, HiveT
 	// indices of fields to be returned, with projection applied (if any)
 	// TODO: push projection into underlying input format that supports it
 	private int[] fields;
-	// Pre-computed partition values that are needed to generate a row.
-	private transient List<Object> partitionValues;
+	// Remember whether a row instance is reused. No need to set partition fields for reused rows
+	private transient boolean rowReused;
 
 	public HiveTableInputFormat(
 			JobConf jobConf,
@@ -143,15 +143,7 @@ public class HiveTableInputFormat extends HadoopInputFormatCommonBase<Row, HiveT
 		} catch (Exception e) {
 			throw new FlinkHiveException("Error happens when deserialize from storage file.", e);
 		}
-		if (!partitionColNames.isEmpty()) {
-			partitionValues = new ArrayList<>();
-			for (int field : fields) {
-				if (field > structFields.size()) {
-					String partition = partitionColNames.get(field - structFields.size());
-					partitionValues.add(hiveTablePartition.getPartitionSpec().get(partition));
-				}
-			}
-		}
+		rowReused = false;
 	}
 
 	@Override
@@ -218,33 +210,40 @@ public class HiveTableInputFormat extends HadoopInputFormatCommonBase<Row, HiveT
 	}
 
 	@Override
-	public Row nextRecord(Row ignore) throws IOException {
+	public Row nextRecord(Row reuse) throws IOException {
 		if (reachedEnd()) {
 			return null;
 		}
-		Row row = new Row(fields.length);
 		try {
 			//Use HiveDeserializer to deserialize an object out of a Writable blob
 			Object hiveRowStruct = deserializer.deserialize(value);
-			int partitionValueIndex = 0;
 			for (int i = 0; i < fields.length; i++) {
-				// non-partition column
+				// set non-partition columns
 				if (fields[i] < structFields.size()) {
 					StructField structField = structFields.get(fields[i]);
 					Object object = HiveInspectors.toFlinkObject(structField.getFieldObjectInspector(),
 							structObjectInspector.getStructFieldData(hiveRowStruct, structField));
-					row.setField(i, object);
-				} else {
-					// partition column
-					row.setField(i, partitionValues.get(partitionValueIndex++));
+					reuse.setField(i, object);
 				}
 			}
-		} catch (Exception e){
+		} catch (Exception e) {
 			logger.error("Error happens when converting hive data type to flink data type.");
 			throw new FlinkHiveException(e);
 		}
+		if (!rowReused) {
+			// set partition columns
+			if (!partitionColNames.isEmpty()) {
+				for (int i = 0; i < fields.length; i++) {
+					if (fields[i] >= structFields.size()) {
+						String partition = partitionColNames.get(fields[i] - structFields.size());
+						reuse.setField(i, hiveTablePartition.getPartitionSpec().get(partition));
+					}
+				}
+			}
+			rowReused = true;
+		}
 		this.fetched = false;
-		return row;
+		return reuse;
 	}
 
 	// --------------------------------------------------------------------------------------------
--- a/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/connectors/hive/HiveTableInputFormat.java
+++ b/flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/connectors/hive/HiveTableInputFormat.java
@@ -52,6 +52,7 @@ import java.io.ObjectOutputStream;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Properties;
+import java.util.stream.IntStream;
 
 import static org.apache.flink.util.Preconditions.checkNotNull;
 import static org.apache.hadoop.mapreduce.lib.input.FileInputFormat.INPUT_DIR;
@@ -73,9 +74,6 @@ public class HiveTableInputFormat extends HadoopInputFormatCommonBase<Row, HiveT
 	protected transient boolean fetched = false;
 	protected transient boolean hasNext;
 
-	// arity of each row, including partition columns
-	private int rowArity;
-
 	//Necessary info to init deserializer
 	private List<String> partitionColNames;
 	//For non-partition hive table, partitions only contains one partition which partitionValues is empty.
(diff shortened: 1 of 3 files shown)


lirui-apache:
cc @JingsongLi @bowenli86 @xuefuz @zjuwangg

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/connectors/hive/HiveTableInputFormat.java
```diff
@@ -207,19 +211,22 @@ public Row nextRecord(Row ignore) throws IOException {
...
+			for (int i = 0; i < fields.length; i++) {
+				// non-partition column
+				if (fields[i] < structFields.size()) {
+					StructField structField = structFields.get(fields[i]);
+					Object object = HiveInspectors.toFlinkObject(structField.getFieldObjectInspector(),
+							structObjectInspector.getStructFieldData(hiveRowStruct, structField));
+					row.setField(i, object);
+				} else {
+					// partition column
+					String partition = partitionColNames.get(fields[i] - structFields.size());
```
bowenli86: minor: the partition column names seem can be precomputed as they stay the same for rows in the same partition?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/connectors/hive/HiveTableSource.java
```diff
@@ -106,7 +113,17 @@ public TableSchema getTableSchema() {
...
 	public DataType getProducedDataType() {
-		return getTableSchema().toRowDataType();
+		TableSchema originSchema = getTableSchema();
+		if (projectedFields == null) {
+			return originSchema.toRowDataType();
+		}
+		String[] names = new String[projectedFields.length];
+		DataType[] types = new DataType[projectedFields.length];
+		for (int i = 0; i < projectedFields.length; i++) {
+			names[i] = originSchema.getFieldNames()[projectedFields[i]];
```
JingsongLi: Use `getFieldDataType(int fieldIndex)` and `getFieldName(int fieldIndex)`.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/connectors/hive/HiveTableInputFormat.java
```diff
@@ -88,17 +86,23 @@
 	private transient InputFormat mapredInputFormat;
 	private transient HiveTablePartition hiveTablePartition;
 
+	// indices of fields to be returned, with projection applied (if any)
+	// TODO: push projection into underlying input format that supports it
```
JingsongLi: I don't think underlying input format can support project push down. Any information or delete this comment?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/connectors/hive/HiveTableSource.java
```diff
@@ -66,6 +68,7 @@
 	private Map<Map<String, String>, HiveTablePartition> partitionSpec2HiveTablePartition = new HashMap<>();
 	private boolean initAllPartitions;
 	private boolean partitionPruned;
+	private int[] projectedFields;
 
 	public HiveTableSource(JobConf jobConf, ObjectPath tablePath, CatalogTable catalogTable) {
 		this.jobConf = Preconditions.checkNotNull(jobConf);
```
JingsongLi: Invoke `this()` to init variable, like `projectedFields`, you can init it as `IntStream.range(0, rowArity).toArray()`

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/connectors/hive/HiveTableInputFormat.java
```diff
@@ -88,17 +86,23 @@
...
+			int[] projectedFields) {
 		super(jobConf.getCredentials());
 		checkNotNull(catalogTable, "catalogTable can not be null.");
 		this.partitions = checkNotNull(partitions, "partitions can not be null.");
 
 		this.jobConf = new JobConf(jobConf);
 		this.partitionColNames = catalogTable.getPartitionKeys();
-		rowArity = catalogTable.getSchema().getFieldCount();
+		int rowArity = catalogTable.getSchema().getFieldCount();
+		fields = projectedFields != null ? projectedFields : IntStream.range(0, rowArity).toArray();
```
JingsongLi: Let `projectedFields` never null?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/connectors/hive/HiveTableInputFormat.java
lirui-apache: Yeah, will pre-compute the partition values.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/connectors/hive/HiveTableInputFormat.java
lirui-apache: I haven't checked the code but I think columnar storage formats should be able to support projection pushdown. For example, these slides imply projection pushdown is supported in ORC and Parquet:
https://www.slideshare.net/oom65/orc-files?qid=a032419f-e6a3-4776-8c9c-03dd752f17fd&v=&b=&from_search=2
https://www.slideshare.net/julienledem/parquet-stratany-hadoopworld2013?qid=94dc3a84-39f0-456b-8766-f623e679a7ca&v=&b=&from_search=4

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/connectors/hive/HiveTableInputFormat.java
```diff
@@ -207,19 +211,22 @@ public Row nextRecord(Row ignore) throws IOException {
 		if (reachedEnd()) {
 			return null;
 		}
-		Row row = new Row(rowArity);
+		Row row = new Row(fields.length);
```
JingsongLi: Why not use the row in `nextRecord(Row ignore)`? with reuse this row, you can never reassign partition values after initialization.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/connectors/hive/HiveTableInputFormat.java
JingsongLi: I think orc/parquet is another story, we should use our implementation instead of hadoop `RecordReader`.
What I want to ask is: can `RecordReader` support project push down?
If it is a easy way, Can we support it now?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/connectors/hive/HiveTableSource.java
lirui-apache: I'd like to keep `projectedFields ` as null to indicate no projection has been pushed down to this table source. So that `explainSource()` doesn't have to display all the column indices if no projection is ever applied. What do you think?

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/connectors/hive/HiveTableInputFormat.java
lirui-apache: > can RecordReader support project push down?

I guess it depends on the implementation. For example, Hive has its own [ORC input format](https://github.com/apache/hive/blob/rel/release-2.3.4/ql/src/java/org/apache/hadoop/hive/ql/io/orc/OrcInputFormat.java). And it seems Hive pushes down the projection via [configurations](https://github.com/apache/hive/blob/rel/release-2.3.4/serde/src/java/org/apache/hadoop/hive/serde2/ColumnProjectionUtils.java#L52).

> Can we support it now?

I'd rather leave it to a separate task because it needs further investigation. Even though this PR doesn't reduce the amount of data to read, it avoids the costs of inspecting unused columns, which is good for queries selecting only a few columns from tables with lots of columns. And for input format that doesn't support projection push down (e.g. text), this is probably the best we can do.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/connectors/hive/HiveTableSource.java
JingsongLi: It is OK

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/connectors/hive/HiveTableSource.java
```diff
@@ -80,23 +83,27 @@ public HiveTableSource(JobConf jobConf, ObjectPath tablePath, CatalogTable catal
 	private HiveTableSource(JobConf jobConf, ObjectPath tablePath, CatalogTable catalogTable,
```
zjuwangg: Maybe it's better to add some method comments here to indicate its effect on projection or partition-prunning.

lirui-apache: Sure

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/connectors/hive/HiveTableInputFormat.java
lirui-apache: Since we've agreed to let projectedFields be null in HiveTableSource, I'll leave this unchanged.

lirui-apache:
PR updated to address comments.
@bowenli86 I'm not pre-computing partition names/values. As @JingsongLi suggested, we can reuse the row instance in `nextRecord`, and therefore we only need to compute the partition values once.
@xuefuz I've created https://issues.apache.org/jira/browse/FLINK-14148 to track the TODO.

flink-connectors/flink-connector-hive/src/main/java/org/apache/flink/connectors/hive/HiveTableInputFormat.java
```diff
@@ -203,30 +210,40 @@ protected void fetchNext() throws IOException {
...
+							structObjectInspector.getStructFieldData(hiveRowStruct, structField));
+					reuse.setField(i, object);
+				}
 			}
-		} catch (Exception e){
+		} catch (Exception e) {
 			logger.error("Error happens when converting hive data type to flink data type.");
 			throw new FlinkHiveException(e);
 		}
+		if (!rowReused) {
```
bowenli86: It seems to be initialization logic to me, as it's always triggered for the first row, but still have to be checked for all other remaining rows regardless?

maybe we can add a `Row` member variable to HiveTableInputFormet, and precompute and set partition values for it based on projection, and just reuse that row in `nextRecord()`?

lirui-apache: > it's always triggered for the first row, but still have to be checked for all other remaining rows regardless?

Yes. But it's still slightly better than populating each partition columns for a row.

> maybe we can add a Row member variable to HiveTableInputFormet, and precompute and set partition values for it based on projection, and just reuse that row in nextRecord()?

That's possible. My concern is we don't have guarantee that the fields of the row we return in `nextRecord` won't be manipulated by the framework or downstream operators. Although both solutions have this problem, the current implementation is a little bit more "compliant" with the interface contract -- that the Row parameter is the instance to be reused.

bowenli86:
ok. thanks, LGTM

PR #9727, opened 2019-09-20
[FLINK-14145] Fix getLatestCheckpoint(true) returns wrong checkpoint

gyfora:

Fixes a bug where the CompletedCheckpointStore would not return the last checkpoint correctly if checkpoints are preferred.

## Verifying this change

Added new test case by @uce to StandaloneCompletedCheckpointStoreTest

  - The serializers: no

## Documentation

### DIFF
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/checkpoint/CompletedCheckpointStore.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/checkpoint/CompletedCheckpointStore.java
@@ -19,7 +19,6 @@
 package org.apache.flink.runtime.checkpoint;
 
 import org.apache.flink.runtime.jobgraph.JobStatus;
-import org.apache.flink.util.FlinkRuntimeException;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -55,31 +54,26 @@ public interface CompletedCheckpointStore {
 	 * added.
 	 */
 	default CompletedCheckpoint getLatestCheckpoint(boolean isPreferCheckpointForRecovery) throws Exception {
-		if (getAllCheckpoints().isEmpty()) {
+		List<CompletedCheckpoint> allCheckpoints = getAllCheckpoints();
+		if (allCheckpoints.isEmpty()) {
 			return null;
 		}
 
-		CompletedCheckpoint candidate = getAllCheckpoints().get(getAllCheckpoints().size() - 1);
-		if (isPreferCheckpointForRecovery && getAllCheckpoints().size() > 1) {
-			List<CompletedCheckpoint> allCheckpoints;
-			try {
-				allCheckpoints = getAllCheckpoints();
-				ListIterator<CompletedCheckpoint> listIterator = allCheckpoints.listIterator(allCheckpoints.size() - 1);
-				while (listIterator.hasPrevious()) {
-					CompletedCheckpoint prev = listIterator.previous();
-					if (!prev.getProperties().isSavepoint()) {
-						candidate = prev;
-						LOG.info("Found a completed checkpoint before the latest savepoint, will use it to recover!");
-						break;
-					}
+		CompletedCheckpoint lastCompleted = allCheckpoints.get(allCheckpoints.size() - 1);
+
+		if (lastCompleted.getProperties().isSavepoint() && isPreferCheckpointForRecovery && allCheckpoints.size() > 1) {
+			ListIterator<CompletedCheckpoint> listIterator = allCheckpoints.listIterator(allCheckpoints.size() - 1);
+			while (listIterator.hasPrevious()) {
+				CompletedCheckpoint prev = listIterator.previous();
+				if (!prev.getProperties().isSavepoint()) {
+					LOG.info("Found a completed checkpoint ({}) before the latest savepoint, will use it to recover!", prev);
+					return prev;
 				}
-			} catch (Exception e) {
-				LOG.error("Method getAllCheckpoints caused exception : ", e);
-				throw new FlinkRuntimeException(e);
 			}
+			LOG.info("Did not find earlier checkpoint, using latest savepoint to recover.");
 		}
 
-		return candidate;
+		return lastCompleted;
 	}
 
 	/**
--- a/flink-runtime/src/test/java/org/apache/flink/runtime/checkpoint/StandaloneCompletedCheckpointStoreTest.java
+++ b/flink-runtime/src/test/java/org/apache/flink/runtime/checkpoint/StandaloneCompletedCheckpointStoreTest.java
@@ -18,8 +18,10 @@
 
 package org.apache.flink.runtime.checkpoint;
 
+import org.apache.flink.api.common.JobID;
 import org.apache.flink.runtime.jobgraph.JobStatus;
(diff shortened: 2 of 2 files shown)


flink-runtime/src/test/java/org/apache/flink/runtime/checkpoint/StandaloneCompletedCheckpointStoreTest.java
```diff
@@ -116,4 +120,35 @@ public void testAddCheckpointWithFailedRemove() throws Exception {
...
 		}
 	}
+
+	@Test
+	public void preferCheckpointForRecoveryWithSavepoint() throws Exception {
+		StandaloneCompletedCheckpointStore store = new StandaloneCompletedCheckpointStore(5);
+		JobID jobId = new JobID();
+		store.addCheckpoint(checkpoint(jobId, 1L));
+		store.addCheckpoint(checkpoint(jobId, 2L));
+		store.addCheckpoint(checkpoint(jobId, 3L));
```
1u0: The `savepoint` method is not used so far. Maybe it should be used by an additional test like:
```
checkpoint 1, checkpoint 2, savepoint 3 -> returns checkpoint 2
```

flink-runtime/src/main/java/org/apache/flink/runtime/checkpoint/CompletedCheckpointStore.java
```diff
@@ -60,6 +60,10 @@ default CompletedCheckpoint getLatestCheckpoint(boolean isPreferCheckpointForRec
 		}
 
 		CompletedCheckpoint candidate = getAllCheckpoints().get(getAllCheckpoints().size() - 1);
+		if (!candidate.getProperties().isSavepoint()) {
+			return candidate;
+		}
```
1u0: This `if` check can be moved into the next `if` block. This would be more explicit that this check is specific only for `isPreferCheckpointForRecovery = true` and has no effect otherwise.

Additional suggestion: although it's not relevant to your PR and the changes in the ticket, you can assign a variable for `getAllCheckpoints()`  once in the top of the method. This would avoid additional method calls, array copies and should make this method more clearer.

flink-runtime/src/test/java/org/apache/flink/runtime/checkpoint/StandaloneCompletedCheckpointStoreTest.java
```diff
@@ -86,7 +90,7 @@ public void testSuspendDiscardsCheckpoints() throws Exception {
 		assertTrue(checkpoint.isDiscarded());
 		verifyCheckpointDiscarded(taskStates);
 	}
-
```
TisonKun: @zentol don't we have a checkstyle rules prevent trailing whitespace?

gyfora:
Sure I can probably clean this up if a bit. I wanted to the minimal change that makes this correct so I added the explicit return in case the last entry is a checkpoint.

gyfora:
@TisonKun @1u0 I pushed a new commit that addresses the comments + adds more tests

flink-runtime/src/main/java/org/apache/flink/runtime/checkpoint/CompletedCheckpointStore.java
```diff
@@ -55,31 +54,26 @@
...
-					CompletedCheckpoint prev = listIterator.previous();
-					if (!prev.getProperties().isSavepoint()) {
-						candidate = prev;
-						LOG.info("Found a completed checkpoint before the latest savepoint, will use it to recover!");
-						break;
-					}
+		CompletedCheckpoint lastCompleted = allCheckpoints.get(allCheckpoints.size() - 1);
+
+		if (lastCompleted.getProperties().isSavepoint() && isPreferCheckpointForRecovery && allCheckpoints.size() > 1) {
+			ListIterator<CompletedCheckpoint> listIterator = allCheckpoints.listIterator(allCheckpoints.size() - 1);
```
1u0: You can simplify code further: as I've mentioned before, you don't need a special `lastCompleted.getProperties().isSavepoint()` check, instead instantiate the iterator from the very last element `allCheckpoints.listIterator(allCheckpoints.size())`.

The whole method can be structured as:
```
if (allCheckpoints.isEmpty()) {
    return null;
}
if (isPreferCheckpointForRecovery) {
    // try to find the latest checkpoint by reverse traversal over allCheckpoints
    // return it if found
}
// return the latest (be it checkpoint or savepoint)
return allCheckpoints.get(allCheckpoints.size() - 1);
```

gyfora: This way it's easy to give a log message when we fall back to a checkpoint

gyfora:
> Thanks for your updates @gyfora. Tests part looks good to me. I will share my version of `getLatestCheckpoint`. You're the man decide which one to adopt.
> 
> ```java
> 	default CompletedCheckpoint getLatestCheckpoint(boolean isPreferCheckpointForRecovery) throws Exception {
> 		List<CompletedCheckpoint> allCheckpoints = getAllCheckpoints();
> 
> 		if (allCheckpoints.isEmpty()) {
> 			return null;
> 		}
> 
> 		Predicate<Integer> isSavepoint = i -> allCheckpoints.get(i).getProperties().isSavepoint();
> 
> 		int i = allCheckpoints.size() - 1;
> 		while (isPreferCheckpointForRecovery && i > 0 && isSavepoint.test(i)) {
> 			i--;
> 		}
> 
> 		CompletedCheckpoint checkpoint = allCheckpoints.get(i);
> 
> 		LOG.info(
> 			"Use {} ({}) to recover.",
> 			checkpoint.getProperties().isSavepoint() ? "savepoint" : "checkpoint",
> 			checkpoint);
> 
> 		return checkpoint;
> 	}
> ```

I don't think one is much better than the other but I feel that from a readability perspective my proposal is slightly better. It also includes some nice logging to know when we actually preferred a checkpoint or when we fell back to a savepoint anyways.

flink-runtime/src/main/java/org/apache/flink/runtime/checkpoint/CompletedCheckpointStore.java
TisonKun: @gyfora Yes. My last concern is also about logging. In your version given the last completed is a checkpoint we actually log nothing :P

gyfora: I would only log if we don't follow the standard fallback-to-last approach

TisonKun: thanks for your explanation :-)

gyfora:
If no objections I will merge this later today :)

flink-runtime/src/main/java/org/apache/flink/runtime/checkpoint/CompletedCheckpointStore.java
1u0: If you want to log situations when another (non last element) is picked, you could have added additional check by reference that the picked element is not the last one.

If you are going to keep your version, could you please move `lastCompleted.getProperties().isSavepoint()` after `isPreferCheckpointForRecovery`, so it's not evaluated eagerly.

PR #9732, opened 2019-09-20
[FLINK-14153][ml] Add to BLAS a method that performs DenseMatrix and SparseVector multiplication.

xuyang1706:

Previously there is "gemv" method in BLAS that performs multiplications between DenseMatrix and DenseVector. Here we add another one that performs multiplications between DenseMatrix and SparseVector.

## Brief change log

- *Add gemv method to BLAS.*
- *Add BLASTest that provides test cases for BLAS.*

## Verifying this change

- run test case pass

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/linalg/BLAS.java
+++ b/flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/linalg/BLAS.java
@@ -19,6 +19,8 @@
 
 package org.apache.flink.ml.common.linalg;
 
+import org.apache.flink.util.Preconditions;
+
 /**
  * A utility class that provides BLAS routines over matrices and vectors.
  */
@@ -145,5 +147,39 @@ public class BLAS {
 		final int lda = matA.numRows();
 		final String ta = transA ? "T" : "N";
 		NATIVE_BLAS.dgemv(ta, m, n, alpha, matA.getData(), lda, x.getData(), 1, beta, y.getData(), 1);
+	}
+
+	/**
+	 * y := alpha * A * x + beta * y .
+	 */
+	public static void gemv(double alpha, DenseMatrix matA, boolean transA,
+							SparseVector x, double beta, DenseVector y) {
+		if (transA) {
+			Preconditions.checkArgument(matA.numCols() == y.size() && matA.numRows() == x.size(),
+				"Matrix and vector size mismatched.");
+		} else {
+			Preconditions.checkArgument(matA.numRows() == y.size() && matA.numCols() == x.size(),
+				"Matrix and vector size mismatched.");
+		}
+		final int m = matA.numRows();
+		final int n = matA.numCols();
+		if (transA) {
+			int start = 0;
+			for (int i = 0; i < n; i++) {
+				double s = 0.;
+				for (int j = 0; j < x.indices.length; j++) {
+					s += x.values[j] * matA.data[start + x.indices[j]];
+				}
+				y.data[i] = beta * y.data[i] + alpha * s;
+				start += m;
+			}
+		} else {
+			scal(beta, y);
+			for (int i = 0; i < x.indices.length; i++) {
+				int index = x.indices[i];
+				double value = alpha * x.values[i];
+				F2J_BLAS.daxpy(m, value, matA.data, index * m, 1, y.data, 0, 1);
+			}
+		}
 	}
 }
--- a/flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/linalg/BLAS.java
+++ b/flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/linalg/BLAS.java
@@ -29,10 +29,31 @@ public class BLAS {
 	private static final com.github.fommil.netlib.BLAS F2J_BLAS = com.github.fommil.netlib.F2jBLAS.getInstance();
 
 	/**
+	 * \sum_i |x_i| .
+	 */
+	public static double asum(int n, double[] x, int offset) {
+		return F2J_BLAS.dasum(n, x, offset, 1);
+	}
+
+	/**
+	 * \sum_i |x_i| .
+	 */
+	public static double asum(DenseVector x) {
+		return asum(x.data.length, x.data, 0);
+	}
+
+	/**
+	 * \sum_i |x_i| .
+	 */
+	public static double asum(SparseVector x) {
+		return asum(x.values.length, x.values, 0);
+	}
+
+	/**
 	 * y += a * x .
 	 */
 	public static void axpy(double a, double[] x, double[] y) {
-		assert x.length == y.length : "Array dimension mismatched.";
+		Preconditions.checkArgument(x.length == y.length, "Array dimension mismatched.");
 		F2J_BLAS.daxpy(x.length, a, x, 1, y, 1);
 	}
 
@@ -40,7 +61,7 @@ public class BLAS {
 	 * y += a * x .
 	 */
 	public static void axpy(double a, DenseVector x, DenseVector y) {
-		assert x.data.length == y.data.length : "Vector dimension mismatched.";
+		Preconditions.checkArgument(x.data.length == y.data.length, "Vector dimension mismatched.");
 		F2J_BLAS.daxpy(x.data.length, a, x.data, 1, y.data, 1);
 	}
 
@@ -57,24 +78,34 @@ public class BLAS {
 	 * y += a * x .
 	 */
 	public static void axpy(double a, DenseMatrix x, DenseMatrix y) {
-		assert x.m == y.m && x.n == y.n : "Matrix dimension mismatched.";
+		Preconditions.checkArgument(x.m == y.m && x.n == y.n, "Matrix dimension mismatched.");
 		F2J_BLAS.daxpy(x.data.length, a, x.data, 1, y.data, 1);
+	}
+
+	/**
+	 * y[yOffset:yOffset+n] += a * x[xOffset:xOffset+n] .
+	 */
+	public static void axpy(int n, double a, double[] x, int xOffset, double[] y, int yOffset) {
+		F2J_BLAS.daxpy(n, a, x, xOffset, 1, y, yOffset, 1);
 	}
 
 	/**
 	 * x \cdot y .
 	 */
 	public static double dot(double[] x, double[] y) {
-		assert x.length == y.length : "Array dimension mismatched.";
-		return F2J_BLAS.ddot(x.length, x, 1, y, 1);
+		Preconditions.checkArgument(x.length == y.length, "Array dimension mismatched.");
+		double s = 0.;
+		for (int i = 0; i < x.length; i++) {
+			s += x[i] * y[i];
+		}
+		return s;
 	}
 
 	/**
 	 * x \cdot y .
 	 */
 	public static double dot(DenseVector x, DenseVector y) {
-		assert x.data.length == y.data.length : "Vector dimension mismatched.";
-		return F2J_BLAS.ddot(x.data.length, x.data, 1, y.data, 1);
+		return dot(x.getData(), y.getData());
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/linalg/BLAS.java
```diff
@@ -146,4 +148,38 @@ public static void gemv(double alpha, DenseMatrix matA, boolean transA,
 		final String ta = transA ? "T" : "N";
 		NATIVE_BLAS.dgemv(ta, m, n, alpha, matA.getData(), lda, x.getData(), 1, beta, y.getData(), 1);
 	}
+
+	/**
+	 * y := alpha * A * x + beta * y .
+	 */
+	public static void gemv(double alpha, DenseMatrix matA, boolean transA,
+							SparseVector x, double beta, DenseVector y) {
+		if (transA) {
```
walterddr: Can we create a `transposePreconditionChecker`? feel like this would be use multiple place/times with duplicate code in such as `gemv`, `gemm`, etc.

flink-ml-parent/flink-ml-lib/src/test/java/org/apache/flink/ml/common/linalg/BLASTest.java
```diff
@@ -0,0 +1,79 @@
...
+public class BLASTest {
+	private static final double TOL = 1.0e-8;
+	private DenseMatrix mat = new DenseMatrix(2, 3, new double[]{1, 4, 2, 5, 3, 6});
+	private DenseVector dv1 = new DenseVector(new double[]{1, 2});
+	private DenseVector dv2 = new DenseVector(new double[]{1, 2, 3});
+	private SparseVector spv1 = new SparseVector(2, new int[]{0, 1}, new double[]{1, 2});
+	private SparseVector spv2 = new SparseVector(3, new int[]{0, 2}, new double[]{1, 3});
+
+	@Test
+	public void testGemvDense() throws Exception {
```
walterddr: why is the dense version not tested until now?

walterddr: missing validator exception test cases: 
1. invalid dimension
2. invalid dimension after transpose

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/linalg/BLAS.java
```diff
@@ -146,4 +148,38 @@ public static void gemv(double alpha, DenseMatrix matA, boolean transA,
...
+				}
+				y.data[i] = beta * y.data[i] + alpha * s;
+				start += m;
+			}
+		} else {
+			scal(beta, y);
+			for (int i = 0; i < x.indices.length; i++) {
+				int index = x.indices[i];
+				double value = alpha * x.values[i];
+				F2J_BLAS.daxpy(m, value, matA.data, index * m, 1, y.data, 0, 1);
```
walterddr: I am not familiar with BLAS internal performance. is this faster than directly coding it up ?
the reason why I am asking is that: there's two step involved. (`scal` and `daxpy`). may have duplicate mem access ?

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/linalg/BLAS.java
xuyang1706: Yes, you are right. We have created a method "gemvDimensionCheck". However, the dimension check in "gemm" and "gemv" are totally different, so "gemm" is left untouched.

flink-ml-parent/flink-ml-lib/src/test/java/org/apache/flink/ml/common/linalg/BLASTest.java
xuyang1706: We just added more test cases.

xuyang1706: Thanks, we just added the test cases.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/linalg/BLAS.java
xuyang1706: There are two reasons I use BLAS here. 
1. BLAS is a mature linear algebra libarary, which pays a lot of attension to data locality, thus it is more cache friendly than naive implementation. Usually we gain a lot of permformace improvement on level-2/level-3 BLAS routine through calling native (JNI) BLAS, while F2J BLAS is better in level-1 BLAS routines.
2. In the case here, y is first scaled by b, then each columns of Ax is added to y. It is inevitable that y would be visited more than once.

xuyang1706:
> thanks for the contribution @xuyang1706 . looks good to me overall. I left a few comments. please kindly take a look if they make sense.

Thanks @walterddr  for your comments and advices. I have added more test cases and precondition checkers.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/linalg/BLAS.java
walterddr: that makes sense. I am thinking that the assumption is for example BLAS can do some sort of SIMD/MIMD optimization based on the data locality so that it can save register/cache loadings and invalidations. 
It is good to call out the rationale here by having some sort of inline comment:
```
// relying on the native implementation and cache locality from BLAS
```

If there's any performance issue later, we can always avoid duplicate register loading by doing the multiplication the addition and variable assignment at the same line similar to 
```
y.data[i] = beta * y.data[i] + alpha * s;
```

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/linalg/BLAS.java
```diff
@@ -55,24 +78,34 @@ public static void axpy(double a, SparseVector x, DenseVector y) {
...
-		assert x.m == y.m && x.n == y.n : "Matrix dimension mismatched.";
+		Preconditions.checkArgument(x.m == y.m && x.n == y.n, "Matrix dimension mismatched.");
 		F2J_BLAS.daxpy(x.data.length, a, x.data, 1, y.data, 1);
 	}
 
+	/**
+	 * y[yOffset:yOffset+n] += a * x[xOffset:xOffset+n] .
+	 */
+	public static void axpy(int n, double a, double[] x, int xOffset, double[] y, int yOffset) {
+		F2J_BLAS.daxpy(n, a, x, xOffset, 1, y, yOffset, 1);
```
walterddr: A higher level question:
Maybe it is good to clarify when to use `F2J_BLAS` vs `NATIVE_BLAS`. 

I found a very [interesting read](https://issues.apache.org/jira/browse/SPARK-21305) from SPARK's mailing list and seems like there are some considerations regarding this.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/linalg/BLAS.java
```diff
@@ -131,19 +168,56 @@ public static void gemm(double alpha, DenseMatrix matA, boolean transA, DenseMat
...
+	 * y := alpha * A * x + beta * y .
+	 */
+	public static void gemv(double alpha, DenseMatrix matA, boolean transA,
+							DenseVector x, double beta, DenseVector y) {
+		gemvDimensionCheck(matA, transA, x, y);
 		final int m = matA.numRows();
 		final int n = matA.numCols();
 		final int lda = matA.numRows();
 		final String ta = transA ? "T" : "N";
 		NATIVE_BLAS.dgemv(ta, m, n, alpha, matA.getData(), lda, x.getData(), 1, beta, y.getData(), 1);
```
walterddr: any reason why `DenseVector` uses `NATIVE_BLAS` while the `SparseVector` uses `F2J_BLAS`

I think at least on a specific level (0,1,2,3 or up) we should probably only use one specific BLAS version unless specific reason comes up (IMO it should be some very strong justifications)

FYI: I am not sure whether this is related. Some suggestions on [stack](https://stackoverflow.com/questions/41825022/why-spark-blas-use-f2jblas-instead-of-native-blas-for-level-1-routines) shows that there are some performance considerations coming from latest development from the JIT compiler

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/linalg/BLAS.java
xuyang1706: Thanks, I have clarified in the inline doc that we should use F2J_BLAS for level-1 routines and NATIVE_BLAS for level-2 and level-3 routines. This is also the practices adopted by SparkML.

The read from SPARK's mailing list you give here indeed shows the pitfalls when using NATIVE_BLAS. It makes clear that the underlying native BLAS libarary should not use multithreading. Fortunately, the default library uses a BLAS version provided by http://www.netlib.org, which is a single-threaded version.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/linalg/BLAS.java
xuyang1706: We consistently use F2J_BLAS for level-1 routines such as scal/axpy/asum, and use NATIVE_BLAS for level-2/level-3 routines such as gemv, gemm. As for the gemv case here, we use NATIVE_BLAS for the dense case. But for the sparse case, the BLAS library is not directly applicable because it is a library for dense linear algebra. So we implement gemv for SparseVector by hand, using F2J_BLAS to do axpy(level-1 routine) during the course.

xuyang1706:
> Hi @xuyang1706 thanks for the prompt update. I think overall the patch looks good.
> 
> I only have 2 higher level questions regarding the usage of BLAS (which on a level not only relates to this PR but also related to previously committed codes). Please kindly take a look. Thanks -Rong

Thanks for your questions and discuss, @walterddr. It is better to declare our rules for using   `F2J_BLAS` and `NATIVE_BLAS`. We have added clear claims in the JavaDoc. Use F2J_BLAS for level-1 routines and NATIVE_BLAS for level-2 and level-3 routines.   
Regards -Xu

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/linalg/BLAS.java
```diff
@@ -19,26 +19,53 @@
...
 /**
  * A utility class that provides BLAS routines over matrices and vectors.
  */
 public class BLAS {
+
+	// For level-1 routines, we use Java implementation.
 	private static final com.github.fommil.netlib.BLAS NATIVE_BLAS = com.github.fommil.netlib.BLAS.getInstance();
+
+	// For level-2 and level-3 routines, we use the native BLAS.
 	private static final com.github.fommil.netlib.BLAS F2J_BLAS = com.github.fommil.netlib.F2jBLAS.getInstance();
```
walterddr: comment and code doesn't align. 
* the field name is `NATIVE_BLAS` for the level-1, 
* the field name is `F2J_BLAS` for level-2,3. 
any naming conversion problem?

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/linalg/BLAS.java
```diff
@@ -131,19 +172,56 @@ public static void gemm(double alpha, DenseMatrix matA, boolean transA, DenseMat
...
+				}
+				y.data[i] = beta * y.data[i] + alpha * s;
+				start += m;
+			}
+		} else {
+			scal(beta, y);
+			for (int i = 0; i < x.indices.length; i++) {
+				int index = x.indices[i];
+				double value = alpha * x.values[i];
+				F2J_BLAS.daxpy(m, value, matA.data, index * m, 1, y.data, 0, 1);
```
walterddr: could you add the explanation you added in this PR comment into the actual code comments? I think it helps others to understand this code in the future.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/linalg/BLAS.java
xuyang1706: Thanks, we have corrected the java doc.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/linalg/BLAS.java
xuyang1706: Thanks, we did it.

walterddr:
@flinkbot run travis

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/linalg/BLAS.java
```diff
@@ -19,26 +19,59 @@
...
+
 /**
  * A utility class that provides BLAS routines over matrices and vectors.
  */
 public class BLAS {
+
+	// For level-1 routines, we use Java implementation.
 	private static final com.github.fommil.netlib.BLAS NATIVE_BLAS = com.github.fommil.netlib.BLAS.getInstance();
+
+	// For level-2 and level-3 routines, we use the native BLAS.
```
walterddr: nit: use `/* */`

xuyang1706: Yes, please refine it when merging. Thanks.

xuyang1706:
> Thanks for the quick reply/update @xuyang1706 . I think the patch looks good overall.
> I will made some minor refinement and merge it soon.

Thanks for your kindly help @walterddr.

walterddr:
@flinkbot run travis

PR #9733, opened 2019-09-20
[FLINK-14154][ml] Add the class for multivariate Gaussian Distribution.

xuyang1706:

Multivariate Gaussian Distribution is a generalization of the one-dimensional (univariate) normal distribution to higher dimensions. In machine learning, multivariate Gaussian Distribution finds usages in algorithms link Gaussian Mixture Model.

## Brief change log

- *Add MultivariateGaussian that provides basic functionality for a Multivariate Gaussian (Normal) Distribution.*
- *Add MultivariateGaussianTest that provides test cases for MultivariateGaussian.*

## Verifying this change

- run test case pass

  - The serializers: (no)

## Documentation

### DIFF
--- a/flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/statistics/basicstatistic/MultivariateGaussian.java
+++ b/flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/statistics/basicstatistic/MultivariateGaussian.java
@@ -94,8 +94,18 @@ public class MultivariateGaussian {
 	/**
 	 * Compute distribution dependent constants.
 	 *
-	 * <p>rootSigmaInv = D^(-1/2)^ * U.t, where sigma = U * D * U.t .
-	 * u = log((2*pi)^(-k/2)^ * det(sigma)^(-1/2)^) .
+	 * <p>The probability density function is calculated as:
+	 * pdf(x) = (2*pi)^(-k/2)^ * det(sigma)^(-1/2)^ * exp((-1/2) * (x-mu).t * inv(sigma) * (x-mu))
+	 *
+	 * <p>Here we compute the following distribution dependent constants that can be reused in each pdf computation:
+	 * A) u = log((2*pi)^(-k/2)^ * det(sigma)^(-1/2)^)
+	 * B) rootSigmaInv = sqrt(inv(sigma)) = D^(-1/2)^ * U.t
+	 *
+	 * <p><ul>
+	 * <li> sigma = U * D * U.t
+	 * <li> inv(sigma) = U * inv(D) * U.t = (D^{-1/2}^ * U.t).t * (D^{-1/2}^ * U.t)
+	 * <li> sqrt(inv(sigma)) = D^(-1/2)^ * U.t
+	 * </ul></p>
 	 */
 	private void calculateCovarianceConstants() {
 		int n = this.mean.size();
--- a/flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/statistics/basicstatistic/MultivariateGaussian.java
+++ b/flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/statistics/basicstatistic/MultivariateGaussian.java
@@ -17,7 +17,7 @@
  * under the License.
  */
 
-package org.apache.flink.ml.statistics.basicstatistic;
+package org.apache.flink.ml.common.statistics.basicstatistic;
 
 import org.apache.flink.ml.common.linalg.BLAS;
 import org.apache.flink.ml.common.linalg.DenseMatrix;
--- a/flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/statistics/basicstatistic/MultivariateGaussian.java
+++ b/flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/statistics/basicstatistic/MultivariateGaussian.java
@@ -25,6 +25,8 @@ import org.apache.flink.ml.common.linalg.DenseVector;
 import org.apache.flink.ml.common.linalg.SparseVector;
 import org.apache.flink.ml.common.linalg.Vector;
 
+import org.apache.flink.shaded.guava18.com.google.common.primitives.Doubles;
+
 import com.github.fommil.netlib.LAPACK;
 import org.netlib.util.intW;
 
@@ -34,6 +36,7 @@ import org.netlib.util.intW;
 public class MultivariateGaussian {
 
 	private static final LAPACK LAPACK_INST = LAPACK.getInstance();
+	private static final com.github.fommil.netlib.BLAS F2J_BLAS_INST = com.github.fommil.netlib.F2jBLAS.getInstance();
 	private static final double EPSILON;
 
 	static {
@@ -69,14 +72,14 @@ public class MultivariateGaussian {
 	}
 
 	/**
-	 * Returns density of this multivariate Gaussian at given point, x.
+	 * Returns density of this multivariate Gaussian at given point x .
 	 */
 	public double pdf(Vector x) {
 		return Math.exp(logpdf(x));
 	}
 
 	/**
-	 * Returns the log-density of this multivariate Gaussian at given point, x.
+	 * Returns the log-density of this multivariate Gaussian at given point x .
 	 */
 	public double logpdf(Vector x) {
 		int n = mean.size();
@@ -87,7 +90,7 @@ public class MultivariateGaussian {
 		} else if (x instanceof SparseVector) {
 			BLAS.axpy(1., (SparseVector) x, delta);
 		}
-		BLAS.gemv(1.0, rootSigmaInv, false, delta, 0., v);
+		BLAS.gemv(1.0, rootSigmaInv, true, delta, 0., v);
 		return u - 0.5 * BLAS.dot(v, v);
 	}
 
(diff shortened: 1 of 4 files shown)


flink-ml-parent/flink-ml-lib/src/test/java/org/apache/flink/ml/common/statistics/basicstatistic/MultivariateGaussianTest.java
```diff
@@ -0,0 +1,57 @@
...
+import org.apache.flink.ml.common.linalg.DenseVector;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+/**
+ * Test cases for MultivariateGaussian.
+ */
+public class MultivariateGaussianTest {
+	private static final double TOL = 1.0e-5;
```
walterddr: Is there a reason why we use a fixed `TOL` while during the actual execution we use the limited bi-section to determine the actual TOL for the system?

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/statistics/basicstatistic/MultivariateGaussian.java
```diff
@@ -0,0 +1,138 @@
...
+			BLAS.axpy(1., (SparseVector) x, delta);
+		}
+		BLAS.gemv(1.0, rootSigmaInv, false, delta, 0., v);
+		return u - 0.5 * BLAS.dot(v, v);
+	}
+
+	/**
+	 * Compute distribution dependent constants.
+	 *
+	 * <p>rootSigmaInv = D^(-1/2)^ * U.t, where sigma = U * D * U.t .
```
walterddr: much more informatively: 
```
 * <p><ul>
 * <li> sigma = U * D * U.t
 * <li> inv(Sigma) = U * inv(D) * U.t = = (D^{-1/2}^ * U.t).t * (D^{-1/2}^ * U.t)
 * <li> sqrt(inv(Sigma)) = D^(-1/2)^ * U.t
 * </ul></p>
```

since the word `rootSigmaInv` is not intuitively understandable to me. (especially "root" doesn't mean square-root)

flink-ml-parent/flink-ml-lib/src/test/java/org/apache/flink/ml/common/statistics/basicstatistic/MultivariateGaussianTest.java
xuyang1706: The limited bi-section is used to compute the threshold in determing whether the singular value of the covariance matrix is larger than zero. It is a way to improve numerical stability when the covariance matrix is singular or nearly singular. The "TOL" in the unit test cases is the tolerance between the computed pdf and theoretical pdf. We could only  achieve the precision at around 1*e-5, because det(sigma) is computed with LAPACK's "dsyev", which uses iterative algorithm to compute eigen values, thus have inherent errors.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/statistics/basicstatistic/MultivariateGaussian.java
xuyang1706: Thanks, we refined the JavaDoc according to your advice.

xuyang1706:
> thanks for the contribution @xuyang1706 . I will go ahead and verify the algorithm once #9732 has been merged and rebased. on a higher level I left a few comments on the details of the javadoc. please kindly take a look.

Thanks for your comments and advice, @walterddr . We refined the JavaDoc according to your advice.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/statistics/basicstatistic/MultivariateGaussian.java
```diff
@@ -0,0 +1,148 @@
...
+			eps /= 2.0;
+		}
+		EPSILON = eps;
+	}
+
+	private final DenseVector mean;
+	private final DenseMatrix cov;
+
+	private DenseMatrix rootSigmaInv;
+	private double u;
```
walterddr: members can be `final`?

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/statistics/basicstatistic/MultivariateGaussian.java
```diff
@@ -0,0 +1,148 @@
...
+		int n = this.mean.size();
+		int lwork = 3 * n - 1;
+		double[] matA = new double[n * n];
+		double[] work = new double[lwork];
+		double[] evs = new double[n];
+		intW info = new intW(0);
+
+		for (int i = 0; i < n; i++) {
+			System.arraycopy(cov.getData(), i * n, matA, i * n, i + 1);
+		}
```
walterddr: `System.arraycopy(cov.getData(), 0, matA, 0, n*n);`?

flink-ml-parent/flink-ml-lib/src/test/java/org/apache/flink/ml/common/statistics/basicstatistic/MultivariateGaussianTest.java
```diff
@@ -0,0 +1,57 @@
...
+
```
walterddr: add test for `Univariate` as well just in case?

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/statistics/basicstatistic/MultivariateGaussian.java
```diff
@@ -0,0 +1,148 @@
...
+			if (ev > tol) {
+				logPseudoDetSigma += Math.log(ev);
+			}
+		}
+
+		this.rootSigmaInv = new DenseMatrix(n, n);
+		for (int i = 0; i < n; i++) {
+			double pinvS = evs[i] > tol ? Math.sqrt(1.0 / evs[i]) : 0.;
+			for (int j = 0; j < n; j++) {
+				this.rootSigmaInv.set(i, j, pinvS * matA[j * n + i]);
```
walterddr: is there a breeze routine that can speed this up. this is basically a scale up from a pseudo inverse of one element of the eigenvector rigth?

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/statistics/basicstatistic/MultivariateGaussian.java
```diff
@@ -0,0 +1,148 @@
...
+		// log(pseudo-determinant) is sum of the logs of all non-zero singular values
+		double logPseudoDetSigma = 0.;
+		for (double ev : evs) {
+			if (ev > tol) {
+				logPseudoDetSigma += Math.log(ev);
+			}
+		}
+
+		this.rootSigmaInv = new DenseMatrix(n, n);
+		for (int i = 0; i < n; i++) {
```
walterddr: nit: this seems to be able to abstract into a subroutine in `BLAS` ? (I am not sure whether this is useful but good to check?

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/statistics/basicstatistic/MultivariateGaussian.java
xuyang1706: Final variables have to be initialized in the constructor. I kind of prefer initializing them in the subroutine calculateCovarianceConstants for readability reasons.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/statistics/basicstatistic/MultivariateGaussian.java
xuyang1706: Yes, your are right. I have made the change according to your suggestion.

flink-ml-parent/flink-ml-lib/src/test/java/org/apache/flink/ml/common/statistics/basicstatistic/MultivariateGaussianTest.java
xuyang1706: Thanks. I have added the univariate test case.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/statistics/basicstatistic/MultivariateGaussian.java
xuyang1706: Yes, you are right. Here we are just scaling the eigen vectors. Now I use BLAS's scal routine to do this for me. More over, by switching rootSigmaInv from D^(-1/2)^ * U.t to U * D^(-1/2)^, we can reuse the data buffer of U.

flink-ml-parent/flink-ml-lib/src/main/java/org/apache/flink/ml/common/statistics/basicstatistic/MultivariateGaussian.java
xuyang1706: I think it could be abstracted as a full matrix multiplies a diagonal matrix. Perhaps we can add to BLAS in another PR.

xuyang1706:
> Thanks for the contribution @xuyang1706 overall the approach looks good to me. I have some questions/comments regarding the detail implementation of the compute. please kindly take a look and see if they make sense? --Rong.

Thanks for your comments and advices @walterddr . I have refined the code according to your advices.  Thanks, --Xu

PR #9735, opened 2019-09-20
[FLINK-14156][runtime] Submit timer trigger letters to task's mailbox with operator's precedence

1u0:

This change submits `StreamTask`'s timer trigger mailbox action with the priority of the operator that the trigger originate from. This should allow chaining multiple operators that use `mailboxExecutor.yield()` internally.

## Brief change log

  - set once and access `ProcessingTimeService` instance at operator level (a new field in `AbstractStreamOperator`);
  - instantiate `ProcessingTimeService` instance with mailbox executor using operator precedence .

## Verifying this change

All existing tests should pass. This change needs to add more tests.

## Documentation

### DIFF
--- a/flink-state-backends/flink-statebackend-rocksdb/src/test/java/org/apache/flink/contrib/streaming/state/RocksDBAsyncSnapshotTest.java
+++ b/flink-state-backends/flink-statebackend-rocksdb/src/test/java/org/apache/flink/contrib/streaming/state/RocksDBAsyncSnapshotTest.java
@@ -69,7 +69,6 @@ import org.apache.flink.streaming.runtime.streamrecord.StreamRecord;
 import org.apache.flink.streaming.runtime.tasks.OneInputStreamTask;
 import org.apache.flink.streaming.runtime.tasks.OneInputStreamTaskTestHarness;
 import org.apache.flink.streaming.runtime.tasks.StreamMockEnvironment;
-import org.apache.flink.streaming.runtime.tasks.StreamTask;
 import org.apache.flink.util.IOUtils;
 import org.apache.flink.util.TestLogger;
 
@@ -85,7 +84,6 @@ import javax.annotation.Nullable;
 
 import java.io.File;
 import java.io.IOException;
-import java.lang.reflect.Field;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.Map;
@@ -216,18 +214,9 @@ public class RocksDBAsyncSnapshotTest extends TestLogger {
 		AtomicReference<Throwable> errorRef = new AtomicReference<>();
 		mockEnv.setExternalExceptionHandler(errorRef::set);
 		testHarness.invoke(mockEnv);
+		testHarness.waitForTaskRunning();
 
 		final OneInputStreamTask<String, String> task = testHarness.getTask();
-
-		// wait for the task to be running
-		for (Field field: StreamTask.class.getDeclaredFields()) {
-			if (field.getName().equals("isRunning")) {
-				field.setAccessible(true);
-				while (!field.getBoolean(task)) {
-					Thread.sleep(10);
-				}
-			}
-		}
 
 		task.triggerCheckpointAsync(new CheckpointMetaData(42, 17), CheckpointOptions.forCheckpointWithDefaultLocation(), false)
 			.get();
@@ -331,18 +320,9 @@ public class RocksDBAsyncSnapshotTest extends TestLogger {
 		blockerCheckpointStreamFactory.setWaiterLatch(new OneShotLatch());
 
 		testHarness.invoke(mockEnv);
+		testHarness.waitForTaskRunning();
 
 		final OneInputStreamTask<String, String> task = testHarness.getTask();
-
-		// wait for the task to be running
-		for (Field field: StreamTask.class.getDeclaredFields()) {
-			if (field.getName().equals("isRunning")) {
-				field.setAccessible(true);
-				while (!field.getBoolean(task)) {
-					Thread.sleep(10);
-				}
-			}
-		}
 
 		task.triggerCheckpointAsync(
 			new CheckpointMetaData(42, 17),
--- a/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SystemProcessingTimeService.java
+++ b/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SystemProcessingTimeService.java
@@ -38,11 +38,11 @@ import java.util.concurrent.atomic.AtomicInteger;
 import static org.apache.flink.util.Preconditions.checkNotNull;
 
 /**
- * A {@link ProcessingTimeService} which assigns as current processing time the result of calling
+ * A {@link TimerService} which assigns as current processing time the result of calling
  * {@link System#currentTimeMillis()} and registers timers using a {@link ScheduledThreadPoolExecutor}.
  */
 @Internal
-public class SystemProcessingTimeService extends ProcessingTimeService {
+public class SystemProcessingTimeService implements TimerService {
 
 	private static final Logger LOG = LoggerFactory.getLogger(SystemProcessingTimeService.class);
 
--- a/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SystemProcessingTimeService.java
+++ b/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SystemProcessingTimeService.java
@@ -190,8 +190,17 @@ public class SystemProcessingTimeService extends ProcessingTimeService {
 		}
 	}
 
-	@Override
-	public boolean shutdownAndAwaitPending(long time, TimeUnit timeUnit) throws InterruptedException {
+	/**
+	 * Shuts down and clean up the timer service provider hard and immediately. This does wait
+	 * for all timers to complete or until the time limit is exceeded. Any call to
+	 * {@link #registerTimer(long, ProcessingTimeCallback)} will result in a hard exception after calling this method.
+	 * @param time time to wait for termination.
+	 * @param timeUnit time unit of parameter time.
+	 * @return {@code true} if this timer service and all pending timers are terminated and
+	 *         {@code false} if the timeout elapsed before this happened.
+	 */
+	@VisibleForTesting
+	boolean shutdownAndAwaitPending(long time, TimeUnit timeUnit) throws InterruptedException {
 		shutdownService();
 		return timerService.awaitTermination(time, timeUnit);
 	}
--- a/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SystemProcessingTimeService.java
+++ b/flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SystemProcessingTimeService.java
@@ -55,17 +55,17 @@ public class SystemProcessingTimeService implements TimerService {
 	/** The executor service that schedules and calls the triggers of this task. */
 	private final ScheduledThreadPoolExecutor timerService;
 
-	private final ScheduledCallbackExecutionContext callbackExecutionContext;
+	private final ExceptionHandler exceptionHandler;
 	private final AtomicInteger status;
 
 	@VisibleForTesting
-	SystemProcessingTimeService(ScheduledCallbackExecutionContext callbackExecutionContext) {
-		this(callbackExecutionContext, null);
+	SystemProcessingTimeService(ExceptionHandler exceptionHandler) {
+		this(exceptionHandler, null);
 	}
 
-	SystemProcessingTimeService(ScheduledCallbackExecutionContext callbackExecutionContext, ThreadFactory threadFactory) {
+	SystemProcessingTimeService(ExceptionHandler exceptionHandler, ThreadFactory threadFactory) {
 
-		this.callbackExecutionContext = checkNotNull(callbackExecutionContext);
+		this.exceptionHandler = checkNotNull(exceptionHandler);
 		this.status = new AtomicInteger(STATUS_ALIVE);
 
 		if (threadFactory == null) {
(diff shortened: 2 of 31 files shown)


flink-streaming-java/src/test/java/org/apache/flink/streaming/util/MockStreamTask.java
```diff
@@ -75,7 +75,7 @@ public MockStreamTask(
 		this.closableRegistry = closableRegistry;
 		this.streamStatusMaintainer = streamStatusMaintainer;
 		this.checkpointStorage = checkpointStorage;
-		this.processingTimeService = processingTimeService;
+		this.processingTimeService = timerService;
```
AHeise: rename field?

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/operators/StreamTaskOperatorTimerTest.java
```diff
@@ -0,0 +1,144 @@
...
+			processingTimeService
+				.registerTimer(
+					processingTimeService.getCurrentProcessingTime() + timeout,
+					timestamp -> {
+						events.add("Timer:" + chainIndex + ":" + count);
+						--count;
+					});
+
+			++count;
+			output.collect(element);
```
AHeise: The test is relying on collect being blocking until the downstream operators (or else yield might be called in the first operator first). Tbh I'm not sure if this a given or an implementation detail. So please double-check if that assumption always holds.

In any case I'd add a comment here. The test seemed rather random in the beginning, but it got much clearer after I realized this blocking behavior.

flink-state-backends/flink-statebackend-rocksdb/src/test/java/org/apache/flink/contrib/streaming/state/RocksDBAsyncSnapshotTest.java
```diff
@@ -331,19 +320,10 @@ public CheckpointStateOutputStream createCheckpointStateOutputStream(Checkpointe
 		blockerCheckpointStreamFactory.setWaiterLatch(new OneShotLatch());
 
 		testHarness.invoke(mockEnv);
+		testHarness.waitForTaskRunning();
 
 		final OneInputStreamTask<String, String> task = testHarness.getTask();
 
-		// wait for the task to be running
```
AHeise: omg. thanks for that hotfix.

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/operators/StreamTaskOperatorTimerTest.java
```diff
@@ -0,0 +1,144 @@
...
+			ProcessingTimeService processingTimeService = getProcessingTimeService();
+			processingTimeService
+				.registerTimer(
+					processingTimeService.getCurrentProcessingTime() + timeout,
+					timestamp -> {
+						events.add("Timer:" + chainIndex + ":" + count);
+						--count;
+					});
+
+			++count;
```
AHeise: Isn't count always 1?

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/operators/StreamTaskOperatorTimerTest.java
```diff
@@ -0,0 +1,144 @@
...
+			output.collect(element);
+
+			while (count > 0) {
```
AHeise: How about processing as many mailbox as possible? `while(tryYield){}`

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/operators/StreamTaskOperatorTimerTest.java
```diff
@@ -0,0 +1,144 @@
...
+import static org.hamcrest.MatcherAssert.assertThat;
+
+/**
+ * Test to verify that timer triggers are run according to operator precedence (combined with yield() at operator level).
+ */
+public class StreamTaskOperatorTimerTest extends TestLogger {
+	private static List<String> events;
+
+	@Test
+	public void testOperatorYieldExecutesSelectedTimers() throws Exception {
```
AHeise: I would love to see another test case, where AWOP1.yield will execute AWOP2.timer.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/TimerService.java
```diff
@@ -0,0 +1,77 @@
...
+ *     <li>After calling {@link #quiesce()}, further calls to
+ *         {@link #registerTimer(long, ProcessingTimeCallback)} will not register any further timers, and will
+ *         return a "dummy" future as a result. This is used for clean shutdown, where currently firing
+ *         timers are waited for and no future timers can be scheduled, without causing hard exceptions.</li>
+ *     <li>After a call to {@link #shutdownService()}, all calls to {@link #registerTimer(long, ProcessingTimeCallback)}
+ *         will result in a hard exception.</li>
+ * </ol>
+ */
+@Internal
+public interface TimerService extends ProcessingTimeService {
```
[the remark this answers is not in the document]
1u0: Although, the commit message says that the lifecycle methods were extracted into a separate interface, my main motivation was to separate "application level api" (`ProcessingTimeService` that exposed to operators) from the system timer provider (implementation of `TimerService`).
It just happens that currently in this PR `TimerService` piggybacks on `ProcessingTimeService`
interface (and `ProcessingTimeCallback`) to define the system timer api.

I can introduce a new, separate timer api for `TimerService` to explicitly separate them. It should not make big changes in the runtime code, only some tests should be affected to be more explicit what they are testing against. I'd like to hear from others if you are ok with this direction.

Regarding the naming, `ProcessingTimeService` is simply the preserved name that operators see and `TimerService` is a common name in Flink code base for classes with similar functionality.

Introducing a common "life-cycle" interface may be too early generalization on my side.
Also, unifying/consolidating the `TimerService` classes may be a separate effort on it's own.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SystemProcessingTimeService.java
```diff
@@ -250,37 +250,36 @@ int getNumTasksScheduled() {
 	// ------------------------------------------------------------------------
 
 	/**
-	 * A context to which {@link ProcessingTimeCallback} would be passed to be invoked when a timer is up.
+	 * An exception handler, called when {@link ProcessingTimeCallback} throws an exception.
 	 */
-	interface ScheduledCallbackExecutionContext {
-
-		void invoke(ProcessingTimeCallback callback, long timestamp);
+	interface ExceptionHandler {
```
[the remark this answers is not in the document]
1u0: I would not use `org.apache.flink.runtime.operators.sort.ExceptionHandler` or `AsyncExceptionHandler` here, as they are both used in their own use case (and were probably introduced on their own to have separation). Reusing them here would be a false code reuse, imo.

I see a value to have a generic interface defined somewhere in `org.apache.flink.util` (in `flink-core`?).
But even for the generic interface, the question is what should be the api (currently, all three mentioned interfaces have some small different details).

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/operators/StreamTaskOperatorTimerTest.java
```diff
@@ -0,0 +1,144 @@
...
+import java.util.List;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+/**
+ * Test to verify that timer triggers are run according to operator precedence (combined with yield() at operator level).
+ */
+public class StreamTaskOperatorTimerTest extends TestLogger {
+	private static List<String> events;
```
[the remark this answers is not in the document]
1u0: I don't like usage of static (mutable) variables myself (including in tests), but I don't see other way how to get "feedback" from the test operators here.
I'm open if you have a suggestion how to avoid it here.

The mailbox is in a different situation, as it's provided by normal (production) code by the runtime.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/TimerService.java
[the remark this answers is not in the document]
pnowojski: I'm not sure if naming this interface `TimerService` is a good idea. Having 4 different interfaces/classes with the same name is pretty confusing. I would prefer to prefix them like `CEPTimerService`.

@AHeise. Regarding the extraction of the lifecycle methods to a `Lifecycle` interface and implementing it in the multiple independent places. If those different classes are are not inter-exchangeably used, so the common base `Lifecycle` interface is used only to make sure that the naming convention remains consistent, I would be strongly against that. Recently I've introduce this with the `AsyncDataInput` that was implemented by couple of independent classes, and honestly that was a mistake. Whenever I was trying to look for usages of the `SingleInputGate#pollNext()` method, IntelliJ was pointing me to tons of the unrelated calls to for example `StreamTaskInput#pollNext()`. (Currently the situation has improved in the code as we introduced `PullingAsyncDataInput` and `PushingAsyncDataInput`.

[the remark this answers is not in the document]
AHeise: Extracting Lifecycle methods is purely optional. @pnowojski you can make it rather small like https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/context/Lifecycle.html or https://logging.apache.org/log4j/2.x/log4j-core/apidocs/org/apache/logging/log4j/core/LifeCycle.html and add specific stuff (`quiensce`) only to specific implementations. But like @1u0 said, it may be premature at this point in time. If we manage to pull life-cycle methods from user-facing interfaces, we can do that later.

Anyways, I'd still like to address the naming. My main source of confusion is the difference between `TimeService` and `TimerService`.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/TimerService.java
```diff
@@ -35,43 +36,43 @@
 @Internal
 public interface TimerService extends ProcessingTimeService {
 
-    /**
-     * Returns <tt>true</tt> if the service has been shut down, <tt>false</tt> otherwise.
-     */
-    boolean isTerminated();
+	/**
```
AHeise: Nit: formatting changed from first commit. This file should not have any changes in this commit.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/SystemProcessingTimeService.java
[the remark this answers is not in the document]
AHeise: 👍

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/operators/StreamTaskOperatorTimerTest.java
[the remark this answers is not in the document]
AHeise: Just create a local list in each test method and pass it in the factory and the operator.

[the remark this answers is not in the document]
AHeise: Ah yes, I thought the test harness is taking a shortcut there. How about emitting the result through output?

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/operators/StreamTaskOperatorTimerTest.java
AHeise: Hm, can we somehow make it deterministic?
Adding an element to first operator, that is forwarded to second operator where it registers the timer.
Then adding another element that does only yield to downstream as long as the timer has not fired.

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/operators/StreamTaskOperatorTimerTest.java
AHeise: Yes, I guess that's a fair assumption and if it is indeed changed somewhere in the future, then we will just adjust the test.

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/operators/StreamTaskOperatorTimerTest.java
AHeise: But how would you register two timers? Control of processElement is never returned unless `count == 0`.

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/operators/StreamTaskOperatorTimerTest.java
AHeise: Yes, that's good. I'm probably still distracted by the count variable. Imho, it more or less acts as a latch which blocks until the only timer (in this operator) has been processed.

flink-streaming-java/src/main/java/org/apache/flink/streaming/runtime/tasks/TimerService.java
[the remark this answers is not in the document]
AHeise: Okay @1u0 convinced my that `ProcessingTime` is well-known name on its own, so that `ProcessingTimeService` makes sense. Similarly, `TimerService` on its own makes sense.
So even though it seems inconsistent, I would keep it as is in this PR.

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/operators/StreamTaskOperatorTimerTest.java
AHeise: Okay, @1u0 discussed the test setup offline and I came to the conclusion that it is already testing what it should. So I'd retract my comment.

flink-streaming-java/src/test/java/org/apache/flink/streaming/runtime/operators/StreamTaskOperatorTimerTest.java
```diff
@@ -0,0 +1,144 @@
...
+
+		TestOperator(int chainIndex, MailboxExecutor mailboxExecutor) {
+			this.chainIndex = chainIndex;
+			this.mailboxExecutor = mailboxExecutor;
+		}
+
+		@Override
+		public void processElement(StreamRecord<T> element) throws Exception {
+			long timeout = (chainIndex + 1) * 1000L + count * 100L;
+			ProcessingTimeService processingTimeService = getProcessingTimeService();
```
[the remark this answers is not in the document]
AHeise: Okay, @1u0 discussed the test setup offline and I came to the conclusion that it is already testing what it should. So I'd retract my comment.

PR #9742, opened 2019-09-23
[FLINK-14076] Ensure CheckpointException can always be deserialized on JobManager

jdm2212:

JobManager would previously fail to deserialize a DeclineCheckpoint if it contained a CheckpointException with a root cause from user-defined code. For example, if a `org.apache.apache.kafka.KafkaException` caused a checkpoint to fail, it would get wrapped in a CheckpointException. The KafkaException is not on the JM classpath, so the (wrapped) KafkaException will cause deserialization to fail on the JM. Now CheckpointException always wraps its cause to ensure deserialization works.

## Brief change log
  - JM is guaranteed to be able to deserialize DeclineCheckpoint from TM

## Verifying this change

  - The serializers: don't know

## Documentation

### DIFF
--- a/flink-core/src/main/java/org/apache/flink/util/ExceptionUtils.java
+++ b/flink-core/src/main/java/org/apache/flink/util/ExceptionUtils.java
@@ -315,6 +315,38 @@ public final class ExceptionUtils {
 	}
 
 	/**
+	 * Checks whether a throwable chain contains a specific type of exception and returns it. It deserializes
+	 * any {@link SerializedThrowable} that are found using the provided {@link ClassLoader}.
+	 *
+	 * @param throwable the throwable chain to check.
+	 * @param searchType the type of exception to search for in the chain.
+	 * @param classLoader to use for deserialization.
+	 * @return Optional throwable of the requested type if available, otherwise empty
+	 */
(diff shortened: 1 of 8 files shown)


tillrohrmann:
@jdm2212 thanks for updating the PR. I'll mark it as ready for review then.

flink-core/src/main/java/org/apache/flink/util/ExceptionUtils.java
```diff
@@ -314,6 +314,37 @@ else if (t instanceof Error) {
...
+		if (throwable == null || searchType == null) {
+			return Optional.empty();
+		}
+
+		Throwable t = throwable;
+		while (t != null) {
+			if (searchType.isAssignableFrom(t.getClass())) {
+				return Optional.of(searchType.cast(t));
+			} else if (t.getClass().isAssignableFrom(SerializedThrowable.class)) {
+				Throwable next = ((SerializedThrowable) t).deserializeError(ClassLoader.getSystemClassLoader());
```
tillrohrmann: I would pass in the class loader as a parameter because one might wanna use a user code class loader here.

jdm2212:
Thanks!

PR #9744, opened 2019-09-23
[FLINK-13515][test] Fix ClassLoaderITCase fails on Java 11

sunhaibotb:

This pull request fixes the failure of `ClassLoaderITCase` on Java 11. For a nested class instance `x`, `x.getCanonicalName()` finds and loads the enclosing class of `x` by calling the methods of the classloader where `x` resides. Thus, it is incorrect to use `candidate.getClass().getCanonicalName()` to test whether there is `SuccessException` in the classpath of the class loader where the `ClassLoaderITCase` class resides, because `candidate.getClass().getCanonicalName()` will not find and load the related classes in the classloader where the `ClassLoaderITCase` class resides. We should use `Class.forName(...)` to test the expected result.

In addition, when calling `getCanonicalName()` , `NoClassDefFoundError` is reported in Java 8, not in Java 11,  for the following reason:

In Java 8, a class named `CheckpointedStreamingProgram$1` is generated at compiling ( because there are nested classes) , and when the nested classes are loaded, the dependent class loaded is `CheckpointedStreamingProgram$1` rather than `CheckpointedStreamingProgram`. Therefore, when `getCanonicalName()` is called in `ClassLoaderITCase`, the `CheckpointedStreamingProgram` class needs to be loaded, but at this time the `ChildFirstClassloader` (an `URLClassloader`) is closed and cannot load it, so `NoClassDefFoundError` is reported. But in Java 11, the `CheckpointedStreamingProgram$1` class is not generated and the `CheckpointedStreamingProgram` class is used directly (that is the `CheckpointedStreamingProgram` class has been loaded into `ChildFirstClassloader` before closing `ChildFirstClassloader`), so no errors are reported.

## Brief change log

  - Change the expected detection logic in `ClassLoaderITCase#testCheckpointedStreamingClassloaderJobWithCustomClassLoader`

## Verifying this change

This change is already covered by existing test.

## Documentation

### DIFF
--- a/flink-tests/src/test/java/org/apache/flink/test/classloading/ClassLoaderITCase.java
+++ b/flink-tests/src/test/java/org/apache/flink/test/classloading/ClassLoaderITCase.java
@@ -142,7 +142,7 @@ public class ClassLoaderITCase extends TestLogger {
 	}
 
 	@Test
-	public void testCustomSplitJobWithCustomClassLoaderJar() throws IOException, ProgramInvocationException {
+	public void testCustomSplitJobWithCustomClassLoaderJar() throws ProgramInvocationException {
 
 		PackagedProgram inputSplitTestProg = new PackagedProgram(new File(INPUT_SPLITS_PROG_JAR_FILE));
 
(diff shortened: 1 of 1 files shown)


sunhaibotb:
Please review this PR for me, thanks @zentol .

sunhaibotb:
Can you review this PR for me @tillrohrmann ?  Thanks.

PR #9745, opened 2019-09-23
[FLINK-14070] [runtime] Use TimeUtils to parse duration configs in flink-runtime

zhuzhurk:

*TimeUtils now can parse all duration configs supported by scala FiniteDuration.
And we'd like to use it to replace scala Duration for duration config parsing.*
*This is one step to make flink-runtime scala free.*

## Brief change log

  - *The duration configs in flink-runtime are now parsed using TimeUtils#parseDuration*

## Verifying this change

This change is already covered by existing tests of changed components.

## Documentation

### DIFF
--- a/flink-runtime/src/main/scala/org/apache/flink/runtime/akka/AkkaUtils.scala
+++ b/flink-runtime/src/main/scala/org/apache/flink/runtime/akka/AkkaUtils.scala
@@ -729,7 +729,7 @@ object AkkaUtils {
   }
 
   def getTimeout(config: Configuration): time.Duration = {
-    return TimeUtils.parseDuration(config.getString(AkkaOptions.ASK_TIMEOUT))
+    TimeUtils.parseDuration(config.getString(AkkaOptions.ASK_TIMEOUT))
   }
 
   def getTimeoutAsTime(config: Configuration): Time = {
@@ -750,11 +750,11 @@ object AkkaUtils {
   }
 
   def getLookupTimeout(config: Configuration): time.Duration = {
-    return TimeUtils.parseDuration(config.getString(AkkaOptions.LOOKUP_TIMEOUT))
+    TimeUtils.parseDuration(config.getString(AkkaOptions.LOOKUP_TIMEOUT))
   }
 
   def getClientTimeout(config: Configuration): time.Duration = {
-    return TimeUtils.parseDuration(config.getString(AkkaOptions.CLIENT_TIMEOUT))
+    TimeUtils.parseDuration(config.getString(AkkaOptions.CLIENT_TIMEOUT))
   }
 
   /** Returns the address of the given [[ActorSystem]]. The [[Address]] object contains
--- a/flink-runtime/src/main/scala/org/apache/flink/runtime/akka/AkkaUtils.scala
+++ b/flink-runtime/src/main/scala/org/apache/flink/runtime/akka/AkkaUtils.scala
@@ -21,7 +21,7 @@ package org.apache.flink.runtime.akka
 import java.io.IOException
 import java.net._
 import java.time
-import java.util.concurrent.{Callable, CompletableFuture, TimeUnit}
+import java.util.concurrent.{Callable, CompletableFuture}
 
 import akka.actor._
 import akka.pattern.{ask => akkaAsk}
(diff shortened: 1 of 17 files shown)


flink-runtime/src/main/scala/org/apache/flink/runtime/akka/AkkaUtils.scala
```diff
@@ -49,8 +51,6 @@ import scala.language.postfixOps
 object AkkaUtils {
```
zhuzhurk: The parsing of Akka configs are not replaced by TimeUtils.parseDuration(..), since they are directly used for ActorSystem creation.
The only exception is "akka.ask.timeout" which is also widely used as Flink RPC timeout config. So I parsed it with TimeUtils when it is used by Flink components and keep it parsed by scala Duration when used by Akka (in AkkaUtils#getRemoteAkkaConfig).
The configs "akka.lookup.timeout" and "akka.client.timeout" are actually only used by Flink components so the parsing logics for them are replaced.

flink-runtime/src/main/scala/org/apache/flink/runtime/akka/AkkaUtils.scala
```diff
@@ -746,38 +744,17 @@ object AkkaUtils {
...
-    val duration = Duration(AkkaOptions.LOOKUP_TIMEOUT.defaultValue())
-    new FiniteDuration(duration.toMillis, TimeUnit.MILLISECONDS)
-  }
-
-  def getClientTimeout(config: Configuration): FiniteDuration = {
-    val duration = Duration(config.getString(AkkaOptions.CLIENT_TIMEOUT))
-
-    new FiniteDuration(duration.toMillis, TimeUnit.MILLISECONDS)
+  def getLookupTimeout(config: Configuration): time.Duration = {
+    return TimeUtils.parseDuration(config.getString(AkkaOptions.LOOKUP_TIMEOUT))
```
tillrohrmann: same here with `return`

flink-runtime/src/main/scala/org/apache/flink/runtime/akka/AkkaUtils.scala
```diff
@@ -746,38 +744,17 @@ object AkkaUtils {
...
   }
 
-  def getDefaultClientTimeout: FiniteDuration = {
-    val duration = Duration(AkkaOptions.CLIENT_TIMEOUT.defaultValue())
-
-    new FiniteDuration(duration.toMillis, TimeUnit.MILLISECONDS)
+  def getClientTimeout(config: Configuration): time.Duration = {
+    return TimeUtils.parseDuration(config.getString(AkkaOptions.CLIENT_TIMEOUT))
```
tillrohrmann: and here

flink-runtime/src/main/scala/org/apache/flink/runtime/akka/AkkaUtils.scala
zhuzhurk: I gave it a thought again and I think we can replace them.

The concern was that scala Duration supports infinite values like Inf, +Inf, etc. Replacing the parsing may break jobs specifying infinite values. However, since we never mentioned infinite durations in the config descriptors, I think we can treat them as finite durations and throw errors for infinite values.

Besides that, do you think we should parse the other duration configs (e.g. "akka.tcp.timeout") which were treated as string and directly set to Akka configs? Parsing them can help to do an early check, rather than expecting Akka to fail due to bad configs.

flink-runtime/src/main/scala/org/apache/flink/runtime/akka/AkkaUtils.scala
zhuzhurk: done.

flink-runtime/src/main/scala/org/apache/flink/runtime/akka/AkkaUtils.scala
```diff
@@ -728,15 +728,13 @@ object AkkaUtils {
     }
   }
 
-  def getTimeout(config: Configuration): FiniteDuration = {
-    val duration = Duration(config.getString(AkkaOptions.ASK_TIMEOUT))
-
-    new FiniteDuration(duration.toMillis, TimeUnit.MILLISECONDS)
+  def getTimeout(config: Configuration): time.Duration = {
+    return TimeUtils.parseDuration(config.getString(AkkaOptions.ASK_TIMEOUT))
```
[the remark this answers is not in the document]
zhuzhurk: done.

flink-runtime/src/main/scala/org/apache/flink/runtime/akka/AkkaUtils.scala
zhuzhurk: A hotfix is added to convert akka duration configs to akka recognizable duration format.
Given that "these two mechanism don't need to stay in sync in the future" and TimeUtils currently supports non-labeled config, I think the hotfix is necessary.

PR #9748, opened 2019-09-23
[FLINK-14016][python][flink-table-planner] Introduce DataStreamPythonCalc for Python function execution

dianfu:

*This pull request introduces DataStreamPythonCalc for Python ScalarFunction execution.*

## Brief change log

  - *Introduces DataStreamPythonCalc for Python ScalarFunction execution*
  - *Introduces PythonScalarFunctionSplitRule which split multiple ScalarFunctions contained in the same node* 
  - *Add a new phase for Python related rules*

## Verifying this change

  - *Added test PythonScalarFunctionSplitRuleTest*

  - The serializers: (no)

## Documentation


flink-table/flink-table-common/src/main/java/org/apache/flink/table/functions/FunctionLanguage.java
```diff
@@ -0,0 +1,32 @@
...
+
+import org.apache.flink.annotation.PublicEvolving;
+
+/**
+ * Categorizes the language of a {@link FunctionDefinition}.
+ */
+@PublicEvolving
+public enum FunctionLanguage {
+
+	JAVA,
```
sunjincheng121: JVM

sunjincheng121:
Thanks a lot for your contribute @dianfu . I have gone through the PR. My main comments are below:
- For the logical node, I think we don't need to create a new node for the python, instead, we can reuse the current `FlinkLogicalCalc` node. In this way, not only we can reuse most of the code, but also I think it matches the semantic that it is a Calc.
- For the physical node, we can create a `DataStreamCalcBase` node and let both `DataStreamPythonCalc` and `DataStreamCalc` extend it.
- For the split rule, maybe we can further split the calc a bit, i.e., don't mix java and python UDFs rather than further split them in translateToPlan. Doing all split in the split rule is more clear.
- For the filter, we calculate these condition UDFs together with other UDFs and do the filter later. I think we can optimize it a bit, i.e., calculate the conditions first and then check whether to call the other UDFs. This can be easily achieved in the SplitRule.

Last for the new optimization phase, I agree with the current design. Another choice is using the current logic optimization phase, but I think it would bring some side effects 
like a) The PythonSplitRule would conflict with the CalcMerge rules and we need to do some hack for the merge rules. b) Adding all rules in logical phase makes the time of optimization longer and longer. This is the reason that we have a lot of Rule based optimization phase in Blink.

What do you think?

dianfu:
@sunjincheng121 I have updated the PR and have addressed your comments with just one exception. 

Regarding to the filter part as I found that it's a big optimization and deserve a separate rule. The functionality of PythonScalarFunctionSplitRule which is introduced in this PR focuses on separating the Python UDFs from the Java UDFs mixed in a Calc node. For the filter optimization, we need introduce a new rule which is responsible for splitting a Calc node into two Calc nodes if it contains condition and the projection contains Python UDFs. The generated bottom Calc contains the filter part and the top Calc contains the projection part. 

Does it make sense to you? If yes, I will create a follow up ticket for this optimization.

dianfu:
@sunjincheng121 As I discussed offline with you, regarding to the filter optimization, thanks for your detail explanation, we are on the same page now. We can improve it in a following PR. I have created a JIRA [FLINK-14202](https://issues.apache.org/jira/browse/FLINK-14202) to track it.

sunjincheng121:
Thanks Dian, we can tracing this filter issue in the  FLINK-14202 as this PR is blocker of the follows PRs.  Good job!

flink-table/flink-table-planner/src/main/scala/org/apache/flink/table/plan/nodes/datastream/DataStreamPythonCalc.scala
```diff
@@ -0,0 +1,173 @@
...
+      .map {
+        case pythonCall: RexCall =>
+          val inputRef = new RexInputRef(input.getRowType.getFieldCount + idx, pythonCall.getType)
+          idx += 1
+          inputRef
+        case node => node
+      }
+  }
+
+  override def translateToPlan(
```
hequn8128: Maybe we can remove the implementation of this method to the next PR. This method can not be tested yet. It's hard to make sure the correctness of these codes. What do you think?

dianfu:
@hequn8128 Thanks a lot for your review. Very appreciated. I have updated the PR accordingly. Regarding to adding tests for translateToPlan, it's not possible for now. So I'm ok to remove it for now. Will add it back in #9766 which will add all the remaining necessary building blocks to make sure Python UDF end to end tests could run.

flink-table/flink-table-planner/src/main/scala/org/apache/flink/table/plan/rules/logical/PythonScalarFunctionSplitRule.scala
```diff
@@ -0,0 +1,211 @@
...
+  override def visitNode(rexNode: RexNode): RexNode = rexNode
+}
+
+object PythonScalarFunctionSplitRule {
+  val INSTANCE: RelOptRule = new PythonScalarFunctionSplitRule
+
+  /**
+    * Extracts the indices of the input fields referred by the specified projects and condition.
+    */
+  def extractRefInputFields(
```
hequn8128: Put this method into the `PythonScalarFunctionSplitRule` class and make it private?

flink-table/flink-table-planner/src/main/scala/org/apache/flink/table/plan/rules/logical/PythonScalarFunctionSplitRule.scala
```diff
@@ -0,0 +1,211 @@
...
+      newProjects, newCondition, extractedFunctionOffset)
+
+    val bottomCalcProjects =
+      accessedFields.map(RexInputRef.of(_, input.getRowType)) ++ extractedRexCalls
+    val bottomCalcFieldNames = SqlValidatorUtil.uniquify(
+      accessedFields.map(i => input.getRowType.getFieldNames.get(i)).toSeq ++
+        extractedRexCalls.indices.map("f" + _),
+      rexBuilder.getTypeFactory.getTypeSystem.isSchemaCaseSensitive)
+
+    val bottomCalc = new FlinkLogicalCalc(
```
hequn8128: It seems we always treat the bottomCalc as the Calc contains python UDFs and the top Calc contains java UDFs. However, in some cases, it would be the opposite, i.e., when outerCallContainsJavaFunction is false.

The error could be reproduced by the following test:
```
  @Test
  def testPythonFunctionAsInputOfJavaFunction(): Unit = {
    val util = streamTestUtil()
    val table = util.addTable[(Int, Int, Int)]("MyTable", 'c, 'a, 'b)
    util.tableEnv.registerFunction("pyFunc1", new PythonScalarFunction("pyFunc1"))
    util.tableEnv.registerFunction("pyFunc2", new PythonScalarFunction2("pyFunc2"))

    val resultTable = table
      .where("pyFunc2(a)")
      .select("pyFunc1(a + 1, b)")

    val expected = unaryNode(
      "DataStreamCalc",
      unaryNode(
        "DataStreamPythonCalc",
        streamTableNode(table),
        term("select", "pyFunc1(a, b) AS f0")
      ),
      term("select", "+(f0, 1) AS _c0")
    )

    util.verifyTable(resultTable, expected)
  }
```

dianfu: Good catch. Fixed.

dianfu:
@hequn8128 Thanks for your review. Updated.

flink-table/flink-table-planner/src/test/scala/org/apache/flink/table/plan/PythonScalarFunctionSplitRuleTest.scala
```diff
@@ -0,0 +1,238 @@
...
+  override def getResultType(signature: Array[Class[_]]): TypeInformation[_] =
+    BasicTypeInfo.INT_TYPE_INFO
+
+  override def getLanguage: FunctionLanguage = FunctionLanguage.PYTHON
+
+  override def toString: String = name
+}
+
+class BooleanPythonScalarFunction(name: String) extends ScalarFunction {
+  def eval(i: Int, j: Int): Int = i + j
```
hequn8128: Change return type to Boolean?

dianfu:
@hequn8128 Thanks for the latest comments. Updated.

flink-table/flink-table-common/src/main/java/org/apache/flink/table/functions/FunctionDefinition.java
```diff
@@ -40,6 +40,13 @@
 	 */
 	FunctionKind getKind();
 
+	/**
+	 * Returns the language of function this definition describes.
+	 */
+	default FunctionLanguage getLanguage() {
```
twalthr: @hequn8128 I just had a brief look at this change. I already mentioned my concern on the ML discussion. Why should users be able to override this property. `UserDefinedFunction` should mark this method `final` similar to `ScalarFunction#getKind`. What do you think?

dianfu: @twalthr Thanks a lot for your comments. Appreciated! Just share my thoughts as following:
1. `Python ScalarFunction` also extends `ScalarFunction` and it will override `getLanguage`. If it's declared as final, there is no way to override it any more. This can be seen from the implementation of [PythonFunctionCodeGenerator](https://github.com/apache/flink/blob/fc4a1df1ad48224ae846fbbdeeb21b4b7b164e02/flink-table/flink-table-planner/src/main/scala/org/apache/flink/table/codegen/PythonFunctionCodeGenerator.scala#L74).
2. Actually users could write a user-defined Python `ScalarFunction` with the `ScalarFunction` interface if they want(In this case, they need to override the getLanguage interface). [PythonFunctionCodeGenerator](https://github.com/apache/flink/blob/fc4a1df1ad48224ae846fbbdeeb21b4b7b164e02/flink-table/flink-table-planner/src/main/scala/org/apache/flink/table/codegen/PythonFunctionCodeGenerator.scala#L74) could be seen as an example how to do that, e.g. they need just implement the `PythonFunction` interface and wrap the serialized Python function inside it.

Does it make sense to you? Looking forward to your feedback.

twalthr: From an architectural point of view, I'm unsure if a Python function should be a `UserDefinedFunction`. Because it is more a system bridging function for executing Python UDFs.

As far as I can see you already do `implements PythonFunction`; isn't that enough information? I still don't understand why users need to be able to override a `getLanguage()` method. When is this method read?

sunjincheng121: +1. If we add the `PythonFunciton` interface for Python, I think `getLanguage()` can be removed.

dianfu: `getLanguage` is used in places where it's needed to determine whether a UDF is Python not not, e.g. in [PythonScalarFunctionSplitRule](https://github.com/apache/flink/blob/2b608e3169ee416c922a8337c2ba5fa873516fde/flink-table/flink-table-planner/src/main/scala/org/apache/flink/table/plan/rules/logical/PythonScalarFunctionSplitRule.scala#L149). As the interface `PythonFunction` has provided enough information for that purpose, so I'm +1 remove it. Have created ticket https://issues.apache.org/jira/browse/FLINK-14342 to track it.

PR #9750, opened 2019-09-24
[FLINK-14139][rest]Fix potential memory leak problem of rest server.

wsry:

The purpose of this pr is to fix the potential memory leak problem of rest server when using session/standalone cluster. More specifically, Flink's rest server uses netty decoder for http request processing and file uploading. However io.netty.handler.codec.http.multipart.DiskAttribute and io.netty.handler.codec.http.multipart.DiskFileUpload class of netty would register some temp files, including post chunks and upload file chunks, to java.io.DeleteOnExitHook which has a potential of memory leak, because the registered file names will never be deleted before the cluster stops. This pr attempts to solve the problem by disabling the registration of temp files to java.io.DeleteOnExitHook and makes Flink's shutdown hook to take the responsibility of temp file clean up.

## Brief change log

  - *The deleteOnExitTemporaryFile filed of io.netty.handler.codec.http.multipart.DiskAttribute and io.netty.handler.codec.http.multipart.DiskFileUpload is set to false to disable netty to register the shutdown hook.*
  - *Set the baseDirectory of io.netty.handler.codec.http.multipart.DiskAttribute to the file upload directory (shares the same directory with file upload), which will be cleaned up by Flink's shutdown hook when jvm exits.*
  - *Add tests to verify that no temp files will be registered to java.io.DeleteOnExitHook after the fix.*

## Verifying this change

  - *Based on the existing FileUploadHandlerTest which will register temp file (both post chunks and uploaded file chunks) to java.io.DeleteOnExitHook, after file uploading, each test case verifies that no file is register to java.io.DeleteOnExitHook. Before the fix, the test will fail and after the fix, the test will pass.*

## Documentation

### DIFF
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/rest/FileUploadHandler.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/rest/FileUploadHandler.java
@@ -38,6 +38,7 @@ import org.apache.flink.shaded.netty4.io.netty.handler.codec.http.HttpResponseSt
 import org.apache.flink.shaded.netty4.io.netty.handler.codec.http.LastHttpContent;
 import org.apache.flink.shaded.netty4.io.netty.handler.codec.http.multipart.Attribute;
 import org.apache.flink.shaded.netty4.io.netty.handler.codec.http.multipart.DefaultHttpDataFactory;
+import org.apache.flink.shaded.netty4.io.netty.handler.codec.http.multipart.DiskAttribute;
 import org.apache.flink.shaded.netty4.io.netty.handler.codec.http.multipart.DiskFileUpload;
 import org.apache.flink.shaded.netty4.io.netty.handler.codec.http.multipart.HttpDataFactory;
 import org.apache.flink.shaded.netty4.io.netty.handler.codec.http.multipart.HttpPostRequestDecoder;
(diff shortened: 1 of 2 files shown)


flink-runtime/src/main/java/org/apache/flink/runtime/rest/RestServerEndpoint.java
```diff
@@ -113,6 +115,15 @@ public RestServerEndpoint(RestServerEndpointConfiguration configuration) throws
...
 		terminationFuture = new CompletableFuture<>();
+
+		// the clean up of temp files when jvm exits is handled by org.apache.flink.util.ShutdownHookUtil; thus,
+		// it's no need to register those files (post chunks and upload file chunks) to java.io.DeleteOnExitHook
+		// which may lead to memory leak.
+		DiskAttribute.deleteOnExitTemporaryFile = false;
+		DiskFileUpload.deleteOnExitTemporaryFile = false;
+
+		// share the same directory with file upload for post chunks storage.
+		DiskAttribute.baseDirectory = uploadDir.normalize().toAbsolutePath().toString();
```
tillrohrmann: Is the `DiskAttribute` and `DiskFileUpload` only used by Flink's `FileUploadHandler`? If yes, would it make sense to move these settings into the constructor of the `FileUploadHandler`? At the moment we also set the `DiskFileUpload.baseDirectory` there. That way it would be easier to understand the context.

wsry: You are right. The DiskAttribute and DiskFileUpload is only used by FileUploadHandler. I will move these settings into the constructor of the FileUploadHandler and update the pr.

wsry: I have updated the pr.

lijiangbo:
Version 1.8 has the same problem. Does this problem converge on the 1.8 branch

PR #9753, opened 2019-09-24
[FLINK-14180][runtime]Enable config of maximum capacity of FileArchivedExecutionGraphStore.

wsry:
The purpose of this pr to enable config of max capacity of FileArchivedExecutionGraphStore. Currently, Flink session cluster uses FileArchivedExecutionGraphStore to keep finished jobs for historic requests. The FileArchivedExecutionGraphStore purges archived ExecutionGraphs only by an expiration time. In a session cluster on which runs many batch jobs, it is hard to config the jobstore.expiration-time, if configured too short, the historical information may have been deleted when the user want to check it, and if configured too long, the web front end may response very slowly when the number of finished job is too large. This pr enables config of maximum capacity of FileArchivedExecutionGraphStore, after which, the expiration time can be set to a relative long value and the maximum capacity can be set to an appropriate value which does not make the web ui become too slow or consume too much memory.

## Brief change log

  - A new config option *jobstore.max-capacity* was added. The default value is *Integer.MAX_VALUE*, which means the capacity of job store is not limited and keeps the default behavior unchanged.
  - Set the maximum capacity of job detail cache through the *maximumSize* interface of Guava Cache.
  - Add a test case to verify the change.

## Verifying this change

  - A ut case was added which works by putting more than the configured capacity number of the finished ExecutionGraphs to FileArchivedExecutionGraphStore and verifying that the size of the FileArchivedExecutionGraphStore is no more than the configured max capacity and the old execution graphs will be purged if the total added execution graph number exceeds the max capacity.

## Documentation

### DIFF
--- a/flink-runtime/src/main/java/org/apache/flink/runtime/dispatcher/FileArchivedExecutionGraphStore.java
+++ b/flink-runtime/src/main/java/org/apache/flink/runtime/dispatcher/FileArchivedExecutionGraphStore.java
@@ -83,6 +83,7 @@ public class FileArchivedExecutionGraphStore implements ArchivedExecutionGraphSt
 	public FileArchivedExecutionGraphStore(
 			File rootDir,
 			Time expirationTime,
+			int maximumCapacity,
 			long maximumCacheSizeBytes,
 			ScheduledExecutor scheduledExecutor,
 			Ticker ticker) throws IOException {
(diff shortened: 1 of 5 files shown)


flink-runtime/src/main/java/org/apache/flink/runtime/dispatcher/FileArchivedExecutionGraphStore.java
```diff
@@ -102,6 +103,7 @@ public FileArchivedExecutionGraphStore(
 			"The storage directory must exist and be a directory.");
 		this.jobDetailsCache = CacheBuilder.newBuilder()
 			.expireAfterWrite(expirationTime.toMilliseconds(), TimeUnit.MILLISECONDS)
+			.maximumSize(maximumCapacity)
```
KarmaGYZ: Since the type of the origin parameter is `long`, we could change it accordingly.

tillrohrmann: I think the int value range should be big enough. If this should become a problem we can change it later.

PR #9756, opened 2019-09-24
[backport to 1.9][FLINK-14145] Fix getLatestCheckpoint(true) returns wrong checkpoint

gyfora:
Backporting of the fix that is already merged to master


wuchong:
Hi @TisonKun , @1u0 , @gyfora  do we have a consensus whether backport this to 1.9?

TisonKun:
@wuchong this pull request fixes the bug where the CompletedCheckpointStore would not return the last checkpoint correctly if checkpoints are preferred. It is definitely a bug fix and I think we are able to include this in a bugfix version.

gyfora:
I agree, this is a critical fix, that fixes a currently completely broken feature.

I have merged this to 1.9 and closing the PR

wuchong:
Thanks @gyfora for merging this.

PR #9758, opened 2019-09-24
[FLINK-14185][tests]Move unstable assertion statement out of close method of QS test server.

wsry:

The purpose of this pr is to fix the unstable test case AbstractServerTest. The test case failed on Travis. The failure reason is that the requestStats is shared by two instance of TestServer which relies on AutoCloseable interface for connection release. If one of the server has been shut down and the anther is not, the assertion in the close method can fail. This pr fix the problem by moving the assertion out of the close method.

## Brief change log

  - The unstable assertion statement was moved out of the close method of TestServer.

## Verifying this change

The change can be verified by the test case itself.

## Documentation

### DIFF
--- a/flink-queryable-state/flink-queryable-state-runtime/src/test/java/org/apache/flink/queryablestate/network/AbstractServerTest.java
+++ b/flink-queryable-state/flink-queryable-state-runtime/src/test/java/org/apache/flink/queryablestate/network/AbstractServerTest.java
@@ -86,8 +86,8 @@ public class AbstractServerTest extends TestLogger {
 	@Test
 	public void testPortRangeSuccess() throws Throwable {
 
-		// this is shared between the two servers.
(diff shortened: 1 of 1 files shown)


zentol:
What if we didn't share the stats between servers?

wsry:
> What if we didn't share the stats between servers?

@zentol From the previous comment and assertion statement below,
```
// the client connects to both servers and the stats object is shared.
Assert.assertEquals(2L, serverStats.getNumConnections());
```
I guess that the test case makes the assumption of shared stats, though, which can also be changed.
Which way of fix do you prefer? Don't share stats and change the above assertion to
```
Assert.assertEquals(1L, serverStats.getNumConnections());
```
or move the unstable assertion statement out of close method as what this pr does. Both are acceptable for me.

